PR #1328, opened 2020-10-22
Add support for model level dataclass/config class

titu1994:
# Changelog
- Adds ModelPT level support for Dataclass and infrastructure to merge dataclasses with OmegaConf/Hydra overrides
- **[Prototype phase]**

# PR overview
 - At the model level, enable potential signature `ModelPT(cfg=Union[OmegaConf, ModelPTConfig], ...).
 - Support hybrid override approach - dataclass for base config, hydra/omegaconf for user provided overrides.
 - Explore potential drawbacks of usage
 - Explore interop with Hydra.

# Examples
- A test case which creates a Model config and then updates with existing model.cfg attributes.
- Two scripts which explore hybrid Dataclass + yaml overrides and full structured config workflow
   - For hybrid workflow, notice reduced number of config parameters (some more can potentially be removed if we dont want to allow users to override easily using Hydra).
   - For full structured config workflow, shows the severe overhead in code to describe model architecture and supporting components etc.

# Limitations
 - This PR does not explore / operate on Hydra's ConfigStore. It uses base dataclass and interops at OmegaConf level only. This is by design.
 - ConfigStore does not interop with the hybrid approach, and it is not practical to provide entire config of model etc from command line. 
 - A middle ground might eventually be possible with a signficantly concretized version of `EncDecCTCModelConfig` - say as a `Quartznet15x5CTCConfig`, which described explicitly the entire model architecture and most of the supporting components, but all in code.
  - This is very high cost to develop for all variants of the Model. Imagine Jasper10x5, Jasper5x5, QN15x5, QN5x5 etc and all that code.


nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,173 @@
...
+    mag_power: float = 2.0
+
+
+@dataclass
+class SpecAugmentConfig:
+    _target_: str = "nemo.collections.asr.modules.SpectrogramAugmentation"
+    freq_masks: int = 0
+    time_masks: int = 0
+    freq_width: int = 0
+    time_width: Optional[Any] = 0
```
titu1994, line 61: Union is not supported by Hydra/OmegaConf

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,173 @@
...
+    stride: List[int] = MISSING
+    dilation: List[int] = MISSING
+    dropout: float = MISSING
+    residual: bool = MISSING
+
+    # Optional arguments
+    groups: int = 1
+    separable: bool = False
+    heads: int = -1
+    residual_mode: str = "add"
```
titu1994, line 108: Code would normally switch automatically from add to whatever was set by default in the ConvASREncoder constructor, however as we are now passing (by default) a value, it is upto user to explicitly set the value when updating the constructor

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,173 @@
...
+    se_context_size: int = -1
+    se_interpolation_mode: str = 'nearest'
+    kernel_size_factor: float = 1.0
+    stride_last: bool = False
+
+
+@dataclass
+class ConvASREncoderConfig:
+    _target_: str = 'nemo.collections.asr.modules.ConvASREncoder'
+    jasper: Optional[JasperEncoderConfig] = field(default_factory=list)
```
titu1994, line 121: List[JasperEncoderConfig] is more appropriate, but cannot be handled by OmegaConf at the current version i am trying, maybe once we bump omegaconf this issue goes away.

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,173 @@
...
+    activation: str = MISSING
+    feat_in: int = MISSING
+    normalization_mode: str = "batch"
+    residual_mode: str = "add"
```
titu1994, line 125: Changing residual mode default here will *not* be reflected in the residual mode at the JasperEncoderConfig level! Code must be changed to support this (can be done easily).

nemo/collections/asr/models/ctc_models.py
```diff
@@ -193,8 +193,14 @@ def change_vocabulary(self, new_vocabulary: List[str]):
                 raise ValueError(f'New vocabulary must be non-empty list of chars. But I got: {new_vocabulary}')
             decoder_config = self.decoder.to_config_dict()
             new_decoder_config = copy.deepcopy(decoder_config)
-            new_decoder_config['params']['vocabulary'] = new_vocabulary
-            new_decoder_config['params']['num_classes'] = len(new_vocabulary)
+
+            if 'params' in new_decoder_config:
```
titu1994, line 197: Necessary to update to `_target_` while maintaining backward compat.

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -141,7 +141,7 @@ def __init__(
             residual_mode = lcfg.get('residual_mode', residual_mode)
             se = lcfg.get('se', False)
             se_reduction_ratio = lcfg.get('se_reduction_ratio', 8)
-            se_context_window = lcfg.get('se_context_window', -1)
+            se_context_window = lcfg.get('se_context_size', -1)
```
titu1994, line 144: All configs use se_context_size... but code checks se_context_window. Thankfully no major experiments were done with limited context. [BUGFIX]

nemo/core/classes/modelPT.py
```diff
@@ -60,6 +62,8 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
 
             trainer (Optional): Pytorch Lightning Trainer instance
         """
+        if is_dataclass(cfg) and isinstance(cfg, ModelPTConfig):
```
titu1994, line 65: Only line really needed to support Data classes. After this, they are treated as OmegaConf / DictConfig object so no other code changes necessary. Serialization and deserialization is unaffected beyond this point.

nemo/core/config/modelPT.py
```diff
@@ -0,0 +1,92 @@
...
+    pin_memory: bool = True
+
+
+@dataclass
+class BaseOptimConfig:
+    pass
+
+
+@dataclass
+class BaseSchedConfig:
```
titu1994, line 49: Providing base classes is not necessary, artifact of earlier attempts. (All subclasses need to override the class type in their signature anyway)

nemo/core/config/modelPT.py
```diff
@@ -0,0 +1,92 @@
...
+    run: Dict[str, Any] = field(default_factory=lambda: {"dir": "."})
+    job_logging: Dict[str, Any] = field(default_factory=lambda: {"root": {"handlers": None}})
+
+
+@dataclass
+class ModelPTConfig:
+    name: str = MISSING
+    model: ModelConfig = MISSING
+    trainer: config.TrainerConfig = config.TrainerConfig()
+    exp_manager: Optional[Any] = exp_manager.ExpManagerConfig()
```
titu1994, line 91: I dont know why but `exp_manager: exp_manager.ExpManagerConfig = exp_manager.ExpManagerConfig()` throws errors, whereas this doesnt.

nemo/utils/config_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+import copy
+from dataclasses import is_dataclass
+
+from omegaconf import DictConfig, OmegaConf, open_dict
+
+from nemo.core.config.modelPT import ModelPTConfig
+
+
+def update_model_config(model_cls: ModelPTConfig, update_cfg: DictConfig):
```
titu1994, line 23: [TODO] document this helper. Simply called by ModelPT to smartly merge the ModelPT dataclass with the "update" class. 

Question - should we provide a flag to avoid dropping the dataloaders / optim subconfigs if the "update_cfg" doesnt have those sections but the ModelPT dataclass has them ? Might be useful / harmful. Needs discussion.

tests/collections/asr/test_asr_ctcencdec_model.py
```diff
@@ -114,3 +110,48 @@ def test_vocab_change(self, asr_model):
...
+        ]
+
+        # Update decoder
+        model_cfg.model.decoder.feat_in = 1024
+        model_cfg.model.decoder.num_classes = 28
+        model_cfg.model.decoder.vocabulary = vocabulary
+
+        # Construct the model
+        asr_cfg = OmegaConf.create({'model': asr_model.cfg})
+        model_cfg = EncDecCTCModel.update_model_dataclass(model_cfg, asr_cfg)
```
titu1994, line 146: This line is just to test the function, we can just pass asr_cfg into EncDecCTCModel and it gives same thing (but with the 3 data loaders which need to have their manifest filepath set to None explicitly + the optim/sched section).

titu1994 (commented):
Comments for reviewers

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,164 @@
...
+from typing import Any, Dict, List, Optional
+
+from omegaconf import MISSING, OmegaConf
+
+from nemo.core.config import modelPT as model_cfg
+
+
+@dataclass
+class BaseASRPreprocessorConfig:
+    pass
```
blisc, line 25: Isn't this redundant?

nemo/collections/asr/modules/conv_asr.py
blisc: Well now that we have dataclasses, can we remove the .get()? I assume they now have smart defaults as opposed to not existing.

blisc (commented):
Minor feedback.
Can you show an example use case of how to use the new dataclasses?
Is there any particular reason, _target_ was added to the dataclasses?

nemo/collections/asr/models/configs/ctc_models_config.py
titu1994: Yes, all the Base* classes are redundant. We can have a placeholder type with Any instead of them, but in hindsight creating empty base classes just is not a good experience so I'll be dropping them in my next iteration.

nemo/core/classes/modelPT.py
```diff
@@ -981,6 +985,44 @@ def cfg(self, cfg):
...
+        and used to override the default values set in the dataclass.
+
+        # Usage
+        @hydra_runner(config_path="conf", config_name="config")
+        def main(cfg):
+            # Create the data class with valid defaults
+            model_dataclass = ModelPTConfig()
+
+            # Override the defaults with hydra / external DictConfig which mirrors the structure
+            cfg = ModelPT.update_model_dataclass(model_dataclass, cfg)
```
ericharper, line 1004: Why do we need these extra steps? I was thinking that our example scripts would stay exactly the same, and the only thing that would be changing for users is there would be less configurations in the config files as everything needed (except paths to data, etc) would be configured from the dataclass.

nemo/collections/asr/modules/conv_asr.py
titu1994: No we cannot remove get. This is by design - 

1) We don't want to refactor old classes to support dataclasses. They are optional by design and dataclasses must interop with existing code, not the other way around. 

Since dataclasses are optional, users are still freely allowed to use yaml based configs, not just as overrides but as full configs. And therefore the get based methods are more stable c

Of course, if you use dataclasses, your yaml config gets validated during config merge.  So that's a useful benefit to having dataclasses. 

2) Updating all of our codebase to handle dataclasses as a special case required a massive effort. Instead, dataclasses are mapped down to dict configs immediately after they are passed to ModelPT so that they interop perfectly with existing code.

nemo/core/config/modelPT.py
ericharper: That's probably worth figuring out.

titu1994 (commented):
@blisc Dataclasses will be used for two primary purposes 

1) Default model config that users can instantiate, update then use to instantiate models all from code without touching yaml at all. 

2) Model configs that provide the global set of parameters and are then updated with user specific overrides - either via Hydra or via a manually construct of a Dictconfig (say from ModelPT.cfg) as shown in the test case. 

3) As explicit validation of Config dicts, passed via Hydra or explicit Dictconfig. 

As to an example, yes I will write a temporary file that will be removed after review showing the above cases.

titu1994:
@blisc  `_target_` is required as a validation. The config merge explicitly fails when trying to merge with the yaml override config otherwise. 

As to why I wrote the class explicitly in string, that can be corrected by import + helper function. I will do that in next iteration.

ericharper (commented):
Are you planning on using Hydra's Structured Config features? Mainly the `ConfigStore`?
```
import hydra
from hydra.core.config_store import ConfigStore
```
Please add some documentation to the README of this PR so that we know what the PR is trying to achieve.

nemo/collections/asr/models/configs/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+from nemo.collections.asr.models.configs.ctc_models_config import (
+    AudioToMelSpectrogramPreprocessorConfig,
+    ConvASRDecoderConfig,
+    ConvASREncoderConfig,
+    EncDecCTCConfig,
+    EncDecCTCDatasetConfig,
+    EncDecCTCModelConfig,
+    JasperEncoderConfig,
+    SpecAugmentConfig,
+)
```
tkornuta-nvidia, line 24: What are the advantages of having:
* a separate file with model configs VS
* specific model config placed in the same file along with the given model

IMO the latter is cleaner, as params as "closer" to model, and whenever you write "from fileX import ModelX" you can change that to "from fileX import ModelX, ModelXConfig".

What do you think?

nemo/collections/asr/models/configs/ctc_models_config.py
tkornuta-nvidia: Yap, IMO this is a big issue, discussed that several times with Omry, seems the only thing stopping him from adding unions is... lack of time:] So hope to see that sometime soon...

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,164 @@
...
+    se_reduction_ratio: int = 8
+    se_context_size: int = -1
+    se_interpolation_mode: str = 'nearest'
+    kernel_size_factor: float = 1.0
+    stride_last: bool = False
+
+
+@dataclass
+class ConvASREncoderConfig:
+    _target_: str = 'nemo.collections.asr.modules.ConvASREncoder'
```
tkornuta-nvidia, line 120: why _target_ is present here and is absent in the configs above?

tkornuta-nvidia: is it because the above are configs for "abstract classes" or sth?

nemo/collections/asr/models/configs/ctc_models_config.py
tkornuta-nvidia: Both thumbs up for this - and generally from moving all defaults from models/datasets to the associated configs

nemo/core/config/modelPT.py
```diff
@@ -0,0 +1,92 @@
...
+class ModelConfig:
+    """
+
+    """
+
+    # ...
+    train_ds: Optional[BaseDatasetConfig] = None
+    validation_ds: Optional[BaseDatasetConfig] = None
+    test_ds: Optional[BaseDatasetConfig] = None
+    optim: Optional[BaseOptimConfig] = None
```
tkornuta-nvidia, line 77: 👍  for optional :)

nemo/core/config/modelPT.py
tkornuta-nvidia: Hmmm, interesting...

nemo/utils/config_utils.py
tkornuta-nvidia: Yeah, generally I think it would be good if we could discuss this function and how should if work in all if/else cases...

nemo/core/classes/modelPT.py
tkornuta-nvidia: What is the difference between `update_model_dataclass` and `update_model_config`? Aren't those two covering the same functionality?

nemo/utils/config_utils.py
tkornuta-nvidia: What is the difference between `update_model_dataclass` and `update_model_config`? Aren't those two covering the same functionality?

nemo/collections/asr/models/configs/ctc_models_config.py
titu1994: @tkornuta-nvidia This is actually *not good*. By default, in Citrinet, I dont add residual mode (therefore it defaults to global default mode of "add"), but in certain places, I am required to override "add" with "stride_add" (if I remember correctly). In yaml, this is perfectly fine and very convenient.

However in Structured Configs, I would have to explicitly update not only the global one here, but every single local one within the list manually. Its a dissapointing limitation to not be able to map from a global to a local inside a dataclass.

nemo/collections/asr/models/ctc_models.py
titu1994: Good point.

nemo/core/config/modelPT.py
titu1994: Tried many things, cant seem to fix it. Nothing seems to be wrong, and I can just as well use the syntax `trainer: config.TrainerConfig = config.TrainerConfig()` just above so dont know whats wrong here.

nemo/utils/config_utils.py
titu1994: I didn't want to clutter up ModelPT + I want to allow users to use this function without being reliant on ModelPT. So they are split and ModelPT simply forwards the call.

nemo/collections/asr/models/configs/ctc_models_config.py
titu1994: It is required when merging with yaml config + Cant have actual "class" without instantiation inside of a dataclass. Ie cant write `_target_ = ConvASREncoder` since `ConvASREncoder` is a class, not an object, and type inference of dataclass doesnt work on class level but on object level

nemo/collections/asr/models/configs/ctc_models_config.py
tkornuta-nvidia: If you have hardcoded override of a parameter in the code, then maybe it shouldn't be a configuration parameter?

I still don't understand what is "not good" in having defaults in Structured Configs...

nemo/core/config/modelPT.py
tkornuta-nvidia: yeah, I know - this is how I am using TranerConfig. What is so different in ExpManagerConfig, not understand that either...

nemo/collections/asr/models/configs/__init__.py
titu1994: Model configs are generally very large, id always keep them entirely seperate from the code itself. We can perform a similar `from fileX import ModelX, ModelXConfig` simply by importing the appropriate config inside the appropriate Model file (with a from import) so that it is accessible to users via the same syntax.

nemo/collections/asr/models/configs/ctc_models_config.py
titu1994: Its useful and required to be overriden from yaml / code. Structured configs should adapt to support our use cases, not the other way around. Thats my opinion on this.

nemo/collections/asr/models/ctc_models.py
titu1994: Done

nemo/collections/asr/parts/perturb.py
```diff
@@ -724,6 +724,9 @@ class CustomPerturbation(perturb.Perturbation):
             disabled and a warning is logged.
     Returns: AudioAugmentor object
     """
+    if augmenter is None:
```
titu1994, line 727: Necessary patch to support `augmentor=None` case from Structured configs 🙃

nemo/core/classes/modelPT.py
titu1994: Updated documentation, but basically `update_model_dataclass ` is a forwarding of the `update_model_config` for easy API. Can be changed if it seems we are polluting the ModelPT namespace.

nemo/core/config/modelPT.py
titu1994: It even breaks autocomplete ... I dont know whats going on with this. Some issue at python dataclasses level, independent of Hydra / Omegaconf.

titu1994 (commented):
Updated code with 2 examples

titu1994:
@ericharper @tkornuta-nvidia @blisc Please find two example scripts with proposed hybrid and full config structure. Also updated readme.

examples/asr/structured/speech_to_text_structured.py
```diff
@@ -0,0 +1,165 @@
...
+    cfg.model.optim.name = 'novograd'
+    cfg.model.optim.lr = 0.01
+
+    cfg.model.optim.betas = [0.8, 0.5]
+    cfg.model.optim.weight_decay = 0.001
+
+    # sched
+    cfg.model.optim.sched.name = "CosineAnnealing"
+    cfg.model.optim.sched.warmup_steps = None
+    cfg.model.optim.sched.warmup_ratio = 0.01
```
titu1994, line 147: I forgot to mention it in the code, but theres an issue here. All of the "additional" configs dont actually get updated and provided to the ModelPT engine for optim and sched !

Ie if you run the code and look at the logs, youll see that `betas` is [0.95, 0.98] (the default for Novograd) and `weight_decay` is 0. Same for sched - `warmup_steps` and `warmup_ratio` are not provided as args.

```python
[NeMo I 2020-10-22 17:31:14 modelPT:587] Optimizer config = Novograd (
    Parameter Group 0
        amsgrad: False
        betas: (0.95, 0.98)
        eps: 1e-08
        grad_averaging: False
        lr: 0.01
        weight_decay: 0
    )
[NeMo I 2020-10-22 17:31:14 lr_scheduler:554] Scheduler "<nemo.core.optim.lr_scheduler.CosineAnnealing object at 0x7f7904106590>" 
    will be used during training (effective maximum steps = 148) - 
    Parameters : 
    (min_lr: 0.0
    last_epoch: -1
    max_steps: 1
```

examples/asr/structured/conf/quartznet_15x5.yaml
```diff
@@ -0,0 +1,236 @@
+name: &name "QuartzNet15x5"
```
okuchaiev, line 1: let's rename "structured"=>"experimental" for now.

examples/asr/structured/speech_to_text_hybrid.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
okuchaiev, line 1: let's rename "structured"=>"experimental" for now.

nemo/collections/asr/models/configs/__init__.py
okuchaiev: It does not look right to me to have things like: ``JasperEncoderConfig``, ``AudioToMelSpectrogramPreprocessorConfig`` under ``nemo.collections.asr.models.configs``.

Let's only have *model* configs under ``nemo.collections.asr.models.configs.*``
I would put *Module* configs in the same file with its module.

nemo/collections/asr/models/configs/ctc_models_config.py
```diff
@@ -0,0 +1,162 @@
...
+    time_masks: int = 0
+    freq_width: int = 0
+    time_width: Optional[Any] = 0
+    rect_masks: int = 0
+    rect_time: int = 0
+    rect_freq: int = 0
+
+
+@dataclass
+class EncDecCTCDatasetConfig(model_cfg.DatasetConfig):
```
okuchaiev, line 63: keeping module's config's under models.configs does not feel right, any good reason for doing so?

nemo/core/classes/modelPT.py
okuchaiev: I don't understand why this method is necessary. It literally just calls ```update_model_config```.
Some people will do things like:
```
mymodel.update_model_dataclass(whatever)
```
and expect ``mymodel`` to change (and, of course, it won't). I would remove this from ``ModelPT``

nemo/core/classes/modelPT.py
```diff
@@ -1115,6 +1119,49 @@ def cfg(self, cfg):
         self._cfg = cfg
         self._set_hparams(cfg)
```
okuchaiev, line 1120: can we save trainer's param's here as well?

nemo/core/config/modelPT.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import Any, Dict, Optional
+
+from omegaconf import MISSING
+
+from nemo.core import config
+from nemo.utils import exp_manager
+
+
+@dataclass
+class DatasetConfig:
```
okuchaiev, line 25: I think this should live in https://github.com/NVIDIA/NeMo/blob/main/nemo/core/classes/dataset.py

okuchaiev (changes requested):
Please see inline.

nemo/collections/asr/models/configs/ctc_models_config.py
titu1994: `EncDecCTCDatasetConfig` is the base config of the dataset underlying each Model, so it should be a part of the Model config. As to the others, they can be moved to their respective modules.

nemo/core/classes/modelPT.py
titu1994: We dont receive the trainer config inside the constructor, just the Trainer object, so it cant be saved easily. Need to look into it.

examples/asr/experimental/structured/speech_to_text_hybrid.py
```diff
@@ -0,0 +1,56 @@
...
+    asr_model_config = configs.EncDecCTCModelConfig()
+
+    # Merge hydra updates with model config
+    # `drop_missing_subconfig=True` is necessary here. Without it, while the data class will instantiate and be added
+    # to the config, it contains test_ds.sample_rate = MISSING and test_ds.labels = MISSING.
+    # This will raise a OmegaConf MissingMandatoryValue error when processing the dataloaders inside
+    # model_utils.resolve_test_dataloaders(model=self) (used for multi data loader support).
+    # In general, any operation that tries to use a DictConfig with MISSING in it will fail,
+    # other than explicit update operations to change MISSING to some actual value.
+    asr_model_config = update_model_config(asr_model_config, cfg, drop_missing_subconfigs=True)
```
blisc, line 45: Why is this not passed as part of hydra_runner()?
https://github.com/NVIDIA/NeMo/blob/main/nemo/core/config/set_config.py#L26-L28

hydra_runner() already accepts schema as a parameter  which is doing what you are doing here.

blisc (changes requested):

nemo/utils/config_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+import copy
+from dataclasses import is_dataclass
+
+from omegaconf import DictConfig, OmegaConf, open_dict
+
+from nemo.core.config.modelPT import ModelPTConfig
+
+
+def update_model_config(model_cls: ModelPTConfig, update_cfg: DictConfig, drop_missing_subconfigs: bool = True):
```
blisc, line 23: Can we merge these functions into hydra_runner()?

okuchaiev (approved):

examples/asr/experimental/structured/speech_to_text_hybrid.py
titu1994: Ok I have a concrete answer for this question now, now that I attempted to implement the above.
Here is the stack trace (truncated) when attempting the above 

```
Key 'betas' not in 'OptimConfig'
        full_key: model.optim.betas
        reference_type=Optional[OptimConfig]
        object_type=OptimConfig
```

This is one of the prime responsibilities of `update_model_config` - merging a `shell` optim + `shell` sched with a more concrete instantiation from yaml. Even with a perfectly specified Model config (say QN15x5) - if we tie one specific optimizer + scheduler config to it then it cannot be changed by the user - they cannot try another optimizer or another scheduler apart from the one specified by the config class.

The lines inside `update_model_config` which explicitly handle this are https://github.com/NVIDIA/NeMo/pull/1328/files#diff-b383ea91b2a34cbb3404da8fd4275a6ff5b036e0520a5b4491ec644b6e59c9b0R127-R170 

This chunk enables there to be a shell config class for optim and sched, and the actual details regarding any optim and sched inside the yaml. fyi @okuchaiev @ericharper @tkornuta-nvidia

titu1994 (commented):
Important note regarding why the `update_model_config` function cannot be dropped form the hybrid variant - unless we want to force only 1 type of optim + sched to the user and they cannot override it.

examples/asr/structured/speech_to_text_structured.py
titu1994: @ericharper @tkornuta-nvidia @okuchaiev @blisc This is another issue - the only way to override using yaml using `update_model_config`. Added a note about this

examples/asr/experimental/structured/speech_to_text_hybrid.py
blisc: If you think hydra_runner() needs to be updated with your new `update_model_config` function, go right ahead.
My main concern is that we have have an API to do what you want to do, but instead of fixing the old API to work, you invented an entirely new API. Let's work to fix old API as opposed to reinventing the wheel.

Side note: I am certain that omegaconf supports what you are trying to do: https://omegaconf.readthedocs.io/en/latest/usage.html#merging-configurations. If it really isn't working, then our next steps should be to ensure that we are using hydra/omegaconf correctly inside of hydra_runner(). If we are using them correctly, and they still don't work, then we should file an issue on hydra/omegaconf in addition to updating hydra_runner() with `update_model_config`.

titu1994: @blisc Hydra runner was not meant for the purpose of merging Dataclass configs. It is fine iff your dataclass is perfectly matching your hydra overrides.

This is not the case for ModelPT configs. Think about the use case of optimizers alone. If you want to fix your optimizer to just Adam (not even AdamW) and not allow the user to change it at all, then its fine to use hydra as it was designed using the ConfigStore with a AdamConfig.

We dont want that. ModelPT config is generic to all Models - and needs to support optimizers and scheduler switching. Different optimizers can have different (disparate) arguments - SGD + momentum does not have betas. Adam doesnt have nesterov etc. So we need a base set of arguments (just the name and lr) inside the base config.

However this base config cannot be merged with an outer config - the keys betas, nesterov, momentum etc dont exist in the base config. I have tried it (and you can see the result above) and that is why I had to perform the merge operation in such a way.

The hydra_runner is not sufficient to work with this generic ModelPT config workflow at all and simply cannot support it. We can hack around it how much ever we want, but it wont support it cleanly. If you feel otherwise, please try to fix hydra runner to satisfy the requirement of a ModelPT level config that is generic enough to handle such cases.

> If it really isn't working, then our next steps should be to ensure that we are using hydra/omegaconf correctly inside of hydra_runner()

We are using private APIs of Hydra to support hydra_runner() - we are using it incorrectly by design in order to inject additional support for our cases, but this has limits. 

Finally, the only thing I can do to add this flexibility is to introduce this new API and simply call it inside hydra_runner(). But  TLT depends on this - and I do not intend to break the workflow that they have settled on. 

Note this is an experimental API - and is marked as such. You are more than welcome to attempt to modify hydra_runner to support this all on its own with just Hydra level support, and we can surely remove all of the `update_model_config` code at that point if it works.

titu1994: Also- the hydra_runner workflow (as we previously discussed) does not even work for non-script environments. So completely removing `update_model_config` is not an option.

nemo/utils/config_utils.py
okuchaiev: I would keep this function and suggest using it inside hydra_runner. This will make this functionality available for those not wishing to use hydra_runner decorator.

titu1994: Agreed, but won't this affect TLT scripts? I would urge @tkornuta-nvidia to comment on this - can you pull this branch and run TLT tests once after updating hydra runner in order to ensure it has no effect on TLT ?

nemo/utils/config_utils.py
```diff
@@ -0,0 +1,170 @@
...
+            if its mirror sub-config does not exist in the `update_cfg`.
+
+    Returns:
+        The updated DictConfig for the ModelPTConfig
+    """
+    with open_dict(model_cfg.model):
+        # If update config has the key, but model cfg doesnt have the key
+        # Add the update cfg subconfig to the model cfg
+        if subconfig_key in update_cfg.model and subconfig_key not in model_cfg.model:
+            model_cfg.model[subconfig_key] = update_cfg.model[subconfig_key]
```
blisc, line 116: Why do we need to do this? Isn't this exactly what omegaconf.merge does?

nemo/utils/config_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # If update config does not the key, but model cfg has the key
+        # Remove the model cfg subconfig in order to match layout of update cfg
+        if subconfig_key not in update_cfg.model and subconfig_key in model_cfg.model:
+            if drop_missing_subconfigs:
+                model_cfg.model.pop(subconfig_key)
```
blisc, line 122: Wait, why would we want to do this. I thought the point of dataclasses was to enforce a structure for the yaml.

nemo/utils/config_utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+    In such a case, we perform a dual merge. The Optim and Sched Dataclass contain the bare minimum essential
+    components. The extra values are provided via update_cfg.
+
+    In order to enable the merge, we first need to update the update sub-config to incorporate the keys,
+    with dummy temporary values (merge update config with model config). This is done on a copy of the
+    update sub-config, as the actual override values might be overriden by the ModelPTConfig defaults.
+
+    Then we perform a merge of this temporary sub-config with the actual override config in a later step
+    (merge model_cfg with original update_cfg, done outside this function).
```
blisc, line 145: I don't think this is a correct use of dataclasses. For the optimizer, you should only have the bare minimum that is the same across all optimizers parameters. If there is nothing, then it should just be empty. In which case, omegaconf.merge is more than sufficient.

blisc (changes requested):

nemo/utils/config_utils.py
blisc: Nvm

nemo/utils/config_utils.py
blisc: nvm

tkornuta-nvidia (approved):

PR #1339, opened 2020-10-24
IntentSlotClassificationModel exportable

drnikolaev:
Signed-off-by: Sergei Nikolaev <snikolaev@nvidia.com>


tkornuta-nvidia (approved):
The PR looks good, one thing that makes me wonder is all those train/test data.
a) do we need all those samples?
b) I know that this is not binary, but do we want to keep it in here? 
Note that along with 0.11 we have moved all test data to files stored/distributed along with release:
https://github.com/NVIDIA/NeMo/releases/tag/v0.11.0
along with methods added to tests that download them from that URL.

Maybe we should do the same for 1.0, "update" the existing test files to and those files to release? @okuchaiev @blisc @titu1994 @ericharper ?

drnikolaev:
Thanks Tom,
at this moment it's unavoidable, but let's talk to Vlad - the owner

vladgets:
What are the questions to me?

drnikolaev:
How can we instantiate the model without providing so much data?

vladgets:
To make inference with a trained model basically we need only Dictionary of Intent and Slots, to know how to interpret the result numbers. They are 2 small text files stored in the data directory that is used for training. I can also add them inside of Config files, if it is better. For testing of training we need some training data as for any other model.

drnikolaev:
No, just to build a graph and export it to ONNX.

nemo/core/classes/common.py
```diff
@@ -356,6 +356,7 @@ def from_pretrained(
         model_name: str,
         refresh_cache: bool = False,
         override_config_path: Optional[str] = None,
+        override_data_dir: Optional[str] = None,
```
okuchaiev, line 359: why this parameter was changed. Can't we control this via setup_training/validation_data methods?

okuchaiev: Also, data_dir is in config file which can be overridden right now anyways.

okuchaiev (changes requested):
can you please package all "dummy_data" as an artifact so that it gets downloaded with the rest of test data. (ping me offline for how-to).

nemo/core/classes/modelPT.py
```diff
@@ -242,6 +243,8 @@ def restore_from(
                     # If override is top level config, extract just `model` from it
                     if 'model' in conf:
                         conf = conf.model
+                if override_data_dir is not None:
```
okuchaiev, line 246: @vladgets isn't data_dir part of config anyways?

vladgets: In Default config it is '???'. User should define it to refer to his dataset to train the model. We possibly had a strange issue with the model that I shared to Nvidia cloud is that I trained using my local data directory and it was automatically saved as a part of config in a .nemo file. Then when you download this model it refer to my local directory. So to use the pretrained model a user need redefine this parameter to his own data directory.

drnikolaev: Sorry about confusion, it's reverted now

nemo/core/classes/common.py
drnikolaev: yes, just mixed with other use case

okuchaiev (approved):

PR #1345, opened 2020-10-26
Fix CI

blisc:


okuchaiev (changes requested):
Let's just disable this test ( - DALI AudioToMelSpectrogramPreprocessor)  for now.

For it to work, we need DALI 0.28, which will be released at the end of November. The PyTorch container could pick that DALI version for 20.12.

So until we are on 20.12 we can't actually test this.

PR #1348, opened 2020-10-26
Patch to test_asr_metrics

ericharper:
Skip assert when reference is empty since empty reference does not make sense for WER.

### DIFF
--- a/tests/collections/asr/test_asr_metrics.py
+++ b/tests/collections/asr/test_asr_metrics.py
@@ -124,7 +124,9 @@ class TestWordErrorRate:
             n2 = random.randint(1, 512)
             s1 = __randomString(n1)
             s2 = __randomString(n2)
-            assert (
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
LGTM

tests/collections/asr/test_asr_metrics.py
```diff
@@ -124,7 +124,12 @@ def __randomString(N):
             n2 = random.randint(1, 512)
             s1 = __randomString(n1)
             s2 = __randomString(n2)
```
blisc, line 126: Can we force it to not generate an empty string? ie, `while s2.strip: s2 = __randomString(n2);`

blisc: Maybe resample n2 too

okuchaiev (approved):

PR #1362, opened 2020-10-28
Reduce CI time NLP

ekmb:
1. set exp_manager to `null` to reduce CI run time for most of the NLP examples - @VahidooX, @vladgets, @yzhang123  please check if it's ok to do for your scipts
2.Reorganized Megatron-LM tests, now we're testing:
- single GPU megatron loading from config and checkpoint
- 2 GPU model parallel 
- 2 GPU loading from ngc


ericharper (approved):
Thanks!

PR #1364, opened 2020-10-28
Remove experimental automodel

ekmb:


okuchaiev:
is this a fix? should we send this to v1.0.0b2 as well?

ekmb:
> is this a fix? should we send this to v1.0.0b2 as well?

both v1.0.0b2 and main are Ok, it's not a bug just removing these to avoid confusion for users that any model from get_lm() works with all NLP tasks.

okuchaiev:
changed base to v1.0.0b2

PR #1367, opened 2020-10-28
Add patches for Notebooks (Model Primer, ASR)

titu1994:
# Bugfixes
- NeMo Model Primer updated to use PTL 1.x api
- Fix WER calculation for ASR notebook
- Speech commands notebook update to use `-last.ckpt` instead of `-end.ckpt` as `-end.ckpt` no longer is created.

Edit: Colab updated some internal format, hence the large diff. No content has been changed.

### DIFF
--- a/tutorials/01_NeMo_Models.ipynb
+++ b/tutorials/01_NeMo_Models.ipynb
@@ -17,9 +17,7 @@
     {
       "cell_type": "code",
       "metadata": {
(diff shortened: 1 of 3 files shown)


redoctopus (approved):
Thanks!

okuchaiev:
are these fixes to 1.0.0b2? Why is this sent to main?

PR #1371, opened 2020-10-29
Updated ambiguous placeholder text

Fackor:
Line 29 in `examples/nlp/glue_benchmark` says `trainer.gpus="[<CHANGE_TO_GPU_YOU_WANT_TO_USE>]` which is not correct since this parameter sets the number of GPUs to use during training. 

I have updated the placeholder text accordingly to say `trainer.gpus="[<CHANGE_TO_NUMBER_OF_GPUS_TO_USE>]`.


okuchaiev:
No, that flag can accept a list of GPU ids. https://pytorch-lightning.readthedocs.io/en/stable/trainer.html#gpus 
So that placeholder looks right to me.

PR #1372, opened 2020-10-29
Add warning message for modelcheckpoint

blisc:


ericharper (approved):
Thanks so much for this, I've ran into this multiple times already.

PR #1377, opened 2020-10-30
Fix http 500 ngc v1.0.0b2

ericharper:
Try and wait in case NGC download fails.


okuchaiev (approved):
thx!

okuchaiev:
looks like it fails on code formatting though

PR #1405, opened 2020-11-03
Fix smoothed cross entropy loss

ericharper:
`SmoothedCrossEntropyLoss` is expecting log probs but was using logits type and logits name for variables.

Changed logits to log_probs and updated types.


okuchaiev (changes requested):
Can you please also change: nemo.collections.nlp.modules.common.sequence_classifier.SequenceClassifier output_types:
```
    @property
    def output_types(self) -> Optional[Dict[str, NeuralType]]:
        if not self.log_softmax:
            return {"logits": NeuralType(('B', 'D'), LogitsType())}
        else:
            return {"log_probs": NeuralType(('B', 'D'), LogprobsType())}
```

nemo/collections/nlp/modules/common/token_classifier.py
```diff
@@ -91,7 +91,10 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
         """
         Returns definitions of module output ports.
         """
-        return {"logits": NeuralType(('B', 'T', 'C'), LogitsType())}
+        if self.log_softmax == False:
```
okuchaiev, line 94: need to set this (self.log_softmax) in constructor

okuchaiev (approved):
LGTM

requirements/requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.18.2
 onnx>=1.7.0
-pytorch-lightning>=1.0.3
+pytorch-lightning==1.0.4
```
okuchaiev, line 3: why strict equality here?

ericharper: PTL 1.0.5 was causing an exp_manager unit test to fail.

@titu1994 has already created a patch: https://github.com/NVIDIA/NeMo/pull/1407

I can revert it or change it, but I fixed it (at least temporarily) so I could unblock the CI.

PR #1410, opened 2020-11-04
Model save/restore based on EFF.NeMoArchive

tkornuta-nvidia:
Rebased branch, reopened the PR


nemo/core/classes/modelPT.py
```diff
@@ -1218,21 +1131,21 @@ def __unpack_nemo_file(path2file: str, out_folder: str) -> str:
...
 
     @staticmethod
-    def __set_model_restore_state(is_being_restored: bool):
+    def _set_model_restore_state(is_being_restored: bool):
         global _MODEL_IS_RESTORED
         _MODEL_IS_RESTORED = is_being_restored
 
     @staticmethod
-    def set_eff_save(use_eff_save: bool):
+    def _set_eff_save(use_eff_save: bool):
```
titu1994, line 1144: Why are these set/get methods for eff private ?

tests/core/test_fileio.py
```diff
@@ -157,18 +157,25 @@ def test_eff_save_restore_from_nemo_file_encrypted(self, asr_model):
...
+            # Try to restore the encrypted archive (weights) without the encryption key.
+            NeMoArchive.set_encryption_key(None)
             with pytest.raises(PermissionError):
                 # Restore the model.
-                asr_model2 = EncDecCTCModel._eff_restore_from(restore_path=filename)
+                asr_model2 = EncDecCTCModel.restore_from(restore_path=filename)
 
             # Restore the model.
-            asr_model3 = EncDecCTCModel._eff_restore_from(restore_path=filename, encryption_key="test_key")
+            NeMoArchive.set_encryption_key("test_key")
```
titu1994, line 175: Global setter and getter is not multiprocessing safe, but I guess we don't expect to encrypt decrypt on the fly in a DDP enc anyway so it's ok I think ?

titu1994 (commented):
Looks good to me just minor comments

tests/core/test_fileio.py
tkornuta-nvidia: Correct. Note this is just setting the encryption key, before any file opening/closing/saving/restoring happens.
The actual "magic" is happening in restore_from, not here.

nemo/core/classes/modelPT.py
tkornuta-nvidia: I needed to call  _set_model_restore_state/_is_model_being_restored in NeMoArchive restore from "recipe", so had to made them "private" instead of "protected". 

Then I decided to treat all setters/getters in the same way ("scope").

If you think it is important I can make eff_save setters/getters "public" once again.

titu1994: I think the other two protected to private is fine (they kinda never should be used outside of internal code management anyway), but eff should be public. 

Once that's done let's merge

tests/core/test_fileio.py
titu1994: Sounds fine then

nemo/core/classes/modelPT.py
tkornuta-nvidia: done

titu1994 (approved):
LGTM

PR #1415, opened 2020-11-05
Classical Transformer NMT architecture

okuchaiev:
PR adding a classical (from Attention-is-all-you-need paper) NMT model


nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -0,0 +1,213 @@
...
+        into batches to minimize the use of padding tokens. Returns a list of
+        batches where each batch contains indices of sentences included into it
+        """
+
+        # create buckets sorted by the number of src tokens
+        # each bucket is also sorted by the number of tgt tokens
+        buckets = {}
+        for i, src_id in enumerate(src_ids):
+            src_len, tgt_len = len(src_id), len(tgt_ids[i])
+            if src_len not in buckets.keys():
```
MaximumEntropy, line 103: Should be better to do `src_len not in buckets` no?

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
```diff
@@ -0,0 +1,332 @@
...
+                learn_positional_encodings=False,
+            )
+            self.tgt_embedding_layer = self.src_embedding_layer
+        else:
+            self.src_tokenizer = get_tokenizer(**cfg.machine_translation.src_tokenizer)
+            self.tgt_tokenizer = get_tokenizer(**cfg.machine_translation.tgt_tokenizer)
+            super().__init__(cfg=cfg, trainer=trainer)
+            # make vocabulary size divisible by 8 for fast fp16 training
+            src_vocab_size = 8 * math.ceil(self.src_tokenizer.vocab_size / 8)
+            tgt_vocab_size = 8 * math.ceil(self.tgt_tokenizer.vocab_size / 8)
```
MaximumEntropy, line 85: I'm not familiar with this idea for fast fp16 training, but with some very low probability, the model might output an invalid index that does not correspond to any vocab item.

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
```diff
@@ -0,0 +1,332 @@
...
+        return log_probs, beam_results
+
+    def training_step(self, batch, batch_idx):
+        """
+        Lightning calls this inside the training loop with the data from the training dataloader
+        passed in as `batch`.
+        """
+        # forward pass
+        for i in range(len(batch)):
+            if batch[i].ndim == 3:
```
MaximumEntropy, line 184: If batch is a valid tensor, then every `batch[i]` should have the same shape, so why do you check for this? Also, why not just squeeze the batch tensor directly?

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
```diff
@@ -0,0 +1,332 @@
...
+
+        Returns:
+            list of translated strings
+        """
+        res = []
+        for txt in text:
+            src = torch.tensor(self.src_tokenizer.text_to_ids(txt), device=self.device)
+            src = torch.unsqueeze(src, 0)
+            src_embeddings = self.src_embedding_layer(input_ids=src)
+            src_embeddings *= src_embeddings.new_tensor(self.emb_scale)
```
MaximumEntropy, line 322: Scaling the embeddings is commented out in the forward function L166 `# src_embeddings *= src_embeddings.new_tensor(self.emb_scale)`, but not here.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
PeganovAnton: Thanks, @MaximumEntropy! Fixed it.

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
PeganovAnton: I thought that there could be another dataset that does not batch data. If such a dataset is used, `batch[i].ndim == 2` and no squeezing is needed.

`batch` is not a tensor but a tuple of tensors.

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
PeganovAnton: The problem is our model is not training if I uncomment scaling.

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
PeganovAnton: Great point! I added the filtering of beam search results. All ids greater or equal to vocabulary size are replaced with UNK id.

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
MaximumEntropy: I see, but you probably want to comment it out here as well when doing inference right?

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
MaximumEntropy: I see, makes sense!

PeganovAnton (approved):

nemo/collections/nlp/models/machine_translation/transformer_mt_model.py
PeganovAnton: I see. It is fixed.

PR #1426, opened 2020-11-07
Add support for replicated shard strategy for tarred datasets

titu1994:
# Changelog
- Support `replicate` strategy for sharding of tarfiles under ddp for Tarred WebDataset.
- Add config line to all ASR models

### DIFF
--- a/examples/asr/conf/config.yaml
+++ b/examples/asr/conf/config.yaml
@@ -17,7 +17,7 @@ model:
     shuffle: True
     is_tarred: False
     tarred_audio_filepaths: null
-    shard_strategy: "scatter"
+    tarred_shard_strategy: "scatter"
 
   validation_ds:
     manifest_filepath: ???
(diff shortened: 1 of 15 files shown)


redoctopus (commented):
General comment for configs: do we want to have `shard_strategy` set by default, when `is_tarred` is false by default?

titu1994:
As in remove the configs shard_strategy = scatter portion? We could drop them, but it reduces visibility. Practically, it should be fine, since the shard strategy is not used unless is_tarred is set

redoctopus:
I was worried about it potentially causing some confusion, since it's not explicitly named in relation to the tarred dataset. How do you feel about a `null` option like for `tarred_audio_filepaths` when `is_tarred` is False?

titu1994:
Oh that's a good option. None by default switches to the current scatter strategy and otherwise the user sets it explicitly. I'll make the changes

titu1994:
Another is to name the parameter `tarred_shard_strategy`

redoctopus:
Either sounds good to me

redoctopus (approved):
LGTM

titu1994:
Thanks !

PR #1431, opened 2020-11-10
Update tarred dataset conversion script

redoctopus:
Bugfix for when max duration is set to None
Add support for min duration filtering


titu1994 (approved):
LGTM

VahidooX (approved):

PR #1434, opened 2020-11-10
Add support for Tarred dataset for Transformer LM

titu1994:
# Changelog
- Adds tarred dataset support for transformer based Language models
- Pre-tokenizes and chunks the dataset for more efficient external dataset handling
- Adds utility script to construct a tarred tokenized LM dataset
- Adds `cache_ids` flag support for transformer based language models

### DIFF
--- a/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
@@ -39,6 +39,9 @@ class L2RLanguageModelingDataset(Dataset):
         batch_step: distance (in tokens) between two successive sequences of
             the text. By default, it is equal to max_seq_length which corresponds
             to splitting text into disjoint segments covering full dataset
+        cache_ids: bool value, defaults to False. Determines whether the preprocessed,
+            tokenized dataset should be cached into a pickle file. If true, cache is saved
+            at the path provided in `dataset`.
     """
 
     def __init__(
--- a/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
+import json
+import pickle
 from typing import Optional
 
 import numpy as np
-import json
 import webdataset as wd
-import io
-import pickle
 from torch.utils.data import Dataset, IterableDataset
 
 from nemo.collections.common.tokenizers.tokenizer_spec import TokenizerSpec
@@ -47,7 +47,7 @@ class L2RLanguageModelingDataset(Dataset):
         dataset: str,
         max_seq_length: Optional[int] = 512,
         batch_step: Optional[int] = None,
-        cache_ids: bool = False
+        cache_ids: bool = False,
     ):
         self.tokenizer = tokenizer
         self.max_seq_length = max_seq_length
--- a/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
@@ -19,7 +19,7 @@ import json
 import webdataset as wd
 import io
 import pickle
-from torch.utils.data import Dataset,IterableDataset
+from torch.utils.data import Dataset, IterableDataset
 
 from nemo.collections.common.tokenizers.tokenizer_spec import TokenizerSpec
 from nemo.collections.nlp.data.data_utils import dataset_to_ids
@@ -105,7 +105,7 @@ class TarredL2RLanguageModelingDataset(IterableDataset):
 
         # Put together WebDataset
         self._dataset = (
-            wd.Dataset(tarpath)
+            wd.Dataset(tarpath, shard_selection=lambda urls: urls)
             .shuffle(shuffle_n)
             .rename(npy='npy', key='__key__')
             .to_tuple('npy', 'key')
@@ -119,7 +119,7 @@ class TarredL2RLanguageModelingDataset(IterableDataset):
         npy.close()
 
         # flatten data
-        idx = np.random.randint(0, (len(data) - self.max_seq_length) // self.batch_step + 1)
+        idx = np.random.randint(0, (len(data) - self.max_seq_length) // self.batch_step)
 
         # random slice of data
         left = idx * self.batch_step
--- a/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
@@ -38,7 +38,7 @@ class L2RLanguageModelingDataset(Dataset):
         batch_step: distance (in tokens) between two successive sequences of
             the text. By default, it is equal to max_seq_length which corresponds
             to splitting text into disjoint segments covering full dataset
-        cache_ids: bool value, defaults to False. Determines whether the preprocessed,
+        use_cache: bool value, defaults to False. Determines whether the preprocessed,
             tokenized dataset should be cached into a pickle file. If true, cache is saved
             at the path provided in `dataset`.
     """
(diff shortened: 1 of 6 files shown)


redoctopus (commented):
Tarred dataset related files look fine to me. Deferring to the others for NLP model-specific changes.

nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
```diff
@@ -59,3 +65,168 @@ def __getitem__(self, idx):
...
+        if shard_strategy not in valid_shard_strategies:
+            raise ValueError(f"`shard_strategy` must be one of {valid_shard_strategies}")
+
+        with open(metadata_path, 'r') as f:
+            metadata = json.load(f)
+
+        self.metadata = metadata
+
+        if isinstance(text_tar_filepaths, str):
+            # Replace '(' and '[' with '{'
```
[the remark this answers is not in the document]
titu1994: Done. Thanks!

nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
```diff
@@ -42,11 +47,12 @@ def __init__(
         dataset: str,
         max_seq_length: Optional[int] = 512,
         batch_step: Optional[int] = None,
+        cache_ids: bool = False,
```
ekmb, line 50: Could you please update the docstring?

ekmb (approved):

scripts/create_tarred_transformer_lm_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+        logging.basicConfig(level=logging.INFO)
+
+    if os.path.exists(data_root):
+        logging.warning(
+            f'Data root {data_root} already potentially contains files. In such a case,'
+            f'please be aware that the tarfiles will be **appended** instead of overridden!'
+        )
+
+    if not os.path.exists(data_root):
+        os.makedirs(data_root)
```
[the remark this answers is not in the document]
titu1994: Good catch, that was the wrong check. Im more worried about the tokenized data existing (and being appended to) than the fact that the base directory already exists.

nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
VahidooX: We used the term use_cache instead of cache_id in nlp collection. Do you think it is OK to change the name to have consistency?

nemo/collections/nlp/models/language_modeling/transformer_lm_model.py
```diff
@@ -163,16 +184,49 @@ def setup_test_data(self, test_data_config: Optional[DictConfig]):
...
+            if ('tarred_text_filepaths' in cfg and cfg['tarred_text_filepaths'] is None) or (
+                'file_name' in cfg and cfg['file_name'] is None
+            ):
+                logging.warning(
+                    "Could not load dataset as `file_name` was None or "
+                    f"`tarred_text_filepaths` is None. Provided config : {cfg}"
+                )
+                return None
+
+            shuffle_n = cfg.get('shuffle_n', 4 * cfg['batch_size']) if shuffle else 0
```
VahidooX, line 202: Do you think we can go higher than 4 times here?

scripts/create_tarred_transformer_lm_dataset.py
```diff
@@ -0,0 +1,303 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
VahidooX, line 1: This scripts folder is getting crowded. Maybe we should organize them based on tasks or collections. We mostly keep the scripts of nlp in their own example folder, under a subfolder named 'data'. But both locations are OK, your choice. Is this script can be used with the transfromer_lm.py example?

nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py
titu1994: Done

scripts/create_tarred_transformer_lm_dataset.py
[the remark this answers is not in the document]
titu1994: Done

VahidooX (changes requested):

nemo/collections/nlp/models/language_modeling/transformer_lm_model.py
titu1994: We can, this is just the default used for ASR. Technically speaking, we have a very small amount of tarfiles (64) for nlp compared to ASR (512/1024) + each tarfile contains very few chunks of samples (75-80 .npy files). So practically it doesnt matter what shuffle size you put, as long as its over 100, the entire tarfile will get shuffled for small amounts of text. Its only when we go to TB of text that this value beomes somewhat important.

scripts/create_tarred_transformer_lm_dataset.py
titu1994: Reorganizing scripts should be done in another PR, its out of scope of this one. Scripts folder should probably be refactored to emulate the current collections hierarchy so that can be tackled in another PR.

scripts/create_tarred_transformer_lm_dataset.py
```diff
@@ -0,0 +1,322 @@
...
+    global_chunk_idx = 0
+    chunk_paths = []
+    chunk_lens = []
+
+    # buffer parameters
+    data_cache = []
+    chunk_idx = 0
+
+    text_generator = iter(__build_dataset_from_text(text_paths, lower_case=lower_case, chunk_size=chunk_size))
+    idx = 0
```
VahidooX, line 151: idx looks like never used.

VahidooX (approved):

scripts/create_tarred_transformer_lm_dataset.py
titu1994: Good catch, fixed

scripts/create_tarred_transformer_lm_dataset.py
```diff
@@ -0,0 +1,318 @@
...
+parser.add_argument('--data_root', required=True, default=None, type=str, help='Output directory')
+
+# General arguments
+parser.add_argument(
+    '--chunk_write_buffer',
+    default=128,
+    type=int,
+    help='Number of chunks of `chunk_size` to buffer for parallel tokenization and serial write to disk',
+)
+parser.add_argument('--lower_case', action='store_true', help='Whether to lower case the corpus')
```
AlexGrinch, line 57: Why do we need this as a separate option? I think that this is a property of specific tokenizer.

AlexGrinch (commented):
Why do we need separate option for lowercasing?

titu1994:
@AlexGrinch The tokenizer can be either cased or uncased - for example LS normalized text corpus is all lower case chars, but other datasets may be cased. When training a LM with a custom vocabulary, it is worthwhile to have an option to lowercase the text prior to tokenization if required, so that additional processing is not required inside the tarred dataset.

AlexGrinch:
@titu1994 Please correct me if I am wrong. What I understood from the code is that in the process of creating tarred dataset, we take all text, transform it into token_ids via _tokenizer.text_to_ids(text)_, and save this numbers into tarred numpy array. If our tokenizer works with lower case text, text.lower() is called somewhere inside _tokenizer.text_to_ids(text)_. It seems that there is no need to do it in advance as we call _tokenizer.text_to_ids(text)_ either way during dataset creation.

titu1994:
For ASR datasets (as an example) - Librispeech LM corpus - the tokekizer is a bert base cased. The language corpus itself is in all Upper case, but the tokekizer is built to support only lower case. Since it is a cased tokekizer, it will convert all text to [UNK] instead of lowering it. 

Since this script directly uses a prebuilt tokekizer vocabulary therefore it is necessary to provide the logic to explicitly lowercase via the flag for compatibility betweenvrae text corpus, the tokekizer vocabulary and the tokekizer construction (cased or uncased)

AlexGrinch:
Okay, I think we can merge this PR after it passes all checks.

titu1994:
Sounds good, I'll fix up the error

PR #1418, opened 2020-11-05
Add GAN-based TTS Vocoder

blisc:
- Adds the fullband MelGAN model as described in https://arxiv.org/abs/2005.05106
- Remove empty *_step() from ModelPT
- Split UniGlowLoss into glow portion and STFTLoss portion
- Simply stft_exact_pad option in FilterBankFeatures

ToDos
- [x] ~Enable FP16~. No currently a priority.
- [x] Add NeMo Loss subclass
- [x] Update Docs
- [x] ~Update TTS tutorials~, can't until I upload the .nemo files

PR #1437, opened 2020-11-10
Update NeMo core for melgan

blisc:
Split off updates to nemo core that are required for #1418 

Changes:
- Remove empty *_step() methods from ModelPT
- Treat None inside of ElementType's parameters as a Void that matches anything
  - Use case being removing the default 16kHz sample_rate from AudioSignal and replacing it with None. This allows a module with type AudioSignal() to match ANY AudioSignal regardless of sample_rate

Minor Changes:
- Add Singleton NeuralType
- Update wrap_training_step such it is doesn't require a dictionary to be passed
- Other cosmetic changes

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -505,7 +505,7 @@ class ModelPT(LightningModule, Model):
     @abstractmethod
     def setup_validation_data(self, val_data_config: Union[DictConfig, Dict]):
         """
-        (Optionally) Setups data loader to be used in validation
+        Setups data loader to be used in validation
         Args:
 
             val_data_layer_config: validation data layer parameters.
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -176,7 +176,7 @@ class ModelPT(LightningModule, Model):
 
     def _default_save_to(self, save_path: str):
         """
-        Saves model instance (weights and configuration) into .nemo file. 
+        Saves model instance (weights and configuration) into .nemo file.
         You can use "restore_from" method to fully restore instance from .nemo file.
 
         .nemo file is an archive (tar.gz) with the following:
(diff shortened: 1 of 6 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -739,23 +739,14 @@ def train_dataloader(self):
         if self._train_dl is not None:
             return self._train_dl
 
-    def training_step(self, batch, batch_ix):
```
titu1994, line 742: Why are the default methods being removed ? It's important to have them for ease of use while building up models partially (without having to define all of the steps at once)

titu1994 (changes requested):
Overall looks fine, but I don't understand the decision to remove the default step functions.

nemo/core/classes/modelPT.py
blisc: Let me ask you this: Why have them? If we require models to have them, they should be abstract.

As to why we shouldn't have them:
- I removed them and nothing fails, which imo makes these functions useless.
- They aren't up-to-date with LightningModule's latest signature.

I suppose this portion isn't strictly necessary for MelGAN, so I can remove it. But I do feel like this is the correct choice.

titu1994: When developing new models, it's not convenient to have to write every single step immediately - say val and test because they may change significantly. 

Especially in cases where models have multiple eval states - eval during train, vs inference during test etc. 

As to signatures not being updated, that can be updated. 

They aren't abstract as we don't want subclasses to have stubs (as they are now). 

Nothing fails right now because all of the models we developed have already reached end stage of development, it's an issue during development itself. 

I suppose going forward enforcing all subclasses to submit all 3 methods or stubs is fine too.

titu1994 (approved):
LGTM, please wait for @okuchaiev review regarding new AxisKind

okuchaiev (approved):
looks good to me

PR #1439, opened 2020-11-11
vad outputs to manifest script for speaker diarization

nithinraok:

### DIFF
--- a/scripts/vad_to_manifest.py
+++ b/scripts/vad_to_manifest.py
@@ -15,6 +15,7 @@
 import argparse
 import glob
 import json
+import os
 
 from nemo.utils import logging
 
@@ -40,8 +41,8 @@ def write_manifest(vad_directory, audio_directory, manifest_file):
                 vad_out = line.strip().split()
                 start, dur, activity = float(vad_out[0]), float(vad_out[1]) - float(vad_out[0]), vad_out[2]
                 if activity.lower() == 'speech':
-                    audio_name = os.path.join(audio_directory, audio_name + '.wav')
-                    meta = {"audio_filepath": audio_name, "offset": start, "duration": dur, "label": 'UNK'}
+                    audio_path = os.path.join(audio_directory, audio_name + '.wav')
+                    meta = {"audio_filepath": audio_path, "offset": start, "duration": dur, "label": 'UNK'}
                     json.dump(meta, outfile)
                     outfile.write("\n")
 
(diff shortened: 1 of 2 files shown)


scripts/vad_to_manifest.py
```diff
@@ -0,0 +1,65 @@
...
+        "--audio_directory",
+        help="path to audio directory of audio files for which vad was computed",
+        type=str,
+        required=True,
+    )
+    parser.add_argument("--manifest_file", help="output manifest file name", type=str, required=True)
+
+    args = parser.parse_args()
+    vad_directory, audio_directory, manifest_file = (args.vad_directory, args.audio_directory, args.manifest_file)
+    wite_manifest(vad_directory, audio_directory, manifest_file)
```
fayejf, line 64: should be `write_manifest`

fayejf (changes requested):
add  `import os`

scripts/vad_to_manifest.py
```diff
@@ -0,0 +1,65 @@
...
+    with open(manifest_file, 'w') as outfile:
+        for vad_file in vad_files:
+            f = open(vad_file, 'r')
+            lines = f.readlines()
+            audio_name = os.path.basename(vad_file).split('.')[0]
+            for line in lines:
+                vad_out = line.strip().split()
+                start, dur, activity = float(vad_out[0]), float(vad_out[1]) - float(vad_out[0]), vad_out[2]
+                if activity.lower() == 'speech':
+                    audio_name = os.path.join(audio_directory, audio_name + '.wav')
```
fayejf, line 43: Gonna be a lot of wav wav wav lol

fayejf (changes requested):

scripts/vad_to_manifest.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
jbalam-nv, line 1: Do we need another script for this? Could this be added as an optional argument in VAD so that VAD can output manifests directly? @fayejf @nithinraok

jbalam-nv (changes requested):
No issues with

scripts/vad_to_manifest.py
nithinraok: This would enable users to use external vad and combine our speaker embeddings for clustering. And also for testing out oracle vad

nithinraok: so has to be an independent script

jbalam-nv: Got it. Thanks!

jbalam-nv (approved):
LGTM

fayejf (approved):

PR #1460, opened 2020-11-16
Bug fixes for text classification model

VahidooX:


nemo/collections/nlp/data/text_classification/text_classification_dataset.py
```diff
@@ -103,7 +104,7 @@ def __init__(
                 f"cached_{filename}_{tokenizer_name}_{max_seq_length}_{vocab_size}_{num_samples}_{self.pad_id}_{shuffle}.pkl",
             )
 
-            if not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
+            if get_envint("LOCAL_RANK", 0) == 0:
```
ericharper, line 107: I think you want global rank here? If you have multi-node, then there's more than one local rank 0.

VahidooX: I need local rank 0, as the data need to process on each node separately as data can is being stored in local node and not a shared space.

vladgets (approved):

nemo/collections/nlp/data/text_classification/text_classification_dataset.py
ericharper: PTL has the option for both with their `prepare_data` hook: https://pytorch-lightning.readthedocs.io/en/stable/lightning_module.html#prepare-data

VahidooX: Actually it is a good idea too move all nlp data processing to prepare_date and use this style, but it is outside of the scope of this PR for the bug fix. We need to add this feature to our TODOs to refactor how we handle data. Som just added the feature for reading and loading from tar datasets for language modelling (better than numpy array as it prevents loading multiple copy of the data in memory per GPU). This effort may need more investigation.

ericharper: Agreed, it's out of scope.

ericharper (approved):

PR #1466, opened 2020-11-18
fix for scripts for audio->manifest in SD/VAD

fayejf:
tiny fix write_long_audio_manifest.py 
Results are same actually, Just the previous code casue negative offsets in manifests which is not good.


nithinraok (approved):
LGTM

PR #1470, opened 2020-11-19
updates for test mode for TLT

ekmb:


examples/nlp/glue_benchmark/glue_benchmark_config.yaml
```diff
@@ -29,18 +29,17 @@ model:
     drop_last: false
 
   train_ds:
-    file_name: 'train.tsv'
+    ds_item: 'train.tsv'
```
tkornuta-nvidia, line 32: hmmm, what ds stands for? dataset? what was wrong with file_name?

titu1994: @ekmb Train_ds does not support multi dataset, so name ds_item isn't treated differently. Probably best to keep it as file_name

ekmb: @titu1994 I don't need multi dataset for training, just wanted to be consistent with validation. 
Discussed with @tkornuta-nvidia offline.

nemo/collections/nlp/models/token_classification/token_classification_model.py
```diff
@@ -181,17 +184,28 @@ def validation_epoch_end(self, outputs):
...
-        return self.validation_epoch_end(outputs)
+        avg_loss = torch.stack([x['test_loss'] for x in outputs]).mean()
+        # calculate metrics and classification report
+        precision, recall, f1, report = self.classification_report.compute()
+        logging.info(report)
+
+        self.log('test_loss', avg_loss, prog_bar=True)
+        self.log('precision', precision)
+        self.log('f1', f1)
+        self.log('recall', recall)
```
tkornuta-nvidia, line 208: Ok, so this is what I needed, awesome! 👍   thaks @ekmb !

tkornuta-nvidia (approved):

PR #1477, opened 2020-11-20
Fixes to IntentSlots init logic making all TLT scripts operational

tkornuta-nvidia:
This is still work in progress - but at least:

- [x] CI passes
- [x] whole TLT workflow (train, finetune, evaluate, inference, export, infer ONNX scripts) work
- [x] setting datadir also "reconfigures" the model (e.g. slots/intents change => heads change too)
- [x] **!! change in the way ModelPT is handing artifacts!!** => saving original path despite whether artifact was restored from that path or from "./"

Still, the whole logic of the model is just horrible. The major problem I see here is improper `separation of concerns`:
1) there is improper separation of parameters - plenty of parameters are shared between model and dataset/dataloader related sections => each `setup_*_ds` should receive all the params required to create ds&dl

2) there is DataDescriptor that does magical operations on slots and values from both training and validation splits => those values should be gathered together during the data preparation phase, and `setup_train_ds` should load them, store in config and other `setup_*_ds` should simply use them

3) DataDescriptor is needed to be created before calling parent class constructor (which calls `setup_*_ds`) and it requires slots and values from both splits to be loaded first. This creates an chicken and egg problem - get rid of DataDescriptor and follow logic from 2) to get rid of that cycle.

4) model reconfiguration should change only last layers of classifier - now it is (re)instantiating the whole classifier object, which IMO is not the best design...


nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
```diff
@@ -53,70 +53,124 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
...
         """
-
-        self.data_dir = cfg.data_dir
         self.max_seq_length = cfg.language_model.max_seq_length
 
-        self.data_desc = IntentSlotDataDesc(
-            data_dir=cfg.data_dir, modes=[cfg.train_ds.prefix, cfg.validation_ds.prefix]
-        )
+        # Check the presence of data_dir.
+        if not cfg.data_dir:
```
ekmb, line 59: Maybe also add `or not os.path.exists(cfg.data_dir)` ? For the models trained on ngc and then restored, the config will have data_dir with an invalid path

tkornuta-nvidia: 👍

tkornuta-nvidia: Still, the "restore path" is handled in a totally different way...

ekmb: But is it working? during restoration `cfg.data_dir` won't be None and `_copy_data_desc_to_cfg` will be attempted, no?

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
```diff
@@ -53,70 +53,124 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
...
         # init superclass
         super().__init__(cfg=cfg, trainer=trainer)
 
+        # Check the presence of data_dir.
+        if cfg.data_dir:
+            # Conditional initialization of modules.
+            self._init_modules_when_tokenizer_set()
+
+        # Enable setup methods.
+        IntentSlotClassificationModel._set_model_restore_state(is_being_restored=False)
```
ekmb, line 78: Why do we need it here?

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
```diff
@@ -53,70 +53,124 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
...
+        # Intents.
+        cfg.data_desc.intent_labels = list(data_desc.intents_label_ids.keys())
+        cfg.data_desc.intent_label_ids = data_desc.intents_label_ids
+        cfg.data_desc.intent_weights = data_desc.intent_weights
+        # Slots.
+        cfg.data_desc.slot_labels = list(data_desc.slots_label_ids.keys())
+        cfg.data_desc.slot_label_ids = data_desc.slots_label_ids
+        cfg.data_desc.slot_weights = data_desc.slot_weights
+
+        cfg.data_desc.pad_label = data_desc.pad_label
```
ekmb, line 94: A good place to register *label_ids as artifacts

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
```diff
@@ -53,70 +53,124 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
...
+
+        Args:
+            data_dir: path to data directory
+        """
+        logging.info(f'Setting data_dir to {data_dir}.')
         self.data_dir = data_dir
-        logging.info(f'Setting model.data_dir to {data_dir}.')
+        # Finish the "conditional initialization" by passing the new data_dir.
+        self._setup_tokenizer(self.cfg.tokenizer)
+        self._init_modules_when_tokenizer_set()
```
ekmb, line 173: Please consider renaming the method, it's doing way more than changing the data_dir + please update the docstring

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
```diff
@@ -53,70 +53,124 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
...
         Args:
             data_dir: path to data directory
         """
+        logging.info(f'Setting data_dir to {data_dir}.')
+        self.data_dir = data_dir
+        # Update configuration with new data.
+        self._copy_data_desc_to_cfg(self.cfg, data_dir, train_ds, validation_ds)
+        # Finish the "conditional initialization" by passing the new data_dir.
+        self._setup_tokenizer(self.cfg.tokenizer)
+        self._init_modules_when_tokenizer_set()
```
ekmb, line 159: Please consider renaming the method, it's doing way more than changing the data_dir + please update the docstring

ekmb (changes requested):

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
tkornuta-nvidia: to disable calling setup when `cfg.data_dir` is not set

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
tkornuta-nvidia: yeah, next iteration - extracting them in TLT export

my goal was not to fix **all** the issues - my goal here is to make it work ;)

ekmb (approved):

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
tkornuta-nvidia: I have tested it in TLT in: training (with two ways injecting data_dir), finetuning, evaluation and inference. Plus it is passing CI in NeMo. So I would say it is working... but still need to be refactored ;)

nemo/core/classes/modelPT.py
```diff
@@ -169,7 +169,9 @@ def register_artifact(self, config_path: str, src: str):
                 used_src = src
             if not os.path.exists(used_src):
                 raise FileNotFoundError(f"Could not find {used_src}")
-            self.artifacts.append((config_path, used_src))
+            # But disregarding whether you use "local" or "remote" artifact - always store the original path.
+            # This fixes issues raising when finetuning NLP models that create and register tokenizer vocabs.
+            self.artifacts.append((config_path, src))
```
tkornuta-nvidia, line 174: @okuchaiev @ericharper @titu1994 @blisc @ekmb 

This is important for TLT fine-tuning where the (save-restore-save) cycle in NLP model tokenizers happen, can go with the full explanation "in person".

All tests are passing, still I think it is important to note this change.

tkornuta-nvidia: @okuchaiev despite green light will wait for your approval with merging this one as it is in "the core of the core".

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
tkornuta-nvidia: done

nemo/collections/nlp/models/intent_slot_classification/intent_slot_classification_model.py
tkornuta-nvidia: done, generally refactored the method, so it is doing that it was supposed to do: reconfigure the model

nemo/core/classes/modelPT.py
titu1994: This needs to be used source only. The tests pass because all of the files exist in the test environment (in the original paths that were used for registration) - the real test is when you run this section on another users environment which doesn't contain those files. Blocking until proper review by @okuchaiev

titu1994 (changes requested):

tkornuta-nvidia:
@titu1994 The fix comes from me doing the cycles (save - restore save), which is not typical NeMo use-case, but it is typical TLT workflow. Thanks for the comment though.

titu1994:
Save restore save is the general cycle of NeMo too - save when we train a model, convert to .Nemo and upload, use from_pretrained(), change vocabulary, fine-tune, save_to() again. 

It's not just a TLT workflow, so please test this on an empty environment which doesn't have the files in exactly the locations that save was called. 

So save on one machine, move to another, restore from in that machine and save as well. If no issues are there during restore on a second machine, all's good.

tkornuta-nvidia:
> So save on one machine, move to another, restore from in that machine and save as well. If no issues are there during restore on a second machine, all's good.

Has _anyone_ tested that use-case?

Just perform a "mental experiment" and go through (restore - save) cycle in your head: any idea where `save_to()` method will take the "restored" artifact that you will need to add to the archive during saving? 

> Save restore save is the general cycle of NeMo too - save when we train a model, convert to .Nemo and upload, use from_pretrained(), change vocabulary, fine-tune, save_to() again.

This is a slightly different case as you are setting new artifact, so the old one, that is in archive, won't be needed anymore.

But let me repeat the same question: has _anyone_ tested this use-case?

Please perform 2nd "mental experiment": If the model reconfiguration is not happening in the model constructor (e.g. ASR Models have a separate method for this) the current `register_artifacts` code will **append** the second artifact, not overwrite the old one... Is that the desired behavior?

tkornuta-nvidia:
Summarizing, I like the idea of artifacts, but I also see plenty of problems with the way they currently work. And I have some ideas how to solve those problems, plus would like to add some additional functionality to them  and hope we will discuss all that some time soon.

But fixing artifacts in general is IMO out of scope of this PR - the goal here was to incorporate changes enabling to have the whole TLT workflow, with artifacts passed along different scripts, operational.

@okuchaiev makes sense?

nemo/core/classes/modelPT.py
blisc: This area is outside of my knowledge with nemo. I can't really comment on this change.

titu1994:
> So save on one machine, move to another, restore from in that machine and save as well. If no issues are there during restore on a second machine, all's good.

ASR Notebooks 1, 2, 4 and I believe 6 does this. ASR 1 (finetuning section) is basically the above workflow - where second environment is Colab and 1st is our local train environment. 

> Just perform a "mental experiment" and go through (restore - save) cycle in your head: any idea where save_to() method will take the "restored" artifact that you will need to add to the archive during saving?

1) Load model using hydra, train, finish, call save_to() - saves the local path of the tokenizer into config + actual tokenizer vocab into the archive.

2) Restore model using restore_from inside colab (or equivalent from_pretrained) - the code reperforms register_artifact() - this time the actual value doesnt exist in the old local path, so it instead searches and finds the new path (inside the tarfile). Returns that path *** This new path is required by all tokenizers and is what is being returned from this function - prior to this PR change.

3) The model gets the new (tarpath filepath) and submits it to create a tokenizer from the vocabulary present in the local path.

The entire workflow above is contingent on the fact that the returned path is the **new** (tarpath filepath) of the tokenizer vocabulary. So your change breaks all of ASR BPE.

The tests dont capture this - because as I said, the test environment still has the original tokenizer vocab after save and durin restore. So it never needs to do the second half of step (2) - the original filepath is accurate and goes through without finding the path from the tarfile.

> But let me repeat the same question: has anyone tested this use-case?

Yes - all of ASR BPE models are using this.

> Please perform 2nd "mental experiment": If the model reconfiguration is not happening in the model constructor (e.g. ASR Models have a separate method for this) the current register_artifacts code will append the second artifact, not overwrite the old one... Is that the desired behavior?

That seperate method is only called by the constructor or change_vocabulary(). They do not append any data - nor is there overwriting. The save_to path is different from the restore_from path. However, even if save_to path is same as restore_from path, the tarfile created will simply overwrite the old tarfile, so still there is no appending, just overwriting.

tkornuta-nvidia:
Som, I am not saying that the model (restore->save) is not working. And maybe I am missing something, but as far as I know the notebooks you mentioned are not doing any of the discussed use-cases. 

Let's take that off-line, this is super inefficient for discussions requiring _alignment on problem definition_.

okuchaiev:
@tkornuta-nvidia  before a proper review, quick question - will this make "save-restore-save" cycle work for "save (machine A)-restore(Machine B)-save" ?

titu1994 (approved):
Needs urgent PR to fix the restoration path

PR #1489, opened 2020-11-23
Add default mix to mono to common voice script

SeanNaren:
Ensures we mix to mono for processing the data. This fixes a bug where I was getting stereo audio files into the audio loader when testing wav2vec.


okuchaiev (approved):
Thank you!
@redoctopus - FYI, MCV could have some stereo files

PR #1494, opened 2020-11-25
Patch for NLP vocab artifacts

ericharper:
Changed setup_tokenizer logic so that if the model is being restored it will try to use the vocab file "tokenizer.vocab_file" first.

If "tokenizer.vocab_file" doesn't exist, then it will fall back to cfg.vocab_file.

If no `vocab_file` is given, we will try to get the vocab from the tokenizer and then register the `vocab_dict` and the `vocab_file`.
When writing the `vocab_dict` and `vocab_file` that we get from the tokenizer we write them to a cache directory based on a hash of the `vocab_dict`.

TODO: create docstring to clarify the flow

### DIFF
--- a/nemo/collections/nlp/models/nlp_model.py
+++ b/nemo/collections/nlp/models/nlp_model.py
@@ -96,6 +96,7 @@ class NLPModel(ModelPT):
                     config_path='tokenizer.tokenizer_model', src=cfg.tokenizer_model
                 ),
             )
+            self.tokenizer = tokenizer
         else:
             tokenizer = get_tokenizer(
                 tokenizer_name=cfg.tokenizer_name,
@@ -105,9 +106,9 @@ class NLPModel(ModelPT):
                     config_path='tokenizer.tokenizer_model', src=cfg.tokenizer_model
                 ),
             )
+            self.tokenizer = tokenizer
             # when there is no vocab file we try to get the vocab from the tokenizer
             self._register_vocab_from_tokenizer(config_path='tokenizer.vocab_file', src=cfg.vocab_file)
-        self.tokenizer = tokenizer
 
     @rank_zero_only
     def _register_vocab_from_tokenizer(self, config_path='tokenizer.vocab_file', src: str = None):
--- a/nemo/collections/nlp/models/nlp_model.py
+++ b/nemo/collections/nlp/models/nlp_model.py
@@ -87,44 +87,29 @@ class NLPModel(ModelPT):
         Args:
             cfg (DictConfig): Tokenizer config
         """
-        if self._is_model_being_restored():
+        if self._is_model_being_restored() and os.path.exists('tokenizer.vocab_file'):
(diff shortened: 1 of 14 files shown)


nemo/collections/nlp/models/nlp_model.py
```diff
@@ -80,44 +81,66 @@ def register_bert_model(self):
...
+            if isinstance(self.tokenizer, AutoTokenizer):
                 # extract vocab from tokenizer
-                vocab_json_src = os.path.join(NEMO_NLP_TMP, vocab_dict_config_path)
                 vocab_dict = self.tokenizer.tokenizer.get_vocab()
+
+                # get hash of vocab_dict to create a unique directory to write vocab_dict and vocab_file
+                m = hashlib.md5()
+                # get string representation of vocab_dict
+                vocab_dict_str = json.dumps(vocab_dict, sort_keys=True).encode()
+                m.update(vocab_dict_str)
```
titu1994, line 131: This will fail for tokekizer name is bert-base-case and bert-base-uncased with exact same vocabulary content.  The likelihood of such an occurrence is low but possible, so let's do a hashlib.updade() before this with the value of cfg.tokenizer_name. 

That way even if vocab is exact match, as long as the Tokenizers name is also different, different hash directory should be created.

titu1994 (changes requested):
LGTM. Minor change regarding hash conflict in an edge case. 

One thing, while I know this code will work, it is worthwhile adding some rudimentary checks for all the 3 paths of execution of this tokekizer saving-restorinf-constructing steps. Can be done in another PR later too.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -80,44 +81,66 @@ def register_bert_model(self):
...
+            tokenizer_model=self.register_artifact(config_path='tokenizer.tokenizer_model', src=cfg.tokenizer_model),
         )
         self.tokenizer = tokenizer
-        self.register_tokenizer(cfg)
+
+        if vocab_file is None:
+            # when there is no vocab file we try to get the vocab from the tokenizer and register it
+            self._register_vocab_from_tokenizer(config_path='tokenizer.vocab_file')
 
     @rank_zero_only
```
okuchaiev, line 111: why rank_zero_only here?

okuchaiev:
shouldn't tokenizer_model be handled similarly?

okuchaiev:
> shouldn't tokenizer_model be handled similarly?

ok, got it.

nemo/collections/nlp/models/nlp_model.py
ericharper: I don't want multiple processes writing to the same file.

okuchaiev (approved):

nemo/collections/nlp/models/nlp_model.py
ericharper: Good idea, added it.

titu1994 (approved):
LGTM

PR #1496, opened 2020-11-25
Add torchscript export for IntentSlotClassificationModel.

egorchakov:
A bit of a roundabout way of allowing `IntentSlotClassificationModel`s to be exported to torchscript. Not immediately sure how to do this in a cleaner way without adding/changing too much.


egorchakov:
Is there a way to check what jenkins is failing with? It seems the details link points to a non-public IP.

tests/collections/nlp/test_nlp_exportables.py
```diff
@@ -103,6 +104,28 @@ def test_IntentSlotClassificationModel(self, dummy_data):
             assert onnx_model.graph.output[0].name == 'intent_logits'
             assert onnx_model.graph.output[1].name == 'slot_logits'
 
+    @pytest.mark.run_only_on('GPU')
+    @pytest.mark.unit
+    def test_IntentSlotClassificationModel_export_to_torchscript(self, dummy_data):
```
drnikolaev, line 109: This method needs to be extended by some basic assertions like the one below, `test_TokenClassificationModel_export_to_onnx`

drnikolaev (changes requested):
Please add assertions

tests/collections/nlp/test_nlp_exportables.py
egorchakov: Added some of those. Though torchscript model internals seem to be a bit more difficult to inspect -- hopefully the assertions make sense.

drnikolaev (approved):

okuchaiev (approved):
lgtm, thanks!

PR #1497, opened 2020-11-25
Add transcription functionality to ASR Classification models

titu1994:
# Changelog
- Refactor TopKClassification accuracy to make `top_k` a property + perform top_k operation seperately
- Add tests for transcription scenarios

### DIFF
--- a/nemo/collections/asr/models/classification_models.py
+++ b/nemo/collections/asr/models/classification_models.py
@@ -18,6 +18,8 @@ from typing import Dict, List, Optional, Union
 
 import onnx
 import torch
+import tempfile
+import json
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from pytorch_lightning import Trainer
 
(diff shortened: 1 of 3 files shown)


nemo/collections/asr/models/classification_models.py
```diff
@@ -60,8 +62,81 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+                    else:
+                        labels_k = []
+                        top_ks = self._accuracy.top_k
+                        for top_k_i in top_ks:
+                            # replace top k value with current top k
+                            self._accuracy.top_k = top_k_i
+                            labels_k_i = self._accuracy.top_k_predicted_labels(logits)
+                            labels_k.append(labels_k_i)
+
+                        # convenience: if only one top_k, pop out the nested list
```
fayejf, line 125: nice

fayejf (commented):
Generally looks good to me! Just have a question for `_setup_dataloader_from_config` called in `_setup_transcribe_dataloader `. Could you please remind me where to input the length of the segment to be sent to NN (say 1.28s used in training)?

titu1994:
For transcription, duration is a dummy value that's not precalculated. It would cost too much compute to precompute the durations of all the samples simply for logging.

titu1994:
Hmm thinking about the vad case, it doesn't work with this dataloader at all. So feel free to push changes to this branch (or an addendum PR to this one) to add vad transcription. 

I think it would be cleaner to have a `transcribe_vad` or equivalent and properly accept or precompute durations for such batched segment inference as in vad case

fayejf:
Hmm VAD is another case and we have a vad_infer and we have vad_frame_seq_collate_fn to handle this.

fayejf:
> For transcription, duration is a dummy value that's not precalculated. It would cost too much compute to precompute the durations of all the samples simply for logging.

emm I didn’t mean precompute the durations of all the files for logging. Just wondering If we have a 5 sec audio, then what if we want my inference window to be 1.28s or 1.0s?

titu1994:
This PR performs offline inference only - and is meant to consume the entire audio sequence and emit one label per sample only. That's why we should have a seperate method to perform segmented (aka infer every T seconds) that is not part of this PR. 

The .transcribe() method comes from the parent ASRModel and as such needs to have similar semantics - offline full scale transcription / classification over an entire sample at once go.

fayejf:
> The .transcribe() method comes from the parent ASRModel and as such needs to have similar semantics - offline full scale transcription / classification over an entire sample at once go.

Got it. Thanks for the explanation!  I thought it gonna perform streaming/average window inference so I was wondering I missed anything from dataloader for it.

fayejf (approved):
LGTM

titu1994:
Yes, let's add another method (let's think of a good name, maybe transcribe_streaming or transcribe_segments or something like that) to perform actual stream classification over limited time segment. It's useful to have a method to do it rather than rely on notebook all the time.

fayejf:
> Yes, let's add another method (let's think of a good name, maybe transcribe_streaming or transcribe_segments or something like that) to perform actual stream classification over limited time segment. It's useful to have a method to do it rather than rely on notebook all the time.

Yes good idea! We have sliced/vad_frame_seq_collate_fn now. It's gonna be awesome to have unified **streaming transcribe** function for classification models, label models, and etc. So we can have more functionality coming in online/offline such as smoothing the inference results by using average window.

Issue #1503, opened 2020-12-01
[Question] External LM and Beam Search Notebook/Guide?


gtcooke94:
I saw that the v1.0.0b2 tagged release says "Added beam search and external LM rescoring support for character-based CTC ASR models."
Is there a notebook on how to do this, or for now is it easiest to dig around in the code?

gtcooke94:
For anyone looking at this, I dug into the PRs and found this notebook added:

https://github.com/NVIDIA/NeMo/blob/v1.0.0b2/tutorials/asr/Offline_ASR.ipynb

PR #1504, opened 2020-12-01
Add tutorial entry for external language model

gtcooke94:
I asked this before I found the tutorial notebook: https://github.com/NVIDIA/NeMo/issues/1503

This question of how to add external language models is very common, so now that it is supported I think it would be worthwhile to add it to the chart of the main README - I'd guess you will all get fewer questions that way as well.

Figured I'd save others some time and go on and submit this small PR.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,6 +207,9 @@ To run tutorials:
    * - ASR
      - Online Noise Augmentation
      - `Online noise augmentation <https://colab.research.google.com/github/NVIDIA/NeMo/blob/v1.0.0b2/tutorials/asr/05_Online_Noise_Augmentation.ipynb>`_
+   * - ASR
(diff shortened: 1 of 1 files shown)


okuchaiev (approved):
thank you!

ShantanuNair:
@okuchaiev Is there going to be any more work on external LMs? I'm asking as this seems like that notebook is not very direct to productionize, especially since we're moving some compute to the CPU (is this not an area where we could use the GPU effectively). 

There's also the issue where the timestamps generated which used for beam search skew heavily for longer files (not talking about the offset inherent to the model), on which I am working on right now, and will create a PR for. Right now, given a longer file, the timestamps are offset from the words more and more later in the transcription.

PR #1509, opened 2020-12-01
Preserve absolute path of model restore file

titu1994:
# Changelog
- Preserve absolute path of provided path in order to correctly check whether its a tarfile - when cwd() is inside of another tarfile.

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -476,7 +476,7 @@ class ModelPT(LightningModule, Model):
             raise FileNotFoundError(f"Can't find {restore_path}")
 
         global _MODEL_RESTORE_PATH
-        _MODEL_RESTORE_PATH = os.path.abspath(restore_path)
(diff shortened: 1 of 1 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -476,7 +476,7 @@ def restore_from(
             raise FileNotFoundError(f"Can't find {restore_path}")
 
         global _MODEL_RESTORE_PATH
-        _MODEL_RESTORE_PATH = restore_path
+        _MODEL_RESTORE_PATH = os.path.abspath(restore_path)
```
tkornuta-nvidia, line 479: how about `os.path.expanduser` so we can properly handle '~'?

titu1994: Oh yeah, good catch

titu1994: Done

tkornuta-nvidia (approved):

PR #1511, opened 2020-12-02
Update model config automatically inside ModelPT

titu1994:
# Changelog
- Immediately convert all Hydra 0.x configs to Hydra 1.x configs inside constructor of ModelPT
- Add tests to ensure that no backward incompatible changes exist
- Update every ASR side config at once - other domains can update as per convenience.

### DIFF
--- a/nemo/collections/asr/models/ctc_bpe_models.py
+++ b/nemo/collections/asr/models/ctc_bpe_models.py
@@ -53,8 +53,8 @@ class EncDecCTCModelBPE(EncDecCTCModel, ASRBPEMixin):
 
     def __init__(self, cfg: DictConfig, trainer=None):
         # Convert to Hydra 1.0 compatible DictConfig
-        cfg = model_utils.setup_model_config(cfg)
-        cfg = model_utils.convert_model_config(cfg)
+        cfg = model_utils.convert_model_config_to_dict_config(cfg)
+        cfg = model_utils.maybe_update_config_version(cfg)
 
         if 'tokenizer' not in cfg:
             raise ValueError("`cfg` must have `tokenizer` config to create a tokenizer !")
--- a/nemo/collections/asr/models/ctc_bpe_models.py
+++ b/nemo/collections/asr/models/ctc_bpe_models.py
@@ -63,16 +63,10 @@ class EncDecCTCModelBPE(EncDecCTCModel, ASRBPEMixin):
 
         # Set the new vocabulary
         with open_dict(cfg):
-            if "params" in cfg.decoder:
-                cfg.decoder.params.vocabulary = ListConfig(list(vocabulary.values()))
-            else:
-                cfg.decoder.vocabulary = ListConfig(list(vocabulary.values()))
+            cfg.decoder.vocabulary = ListConfig(list(vocabulary.values()))
 
         # Override number of classes if placeholder provided
-        if "params" in cfg.decoder:
-            num_classes = cfg.decoder["params"]["num_classes"]
-        else:
-            num_classes = cfg.decoder["num_classes"]
+        num_classes = cfg.decoder["num_classes"]
 
         if num_classes < 1:
             logging.info(
@@ -80,10 +74,7 @@ class EncDecCTCModelBPE(EncDecCTCModel, ASRBPEMixin):
                     num_classes, len(vocabulary)
                 )
             )
-            if "params" in cfg.decoder:
-                cfg.decoder["params"]["num_classes"] = len(vocabulary)
-            else:
-                cfg.decoder["num_classes"] = len(vocabulary)
+            cfg.decoder["num_classes"] = len(vocabulary)
 
         super().__init__(cfg=cfg, trainer=trainer)
 
@@ -207,10 +198,7 @@ class EncDecCTCModelBPE(EncDecCTCModel, ASRBPEMixin):
         decoder_config = copy.deepcopy(self.decoder.to_config_dict())
         decoder_config.vocabulary = ListConfig(list(vocabulary.values()))
 
-        if "params" in decoder_config:
-            decoder_num_classes = decoder_config['params']['num_classes']
-        else:
-            decoder_num_classes = decoder_config['num_classes']
+        decoder_num_classes = decoder_config['num_classes']
 
         # Override number of classes if placeholder provided
         logging.info(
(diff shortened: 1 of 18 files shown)


nemo/collections/asr/models/ctc_bpe_models.py
```diff
@@ -63,27 +65,18 @@ def __init__(self, cfg: DictConfig, trainer=None):
 
         # Set the new vocabulary
         with open_dict(cfg):
-            if "params" in cfg.decoder:
-                cfg.decoder.params.vocabulary = ListConfig(list(vocabulary.values()))
-            else:
-                cfg.decoder.vocabulary = ListConfig(list(vocabulary.values()))
+            cfg.decoder.vocabulary = ListConfig(list(vocabulary.values()))
```
tkornuta-nvidia, line 68: OMG... We really had those cases all over the models...? wow...

nemo/collections/asr/models/rnnt_bpe_models.py
```diff
@@ -62,6 +62,8 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
                 "container that supports RNN-T loss."
             )
 
+        cfg = model_utils.setup_model_config(cfg)
```
tkornuta-nvidia, line 65: what is the difference between `setup_model_config` and `convert_model_config`?

tkornuta-nvidia: (guess will learn that soon ;))

nemo/core/classes/common.py
```diff
@@ -274,6 +277,8 @@ def to_config_dict(self) -> DictConfig:
             config = OmegaConf.create(config)
             OmegaConf.set_struct(config, True)
 
+            config = convert_model_config(config)
+
```
tkornuta-nvidia, line 281: hmmmm, I somehow think this is redundant... but maybe it is good to have it just in case also when saving config...

nemo/collections/asr/models/rnnt_bpe_models.py
titu1994: Im bad at naming things. But basically setup converts {DictConfig, ModelPTConfig dataclass} to DictConfig (which earlier was hardcoded into ModelPT and problematic for cases like *BPE where we need to inject stuff before actually callin super()).

convert will take a config and map it to Hydra 1.x config if necessary.

nemo/core/classes/common.py
titu1994: Redundant is better than missing some weird edge case - its a no-op anyway if it finds an already new type config

nemo/collections/asr/models/ctc_bpe_models.py
titu1994: Yeah, Vahid took a lot of pain to add these. Glad they can go away (hopefully)

nemo/utils/model_utils.py
```diff
@@ -340,3 +341,76 @@ def wrap_training_step(wrapped, instance: pl.LightningModule, args, kwargs):
         instance.log_dict(log_dict, on_step=True)
 
     return output_dict
+
+
+def setup_model_config(cfg: Union[DictConfig, 'ModelPTConfig']) -> DictConfig:
```
tkornuta-nvidia, line 346: ok, got it :)

tests/core/test_serialization.py
```diff
@@ -20,7 +20,7 @@
 from nemo.core.classes.common import Serialization
 
 
-class SerializationTest:
+class TestSerialization:
```
tkornuta-nvidia, line 23: Good catch! Those tests were skipped before, right?

titu1994: Yep, I tried running it locally and it skipped all of them 😓

tkornuta-nvidia: I have covered more cases in EFF:
https://gitlab-master.nvidia.com/tlt/eff/-/blob/main/tests/test_nemo_convert.py

@titu1994 You can simply grab them from there if you want 👍

nemo/collections/asr/models/rnnt_bpe_models.py
tkornuta-nvidia: so how about simply go with `to_dict_config()`?

tkornuta-nvidia: ehehe, it be too close to `to_config_dict()` that we already have in Serialization... :]

tkornuta-nvidia (approved):
LGTM, good changes. Please merge once CI passed.

Ps. After the mind-bending experience of spending few nights on fixing part of the logic in one of NLP models I am for any change that is simplifying any part of any logic in any Model(s) init() ;)

nemo/collections/asr/models/rnnt_bpe_models.py
titu1994: > to_config_dict()
Yep thats why I avoided that name, safer to have "common" things inside model_utils.py

tkornuta-nvidia:
Hey @titu1994 I am trying to deal with some circular/lazy imports in EFF and noticed in my NeMo Cookbook unit tests I have 
`asr_model()` fixture with old structure - and I took this fixture from NeMo.

So this is a reminder - please update the structure (params etc.) also down there :)

titu1994:
I actually left the unit tests intact in order to properly test that the convertion occurs properly - ie old model config to new model config. I think it's easier to replicate the fixture in EFF ? At least until all domains of Nemo have moved into new config style ?

ericharper:
NLP Jenkins test changes look to me.

Thanks!

tkornuta-nvidia:
> I actually left the unit tests intact in order to properly test that the convertion occurs properly - ie old model config to new model config. I think it's easier to replicate the fixture in EFF ? At least until all domains of Nemo have moved into new config style ?

EFF is not the problem here, what I think is this might be misleading to some people, i.e. when they will see this won't they simply think "ok, let me copy and adapt"... (I did this in EFF, so you know ;) )

titu1994:
Let's update it once we have uniformly updated all configs.

nemo/collections/asr/models/ctc_bpe_models.py
```diff
@@ -52,6 +52,10 @@ def list_available_models(cls) -> Optional[PretrainedModelInfo]:
         return result
 
     def __init__(self, cfg: DictConfig, trainer=None):
+        # Convert to Hydra 1.0 compatible DictConfig
+        cfg = model_utils.setup_model_config(cfg)
```
okuchaiev, line 56: (1) are these two always called in sequence one after another?
(2) should every child of ModelPT do these calls now?

titu1994: (1)

The first call is required if the Model needs to change something's before calling super().init(). (aka inside BPE models that change decoder tokens on the fly). Most subclasses of ModelPT won't need to explicitly call this as super () will do it for them. 

The second fall needs to be done after the first one (and is done by ModelPT as well). But for models that need this before actually falling super() it needs to be done manually. 

(2) Very special cases are required to call these two manually. Most models can do without them (aka rest or entire CTC/RNNT for char based, NLP, NMT and TTS)

blisc (approved):

PR #1514, opened 2020-12-03
Fix quantization dependency

skyw:
There is only one python version of prebuild wheel of TensorRT's pytorch-quantization on nvidia pyindex, which causes install with all other version of python to fail with a confusion error message.

Removed requirement_quantization.txt. Instead, install from source in Dockerfile. 

Also added error message if running quantization scripts without pytorch-quantization.

@okuchaiev

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,13 @@ WORKDIR /tmp/nemo
 COPY requirements .
 RUN for f in $(ls requirements/*.txt); do pip install --disable-pip-version-check --no-cache-dir -r $f; done
 
+# install quantization support
+RUN git clone https://github.com/NVIDIA/TensorRT.git && \
+    cd TensorRT/tools/pytorch-quantization && \
+    python setup.py install && \
+    cd - && \
+    rm -rf TensorRT
+
 # build CTC beam search decoder
 COPY scripts/install_ctc_decoders.sh .
 RUN ./install_ctc_decoders.sh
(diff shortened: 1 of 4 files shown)


examples/asr/speech_to_text_quant_infer.py
```diff
@@ -21,13 +21,21 @@
 from pprint import pprint
 
 import torch
-from pytorch_quantization import nn as quant_nn
 from pytorch_quantization import quant_modules
```
titu1994, line 24: There's an additional import here

skyw: black moved things around, the check was on the first import. will move all the pytorch_quantization import under the try clause.

Dockerfile
```diff
@@ -73,6 +73,13 @@ WORKDIR /tmp/nemo
 COPY requirements .
 RUN for f in $(ls requirements/*.txt); do pip install --disable-pip-version-check --no-cache-dir -r $f; done
 
+# install quantization support
+RUN git clone https://github.com/NVIDIA/TensorRT.git && \
```
okuchaiev, line 77: @skyw do you want to take a specific commit here. Or you are confident that TOT will be fine?

skyw: TOT should be fine. It is only updated once a month, together with other DLFW update.

okuchaiev (approved):

PR #1523, opened 2020-12-05
changes to NLP config for ejrvs

ekmb:


nemo/collections/nlp/models/nlp_model.py
```diff
@@ -229,9 +237,6 @@ def setup(self, stage: str) -> None:
         # TODO: implement model parallel for test stage
         if stage == 'fit':
 
-            # adds self.bert_model config to .nemo file
-            self.register_bert_model()
```
ericharper, line 233: I'm not sure why you want to remove this. I made it so that any model using NLP Model will automatically register the bert_model config if it's a Megatron or HF BERT.  It's grabbing the config from the instantiated encoder so the config should be the correct one even if the user specified the config manually.

ericharper (approved):
LGTM, thanks!

PR #1526, opened 2020-12-07
Megatron-lm checkpoints need to have version set.

ericharper:
Checking for version in restore and then setting it. 

Need to save checkpoint version when saving NeMo Megatron-lm models.

### DIFF
--- a/nemo/collections/nlp/modules/common/megatron/megatron_bert.py
+++ b/nemo/collections/nlp/modules/common/megatron/megatron_bert.py
@@ -15,10 +15,10 @@
 # limitations under the License.
 
 import os
-from megatron.checkpointing import set_checkpoint_version
 
 import torch
 from megatron import get_args, initialize_megatron
+from megatron.checkpointing import set_checkpoint_version
 from megatron.model import get_language_model
 from megatron.model.bert_model import bert_attention_mask_func, bert_extended_attention_mask, bert_position_ids
 from megatron.mpu import get_model_parallel_group, model_parallel_is_initialized
--- a/nemo/collections/nlp/modules/common/megatron/megatron_bert.py
+++ b/nemo/collections/nlp/modules/common/megatron/megatron_bert.py
@@ -143,6 +143,7 @@ class MegatronBertEncoder(BertModule):
                 if state_dict['checkpoint_version'] is not None:
                     set_checkpoint_version(state_dict['checkpoint_version'])
             else:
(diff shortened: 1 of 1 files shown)


nemo/collections/nlp/modules/common/megatron/megatron_bert.py
```diff
@@ -151,6 +157,11 @@ def restore_weights(self, restore_path: str):
                 mp_restore_path = f'{restore_path}/mp_rank_{model_parallel_rank:02d}/model_optim_rng.pt'
                 logging.info(f'Restoring model parallel checkpoint from: {mp_restore_path}')
                 state_dict = torch.load(mp_restore_path, map_location='cpu')
+                if 'checkpoint_version' in state_dict:
+                    if state_dict['checkpoint_version'] is not None:
+                        set_checkpoint_version(state_dict['checkpoint_version'])
+                else:
+                    set_checkpoint_version(0)
```
ekmb, line 164: Is the `checkpoint_version` added to the state_dict during this call?

nemo/collections/nlp/modules/common/megatron/megatron_bert.py
```diff
@@ -151,6 +157,11 @@ def restore_weights(self, restore_path: str):
                 mp_restore_path = f'{restore_path}/mp_rank_{model_parallel_rank:02d}/model_optim_rng.pt'
                 logging.info(f'Restoring model parallel checkpoint from: {mp_restore_path}')
                 state_dict = torch.load(mp_restore_path, map_location='cpu')
+                if 'checkpoint_version' in state_dict:
+                    if state_dict['checkpoint_version'] is not None:
+                        set_checkpoint_version(state_dict['checkpoint_version'])
```
ekmb, line 162: Should we throw an error/warning if this check fails?

ekmb (approved):

nemo/collections/nlp/modules/common/megatron/megatron_bert.py
ericharper: No, it's not, I'm still trying to figure out how to persist it when we save on the NeMo side.

nemo/collections/nlp/modules/common/megatron/megatron_bert.py
ericharper: Yes, i'll add a warning message noting that we're defaulting to checkpoint_version=0.

PR #1527, opened 2020-12-07
Update conformer with bug fixes.

VahidooX:


titu1994 (changes requested):
Could you revert the changes related to WER for the time being? That needs further discussion and is not necessary for bug fixes to Conformer model itself. Everything else related to conformer modules looks good to me.

VahidooX:
> 
> 
> Could you revert the changes related to WER for the time being? That needs further discussion and is not necessary for bug fixes to Conformer model itself. Everything else related to conformer modules looks good to me.

Dropped that part but keep in mind that predictions_len comes from the output of the decoder. So if a model return that, it is used otherwise it is ignored. For some models like encoder/decoder with cross attention it should return None for instance. So it would just used for the models who returns the valid length of their outputs.

titu1994:
> Dropped that part but keep in mind that predictions_len comes from the output of the decoder. So if a model return that, it is used otherwise it is ignored. For some models like encoder/decoder with cross attention it should return None for instance. So it would just used for the models who returns the valid length of their outputs.

All CTC based encoders return this length value, which would affect WER calculations wrt QuartzNet/Jasper as well. In the future, yes we can add this but there needs to be a special flag passed by only certain models - say Attention based models - which then submits the lengths for masking to the WER decoding. This is so that unrelated models such as QN/Jasper are not affected by this masking during wer.

nemo/collections/asr/models/ctc_models.py
```diff
@@ -393,7 +393,7 @@ def forward(
     def training_step(self, batch, batch_nb):
         signal, signal_len, transcript, transcript_len = batch
         if isinstance(batch, DALIOutputs) and batch.has_processed_signal:
-            log_probs, encoded_len, predictions = self.forward(
+            log_probs, predictions_len, predictions = self.forward(
```
titu1994, line 396: This needs to remain encoded_len

titu1994 (commented):
One last change, then good to go.

nemo/collections/asr/models/ctc_models.py
VahidooX: Updated.

titu1994 (approved):
LGTM

PR #1531, opened 2020-12-08
Text classification label added.

VahidooX:


examples/nlp/text_classification/conf/text_classification_config.yaml
```diff
@@ -50,6 +50,9 @@ model:
     num_output_layers: 2
     fc_dropout: 0.1
 
+  class_labels:
+    class_labels_file : null # optional to specify a file containing the list of the labels
```
tkornuta-nvidia, line 54: yes! :)

examples/nlp/text_classification/data/import_datasets.py
```diff
@@ -64,9 +64,62 @@ def process_imdb(infold, outfold, uncased, modes=['train', 'test']):
                     review = review.lower()
                 review = review.replace("<br />", "")
                 outfiles[mode].write(f'{review}\t{label}\n')
+
     for mode in modes:
         outfiles[mode].close()
 
+    class_labels_file = open(os.path.join(outfold, 'label_ids.tsv'), 'w')
+    class_labels_file.write('negative\npositive\n')
+    class_labels_file.close()
```
tkornuta-nvidia, line 73: ++

tkornuta-nvidia: so is mapping (0:negative) correct?

tkornuta-nvidia:
@VahidooX to discuss:
1. should we make class_labels_file mandatory? and it so, then:
2. could we get rid of num_classes and get it as len(labels)  when loaded from class_labels_file?

The PR looks good and we can absorb it to TLT->JARVIS the way it is , but I am just thinking loud here...

tkornuta-nvidia:
2nd set of thoughts:

1) in fine-tuning we are effectively changing the dataset, hence should change the "class labels" too. We are using the `setup_*_ds`  for that purpose. Do you think those methods should re-load/overwrite those sections in model?
2) should we leverage that information in inference (`classifytext`) ?

examples/nlp/text_classification/data/import_datasets.py
tkornuta-nvidia: phew :D

tkornuta-nvidia (approved):

tkornuta-nvidia (approved):

tkornuta-nvidia (approved):

PR #1535, opened 2020-12-08
Fix max step calculation

blisc:
Addresses https://github.com/PyTorchLightning/pytorch-lightning/issues/5029


titu1994 (approved):
Fantastic, thanks Jason!

PR #1536, opened 2020-12-09
VAD and ASR LM beam-search notebook fixes

ShantanuNair:
- Removed unnecessary pyaudio stream from breaking VAD offline example
- Fixed issue in timestamp calculation for boundaries of contiguous of space (" ") characters. It did not account for blank tokens that could appear after a space, increasing the boundary.


vsl9 (approved):
Thanks a lot for the PR!

PR #1543, opened 2020-12-10
--result-dir enabled in TLT

tkornuta-nvidia:
Modifications enabling TLT text classification train to use ExpManager

- [x] added postfix to ckpt callback structured config + its handling
- [x] minimally polished logic for explicit_log_dir
- [x] minor polish of names
- [x] Bumped versions in requirements: omegaconf, hydra to make Optional[List]] work
- [x] Bumped version of PTL to 1.1 to enable CheckpointConnector overload in TLT


nemo/utils/exp_manager.py
```diff
@@ -66,7 +66,8 @@ class CallbackParams:
     mode: Optional[str] = "auto"
     period: Optional[int] = 1
     prefix: Optional[str] = None  # If None, exp_manager will attempt to handle the filepath
-    nemo_save_best_model: bool = False
+    postfix: str = ".nemo"
+    save_best_model: bool = False
```
blisc, line 71: Why rename this? Actually in TLT, why are you using `NeMoModelCheckpoint`? Why not just use Lightning's `NeMoModelCheckpoint`

nemo/utils/exp_manager.py
```diff
@@ -538,19 +542,24 @@ class NeMoModelCheckpoint(ModelCheckpoint):
     """ Light wrapper around Lightning's ModelCheckpoint to force a saved checkpoint on train_end
```
blisc, line 542: In TLT, why are you using `NeMoModelCheckpoint`? Why not just use Lightning's `ModelCheckpoint`? Do you need the final pl_module.save_to() call? Are the "ckpt" saved by Lightning not enough?

blisc (commented):
See review

nemo/utils/exp_manager.py
tkornuta-nvidia: PTL chkpt 1) doesn't have artifacts (that I need), but 2) contains also trainer/optim params (I do not need).

PTL ckpt is not EFF archive.

blisc: All `NeMoModelCheckpoint` does is call `save_to` on `on_train_end`.

It DOES NOT save model checkpoints during training. This is delegated to Lightning's callback which saves Lightning's "ckpt"s

> PTL ckpt is not EFF archive.

It is my understanding that you are patching this so that they are.

tkornuta-nvidia: > All NeMoModelCheckpoint does is call save_to on on_train_end.

Exactly - and  in TLT I am relying on  NeMoModelCheckpoint  on_train_end -> calls model save -> calls EFF NeMoCookbook save.

blisc (approved):

PR #1551, opened 2020-12-12
Correcting couple of bugs in example an tutorial

vladgets:


tkornuta-nvidia (approved):
@vladgets Just to make sure - I do not see any changes in the Intent and Slot model per se, so I guess this won't need to be integrated/won't have any impact on TLT, correct?

okuchaiev:
@vladgets  can you please fix CI on this PR?

vladgets:
Correcting it. The problem was in the previous commits exp_manager was removed from testing my model in Jenkins file. And my code was relying for it to save the model.

PR #1559, opened 2020-12-14
Sphinx Bibtext Fix

blisc:


docs/update_docs.sh
```diff
@@ -1,3 +1,4 @@
 rm -rf build
 make clean
 make html
+make html
```
blisc, line 4: Kinda of a "dumb" fix. Version 2.0.0 of sphinxcontrib-bibtex requires users to either build sphinx twice or add "bibtext.json" to the version control.
Note that this fix will not fix the RTD build, which is equally as problematic.

blisc: Alternatively, we can just pin sphinxcontrib-bibtex==1.0.0. @tkornuta-nvidia thoughts?

tkornuta-nvidia: @blisc those changes are good for local build, but as you pointed out won't solve RTD build - so be effectively loose access to the latest documentation. So how about pinning to `sphinxcontrib-bibtex<2.0.0` as suggested:
https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/214#issuecomment-744760613

blisc: I am ok with either solution:
1) Pin `sphinxcontrib-bibtex<2.0.0`
2) Add `bibtext.json` to github
@okuchaiev, thoughts?

okuchaiev: @blisc up to you

tkornuta-nvidia (approved):

PR #1563, opened 2020-12-15
Adding the feature of variational noise to the RNNT-based models

VahidooX:
It is going to add Gaussian noise to the decoder's weights of rnnt models after each step to avoid overfitting.
This section need to be added to the model section of the config:

```
model:
    var_noise:
      std: 0.075
      start_step: 2000
```

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -48,7 +48,6 @@ try:
 except ImportError:
     _EFF_PRESENT_ = False
 
-
 __all__ = ['ModelPT']
 
 """
@@ -68,7 +67,7 @@ _MODEL_RESTORE_PATH:
     artifact objects.
     If it is an archive file, during restoration, the cwd will be temporarily moved to inside the
     archive itself.
-    
+
 _MODEL_EFF_SAVE:
     A global flag that switches the format of the archive file that will be stored.
     This flag only enables EFF when the package support is available.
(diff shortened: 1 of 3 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -791,6 +793,14 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
         if lr is not None:
             optimizer_args['lr'] = lr
 
+        if 'var_noise' in optimizer_args:
+            self._var_noise_std = optimizer_args['var_noise'].get('std', None)
+            self._var_noise_start = optimizer_args['var_noise'].get('start_step', 0)
```
titu1994, line 798: Rename these to `_optim_variational_noise_std` and `_optim_variational_noise_start` since its part of optim inside ModelPT

nemo/core/classes/modelPT.py
```diff
@@ -791,6 +793,14 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
...
+            optimizer_args.pop('var_noise')
+        else:
+            self._var_noise_std = None
+            self._var_noise_start = 0
```
titu1994, line 802: Same rename here

titu1994 (commented):
Looks good for now.

titu1994 (approved):
LGTM

Issue #1561, opened 2020-12-15
Nemo ASR silently trims silence on valid/test set


rbracco:
**Describe the bug**

My assumption would be that `trim_silence` would be False for valid/test unless explicitly turned on. It appears that that is the intention [here in the CTC Models Config](https://github.com/NVIDIA/NeMo/blob/f0378b10cdadc82752d70032bd80219e531519c0/nemo/collections/asr/models/configs/ctc_models_config.py#L66) is to have `trim_silence=True` for training but not validation. Also all the ASR configs for the tutorials have an explicit `trim_silence=True` for train and no entry for validation, implying that the default is False.

I believe the default is set to True [here in audio_to_text_dataset.py](https://github.com/NVIDIA/NeMo/blob/c0fedd91e316a4ba95ac6b14330ff92844c56097/nemo/collections/asr/data/audio_to_text_dataset.py#L46).

**Steps/Code to reproduce bug**

Print out the duration of the audio in the validation loop for some dataset, then set `params["model"]["validation_ds"]["trim_silence"] = False` and try it again and some of the audios will be of different length because they are no longer being trimmed.

**Expected behavior**

Validation silence trimming defaults to False, or the documentation changes to make it more explicit that silence is being trimmed during validation/inference. 

**Environment overview (please complete the following information)**

 -Colab + Nemo 1.0.0b2

titu1994:
Hey @rbracco, thanks for pointing this out! I submitted a PR to patch this bug.

rbracco:
Thank you for all that you do for the repo!

PR #1565, opened 2020-12-16
Correct ASR issues + Patch for Pytorch 1.8

titu1994:
# Changelog
- `trim_silence` will now be False by default for all dataloaders in ASR and only active for train_ds (set by flag in config).
- Update ASR stft, ifft ops to be compatible with Pytorch 1.8 deprecation warnings

Note:
- Due to change in `trim_silence`, scores obtained via trainer.test(), or model.transcribe() or any other evaluation process will change slightly. We expect only minor changes to WER due to this bugfix.

Closes https://github.com/NVIDIA/NeMo/issues/1561


titu1994:
@blisc Could you review the TTS side changes to `stft` and `torch.ifft`?

blisc (approved):

jbalam-nv (approved):
LGTM

mohsenoon:
trim_silence default in the model.transcribe method is still True!
This issue can be seen here:
collections/asr/models/ctc_models.py: _setup_transcribe_dataloader

PR #1569, opened 2020-12-16
change squad output format

yzhang123:

### DIFF
--- a/tutorials/nlp/Question_Answering_Squad.ipynb
+++ b/tutorials/nlp/Question_Answering_Squad.ipynb
@@ -46,7 +46,7 @@
         "id": "uRLPr0TnIAHO"
       },
       "source": [
-        "BRANCH = 'v1.0.0b3'"
+        "BRANCH = 'main'"
       ],
       "execution_count": null,
       "outputs": []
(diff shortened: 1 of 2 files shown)


tutorials/nlp/Question_Answering_Squad.ipynb
```diff
@@ -46,7 +46,7 @@
         "id": "uRLPr0TnIAHO"
       },
       "source": [
-        "BRANCH = 'v1.0.0b3'"
+        "BRANCH = 'main'"
```
tkornuta-nvidia, line 49: I am not sure what is the NeMo policy for the notebooks now... @okuchaiev ?

yzhang123: this change will not work in the Notebook if this is not changed

okuchaiev: looks good to me, see Evelina's PR on notebook updates

tkornuta-nvidia (approved):

tutorials/nlp/Question_Answering_Squad.ipynb
okuchaiev: I'd say it's ok to have an exception like this. But just keep in mind that this makes this notebook somewhat vulnerable because it is loaded from main

PR #1577, opened 2020-12-17
Speaker Diarization pipeline

nithinraok:


nemo/collections/asr/models/clustering_diarizer.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+__all__ = ['ClusteringDiarizer']
+
+_MODEL_CONFIG_YAML = "model_config.yaml"
+_VAD_MODEL = "vad_model_.nemo"
+_SPEAKER_MODEL = "speaker_model.nemo"
+
+
+class ClusteringDiarizer(Model, DiarizationMixin):
```
jbalam-nv, line 60: @okuchaiev @titu1994 Here are the changes that were suggested to make this "model" inherit from Model instead of ModelPT. Please take a look when you get a chance. I have tested saving and restoring functionalities and they work fine.

PR #1580, opened 2020-12-18
Correcting few corner case in validation evaluation during training

vladgets:


nemo/collections/nlp/data/question_answering_squad/qa_dataset.py
```diff
@@ -136,6 +130,14 @@ def __init__(
                     with open(cached_features_file, "wb") as writer:
                         pickle.dump(self.features, writer)
 
+        # check number of samples. Should be either -1 not to limit or positive number
+        if num_samples == 0:
+            raise ValueError(
```
yzhang123, line 135: Could you move this section up before line 110, then you can remove index checks down below

yzhang123 (changes requested):

nemo/collections/nlp/data/question_answering_squad/qa_dataset.py
tkornuta-nvidia: wait... you mean: before use_cache?

yzhang123: I think its best to move up line 134- 139 before line 110

yzhang123 (approved):
Thank you @vladgets !

nemo/collections/nlp/data/question_answering_squad/qa_dataset.py
vladgets: Corrected

PR #1587, opened 2020-12-23
fix container version - no leading v

junkin:
readme has wrong container, there is no leading v on the version for the container on NGC.


okuchaiev (approved):
Thanks!

PR #1593, opened 2020-12-28
Megatron model parallel autoresume

ericharper:
Adds training auto-resume for NLP models that use model parallel megatron-lm. The existing autoresume logic depends on there being only one checkpoint, this is not the case for model parallel models. There is a checkpoint for each model parallel rank.

We were able to add this feature with minimal changes to exp_manager. The only functional change is to log a warning rather than throw an error when exp_manager detects multiple `last` checkpoints. We also use the exp_manager to set some properties in `AppState` that we need in `NLPModel`.

# Usage
```
python examples/nlp/text_classification/text_classification_with_bert.py \
	--config-path=${CONFIG_PATH} \
	--config-name=${CONFIG_NAME} \
        ...
	exp_manager.exp_dir=${EXP_DIR}/nemo_experiments \
	exp_manager.create_checkpoint_callback=true \
	+exp_manager.create_wandb_logger=true \
	+exp_manager.wandb_logger_kwargs.name=${EXP_NAME} \
	+exp_manager.wandb_logger_kwargs.project=${PROJECT_NAME} \
	exp_manager.name=${EXP_NAME} \
	+exp_manager.resume_if_exists=true \
	+exp_manager.resume_ignore_no_checkpoint=true \
	+exp_manager.version=0 \
	+exp_manager.use_datetime_version=false \
```


nemo/utils/exp_manager.py
```diff
@@ -309,8 +321,9 @@ def check_resume(
     elif len(end_checkpoints) > 0:
         if resume_past_end:
             if len(end_checkpoints) > 1:
-                raise ValueError(f"Multiple multiple checkpoints {end_checkpoints} that matches *end.ckpt.")
-            logging.info(f"Resuming from {end_checkpoints[0]}")
+                logging.warning(
+                    f"Multiple checkpoints {end_checkpoints} that matches *end.ckpt. This should only happen if using model parallelism."
+                )
```
blisc, line 326: Is it possible to add a check to see if this is a model parallel checkpoint? Maybe even a dumb check if "mp_rank" in checkpoint_name OR if AppState has model parallel enabled.
I am weary of making a change in behaviour without a deprecation period.

blisc (changes requested):

nemo/utils/exp_manager.py
ericharper: The issue is that exp_manager is called before DDP is initialized. So there's no awareness of model parallelism at this point.

ericharper: Your'e right though, we can check if mp_rank is there. I'll add that.

blisc: Another alternative is to force model parallel users to set resume_past_end to True since you are effectively ignoring this flag when running in model parallel.

blisc (approved):

PR #1599, opened 2021-01-04
Patch ASR to support < PT 1.7

titu1994:
# Changelog
- Wrap torch.stft to switch signature between  PT< 1.7 and >= 1.7

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,6 +29,12 @@ pipeline {
       }
     }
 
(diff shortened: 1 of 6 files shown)


blisc:
Can we add the following?
1) A reminder to remove this patch once we require pytorch >= 1.7
2) A unittest that fails or gives a warning if a PR contains "torch.stft" while we still allow pytorch < 1.7

blisc (approved):

Issue #1606, opened 2021-01-06
ConfigKeyError: str interpolation key 'model.language_model.pretrained_model_name' not found


amirj:
I'm trying to reproduce [Named Entity Recognition - BioMegatron colab](https://colab.research.google.com/github/NVIDIA/NeMo/blob/r1.0.0b3/tutorials/nlp/Token_Classification-BioMegatron.ipynb), everything is fine until the following cell:

`model_ner = nemo_nlp.models.TokenClassificationModel(cfg=config.model, trainer=trainer)`

Output:
```
[NeMo I 2021-01-06 10:36:00 megatron_utils:235] Downloading from https://api.ngc.nvidia.com/v2/models/nvidia/biomegatron345mcased/versions/0/files/vocab.txt
Downloading: 100%
625/625 [00:00<00:00, 1.31kB/s]

Downloading: 100%
213k/213k [00:00<00:00, 3.31MB/s]
Using bos_token, but it is not set yet.
Using eos_token, but it is not set yet.

---------------------------------------------------------------------------
ConfigKeyError                            Traceback (most recent call last)
<ipython-input-25-4645443f78bc> in <module>()
----> 1 model_ner = nemo_nlp.models.TokenClassificationModel(cfg=config.model, trainer=trainer)

8 frames
/usr/local/lib/python3.6/dist-packages/omegaconf/base.py in _resolve_simple_interpolation(self, key, inter_type, inter_key, throw_on_missing, throw_on_resolution_failure)
    326                 if throw_on_resolution_failure:
    327                     raise ConfigKeyError(
--> 328                         f"{inter_type} interpolation key '{inter_key}' not found"
    329                     )
    330                 else:

ConfigKeyError: str interpolation key 'model.language_model.pretrained_model_name' not found
```

I believe that `config.model.language_model.pretrained_model_name = 'biomegatron-bert-345m-cased'`. In addition, I'm getting an error in the installation phase though I'm not sure it would be related to this issue:

```
# install NeMo
BRANCH = 'main'
!python -m pip install git+https://github.com/NVIDIA/NeMo.git@$BRANCH#egg=nemo_toolkit[nlp]
```

Output:
```
Collecting nemo_toolkit[nlp]
  Cloning https://github.com/NVIDIA/NeMo.git (to revision main) to /tmp/pip-install-hz__8v6m/nemo-toolkit
[...]

PR #1610, opened 2021-01-07
Patch colab omegaconf resolution

titu1994:
# Bugfix
- Resolution for dataclass somehow interferes with key interpolation
- Patch via check for type explicitly

Resolves https://github.com/NVIDIA/NeMo/issues/1606
@ekmb for review


ekmb (approved):
Thanks!

Issue #1549, opened 2020-12-11
Basic support for additional audio file formats (.mp3, .flac...etc)


rbracco:
**Is your feature request related to a problem? Please describe.**

It appears that NeMo only works with wav files, but they are 10x larger than equivalent MP3s. Many popular datasets like Mozilla Commonvoice are stored as MP3 and have to be converted before being used. This is time consuming and also makes it harder for those using limited resources like Google Colab. 

I was wondering if there's a specific reason that only WAV is used. 

**Describe the solution you'd like**

For NeMo to work with non wav files.

rbracco:
Any word on this @titu1994? I'm curious if theres a specific reason. I'm starting to use CommonVoice English and it's 56GB when stored as MP3 so it will be unusable in wav format. Thank you!

titu1994:
@vsl9 Any thoughts about this ?

mthrok:
Hi

torchaudio developer here. Recently [I am adding support for file-like object in torchaudio's I/O functions](https://github.com/pytorch/audio/pull/1158), and I am wondering what extra features would be required for NeMo to consider using `torchaudio.load`. In #1616, I see that tarfile support is one, which is working now. Do you have any other requirements? What about Windows support?

titu1994:
@mthrok Sorry for the delay, just saw the previous #1616 conversation regarding this. wav, tarfiles via webdataset, and mp3 are currently supported, but it would be nice for torchaudio to also support FLAC if possible, since MLS dataset is in that format - http://www.openslr.org/94/. 

The current reason of why we don't add torchaudio as an explicit dependency, and instead make it optional for certain modules, is the fact that torchaudio previously pinned it's version to a specific pytorch version. I am unsure of whether this is still the case. 

This is generally fine for day to day use, but interferes with containers that utilize commit hash version of pytorch (Torchaudio pinning forces reinstallation of pytorch). This issue goes away by building torchaudio from main branch, but building torchaudio is a time-consuming process which we don't want to constantly do on the fly. 

One way to alleviate this is to have the Nemo container build torchaudio, but our ci uses pytorch container and pip installs the Nemo library in order to simulate user workflow  - and torchaudios pinned python version interfered with this. 

We would be very happy to add torch audio data loading as an option for various audio formats, but it will have to remain as an optional dependency for users until the pinned version issue is resolved.

mthrok:
[...]

PR #1616, opened 2021-01-09
support other audio format

khursani8:
Might solve #1549 

Currently I let pydub handle file that is not .wav format because there is something I'm not understand with current implementation.
1. "samples.transpose()" before init with cls
2. User also need to install [ffmpeg](https://github.com/jiaaro/pydub#getting-ffmpeg-set-up) to be able to load non wav file format, I don't know where to put the warning or write a script when user pip install
3. I'm not sure if NeMo dataloader will cache the loaded samples because it might convert using ffmpeg everytime it try to load non wav file.

So far I only tested inference with mp3 audio from Common Voice with model trained on wav file format, there is no problem with the prediction result.

### DIFF
--- a/nemo/collections/asr/parts/segment.py
+++ b/nemo/collections/asr/parts/segment.py
@@ -123,8 +123,9 @@ class AudioSegment(object):
         :param duration: duration in seconds when loading audio
         :return: numpy array of samples
         """
-        ext = Path(audio_file).suffix
-        if ext == ".wav":
+        if isinstance(audio_file, (str, Path)):
+            ext = Path(audio_file).suffix
+        if not isinstance(audio_file, (str, Path)) or ext == ".wav":
             with sf.SoundFile(audio_file, 'r') as f:
                 dtype = 'int32' if int_values else 'float32'
                 sample_rate = f.samplerate
--- a/nemo/collections/asr/parts/segment.py
+++ b/nemo/collections/asr/parts/segment.py
@@ -34,10 +34,12 @@
 # This file contains code artifacts adapted from https://github.com/ryanleary/patter
 
 import random
+from pathlib import Path
 
 import librosa
 import numpy as np
 import soundfile as sf
+from pydub import AudioSegment as Audio
 
 
 class AudioSegment(object):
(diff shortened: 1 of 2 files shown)


blisc (approved):
Thanks for the PR. It looks good to me.
I just added a slight check to ensure that it will continue to work with io.bufferedreader that we use for tarred datasets.

yoks:
This MR, make training so much slower. Pydub takes too much of CPU to convert media files. Need to make sure pydub is only used for unsupported media formats, right now we using FLAC, and it goes through pydub, but libsnd supports flac.

image

Part in the middle is with MR, and parts before and after are without it. Training done on DGX A100.

blisc:
Ok that's a problem. I'll go ahead and change the logic a bit

khursani8:
Hi @blisc,
I create hotfix for this bug https://github.com/NVIDIA/NeMo/pull/1630

blisc:
@yoks, can you try #1630 and see if performance returns?

yoks:
Will test as soon as we will have free video card available on DGX (tomorrow most likely)

yoks:
Yea looks like it works as before (maybe a little bit slower, but can be related to task/node load).

rbracco:
Is there a reason this isn't using torchaudio? It supports multiple formats in a single library.

khursani8:
I don’t have experience using torchaudio, after take a look at their repo I think torchaudio will be better for this, support multiple audio format including kaldi format. Maybe someone know how to use torchaudio better can contribute to this

mthrok:
> I don’t have experience using torchaudio, after take a look at their repo I think torchaudio will be better for this, support multiple audio format including kaldi format. Maybe someone know how to use torchaudio better can contribute to this

Hi

torchaudio maintainer here. I commented in #1549 but the current release of torchaudio does not support file-like object. I am adding that support right now and if will be released in the next release. (Early March). 
Is there any other requirements that NeMo has?

khursani8:
Hmmm, I'm not related to NeMo actually, I also don't know the requirements, I think @blisc can answer this question.

rbracco:
Theres a [very detailed and helpful response](https://github.com/NVIDIA/NeMo/issues/1549#issuecomment-767058227) from @titu1994 on why torchaudio isn't used as a dependency.

PR #1608, opened 2021-01-07
Workaround for TRT lack of support for DOUBLE

borisfom:
Background: 
https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3221866&cmtNo=

PR #1619, opened 2021-01-11
Add enemo export for Quartznet

purn3ndu:
For direct usage within ServiceMaker (Nemo to Jarvis) without TLT. (Experimental)

### DIFF
--- a/scripts/convasr_to_enemo.py
+++ b/scripts/convasr_to_enemo.py
@@ -1,0 +1,78 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+
+from nemo.collections.asr.models import EncDecClassificationModel, EncDecCTCModel, EncDecSpeakerLabelModel
+from nemo.utils import logging
+import os
+import tarfile
+
+def get_parser():
+    parser = argparse.ArgumentParser(description="Convert .nemo file to onnx file (encoder and decoder fused)")
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


okuchaiev:
@purn3ndu, thanks for the PR. Please fix code formatting, to unblock the CI. Here is how https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md

tkornuta-nvidia (changes requested):
1) the solution is clearly not exporting artifacts, so the resulting files won't be able to be used in JARVIS ServiceMaker (!)

2) if this is export to ONNX - please call the file .eonnx (for ONNX) or .ejrvs (for JARVIS) to be consistent with naming convention of TLT

3) EFF ONNXCookbook contains recipes for creating ONNX archives. Any reason why to duplicate code/solve it in a different way?

scripts/convasr_to_enemo.py
```diff
@@ -0,0 +1,78 @@
...
+        logging.info("Preparing Speech Label Classification model")
+        model = EncDecClassificationModel.restore_from(nemo_file)
+    elif model_type == 'speaker':
+        logging.info("Preparing Speaker Recognition model")
+        model = EncDecSpeakerLabelModel.restore_from(nemo_file)
+    else:
+        raise NameError("Available model names are asr, speech_label and speaker")
+
+    logging.info("Writing onnx file")
+    model.export(onnx_file, onnx_opset_version=12)
```
tkornuta-nvidia, line 58: @purn3ndu a week ago you were asking/pushing for changes in NeMo export defaults as `opset=12` was supposed to be error prone for TensorRT 7.2. What is the status?

tkornuta-nvidia (approved):
Required by client this week, no additional artifacts required, approving.

scripts/convasr_to_enemo.py
purn3ndu: That was specific to Megatron model only. A fix has been worked out by TRT/Jarvis team using Graph surgeon. @borisfom is working on that.

tkornuta-nvidia: ok, thx. So just to make sure - the fix will be on TRT side/Jarvis side only?

borisfom: Well, I filed TRT bug - but that will take time. There is a workaround though, that will go into NeMo (it may also improve quality of exported ONNX) - essentially we do a pass of onnx-surgeon on the graph with constant folding after export and save the result. PR: https://github.com/NVIDIA/NeMo/pull/1608

tkornuta-nvidia: So it is good that I have asked... Do I guess correctly that this change should be "absorbed" in TLT as well?

PR #1620, opened 2021-01-11
Correct onesided for PT < 1.7 compatibility

titu1994:
# Bugfix
- PT 1.6 expects explicit bool type for `onesided` (defaults to true)


fayejf (approved):
LGTM

PR #1626, opened 2021-01-12
Loss metric

PeganovAnton:
Add a PTL metric for averaging loss across processes in ddp mode.

Motivation.
If ddp is used, the validation loss for only one gpu is logged. This behavior impairs "Reduce on plateau" and "Early stopping" scheduling strategies. If many GPUs are used and the validation dataset is small, logged validation loss can differ from real validation loss a lot.

### DIFF
--- a/nemo/collections/common/metrics/loss.py
+++ b/nemo/collections/common/metrics/loss.py
@@ -1,0 +1,73 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from pytorch_lightning.metrics import Metric
+
+__all__ = ['Loss']
+
+
+class Loss(Metric):
+    """
+    This class is for averaging loss across multiple processes if a distributed backend is used. It does not accumulate
+    gradients so the averaged loss cannot be used for optimization. If ``take_avg_loss`` is ``True``, the :meth:`update`
+    method ``loss`` argument has to be a mean loss. If ``take_avg_loss`` is ``False`` then the :meth:`update` method
+    ``loss`` argument has to be a sum of losses.
+
+    See :doc:`PyTorch Lightning Metrics<pytorch-lightning:metrics>` for the metric usage instruction.
+
+    Args:
+        compute_on_step:
+            Forward only calls ``update()`` and returns ``None`` if this is set to ``False``. default: ``True``
+        dist_sync_on_step:
+            Synchronize metric state across processes at each ``forward()``
+            before returning the value at the step.
+        process_group:
+            Specify the process group on which synchronization is called. default: ``None`` (which selects the entire
+                world)
+        take_avg_loss:
+            If ``True`` values of :meth:`update` method ``loss`` argument has to be a mean loss. If ``False``
+            values of :meth:`update` method ``loss`` argument has to be a sum of losses. default: ``True``
+    """
+
+    def __init__(self, compute_on_step=True, dist_sync_on_step=False, process_group=None, take_avg_loss=True):
+        super().__init__(
+            compute_on_step=compute_on_step, dist_sync_on_step=dist_sync_on_step, process_group=process_group
+        )
+        self.add_state("loss_sum", torch.tensor(0.0, dtype=torch.float64), dist_reduce_fx='sum')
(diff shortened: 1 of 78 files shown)


nemo/collections/common/metrics/loss.py
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+from pytorch_lightning.metrics import Metric
+
+__all__ = ['Loss']
+
+
+class Loss(Metric):
```
titu1994, line 21: The name is too generic. It's supposed to be a metric not an actual loss function / class. Please call it something like GlobalAverageLossMetric or something similar to denote that 1) it is a metric, not an actual loss function 2) that it computed true global average and not running average of a metric. 

Same with the filename

tests/collections/common/loss_inputs.py
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+
+import torch
+
+from .pl_utils import BATCH_SIZE, NUM_BATCHES, NUM_CLASSES
+
+LossInput = namedtuple('Input', ["loss_sum_or_avg", "num_measurements"])
```
titu1994, line 35: Can be replaced with a typed dataclass. That would offer more info + docstring to explain what this class is c

tests/collections/common/loss_inputs.py
```diff
@@ -0,0 +1,54 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+
+import torch
+
+from .pl_utils import BATCH_SIZE, NUM_BATCHES, NUM_CLASSES
```
titu1994, line 33: No relative imports. Please do an absolute import everywhere.

tests/collections/common/loss_inputs.py
```diff
@@ -0,0 +1,54 @@
...
+        (
+            torch.randint(1, 100, (NUM_BATCHES // 2,), dtype=torch.int32),
+            torch.zeros(NUM_BATCHES - NUM_BATCHES // 2, dtype=torch.int32),
+        )
+    ),
+)
+
+_all_num_measurements_are_zero = LossInput(
+    loss_sum_or_avg=torch.rand(NUM_BATCHES) * 2.0 - 1.0, num_measurements=torch.zeros(NUM_BATCHES, dtype=torch.int32),
+)
```
titu1994, line 54: If all 3 instantiations are in this classes, why are they private? And since they are global constants in the file, please use all caps in names to designate them as constants (either private or public).

tests/collections/common/loss_inputs.py
titu1994: More so than this, this import is cyclic. Please copy the constants from that file here onto this file to avoid it.

tests/collections/common/pl_utils.py
```diff
@@ -405,3 +405,132 @@ def run_class_perplexity_test(
                 check_batch=check_batch,
                 atol=self.atol,
             )
+
+
+def reference_loss_func(loss_sum_or_avg, num_measurements, take_avg_loss):
+    if take_avg_loss:
```
titu1994, line 411: Add documentation of what this function does

tests/collections/common/test_metrics.py
```diff
@@ -19,8 +19,9 @@
 import pytest
 import torch
 
+from .loss_inputs import _all_num_measurements_are_zero, _no_zero_num_measurements, _some_num_measurements_are_zero
```
titu1994, line 22: Avoid relative imports

titu1994 (changes requested):
Overall looks good. Minor semantic changes.

tests/collections/common/loss_inputs.py
PeganovAnton: I have a problem with using absolute imports. Command `pytest` fails with import error.
```
ImportError while importing test module '/home/apeganov/NeMo/tests/collections/common/test_metrics.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../anaconda3/envs/translation_new_py37/lib/python3.7/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/collections/common/test_metrics.py:22: in <module>
    from tests.collections.commmon.loss_inputs import _all_num_measurements_are_zero, _no_zero_num_measurements, _some_num_measurements_are_zero
E   ModuleNotFoundError: No module named 'tests.collections.commmon'
```
@titu1994 could you please help?

This import is not cyclic in my opinion.

titu1994: Ok seems like pytest wants relative imports. As to cyclic import, since tests are running and only constants are being imported, I guess it's fine. 

Thanks for looking into this c

titu1994:
Overall, the PR looks good to merge. @MaximumEntropy sign off please.

MaximumEntropy:
Looks good to me as well

MaximumEntropy (approved):
Looks good to me as well.

titu1994 (approved):
LGTM

PR #1630, opened 2021-01-13
only use pydub for unsupported format

khursani8:
Signed-off-by: khursani8 <khursani8@gmail.com>

### DIFF
--- a/nemo/collections/asr/parts/segment.py
+++ b/nemo/collections/asr/parts/segment.py
@@ -41,6 +41,8 @@ import numpy as np
 import soundfile as sf
 from pydub import AudioSegment as Audio
 
+available_formats = sf.available_formats()
+sf_supported_formats = ["."+i.lower() for i in available_formats.keys()]
 
 class AudioSegment(object):
     """Monaural audio segment abstraction.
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/parts/segment.py
```diff
@@ -123,21 +128,25 @@ def from_file(
...
-                    f.seek(int(offset * sample_rate))
-                if duration > 0:
-                    samples = f.read(int(duration * sample_rate), dtype=dtype)
-                else:
-                    samples = f.read(dtype=dtype)
-
-            samples = samples.transpose()
-        else:
+        samples = None
+        if not isinstance(audio_file, (str, Path)) or Path(audio_file).suffix.lower() in sf_supported_formats:
```
titu1994, line 132: Why not just use os.path.splitext() here instead of creating path objects. This line is called for every single sample in the training run, any efficiency issue is going to become a bottleneck.

khursani8: Ohhh, I don't know about splitext, I will change the code to use splitext

blisc (approved):

PR #1636, opened 2021-01-15
Make onnx-graphsurgery an optional deferred import

titu1994:
# Changelog
- onnx-graphsurgery is not available on all platforms, and breaks pip install (currently tested on darwin and some versions of ubuntu).
- Make it an optional import that performs runtime check instead.

### DIFF
--- a/nemo/collections/asr/models/classification_models.py
+++ b/nemo/collections/asr/models/classification_models.py
@@ -478,7 +478,7 @@ class EncDecClassificationModel(ASRModel, Exportable):
         output_example=None,
         verbose=False,
         export_params=True,
-        do_constant_folding=True,
+        do_constant_folding: Optional[bool] = None,
         keep_initializers_as_inputs=False,
         onnx_opset_version: int = 12,
         try_script: bool = False,
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -489,7 +489,7 @@ class EncDecCTCModel(ASRModel, Exportable):
         output_example=None,
         verbose=False,
         export_params=True,
-        do_constant_folding=True,
+        do_constant_folding: Optional[bool] = None,
         keep_initializers_as_inputs=False,
         onnx_opset_version: int = 12,
         try_script: bool = False,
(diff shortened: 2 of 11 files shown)


tkornuta-nvidia:
Hmmm, I think the line that was breaking NeMo installation on my MacBook was in `setup.py`:
https://github.com/NVIDIA/NeMo/blob/45f60f03e5d15a4b6e7db145b5d731d73f3a891d/setup.py#L235

@borisfom ideas?

okuchaiev (approved):
lgtm

okuchaiev:
> Hmmm, I think the line that was breaking NeMo installation on my MacBook was in `setup.py`:
> https://github.com/NVIDIA/NeMo/blob/45f60f03e5d15a4b6e7db145b5d731d73f3a891d/setup.py#L235
> 
> @borisfom ideas?

@tkornuta-nvidia Doesn't seem like this PR breaks MacOS compatibility. It looks like it was broken before? This PR change reduces ```setup_requires=['pytest-runner', 'nvidia-pyindex'],``` to ```setup_requires=['pytest-runner']```

tkornuta-nvidia (approved):

titu1994:
Awaiting @borisfom review before merge

titu1994:
Merged as main branch is broken without this, preventing other prs

borisfom:
Sorry to be late for the game - I am on vacation right now ...
Too bad onnx-graphsurgeon is not available on all platforms - we would need to think how to handle those cases properly in general. It makes it rather difficult  to have seamless setup.py, without hairy scripts around. As my original change was primarily to enable a workaround for TRT issue that will hopefiully be fixed, it's probably fine in this case. 
The way it's left after this PR merged is not right though: 
1. constant_folding argument is defunct (always overridden). We should revert changes to it and only have onnx-graphsurgeon invocation under the GRAPGSURGEION_AVAILABLE conditional.
2. default setup.py would not install onnx-graphsurgeon on any platform. That's probably the best we can do at the moment and have developers who need to use this WAR, to install it manually - but we should think of general way of handling optional components to have best experience on all platforms. It sees that, besides WAR for TRT 'double' issue, onnx-graphsurgeon pass makes most of the graphs more compact (potentially faster, one needs to check), so it would probably be nice to keep it enabled for the platforms that have it, even when TRT issue is fixed.

PR #1656, opened 2021-01-22
pre-trained models performance test added and masking fix

ekmb:


nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
```diff
@@ -382,8 +382,8 @@ def create_label_ids(unique_labels, pad_label=pad_label):
         self.all_input_ids = features[0]
         self.all_segment_ids = features[1]
         self.all_input_mask = features[2]
-        self.all_loss_mask = features[3]
-        self.all_subtokens_mask = features[4]
+        self.all_subtokens_mask = features[3]
+        self.all_loss_mask = features[4]
```
tkornuta-nvidia, line 386: wow, how the training could even work with masks messed up...?

ekmb: @tkornuta-nvidia depending on the parameters, the masks could be the same. This fix won't require retraining the models.

tkornuta-nvidia (approved):
LGTM

tests/collections/nlp/test_pretrained_models_performance.py
```diff
@@ -50,51 +50,55 @@ def get_metrics(data_dir, model):
     return metrics
 
 
+def data_exists(data_dir):
+    return os.path.exists(data_dir)
+
+
 class TestPretrainedModelPerformance(TestCase):
     @pytest.mark.unit
```
tkornuta-nvidia, line 58: btw. Why this is unit test? You take a pre-trained model and then verify if it is good enough - isn't it an Acceptance Test?

Issue #1649, opened 2021-01-19
Key 'window_stride' is not in struct


shfaizan:
I am trying to run the[ VAD notebook](https://github.com/NVIDIA/NeMo/blob/9aa4ada92b2e6ddd418704dc392e8576d44e5f41/tutorials/asr/07_Online_Offline_Microphone_VAD_Demo.ipynb). 

I followed every step but got this error:
![image](https://user-images.githubusercontent.com/39994105/105074869-f4a20400-5aae-11eb-85f0-6512c0f56d2a.png)

fayejf:
Thanks for reporting this. We will fix it asap and let you know.

fayejf:
@shfaizan  [Fixed](https://github.com/NVIDIA/NeMo/pull/1659)! Please pull and update. Thanks!

PR #1659, opened 2021-01-23
Fix configs in ASR tutorials

fayejf:
Update and fix the **outside of the model constructor configs** in ASR tutorials. Converting outside Hydra 0.x configs to Hydra 1.x configs.
Tiny bug fix for notebook 03 speech command

Related bug: [1649](https://github.com/NVIDIA/NeMo/issues/1649)

### DIFF
--- a/scripts/write_long_audio_manifest.py
+++ b/scripts/write_long_audio_manifest.py
@@ -19,7 +19,6 @@ from argparse import ArgumentParser
 from itertools import repeat
 
 import librosa
-import numpy as np
 
 
 """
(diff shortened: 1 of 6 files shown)


titu1994 (changes requested):
LGTM but revert branch to b3.

titu1994:
Hmm wait, the function exists in main but the error exists in b3? If the error exists only in main, then for now comment out these patches with a note that it needs to be uncommented

fayejf:
> Hmm wait, the function exists in main but the error exists in b3? If the error exists only in main, then for now comment out these patches with a note that it needs to be uncommented. Or use a try catch.
> 
> Also, tests are failing on a notebook only pr?

@titu1994 the error exists in b3 well the patch function exists in main. Is it safe to change branch in some notebooks to main now?  If it's not, I can revert it back to b3 and fix it by manually accessing params or writing the patch function explicitly in notebooks. 

Right not sure why the tests fails for notebook changes

titu1994:
We will be changing branch later, main is not suitable for the notebooks cause it's not stable and tested as much as the exact branches.

For now, It might be best to do manual fixes by back porting the config updater to the notebooks.

fayejf:
> We will be changing branch later, main is not suitable for the notebooks cause it's not stable and tested as much as the exact branches.
> 
> For now, It might be best to do manual fixes by back porting the config updater to the notebooks.

Sounds great. will update it.

titu1994 (approved):
@fayejf LGTM

PR #1665, opened 2021-01-26
Add preliminary support for ASR Model dataclasses

titu1994:
# Changelog
- Adds nearly concrete implementations of dataclasses for QuartzNet 15x5, Jasper 10x5-DR and MatchboxNet-3x1x64 and VAD modes.
- Adds *ConfigBuilder class supporting the construction of arbitrarily complex interdependencies between sub-configs
- Adds various dataclasses for ASR datasets + ASR Classification datasets
- Adds core support for class-dataclass testing

### DIFF
--- a/examples/asr/experimental/structured/speech_to_text_structured_v2.py
+++ b/examples/asr/experimental/structured/speech_to_text_structured_v2.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import asdict
+
+import pytorch_lightning as pl
+
+from nemo.collections.asr.models import EncDecCTCModel, configs
+from nemo.core.config import optimizers, schedulers
+from nemo.utils.exp_manager import exp_manager
+
+"""
+python speech_to_text_structured_v2.py
+"""
+
+# fmt: off
+LABELS = [
+    " ", "a", "b", "c", "d", "e",
+    "f", "g", "h", "i", "j", "k",
+    "l", "m", "n", "o", "p", "q",
+    "r", "s", "t", "u", "v", "w",
+    "x", "y", "z", "'",
+]
+
+optim_cfg = optimizers.NovogradParams(
+    lr=0.01,
+    betas=(0.8, 0.5),
+    weight_decay=0.001
+)
+
+sched_cfg = schedulers.CosineAnnealingParams(
+    warmup_steps=None,
+    warmup_ratio=None,
+    min_lr=0.0,
+)
+# fmt: on
+
+
+def main():
+    # Generate default asr model config
+    builder = configs.QuartzNetConfigBuilder(name='quartznet_15x5')
+
+    # set global values
+    builder.set_repeat(5)
+    builder.set_labels(LABELS)
+    builder.set_optim(cfg=optim_cfg, sched_cfg=sched_cfg)
+
+    cfg = builder.build()
+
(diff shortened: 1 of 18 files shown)


examples/asr/experimental/structured/speech_to_text_structured_v2.py
```diff
@@ -0,0 +1,91 @@
...
+
+def main():
+    # NeMo Model config
+    cfg = modelPT.NemoConfig(name='Custom QuartzNet')
+
+    # Generate default asr model config
+    builder = configs.EncDecCTCModelConfigBuilder(name='quartznet_15x5')
+
+    # set model global values
+    builder.set_repeat(5)
```
okuchaiev, line 58: shouldn't this already be 5 if config is quartznet_15x5?

okuchaiev: if yes, let's remove this line

nemo/collections/asr/models/configs/quartznet_config.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # Model general component configs
+    preprocessor: AudioToMelSpectrogramPreprocessorConfig = AudioToMelSpectrogramPreprocessorConfig()
+    spec_augment: Optional[SpectrogramAugmentationConfig] = SpectrogramAugmentationConfig()
+    encoder: ConvASREncoderConfig = ConvASREncoderConfig(activation="relu")
+    decoder: ConvASRDecoderConfig = ConvASRDecoderConfig()
+
+
+@dataclass
+class JasperModelConfig(QuartzNetModelConfig):
```
okuchaiev, line 196: small nitpick. Jasper appeared first, so it would be better to have QuartzNetModelConfig inherit JasperModelConfig and set separable=True

nemo/collections/asr/models/configs/quartznet_config.py
```diff
@@ -0,0 +1,308 @@
...
+from nemo.collections.asr.modules.audio_preprocessing import (
+    AudioToMelSpectrogramPreprocessorConfig,
+    SpectrogramAugmentationConfig,
+)
+from nemo.collections.asr.modules.conv_asr import ConvASRDecoderConfig, ConvASREncoderConfig, JasperEncoderConfig
+from nemo.core.config import modelPT as model_cfg
+
+
+# fmt: off
+def qn_15x5():
```
okuchaiev, line 30: should we start the name with " _" here as we don't expect users to call this themselves?

nemo/core/config/modelPT.py
```diff
@@ -64,3 +63,53 @@ class NemoConfig:
     )
     exp_manager: Optional[Any] = exp_manager.ExpManagerConfig()
     hydra: HydraConfig = HydraConfig()
+
+
+class ModelConfigBuilder:
+    def __init__(self, model_cfg: ModelConfig):
```
okuchaiev, line 69: can you please add docstring to this class?

okuchaiev (changes requested):
few small comments. and thanks for adding unittests too!

examples/asr/experimental/structured/speech_to_text_structured_v2.py
titu1994: Hmm good point. The default is 1 inside the config class since that is generic, but the builder should auto set this to 5 if its a valid config name.

nemo/core/config/modelPT.py
titu1994: Ah forgot about that one.

examples/asr/experimental/structured/speech_to_text_structured_v2.py
titu1994: Done

nemo/collections/asr/models/configs/quartznet_config.py
titu1994: Hmm technically users can override this function which returns a constructor using `encoder_cfg_func` in the constructor. So they may or may not use this function explicitly. They can also import this function to get just the encoder config. 

So I think its better to keep them "public" methods, but to access them youd need to import the actual module since its not provided inside `__init__`

nemo/core/config/modelPT.py
titu1994: Done

nemo/collections/asr/models/configs/quartznet_config.py
titu1994: Done

okuchaiev (approved):
lgtm

PR #1669, opened 2021-01-27
Patch for NMT accuracy regression from PTL 1.1.3 to PTL 1.1.4+

ericharper:
The default for `find_unused_parameters` was changed from `True` to `False` in PTL. We're seeing higher accuracy in NMT with `find_unused_parameters=True`. Overriding this in NMT model until we learn more.


nemo/core/classes/modelPT.py
```diff
@@ -1206,6 +1208,16 @@ def _update_dataset_config(self, dataset_name: str, config: Optional[Union[DictC
             else:
                 raise ValueError("`dataset_name` when updating config must be one of [train, validation, test]")
 
+    # def configure_ddp(self, model: LightningModule, device_ids: List[int]) -> DistributedDataParallel:
```
titu1994, line 1211: Why is this not a core level fix ?

ericharper: It was failing unit tests.

titu1994: Hmm wasn't it the default in 1.1.3 anyway ? It's odd that unit tests should fail

requirements/requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.18.2
 onnx>=1.7.0
-pytorch-lightning>=1.1.0,<=1.1.3
+pytorch-lightning>=1.1.0,<=1.1.5
```
okuchaiev, line 3: do we need the <=1.1.5 part?

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
```diff
@@ -188,7 +190,7 @@ def __init__(self, embedding, decoder, log_softmax, beam_size=1, temperature=1.0
         self.beam_size = beam_size
         self.temp = temperature
 
-    @torch.no_grad()
+    # @torch.no_grad()
```
okuchaiev, line 193: just wondering - why these no_grad() are no longer needed?

requirements/requirements.txt
ericharper: 1.1.6 was breaking MelGan.

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
ericharper: I was thinking that weights should be frozen anyway during eval, so that these calls would be redundant.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_config.py
```diff
@@ -34,6 +34,7 @@ class MTEncDecModelConfig(EncDecNLPModelConfig):
     len_pen: float = 0.0
     max_generation_delta: int = 3
     label_smoothing: Optional[float] = 0.0
+    find_unused_parameters: Optional[bool] = False
```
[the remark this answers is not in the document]
ericharper: Yes, the goal was to get the same accuracy with False.

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
ericharper: PTL calls .eval() and .no_grad() for val and test: https://pytorch-lightning.readthedocs.io/en/stable/converting.html#find-the-val-loop-meat

titu1994: What about using these classes outside of PTL context?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_config.py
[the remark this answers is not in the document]
ericharper: Changed to True and accuracy with False is still low.

okuchaiev (approved):
lgtm, thanks!

PR #1670, opened 2021-01-27
Add ASR with Subword Tokenization Tutorial

titu1994:
# Changelog
- Adds ASR with subword tokenization notebook
- Move configs from experimental to conf/citrinet/
- Updates change_vocabulary() in CTCBPE and RNNTBPE so that vocabulary updated is actual tokens instead of token indices.
- Update readme with link to colab


titu1994:
@okuchaiev Could you review the move of citrinet configs from experimental to conf/citrinet/ and update to Jenkinsfile?

VahidooX (approved):

PR #1673, opened 2021-01-28
Tarred Datasets for NMT

MaximumEntropy:
This PR adds support for Tarred Datasets when training NMT models on very large datasets. It avoids loading the entire dataset into CPU memory for every DDP process. It uses webdataset (https://github.com/tmbdev/webdataset) similar to ASR and language modeling.

To use this functionality

1. Run `python examples/nlp/machine_translation/create_tarred_dataset.py --clean --bpe_dropout 0.1 --shared_tokenizer --src_fname src_file.txt --tgt_fname tgt_file.txt --out_dir /path/to/out_dir --vocab_size 32000 --tokens_in_batch 16000 --lines_per_dataset_fragment 5000000 --num_batches_per_tarfile 1000` to create a folder that contains many tar files. Each tar file consists of a certain number of batches (specified by `num_batches_per_tarfile`), where each batch is a pickle file. The output folder will also contain `metadata.json` that contains the total number of training batches and YTTM tokenizer model files.
2. Point to the tarred dataset folder when training like so

```
python examples/nlp/machine_translation/enc_dec_nmt.py -c aayn_base \
     trainer.gpus=8 \
     +trainer.max_steps=300000
     ........... 
     model.train_ds.src_file_name=/path/to/out_dir/batches.tokens.16000._OP_1..100_CL_.tar \
     model.train_ds.tgt_file_name=/path/to/out_dir/batches.tokens.16000._OP_1..100_CL_.tar \
     +model.train_ds.load_from_tarred_dataset=true \
     +model.train_ds.metadata_path=/path/to/out_dir/metadata.json \
     +model.train_ds.tar_shuffle_n=100 \
     .............
```


examples/nlp/machine_translation/create_tarred_dataset.py
```diff
@@ -0,0 +1,185 @@
...
+import tarfile
+import numpy as np
+import youtokentome as yttm
+import math
+import json
+
+from nemo.collections.nlp.data import TranslationDataset
+from nemo.collections.nlp.modules.common.tokenizer_utils import get_tokenizer
+
+def create_pickled_dataset(args, src_fname, tgt_fname, num_tokens, encoder_tokenizer, decoder_tokenizer, tar_file_ctr, global_batch_ctr):
```
ericharper, line 30: Let's add doc strings to functions and classes.

ericharper (commented):
Could you add some high level description to the PR with example usage?

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -260,3 +268,155 @@ def clean_src_and_target(
             src_ids_.append(src_ids[i])
             tgt_ids_.append(tgt_ids[i])
         return src_ids_, tgt_ids_
+
+
+class TarredTranslationDataset(IterableDataset):
+    """
+    A similar Dataset to the TranslationDataset, but which loads tarred tokenized pickle files.
+    Accepts a single JSON metadata manifest file as well as the path(s) to the tarball(s) containing the wav files.
```
ericharper, line 276: Is this line left over from copy paste?

examples/nlp/machine_translation/create_tarred_dataset.py
```diff
@@ -0,0 +1,185 @@
...
+    parser.add_argument(
+        '--tokens_in_batch', type=int, default=16000, help='# Tokens per batch per GPU'
+    )
+    parser.add_argument(
+        '--lines_per_dataset_fragment', type=int, default=1000000, help='# Tokens per dataset fragment'
+    )
+    parser.add_argument(
+        '--num_tar_files_per_dataset_fragment', type=int, default=4, help='Number of tarfiles'
+    )
+
```
titu1994, line 101: You cannot have uneven number of files inside each tar file, and the contents of each file should approximately be the same. Uneven data size inside tarfile will cause one or more workers to terminate before the rest, leaving DDP in a locked state. Look at https://github.com/NVIDIA/NeMo/blob/main/scripts/convert_to_tarred_audio_dataset.py#L127-L135 to see how ASR slices off small amounts of data to ensure that each tarfile has roughly the same amount of batches in it.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -43,7 +50,8 @@ class TranslationDataConfig:
     num_workers: int = 8
     load_from_cached_dataset: bool = False
     reverse_lang_direction: bool = False
-
+    load_from_tarred_dataset: bool = False
+    metadata_path: str = ""
```
titu1994, line 54: This can be Optional[str] = None no?

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -260,3 +268,155 @@ def clean_src_and_target(
...
+        if shard_strategy == 'scatter':
+            logging.info("All tarred dataset shards will be scattered evenly across all nodes.")
+            if len(text_tar_filepaths) % world_size != 0:
+                logging.warning(
+                    f"Number of shards in tarred dataset ({len(text_tar_filepaths)}) is not divisible "
+                    f"by number of distributed workers ({world_size})."
+                )
+            begin_idx = (len(text_tar_filepaths) // world_size) * global_rank
+            print('Begin Index : %d' % (begin_idx))
+            print('World Size : %d' % (world_size))
```
titu1994, line 377: Remove prints and use logging.info()

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -260,3 +268,155 @@ def clean_src_and_target(
...
+            end_idx = begin_idx + (len(text_tar_filepaths) // world_size)
+            print('End Index : %d' % (end_idx))
```
titu1994, line 379: Same as above

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -280,6 +286,29 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
                 raise ValueError("src must be equal to target for cached dataset")
             dataset = pickle.load(open(cfg.src_file_name, 'rb'))
             dataset.reverse_lang_direction = cfg.get("reverse_lang_direction", False)
+        elif cfg.get("load_from_tarred_dataset", False):
+            logging.info('Loading from tarred dataset %s' % (cfg.src_file_name))
+            if cfg.src_file_name != cfg.tgt_file_name:
+                raise ValueError("src must be equal to target for tarred dataset")
+            if not cfg.get("metadata_path", ""):
```
titu1994, line 293: use `if not cfg.get(..., None) is None:` instead

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -280,6 +286,29 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
+                decoder_tokenizer=self.decoder_tokenizer,
+                shuffle_n=cfg.get("shuffle_n", 100),
+                shard_strategy=cfg.get("shard_strategy", "scatter"),
+                global_rank=self.global_rank,
+                world_size=self.world_size
+            )
+            return torch.utils.data.DataLoader(
+                dataset=dataset,
+                batch_size=1,
+                num_workers=cfg.get("num_workers", 2),
```
titu1994, line 308: You need to set the collate function for typechecking to work (when you add types that is)

titu1994 (changes requested):
Theres an issue with how the tarfiles are being constructed - please look at how ASR does it so as to evenly distribute the number of samples across each tarfile. If uneven number of files exist in the tarfiles, DDP will hang.

examples/nlp/machine_translation/create_tarred_dataset.py
MaximumEntropy: Fixed! Thanks.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
MaximumEntropy: Yeah, my bad. Fixed now

examples/nlp/machine_translation/create_tarred_dataset.py
MaximumEntropy: Done

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: Not sure what I should be adding types to here?

titu1994: Let's skip it for now, it can be added later if needed.

titu1994 (approved):
LGTM

titu1994:
@MaximumEntropy Try the final dataloader on a multigpu setup to full convergence before merge. Otherwise, it looks good.

okuchaiev:
this should go to "main" (I should probably delete nmtmodel branch). And now there are a bunch of merge conflicts. Any thoughts why?

MaximumEntropy:
@okuchaiev I guess its because this branch was created from nmtmodel rather than main and there have been some commits to main after the nmtmodel merge (like typing). I can either resolve these conflicts or just submit another PR starting from main.

MaximumEntropy:
Closing this PR since it started from the wrong branch.

PR #1675, opened 2021-01-28
Adding Jarvis compatibility to WaveGlow

borisfom:

### DIFF
--- a/nemo/collections/tts/modules/waveglow.py
+++ b/nemo/collections/tts/modules/waveglow.py
@@ -63,7 +63,7 @@ class WaveGlowModule(NeuralModule, Exportable):
         self.wavenet = torch.nn.ModuleList()
         self.convinv = torch.nn.ModuleList()
         self.mode = OperationMode.infer
-        
+
         n_half = n_group // 2
 
         # Set up layers with the right sizes based on how many dimensions
@@ -85,7 +85,7 @@ class WaveGlowModule(NeuralModule, Exportable):
             )
         self.n_remaining_channels = n_remaining_channels
         self.removed_weightnorm = False
-        
+
     def _prepare_for_export(self):
         """
         Override this method to prepare module for export. This is in-place operation.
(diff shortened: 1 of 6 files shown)


nemo/collections/tts/modules/waveglow.py
```diff
@@ -85,51 +85,62 @@ def __init__(
...
-    @typecheck()
+    @typecheck(
+        input_types={
+            "spec": NeuralType(('B', 'D', 'T'), MelSpectrogramType()),
+            "z": NeuralType(('B', 'D', 'T'), MelSpectrogramType(), optional=True),
+            "sigma": NeuralType(optional=True),
+        },
+        output_types={"audio": NeuralType(('B', 'T'), AudioSignal())},
+    )
+    def infer(self, spec, z=None, sigma=1):
```
blisc, line 96: @borisfom, I changed symbolic to infer, and added sigma to the function signature. Otherwise, this PR looks good to me. Let's wait for CI and merge?

blisc (approved):

nemo/core/classes/exportable.py
```diff
@@ -123,11 +123,16 @@ def export(
             if len(dynamic_axes) == 0:
                 dynamic_axes = None
 
+            if hasattr(self, 'symbolic'):
+                subj = self.symbolic
+            else:
+                subj = self.forward
```
blisc, line 129: I guess this part will have to change? Should I keep it as symbolic instead of infer?

nemo/collections/tts/modules/waveglow.py
```diff
@@ -101,7 +110,7 @@ def forward(self, spec, audio=None, run_inverse=True, sigma=1.0):
         if run_inverse:
             # norm_dist_to_audio is used to predict audio from spectrogram so only used in val or infer mode
             # Could also log train audio but currently not done
-            audio_pred = self.norm_dist_to_audio(spec=spec, sigma=sigma)
+            audio_pred = self.norm_dist_to_audio(spec=spec, sigma=sigma, z=z)
```
blisc, line 113: Why add z here? This breaks waveglow evaluation since this is no longer the z that you pass to `forward`. It is the `z` that gets computed during the forward pass.

borisfom: I have added optional z parameter to forward() as Jarvis needs it. It has to be None in this case - but clearly the error happens in 'not None' conditional section ... This may be PT bug : I am trying to rename outer scope 'z' variable first - if that does not fix things, I will change to separate calls for inference and validation mode.

borisfom: Oh, figured - no PT bug mystery here - there was 'z' local variable assigned right there in forward - but only for validation. Renamed local variable, should be all good now.

PR #1678, opened 2021-01-29
add new quantization nodes in QuartzNet

ttyio:


okuchaiev:
Could you please sign your commits. See https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md for general guidelines.
To sign already committed commits, checkout DCO's bot output: https://github.com/NVIDIA/NeMo/pull/1678/checks?check_run_id=1789269686

okuchaiev:
CI failed on style check. 
To check for style do:
``python setup.py style``
To fix style:
``python setup.py style --fix``

titu1994:
This PR requires a significant refactor - and does not need to add explicit global support for quantization in the constructor. If you would agree to it, could I push some changes to this branch directly ?

nemo/collections/asr/models/ctc_models.py
```diff
@@ -86,7 +86,7 @@ def list_available_models(cls) -> Optional[PretrainedModelInfo]:
         result.append(model)
         return result
 
-    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None, quantize: bool = False):
```
titu1994, line 89: No need to add `quantize` to constructor.

nemo/collections/asr/modules/audio_preprocessing.py
```diff
@@ -208,6 +208,7 @@ def __init__(
         stft_conv=False,
         pad_value=0,
         mag_power=2.0,
+        quantize=False,
```
titu1994, line 211: quantize argument added but it doesnt do anything to `AudioToMelSpectrogramPreprocessor` ? And is `AudioToMFCCPreprocessor` not supported?

nemo/collections/asr/parts/jasper.py
```diff
@@ -21,6 +21,18 @@
 
 from nemo.collections.asr.parts.activations import Swish
 
+try:
+    from pytorch_quantization import calib
+    from pytorch_quantization import nn as quant_nn
+    from pytorch_quantization import quant_modules
+    from pytorch_quantization.tensor_quant import QuantDescriptor
+except ImportError:
```
titu1994, line 29: Needs to be an optional import under a try catch guard
```python
try:
    from pytorch_quantization import calib
    from pytorch_quantization import nn as quant_nn
    from pytorch_quantization import quant_modules
    from pytorch_quantization.tensor_quant import QuantDescriptor
    PYTORCH_QUANTIZATION_AVAILABLE = True
except (ImportError, ModuleNotFoundError):
    PYTORCH_QUANTIZATION_AVAILABLE  = False
```

After this, every `if quantize` flag must be replaced with `if PYTORCH_QUANTIZATION_AVAILABLE  and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -139,16 +152,28 @@ def __init__(
...
-            in_channels,
-            out_channels,
-            kernel_size,
-            stride=stride,
-            padding=padding,
-            dilation=dilation,
-            groups=groups,
-            bias=bias,
-        )
+        if quantize:
```
titu1994, line 155: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -227,18 +253,31 @@ def __init__(
         self.interpolation_mode = interpolation_mode
 
         if self.context_window <= 0:
-            self.pool = nn.AdaptiveAvgPool1d(1)  # context window = T
+            if quantize:
```
titu1994, line 256: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -227,18 +253,31 @@ def __init__(
...
+                self.pool = quant_nn.QuantAdaptiveAvgPool1d(1)  # context window = T
+            else:
+                self.pool = nn.AdaptiveAvgPool1d(1)  # context window = T
         else:
-            self.pool = nn.AvgPool1d(self.context_window, stride=1)
+            if quantize:
```
titu1994, line 261: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -227,18 +253,31 @@ def __init__(
...
 
         if activation is None:
             activation = nn.ReLU(inplace=True)
 
-        self.fc = nn.Sequential(
-            nn.Linear(channels, channels // reduction_ratio, bias=False),
-            activation,
-            nn.Linear(channels // reduction_ratio, channels, bias=False),
-        )
+        if quantize:
```
titu1994, line 269: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -385,12 +429,15 @@ def __init__(
...
                         norm_groups=norm_groups,
                         stride=stride_val,
+                        quantize=quantize,
                     )
                 )
 
                 res_list.append(res)
 
             self.res = res_list
+            if self.quantize:
```
titu1994, line 439: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -422,18 +470,31 @@ def _get_conv(
...
-                in_channels,
-                out_channels,
-                kernel_size,
-                stride=stride,
-                dilation=dilation,
-                padding=padding,
-                bias=bias,
-                groups=groups,
-            )
+            if quantize:
```
titu1994, line 476: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/collections/asr/parts/jasper.py
```diff
@@ -545,7 +610,10 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]):
                         res_out = res_layer(res_out)
 
                 if self.residual_mode == 'add' or self.residual_mode == 'stride_add':
-                    out = out + res_out
+                    if self.quantize:
```
titu1994, line 613: Replace with the `if PYTORCH_QUANTIZATION_AVAILABLE and quantize:`

nemo/core/classes/common.py
```diff
@@ -245,7 +245,7 @@ def __attach_neural_type(self, obj, type_val, name=None):
 
 class Serialization(ABC):
     @classmethod
-    def from_config_dict(cls, config: DictConfig):
+    def from_config_dict(cls, config: DictConfig, quantize: bool=False):
```
titu1994, line 248: No need for adding `quantize: bool = False`

nemo/core/classes/common.py
```diff
@@ -257,13 +257,21 @@ def from_config_dict(cls, config: DictConfig):
 
         if ('cls' in config or 'target' in config) and 'params' in config:
             # regular hydra-based instantiation
+            if config.cls == 'nemo.collections.asr.modules.ConvASREncoder':
+                OmegaConf.set_struct(config, False)
+                config.quantize = quantize
+                OmegaConf.set_struct(config, True)
```
titu1994, line 263: Not needed to specialize code like this for a certain module, and config.cls is not guarenteed to exist since Hydra 1.x has `_target_` now. Better not override any such core method to inject parameters and handle it inside `ConvASREncoder` itself.

nemo/core/classes/common.py
```diff
@@ -257,13 +257,21 @@ def from_config_dict(cls, config: DictConfig):
...
             instance = hydra.utils.instantiate(config=config)
         elif '_target_' in config:
             # regular hydra-based instantiation
+            if config._target_ == 'nemo.collections.asr.modules.ConvASREncoder':
+                OmegaConf.set_struct(config, False)
+                config.quantize = quantize
+                OmegaConf.set_struct(config, True)
```
titu1994, line 270: Same comment as above

nemo/core/classes/common.py
```diff
@@ -257,13 +257,21 @@ def from_config_dict(cls, config: DictConfig):
...
             instance = hydra.utils.instantiate(config=config)
         else:
             # models are handled differently for now
-            instance = cls(cfg=config)
+            instance = cls(cfg=config, quantize=quantize)
```
titu1994, line 274: Please revert as this will break all other nemo models since they dont support this quantize flag. There are other ways to do this.

nemo/core/classes/modelPT.py
```diff
@@ -444,6 +445,7 @@ def restore_from(
         override_config_path: Optional[str] = None,
         map_location: Optional[torch.device] = None,
         strict: bool = False,
+        quantize: bool = False,
```
titu1994, line 448: We can avoid such a deep integration with the core by refactoring out some component of restore_from - mainly its config loading and config overriding. We can add a method to extract and instantiate just the model's config (and not the model itself) and return that - this will give you a model's config prior to instantiation to inject the `model.*.quantize = True` flag and then restore from using this overriden config. 

The primary reason for doing this is 
 - It is generic and allows overriding parts of the model before restoring weights (as long as changes dont affect weights themselves)
 - quantize is a property of just 1 model right now and not general to all Nemo models, so it doesnt seem right to expose the `quantize=False` flag in core methods across all models.
 - The way quantize is being propagated right now, it will prevent instantiation of any other model since no other NeMo model constructor has `quantize` argument in it

titu1994 (changes requested):
I've added a few comments as to why this PR would need a heavy refactor. Please advice if its alright for me to push commits directly to it

ttyio:
> This PR requires a significant refactor - and does not need to add explicit global support for quantization in the constructor. If you would agree to it, could I push some changes to this branch directly ?

Thanks @titu1994 , I do not know other ways to pass the quantization flag from the upper script to model, please help me to refactor the code, thanks!

titu1994:
Sounds good, thanks for the go ahead.

titu1994:
@okuchaiev for review regarding core ModelPT changes to support 1) returning of config instead of model 2) overriding config with either str path or with OmegaConf object

titu1994 (approved):
Refactored code. Everything works well. @okuchaiev Final review on design of return_config parameter as well as override_config_path. Note that EFF save/restore will not support this new parameter at the moment until it is updated in the future.

okuchaiev (approved):
looks good to me. Thanks!

PR #1684, opened 2021-01-30
fix eval_model in IntentSlot tutorial

carolmanderson:
Ensures the reloaded (best) model is used for evaluation as intended.


okuchaiev (approved):
looks good. thank you.

PR #1687, opened 2021-02-01
Update NeMo model filepath

blisc:


titu1994:
As long as the pretrained tests don't fail, this change should be ok. I believe we have some save - restore - save - restore tests too which should test if the checkpoints are correctly being saves.

titu1994 (approved):

PR #1707, opened 2021-02-03
Speech Regression Support

diego-fustes:
Support for Encoder-Decoder models that learn to predict numeric values in a regression model. An example application is non-intrusive speech-quality assessment as presented here:

https://www.microsoft.com/en-us/research/uploads/prod/2019/04/preprint_ICASSP2019_MOS_estimation.pdf


diego-fustes:
Hi @titu1994. Well, this was the hard way I guess, but I think that the result is what I wanted. Please let me know how it's coming out

titu1994 (changes requested):
PR is overall good, but needs a significant refactor for clarity and extensibility for future cases. It 
 needs a closer look its data input pipeline and overall structure. Adding @fayejf for final review once initial comments are resolved.

titu1994:
And error case is caught in the EncDecRegressionModel - 
```
=================================== FAILURES ===================================
_________________ TestEncDecRegressionModel.test_transcription _________________

self = <test_asr_regression_model.TestEncDecRegressionModel object at 0x7f3de006aac8>
speech_regression_model = EncDecRegressionModel(
  (preprocessor): AudioToMelSpectrogramPreprocessor(
    (featurizer): FilterbankFeatures()
  )...coder_layers): Sequential(
      (0): Linear(in_features=32, out_features=1, bias=True)
    )
  )
  (loss): MSELoss()
)
test_data_dir = '/home/mrjenkins/.jenkins/workspace/NeMo_PR-1707/tests/.data'

    @pytest.mark.unit
    def test_transcription(self, speech_regression_model, test_data_dir):
        # Ground truth labels = ["yes", "no"]
        audio_filenames = ['an22-flrp-b.wav', 'an90-fbbh-b.wav']
        audio_paths = [os.path.join(test_data_dir, "asr", "train", "an4", "wav", fp) for fp in audio_filenames]
        model = speech_regression_model.eval()

        # Test Top 1 classification transcription
        results = model.transcribe(audio_paths, batch_size=2)

>       assert len(results) == 2
E       assert 0 == 2
E         +0
E         -2
```

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -282,17 +282,31 @@ class _AudioLabelDataset(Dataset):
...
-                AudioSignal(freq=self._sample_rate)
-                if self is not None and hasattr(self, '_sample_rate')
-                else AudioSignal(),
-            ),
-            'a_sig_length': NeuralType(tuple('B'), LengthsType()),
-            'label': NeuralType(tuple('B'), LabelsType()),
-            'label_length': NeuralType(tuple('B'), LengthsType()),
-        }
+
+        if self.regression:
```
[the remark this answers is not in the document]
diego-fustes: ok

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -304,26 +318,36 @@ def __init__(
...
         )
 
         self.featurizer = featurizer
         self.trim = trim
         self.load_audio = load_audio
+        self.regression = regression
 
-        self.labels = labels if labels else self.collection.uniq_labels
-        self.num_classes = len(self.labels)
+        if self.regression:
```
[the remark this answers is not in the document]
diego-fustes: sure

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -343,10 +367,15 @@ def __getitem__(self, index):
...
+            t = sample.label
+            tl = 1  # For compatibility with collate_fn used later
+            return f, fl, torch.tensor(t).float(), torch.tensor(tl).long()
+        else:
+            t = self.label2id[sample.label]
+            tl = 1  # For compatibility with collate_fn used later
 
-        return f, fl, torch.tensor(t).long(), torch.tensor(tl).long()
+            return f, fl, torch.tensor(t).long(), torch.tensor(tl).long()
```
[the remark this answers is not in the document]
diego-fustes: that's true

nemo/collections/asr/data/audio_to_label.py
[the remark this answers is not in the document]
diego-fustes: ok

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -343,10 +367,15 @@ def __getitem__(self, index):
         else:
             f, fl = None, None
 
-        t = self.label2id[sample.label]
-        tl = 1  # For compatibility with collate_fn used later
+        if self.regression:
+            t = sample.label
```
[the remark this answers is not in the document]
diego-fustes: Here I am referring to the 'target" value in the manifest for training/testing, not to the labels in the config

tests/collections/asr/test_asr_regression_model.py
```diff
@@ -0,0 +1,89 @@
...
+    def test_transcription(self, speech_regression_model, test_data_dir):
+        # Ground truth labels = ["yes", "no"]
+        audio_filenames = ['an22-flrp-b.wav', 'an90-fbbh-b.wav']
+        audio_paths = [os.path.join(test_data_dir, "asr", "train", "an4", "wav", fp) for fp in audio_filenames]
+
+        model = speech_regression_model.eval()
+
+        # Test Top 1 classification transcription
+        results = model.transcribe(audio_paths, batch_size=2)
+        assert len(results) == 2
```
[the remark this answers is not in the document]
diego-fustes: transcribe returns a list of floats

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -304,26 +318,36 @@ def __init__(
...
         self.featurizer = featurizer
         self.trim = trim
         self.load_audio = load_audio
+        self.regression = regression
 
-        self.labels = labels if labels else self.collection.uniq_labels
-        self.num_classes = len(self.labels)
+        if self.regression:
+            self.labels = []
+            self.num_classes = 1
```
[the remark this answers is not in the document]
diego-fustes: ok

nemo/collections/asr/data/audio_to_label_dataset.py
```diff
@@ -15,24 +15,28 @@
 from nemo.collections.asr.data import audio_to_label
 
 
-def get_classification_label_dataset(featurizer, config: dict,) -> audio_to_label.AudioToClassificationLabelDataset:
+def get_classification_label_dataset(
+    featurizer, config: dict, regression=False
```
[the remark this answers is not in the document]
diego-fustes: good point

nemo/collections/asr/data/audio_to_label_dataset.py
```diff
@@ -15,24 +15,28 @@
...
         config: Config of the AudioToClassificationLabelDataset.
+        regression: True if the target is a numeric value for regression
 
     Returns:
         An instance of AudioToClassificationLabelDataset.
     """
     dataset = audio_to_label.AudioToClassificationLabelDataset(
         manifest_filepath=config['manifest_filepath'],
-        labels=config['labels'],
+        labels=config.get('labels', []),
```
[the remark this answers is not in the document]
diego-fustes: ok

nemo/collections/asr/models/classification_models.py
```diff
@@ -28,21 +29,21 @@
...
+__all__ = ['EncDecClassificationModel', 'EncDecRegressionModel', 'MatchboxNet']
 
 
-class EncDecClassificationModel(ASRModel, Exportable):
-    """Encoder decoder Classification models."""
+class _EncDecBaseModel(ASRModel, Exportable):
+    """Abstract class to represent an ASR encoder-decoder model."""
 
-    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None, regression=False):
```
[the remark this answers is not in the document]
diego-fustes: got it

nemo/collections/asr/models/classification_models.py
```diff
@@ -52,24 +53,176 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
-            self.crop_or_pad = EncDecClassificationModel.from_config_dict(self._cfg.crop_or_pad_augment)
+            self.crop_or_pad = ASRModel.from_config_dict(self._cfg.crop_or_pad_augment)
         else:
             self.crop_or_pad = None
 
-        # Setup metric objects
-        self._accuracy = TopKClassificationAccuracy(dist_sync_on_step=True)
+        self.preprocessor = self._load_preprocessor()
+        self.encoder = self._load_encoder()
+        self.decoder = self._load_decoder()
```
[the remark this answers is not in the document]
diego-fustes: makes sense

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +619,125 @@ def _update_decoder_config(self, cfg):
...
-        Args:
-            config: A python dictionary which contains the following keys:
-            paths2audio_files: (a list) of paths to audio files. The files should be relatively short fragments. \
-                Recommended length per file is between 5 and 25 seconds.
-            batch_size: (int) batch size to use during inference. \
-                Bigger will result in better throughput performance but would use more memory.
-            temp_dir: (str) A temporary directory where the audio manifest is temporarily
-                stored.
+class EncDecRegressionModel(_EncDecBaseModel):
+    """Encoder decoder class for speaker regression models.
```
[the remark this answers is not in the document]
diego-fustes: speech regression is better as being more general

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +619,125 @@ def _update_decoder_config(self, cfg):
...
+    Model class creates training, validation methods for setting up data
+    performing model forward pass.
+    Expects config dict for
+    * preprocessor
+    * Jasper/Quartznet Encoder
+    * Speaker Decoder
+    """
 
+    def _update_decoder_config(self, cfg):
```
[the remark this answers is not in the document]
diego-fustes: It's only related to classification, so I've moved it ton EncDecClassificationModel. I think that the idea is to rebuild the decoder in the case that the labels in the config are different from the ones in the pretrained model / checkpoint. That can be also achieved by calling the method "change_labels"... should be do it in the constructor?

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +619,125 @@ def _update_decoder_config(self, cfg):
...
-            'shuffle': False,
+        result = []
+
+        return result
+
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
+        super().__init__(cfg=cfg, trainer=trainer, regression=True)
+        self.preprocessor = EncDecRegressionModel.from_config_dict(cfg.preprocessor)
+        self.encoder = EncDecRegressionModel.from_config_dict(cfg.encoder)
+        self.decoder = EncDecRegressionModel.from_config_dict(cfg.decoder)
```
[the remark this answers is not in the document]
diego-fustes: that's bad, fixing

diego-fustes:
Hi again @titu1994. I hope that I've addressed all your concerns above

nemo/collections/asr/models/classification_models.py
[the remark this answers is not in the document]
titu1994: Hmm the constructor shouldnt go updating decoder like that - its already instantiated once when from_config_dict() is called. Lets let the user call it explicitly with change_labels().

tests/collections/asr/test_asr_regression_model.py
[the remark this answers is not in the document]
titu1994: A ok then thats fine. Floats are hard to assert exactly.

nemo/collections/asr/data/audio_to_label_dataset.py
```diff
@@ -15,12 +15,13 @@
...
 
 
-def get_classification_label_dataset(featurizer, config: dict,) -> audio_to_label.AudioToClassificationLabelDataset:
+def get_classification_label_dataset(featurizer, config: dict) -> audio_to_label.AudioToClassificationLabelDataset:
     """
     Instantiates a Classification AudioLabelDataset.
 
     Args:
         config: Config of the AudioToClassificationLabelDataset.
+        regression: True if the target is a numeric value for regression
```
titu1994, line 24: This can be removed now

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -304,26 +318,37 @@ def __init__(
...
+            self.label2id, self.id2label = {}, {}
+            for label_id, label in enumerate(self.labels):
+                self.label2id[label] = label_id
+                self.id2label[label_id] = label
+
+            for idx in range(len(self.labels[:5])):
+                logging.debug(" label id {} and its mapped label {}".format(idx, self.id2label[idx]))
+
+        else:
+            self.num_classes = len(self.labels) if self.labels is not None else 1
```
titu1994, line 351: self.labels is not set before this else clause, its set inside the if clause. So this will raise an error since self doesnt contain the attribute `labels`.

nemo/collections/asr/models/classification_models.py
```diff
@@ -53,23 +54,199 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+    def input_types(self) -> Optional[Dict[str, NeuralType]]:
+        if hasattr(self.preprocessor, '_sample_rate'):
+            audio_eltype = AudioSignal(freq=self.preprocessor._sample_rate)
+        else:
+            audio_eltype = AudioSignal()
+        return {
+            "input_signal": NeuralType(('B', 'T'), audio_eltype),
+            "input_signal_length": NeuralType(tuple('B'), LengthsType()),
+        }
+
```
titu1994, line 127: Can you make the `output_types` attribute here abstract, so  that subclasses are forced to implement it (even a return None). Theres a high chance newer subclasses of this will forget to add the `output_types` since `input_types` is already defined.

nemo/collections/asr/models/classification_models.py
```diff
@@ -147,112 +325,115 @@ def transcribe(self, paths2audio_files: List[str], batch_size: int = 4, logprobs
...
-            val_data_config['shuffle'] = False
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
 
-        # preserve config
-        self._update_dataset_config(dataset_name='validation', config=val_data_config)
+        super().__init__(cfg=cfg, trainer=trainer)
 
-        self._validation_dl = self._setup_dataloader_from_config(config=val_data_config)
+        # Change labels if needed
+        self._update_decoder_config(self._cfg.decoder)
```
titu1994, line 421: This should be removed - since the decoder is already created inside the `_setup_decoder` method. Users must explicitly call `change_labels` if they want to change the decoder outside the constructors scope.

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +647,109 @@ def _update_decoder_config(self, cfg):
...
-            'labels': self.cfg.labels,
-            'batch_size': min(config['batch_size'], len(config['paths2audio_files'])),
-            'trim_silence': False,
-            'shuffle': False,
+        result = []
+
+        return result
+
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
+        OmegaConf.set_struct(cfg, False)
```
titu1994, line 669: I don't think this should be done actually - the user should explicitly set this argument. What can (and should) be done is to check if the `cfg.is_regression_task` value is set to True in the config prior to building the regression model. Else raise a ValueError() stating that the config does not contain the required value of `is_regression_task=True`.

Similarly, the classification model should inspect the config for the absence of `cfg.is_regression_task` or that if it exists in the config, it is False, otherwise raise a ValueError() stating that the config contains an invalid value of `is_regression_task` and it should be False for classification models.

Also, `cfg` is not guarenteed to be `OmegaConf` yet before calling `super().__init__`. Please import model_utils and perform the following

```python
from nemo.utils import  model_utils
...

def __init__(self, cfg: DictConfig, trainer: Trainer = None):
        # Convert config to a DictConfig
        cfg = model_utils.convert_model_config_to_dict_config(cfg)

        # Convert config to support Hydra 1.0+ instantiation
        cfg = model_utils.maybe_update_config_version(cfg)
        
        ....
```

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +647,109 @@ def _update_decoder_config(self, cfg):
...
+        return EncDecRegressionModel.from_config_dict(self._cfg.encoder)
+
+    def _setup_decoder(self):
+        return EncDecRegressionModel.from_config_dict(self._cfg.decoder)
+
+    def _setup_loss(self):
+        return MSELoss()
+
+    def _setup_metrics(self):
+        pass
```
titu1994, line 687: I believe PTL supports MSE as a metric yes? Can we add that here?

nemo/collections/asr/models/classification_models.py
```diff
@@ -489,95 +647,109 @@ def _update_decoder_config(self, cfg):
...
-        onnx.save(output_model, output)
-        return ([output, output1, output2], [output_descr, output1_descr, output2_descr])
+        return {'test_loss': logs['val_loss']}
+
+    def multi_test_epoch_end(self, outputs, dataloader_idx: int = 0):
+        test_loss_mean = torch.stack([x['test_loss'] for x in outputs]).mean()
+
+        tensorboard_logs = {'test_loss': test_loss_mean}
+
+        return {'log': tensorboard_logs}
```
titu1994, line 734: Pass `{"test_loss": test_loss_mean, "log": tensorboard_logs}` here

nemo/collections/asr/models/configs/matchboxnet_config.py
```diff
@@ -217,6 +222,9 @@ def _finalize_cfg(self):
         # propagate num classes
         self.model_cfg.decoder.num_classes = len(self.model_cfg.labels)
 
+        # propagate task type (regression or classification)
+        self.model_cfg.is_regression_task = self.model_cfg.is_regression_task
```
titu1994, line 226: For now, lets revert this, I will update this once the PR merges, theres a little issue with the current approach that I need to think about.

nemo/collections/asr/parts/collections.py
```diff
@@ -289,7 +290,8 @@ def __init__(self, manifests_files: Union[str, List[str]], *args, **kwargs):
         for item in manifest.item_iter(manifests_files, parse_func=self.__parse_item):
             audio_files.append(item['audio_file'])
             durations.append(item['duration'])
-            labels.append(item['label'])
+            if not is_regression_task:
+                labels.append(item['label'])
```
titu1994, line 294: This is problematic. Your earlier code calls `sample.label` in the `_AudioLabelDataset` class - even for regression ... but `sample` is built from this list of values inside `item` - ie since you dont add `label` here for `is_regression_task=True`, `sample.label` raises an exception.

You _must pass ground truth value in the manifest_ - whether its called `label` or `target` or `value` makes little difference, but it must be added here. I would suggest adding the ground truth regression value here as just the `label` - but for regression case, explicitly cast the value to float.

So basically : 
```python
else:
    labels.append(float(item['label']))
```

tests/tools/test_text_normalization.py
```diff
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
```
titu1994, line 15: This should be reverted

titu1994 (commented):
Really nice fixes! There remain some core details that need to be fixed - specifically how the ground truth regression value is loaded by the `_AudioLabelDataset` class. Once that is fixed, it should be only minor fixes here and there.

titu1994:
Ah and the error for this one is 
```
  File "examples/asr/speech_to_label.py", line 129, in <module>
    main()  # noqa pylint: disable=no-value-for-parameter
  File "/opt/conda/lib/python3.6/site-packages/nemo/core/config/hydra_runner.py", line 103, in wrapper
    strict=None,
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 347, in _run_hydra
    lambda: hydra.run(
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 201, in run_and_report
    raise ex
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 198, in run_and_report
    return func()
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 350, in <lambda>
    overrides=args.overrides,
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/hydra.py", line 112, in run
    configure_logging=with_log_configuration,
  File "/opt/conda/lib/python3.6/site-packages/hydra/core/utils.py", line 127, in run_job
    ret.return_value = task_function(task_cfg)
  File "examples/asr/speech_to_label.py", line 117, in main
    asr_model = EncDecClassificationModel(cfg=cfg.model, trainer=trainer)
  File "/opt/conda/lib/python3.6/site-packages/nemo/collections/asr/models/classification_models.py", line 418, in __init__
  File "/opt/conda/lib/python3.6/site-packages/nemo/collections/asr/models/classification_models.py", line 70, in __init__
    self.decoder = self._setup_decoder()
  File "/opt/conda/lib/python3.6/site-packages/nemo/collections/asr/models/classification_models.py", line 430, in _setup_decoder
    return EncDecClassificationModel.from_config_dict(self._cfg.decoder)
  File "/opt/conda/lib/python3.6/site-packages/nemo/core/classes/common.py", line 263, in from_config_dict
    instance = hydra.utils.instantiate(config=config)
  File "/opt/conda/lib/python3.6/site-packages/hydra/utils.py", line 70, in call
    raise HydraException(f"Error calling '{cls}' : {e}") from e

hydra.errors.HydraException: Error calling 'nemo.collections.asr.modules.ConvASRDecoderClassification' : __init__() missing 1 required positional argument: 'num_classes'

script returned exit code 1
```

nemo/collections/asr/models/classification_models.py
titu1994, line 421: This should be done *before* calling `super.__init__` so that the decoder config is updated properly. Otherwise, let baseclass call it *before* it calls the `_setup_decoder`.

nemo/collections/asr/parts/collections.py
diego-fustes: You are right about this. I've missed that because it was working anyway, and the reason is that the flag is_regression_task is not being populated to  `get_classification_label_dataset`.

I'm fixing this problem by adding the flag to the data loader config in the method `_setup_dataloader_from_config` in the class `_EncDecBaseModel,` what do you think?

nemo/collections/asr/models/classification_models.py
diego-fustes: Sure, indeed I'm adding also MAE as it's a popular metric for regression. MSE will be the same as loss for this model, but it's possible to create regression models with other loss functions.

Just one comment on this. The loss that is computed and the end of validation/testing is the average loss of the losses computed for every batch, which is not the same as the total loss for the validation/test dataset. In the case of metrics, I'm doing the later and the output is the actual MSE/MAE for the complete dataset.

diego-fustes:
ready for review

nemo/collections/asr/models/classification_models.py
```diff
@@ -52,24 +54,218 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+        # Change labels if needed
+        self._update_decoder_config(cfg.labels, cfg.decoder)
         super().__init__(cfg=cfg, trainer=trainer)
-        self._update_decoder_config(self._cfg.decoder)
 
-        self.preprocessor = EncDecClassificationModel.from_config_dict(self._cfg.preprocessor)
-        self.encoder = EncDecClassificationModel.from_config_dict(self._cfg.encoder)
-        self.decoder = EncDecClassificationModel.from_config_dict(self._cfg.decoder)
-        self.loss = CrossEntropyLoss()
+        self.is_regression_task = self._cfg.get('is_regression_task', False)
```
titu1994, line 68: Same as line 63?

nemo/collections/asr/models/classification_models.py
```diff
@@ -147,112 +344,119 @@ def transcribe(self, paths2audio_files: List[str], batch_size: int = 4, logprobs
...
+    """Encoder decoder Classification models."""
 
-    def setup_validation_data(self, val_data_config: Optional[Union[DictConfig, Dict]]):
-        if 'shuffle' not in val_data_config:
-            val_data_config['shuffle'] = False
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
 
-        # preserve config
-        self._update_dataset_config(dataset_name='validation', config=val_data_config)
+        if cfg.is_regression_task:
```
titu1994, line 441: Do `if 'is_regression_task' in cfg and cfg.is_regression_task:` so that it doesnt crash if `is_regression_task` doesnt exist in config.

nemo/collections/asr/models/classification_models.py
```diff
@@ -471,113 +652,149 @@ def change_labels(self, new_labels: List[str]):
...
+        This method returns a list of pre-trained model which can be instantiated directly from NVIDIA's NGC cloud.
+        Returns:
+            List of available pre-trained models.
+        """
+        result = []
+
+        return result
+
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None):
+        if not cfg.is_regression_task:
```
titu1994, line 691: Do `if 'is_regression_task' in cfg and not cfg.is_regression_task:` to first check if `is_regression_task` exists or not.

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -282,18 +282,32 @@ class _AudioLabelDataset(Dataset):
...
+            output_types.update(
+                {
+                    'targets': NeuralType(tuple('B'), RegressionValuesType()),
+                    'targets_length': NeuralType(tuple('B'), LengthsType()),
+                }
+            )
+        else:
+
+            output_types.update(
+                {'label': NeuralType(tuple('B'), LabelsType()), 'label_length': NeuralType(tuple('B'), LengthsType()),}
```
titu1994, line 306: `label` for regression should be `('B', 'D')` to support multivariate regression in the future (even if D = 1 for now).

nemo/collections/asr/models/classification_models.py
```diff
@@ -471,113 +652,149 @@ def change_labels(self, new_labels: List[str]):
...
-            'trim_silence': False,
-            'shuffle': False,
-        }
 
-        temporary_datalayer = self._setup_dataloader_from_config(config=DictConfig(dl_config))
-        return temporary_datalayer
+            A list of predictions in the same order as paths2audio_files
+        """
+        predictions = super().transcribe(paths2audio_files, batch_size, logprobs=True)
+        return [float(pred) for pred in predictions]
```
titu1994, line 786: Simply return predictions - there might be further processing done on it as a torch.tensor

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -295,7 +295,7 @@ def output_types(self):
     def __init__(
         self,
         feat_in: int,
-        num_classes: int,
+        num_classes,
```
titu1994, line 298: Add back the `: int`

tests/collections/asr/test_asr_classification_model.py
```diff
@@ -61,6 +61,7 @@ def speech_classification_model():
             'encoder': DictConfig(encoder),
             'decoder': DictConfig(decoder),
             'labels': ListConfig(["dummy_cls_{}".format(i + 1) for i in range(30)]),
+            'is_regression_task': False,
```
titu1994, line 64: Revert this since in most cases, older configs (and related tests) must pass without modification - so that older models like MatchboxNet and MarbleNet restoration is supported.

titu1994 (commented):
Overall looking really great ! One final review from @fayejf to make sure everything works as intended for Speaker Diarization pipeline.

titu1994:
The error is due to missing keys - following the review comments should resolve it
```

  File "examples/asr/speech_to_label.py", line 129, in <module>
    main()  # noqa pylint: disable=no-value-for-parameter
  File "/opt/conda/lib/python3.6/site-packages/nemo/core/config/hydra_runner.py", line 103, in wrapper
    strict=None,
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 347, in _run_hydra
    lambda: hydra.run(
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 201, in run_and_report
    raise ex
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 198, in run_and_report
    return func()
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/utils.py", line 350, in <lambda>
    overrides=args.overrides,
  File "/opt/conda/lib/python3.6/site-packages/hydra/_internal/hydra.py", line 112, in run
    configure_logging=with_log_configuration,
  File "/opt/conda/lib/python3.6/site-packages/hydra/core/utils.py", line 127, in run_job
    ret.return_value = task_function(task_cfg)
  File "examples/asr/speech_to_label.py", line 117, in main
    asr_model = EncDecClassificationModel(cfg=cfg.model, trainer=trainer)
  File "/opt/conda/lib/python3.6/site-packages/nemo/collections/asr/models/classification_models.py", line 441, in __init__
    if cfg.is_regression_task:
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/dictconfig.py", line 305, in __getattr__
    self._format_and_raise(key=key, value=None, cause=e)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/base.py", line 101, in _format_and_raise
    type_override=type_override,
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/_utils.py", line 629, in format_and_raise
    _raise(ex, cause)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/_utils.py", line 610, in _raise
    raise ex  # set end OC_CAUSE=1 for full backtrace
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/dictconfig.py", line 303, in __getattr__
    return self._get_impl(key=key, default_value=DEFAULT_VALUE_MARKER)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/dictconfig.py", line 361, in _get_impl
    node = self._get_node(key=key)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/dictconfig.py", line 383, in _get_node
    self._validate_get(key)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/dictconfig.py", line 136, in _validate_get
    key=key, value=value, cause=ConfigAttributeError(msg)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/base.py", line 101, in _format_and_raise
    type_override=type_override,
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/_utils.py", line 694, in format_and_raise
    _raise(ex, cause)
  File "/opt/conda/lib/python3.6/site-packages/omegaconf/_utils.py", line 610, in _raise
    raise ex  # set end OC_CAUSE=1 for full backtrace
omegaconf.errors.ConfigAttributeError: Key 'is_regression_task' is not in struct
	full_key: model.is_regression_task
	reference_type=Any
	object_type=dict
```

nemo/collections/asr/models/classification_models.py
diego-fustes: copy/paste issue, fixing

nemo/collections/asr/models/classification_models.py
diego-fustes: Shouldn't it be `if 'is_regression_task' not in cfg or cfg.is_regression_task:`? That way we enforce it to be in config with the right value

nemo/collections/asr/models/classification_models.py
diego-fustes: Same as above

nemo/collections/asr/models/classification_models.py
diego-fustes: the problem in that case is with the type annotations as it's expecting to return `List[Float]`, should it return `List[str]`, well is not the case, but functionality is not affected...

nemo/collections/asr/models/classification_models.py
diego-fustes: Looking at the posterior comments below, I guess that you prefer to assume that the absence of the flag `is_regression_task` in the config is equivalent to set it to false. Is that correct?

titu1994: > if 'is_regression_task' not in cfg or cfg.is_regression_task:
Yes this is correct. My bad. 
The assumption is that classification is the default - and therefore does not need the `is_regression_task` flag to be present in the config. 

If it is present, and if it is set to True, raise an error in the constructor of the Classification model.

In the regression models, if the flag doesnt exist or if it does exist but is set to False, raise an error.

nemo/collections/asr/models/classification_models.py
titu1994: Yes my bad, it should mandatorily check that 1) cfg.is_regression_task exists, and if it does exist, it is set to True. If either condition is not met, raise an error in the regression model.

nemo/collections/asr/models/classification_models.py
titu1994: Hmm agreed, parallel to List[str] for classificaiton, lets keep this as List[float] then.

tests/collections/asr/test_asr_classification_model.py
titu1994: ++

titu1994 (commented):
Sorry for the delay, rectified my if cases in the comments. In general the classification models should not require the presence of the `cfg.is_regression_task` and if present, it should be False.

On the other hand, regression models must require the presence of `cfg.is_regression_task` and  when present, must be set to True.

titu1994:
Oh and there are some code formatting issues, could you run the `python setup.py style --fix --scope=nemo/`

diego-fustes:
ok, done, could you check if all is good now?

fayejf:
@diego-fustes Hi Thanks for this excellent contribution. Could you run `python setup.py style --fix `? There are some fomart issue in `tests` fails the CI tests.

diego-fustes:
I always try to run the style checks... anyway I've done that and also renamed back the class `_EncDecBaseModel`, it looks like it was accidentally renamed in a previous merge, causing errors in the module...

nemo/collections/asr/models/classification_models.py
```diff
@@ -49,24 +53,217 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+
+        self._test_dl = self._setup_dataloader_from_config(config=test_data_config)
+
+    def test_dataloader(self):
+        if self._test_dl is not None:
+            return self._test_dl
+
+    def _setup_dataloader_from_config(self, config: Optional[Dict]):
+
+        OmegaConf.set_struct(config, False)
```
fayejf, line 202: Hi some of the configs sent to `_setup_dataloader_from_config` are not `DictConfig` type and it will raise error. Could you cast them either inside  `_setup_dataloader_from_config` or in argument?

fayejf (changes requested):
Overall LGTM. Thanks for this awesome contribution!  Just need to fix the  `_setup_dataloader_from_config` issue and pass the CI tests.

nemo/collections/asr/models/classification_models.py
diego-fustes: I've tried to do the casting in the argument, but I'm not sure how to test it

fayejf: Some config passed into `_setup_dataloader_from_config` is still type **dict** and set_struct for it will raise error. You can either have a check and caset before set_struct inside `_setup_dataloader_from_config`; or do similar thing as in `_setup_transcribe_dataloader` in classification_models:  `temporary_datalayer = self._setup_dataloader_from_config(config=DictConfig(dl_config))`

What do you mean test? test case? It's better to have it but please at least make sure to pass CI jenkins tests. Jobs failed due to this `type` issue.

fayejf (approved):
LGTM. Thanks for this great contribution and sorry for the delay

nemo/collections/asr/models/classification_models.py
```diff
@@ -468,51 +654,149 @@ def change_labels(self, new_labels: List[str]):
...
+        train_mse = self._mse(preds=logits, target=targets)
+        train_mae = self._mae(preds=logits, target=targets)
+
+        tensorboard_logs = {
+            'train_loss': loss,
+            'train_mse': train_mse,
+            'train_mae': train_mae,
+            'learning_rate': self._optimizer.param_groups[0]['lr'],
+        }
+
```
titu1994, line 737: @fayejf Could you send a PR to change use self.log(tensorboard_logs) and remove the 'log' part in the return below, for the train step? Thanks.

titu1994 (approved):
Everything looks great! Thanks for the wait, and for being patient with the ultra long review, but really nice PR in the end !

nemo/collections/asr/models/classification_models.py
diego-fustes: Should I use self.log(tensorboard_logs)?, it's asking for a key value. Maybe we should do:

`  for key in tensorboard_logs:
            self.log(f'{key}', tensorboard_logs[key], prog_bar=True)`

diego-fustes:
Thanks a lot for your support guys! I'll keep contibuting as much as I can :)

nemo/collections/asr/models/classification_models.py
titu1994: There's self.log_dict(). It's fine, we already merged it and it shoupd be fine for now

fayejf: @titu1994  Sure!

PR #1708, opened 2021-02-04
export() cleanup

borisfom:


nemo/collections/asr/models/asr_model.py
```diff
@@ -47,3 +51,61 @@ def multi_test_epoch_end(self, outputs, dataloader_idx: int = 0):
...
+            do_constant_folding=do_constant_folding,
+            keep_initializers_as_inputs=keep_initializers_as_inputs,
+            onnx_opset_version=onnx_opset_version,
+            try_script=try_script,
+            set_eval=set_eval,
+            check_trace=check_trace,
+            use_dynamic_axes=use_dynamic_axes,
+        )
+
+        output_model = attach_onnx_to_onnx(encoder_onnx, decoder_onnx, "DC")
```
drnikolaev, line 108: please use something unique here which is similar to class name like "ASRM" 
DC most probably taken and if someone would merge this onnx it's theoretically possible to get names clashed.

tkornuta-nvidia (commented):
@borisfom hi, I see that you removed "specialized export methods", as they are similar, so code cleanup.

Generally this is good, but my question is: should we do it now?

Yesterday people were testing if their trained models work, then exporting to Jarvis etc. This PR literally touches every export, so now what? Should we absorb this in TLT and ask them to do it once again?

@okuchaiev @ryanleary thoughts? What changes should be allowed for b4 in a given "stage"? (maybe good question is: in what stage NeMo b4 is?)

tkornuta-nvidia (changes requested):
Comment from Negar: We should not do this change. This is mostly for “cleanup”. At this stage, unless we have a serious blocker issue that needs to be fixed, we should not take any other changes. As you said we are in code freeze.

@borisfom  Those changes are still good, but not now and not for b4. Please push that PR towards `main`.

borisfom:
Well, those changes were not only for cleanup - I only did refactoring because I did not want to make same changes in many files. There were functional issues affecting the clients - input_example and output_example parameters to export() would be rejected. Essentially, no user would be able to export a model with custom fixed size inputs - like they can do with modules. I think it's essential feature and we should not be losing it. Another lack of functionality was that most export's did not support TS format at all (only one I think was).

borisfom:
Re QA effort - there is no need to re-run re-export of those models that were exported - my changes would not change the result of those. 
As for regression - they are covered by unit tests.

okuchaiev:
re-route to main

nemo/collections/asr/models/asr_model.py
borisfom: Well, we do not do multiple-level models yet, and actually, the collision would be guaranteed if we try to concat any 2 of the same models. A real fix, I think,  should be using "a." prefix on first and "b." prefix on second argument inside attach_onnx_to_onnx - no prefix arg needed.

borisfom:
Closing, re-routed to main.

PR #1711, opened 2021-02-04
WIP: Adding source and target language as a model parameter and removing i…

RPrenger:
…t as an option during inference. Also adding a different Detokenizer for Japanese based on this option

Signed-off-by: rprenger <rprenger@nvidia.com>

### DIFF
--- a/nemo/collections/common/tokenizers/ja_tokenizer.py
+++ b/nemo/collections/common/tokenizers/ja_tokenizer.py
@@ -1,6 +1,8 @@
-from sacremoses import MosesDetokenizer
 import re
 from typing import List
+
+from sacremoses import MosesDetokenizer
+
 
 class JADetokenizer:
     def __init__(self):
--- a/nemo/collections/common/tokenizers/ja_tokenizer.py
+++ b/nemo/collections/common/tokenizers/ja_tokenizer.py
@@ -1,0 +1,11 @@
+from sacremoses import MosesDetokenizer
+import re
+from typing import List
+
+class JADetokenizer:
+    def __init__(self):
+        self.moses_detokenizer = MosesDetokenizer(lang="ja")
+
+    def detokenize(self, translation: List[str]):
+        translation = self.moses_detokenizer.detokenize(translation)
+        return re.sub('▁', ' ', re.sub(' ', '', translation))
--- a/nemo/collections/common/tokenizers/ja_tokenizer.py
+++ b/nemo/collections/common/tokenizers/ja_tokenizer.py
@@ -1,3 +1,16 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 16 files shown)


nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -325,7 +331,7 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
         )
 
     @torch.no_grad()
-    def translate(self, text: List[str], source_lang: str = None, target_lang: str = None) -> List[str]:
```
ericharper, line 328: It looks like we have to keep these for backwards compatiblity. Can we check if they are None? And if so then we use the model attributes?

nemo/collections/common/tokenizers/ja_tokenizer.py
```diff
@@ -0,0 +1,13 @@
+import re
```
ericharper, line 1: Make sure to add the license header to this file.

ericharper (approved):
LGTM. Thanks for the contribution!

nemo/collections/common/tokenizers/ja_tokenizer.py
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import re
+from typing import List
+
+from sacremoses import MosesDetokenizer
+
+
+class JADetokenizer:
```
MaximumEntropy, line 20: Since MosesDetokenizer() is being run for all languages anyway on L359 of `mt_enc_dec_model.py` you could undo only sentencepiece within this class?

MaximumEntropy (changes requested):
LGTM overall, just a couple of minor things like adding target lang to MosesDetokenizer within the eval_step and potentially undoing sentencepiece within JADetokenizer.

nemo/collections/common/tokenizers/ja_tokenizer.py
RPrenger: Do you mean instead of having a JADetokenizer, modify the MosesDetokenizer?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -226,7 +229,10 @@ def eval_epoch_end(self, outputs, mode):
         ground_truths = list(itertools.chain(*[x['ground_truths'] for x in outputs]))
 
         # TODO: add target language so detokenizer can be lang specific.
-        detokenizer = MosesDetokenizer()
+        if self.tgt_language in ['ja']:
+            detokenizer = JADetokenizer()
+        else:
+            detokenizer = MosesDetokenizer()
```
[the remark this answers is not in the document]
RPrenger: Ok.  The original code doesn't pass in the language string, so I wasn't sure if it took a parameter, but it looks like the class just assumes `en` if no language is passed.  Fixing now

MaximumEntropy (approved):

PR #1714, opened 2021-02-04
CER computation added to speech_to_text_infer.py

jbalam-nv:
* Added CER option to speech_to_text_infer.py
* Cleaned up unused code in speech_to_text_sclite.py


examples/asr/speech_to_text_infer.py
```diff
@@ -49,7 +49,10 @@ def main():
     parser.add_argument("--batch_size", type=int, default=4)
     parser.add_argument("--wer_tolerance", type=float, default=1.0, help="used by test")
     parser.add_argument(
-        "--normalize_text", default=True, type=bool, help="Normalize transcripts or not. Set to False for non-English."
+        "--dont_normalize_text", default=False, action='store_true', help="Turn off trasnscript normalization. Recommended for non-English."
```
titu1994, line 52: Why is this arg being inverted ? It's only used for en, and every other language doesn't have normalization ?

examples/asr/speech_to_text_infer.py
```diff
@@ -90,10 +93,14 @@ def main():
...
-    wer_value = word_error_rate(hypotheses=hypotheses, references=references)
-    if wer_value > args.wer_tolerance:
-        raise ValueError(f"Got WER of {wer_value}. It was higher than {args.wer_tolerance}")
-    logging.info(f'Got WER of {wer_value}. Tolerance was {args.wer_tolerance}')
+
+    wer_value = word_error_rate(hypotheses=hypotheses, references=references, use_cer=args.use_cer)
+    if not args.use_cer:
+        if wer_value > args.wer_tolerance:
+            raise valueerror(f"got wer of {wer_value}. it was higher than {args.wer_tolerance}")
+        logging.info(f'Got WER of {wer_value}. Tolerance was {args.wer_tolerance}')
```
titu1994, line 101: What's `valueerror`?

titu1994 (changes requested):

examples/asr/speech_to_text_infer.py
jbalam-nv: No idea how that happened. Thanks!

examples/asr/speech_to_text_infer.py
jbalam-nv: That's right.  Previous default was True and it is the same behavior. Although previously to turn it off, we would have to pass '--normalize_text ""' to pass an empty string as bool False, which is annoying.

titu1994: Annoying yes, but we don't need to invert the flag to support CER. It will silently create bugs in measuring wer if you forget to pass this flag so it should be reverted.

titu1994 (approved):
LGTM, resolved confusion of default behaviour

PR #1715, opened 2021-02-04
Add save restore tests for .nemo checkpoints

titu1994:
# Changelog
- Adds tests for restore_from, overriden config restore_from, support for OmegaConf object instead of filepath for overrides and support for returning models internal config instead of instantiation of model

### DIFF
--- a/tests/core/test_save_restore.py
+++ b/tests/core/test_save_restore.py
@@ -14,9 +14,11 @@
 import os
 import shutil
 import tempfile
-from typing import Set
+from typing import Dict, Optional, Set, Union
 
 import pytest
+import torch
+from omegaconf import DictConfig, OmegaConf, open_dict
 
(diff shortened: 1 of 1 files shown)


titu1994:
@okuchaiev fyi

tests/core/test_save_restore.py
```diff
@@ -31,8 +33,54 @@ def getattr2(object, attr):
...
+    def __test_restore_elsewhere(
+        self,
+        model: ModelPT,
+        attr_for_eq_check: Set[str] = None,
+        override_config_path: Optional[Union[str, DictConfig]] = None,
+        map_location: Optional[torch.device] = None,
+        strict: bool = False,
+        return_config: bool = False,
+    ):
         """Test's logic:
```
tkornuta-nvidia, line 84: I like this logic. 👍

PR #1716, opened 2021-02-04
Waveglow model export

borisfom:
Also: multiple fixes to export_utils and refactoring of Exportables.


nemo/collections/asr/models/asr_model.py
```diff
@@ -47,3 +51,61 @@ def multi_test_epoch_end(self, outputs, dataloader_idx: int = 0):
         wer_denom = torch.stack([x['test_wer_denom'] for x in outputs]).sum()
         tensorboard_logs = {'test_loss': val_loss_mean, 'test_wer': wer_num / wer_denom}
         return {'test_loss': val_loss_mean, 'log': tensorboard_logs}
+
+    def export(
+        self,
```
titu1994, line 56: This will work - as long as the model has just encoder - decoder architecture. As this is the base class of all ASR Models - including RNNT models - this assumption is invalid. However, RNNT is under development so could you override the export() method in both RNNT model classes and explicitly raise a NotImplementedError()?

nemo/utils/export_utils.py
```diff
@@ -69,14 +69,16 @@ def expand_Conv1D(conv1d: nn.Module) -> Optional[nn.Conv2d]:
...
     conv2d.weight = nn.Parameter(conv1d.weight.unsqueeze(-1))
     # check that expansion is valid
     for _ in range(2):
-        sample_input = torch.rand(1, conv1d.in_channels, 256)
+        sample_input = torch.rand(1, conv1d.in_channels, 256).to(
+            device=conv1d.weight.device, dtype=conv1d.weight.dtype
+        )
         close = conv1d(sample_input).mean() - conv2d(sample_input.unsqueeze(-1)).squeeze().mean()
-        if close.abs() > 1e-6:
+        if close.abs() > 1.0:
```
titu1994, line 81: Isn't this tolerance way too high?

nemo/utils/export_utils.py
```diff
@@ -97,12 +99,45 @@ def expand_BatchNorm1d(bn1d: nn.Module) -> Optional[nn.BatchNorm2d]:
...
+    ).to(device=conv1d.weight.device, dtype=conv1d.weight.dtype)
+    conv2d.bias = conv1d.bias
+    conv2d.weight = nn.Parameter(conv1d.weight.unsqueeze(-1))
+    # check that expansion is valid
+    for _ in range(2):
+        sample_input = torch.rand(1, conv1d.in_channels, 256).to(
+            device=conv1d.weight.device, dtype=conv1d.weight.dtype
+        )
+        close = conv1d(sample_input).mean() - conv2d(sample_input.unsqueeze(-1)).squeeze().mean()
+        if close.abs() > 1.0:
```
titu1994, line 136: Isnt tolerance too high here?

nemo/collections/asr/models/asr_model.py
```diff
@@ -47,3 +51,61 @@ def multi_test_epoch_end(self, outputs, dataloader_idx: int = 0):
         wer_denom = torch.stack([x['test_wer_denom'] for x in outputs]).sum()
         tensorboard_logs = {'test_loss': val_loss_mean, 'test_wer': wer_num / wer_denom}
         return {'test_loss': val_loss_mean, 'log': tensorboard_logs}
+
```
titu1994, line 54: Doesnt ASRModel need to extend Exportable?

nemo/collections/asr/models/asr_model.py
titu1994: In fact, it can be more cleanly managed by making this ASR export method be in a mixin class - one for `ExportableEncDec` (which has this method in it) and another `ExportableRNNT` (which raises NotImplemented for now).

After that, let all the CTC and classification / label models extend the `ExportableEncDec` class which gives them proper export, and RNNT extend the `ExportableRNNT` which raises an error during export (for now).

This is infact how we add Subword support to ASR Models - https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/mixins.py

titu1994 (changes requested):
Overall looks good, but makes a wrong assumption for ASR models that they are all enc-dec type architectures - RNNT is not such a model. Minor refactoring to fix this.

Another thing is tolerance is quite high in some of the checks (>= 1) - is that expected / necessary or just an artifact during the refactor?

borisfom:
> Overall looks good, but makes a wrong assumption for ASR models that they are all enc-dec type architectures - RNNT is not such a model. Minor refactoring to fix this.
> 
Thanks! See my notes below.

> Another thing is tolerance is quite high in some of the checks (>= 1) - is that expected / necessary or just an artifact during the refactor?
I have found this high tolerance necessary for FP16 tests.

nemo/collections/asr/models/asr_model.py
borisfom: I am not quite happy with current approach to Model vs. Exportable "interface", actually. 
Current default implementation of Exportable mixin assumes Module and does not even always fit Model. 
Return value of export() is PT/ONNX graph for the module and 2 lists (file names and doc strings) for Model.
I believe Model baseclass should have export() method (returning 2 lists) that throws NotImplemented by default; the models that do implement export() can use additional mixins or utility functions to share export code patterns.

nemo/utils/export_utils.py
borisfom: Original tolerance failed on FP16 tests so I cranked it up. This is just a sanity check (which we may as well remove).

nemo/collections/asr/models/asr_model.py
titu1994: This makes sense, Exportable was ported by taking Nemo 0.x export example - which were at module level only, so the Exportable interface was designed to support module level export as well. 

I'm ok with the ASR side changes, just needs review by others to make sure export works properly.

nemo/core/classes/exportable.py
```diff
@@ -63,6 +63,60 @@ def get_format(filename: str):
...
+        if isinstance(input_example, Dict):
+            input_names = list(input_example.keys())
+        else:
+            if not (hasattr(self, 'input_types')):
+                raise NotImplementedError(
+                    'For export to work you must define input_types or pass names in input_example'
+                )
+            input_names = list(self.input_types.keys())
+        # remove unnecessary inputs for input_ports
+        for name in self.disabled_deployment_input_names:
```
titu1994, line 84: Where is the value `self.disabled_deployment_input_names` set ? Probably needs to check if hasattr(self, ...)?

nemo/core/classes/exportable.py
```diff
@@ -63,6 +63,60 @@ def get_format(filename: str):
...
+        if isinstance(output_example, Dict):
+            output_names = list(output_example.keys())
+        else:
+            if not (hasattr(self, 'output_types')):
+                raise NotImplementedError(
+                    'For export to work you must define output_types or pass names in output_example'
+                )
+            output_names = list(self.output_types.keys())
+            # remove unnecessary inputs for input_ports
+        for name in self.disabled_deployment_output_names:
```
titu1994, line 98: Same comments as above - maybe needs a hasattr(self, ...) first ?

titu1994 (commented):
Overall, looks ok from ASR side. Awaiting other reviews before final approval.

nemo/core/classes/exportable.py
borisfom: Why, it's there in Exportable ?

titu1994: Oh I missed that. LGTM then.

blisc (approved):
WaveGlow changes looks good

titu1994 (approved):
LGTM for ASR

PR #1717, opened 2021-02-04
Add save restore tests for .nemo checkpoints 

titu1994:
# Changelog
- Adds tests for restore_from, overriden config restore_from, support for OmegaConf object instead of filepath for overrides and support for returning models internal config instead of instantiation of model


okuchaiev (approved):
lgtm, thx

PR #1719, opened 2021-02-05
fix squad eval

yzhang123:
Fix error from Dec 17, 2020 https://github.com/NVIDIA/NeMo/pull/1569/files


ekmb:
@yzhang123 it's this needed for tlt? if so, it should go to b4

yzhang123:
> @yzhang123 it's this needed for tlt? if so, it should go to b4

I'm not sure about that status. @vladgets  could you take comment?

vladgets (approved):

PR #1720, opened 2021-02-05
fix AudioSegment for stereo audio

s-omranpour:
A fix to audio channel reduction in segment.py


titu1994 (commented):
This seems like an incorrect change? samples is a flattened vector of shape [num samples x num channels] - so if reshaped, it would be [num samples, channels]. So reducing over axis 1 is correct (averaging over channel)

PR #1721, opened 2021-02-05
fix squad eval

yzhang123:


tkornuta-nvidia (approved):
Discussed on Slack, approving

PR #1723, opened 2021-02-05
Fix qa config

yzhang123:
this is rebased on https://github.com/NVIDIA/NeMo/pull/1721
please merge after that


tkornuta-nvidia (approved):
Approving, but is this change related to TLT at all?

Seems you are just polishing a bit functionality of a QA training NeMo script...

yzhang123:
> Approving, but is this change related to TLT at all?
> 
> Seems you are just polishing a bit functionality of a QA training NeMo script...

Should not affect tlt

Issue #1700, opened 2021-02-03
Error when running evaluation and inference on QAModel


avikpal00:
**Bug Description**

Getting _Exception: cuda rng state model-parallel-rng is not added_ error when trying to evaluate/inference a pretrained QAModel with biomegatron lm after restoring the model from a .nemo file. However I am able to run the evaluate/inference right after training (i.e. trainer.fit).
Code :
`trainer.test(restored_model)`

### Traceback
<details>

``` python
Exception                                 Traceback (most recent call last)
<ipython-input-7-b3320e12a098> in <module>
      1 restored_model.setup_test_data(test_data_config=config.model.test_ds)
----> 2 trainer.test(restored_model)

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py in test(self, model, test_dataloaders, ckpt_path, verbose, datamodule)
    753 
    754         if model is not None:
--> 755             results = self.__test_given_model(model, test_dataloaders)
    756         else:
    757             results = self.__test_using_best_weights(ckpt_path, test_dataloaders)

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py in __test_given_model(self, model, test_dataloaders)
    818         self.testing = True
    819         self.model = model
--> 820         results = self.fit(model)
    821         self.testing = False
    822 

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py in fit(self, model, train_dataloader, val_dataloaders, datamodule)
    471         self.call_hook('on_fit_start')
    472 
--> 473         results = self.accelerator_backend.train()
    474         self.accelerator_backend.teardown()
    475 

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/accelerators/gpu_accelerator.py in train(self)
     64 
     65         # train or test
---> 66         results = self.train_or_test()
     67         return results
     68 

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/accelerators/accelerator.py in train_or_test(self)
     65     def train_or_test(self):
     66         if self.trainer.testing:
---> 67             results = self.trainer.run_test()
     68         else:
     69             results = self.trainer.train()

/opt/conda/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py in run_test(self)
    660         # self.reset_test_dataloader(ref_model)
    661         with self.profiler.profile("run_test_evaluation"):
[...]

PR #1724, opened 2021-02-05
Add test stage to setup to finish megatron init

ericharper:
If only trainer.test was being called then megatron was not finishing it's init.

This should fix this issue: https://github.com/NVIDIA/NeMo/issues/1700


yzhang123 (approved):
Thank you!

PR #1727, opened 2021-02-08
Adding working Japanese option to SacreBleu

RPrenger:
Needed to uninstall the version of `sacrebleu` that was in the existing container and then install `sacrebleu[ja]`.  Also changed the validation scoring to use the Japanese `ja-mecab` tokenizer with sacrebleu instead of `13a`.


okuchaiev (approved):
lgtm, thanks!

PR #1732, opened 2021-02-08
Dev deps cnt

tkornuta-nvidia:
- [x] Bumping version of EFF to 0.2.6 (changes required by Yang's PR to TLT)
- [x] Added 'nvidia-pyindex' to setup_requires

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -13,4 +13,3 @@ transformers>=4.0.1
 sentencepiece<1.0.0
 webdataset
 nvidia-eff>=0.2.6
-onnx-graphsurgeon>=0.2.7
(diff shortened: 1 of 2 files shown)


borisfom (approved):

borisfom (changes requested):
Actually. this would break Mac and other platforms Nemo is being used on.

requirements/requirements.txt
```diff
@@ -12,3 +12,5 @@ hydra-core>=1.0.4
 transformers>=4.0.1
 sentencepiece<1.0.0
 webdataset
+nvidia-eff>=0.2.6
+onnx-graphsurgeon>=0.2.7
```
borisfom, line 16: graph_surgeon should not go into requirements.txt but listed in ‘install_requires’ arg of setup.py - conditionally : if platform is one of those we currently have wheel published.

borisfom:
We should close this PR - its sole purpose was to enable ONNX export for Megatron (in Nemo b3 environment) - which we have found is broken anyway in b4 (due to Pytorch version).
So it does not solve the issue for our standard containers and introduces risks.

tkornuta-nvidia:
please note I still need EFF.

PR #1743, opened 2021-02-10
Add container support to `typecheck`

titu1994:
# Changelog
- Add proper depth analysis for nested type definition
- Add several tests to capture cases occurring due to container checks
- Add typecheck argument `ignore_collections` in order to fall back to previous behavior - ignore nesting entirely
- Update type definition of older code to correctly declare nested types


okuchaiev (approved):
LGTM. Thank you!

blisc (approved):
The code LGTM, but I think we need a documentation update for neural types somewhere that concisely and accurately describe the current behaviour.

titu1994:
Agreed about docs - maybe in a notebook specifically for types ? Or maybe the api docs in RTD

PR #1749, opened 2021-02-11
Chinese normalization and tokenization at inference

MaximumEntropy:
Adds Jieba word segmentation and opencc traditional to simplified Chinese normalization


okuchaiev (approved):
lgtm, thx

PR #1750, opened 2021-02-11
Megatron fixes: lazy init moved back to module for inference to work

borisfom:
lazy int() only during fit() led to failures in forward() for inference scripts.


vladgets (dismissed):
I generally approve since I was part of this discussion on Slack. But the validity of specific changes should be approved by @ericharper

ericharper (approved):
LGTM. Thanks!

PR #1760, opened 2021-02-16
Update webdataset call to support v0.1.48

titu1994:
# Changelog
- Update API for webdataset calls
- Add lower bound for webdataset requirements (may want to pin it for release, @okuchaiev)


titu1994:
@ericharper Can you pull this branch and verify NMT models have no issue running their WebDataset training on some toy / small dataset?

@fayejf Same for ASR Classification models

I've verified that ASR CTC/RNNT work 1:1 as before.

ericharper (approved):
LGTM. I tested on both Toy WMT and En-De and everything is working as before.

Thanks!

PR #1762, opened 2021-02-17
Moving calculating the class weights to the data loader creation method.

VahidooX:
It is necessary to support class_balancing in TTL.

### DIFF
--- a/nemo/collections/nlp/models/text_classification/text_classification_model.py
+++ b/nemo/collections/nlp/models/text_classification/text_classification_model.py
@@ -77,13 +77,7 @@ class TextClassificationModel(NLPModel, Exportable):
             idx_conditioned_on=0,
         )
 
-        # create the loss module if it is not yet created by the training data loader
-        if not self.loss:
-            if self.class_weights:
-                # You may need to increase the number of epochs for convergence when using weighted_loss
-                self.loss = CrossEntropyLoss(weight=self.class_weights)
-            else:
-                self.loss = CrossEntropyLoss()
+        self.create_loss_module()
 
         # setup to track metrics
         self.classification_report = ClassificationReport(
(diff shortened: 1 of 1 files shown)


vladgets (approved):

nemo/collections/nlp/models/text_classification/text_classification_model.py
```diff
@@ -200,6 +190,14 @@ def setup_training_data(self, train_data_config: Optional[DictConfig]):
...
         self._train_dl = self._setup_dataloader_from_config(cfg=train_data_config)
 
+        # we need to create/update the loss module by using the weights calculated from the training data
+        if self.cfg.dataset.class_balancing == 'weighted_loss':
+            self.class_weights = calc_class_weights(train_data_config.file_path, self.cfg.dataset.num_classes)
+            self.loss = CrossEntropyLoss(weight=self.class_weights)
+        else:
+            self.class_weights = None
+            self.loss = CrossEntropyLoss()
+
```
ekmb, line 200: I would rather have this in a separate method for the users to call but this would require changes on the TLT side.

ekmb:
Can we still make changes to b4 branch? @tkornuta-nvidia @okuchaiev ?

VahidooX:
> 
> 
> Can we still make changes to b4 branch? @tkornuta-nvidia @okuchaiev ?

Which branch should we use for bug fixes needed for TLT?

nemo/collections/nlp/models/text_classification/text_classification_model.py
VahidooX: Moved it into a method.

ekmb (approved):

tkornuta-nvidia (approved):
ok, looks good.

PR #1772, opened 2021-02-19
Add support for computing CTC / RNNT alignments

titu1994:
# Changelog
- Adds support for CTC/RNNT models to compute alignments for their respective decoding schemes. Valid schemes listed below.
- Adds support for CTC models to return a `Hypothesis` object that contains additional information about the decoded / transcribed tensor.
- Adds test for hypothesis in CTC

## Valid alignment computations
- CTC
  - CTC + Char encoding
  - CTC + Subword encoding
- RNNT
  - RNNT with Greedy (sample) and Greedy (batched, both variants)
  - RNNT with Basic Beam (beam size >= 1)

## Invalid alignment computations
- RNNT
  - RNNT with TSD / ALSD

# Example (CTC / CTC-BPE)

```python
import torch
...

trainer = Trainer(...)
asr_model = EncDecCTCModel.from_pretrained(pretrained_name, map_location=torch.device('cpu'), strict=True)
asr_model.set_trainer(trainer)

device = torch.device('cuda:0') if torch.cuda.is_available() else torch.device('cpu')
asr_model = asr_model.to(device)

# Compute alignments
audio_paths = [...]  # path of audio files
hypotheses = asr_model.transcribe(audio_paths , batch_size=32, return_hypotheses=True)

SAMPLE_IDX = 0
h = hypotheses[SAMPLE_IDX]  # type: Hypothesis
alignments = h.alignments  # log probs of first sample

token_set = []
for p in alignments:
    tokens = asr_model._wer.decode_tokens_to_str([p])
    token_set.append(tokens)

print("Predicted text :", h.text)
print()
print("Alignments :")
for t, p in enumerate(token_set):
    print(f"t = {t:4d} : {p}")
```

# Example (RNNT / RNNT-BPE)

```python
import torch
from omegaconf import OmegaConf, open_dict
...

trainer = Trainer(...)
asr_model = EncDecRNNTBPEModel.restore_from(<path to RNNT model>, map_location=torch.device('cpu'), strict=True)
asr_model.set_trainer(trainer)

device = torch.device('cuda:0') if torch.cuda.is_available() else torch.device('cpu')
asr_model = asr_model.to(device)

# Change decoding strategy to preserve alignments
decoding_cfg = asr_model.cfg.decoding

# Add `preserve_alignments` flag to RNNT models.
with open_dict(decoding_cfg):
    decoding_cfg.preserve_alignments = True

asr_model.change_decoding_strategy(decoding_cfg)  # cfg.model.decoding is the `decoding` portion of an RNNT yaml config.

# Compute alignments
audio_paths = [...]  # path of audio files
hypotheses = asr_model.transcribe(audio_paths , batch_size=32, return_hypotheses=True)

# Above hypotheses is a list of 2 items
# [0] - best Hypothesis. This is always available, for both greedy and beam search
# [1] - NBestHypothesis (if beam search size > 1). None for greedy and beam search size of 1.

SAMPLE_IDX = 0
hypotheses = hypotheses[0]  # select best hypothesis
h = hypotheses[SAMPLE_IDX]  # type: Hypothesis
alignments = h.alignments  # log probs of first sample

token_set = []
for p in alignments:
    tokens = [asr_model.decoding.decode_tokens_to_str([p_i]) for p_i in p]
    token_set.append(tokens)

print("Predicted text :", h.text)
print()
print("Alignments  :")
for t, p in enumerate(token_set):
    print(f"t = {t:4d} : {p}")
```

### DIFF
--- a/nemo/collections/asr/metrics/rnnt_wer.py
+++ b/nemo/collections/asr/metrics/rnnt_wer.py
@@ -95,7 +95,7 @@ class AbstractRNNTDecoding(ABC):
                 joint_model=joint,
                 blank_index=self.blank_id,
                 max_symbols_per_step=self.cfg.greedy.get('max_symbols', None),
-                preserve_logprobs=self.cfg.greedy.get('preserve_logprobs', False),
+                preserve_alignments=self.cfg.greedy.get('preserve_alignments', False),
             )
 
         elif self.cfg.strategy == 'greedy_batch':
@@ -104,7 +104,7 @@ class AbstractRNNTDecoding(ABC):
                 joint_model=joint,
                 blank_index=self.blank_id,
                 max_symbols_per_step=self.cfg.greedy.get('max_symbols', None),
-                preserve_logprobs=self.cfg.greedy.get('preserve_logprobs', False),
+                preserve_alignments=self.cfg.greedy.get('preserve_alignments', False),
             )
 
         elif self.cfg.strategy == 'beam':
@@ -116,7 +116,7 @@ class AbstractRNNTDecoding(ABC):
                 return_best_hypothesis=decoding_cfg.beam.get('return_best_hypothesis', True),
                 search_type='default',
                 score_norm=self.cfg.beam.get('score_norm', True),
-                preserve_logprobs=self.cfg.greedy.get('preserve_logprobs', False),
+                preserve_alignments=self.cfg.greedy.get('preserve_alignments', False),
             )
 
         elif self.cfg.strategy == 'tsd':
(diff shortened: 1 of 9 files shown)


VahidooX:
> ```python
> change_decoding_strategy
> ```

Can we call the change_decoding_strategy automatically in the transcribe method if necessary so that user does not need to do that?

titu1994:
1) Calling it automatically is wrong - user might not want to change config or might not want to silently update config. That's why there needs to be explicit call. 

2) CTC is cheap to decode, just argmax. RNNT is Autoregressive and expensive to compute alignments and preserve them (each alignment gets pushed onto CPU). So it needs a special flag to prevent degradation of performance in general case when used doesn't care about alignments.

3) This is unified already per model type. CTC doesn't need an additional decoding class. RNNT does. The two models are too different to share more attributes. 

4) we don't need to add preserve_alignments to config, it's optional. Eventually we will have a tutorial for RNNT models which will demonstrate this.

VahidooX (approved):

PR #1781, opened 2021-02-22
Ja Source Language Preprocessing

mchrzanowski:
This PR adds preprocessing for Japanese as a source language to nemo. Previously, support only included Japanese when it was a target language.


nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -77,6 +78,7 @@ def __init__(self, cfg: MTEncDecModelConfig, trainer: Trainer = None):
 
         self.src_language: str = cfg.get("src_language", None)
         self.tgt_language: str = cfg.get("tgt_language", None)
+        self.sentencepiece_model = cfg.get("sentencepiece_model", None)
```
okuchaiev, line 81: @ericharper, I think "sentencepiece_model" should be registered as artifact, so that it gets included inside .nemo file, right?

okuchaiev (approved):
lgtm

PR #1786, opened 2021-02-24
Add Transcription script for all ASR models 

titu1994:
# Changelog
- Add single script to allow transcription of all ASR models
- Add subclass walk for finding all Pretrained models of ASRModel. 
- Add dataclasses for RNNT Decoding strategy


examples/asr/transcribe_speech.py
```diff
@@ -0,0 +1,122 @@
...
+        # restore model from .nemo file path
+        model_cfg = ASRModel.restore_from(restore_path=cfg.model_path, return_config=True)
+        classpath = model_cfg.target  # original class path
+        imported_class = model_utils.import_class_by_path(classpath)  # type: ASRModel
+        logging.info(f"Restoring model : {imported_class.__name__}")
+
+        asr_model = imported_class.restore_from(restore_path=cfg.model_path, map_location=device)  # type: ASRModel
+    else:
+        # restore model by name
+        asr_model = ASRModel.from_pretrained(model_name=cfg.pretrained_name, map_location=device)  # type: ASRModel
```
blisc, line 83: Can you add an unittest that checks if this works?

titu1994: Added, trying to get Jenkins to actually run these tests

titu1994: Unittests dont have access to examples/asr/* so had to put the test in Jenkins instead

titu1994: Tests now run on this branch - but I had to explicitly set the branch name. Needs to be reverted to main when merged back. @okuchaiev

blisc (approved):

PR #1794, opened 2021-02-24
Adding option to always create .nemo file when writing checkpoint

RPrenger:
Right now we only get a `.nemo` file when the training is finished.  However, it is common to let training continue for an unspecified amount of epochs and just stop when the training stops improving, in which case we'd never get a `.nemo` file.

This commit adds the option save the nemo file whenever a checkpoint is saved.

Use:

`+exp_manager.checkpoint_callback_params.always_save_nemo = True`

Signed-off-by: rprenger <rprenger@nvidia.com>


titu1994:
Could you make sure that at any point only 1 ".nemo" file exists ? We dont want to clutter with dozens of nemo + dozens of PTL checkpoints for consecutive runs.

RPrenger:
> Could you make sure that at any point only 1 ".nemo" file is saved ? We dont want to clutter with dozens of nemo + dozens of PTL checkpoints for consecutive runs.

That's how it works now.  It'll either overwrite the old file with the latest or overwrite it with the best each time depending on your `save_best_model` option.

RPrenger:
I just realized that if "save_best_model" is

titu1994:
Cool that sounds nice. Maybe a good default for this is True ? @blisc for final review

okuchaiev:
also, @blisc please decide whether this OK for rc (looks small enough of a change to me) or should stay on main

nemo/utils/exp_manager.py
```diff
@@ -564,12 +565,26 @@ class NeMoModelCheckpoint(ModelCheckpoint):
...
+        output = super().on_save_checkpoint(trainer, pl_module)
+         
+        if not self.always_save_nemo:
+            return output
+
+        # Load the best model and then re-save it
+        if self.save_best_model:
+            trainer.checkpoint_connector.restore(self.best_model_path, on_gpu=trainer.on_gpu)
+        pl_module.save_to(save_path=os.path.join(self.dirpath, self.prefix + self.postfix))
+        return output
```
blisc, line 587: Does this save only 1 .nemo file? Or will it save 1 nemo everytime this function is called (since we don't delete the old checkpoints)?

nemo/utils/exp_manager.py
```diff
@@ -564,12 +565,26 @@ class NeMoModelCheckpoint(ModelCheckpoint):
...
+    @rank_zero_only
+    def on_save_checkpoint(self, trainer, pl_module):
+        output = super().on_save_checkpoint(trainer, pl_module)
+         
+        if not self.always_save_nemo:
+            return output
+
+        # Load the best model and then re-save it
+        if self.save_best_model:
+            trainer.checkpoint_connector.restore(self.best_model_path, on_gpu=trainer.on_gpu)
```
blisc, line 585: This doesn't seem correct. As we will be restoring during training and if your validation error doesn't decrease in that epoch, it would continuously loop here.

blisc (changes requested):
Overall, I really don't like this change. Lightning already has a robust checkpoint saving callback, and we really should just leverage what they have rather than re-writing existing code.

If for whatever reason, you want a .nemo file, then it's probably better to intercept the keyboardinterrupt or other killing signal, and then calling NeMoModelCheckpoint.on_train_end()

nemo/utils/exp_manager.py
blisc: Nvm, I see Som already asked this.

nemo/utils/exp_manager.py
blisc: I see that you already commented on this issue. Will wait until it to be resolved

blisc:
Lightning's callbacks have `on_keyboard_interrupt`, https://pytorch-lightning.readthedocs.io/en/stable/extensions/callbacks.html#on-keyboard-interrupt. Maybe we should just leverage that instead.

okuchaiev:
@blisc  this is for scenario when training takes several days, but you want to be able to grab intermediate results and evaluate them/deploy intermediate result to debugging tool, etc.

nemo/utils/exp_manager.py
RPrenger: That problem is fixed below.

RPrenger:
> Lightning's callbacks have `on_keyboard_interrupt`, https://pytorch-lightning.readthedocs.io/en/stable/extensions/callbacks.html#on-keyboard-interrupt. Maybe we should just leverage that instead.

If we use the keyboard interrupt and we have a long running job, we can't have a .nemo file unless we kill our job.  We should be able to deploy our best model while it's still training.

blisc:
> > Lightning's callbacks have `on_keyboard_interrupt`, https://pytorch-lightning.readthedocs.io/en/stable/extensions/callbacks.html#on-keyboard-interrupt. Maybe we should just leverage that instead.
> 
> If we use the keyboard interrupt and we have a long running job, we can't have a .nemo file unless we kill our job. We should be able to deploy our best model while it's still training.

I'm confused as to why the lightning .ckpt isn't sufficient for your needs.

mchrzanowski:
for translation, we have to use a model to produce segmented text before it's fed into the Nemo model (ie, sentencepiece processor). this must be stored alongside the model, but it is not itself part of the model we're training. a Nemo file contains this, but I dont think ckpt files would.

nemo/utils/exp_manager.py
```diff
@@ -564,12 +567,38 @@ class NeMoModelCheckpoint(ModelCheckpoint):
...
+        if not self.always_save_nemo:
+            return output
+
+        # Load the best model and then re-save it
+        if self.save_best_model:
+            if not os.path.exists(self.best_model_path):
+                return output
+           
+            old_state_dict = deepcopy(pl_module.state_dict())
+            checkpoint = torch.load(self.best_model_path, map_location='cpu')
```
blisc, line 591: Can we cache `best_model_path`, so that we can return if there was no change in best model?

RPrenger: Yep that's a good idea.  I'll add a field in the NeMoModelCheckpoint to hold the previous one.

blisc (approved):

PR #1803, opened 2021-02-25
Fix megatron vocab file

ericharper:
Megatron-LM is not using the vocab file specified by the user. This PR fixes that.


ekmb (approved):
LGTM. Thanks!

PR #1804, opened 2021-02-25
Hotfix for en/ ja preprocessing

mchrzanowski:
1) hotfix for https://github.com/NVIDIA/NeMo/pull/1801. one call to `get_detokenizer` wasn't updated.
2) @MaximumEntropy noticed that the previous logic would apply the en/ja preprocessing in the case that the user did src == tgt (for whatever reason they wanted to do this). The current logic fixes this.


RPrenger (approved):
LGTM

PR #1805, opened 2021-02-26
[WIP] Refactoring translation routines

okuchaiev:
This PR does the following:
* Makes sure that all pipeline pieces are create inside the constructor: src_normalizer, src_pre_tokenizer, encoder/decoder_tokenizers, tgt_post_tokenizer. This is necessary for their artifacts to be handled correctly and be included in .nemo file
* Refactors "batch_translate", "translate" and "eval_step" so that:
  - eval_step relies on batch_translate
  - translate and batch_translate call pipeline pieces, hopefully, in a correct order 

TODOs:
translate method should really just prepare input batch and call batch_translate so that core pipeline logic is implemented in a single place (batch_translate) 

@ericharper I suspect if we the conversion of raw text training data to tarred files might be missing corrsponding src_normalizer and src_pre_tokenizer calls.


MaximumEntropy (approved):
LGTM. Needs fixes for En <-> Ja, which will happen in a subsequent PR.

ericharper (approved):
Really nice PR, thanks!

PR #1806, opened 2021-02-26
Concatenation of tarred datasets for NMT

MaximumEntropy:
This PR lets you combine multiple TarredTranslationDatasets and sample minibatches from them with different ratios. Right now our config setup allows specifying only two tarred datasets at a time, but the `ConcatTarredTranslationDataset` should support an arbitrary number of datasets.


ericharper:
Can you add example usage to the README?

titu1994 (commented):
this looks fine, but have you tested it to make sure it doesnt hang under DDP?

PR #1807, opened 2021-02-26
Add basic grpc MT server

ryanleary:
Add readme, server updates

Signed-off-by: Ryan Leary <rleary@nvidia.com>


examples/nlp/machine_translation/nmt_server/server.py
```diff
@@ -0,0 +1,102 @@
...
+        lang_pair = f"{request.source_language}-{request.target_language}".lower()
+        if lang_pair not in self._models:
+            context.set_code(grpc.StatusCode.INVALID_ARGUMENT)
+            context.set_details(f"Invalid lang pair {lang_pair} requeste")
+            return nmt.TranslateTextResponse()
+
+        request_strings = [x for x in request.texts]
+
+        for batch in batches(request_strings, self._batch_size):
+            batch_results = self._models[lang_pair].translate(text=batch, source_lang=None, target_lang=None)
```
MaximumEntropy, line 83: Is there a reason why source and target lang is None here? This won't use the correct input tokenization and target detokenization.

ryanleary: Nope, no good reason. Better question is why doesn't the model already know the `source_lang` and `target_lang` if it matters? It's being restored from a nemo model.. that data should already be there.

MaximumEntropy: We've added that now and new models won't require this (#1805). But for existing models, and for temporary backward compatibility, the translate method has a source_lang and target_lang that need to be set.

tools/nmt_grpc_service/api/nmt_pb2.py
```diff
@@ -0,0 +1,286 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
ericharper, line 1: 2021

tools/nmt_grpc_service/api/nmt_pb2_grpc.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
ericharper, line 1: 2021

tools/nmt_grpc_service/nmt.proto
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+//
+// NVIDIA CORPORATION and its licensors retain all intellectual property
+// and proprietary rights in and to this software, related documentation
+// and any modifications thereto.  Any use, reproduction, disclosure or
+// distribution of this software and related documentation without an express
+// license agreement from NVIDIA CORPORATION is strictly prohibited.
```
ericharper, line 7: Why is this license different than the others?

tools/nmt_grpc_service/server.py
```diff
@@ -0,0 +1,175 @@
...
+            model = nemo_nlp.models.machine_translation.MTEncDecModel.restore_from(restore_path=model_path)
+            model.beam_search.beam_size = self._beam_size
+            model.beam_search.len_pen = self._len_pen
+            model.beam_search.max_delta_length = self._max_delta_length
+        else:
+            raise NotImplemented(f"Only support .nemo files, but got: {model_path}")
+
+        if not hasattr(model, "src_language") or not hasattr(model, "tgt_language"):
+            raise ValueError(f"Could not find src_language and tgt_language in model attributes")
+
```
ericharper, line 113: Can we add these to the argparser? Then if a model doesn't have them, it could be updated by the user specified languages?

ericharper (approved):
LGTM! Thanks!

Only some minor comments.

tools/nmt_grpc_service/nmt.proto
MaximumEntropy: Fixed, thanks!

PR #1820, opened 2021-03-01
Fix enja decoding

mchrzanowski:
the current en<>ja preprocessing pipeline is being shifted to using integers for intermediate steps. this will avoid tokenization issues between yttm, sentencepiece, and sacremoses due to parsing incomptabilities. this PR changes the detokenization to deal with integer inputs and output tokens only at the end.


mchrzanowski:
Will do at some point soon. don’t let me block you for now and merge away;
I’ll fix things in a subsequent pr later

mchrzanowski:
i think this PR hopefully wont have to be merged. i think the path of least resistance for en<>ja will be to have the preprocessing pipeline produce integer ids as opposed to tokens. that way, there's no ambiguity between sequences of chars or tokens. detokenization will i think at that point just work with the default pathway and we wont have to special case en<>ja like this draft does.

PR #1829, opened 2021-03-02
Fixing NMT DDP override that no longer works with PTL 1.2 

ericharper:
Also, need to update/add tests that would have caught this problem.


mchrzanowski:
thanks @ericharper !

blisc (approved):

PR #1835, opened 2021-03-03
Fix enja decoding (round 2)

mchrzanowski:
after trying hard to make yttm and spm play nicely together, i am giving up on the latter. the en<>ja pipeline is now moses -> yttm. from what i can see, this fixes tokenization & detokenization issues we were seeing (eg, `&apos;` was being split apart incorrectly during tokenization, preventing it from being detokenized correctly).


nemo/collections/common/tokenizers/en_ja_tokenizers.py
```diff
@@ -43,22 +38,13 @@ def detokenize(self, tokens: List[str]) -> str:
...
         Tokenizes text using Moses -> Sentencepiece.
         """
-        if self.sp_tokenizer is None:
-            raise ValueError("Need valid sp_tokenizer_model_path, found None")
-        text = self.moses_tokenizer.tokenize(text, escape=False, return_str=True)
-        return self.sp_tokenize(text)
+        return self.moses_tokenizer.tokenize(text)
 
     def normalize(self, text):
         return self.normalizer.normalize(text)
```
MaximumEntropy, line 50: Do you do punct normalization for Japanese? When I did this for Chinese, it replaced the period "。" with the English one ".".

nemo/collections/common/tokenizers/en_ja_tokenizers.py
```diff
@@ -43,22 +38,13 @@ def detokenize(self, tokens: List[str]) -> str:
...
 
     def tokenize(self, text):
         """
         Tokenizes text using Moses -> Sentencepiece.
         """
-        if self.sp_tokenizer is None:
-            raise ValueError("Need valid sp_tokenizer_model_path, found None")
-        text = self.moses_tokenizer.tokenize(text, escape=False, return_str=True)
-        return self.sp_tokenize(text)
+        return self.moses_tokenizer.tokenize(text)
```
MaximumEntropy, line 47: Do you need to set `escape=False` here?

MaximumEntropy (commented):
Just a couple of questions regarding Moses normalization and tokenization for Japanese, but LGTM otherwise.

nemo/collections/common/tokenizers/en_ja_tokenizers.py
mchrzanowski: i think escaping is fine for en<>ja. the preprocessing pipeline escapes, do we have to unescape. and i checked to make sure it is undone correctly. but are you concerned about this behavior in general?

nemo/collections/common/tokenizers/en_ja_tokenizers.py
mchrzanowski: ah good catch, japanese uses that period and the normalization was screwing it up

PR #1853, opened 2021-03-05
Freeze modules during transcribe to prevent graph accumulation during transcribe loop

titu1994:
# Changelog
- Freeze internal modules when transcribing data so as to prevent graph from accumulating data during loop


VahidooX (approved):

okuchaiev (changes requested):
shouldn't this go to r1.0.0rc1 ?

titu1994:
@okuchaiev done

okuchaiev (approved):

PR #1863, opened 2021-03-08
EnJa tokenize output format fix

mchrzanowski:
~preprocessor `tokenize(txt)` methods should probably return a list of tokens, not a string. these can then be converted into ids without the intermediate steps of producing a string and then immediately splitting it for the tokens again, as was being done in `translate`~

it turns out that `text_to_ids(txt)` acts differently than `tokens_to_ids(txt.split(' '))`, so i am changing the output of the `processsor.tokenize` to be `str` again.


mchrzanowski:
buncha tests are failing but i dont think theyre related to this pr

okuchaiev (approved):

Issue #1868, opened 2021-03-09
torch 1.8.0 incompatibility: pytorch-lightening needs to be updated to 1.2.3, but is not enough.


tiborkiss:
**Describe the bug**

Due to `torch` 1.8.0 just was released a few days ago, `pytorch-lightning` is still incompatible with the new torch.
Since NeMo setup pulls the latest stable torch release, imports this problem.
The problem is not visible during the install, but when using `pytorch-lightning` within NeMo tutorials for example.
 
**Steps/Code to reproduce bug**

I installed NeMo from r1.0.0rc17
```
python -m pip install git+https://github.com/NVIDIA/NeMo.git@r1.0.0rc1#egg=nemo_toolkit[all]
```
It was working while the official stable pytorch was 1.7.1.
Since PyTorch 1.8.0 came out, in fresh venv NeMo's requirements are pulling the latest torch, which is 1.8.0,
With torch 1.8.0 the second ASR tutorial notebook fails while using `pytorch-lightning` because `torchtext.data.Batch` was deprecated (and moved to` torchtext.legacy.data.Batch`, which breaks PyTorch Lightning. For more information read: https://github.com/PyTorchLightning/pytorch-lightning/pull/6168

**Expected behavior**

Since the bug is indirectly caused by the incompatibility between the current `pytorch-lightning` installed by NeMo and the latest `torch` 1,8.0, the only temporary solution I see that NeMo should stick with `torch<=1.7.1` and it's associated libraries.
I tried to edit the requirements by myself, then `./reinstall.sh`, unfortunately something is still pulling 1.8.0 somewhere at the end of the setup process.
The expected behavior is to be able to stick to torch==1.7.1

**Environment overview (please complete the following information)**

 - Environment location: WSL2 with GPU support and Ubuntu 20.04

**Environment details**

- OS version: Ubuntu 20.04 running in WSL2 with GPU support.
- PyTorch version: 1.8.0
- Python version: 3.8.5

**Additional context**

I think, I completely described the problem and the possible solution.

blisc:
@tiborkiss Please note that NeMo RC1 does not work on lightning 1.2.0. In fact, RC1 pins Lightning <= 1.1.5. If you need Lightning 1.2.0 or high, please use the NeMo main branch

tiborkiss:
Thank you for the tips @blisc .
Looks like the shortest workaround is to stick with torch 1.7.1

tiborkiss:
[...]

PR #1869, opened 2021-03-09
After upgrading P.Lightening to 1.2.3, remains other incompatibilities with torch 1.8.0. A workaround until is sorted out all, is to stick with torch 1.7.1.

tiborkiss:
… pull down 1.8.0. https://github.com/PyTorchLightning/pytorch-lightning/pull/6168

This diff temporarily solves the imported incompatibility reported in https://github.com/NVIDIA/NeMo/issues/1868


requirements/requirements.txt
```diff
@@ -2,7 +2,7 @@ numpy>=1.18.2
 onnx>=1.7.0
 pytorch-lightning>=1.1.0,<=1.1.5
 python-dateutil
-torch
```
okuchaiev, line 5: we can't pin torch to the exact version because it will force torch re-install if the user has slightly different version (e.g. 1.7.0) or inside NVIDIA's container (20.11-py3) which has version like (1.8.0a0+17f8c32 --- this is before upstream's 1.8)

okuchaiev (changes requested):
@tiborkiss thanks for the PR. However, we can't pin torch version exactly. PLT team is going to address 1.8 incompatibility soon, so giving a different range for PTL version could be a better solution here.

requirements/requirements.txt
tiborkiss: @okuchaiev thanks for the suggestion. 
I tried to let torch as it was and just changed the range for `pytorch-lightning>=1.2.0,<=1.2.2`.
At this moment 1.2.2 is the latest and 1.2.0 seems to have some PyTorch 1.8 compatibility fixes.

I recreate the env from the scratch with `./reinstall.sh` with the above PLT range changes, but the issue is still there. Cannot choose newer version from PLT.
Then I found somewhere a workaround like
```
torchtext>=0.5, <0.7 # TODO: temporary fix fix for compatibility
```
I downgraded `torchtext` to 0.7, then complained about `_torchtext.so` undefined symbol. Probably it will not be possible to downgrade just `torchtest`.

I have no other idea, then just for myself I come back to torch 1.7.1 and if somebody will install these days, may run in the same issue and do like myself.

tiborkiss: Now, `pytorch-lightening` 1.2.3 came out, which has the fix. I did a test. This PLT version incompatibility is fixed, but came out another. You can read https://github.com/NVIDIA/NeMo/issues/1868

okuchaiev:
thanks again @tiborkiss  but we'll close this one for now because we can't pin torch version to the exact number like 1.7.1 We could add something like this right before upload to pypi though.

PR #1777, opened 2021-02-20
Update/Fix for classification models

fayejf:
1. Change self.log(tensorboard_logs) and remove the 'log' part in the return  requested in  PR #[1707](https://github.com/NVIDIA/NeMo/pull/1707) 
2. Fix for logging for top-k acc metrics
3. Fix **tarred** dataloader for speech regression support

PR #1878, opened 2021-03-10
self.log and fix top-k acc metrics 

fayejf:
Part of [PR#1777](https://github.com/NVIDIA/NeMo/pull/1777) that should go to 1.0.0RC


nemo/collections/asr/models/classification_models.py
```diff
@@ -396,11 +397,13 @@ def multi_validation_epoch_end(self, outputs, dataloader_idx: int = 0):
         self._accuracy.total_counts_k = total_counts
         topk_scores = self._accuracy.compute()
 
-        tensorboard_log = {'val_loss': val_loss_mean}
+        self.log('val_loss', val_loss_mean)
```
titu1994, line 400: For multi validation and multi test, return the dictionary as before. Internally it will prepend dataset names and then call self.log_dict()

nemo/collections/asr/models/classification_models.py
```diff
@@ -411,11 +414,13 @@ def multi_test_epoch_end(self, outputs, dataloader_idx: int = 0):
         self._accuracy.total_counts_k = total_counts
         topk_scores = self._accuracy.compute()
 
-        tensorboard_log = {'test_loss': test_loss_mean}
+        self.log('test_loss', test_loss_mean)
```
titu1994, line 417: Revert, same comment as above.

titu1994 (changes requested):
Overall looks good. The multi methods should not use self.log, it is called internally.

titu1994 (approved):
LGTM

PR #1887, opened 2021-03-11
removing loading of the trained model from the example to prevent mul…

vladgets:
…ti gpu bug

Signed-off-by: Vlad Getselevich <vgetselevich@nvidia.com>


ericharper (approved):
Thanks!

PR #1895, opened 2021-03-11
Clean up ASR augmentor config docs

redoctopus:
Removed redundant parameter table

### DIFF
--- a/docs/source/asr/configs.rst
+++ b/docs/source/asr/configs.rst
@@ -88,28 +88,8 @@ Augmentation Configurations
 There are a few on-the-fly spectrogram augmentation options for NeMo ASR, which can be specified by the
 configuration file using a ``spec_augment`` section.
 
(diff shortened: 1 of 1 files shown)


titu1994 (approved):

blisc:
It's up to you if you want to merge it or not. My concern is that we are lacking good docstrings and are trying to make up for it by adding it in the documentation

titu1994:
LGTM merge when ready

titu1994:
Ideally the API documentation for this module should incorporate the above information. So just adding it to the docstring will add it to the API

redoctopus:
> It's up to you if you want to merge it or not. My concern is that we are lacking good docstrings and are trying to make up for it by adding it in the documentation

I think this one is fine to remove from the docs; everything here is already covered in the `SpectrogramAugmentation` docstring so we don't lose anything by deleting it from this page.
(https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/asr/api.html#nemo.collections.asr.modules.SpectrogramAugmentation)

PR #1898, opened 2021-03-11
Updates to conformer.

VahidooX:
+ Made positional embedding more efficient.
+ Updated the rel_shift() method to avoid flipping which needs significant memory.
+ Fixed the onnx issue.
+ Added att_context_size feature to limit the context of attention with masking.


VahidooX:
@borisfom would you please checkout the asr_model.py? 
I needed to make some changes to the export part. We did not pass the length to the encoder and it was fine for convolutional modules which do not use that input. But Conformer would use that input and export would not work on this model without passing lengths.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -122,6 +123,7 @@ def __init__(
         xscaling=True,
         untie_biases=True,
         conv_kernel_size=31,
+        conv_norm_type="batch_norm",
```
titu1994, line 126: For consistency with jasperblock, maybe just use `batch` - `layer` etc

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -203,7 +207,9 @@ def __init__(
             self._feat_out = d_model
 
     @typecheck()
-    def forward(self, audio_signal, length):
+    def forward(self, audio_signal, length=None):
+        if length is None:
```
titu1994, line 211: Do we ever expect length to be None?

nemo/collections/asr/parts/conformer_modules.py
```diff
@@ -38,13 +38,14 @@ def __init__(
...
-        self_attention_model,
-        n_heads,
-        dropout,
-        dropout_att,
-        pos_bias_u,
-        pos_bias_v,
+        self_attention_model='rel_pos',
+        n_heads=4,
+        conv_kernel_size=31,
+        conv_norm_type="batch_norm",
```
titu1994, line 44: Same as above - maybe `batch` for consistency

nemo/collections/asr/parts/conformer_modules.py
```diff
@@ -69,7 +70,7 @@ def __init__(
         elif self_attention_model == 'abs_pos':
             self.self_attn = MultiHeadAttention(n_head=n_heads, n_feat=d_model, dropout_rate=dropout_att)
         else:
-            raise ValueError(f"Not valid self_attention_model: '{self_attention_model}'!")
+            raise ValueError(f"Not a valid value for 'self_attention_model': {self_attention_model}!")
```
titu1994, line 73: Also list out the actual valid values so users dont need to dig into docs

nemo/collections/asr/parts/conformer_modules.py
```diff
@@ -124,7 +125,7 @@ class ConformerConvolution(nn.Module):
         kernel_size (int): kernel size for depthwise convolution
     """
 
-    def __init__(self, d_model, kernel_size):
+    def __init__(self, d_model, kernel_size, norm_type="batch_norm"):
```
titu1994, line 128: Same comment as above

nemo/collections/asr/parts/conformer_modules.py
```diff
@@ -141,7 +142,13 @@ def __init__(self, d_model, kernel_size):
             groups=d_model,
             bias=True,
         )
-        self.batch_norm = nn.BatchNorm1d(d_model)
+        if norm_type == "batch_norm":
+            self.batch_norm = nn.BatchNorm1d(d_model)
+        elif norm_type == "group_norm":
+            self.batch_norm = nn.GroupNorm(num_groups=1, num_channels=d_model)
```
titu1994, line 148: Same comments as above. Btw - why is it "group_norm" when in fact this is the "layer" norm settings with num_groups=1?

nemo/collections/asr/parts/conformer_modules.py
```diff
@@ -141,7 +142,13 @@ def __init__(self, d_model, kernel_size):
...
+        else:
+            raise ValueError(f"Not a valid conv_norm_type: {norm_type}!")
```
titu1994, line 150: + show valid values list

nemo/collections/asr/parts/multi_head_attention.py
```diff
@@ -163,12 +163,13 @@ def rel_shift(self, x):
...
-        x = x.view(x.size(0), x.size(1), -1)
-        x = torch.nn.functional.pad(x, pad=(0, qlen))
-        x = x.view(x.size(0), x.size(1), qlen, pos_len + 1)
-        return x[:, :, :, 0:qlen].flip(dims=[-1])
+        x_size = x.size()
+        qlen = x_size[-2]
+        pos_len = x_size[-1]
+        x = torch.nn.functional.pad(x, pad=(1, 0, 0, 0))
+        x = x.view(*x.size()[:-2], pos_len + 1, qlen)
+        x = x[:, :, 1:].view(*x_size)
```
titu1994, line 171: Document what these lines are doing, too hard to keep track of the tensor shapes mentally here.

nemo/collections/asr/parts/multi_head_attention.py
```diff
@@ -306,12 +315,17 @@ def forward(self, x):
...
+
         if self.xscale:
             x = x * self.xscale
 
-        start_pos = (self.pe.size(1) + 1) // 2 - x.size(1)
-        pos_emb = self.pe[:, start_pos:-start_pos]
+        center_pos = self.pe.size(1) // 2
+        start_pos = center_pos - x.size(1) + 1
+        end_pos = center_pos + x.size(1)
+        pos_emb = self.pe[:, start_pos:end_pos]
```
titu1994, line 328: Document what these steps are doing, too hard to mentally keep track of tensor shapes here.

titu1994 (changes requested):
Overall looks good, little bit improved docs is required.

nemo/collections/asr/parts/conformer_modules.py
VahidooX: done.

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: Actually it can be None for Jasper/QuartzNet as they do not use this in the current implementation. That is why the export method was working without passing the length. Take a look here: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/modules/conv_asr.py#L189 . We have None as default.

If user does not pass the length, we would not mask-out anything assuming all inputs are speech. It can be just one sample or a batch. It is a optional feature. With the fix in this PR on the export, I can drop it, but good to have I think.

nemo/collections/asr/parts/conformer_modules.py
VahidooX: dropped the norm_type as it was experimental.

nemo/collections/asr/parts/multi_head_attention.py
VahidooX: Done.

nemo/collections/asr/parts/multi_head_attention.py
VahidooX: It is a little hard to explain why they are done,  but added some comments.

nemo/collections/asr/parts/conformer_modules.py
VahidooX: dropped conv_nomr_type.

nemo/collections/asr/parts/conformer_modules.py
VahidooX: dropped it. group_norm with 1 group is eq to layer_norm when both last two dimensions are passed to the layer, not just the last layer (like the ones used in other layers). group_norm here normalizes over both the time and feature.

nemo/collections/asr/parts/conformer_modules.py
VahidooX: dropped the feature.

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: dropped the feature.

nemo/collections/asr/models/asr_model.py
```diff
@@ -79,8 +79,12 @@ def input_module(self):
     def output_module(self):
         return self.decoder
 
-    def forward_for_export(self, input):
-        return self.output_module(self.input_module(input))
+    def forward_for_export(self, input, length=None):
+        encoder_output = self.input_module(input, length)
+        if isinstance(encoder_output, tuple):
```
borisfom, line 84: I believe, that would be more consistent if you modify forward() of the new encoder to return a single value if called with Len=None ? Or do you actually have a case where forward() does not receive length, but should return it anyway ?

VahidooX: Actually in the new models like Conformer, we have sub-sampling which reduces the lengths, so we need to always return the updated lengths. In case of None for length, we assume the whole input is valid and produce the lengths automatically.  Due to the sub-sampling, the lengths get reduces, so need to return it. Older models did not change the lengths inside the encoder.

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Such an interesting issue.. Cool fix. Yes, lets keep it.

nemo/collections/asr/parts/multi_head_attention.py
titu1994: Cool thanks !

titu1994 (approved):
LGTM. If @borisfom gives the go ahead, merge.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -117,10 +118,11 @@ def __init__(
         subsampling_conv_channels=-1,
         ff_expansion_factor=4,
         self_attention_model='rel_pos',
-        pos_emb_max_len=5000,
         n_heads=4,
+        att_context_size=[-1, -1],
```
titu1994, line 122: Dont create lists inside of functions, use None and then check if None and set `att_context_size=[-1, -1]`

titu1994 (changes requested):
Minor patch, python has issues with lists inside functions

nemo/collections/asr/models/asr_model.py
borisfom: OK! The change is good then.

borisfom (approved):

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: Done. Thanks!

titu1994 (approved):
LGTM !

PR #1903, opened 2021-03-12
Speedup librispeech dataset processing by using multiple processes.

mousebaiker:
It is 2021, we should use parallel computations for slow scripts when possible.

As each directory contains independent data, we can easily add more processes to get_librispeech_data script to decrease waiting time.

On my machine the speedup when using 4 processes was 7-8 fold.
The whole download process went from >2h waiting time to 15 minutes.

The downloaded data is identical, so there should not be any problems introduced by parallelizing the script.


scripts/dataset_processing/get_librispeech_data.py
```diff
@@ -93,27 +129,13 @@ def __process_data(data_folder: str, dst_folder: str, manifest_file: str):
...
-                entry = {}
-                entry['audio_filepath'] = os.path.abspath(wav_file)
-                entry['duration'] = float(duration)
-                entry['text'] = transcript_text
-                entries.append(entry)
+            files.append(os.path.join(root, filename))
+
+    with multiprocessing.Pool(num_workers) as p:
+        processing_func = functools.partial(__process_transcript, dst_folder=dst_folder)
+        results = p.imap_unordered(processing_func, files)
```
titu1994, line 136: This is unordered map - changes order of validation data files which will make it hard to process data (equivalent to shuffling validation and text data). Use joblib.Parallel or mp.map() for ordered results.

titu1994 (changes requested):

scripts/dataset_processing/get_librispeech_data.py
mousebaiker: Thanks for the catch about the order.

I decided to got with imap. It guarantees order and I checked on my machine that the order is preserved. 
Compared to joblib, it is much easier to couple tqdm with it. And having progress bar is very informative for these long scripts.

mousebaiker:
PTAL. I addressed the ordering issue.

titu1994 (approved):

titu1994:
LGTM will merge after tests pass

PR #1906, opened 2021-03-12
Update RC1 to have joint installation of PT and PTL

titu1994:


tkornuta-nvidia (approved):
wow, there is `b` as well? in TLT PT/NeMo docker image  we got:
```
>>> torch.__version__
'1.8.0a0+17f8c32'
```
It is fascinating how much chaos a "simple internal torch release caused... ;)

DEKHTIARJonathan:
Unless you have a good reason to do this, I will have to oppose this one.

This PR will have significant consequences:
- You can ship as a wheel anymore
- Or the dependency of the wheel will depends on what is installed on the machine that is building the package => And it's terrible for reproducible environment.
- Wheel and SDist will behave differently.

So no, no and no.

Unless you guys make the active choice to actively only ship SDist, knowing that you break the standard as the python ecosystem tries to rely more & more on wheels not sdist.

DEKHTIARJonathan (changes requested):
@tkornuta-nvidia @okuchaiev @titu1994

Unless you have a good reason to do this, I will have to oppose this one.

This PR will have significant consequences:

You can not ship as a wheel anymore
Or the dependency of the wheel will depends on what is installed on the machine that is building the package => And it's terrible for reproducible environment.
Wheel and SDist will behave differently.
So no, no and no.

Unless you guys make the active choice to actively only ship SDist, knowing that you break the standard as the python ecosystem tries to rely more & more on wheels not sdist.

titu1994:
@DEKHTIARJonathan The code here is installed via reinstall.sh - which does pip install .[all]. 

It's still following all wheel semantics. The issue is that the pytorch container 21.02 is masquerading as a Pytorch 1.8.0a* - but it is in fact NOT pytorch 1.8. 

This confuses Pytorch Lightning which has code switching for deprecated torchtext functionality. 

The pip release is explicitly handled by @okuchaiev - he can remove this patch. But for the moment, none of the RC1 branch code will run or can be tested on pytorch 1.8 since 21.03 container has not been released yet.

DEKHTIARJonathan:
I still oppose this PR. Code relevant to your testing environment should not pollute the core of the package. This approach is just bad practices.

okuchaiev (approved):
Thanks @titu1994  and @DEKHTIARJonathan. 

I agree this is not an ideal solution. However, this isn't about our testing environment. In the order of priority, the "pip install nemo_toolkit==r1.0.0rc1" should:
1. work outside NV container
2. inside NV 20.11 container

Looks like this PR is the a way to achieve both (1) and (2). 
As soon as 21.03 is out we will do a new dot release without this workaround.

PR #1907, opened 2021-03-12
Minor update for speech regression model

fayejf:
The second part of https://github.com/NVIDIA/NeMo/pull/1777

1. Change self.log(tensorboard_logs) and remove the 'log' part in the return requested for the regression model
2. Fix tarred dataloader for speech regression support


titu1994 (approved):
LGTM

PR #1916, opened 2021-03-16
SentencePiece support for NMT

aklife97:
This PR adds support for training (as well as providing pre-trained) SentencePiece tokenizer to the **MTEncDecModel**.

~~Minor changes in SentencePiece wrapper to retrieve _unk_id_ and remove _special_tokens_ at the time of "ID to Text" conversion.~~ Changes to the SentencePiece wrapper have been moved to a new, more generic PR #1973.

### SentencePiece for NMT Usage ### 
To use sentencepiece with MTEncDecModel, just set:

`model.encoder_tokenizer.library=sentencepiece` and
  `model.decoder_tokenizer.library=sentencepiece `

A pre-trained SentencePiece model can be provided using the:

`model.encoder_tokenizer.tokenizer_model` and `model.decoder_tokenizer.tokenizer_model` tags.

In case no pre-trained SentencePiece model is provided, it will be trained based on the configurations provided. 

The SentencePiece tokenizer can also be used for only the encoder while using a different tokenizer for the decoder, and vice versa. Similar mix-n-matches can also be done for training a tokenizer for encoder/decoder only (while providing a pre-trained version of the other) etc.


titu1994 (commented):
The sentence piece change is fine by me but seems like a possibility of breaking change for backward compatibility. @yzhang123 for final say

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -503,14 +514,23 @@ def train_yttm_tokenizers(
...
         decoder_tokenizer_coverage,
         global_rank,
+        encoder_training_sample_size=-1,
+        decoder_training_sample_size=-1,
     ):
         encoder_tokenizer_model = None
         decoder_tokenizer_model = None
         os.makedirs(out_dir, exist_ok=True)
 
+        supported_tokenizers = ['yttm', 'sentencepiece']
```
ericharper, line 524: Maybe call this `supported_shared_tokenizers`?

nemo/collections/nlp/modules/common/tokenizer_utils.py
```diff
@@ -113,18 +113,29 @@ def get_nmt_tokenizer(
...
+            "pad_token": "<pad>",
+            "bos_token": "<s>",
+            "eos_token": "</s>",
+        }
+        special_tokens_dict.update(special_nmt)
+        return nemo.collections.common.tokenizers.sentencepiece_tokenizer.SentencePieceTokenizer(
+            model_path=tokenizer_model, special_tokens=special_tokens_dict
+        )
     else:
         raise NotImplementedError('Currently we only support "yttm" and "huggingface" tokenizer library.')
```
ericharper, line 141: Can add sentencepiece here.

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
```diff
@@ -105,17 +106,11 @@ def tokens_to_text(self, tokens):
...
-        for i, id in enumerate(ids):
-            if id in self.id_to_special_token:
-                text += self.tokenizer.decode_ids(ids[last_i:i]) + " "
-                text += self.id_to_special_token[id] + " "
-                last_i = i + 1
-
-        text += self.tokenizer.decode_ids(ids[last_i:])
-        return text.strip()
+        ids_ = [id_ for id_ in ids if id_ not in self.id_to_special_token]
+        return self.tokenizer.decode_ids(ids_)
```
ericharper, line 113: Does this return the same thing?

ericharper (commented):
Overall looks really good to me. Nice PR. Like @titu1994  said, we should make sure any existing models using sentencepiece aren't breaking with the `ids_to_text` change. Shouldn't be bad as that would be used for validation and inference only I'm guessing.

I only found two usages in NeMo:
https://github.com/NVIDIA/NeMo/blob/39e2e4a86ec76fb96ac79e47b523e89b6bd3e7a0/Jenkinsfile#L778
https://github.com/NVIDIA/NeMo/blob/39e2e4a86ec76fb96ac79e47b523e89b6bd3e7a0/nemo/collections/asr/parts/mixins.py#L68

ericharper:
Also please add example usage to the README.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -549,15 +584,28 @@ def train_yttm_tokenizers(
...
+                                model=encoder_tokenizer_model,
+                                coverage=encoder_tokenizer_coverage,
+                                n_threads=-1,
+                            )
+                        else:
+                            dir_name = os.path.dirname(encoder_tokenizer_model)
+                            create_spt_model(
+                                data_file=src_fname,
+                                vocab_size=encoder_tokenizer_vocab_size,
+                                sample_size=encoder_training_sample_size,
```
MaximumEntropy, line 600: Is the reason for specifying a sample_size here that sentencepiece cannot be trained on very large corpora?

aklife97: Yes. The variable though can be set to -1 (the default value) to fetch the entire data.

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
aklife97: The update still returns decoded text, just with special tokens removed. This makes it consistent with the YTTM and word tokenizers. In general, the way we're handling special tokens in sentencepiece wrapper is kind of dicey (as it is dependent on the order of the passed tokens), and I believe we might want to take some other route for it.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
aklife97: I have changed it to `supported_train_tokenizers` - also maintains consistency with the method name (`train_tokenizers`).

titu1994 (changes requested):
Blocking because this is a breaking change that impacts ASRBPEMixin - which is basically every CTCBPE and RNNTBPE model. 

Use a new flag that switches between the two modes, defaulting to the old one. NNT will just have to specify this flag always.

titu1994:
After discussion, default can be to drop special tokens, and asr can just use a constructor level flag to default to old behaviour as needed. 
Unblocked, address other reviewers and then ping me

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
yzhang123: is this also consistent with autotokenizer? does that also remove special tokens?

aklife97: Yes, autotokenizer removes the special tokens too.

yzhang123 (approved):

yzhang123 (approved):

yzhang123 (approved):

titu1994 (approved):
SPE Changes look fine now

PR #1919, opened 2021-03-16
adding bert pretraining docs

yzhang123:


docs/source/nlp/bert_pretraining.rst
```diff
@@ -0,0 +1,104 @@
...
+| model.mask_prob                           | float           | Probability of masking a token in the input text during data processing                                      |
++-------------------------------------------+-----------------+--------------------------------------------------------------------------------------------------------------+
+| model.short_seq_prob                      | float           | Probability of having a sequence shorter than the maximum sequence length                                    |
++-------------------------------------------+-----------------+--------------------------------------------------------------------------------------------------------------+
+
+.. note::
+
+    For offline data preprocessing **model.tokenizer** is null.
+    For online data preprocessing **model.tokenizer** needs to be specified.
+
```
ekmb, line 83: maybe specify what tokenizer is going to be used and what tokenizer/vocab users would need when using this model on downstream tasks

How about adding a line here on how to use this pre-trained bert model + tokenizer for downstream tasks and maybe cross-reference in the https://github.com/NVIDIA/NeMo/blob/r1.0.0rc1/docs/source/nlp/nlp_model.rst

ekmb (changes requested):
Please add a quick start section on how to load a pre-trained model once they're available

docs/source/nlp/bert_pretraining.rst
yzhang123: done

yzhang123:
@ekmb  also updated PretrainedModelInfo for BERT and QAModel, also added quick start guide with available models for both to docs

ekmb (approved):

PR #1924, opened 2021-03-16
NMT revert change to yaml cfg

ericharper:
NMT config was accidentally updated with some wrong values for train_ds. Reverting those.


okuchaiev (approved):

MaximumEntropy (approved):
LGTM

PR #1945, opened 2021-03-19
Add ASR Evaluation Service

titu1994:
# Changelog
- Adds Flask based ASR Service with support for gunicorn and docker container
- Enables inference via pretrained checkpoint to evaluate models on audio data that is formatted to 16 KHz and mono-channel wav files.
- Enables upload of trained checkpoints for evaluation of custom models (can be disabled inside `templates/main.html`)

# Usage
## Flask Server

```bash
$ python asr_service.py
```

## Gunicorn Server
By default, the gunicorn server config is loaded from `gunicorn.conf.py`. Command line values will override these arguments.

**NOTE**: Limit the number of workers since each worker will maintain the cache of models in memory.

```bash
# wsgi.py is a thin wrapper over `asr_service.py` to enable the service without debugging mode
$ gunicorn wsgi:app
```

## Docker Server
The app can be built and run easily using docker. 

Build the docker container using -
```bash
$ bash docker_container_build.sh
```

Run the docker container (in detached mode with all available GPUs) using - 
```bash
$ bash docker_container_run.sh
```


VahidooX:
> 
> 
> This pull request **introduces 2 alerts** when merging [82fa2be](https://github.com/NVIDIA/NeMo/commit/82fa2be0844042c845af3cf43e7e5e2cf989eab0) into [da48c99](https://github.com/NVIDIA/NeMo/commit/da48c997177f89a9e305e508ab27b15954044814) - [view on LGTM.com](https://lgtm.com/projects/g/NVIDIA/NeMo/rev/pr-87eabc38d29520a4b3fc1bcfa066c69c4774062f)
> 
> **new alerts:**
> 
>     * 1 for Unnecessary pass
> 
>     * 1 for Unused import

Some LGTM warnings to address.

examples/asr/asr_webapp/README.rst
```diff
@@ -0,0 +1,70 @@
...
+Usage
+-----
+
+A simple service that will load a pre-trained model from NeMo's ASR collection and after uploading some files, will transcribe them. Any pre-trained model can be selected and if the local / docker environment supports GPUs then it will be possible to transcribe the audio segments with GPU.
+
+All uploaded files are immediately deleted after transcriptions are obtained.
+
+ .. note::
+
+    When using gunicorn, you might notice that each pretrained checkpoint takes a long time for the first transcription. This is because that worker is instantiating the model into memory and then moving the model onto the GPU. For large models, this step might take a significant amount of time. However, this step is cached, and subsequent transcription requests should be much faster (especially if a GPU is utilized).
```
VahidooX, line 13: Gunicorn.
Have a link or refs here?

examples/asr/asr_webapp/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+There are three options to run the ASR Service to transcribe audio -
+
+Local Setup
+-----------
+
+1) Install all the dependencies (``pip install -r requirements.txt``). Note : It will utilize the latest branch set in the requirements and will override current nemo installation.
```
VahidooX, line 20: Where are these files located? Maybe give the path to the folder?

examples/asr/asr_webapp/README.rst
```diff
@@ -0,0 +1,70 @@
...
+The cleanest approach of the three, and requires simply building and running a docker container.
+
+1) Build the docker by executing ``bash docker_container_build.sh``. This will build a container using the latest branch of nemo.
+
+2) Run the container by executing ``bash docker_container_run.sh``. This will run a detached container that can be used by visiting ``0.0.0.0:8000`` on a modern browser.
+
+Note About Uploading Models
+---------------------------
+
+Uploading models is a useful method to evaluate models quickly without having to write scripts.
```
VahidooX, line 49: Should we clarify that by all users, we mean the users who have access to that specific server?
Some people with limited knowledge may think it would be seen by everybody.

examples/asr/asr_webapp/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+There are several reasons this approach was taken -
+
+* It allows for easy evaluation amongst peers without duplicating model uploads (which may be very big).
+* When using ``gunicorn``, every worker has access to a central cache of models which can be loaded into memory.
+* Model upload mode is meant to be used in a local, protected environment.
+
+If you wish to disable model uploading, please open ``templates/main.html`` and comment out the following section:
+
+.. code-block:: html
```
VahidooX, line 63: Can it be a variable on the top of the HTML file so that it can be done by just updating that?

examples/asr/asr_webapp/docker_container_build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
VahidooX, line 1: Isn't it easier to put the building and running docker commands in the docs directly? Both ways are actually OK.

examples/asr/asr_webapp/model_api.py
```diff
@@ -0,0 +1,131 @@
...
+    else:
+        model = initialize_model(model_name)
+
+    if torch.cuda.is_available() and use_gpu_if_available:
+        model = model.cuda()
+
+    # transcribe audio
+    logging.info("Begin transcribing audio...")
+    try:
+        with autocast():
```
VahidooX, line 100: Does it always use AMP? Is it useful to make it an option or it does not matter?

examples/asr/asr_webapp/README.rst
```diff
@@ -0,0 +1,70 @@
+**ASR SERVICE **
+================
+
+Usage
+-----
+
+A simple service that will load a pre-trained model from NeMo's ASR collection and after uploading some files, will transcribe them. Any pre-trained model can be selected and if the local / docker environment supports GPUs then it will be possible to transcribe the audio segments with GPU.
```
VahidooX, line 7: They can also upload and perform inference on their own pre-trained model, yes?

VahidooX (approved):

examples/asr/asr_webapp/README.rst
titu1994: They are local to the service directory. Good point, ill note it somewhere.

examples/asr/asr_webapp/README.rst
titu1994: It is technically seen by everybody .. say the webserver is on a public URL then everyone who can access that public url can upload models.

examples/asr/asr_webapp/README.rst
titu1994: It will need more JS than I have knowledge about. For now, manual deletion is preferred.

examples/asr/asr_webapp/docker_container_build.sh
titu1994: In docs means it cant be executed. Rather keep it a runnable bash script.

examples/asr/asr_webapp/model_api.py
titu1994: AMP is used only if torch amp is possible - it defaults to False for all hardware that doesnt support mixed precision. If it exists, it should be safe to use for inference - it preserves memory as well as improves speed of transcription.

examples/asr/asr_webapp/README.rst
titu1994: Ah good catch

PR #1965, opened 2021-03-25
Correct hyper parameter saving for WandB logs

titu1994:
# Changelog
- Preserve updated logs inside parameter used by wandb and tensorboard.


okuchaiev (approved):
lgtm

PR #1967, opened 2021-03-25
NMT add Multiprocessing to Data Preprocessing

ericharper:
Tokenizing data and saving it to a tarred dataset is an embarrassingly parallel task.
We parallelize the data processing by partioning the src and tgt text lines into chunks. 
Each chunk is then written to a tarred dataset in parallel using the joblib library.

Also note that tarred datasets are to be used when the dataset is too large to be handled within system memory so
we made sure that our approach was also memory friendly. The memory consumed while multiprocessing depends 
only on the number of jobs and the chunk size (n_jobs * chunk_size).

# Results
We tested the end-to-end time of automatic data processing for NMT (do_training=False) with multiprocessing for two language pairs. We used a chunk size of 1M lines.

En-De used a parallel training set with 4524868 lines (~4.5M):
```
n_preproc_jobs: 4, elapsed seconds: 122
n_preproc_jobs: 2, elapsed seconds: 207
n_preproc_jobs: 1, elapsed seconds: 373
```
En-Es used a parallel training set with 65630789 lines (~65M):
```
n_preproc_jobs: 64, elapsed seconds: 514
n_preproc_jobs: 32, elapsed seconds: 602
n_preproc_jobs: 16, elapsed seconds: 837
n_preproc_jobs: 8, elapsed seconds: 1326
n_preproc_jobs: 1, elapsed seconds: 8168
```
As we can see, for larger datasets the speedup can save a lot of time! ~15x speed up or 2.2 hours vs 8.5 minutes.
Since we have to preprocess our dataset for different tokenizers, vocabulary sizes, and batch sizes, it's process that we end up doing quite a lot when experimenting and it can be very useful to save a couple of hours per experiment.

# Usage
Multi-processing will now happen automatically. By default NMT will attempt to use every process available (except for 2):
```
train_ds.n_preproc_jobs=-2
```
Not every process will necessarily be used as there has to be at least one chunk per process. To configure multi-processing manually simply set:
```
train_ds.n_preproc_jobs=4
```
# TODO
- [x] Testing for backwards compatibility and convergence.
- [x] General clean up
- [x] Add jenkins tests
- [x] Update script version


nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -366,6 +348,91 @@ def preprocess_parallel_dataset(
...
+        return i + 1
+
+    @staticmethod
+    def _get_lines_partition(num_lines, lines_per_dataset_fragment):
+        # create partition based on fragment size
+        fragment_indices = []
+        for i in range(0, num_lines, lines_per_dataset_fragment):
+            fragment_indices.append([i, i + lines_per_dataset_fragment])
+        # remove last indices if needed
+        if fragment_indices[-1][1] >= num_lines:
```
MaximumEntropy, line 365: For large values of lines_per_dataset_fragment, this will potentially result in a lot of lines lost right? I think we should still process this fragment and just throw away batches in `_process_fragment()`

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -250,107 +272,67 @@ def preprocess_parallel_dataset(
...
+                    for fragment_index, lines_indices in enumerate(lines_partition)
+                )
+                total_batches = sum(total_batches_list)
 
-                json.dump({'num_batches': global_batch_ctr}, open(metadata_path, 'w'))
+                # dump metadata to json
+                metadata = {}
+                metadata['num_batches'] = total_batches
+                tar_file_paths = glob.glob(f'{out_dir}/*.tar')
+                metadata['tar_files'] = tar_file_paths
```
MaximumEntropy, line 332: Adding this path to the metadata may make moving an already created tarred dataset around tricky. Since you disallow `cfg.tar_files` and metadata['tar_files']. How do I override what's in the metadata to use what I provide via hydra?

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -366,6 +348,91 @@ def preprocess_parallel_dataset(
...
+    ):
+        start = lines_indices[0]
+        stop = lines_indices[1]
+
+        # write lines in partition to temporary files to be consumed by write_parallel_batches_to_tarfiles
+        tmp_f_src = tempfile.NamedTemporaryFile(delete=False, mode='w')
+        tmp_f_tgt = tempfile.NamedTemporaryFile(delete=False, mode='w')
+
+        with open(src_filename, 'r') as src_in, open(tgt_filename) as tgt_in:
+            src_line = next(src_in)
```
MaximumEntropy, line 398: Why do you call next on the file before enumerate? Won't that discard the first line before you start enumerating the file?

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -366,6 +348,91 @@ def preprocess_parallel_dataset(
...
+            tgt_line = next(tgt_in)
+            for line_number, lines in enumerate(zip(src_in, tgt_in)):
```
MaximumEntropy, line 400: Silly nitpick but you could do `for line_number, (src_line, tgt_line) in enumerate(zip(src_in, tgt_in)):` to save two lines below

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -250,107 +272,67 @@ def preprocess_parallel_dataset(
...
+                        encoder_bpe_dropout=encoder_bpe_dropout,
+                        encoder_model_name=encoder_model_name,
+                        decoder_tokenizer_name=decoder_tokenizer_name,
+                        decoder_tokenizer_model=decoder_tokenizer_model,
+                        decoder_bpe_dropout=decoder_bpe_dropout,
+                        decoder_model_name=decoder_model_name,
+                        pkl_file_prefix=fragment_index,
                     )
-                    global_batch_ctr -= num_files_in_tar
-                    logging.info('Dropping %d batches because of overflow' % (num_files_in_tar))
```
MaximumEntropy, line 349: You need to move this dropping logic of batches into `_process_fragment()` otherwise tarfiles create from a particular fragment may be imbalanced.

MaximumEntropy (changes requested):

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: You can still specify your tarred dataset via hydra. But now you can also specify it via a manifest file.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: Good point, I can make that change.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: I believe that gives you the first line. I can double check.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: I like that! Will do.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: I drop batches here: https://github.com/NVIDIA/NeMo/blob/146952ca71c98f2274c49f2472a080f786157e45/nemo/collections/nlp/data/machine_translation/preproc_mt_data.py#L729

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: Done.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: Done.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: Done.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
ericharper: Updated the batch dropping logic. Now we only discard at most num_batches_per_tarfile

Jenkinsfile
```diff
@@ -994,6 +994,57 @@ pipeline {
...
+    stage('L2: NMT Tarred Dataset Creation') {
+      when {
+        anyOf{
+          branch 'main'
+          changeRequest target: 'main'
+        }
+      }
+      failFast true
+      parallel {
+        stage('L2: NMT Auto Tarred Dataset Creation') {
```
MaximumEntropy, line 1006: Should we want to set `do_training=False` for this test?

examples/nlp/machine_translation/create_tarred_parallel_dataset.py
```diff
@@ -116,14 +123,21 @@
...
+        decoder_model_name=args.decoder_model_name,
+        decoder_tokenizer_model=decoder_tokenizer_model,
+        decoder_bpe_dropout=args.decoder_tokenizer_bpe_dropout,
         max_seq_length=args.max_seq_length,
         min_seq_length=args.min_seq_length,
         tokens_in_batch=args.tokens_in_batch,
         lines_per_dataset_fragment=args.lines_per_dataset_fragment,
         num_batches_per_tarfile=args.num_batches_per_tarfile,
-        pkl_file_prefix=args.pkl_file_prefix,
+        tar_file_prefix=args.tar_file_prefix,
```
MaximumEntropy, line 139: Should we add this variable to the yaml config as well?

MaximumEntropy (approved):
LGTM. Potentially a couple of minor changes only.

Jenkinsfile
ericharper: Done.

examples/nlp/machine_translation/create_tarred_parallel_dataset.py
ericharper: Done.

PR #1969, opened 2021-03-25
Update tarred dataset preprocessing script to use joblib

titu1994:
# Changelog
- Update ASR tarred dataset preprocessing script to enable more than 1 workers via joblib
- Maintaining backward compatibility with exact same results even when `workers` > 1 as compared to `workers` == 1.

# Caveats
- The new manifest itself will temporarily require double the memory in ram now until all of the tarfiles are processed as we need to flatten the intermediate representation of each create tarfile output manifest


redoctopus (approved):
LGTM

PR #1973, opened 2021-03-26
SentencePiece changes + UnitTests + Making tokenizer APIs consistent

aklife97:
### Summary of the PR ###
Broadly, this PR does the following:
* Changes `SentencePiece` wrapper and `create_spt_model` function to handle special tokens inside SentencePiece model (instead of inside the wrapper that happens currently).
* Makes all TokenizerSpec tokenizer wrappers (AutoTokenizer, CharTokenizer, WordTokenizer, YTTM, SentecePiece) APIs consistent - except for a few differences as discussed later.
* Adds tests for the current SentencePiece and YTTM tokenizers.

### SentencePiece Changes ###
Current implementation of the SentencePiece wrapper handles special tokens inside the wrapper using a special token dictionary - rather than inside the tokenizer model. This makes the special tokens created non-persistent and volatile to change during different runs. 

To better handle this situation, the `create_spt_model` function now supports the addition of `bos`, `eos`, and `pad` tokens directly and special tokens can only be added during train time. More custom special tokens can be passed as `Control symbols` or `User defined symbols` as defined by sentencepiece. `bos`, `eos`, and `pad` are by default `custom symbols`. The major difference between setting a special token as `control` vs `user defined` is that `user defined symbols` can encoded and decoded directly from the text, whereas `control symbols` can only be added manually.  For more information on the differnce, refer [SentencePiece](https://github.com/google/sentencepiece/blob/master/doc/special_symbols.md).  Despite these changes, any previously coded calls to `create_spt_model` would work in the **_same fashion and do not need to be changed_**. 

The SentencePiece wrapper has also been changed to accommodate the above. While the default behavior of the SentencePiece wrapper has been changed with this, older behavior for previously trained models can be brought back by setting `legacy = True` while instantiating the wrapper.

### Making TokenizerSpec tokenizer APIs consistent ###

This PR also tries to make all the tokenizer wrappers that inherit from TokenizerSpec consistent. It changes the YTTM tokenizer's `ids_to_tokens` function to make it consistent with others. Again, while this has been set as the default behavior, previous behavior can be restored by setting `legacy = True`.

While the tokenizers are now mostly similar, there are some exceptions:
* AutoTokenizer differs from all other tokenizers in the fact that the AutoTokenizer's `tokens_to_text` function _**does not**_ remove special tokens. This is probably a bug, and should ideally be addressed when it's certain that it won't break anything. 
* YTTM and SentecePiece tokenizers do not encode special tokens (control symbols for SentencePiece) when `text_to_ids` or `text_to_tokens` is called. While this should ideally be consistent across all tokenizers, as this inhibits user text to change the programmatic meaning of the text, other tokenizers (char, word, and auto) encode the special tokens. An example:
When trying to tokenize (`text_to_tokens`)
`text: a b [UNK] c` 
yttm and sentencepice: `['a', 'b', '[', 'U', 'N', 'K', ']', 'c']`
others: `['a', 'b', '[UNK]', 'c']`

#### Behavior of TokenizerSpec API ####
* `[tokens|ids]_to_[ids|tokens]`: Converts tokens to ids and vice versa while preserving all special token infomation, i.e. `len(ids) == len(tokens)`
* `text_to_[tokens|ids]`: Encodes text as tokens/ids. Exception: Auto/word/char encode special tokens in text as is.
* `[tokens|id]_to_text`:  Converts tokens/ids to text and removes special tokens (only control symbols for SentencePiece). Exception: `tokens_to_text` of `AutoTokenizer`.

### SentencePiece and YTTM tests ###
Added tests for testing YTTM and SentencePiece.


nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
```diff
@@ -206,6 +260,11 @@ def create_spt_model(
             to build the tokenizer.
         max_sentencepiece_length: Limits the maximum length of the SentencePiece subword that can be constructed.
             By default, no limit is placed.
+        bos: when True, bos token "<s>" is added to the vocabulary.
+        eos: when True, eos token "</s>" is added to the vocabulary.
+        pad: when True, pad token "<pad>" is added to the vocabulary.
+        control_symbols: control symbols to add to tokenizer, as defined by sentencepiece.
```
titu1994, line 266: Document difference between `control_symbols` and `user_defined_symbols` - maybe with example.

titu1994 (commented):
SentencePiece changes look fine. @yzhang123  for final approval

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
aklife97: Added explanation for both in the docstring.

ericharper:
Doesn't have to be this PR, but I think the contents of the README would be a great addition to the NeMo NLP Docs.

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
```diff
@@ -30,7 +31,9 @@ class SentencePieceTokenizer(TokenizerSpec):
     Sentencepiecetokenizer https://github.com/google/sentencepiece.
     '''
 
-    def __init__(self, model_path: str, special_tokens: Optional[Union[Dict[str, str], List[str]]] = None):
+    def __init__(
+        self, model_path: str, special_tokens: Optional[Union[Dict[str, str], List[str]]] = None, legacy: bool = False
+    ):
         """
         Args:
             model_path: path to sentence piece tokenizer model. To create the model use create_spt_model()
```
okuchaiev, line 39: can you please add "legacy" to docstring. Say something brief.

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
```diff
@@ -42,84 +45,105 @@ def __init__(self, model_path: str, special_tokens: Optional[Union[Dict[str, str
         self.tokenizer.Load(model_path)
         self.original_vocab_size = self.tokenizer.get_piece_size()
         self.vocab_size = self.tokenizer.get_piece_size()
+        self.legacy = legacy
         self.special_token_to_id = {}
         self.id_to_special_token = {}
         if special_tokens:
+            if not self.legacy:
+                raise ValueError(
+                    "Special tokens must be None when legacy is set to False. Provide special tokens at train time, or set legacy=True."
```
okuchaiev, line 54: let's remove the suggestion: ", or set legacy=True."

okuchaiev (approved):
looks good to me. few small comments

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
aklife97: Addressed both comments. Waiting for @yzhang123 's review. Will merge once approved.

nemo/collections/nlp/modules/common/tokenizer_utils.py
```diff
@@ -82,7 +82,7 @@ def get_tokenizer(
 
     if tokenizer_name == 'sentencepiece':
         return nemo.collections.common.tokenizers.sentencepiece_tokenizer.SentencePieceTokenizer(
-            model_path=tokenizer_model, special_tokens=special_tokens
+            model_path=tokenizer_model, special_tokens=special_tokens, legacy=True
```
yzhang123, line 85: could you add docstring for bpe_dropout?

yzhang123 (changes requested):

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
yzhang123: could you move docstring outside of constructor for automatic doc generation?

aklife97: Moved the args docstring outside. Is it fine now?

nemo/collections/nlp/modules/common/tokenizer_utils.py
aklife97: Added.

yzhang123 (approved):

yzhang123 (approved):

PR #1978, opened 2021-03-27
Adding the feature of prelaoding weights a checkpoint.

VahidooX:
It enables user to start training by loading weights from another checkpoint ('.ckpt') or '.nemo' file. 

It can be done by setting the 'model.init_weights_from_model' parameter in the config or by:
model.load_weights_from_checkpoint(init_model_path='/path/to/model.nemo', strict=True)

It is a feature which enables fine-tuning out of the box for all nemo model/scripts/examples by just using the config.

### DIFF
--- a/docs/source/starthere/intro.rst
+++ b/docs/source/starthere/intro.rst
@@ -72,7 +72,7 @@ Below we is the code snippet of Audio Translator application.
     # Speech Recognition model - QuartzNet trained on Russian part of MCV 6.0
     quartznet = nemo_asr.models.EncDecCTCModel.from_pretrained(model_name="stt_ru_quartznet15x5").cuda()
     # Neural Machine Translation model
-    nmt_model = nemo_nlp.models.MTEncDecModel.from_pretrained(model_name='nmt_ru_en_transformer6x6', strict=False).cuda()
+    nmt_model = nemo_nlp.models.MTEncDecModel.from_pretrained(model_name='nmt_ru_en_transformer6x6').cuda()
     # Spectrogram generator which takes text as an input and produces spectrogram
     spectrogram_generator = nemo_tts.models.Tacotron2Model.from_pretrained(model_name="tts_en_tacotron2").cuda()
     # Vocoder model which takes spectrogram and produces actual audio


nemo/core/classes/modelPT.py
```diff
@@ -1246,3 +1245,22 @@ def set_eff_save(use_eff_save: bool):
     def use_eff_save() -> bool:
         global _MODEL_EFF_SAVE
         return _MODEL_EFF_SAVE
+
+    def load_weights_from_checkpoint(self, model_path, strict=True):
```
titu1994, line 1249: Rename this function - checkpoint implies `PTL .ckpt` but this function takes both `ckpt` as well as `.nemo`. Also, ckpt can be confused for Pytorch .pt - which is not supported.

This method has a fatal flaw - it tries to restore model using `self` - but current context of `self` is ModelPT. So while the method exists in ModelPT, trying `ModelPT.load_weights_from_checkpoint(...)` will fail since `self` doesnt resolve to the actual class that needs to be loaded up. I will send a PR to generalize this (at least for .nemo files). However this will fail for PTL checkpoints.

I think we should not allow PTL checkpoints at all in such a function, only .nemo files. For PTL files, experiment manager has its `resume*` flags and only those should be the supported path to avoid duplicating complexity. @okuchaiev does that sound fine? We mostly release pretrained .nemo files only anyway.

nemo/core/classes/modelPT.py
```diff
@@ -1246,3 +1245,22 @@ def set_eff_save(use_eff_save: bool):
...
+        if model_path.endswith(".nemo"):
+            checkpoint = self.__class__.restore_from(model_path, map_location=torch.device('cpu'))
+            self.load_state_dict(checkpoint.state_dict(), strict=strict)
```
titu1994, line 1252: `del checkpoint` after this

nemo/core/classes/modelPT.py
```diff
@@ -1246,3 +1245,22 @@ def set_eff_save(use_eff_save: bool):
...
+        elif model_path.endswith(".ckpt"):
+            checkpoint = torch.load(model_path, map_location=torch.device('cpu'))
+            self.load_state_dict(checkpoint["state_dict"], strict=strict)
```
titu1994, line 1255: `del checkpoint` after this

nemo/core/classes/modelPT.py
```diff
@@ -1246,3 +1245,22 @@ def set_eff_save(use_eff_save: bool):
...
+        else:
+            raise ValueError(
+                "'{model_path}' is not a valid file as we need either a NeMo file ('.nemo') or "
+                "a PyTorch checkpoint ('.ckpt') to load weights from."
+            )
+
+    def on_pretrain_routine_start(self):
```
titu1994, line 1262: Remove this option to init model from config defined path. Too many edge cases for no benefit when its better left to be done manually using the above function by the user - in either case its a one liner.

One of the most problematic cases is that the checkpoint path is not removed from the config - so when the .nemo model gets saved, it still has this path inside it. When someone restores and tries to finetune or train from scratch, even if they dont specify this flag, the old value will still be there and weights loading will fail (since on different machine, different paths).

Another issue is that this can cause an issue with model parallel models. For now, lets revert this function entirely.

nemo/core/classes/modelPT.py
titu1994: Add docstrings to this function.

titu1994 (changes requested):
Needs some changes - mainly we already have experiment manager to handle fine tuning with .ckpt checkpoints. So this should only solve the .nemo case and nothing else.

The user can add this function call outside the trainer.train() call so that model weight loading is done safely in non-ddp env. No need for config management which adds more problems than it solves. 

For .nemo files, this PR is currently insufficient. It needs to depend on a PR ill send in some time to generalize the `restore_from` implementation to ModelPT itself. 

Other than that, some docstrings and its good to go.

jarivk:
Can we get this merged?

nemo/core/classes/modelPT.py
VahidooX: + I have tested these functions with both nemo and PTL checkpoints. They work fine. Python has dynamic binding so when we call that methosd it calls the right method based on the real type of the object.

+ Current exp_manager does not support this feature currently. This feature is different from what we have currently. The one we have is to continue training which loads both the weights and optimizer state, etc. But this feature just loads the weights which is used for fine-tuning. Fine-tuning is different from continuing previous runs as we should not just load the weights and nothing else. I think currently the exp_manager can not be used for fine-tuning without adding some lines to the speech_to_text_bpe.py. Do you currently fine-tune with speech_to_text.py and exp_manager?

nemo/core/classes/modelPT.py
VahidooX: done

nemo/core/classes/modelPT.py
VahidooX: done

nemo/core/classes/modelPT.py
VahidooX: The whole point of this PR is to enable initializing the weights by using the config file which makes the current speec_to_text.py script able to fine-tune without any need to add some extra lines. If I remove it, the whole PR becomes mostly absolute as users can always loads weights with two lines of code.

We can drop it from the config to address such cases. It does not look to me a not solvable issue. But even in this state, it does not make serious issue. No one can train from scratch or fine-tune with a nemo file as it is already trained and step is not zero anymore. It just tried to load the weights if global step is zero. The only way for such a case is load the nemo model, recreate the optimizer, data loader and all stuff and etc. We have the same issue with tokenizer, data loaders, load_from_checkpoint too.

nemo/core/classes/modelPT.py
VahidooX: + Currently I think PTL handles resume_from_checkpoint with the same approach. The only difference is that they load the whole state including config and optimizer state while this feature loads just weights.

+ We can add this feature of just loading the weights to the current load models/checkpoints method. It is fine. Then we can call them. But that is not the main goal of this PR. This PR is to enable all the examples in the NeMo to support fine-tuning without the need of the user to add extra lines to the examples. It is not just for text_to_speech. It is a general feature.  Currently all NeMo users who do fine-tuning need to add extra lines (it they have enough knowledge) to the examples to perform fine-tuning.

VahidooX: + added @ericharper as we are making changes to the core.

titu1994: .nemo files are meant to be used for finetuning - you are not supposed to load up a model seperately, then load state dict for this model from yet another .nemo / PTL checkpoint. First - you should not even be using .ptl checkpoint for finetuning, since as you mentioned it contains the optimizer state which is incorrect for finetuning!

If this is meant for just finetuning, then this function should drop support for PTL entirely since it is incorrect.

This PR adds yet another approach to fine tune - which is by loading one model via config, restoring some/all of its weights from another checkpoint (this can be .nemo / ptl .ckpt but not pt ckpt), and then continue training. I dont see how this is a good practice, but there are cases for it I guess. @okuchaiev needs to decide if we want to support this. Only exp manager should work on .ptl checkpoints.

To note, we will now have 4 different ways of handling checkpoints - and that doesnt make sense to me.
1) restore_from for .nemo  (for finetuning or eval)
2) from pretrained for .nemo (for finetuning or eval)
3) load_state_dict - for PT / PTL checkpoints  (for finetuning or eval)
4) and now load_state_from_checkpoints which is for both .nemo and ptl checkpoints (but not PT checkpoints) - (for finetuning or eval).

titu1994: This PR also doesnt change the fact that users will still need to add these lines to the example scripts - so other than hiding 3-4 lines inside ModelPT - it has no major benefit. All of the example scripts will need to update to use this function anyway.

nemo/core/classes/modelPT.py
titu1994: > No one can train from scratch or fine-tune with a nemo file as it is already trained and step is not zero anymore

.nemo files are the **primary** way to finetune or scratch train - there are multiple examples all throughout notebooks of exactly this scenario. The step is exactly 0 for finetuning - since the PTL optimizer state is not a part of .nemo files!

And yes, this should most definitely be removed from config because it will break fine tuning after restoration (reasons mentioned in the previous message)

ericharper:
I'm not sure what we're trying to accomplish with this PR. As @titu1994 has said, we can already restore weights and finetune in NeMo with .nemo files. Or to resume training, we use PTL.

Can you please make it more clear in the README what feature is being added and how to use it?

Also, this PR is adding a new feature and should not be going into RC1. RC1 PRs are only for bug fixes and documentation.

Please rebase to main.

ericharper:
I'm not sure what we're trying to accomplish with this PR. As @titu1994 has said, we can already restore weights and finetune in NeMo with .nemo files. Or to resume training, we use PTL.

Can you please make it more clear in the README what feature is being added and how to use it?

Also, this PR is adding a new feature and should not be going into RC1. RC1 PRs are only for bug fixes and documentation.

Please rebase to main.

PR #1984, opened 2021-03-30
NLP Update to Model Parallel

ericharper:
PTL significantly changed their API which broke our existing model parallel support.

This PR updates `NLPModel` so that model parallel training and auto-resume works again.

# TODO:

- Add model parallel support to .nemo files.


ericharper:
Close in favor of #2015

PR #1986, opened 2021-03-30
Fix formula to compute same padding for Jasper-based models.

mousebaiker:
Notation: 
k - kernel size
p - padding
d - dilation

The old formula is only correct when d <= 3. 

This can be verified by choosing d = 4, k = 3 and computing the value of padding with an old formula _p_old = (d * k) // 2 - 1 = 5._ Whereas the correct answer is _(d * (k - 1)) // 2 = 4_.
This result can be verified in [convolution visualizer](https://ezyang.github.io/convolution-visualizer/index.html).

The change **does not affect** a case when **d=1 and k is odd**. In that case, (k - 1) // 2 = k // 2. Also **d=2** and **d=3** are unaffected due to coincidental correct answers that old formula gave.

The behaviour **changes for even kernels** but there's no correct symmetrical padding in that case, so we don't use it in practice.


titu1994 (commented):
LGTM, as long as dilation < 3 this formula converges to old formula.

PR #1992, opened 2021-03-30
Make numba dynamically checked and remove hard dependency in requirements.txt

titu1994:
# Changelog
- Remove hard numba version requirements
- Dynamically check for version at runtime when it is required
- Update tests to dynamically switch off for cuda if version is incompatible (either due to numba version itself, or due to mismatch between supported CUDA version and actual CUDA version installed).
- Update docker file to forcibly update numba temporarily (to be removed at 20.04 release)
- Update docker file to preemptively import the domains to download whatever is necessary.
- Patch a bug in dynamic resolution of numba losses

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN apt-get update && \
 
 # build torchaudio (change latest release version to match pytorch)
 WORKDIR /tmp/torchaudio_build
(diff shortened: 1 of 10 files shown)


okuchaiev (approved):
lgtm, thx

PR #1995, opened 2021-03-31
Prepare RNNT to switch to Numba loss for compatibility

titu1994:
# Changelog
- Add rnnt loss switching capability.
- Enable passing arbitrary kwargs to the inner loss function.
- Add jenkins tests (commented out until 21.04 container is released which supports minimum numba version 0.53 for CUDA 11.2)
- Enable all RNNT model tests using the numba default loss

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -291,6 +291,7 @@ pipeline {
       }
     }
 
+//  TODO: UNCOMMENT TESTS AFTER 21.04 release (numba 0.53 min requirement)
 //     stage('L2: ASR RNNT dev run') {
 //       when {
 //         anyOf {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -291,6 +291,44 @@ pipeline {
       }
     }
 
+//     stage('L2: ASR RNNT dev run') {
+//       when {
+//         anyOf {
+//           branch 'main'
+//           changeRequest target: 'main'
+//         }
+//       }
+//       failFast true
+//       parallel {
+//         stage('Speech to Text - RNNT') {
+//           steps {
+//             sh 'python examples/asr/speech_to_text_rnnt.py \
+//             model.train_ds.manifest_filepath=/home/TestData/an4_dataset/an4_train.json \
+//             model.validation_ds.manifest_filepath=/home/TestData/an4_dataset/an4_val.json \
+//             model.train_ds.batch_size=8 \
+//             trainer.gpus=[0] \
+//             +trainer.fast_dev_run=True \
+//             exp_manager.exp_dir=examples/asr/speech_to_text_rnnt_results'
+//             sh 'rm -rf examples/asr/speech_to_text_rnnt_results'
+//           }
+//         }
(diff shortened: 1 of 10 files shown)


nemo/collections/asr/models/rnnt_models.py
```diff
@@ -130,6 +136,44 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             self._optim_variational_noise_std = 0
             self._optim_variational_noise_start = 0
 
+    def _setup_loss(self, cfg: Optional[DictConfig]):
```
VahidooX, line 139: name is a little misleading. How about extract_rnnt_cfg?

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -130,6 +136,44 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+
+                Note that whichever loss_name is selected, that corresponding kwargs will be
+                selected. For the "default" case, the "default_kwargs" will be used, not the
+                "{resolved_default}_kwargs".
+
+        Examples:
+            .. code-block:: yaml
+                loss_name: "default"
+
+                default_kwargs:
```
VahidooX, line 158: What happens if later we change the default rnnt loss later?
is it necessary to have this default?

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -130,6 +136,44 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+                    kwargs2: some_other_val
+
+        Returns:
+            A tuple, the resolved loss name as well as its kwargs (if found).
+        """
+        if cfg is None:
+            cfg = DictConfig({})
+
+        loss_name = cfg.get("loss_name", "default")
+        loss_kwargs = cfg.get(f"{loss_name}_kwargs", None)
```
VahidooX, line 171: Do we need to have the name of the loss in the param?
Is there a way we use target as the class name of the loss and let hydra create the appropriate loss?

Jenkinsfile
```diff
@@ -286,6 +286,44 @@ pipeline {
       }
     }
 
+//     stage('L2: ASR RNNT dev run') {
```
VahidooX, line 289: How about adding a comment why this section is commented out and when it comes back?

VahidooX (changes requested):

Jenkinsfile
titu1994: Its commented out until 21.04 is released, which has Numba 0.53 installed in it. I think well just send a PR next month once we get that container.

nemo/collections/asr/models/rnnt_models.py
titu1994: Done

Jenkinsfile
titu1994: Done

VahidooX (approved):

PR #1998, opened 2021-04-01
Speed up training of Citrinet models by more efficient masking.

mousebaiker:
The proposed change includes three optimization steps that I found were beneficial for model training efficiency.

### 1. Mask caching
Currently `arange` for mask is computed at each step from scratch. While it might be cached by GPU, there is no guarantee as it is not needed after forward function execution. 
By caching it dynamically, it will be easier to strike balance between numerous short sequences that will benefit from caching and arbitrarily long sequences during inference.
 

### 2. Multiplication by mask instead of `masked_fill`
`masked_fill `supports arbitrary values to fill into the tensor as, thus, implemented as an iterator over Tensor memory.
In our models masks are binary, so we can leverage vectorized multiplication operation to speedup mask application.

### 3. Shortcutting new sequence length computation
Given that Citrinet uses same padding everywhere and that striding only happens at certain levels. We can avoid doing arithmetic with sequence lengths when not needed.
Additionally, NeMo DatasetLoaders guarantee that lengths are represented as longs, so we can remove redundant `to` conversion.

### NGC
Time per epoch, 384 filters model, 8 GPUs, average across 10 epochs:

| baseline  |  +Mask caching  | +Multiplication   | +Shortcuts  |
|---|---|---|---|
|50 m   | 46.5 m   | 45 m   | 44.5m   |

 
Time per epoch, 384 filters model, 4 GPUs, aggregate grads from 2 batches, average across 34 epochs:

| baseline  |  +Mask caching  | +Multiplication   | +Shortcuts  |
|---|---|---|---|
|78 m   | 71 m   | 69 m   | 70m   |

**Accuracy** for all improvement runs **is in bounds of statistical noise** compared to the baseline.


mousebaiker:
I am still waiting for the full run to complete in order to confirm that there is no degradation in quality.  For now, keeping in draft state.

titu1994 (commented):
LGTM, @blisc for final review
A little comment block explaining that self.lens reflects the "current" maximum length of the sequence and is updated as a new "max_len" is found would be good.

nemo/collections/asr/parts/jasper.py
```diff
@@ -178,19 +178,34 @@ def __init__(
...
-            lens = lens.to(dtype=torch.long)
             max_len = x.size(2)
-            mask = torch.arange(max_len).to(lens.device).expand(len(lens), max_len) >= lens.unsqueeze(1)
-            x = x.masked_fill(mask.unsqueeze(1).to(device=x.device), 0)
-            # del mask
+            if max_len > self.max_len:
+                self.lens = torch.arange(max_len)
+                self.max_len = max_len
+
+            self.lens = self.lens.to(lens.device)
```
blisc, line 206: Why not move this instead the `if` statement?

nemo/collections/asr/parts/jasper.py
```diff
@@ -178,19 +178,34 @@ def __init__(
...
+            mask = self.lens[:max_len].unsqueeze(0) < lens.unsqueeze(1)
+            x = x * mask.unsqueeze(1).to(device=x.device)
```
blisc, line 208: The `to(x.device)` is probably redundant

blisc (approved):

nemo/collections/asr/parts/jasper.py
mousebaiker: `lens` input can come from different device during inference. This will cause mismatch if the inference utterance is shorter than `self.max_len`.

nemo/collections/asr/parts/jasper.py
mousebaiker: In the unlikely case of `x` and `lens` being on different devices, the forward pass will fail without this `to`. So let's err on the side of safety here.

mousebaiker:
Thanks a lot everybody for the review!

PR #2001, opened 2021-04-01
Actually install NeMo requirements in Docker.

mousebaiker:
When building docker container locally: 
```
#17 [nemo-deps 10/12] RUN for f in $(ls requirements/*.txt); do pip install ...
#17 0.515 ls: cannot access 'requirements/*.txt': No such file or directory
#17 DONE 0.6s

```

The files are copied not into a separate `/tmp/nemo/requirements/` folder but directly into `/tmp/nemo/`, so we either have to create new directory first or install directly from where we copied.


okuchaiev (approved):
lgtm

PR #2007, opened 2021-04-01
Update container version in Jenkinsfile

ericharper:


okuchaiev (approved):
lgtm

Dockerfile
```diff
@@ -89,7 +89,7 @@ COPY . .
 
 # start building the final container
 FROM nemo-deps as nemo
-ARG NEMO_VERSION=1.0.0rc1
+ARG NEMO_VERSION=main
```
titu1994, line 92: This should revert to 1.0.0rc1

okuchaiev: why though? 1.0.0rc1 was never tested with 21.03 container?

Dockerfile
```diff
@@ -68,7 +77,7 @@ RUN COMMIT_SHA=f546575109111c455354861a0567c8aa794208a2 && \
     rm -rf ../tests test ../tensorflow_binding
 
 # uninstall stuff from base container
-RUN pip uninstall -y sacrebleu
+RUN pip uninstall -y sacrebleu torchtext
```
titu1994, line 80: Wait, torchtext is being deleted here again, this is unecessary now that we built from source.

titu1994 (commented):
LGTM

Dockerfile
ericharper: I moved it earlier now.

titu1994:
LGTM

PR #2015, opened 2021-04-02
[NLP] Updating model parallel 

ericharper:
Updating model parallel to work with new PTL DDP API and adding .nemo support.

# Usage
We support 4 main workflows: 
- Training from megatron-lm checkpoint
- Auto-resume training from experiment manager
- Training from model parallel .nemo file
- Evaluation from model parallel .nemo file

To enable these workflows in a NeMo NLP example training script, the PyTorch Lightning `Trainer` must be instantiated with the `NLPDDPPlugin`. This plugin contains the necessary logic to instantiate model parallel Megatron-LM models and train them.

```python
trainer = Trainer(plugins=[NLPDDPPlugin()], **cfg.trainer)
```
Since model parallel models always require multiple GPUs, we must always have a `Trainer` object.
This means that when we restore a model parallel .nemo file, we must also pass in the `Trainer`.
```python
model = TextClassificationModel.restore_from(cfg.model.nemo_path, trainer=trainer)
```

## Training from Megatron-LM checkpoint
Megatron-LM checkpoints have the following specific format:
```
iter_0080000/
├── mp_rank_00
│   └── model_optim_rng.pt
└── mp_rank_01
    └── model_optim_rng.pt
```
In order to start training from these checkpoints, pass in the parent directory and model configuration file:
```
model.language_model.lm_checkpoint=/raid/megatron/bert/iter_0080000 \
model.language_model.config_file=/raid/data/megatron/bert/config.json \
```
The model configuration can be specified in YAML:
```yaml
model:
  language_model:
    config:
        hidden_size: 1024
        num_attention_heads: 16
        num_layers: 24
        max_position_embeddings: 512
```
Additionally, the vocab file must be specified for megatron-lm models:
```
model.tokenizer.vocab_file=/path/to/vocab.txt
```
or if using the Megatron-LM default tokenizer for training BERT:
```
model.tokenizer.tokenizer_name=megatron-bert-uncased
```

## Auto-resume from experiment manager
Auto resume works exactly the same as non-model parallel models. 
While training with PTL, model parallel checkpoint will be saved and loaded properly.
```
checkpoints/
├── mp_rank_00
│   ├── mp_autoresume-last.ckpt
│   ├── mp_autoresume---val_loss=0.35-epoch=0.ckpt
│   ├── mp_autoresume---val_loss=0.38-epoch=1.ckpt
│   └── mp_autoresume---val_loss=0.39-epoch=2.ckpt
└── mp_rank_01
    ├── mp_autoresume-last.ckpt
    ├── mp_autoresume---val_loss=0.35-epoch=0.ckpt
    ├── mp_autoresume---val_loss=0.38-epoch=1.ckpt
    └── mp_autoresume---val_loss=0.39-epoch=2.ckpt
```
## Training from model parallel .nemo
Model parallel checkpoints are saved in the correct format:
```
text_class_350m
├── megatron-bert-uncased_encoder_config.json
├── megatron_checkpoint_version.json
├── model_config.yaml
├── mp_rank_00
│   └── model_weights.ckpt
├── mp_rank_01
│   └── model_weights.ckpt
├── tokenizer_vocab_dict.json
└── tokenizer.vocab_file
```
Training from the .nemo is the same as other NeMo NLP example scripts:
```python
    trainer = pl.Trainer(plugins=[NLPDDPPlugin()], **cfg.trainer)
    model = TokenClassificationModel.restore_from(cfg.pretrained_model, trainer=trainer)
    trainer.fit(model)
```

## Evaluation from model parallel .nemo
Current NeMo NLP example scripts often assume using a single GPU for evaluation, which will not work.
However, evaluation with model parallel .nemo is very simple:
```python
    trainer = pl.Trainer(plugins=[NLPDDPPlugin()], **cfg.trainer)
    exp_manager(trainer, cfg.get("exp_manager", None))

    model = TextClassificationModel.restore_from(cfg.model.nemo_path, trainer=trainer)
    model.setup_test_data(test_data_config=cfg.model.test_ds)

    trainer.test(model=model, ckpt_path=None)
```


examples/nlp/text_classification/model_parallel_text_classification_evaluation.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
okuchaiev, line 1: it's 2021 ;)

examples/nlp/text_classification/text_classification_with_bert.py
```diff
@@ -107,7 +108,7 @@
 @hydra_runner(config_path="conf", config_name="text_classification_config")
 def main(cfg: DictConfig) -> None:
     logging.info(f'\nConfig Params:\n{cfg.pretty()}')
-    trainer = pl.Trainer(**cfg.trainer)
+    trainer = pl.Trainer(plugins=[NLPDDPPlugin()], **cfg.trainer)
```
okuchaiev, line 111: Is this script for model parallel only? Or, if it is for both, will NLPDDPPlugin() affect things for non-model parallel case?

examples/nlp/token_classification/token_classification_train.py
```diff
@@ -101,15 +102,15 @@
 
 @hydra_runner(config_path="conf", config_name="token_classification_config")
 def main(cfg: DictConfig) -> None:
-    trainer = pl.Trainer(**cfg.trainer)
+    trainer = pl.Trainer(plugins=[NLPDDPPlugin()], **cfg.trainer)
```
okuchaiev, line 105: same question here - how does this interact with non-model-parallel

examples/nlp/token_classification/token_classification_train.py
```diff
@@ -101,15 +102,15 @@
...
 
     if not cfg.pretrained_model:
         logging.info(f'Config: {OmegaConf.to_yaml(cfg)}')
         model = TokenClassificationModel(cfg.model, trainer=trainer)
     else:
         if os.path.exists(cfg.pretrained_model):
-            model = TokenClassificationModel.restore_from(cfg.pretrained_model)
+            model = TokenClassificationModel.restore_from(cfg.pretrained_model, trainer=trainer)
         elif cfg.pretrained_model in TokenClassificationModel.get_available_model_names():
             model = TokenClassificationModel.from_pretrained(cfg.pretrained_model)
```
okuchaiev, line 115: so this case is guaranteed to be non-model parallel, right?

nemo/collections/nlp/modules/common/megatron/megatron_utils.py
```diff
@@ -144,10 +144,13 @@ def get_megatron_lm_model(
...
-    if os.path.isdir(checkpoint_file):
-        app_state = AppState()
+    app_state = AppState()
+    if app_state.model_parallel_size is not None and app_state.model_parallel_rank is not None:
+        # model parallel already known from .nemo restore
+        model_parallel_size = app_state.model_parallel_size
+        model_parallel_rank = app_state.model_parallel_rank
+    elif os.path.isdir(checkpoint_file):
+        # starting training from megatron-lm checkpoint
         model_parallel_size = len(os.listdir(checkpoint_file))
```
okuchaiev, line 154: I am wondering if something else can appear in this folder rather than checkpoints per rank (for example, some config)? Maybe do some file extension check?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -410,3 +522,160 @@ def save_checkpoint(self, filepath, weights_only: bool):
                     )
                     atomic_save(checkpoint, filepath)
         return None
+
+
+class NLPDDPPlugin(DDPPlugin):
```
okuchaiev, line 527: Consider moving this out of nlp_model.py into ints own file.

nemo/utils/app_state.py
```diff
@@ -45,6 +45,12 @@ def __init__(self):
         self._create_checkpoint_callback = None
         self._checkpoint_callback_params = None
 
+        # Save and Restore (.nemo)
+        self._tmpdir_name = None
+        self._model_config_yaml = "model_config.yaml"
```
okuchaiev, line 50: This logic assumes that there is a single model per application.
We currently don't have other use-cases but in future that assumption will break when we will need to add teacher-student training - in those cases it is likely to have 2 models per application.

okuchaiev (commented):
Is it possible to not make an assumption (within AppState) that we'll have 1 application == 1 NeMo model?

examples/nlp/text_classification/text_classification_with_bert.py
ericharper: Non-model parallel case works as normal with NLPDDPPLugin.

examples/nlp/token_classification/token_classification_train.py
ericharper: Non-model parallel case works as normal with NLPDDPPLugin.

examples/nlp/token_classification/token_classification_train.py
ericharper: No this can be model parallel or non-model parallel.

nemo/collections/nlp/modules/common/megatron/megatron_utils.py
ericharper: Yes, I can check for `mp_rank`.

nemo/collections/nlp/models/nlp_model.py
ericharper: I think that makes sense. I'll have to figure out how to organize it.

nemo/utils/app_state.py
ericharper: Good point. I think we can generalize it to multiple models when we add teacher-student.

examples/nlp/token_classification/token_classification_train.py
okuchaiev: I meant elif part where trainer instance isn't passed:
```python
elif cfg.pretrained_model in TokenClassificationModel.get_available_model_names():
            model = TokenClassificationModel.from_pretrained(cfg.pretrained_model)
```

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
-                    raise NotImplementedError(
-                        f'The BERT encoder: {self.bert_model} does not support model parallelism yet.'
-                    )
-
     def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
         if hasattr(self, "bert_model") and isinstance(self.bert_model, MegatronBertEncoder):
             checkpoint['checkpoint_version'] = get_checkpoint_version()
         return None
 
+    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
```
titu1994, line 260: Is this a PTL override? Can you add a comment as to its origin - internal to nemo or overriden from nemo/ptl.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+        if hasattr(self, "bert_model") and isinstance(self.bert_model, MegatronBertEncoder):
+            set_checkpoint_version(checkpoint['checkpoint_version'])
+            logging.info(f"Setting Megatron checkpoint version: {checkpoint['checkpoint_version']}")
+        return None
+
+    # no rank check as model parallel models need to be saved on data parallel rank 0
```
titu1994, line 266: Its true that model parallel models dont need to be saved on rank 0, *non* model parallel models would suffer issues we had before - all nodes try to save to same location and it corrupts the file. Its fine to remove the decorator but have an equivalent such that when 1) its a non-model parallel model 2) its any rank != 0 ;; immediately return and dont try to save anything.

titu1994: Or is this implicitly handled by `super().save_to()`? In that case, disregard above comment.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+        if app_state.model_parallel_size is not None:
+            # each model parallel rank creates a .nemo file
+            # after all .nemo files are created, each rank
+            # will add their checkpoint to global rank 0
+
+            base_dir = os.path.dirname(save_path)  # use the directory to merge mp_rank .nemo files into one
+
+            # TODO: accept general paths
+            # update save_path based on model parallel_rank
+            base_path = save_path[0:-5]  # everything excpe the .nemo extension
```
titu1994, line 297: Lets correct this before merging. os.path.splitext(X)[0] should work here ?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+            mp_save_path = f'{base_path}_mp_rank_{app_state.model_parallel_rank:02d}.nemo'
+
+            if app_state.data_parallel_rank == 0:
+                super()._default_save_to(mp_save_path)
+
+            # barrier so that all processes have finished writing their weights before creating .nemo file
+            torch.distributed.barrier()
+
+            if is_global_rank_zero():
+                # TODO: do in a temporary directory
```
titu1994, line 308: I think (given the huge size of the models + the time it would take to 1) write to temp dir 2) copy from temp dir to required dir), this can be done in place. Upto you. Its cleaner to do in a temp dir, but with model parallel, we are talking about non trivial model checkpoint write times.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+                By default (None), it will select a GPU if available, falling back to CPU otherwise.
+            strict: Passed to load_state_dict.
+            return_config: If set to true, will return just the underlying config of the restored
+                model as an OmegaConf DictConfig object without instantiating the model.
+            trainer: Must be passed in to use model parallel .nemo
+
+            Example:
+                ```
+                model = nemo.collections.asr.models.EncDecCTCModel.restore_from('asr.nemo')
+                assert isinstance(model, nemo.collections.asr.models.EncDecCTCModel)
```
titu1994, line 373: Should probably reflect NLPModels?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+
+        app_state = AppState()
+        app_state.model_restore_path = os.path.abspath(os.path.expanduser(restore_path))
+
+        # detect if we have a model parallel .nemo file
+        with tempfile.TemporaryDirectory() as tmpdir:
+            cwd = os.getcwd()
+            os.chdir(tmpdir)
+            cls.__unpack_nemo_file(path2file=restore_path, out_folder=tmpdir)
+            mp_ranks = glob.glob(os.path.join(tmpdir, 'mp_rank*'))
```
titu1994, line 390: Isnt this unpack check super expensive for say non-model parallel - large BERT models? No other choice?

titu1994: Can this be moved below so that its done only if `app_state.model_parallel_size is not None`?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+                else:
+                    checkpoint_version_dict = {'checkpoint_version': checkpoint_version}
+                    checkpoint_version_path = 'megatron_checkpoint_version.json'
+                    checkpoint_version_src = os.path.join(NEMO_NLP_TMP, checkpoint_version_path)
+                    with open(checkpoint_version_src, 'w') as f:
+                        f.write(json.dumps(checkpoint_version_dict))
+                    self.register_artifact(checkpoint_version_path, checkpoint_version_src)
+
+    @staticmethod
+    def __unpack_nemo_file(path2file: str, out_folder: str) -> str:
```
titu1994, line 461: isnt this same as ModelPT impl? Rather replace the `__unpack_nemo_file` with `_unpack_nemo_file` and call super?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+        else:
+            # get encoder config and create source for artifact
+            if isinstance(self.bert_model, MegatronBertEncoder):
+                checkpoint_version = get_checkpoint_version()
+                if checkpoint_version is None:
+                    raise ValueError('Unable to get megatron checkpoint version.')
+                else:
+                    checkpoint_version_dict = {'checkpoint_version': checkpoint_version}
+                    checkpoint_version_path = 'megatron_checkpoint_version.json'
+                    checkpoint_version_src = os.path.join(NEMO_NLP_TMP, checkpoint_version_path)
```
titu1994, line 455: This does not easily scale to future changes. What I would suggest is have a minimal dataclass that can be loaded up with these default values set for everything, so that in future if changes are necessary they can be made to the data class and then loaded up here. You can convert dataclasses to DictConfigs and then get back dict to write to json and it scales to future changes.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,53 +252,225 @@ def setup(self, stage: str) -> None:
...
+
+    @staticmethod
+    def __unpack_nemo_file(path2file: str, out_folder: str) -> str:
+        if not os.path.exists(path2file):
+            raise FileNotFoundError(f"{path2file} does not exist")
+        tar = tarfile.open(path2file, "r:gz")
+        tar.extractall(path=out_folder)
+        tar.close()
+        return out_folder
+
```
titu1994, line 468: Same here, this is nearly identical to ModelPT so make it `_` instead of `__` and call super?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -410,3 +522,160 @@ def save_checkpoint(self, filepath, weights_only: bool):
...
+    distributed_backend = "ddp"
+
+    def __init__(
+        self,
+        parallel_devices: Optional[List[torch.device]] = None,
+        num_nodes: int = 1,
+        cluster_environment: ClusterEnvironment = None,
+        sync_batchnorm: bool = False,
+        **kwargs: Union[Any, Dict[str, Any]],
+    ) -> None:
```
titu1994, line 540: Can you add `self.__doc__ = super().__doc__` just to copy the dicstring from the original impl?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -410,3 +522,160 @@ def save_checkpoint(self, filepath, weights_only: bool):
...
+        if isinstance(self.lightning_module.bert_model, MegatronBertEncoder):
+            app_state = AppState()
+            if app_state.model_parallel_size is not None:
+                # mpu grad clipping needs parameters to have the attribute model_parallel
+                parameters = self.lightning_module.parameters()
+                for p in parameters:
+                    if not hasattr(p, 'model_parallel'):
+                        p.model_parallel = False
+
+                # TODO: figure out how to override clip gradients again
```
titu1994, line 569: Is it fine to train Megatron without it? If not, maybe setup a warning here for the timebeing for megatron models.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -410,3 +522,160 @@ def save_checkpoint(self, filepath, weights_only: bool):
...
+                    logging.info(
+                        f"Restoring from pretrained model parallel checkpoint: {self.lightning_module.bert_model._restore_path}"
+                    )
+                    self.lightning_module.bert_model.restore_weights(self.lightning_module.bert_model._restore_path)
+
+            self.lightning_module.register_megatron_checkpoint_version()
+
+        return super().start_training(trainer)
+
+    def start_testing(self, trainer: 'Trainer') -> None:
```
titu1994, line 602: Just a question, theres no `start_validation`?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -410,3 +522,160 @@ def save_checkpoint(self, filepath, weights_only: bool):
...
+        app_state = AppState()
+
+        if app_state.model_parallel_size is not None:
+            logging.info(f"Configuring DDP for model parallelism.")
+
+            # With model parallelism, multiple GPUs form a large "logical GPU"
+            # this means that data parallel groups span multiple GPUs
+            # and are non-trivial
+            device_ids = self.determine_ddp_device_ids()
+            self._model = DistributedDataParallel(
```
titu1994, line 630: Im afraid future versions of PTL signature for this code will change, and it will cause issues. Theres a way to inspect signatures of classes using `inspect` ive used for comparing classes and their respective dataclasses, can a similar technique be used here to ensure that signature doesnt diverge too much ?

nemo/collections/nlp/modules/common/megatron/megatron_bert.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, model_name, config, vocab_file, model_parallel_size=None, mod
         if not os.path.exists(vocab_file):
             raise ValueError(f'Vocab file not found at {vocab_file}')
 
+        # convert config to dictionary
+        if isinstance(config, DictConfig):
```
titu1994, line 72: This is necessary? Maybe just remove its struct mode? Anyway its fine.

nemo/core/classes/modelPT.py
```diff
@@ -1236,11 +1241,17 @@ def _is_restore_type_tarfile() -> bool:
...
             return False
         else:
-            if tarfile.is_tarfile(_MODEL_RESTORE_PATH):
-                return True
+            if _MODEL_RESTORE_PATH:
+                if tarfile.is_tarfile(_MODEL_RESTORE_PATH):
+                    return True
+            elif app_state.model_restore_path:
+                if tarfile.is_tarfile(app_state.model_restore_path):
+                    return True
```
titu1994, line 1254: This `_MODEL_RESTORE_PATH` ... It was a bad idea to make it a single string. I will need to refactor it to a list of strings in the future to support more than one model being restored in a sequence, and it will affect NLPModel restore_from impl. Will request your review at that time.

nemo/utils/app_state.py
titu1994: Yes AppState in general always (rightly) assumes only 1 model is being constructed ever in entire runtime - because its dealing with global defaults. This setup here will need to change to support more than 1 model being restored.

titu1994 (commented):
Overall it looks great. Lots of things need to be documented better and minor changes here and there if possible.

nemo/utils/app_state.py
okuchaiev: So we have two options:

1. Re-design this part of this PR now.
2. Re-design this part when we start adding teacher-student

I think both options are OK.

examples/nlp/text_classification/model_parallel_text_classification_evaluation.py
ericharper: Done.

examples/nlp/token_classification/token_classification_train.py
ericharper: There's no model parallel models on NGC yet...

nemo/collections/nlp/models/nlp_model.py
ericharper: Done. Also moved NLPCheckpointConnector.

nemo/utils/app_state.py
ericharper: I'll go with 2 as this PR is already too big.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -47,6 +54,10 @@
 
 os.makedirs(NEMO_NLP_TMP, exist_ok=True)
 
+_MODEL_RESTORE_PATH = None
+_MODEL_CONFIG_YAML = "model_config.yaml"
```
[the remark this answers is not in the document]
ericharper: I couldn't figure out a way to use them from ModelPT directly. 

I want to have these kind of attributes in AppState so let me see if I can clean this up some.

nemo/collections/nlp/models/nlp_model.py
ericharper: Good catch. Comment added.

nemo/collections/nlp/models/nlp_model.py
ericharper: That's correct. `super().save_to()` has the decorator for the non-model parallel case.

nemo/collections/nlp/models/nlp_model.py
ericharper: Yes, I will fix this before we merge.

nemo/collections/nlp/models/nlp_model.py
ericharper: That is a good point, with 4-8B parameter models that will be a huge copy. Let's leave it as is for now and we can revisit it later if needed.

nemo/collections/nlp/models/nlp_model.py
ericharper: Done.

nemo/collections/nlp/models/nlp_model.py
ericharper: This is used to detect model parallel. 

The other option is to have users pass in model parallel size via an argument. I was not going this route because I wanted to make model parallelism automatic in NeMo.

nemo/collections/nlp/models/nlp_model.py
ericharper: I tried so many ways but could not make it work. Because it's a class method and a double underscore, nothing I did would work.

nemo/collections/nlp/models/nlp_model.py
ericharper: I will try that.

nemo/collections/nlp/models/nlp_model.py
ericharper: It's used in the megatron-lm gradient clipping code.

nemo/collections/nlp/models/nlp_model.py
ericharper: That's correct.

nemo/collections/nlp/models/nlp_model.py
ericharper: Done.

nemo/collections/nlp/models/nlp_model.py
ericharper: I tried doing this and get this error: 
```
AttributeError: 'TextClassificationModel' object has no attribute '_NLPModel__make_nemo_file_from_folder'
```

nemo/collections/nlp/models/nlp_model.py
ericharper: That's a fine suggestion but I'd like to move it to a later PR.

nemo/collections/nlp/models/nlp_model.py
ericharper: Another good suggestions, but can we defer to a later PR?

examples/nlp/token_classification/token_classification_train.py
titu1994: I meant for the future. Better to have it in from_pretrained at the ModelPT level and then forward the trainer to the actual restore_from call inside.

nemo/collections/nlp/models/nlp_model.py
titu1994: Ok perfect then

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
+        """
+        Saves model instance (weights and configuration) into .nemo file
+         You can use "restore_from" method to fully restore instance from .nemo file.
 
-class NLPCheckpointConnector(CheckpointConnector):
-    """ Override PTL CheckpointConnector to support model parallel checkpoints from Megatron-LM.
-    """
+        .nemo file is an archive (tar.gz) with the following:
+            model_config.yaml - model configuration in .yaml format. You can deserialize this into cfg argument for model's constructor
+            model_wights.chpt - model checkpoint
```
titu1994, line 272: Minor typo - **.ckpt**

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
+            save_path: Path to .nemo file where model instance should be saved
+        """
 
-    def save_checkpoint(self, filepath, weights_only: bool):
-        """Slightly modified version of PyTorch Lightning's save_checkpoint.
+        app_state = AppState()
+        if app_state.model_parallel_size is not None:
+            self._default_save_to(save_path)
+        else:
+            return super().save_to(save_path)
```
titu1994, line 282: Could you add a comment here saying this super()... line will execute only on rank 0?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
-            weights_only (bool): [description]
+            restore_path: path to .nemo file from which model should be instantiated
+            override_config_path: path to a yaml config that will override the internal
+                config file or an OmegaConf / DictConfig object representing the model config.
+            map_location: Optional torch.device() to map the instantiated model to a device.
+                By default (None), it will select a GPU if available, falling back to CPU otherwise.
+            strict: Passed to load_state_dict.
+            return_config: If set to true, will return just the underlying config of the restored
+                model as an OmegaConf DictConfig object without instantiating the model.
+            trainer: PyTorch Lightning trainer. Must be passed in to use model parallel .nemo
```
titu1994, line 363:
```suggestion
            trainer: PyTorch Lightning trainer. Must be passed in order to use model parallel .nemo
```

nemo/collections/nlp/models/nlp_model.py
titu1994: Hmm youre right we should automate it. We might need to think of a more efficient way - as in just directly using tarfile to open the file and parse the number of folders or something in the future. The extraction step here will be executed on every rank, therefore all ranks will waste a very long time unpacking large chunk of data just to find number of rank files inside it.

Can be done later though. Leaving the comment open as a todo reminder for future work

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
-                    )
-                    atomic_save(checkpoint, filepath)
-        return None
+            app_state.world_size = trainer.num_gpus * trainer.num_nodes
+
+            # try to get local rank from global
+            local_rank = None
+            try:
+                local_rank = int(os.environ['LOCAL_RANK'])
+            except:
```
titu1994, line 419: Assumes SLURM env? trainer.local_rank should have this info no?

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
+                logging.info('Global variable LOCAL_RANK not yet specified. Assuming LOCAL_RANK is 0.')
+
+            if local_rank is not None:
+                app_state.local_rank = local_rank
+            else:
+                # if local is None then we are on the main process
+                local_rank = 0
```
titu1994, line 426: This assumption is risky - on other envs all of the nodes will assume they are local rank 0 and load that checkpoint multiple times. Everything works but inference would give random results.

Best to simply raise a hard error here.

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -312,101 +241,238 @@ def setup(self, stage: str) -> None:
...
+            model_parallel_rank = compute_model_parallel_rank(local_rank, app_state.model_parallel_size)
+            app_state.model_parallel_rank = model_parallel_rank
+
+            restored_model = cls._default_restore_from(
+                restore_path, override_config_path, map_location, strict, return_config
+            )
+            restored_model._trainer = trainer
+            return restored_model
+        else:
+            return super().restore_from(restore_path, override_config_path, map_location, strict, return_config)
```
titu1994, line 437: Lets add a note that this super() line will execute only on rank 0?

nemo/collections/nlp/models/nlp_model.py
titu1994: Sure thats fine, keeping comment open for future work.

nemo/collections/nlp/models/nlp_model.py
titu1994: Oh I meant change the ModelPT double underscore to single underscore, and use it here. Theres no particular reason the ModelPT impl of this should be name mangled and unavailable to its subclasses.

nemo/collections/nlp/models/nlp_model.py
titu1994: Yep same idea as above, replace the double underscore in ModelPT itself with single underscore and then use single underscore version in this class. No reason this method should be name mangled and unavailable to its subclasses.

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -0,0 +1,233 @@
...
+            if app_state.model_parallel_size is not None:
+                # mpu grad clipping needs parameters to have the attribute model_parallel
+                parameters = self.lightning_module.parameters()
+                for p in parameters:
+                    if not hasattr(p, 'model_parallel'):
+                        p.model_parallel = False
+
+                # TODO: figure out how to override clip gradients again
+                # Update PTL trainer to use our _clip_gradients
+                # self._trainer.accelerator_backend._clip_gradients = self._clip_gradients
```
titu1994, line 79: Lets raise a user visible warning if they try to clip grads and its not implemented yet.

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -0,0 +1,233 @@
...
+                num_replicas=app_state.data_parallel_size, rank=app_state.data_parallel_rank
+            )
+            return distributed_sampler_kwargs
+
+        else:
+            return super().distributed_sampler_kwargs
+
+
+class NLPCheckpointConnector(CheckpointConnector):
+    """ Override PTL CheckpointConnector to support model parallel checkpoints from Megatron-LM.
```
titu1994, line 193: So this class overrides exp_manager checkpoint connecter right? Doesnt it lose a lot of functionality?

nemo/utils/app_state.py
titu1994: (2) is fine with me too.

titu1994 (commented):
Overall its ready to merge. Minor changes : 
1) Make the double underscore methods of modelPT to single underscore and reuse them in NLPModel
2) Some comment strings to show that super().restore or save is going to be executed on rank 0 only
3) Some cleanup which can be done in the next PR.

@okuchaiev for final review

nemo/collections/nlp/parts/nlp_overrides.py
ericharper: Oh you can still clip grads just won't be over model parallel groups until we add that back.

nemo/collections/nlp/parts/nlp_overrides.py
ericharper: No this is overriding PTL CheckpointConnector. exp_manager overrides ModelCheckpoint. No functionality should be loss, just the added functionality of model parallel checkpoints.

Please let me know if you see some functionality being lost.

ericharper:
@titu1994 I was able to detect model parallel without unpacking the .nemo file. Thanks for catching that.

nemo/collections/nlp/models/nlp_model.py
okuchaiev: agree with @titu1994 here, let's raise an exception here.

okuchaiev (approved):
lgtm

nemo/collections/nlp/parts/nlp_overrides.py
titu1994: Awesome, I think it should be fine if this is not overriding exp_manager (it also overrides some of PTLs connectors like this) so lets test it out on a cluster and if it works as expected, no issues.

PR #2016, opened 2021-04-02
Fix the versioning on multi-node training.

VahidooX:
Used resume_if_exists instead of checking the Slurm to void creating the version folders. So whenever resume_if_exists  is enabled, we should skip creating the versions otherwise training can not continue.


titu1994 (commented):
LGTM, @blisc for final review

nemo/utils/exp_manager.py
```diff
@@ -457,9 +460,11 @@ def get_log_dir(
...
         version = version or os.environ.get(NEMO_ENV_VARNAME_VERSION, None)
 
-        if version is None:
-            if check_slurm(trainer):
-                logging.warning("Running on a slurm cluster. exp_manager will not add a version number.")
+        if not version:
+            if resume_if_exists:
+                logging.warning(
+                    "No version folders would be created under the log folder as 'resume_if_exists' is enabled."
+                )
```
blisc, line 467: Wouldn't it be more accurate to tell users to use `explicit_log_dir` when resuming, instead of using `exp_dir`/`name`/`version `?

blisc (changes requested):

blisc (changes requested):
Please update the docstring of exp_manger so that it matches the new behaviour.
Please add a note that the behaviour of this argument changed as of version 1.0.0.

VahidooX:
> 
> 
> Please update the docstring of exp_manger so that it matches the new behaviour.
> Please add a note that the behaviour of this argument changed as of version 1.0.0.

Added to the docstring.

blisc (approved):

titu1994 (approved):
LGTM

PR #2023, opened 2021-04-06
Fix speech to text scripts.

VahidooX:
We need to use the trainer object, no the config. It forces the user to have those in the config file.
Also fixed the check_slurm method to detect the slurm.

### DIFF
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -473,7 +473,7 @@ def get_log_dir(
                 logging.warning(
                     "No version folders would be created under the log folder as 'resume_if_exists' is enabled."
                 )
-                version = ""
+                version = None
             elif is_global_rank_zero():
                 if use_datetime_version:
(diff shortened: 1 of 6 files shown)


titu1994 (approved):
Makes sense, LGTM

nemo/utils/exp_manager.py
```diff
@@ -473,7 +473,7 @@ def get_log_dir(
                 logging.warning(
                     "No version folders would be created under the log folder as 'resume_if_exists' is enabled."
                 )
-                version = ""
+                version = None
```
titu1994, line 476: Ah this should have remained "". I'll fix this in another pr

VahidooX: I was testing something. It works without error now. Let me try something before reverting it back. My suspicion is that w&b handler methods think we want to use version even when we pass "" as the version.

titu1994: Ok I'll leave that to you then, I haven't attempted the wandb fox yet.

PR #2030, opened 2021-04-07
 Compute SE in float32 for stability in AMP training.

mousebaiker:
Squeeze-and-Excitation block introduced instabilities to Citrinet when training with mixed precision.
This change avoids computing SE in float16 all together.

As SE block does not take much of computational time, the time and memory benefits of other computations done in float16 can allow to train model faster.  

The runs on NGC and multinode cluster confirm that there is no performance degradation or instabilities when using fp16 with current patch. The speedup is 15% and memory savings are around 3Gb for 21M parameter model.


nemo/collections/asr/parts/jasper.py
```diff
@@ -309,21 +309,22 @@ def __init__(
                 nn.Linear(channels, channels // reduction_ratio, bias=False),
                 activation,
                 nn.Linear(channels // reduction_ratio, channels, bias=False),
-            )
+            ).float()
```
blisc, line 312: Do you need this .float()?

nemo/collections/asr/parts/jasper.py
```diff
@@ -309,21 +309,22 @@ def __init__(
...
 
     def forward(self, x):
         # The use of negative indices on the transpose allow for expanded SqueezeExcite
         batch, channels, timesteps = x.size()[:3]
+        x = x.float()
```
blisc, line 317: Do you need this .float()?

mousebaiker: Yes, definitely.

Otherwise, pooling will be done in fp16 and will cause instabilities.

nemo/collections/asr/parts/jasper.py
mousebaiker: If AMP is enabled, the parameters of Linear can be initialized as float16. 
This is not desirable as we use float32 in block.

titu1994: Can you use the torch amp decorator and set enabled to False instead ?

nemo/collections/asr/parts/jasper.py
titu1994: Same as above?

titu1994 (commented):
LGTM since the .float() seems necessary, @blisc for final review

nemo/collections/asr/parts/jasper.py
blisc: > If AMP is enabled, the parameters of Linear can be initialized as float16.

Are you sure about this? Doesn't amp autocast them as needed?

nemo/collections/asr/parts/jasper.py
blisc: That's the point of autocast. If it is stable in fp16, it will be autocasted. Otherwise it wouldn't be.
Are you claiming that our pooling layer isn't stable in fp16? If so, you need to open a bug on amp.

mousebaiker: `AvgPool` keeps the dtype of the input. But I see what you mean by stability. 

**What I observe is:** 
1. Train without any changes with AMP -> divergence. 
2. Train with just `x = x.float()` added  -> stable training. 

That's why I suspected that the pooling is the issue.

I've just dug deep into the pooling implementation. `AdaptiveAvgPool1d` is implemented via `AdaptiveAvgPool2d`. 
`AdaptiveAvgPool2d` has a short cut for kernels of size 1, calling `torch.mean` operation directly.

I didn't look further but my understanding is that `torch.mean` is implemented via float32 aggregation and then casting, so it should not cause any issues. 

So it's ever weirder that I see such behaviour. 

What would you recommend doing to pinpoint the exact issue?

nemo/collections/asr/parts/jasper.py
mousebaiker: After going through the documentation, I think you're correct. Explicit conversion is not needed.

blisc:
Can you open a new PR?

nemo/collections/asr/parts/jasper.py
blisc: What happens if you move the pool inside the disable autocast context?

jprobichaud:
I'm looking forward for this!  I'm hoping to train a citrinet model from scratch soon on A100 GPUs which could leverage really well mix-precision training.  Is there an ETA for this patch to land in the main trunk?

nemo/collections/asr/parts/jasper.py
mousebaiker: Without float it fails due to precision mismatch.

mousebaiker:
As discussed with Jason and Som, we're leaving the workaround for the SE with AMP right now. Once I get to repro the exact issue, we can follow up with PyTorch team.

@jprobichaud, I am a bit absorbed by watching GTC this week :) 
Hopefully, the patch will land by the beginning of the next week.

blisc (approved):

PR #2037, opened 2021-04-09
Set number of token types to 0 for NMT

MaximumEntropy:
This PR sets the number of token types to 0 by default for NMT since it wasn't being used. Parameters, however, were still being created for it in the model, making it impossible to set`find_unused_parameters=False` in DDP.


nemo/collections/nlp/modules/common/transformer/transformer_utils.py
```diff
@@ -78,6 +78,7 @@ def get_nemo_transformer(
             hidden_act=cfg.get('hidden_act', 'relu'),
             mask_future=cfg.get('mask_future', False),
             pre_ln=cfg.get('pre_ln', False),
+            num_token_types=cfg.get('num_token_types', 2)
```
ericharper, line 81: Should the default be 2? Or 0?

nemo/collections/nlp/modules/common/transformer/transformer_utils.py
```diff
@@ -94,6 +95,7 @@ def get_nemo_transformer(
             attn_layer_dropout=cfg.get('attn_layer_dropout', 0.0),
             hidden_act=cfg.get('hidden_act', 'relu'),
             pre_ln=cfg.get('pre_ln', False),
+            num_token_types=cfg.get('num_token_types', 2)
```
ericharper, line 98: Should the default be 2? Or 0?

nemo/collections/nlp/modules/common/transformer/transformer_utils.py
MaximumEntropy: Not sure what the right default should be. AFAIK, there are maybe 3 different models that use this - NMT, TransformerXL LM, and BERT. NMT and TransformerXL need 0 but BERT needs 2. Neither BERT nor TransformerXL set `num_token_types` in their yaml configs. So I figured it's best to leave this at 2. The other option is to set this to 0 and explicitly set 2 for BERT.

ericharper: Okay, that's sounds good to me.

ericharper (approved):
LGTM. Thanks!

okuchaiev (approved):

PR #2047, opened 2021-04-10
Fix padding formula

nithinraok:


nemo/collections/asr/parts/jasper.py
```diff
@@ -71,7 +71,7 @@ def compute_new_kernel_size(kernel_size, kernel_width):
 def get_same_padding(kernel_size, stride, dilation) -> int:
     if stride > 1 and dilation > 1:
         raise ValueError("Only stride OR dilation may be greater than 1")
-    return (dilation * (kernel_size - 1)) // 2
+    return (kernel_size + (kernel_size - 1) * (dilation - 1) - 1) // 2
```
mousebaiker, line 74: This formula is the exact same formula as the old one. Just expand:
`k + (k - 1) * (d - 1) - 1 = k + kd - k - d  + 1 - 1 = kd - d = d * (k - 1)`

You can also verify numerically:
```
def p1(k, d):
  return (d * (k - 1))//2

def p2(k ,d):
   return (k + (k - 1) * (d - 1) - 1) // 2


for k in range(1, 1001):
   for d in range(1, 1000):
     if p1(k, d) != p2(k, d):
       print(k, d)
```

I would suggest keeping the old formula and just adding tests in this PR.

tests/collections/asr/test_jasper_block.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import random
```
mousebaiker, line 15: This is probably not needed since you're using `torch.rand`.

mousebaiker (changes requested):

nemo/collections/asr/parts/jasper.py
titu1994: @nithinraok The new formula here and the one before are identical. The expanded formula is far harder to interpret and extend to the asymmetric case as well. Revert and just add the tests. As a sanity check, also add the above code snippet as a test.

nithinraok: Sorry was basing my PR on my local changes from rc1 branch which has a previous formula of `(d*k)//2 -1` didn't notice it was fixed in main. changed the formula to its shorter form of `d*(k-1)` funny that I missed it. Will update with tests only

tests/collections/asr/test_jasper_block.py
nithinraok: I thought of using random.randrange to generate random kernel and dilation sizes, but later found out we cannot always find a padding size for conv1d to make input output same for some even kernel sizes like 4 and dilation 3. 
There must be a reason why pytorch didn't support this to append zeros?

titu1994: We force all kernels to odd size so don't consider even kernel shapes.

tests/collections/asr/test_jasper_block.py
```diff
@@ -273,3 +273,23 @@ def test_residual_block_asymmetric_pad_future_context_fallback(self):
         assert ylen[0] == 131
         assert block.mconv[0].pad_layer is None
         assert block.mconv[0]._padding == config['kernel_size'][0] // 2
+
+    @pytest.mark.unit
+    def test_padding_size_conv1d(self):
+        input_size = 1024
```
titu1994, line 279: For a test do we really need such large arrays?

tests/collections/asr/test_jasper_block.py
```diff
@@ -273,3 +273,23 @@ def test_residual_block_asymmetric_pad_future_context_fallback(self):
...
+        output_size = 1024
```
titu1994, line 280: Same as ^

titu1994 (approved):

mousebaiker (approved):
LGTM.

PR #2050, opened 2021-04-12
Entity linking

vadam5:
This pull request adds an entity linking model, dataset, loss, example training code, and demo notebook based on the paper "Self-Alignment Pretraining for Biomedical Entity Representations"

You can run through the notebook tutorial or run everything on the command line.

To run everything on the command line follow these steps

1) Download the UMLS datset file `MRCONSO.RRF` from the NIH website and place it in the `examples/nlp/entity_linking/data` directory.

2) Run the following commands from the `examples/nlp/entity_linking` directory
```
python data/umls_dataset_processing.py
python self_alignment_pretraining.py project_dir=. 
python data/umls_dataset_processing.py --index
python build_index.py --restore
python query_index.py --restore
```
The model will take ~24hrs to train on two GPUs and ~48hrs to train on one GPU. By default the project directory will be "." but can be changed by adding the flag `--project_dir=<PATH>` after each of the above commands and changing `project_dir=<PATH>` in the `self_alignment_pretraining.py` command.


nemo/collections/common/losses/multi_similarity_loss.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
MaximumEntropy, line 1: We should be changing all copyright headers to 2021 :)

nemo/collections/nlp/models/entity_linking/entity_linking_model.py
```diff
@@ -0,0 +1,193 @@
...
+            pretrained_model_name=cfg.language_model.pretrained_model_name,
+            config_file=cfg.language_model.config_file,
+            config_dict=cfg.language_model.config,
+            checkpoint_file=cfg.language_model.lm_checkpoint,
+        )
+
+        # Token to use for the self-alignment loss, typically the first token, [CLS]
+        self._idx_conditioned_on = 0
+        self.loss = MultiSimilarityLoss()
+        self.index = None
```
MaximumEntropy, line 70: Lets see if we can try and register the index as an artifact so it gets saved in the `.nemo` file

tests/collections/nlp/test_entity_linking_model.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+import wget
+from omegaconf import OmegaConf
+
+from nemo.collections.nlp.models import EntityLinkingModel
+
+
+def get_cfg(save_dir):
+    wget.download(
+        "https://raw.githubusercontent.com/vadam5/NeMo/main/examples/nlp/entity_linking/conf/umls_medical_entity_linking_config.yaml",
```
MaximumEntropy, line 28: Why are you using wget to download this file instead of having the yaml file a part of the repo?

MaximumEntropy (changes requested):
Thanks for the PR! It looks great. Let's try and address a few things before we merge.

examples/nlp/entity_linking/data/umls_dataset_processing.py
```diff
@@ -0,0 +1,184 @@
...
+    'SCUI',
+    'SDUI',
+    'SAB',
+    'TTY',
+    'CODE',
+    'STR',
+    'SRL',
+    'SUPPRESS',
+    'CVF',
+]
```
cparisien, line 43: Include a pointer to the UMLS doc page explaining what these headers are.

examples/nlp/entity_linking/data/umls_dataset_processing.py
```diff
@@ -0,0 +1,184 @@
...
+
+            if len(pairs) == 0:
+                # Not enough concepts gathered to make a pair
+                cui = df["CUI"].iloc[idx]
+                names = [df["STR"].iloc[idx]]
+                continue
+
+            # Removing leading C to convert label string to int
+            cui = int(cui[1:])
+            random.shuffle(pairs)
```
cparisien, line 94: Make sure you seed the generator for replicability (I can't see if you did elsewhere).

nemo/collections/nlp/models/entity_linking/entity_linking_model.py
```diff
@@ -0,0 +1,193 @@
...
+        Lightning calls this inside the validation loop with the data from the validation dataloader
+        passed in as `batch`.
+        """
+        input_ids, input_type_ids, input_mask, concept_ids = batch
+        with torch.no_grad():
+            logits = self.forward(input_ids=input_ids, token_type_ids=input_type_ids, attention_mask=input_mask)
+            val_loss = self.loss(logits=logits, labels=concept_ids)
+
+        # No hard examples found in batch,
+        # shouldn't use this batch to update model weights
```
cparisien, line 120: Reword the comment -- val_loss wouldn't update weights anyway

tests/collections/nlp/test_entity_linking_model.py
cparisien: Shouldn't point to your own repo anyway

examples/nlp/entity_linking/build_and_query_index.py
```diff
@@ -0,0 +1,290 @@
...
+                          the concept
+    Returns:
+        A dictionary with the concept ids of the index's most similar 
+        entities as the keys and a tuple containing the string 
+        representation of that concept and its cosine similarity to 
+        the query as the values. 
+    """
+    query_emb = get_query_embedding(query, model).detach().cpu().numpy()
+
+    if cfg.apply_pca:
```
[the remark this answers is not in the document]
cparisien: It's also used for cfg.query_num_factor below

examples/nlp/entity_linking/build_and_query_index.py
```diff
@@ -0,0 +1,290 @@
...
+            unique_ids[concept_id] = (concept, 1 - dist[neighbor_idx])
+
+        neighbor_idx += 1
+
+    unique_ids = dict(unique_ids)
+
+    return unique_ids
+
+
+def main(cfg: DictConfig, top_n: int, restore: bool):
```
cparisien, line 229: Shouldn't top_n belong in either the config file or in the query itself? Feels like a config param. Scope-wise it's more relevant to the query than the index building and could change from one client to the next.

cparisien: Oh I see. It's really "load and query" and if it's not built, go build it first. I think this should be separated into a build stage and a query utility, where the query utility gives a constructive error if not already built. Best to avoid the "surprise! I'm going to go build now."

examples/nlp/entity_linking/conf/tiny_example_entity_linking_config.yaml
```diff
@@ -0,0 +1,90 @@
...
+  precision: 16
+  amp_level: O1
+  accelerator: ddp
+  gradient_clip_val: 0.0
+  log_every_n_steps: 1
+  val_check_interval: 2
+  checkpoint_callback: false
+  logger: false
+model:
+  nemo_path: tiny_example_sap_bert_model.nemo
```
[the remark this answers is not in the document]
cparisien: Why? If this is an example config that aligns with the tutorial notebook, then shouldn't it provide the expected value? At least provide the expected value in a comment, along with something like "The path to the .nemo file created in step X of the tutorial."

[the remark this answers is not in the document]
cparisien: Or ensure they are correctly overridden in the tutorial, if using '???'.
Note that the UMLS yaml below uses paths relative to the project_dir.

cparisien (changes requested):
Looking good. Just a few straightforward recommendations.

nemo/collections/nlp/data/entity_linking/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2021 The Google AI Language Team Authors and
+# The HuggingFace Inc. team.
```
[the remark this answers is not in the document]
MaximumEntropy: Good point, if parts of any of the files you've added use code from Google's repos or Huggingface then we should have this in. If not, we should remove it.

nemo/collections/common/losses/multi_similarity_loss.py
```diff
@@ -0,0 +1,95 @@
...
+        """Returns definitions of module output ports.
+        """
+        return {"loss": NeuralType(elements_type=LossType())}
+
+    def __init__(
+        self,
+        scale_pos: Optional[float] = 2.0,
+        scale_neg: Optional[float] = 40.0,
+        offset: Optional[float] = 0.5,
+        margin: Optional[float] = 0.1,
```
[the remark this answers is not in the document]
vadam5: I should probably add a place in the config where these values can be specified.

nemo/collections/nlp/data/entity_linking/__init__.py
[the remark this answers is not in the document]
vadam5: I hadn't noticed this either. I'll make sure its accurate for all the files in this PR

nemo/collections/nlp/models/entity_linking/entity_linking_model.py
vadam5: I actually think index doesn't need to be registered as an artifact. Its optionally created after the model is fully trained and saved. I removed the index attribute from the entity linking encoder.

tests/collections/nlp/test_entity_linking_model.py
vadam5: about using wget, how should I specify the path to the config in a way that works with Jenkins? The unit test fails if I put the relative path to the config file from the test file. I was using the tests in tests/collections/nlp/test_nlp_exportables.py where wget is used to load the config as an example.

vadam5: @cparisien Its pointing to my own repo b/c the config file won't be in NVIDIA nemo until after the PR is merged in. There must be a better way to go about doing this.

nemo/collections/nlp/data/language_modeling/lm_bert_dataset.py
```diff
@@ -65,42 +66,14 @@ def __init__(
...
         # from main memory when needed during training.
 
-        if sentence_idx_file is None:
-            data_dir = data_file[: data_file.rfind('/')]
-            mode = data_file[data_file.rfind('/') + 1 : data_file.rfind('.')]
-            sentence_idx_file = f"{data_dir}/{mode}_sentence_indices.pkl"
+        # Try and load sentence indices file if already exists
+        sentence_indices, sentence_idx_file, data_dir = load_data_indices(
+            sentence_idx_file, data_file, "sentence_indices"
+        )
```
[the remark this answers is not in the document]
vadam5: Tested it myself by running the BERT language modeling example and the unit and CI tests also test it.

tests/collections/nlp/test_entity_linking_model.py
vadam5: I decided to just define a much simpler config in the unit test, so no need to load in a config from anywhere

examples/nlp/entity_linking/query_index.py
```diff
@@ -0,0 +1,166 @@
...
+
+    unique_ids = dict(unique_ids)
+
+    return unique_ids
+
+
+def main(cfg: DictConfig, restore: bool):
+    """
+    Loads faiss index and allows commandline queries 
+    to the index. Builds new index if one hasn't been built yet.
```
cparisien, line 113: Just need to fix the "build" comment since that doesn't happen here anymore

cparisien (approved):
Looks good to me! Splitting the build/query is more clear now.

MaximumEntropy (approved):
LGTM. Thanks!

PR #2064, opened 2021-04-16
Update preproc_mt_data.py tarred dataset

vadam5:
If just a metadata_file is passed in through the config, it won't enter the correct elif statement

### DIFF
--- a/nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
+++ b/nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
@@ -184,7 +184,7 @@ class MTDataPreproc:
                         f"Using tarred dataset created at {self.train_tar_files} and metadata created at {self._cfg.train_ds.metadata_file}"
                     )
                 elif cfg.train_ds.get('tar_files') is not None and cfg.train_ds.get('metadata_file') is None:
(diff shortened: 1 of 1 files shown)


MaximumEntropy:
I think @ericharper may have had this in place on purpose to make sure both metadata and tarfiles were provided and not just the metadata. I'm fine with providing just the metadata JSON file (even better would be to allow specifying preproc_out_dir to avoid specifying tokenizers as well). 

Can you also handle the case where `cfg.train_ds.get('tar_files')` is not None and `cfg.train_ds.get('metadata_file')` is None and throw an error in that case since we always require the metadata file.

vadam5:
> I think @ericharper may have had this in place on purpose to make sure both metadata and tarfiles were provided and not just the metadata. I'm fine with providing just the metadata JSON file (even better would be to allow specifying preproc_out_dir to avoid specifying tokenizers as well).
> 
> Can you also handle the case where `cfg.train_ds.get('tar_files')` is not None and `cfg.train_ds.get('metadata_file')` is None and throw an error in that case since we always require the metadata file.

Yes, I can handle that case as well.

MaximumEntropy (approved):
LGTM. Thanks!

PR #2080, opened 2021-04-20
Add Multi GPU Rank tests

titu1994:
# Changelog
- Add jenkins test to run DDP predict and measure the model internal ranks calculated
- Exit early if these ranks in logs dont correspond to expected global ranks.

### DIFF
--- a/tests/core_ptl/check_for_ranks.py
+++ b/tests/core_ptl/check_for_ranks.py
@@ -1,0 +1,155 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/core_ptl/check_for_ranks.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def validation_epoch_end(self, loss):
+        self.log("val_loss", torch.stack(loss).mean())
+
+
+def instantiate_multinode_ddp_if_possible():
+    num_gpus = torch.cuda.device_count()
+    trainer = Trainer(gpus=num_gpus, accelerator='ddp', logger=None, checkpoint_callback=None)
+
+    exp_manager_cfg = ExpManagerConfig(exp_dir='./ddp_check/', use_datetime_version=False, version="")
```
ericharper, line 83: Could also use `explicit_log_dir` here.

ericharper (approved):
LGTM! Thanks!

tests/core_ptl/check_for_ranks.py
titu1994: I just realized the first arg is explicit_log_dir. Oh well, will update in future if script needs update.

PR #2085, opened 2021-04-21
remove sil tag for punct SH

ekmb:


yzhang123 (approved):
I assume this is tested with SH? Then green light here.
Thanks!

PR #2087, opened 2021-04-22
configure input capitalization

yzhang123:
* adding configuration to distinguish between lower_cased and cased input, affects whitelist, word classes
* added test
* added pynini sentence predict test to jenkins
* deleted sentence boundary exception


nemo_text_processing/text_normalization/data/sentence_boundary_exceptions.txt
```diff
@@ -1,15 +1,4 @@
-mr.
```
ekmb, line 1: Do we need this file or can we use the whitelist? added to our to-do

nemo_text_processing/text_normalization/normalize.py
```diff
@@ -22,10 +22,11 @@
 from nemo_text_processing.text_normalization.verbalizers.verbalize_final import VerbalizeFinalFst
 from tqdm import tqdm
 
+INPUT_CASE = "lower_cased"  # cased
```
ekmb, line 25: can we infer this from the input itself?  and in the case when it's not possible, use this param? could you add it to the arguments?

ekmb (changes requested):
Please add test cases to cover the changes.

nemo_text_processing/text_normalization/data/sentence_boundary_exceptions.txt
yzhang123: thanks, yes its todo

nemo_text_processing/text_normalization/normalize.py
yzhang123: let me figure out how to add it to args

PR #2088, opened 2021-04-22
Teacher student model

titu1994:
# Changelog
- Prototype generic TeacherStudentModelPT

# Usage
In any current NeMo model config, attach the following section

```yaml
model:
  # Teacher details
  teacher:
    model_path: ???
    
    # optional - distillation loss config
    loss:
      _target_: 'torch.nn.KLDivLoss'
      log_target: true
      reduction: 'batchmean'

  ... rest of student model config as usual
```

Then build a model using `TeacherStudentModelPT` with this config and call `trainer.fit()`.

# Working
To be expanded once API is finalized.


nemo/collections/asr/models/ctc_models.py
```diff
@@ -544,7 +545,20 @@ def training_step(self, batch, batch_nb):
...
-        tensorboard_logs = {'train_loss': loss_value, 'learning_rate': self._optimizer.param_groups[0]['lr']}
+        # Distillation support
+        if self.is_being_distilled():
+            if self.is_student_model():
+                loss_key = 'input'
+            else:
+                loss_key = 'target'
+
+            # Register the tensor for the loss function
+            self.register_distillation_tensor(loss_key=loss_key, tensor=log_probs)
```
blisc, line 556: We should aim to minimize this. Can we make some assumptions like all distillation loss requires keys to be prepended with `teacher_` and `student_`? Then we can just make this part 1 line: `return {"log_probs": log_probs}`, and prepend it with student or teacher as necessary in the parent class.

titu1994: https://pytorch.org/docs/stable/generated/torch.nn.KLDivLoss.html in its forward call requires (input=X, target=Y). Both must be tensors of same shape. So the `loss_key` here refers to the actual keys of the loss itself.

`return {"log_probs": log_probs}` is a special case of just 1 loss, but the codebase makes no assumption of what loss you use - and this return semantic will fail for all other losses. Since KLDiv is the most common loss that must be reduced anyway, we can extract this section out of here into a default implementation for the default loss. However, the generality of loss key cant be lost, especially once I add support for secondary losses.

blisc: You don't have to make an assumption of what loss the model writer uses. They know what loss they will use and what keys are needed. It's up to the model writers to make sure than their keys are correct.

titu1994: Returning values causes multiple issues - I can't add support for secondary losses (cosine similarity loss) cause I don't know which tensor belongs to primary loss and which to second. 

Another issue is implementer will need to return somehow the actual student loss if they want to add it to distillation loss. 

Finally, returning something makes it harder to change loss from config and I'd have to remove that support entirely. 

There's too many implementation issues with it.

agemagician:
Hi @titu1994 ,

Thanks a lot for this awesome work.
Do you have any idea when this pull request will be ready and merged to the main ?

titu1994:
It will take time. There are experiments underway to see the effectiveness of this approach.

PR #2099, opened 2021-04-23
NMT Eval Loss Fixes

MaximumEntropy:
This PR does the following

1. Implements NLL Loss to be used at validation/test time for NMT (without label smoothing)
2. Removes logic that changes the attention mask to not be causal at test time for beam search, which was causing high validation loss


okuchaiev (approved):
lgtm

Issue #2094, opened 2021-04-22
Inverse Text Normalization text-processing breaks when there's an empty line the input text


taufique74:
**Bug Description**

While running `python run_predict.py --input=input.txt --output=output.txt`, it breaks saying `pywrapfst.FstOpError: Operation failed`. I noticed there is an empty line in `input.txt`, and after removing that empty line the error was gone.

**Steps to reproduce the bug**

Let's first create a text file `input.txt` with some examples along with an empty line
```bash
$ echo "one hundred and twenty four" >> input.txt
$ echo "" >> input.txt
$ echo "may third nineteen sixty" >>
```

Now if we run `python run_predict.py --input=input.txt --output=output.txt`, it will throw an error.
Here is the complete traceback:
```python-traceback
Traceback (most recent call last):
  File "/home/ubuntu/prtc/itn/NeMo/nemo_text_processing/inverse_text_normalization/inverse_normalize.py", line 209, in inverse_normalize_nemo
    text = inverse_normalize(input, verbose=verbose)
  File "/home/ubuntu/prtc/itn/NeMo/nemo_text_processing/inverse_text_normalization/inverse_normalize.py", line 169, in inverse_normalize
    tagged_text = select_tag(tagged_lattice)
  File "/home/ubuntu/prtc/itn/NeMo/nemo_text_processing/inverse_text_normalization/inverse_normalize.py", line 125, in select_tag
    tagged_text = pynini.shortestpath(lattice, nshortest=1, unique=True).string()
  File "extensions/_pynini.pyx", line 464, in _pynini.Fst.string
  File "extensions/_pynini.pyx", line 509, in _pynini.Fst.string
_pywrapfst.FstOpError: Operation failed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "run_predict.py", line 77, in <module>
    inverse_normalizer_prediction = inverse_normalizer(data, verbose=args.verbose)
  File "/home/ubuntu/prtc/itn/NeMo/nemo_text_processing/inverse_text_normalization/inverse_normalize.py", line 211, in inverse_normalize_nemo
    raise Exception
Exception
```

**Environment overview**

 - Environment location:  Cloud(AWS)
 - Method of NeMo install: from source

**Environment details**

- OS version: Ubuntu 18.04.5 LTS
- PyTorch version: 1.8.1+cu102
- Python version: 3.7

taufique74:
Hey @yzhang123 
[...]

PR #2100, opened 2021-04-23
bug fix for handling empty lines in input text

taufique74:
This fixes the empty line issue of `nemo_text_processing/inverse_text_normalization/run_predict.py` discussed in https://github.com/NVIDIA/NeMo/issues/2094

@yzhang123


yzhang123:
@taufique74 first of all, thanks for catching the bug and PR! Since we would prefer to keep the line breaks which fits more the user's expectations i will replace this PR my this one https://github.com/NVIDIA/NeMo/issues/2094.
And close this one.

PR #2114, opened 2021-04-25
[docs] Reviewed/edited NeMo doc files

LynseyFabel:
I've reviewed and edited the NeMo doc files.  All documentation have been reviewed.  Majority of them were edited/updated.

### DIFF
--- a/docs/source/asr/speaker_recognition/configs.rst
+++ b/docs/source/asr/speaker_recognition/configs.rst
@@ -1,32 +1,27 @@
 NeMo ASR Configuration Files
 ============================
 
-This page covers NeMo configuration file setup that is specific to speaker recognition models.
-For general information about how to set up and run experiments that is common to all NeMo models (e.g.
-experiment manager and PyTorch Lightning trainer parameters), see the :doc:`../../starthere/core` page.
+The model section of NeMo ASR configuration files requires information about the datasets being used, the preprocessor for audio files, 
+parameters for any augmentation being performed, as well as the model architecture specification. The sections on this page cover each 
+of these in more detail.
 
-The model section of NeMo ASR configuration files will generally require information about the dataset(s) being
-used, the preprocessor for audio files, parameters for any augmentation being performed, as well as the
-model architecture specification.
-The sections on this page cover each of these in more detail.
+Example configuration files for all of the Speaker related scripts can be found in the config directory of the examples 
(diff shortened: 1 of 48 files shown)


docs/source/core/core.rst
```diff
@@ -233,13 +233,13 @@ YAML
...
 
-- trainer
-- exp_manager
-- model
+- ``trainer``
+- ``exp_manager``
+- ``model``
 
-Model configuration always contain ``train_ds``, ``validation_ds``, ``test_ds``, and ``optim``.  Model architectures vary across 
+The model configuration section always contain ``train_ds``, ``validation_ds``, ``test_ds``, and ``optim``.  Model architectures vary across
```
redoctopus, line 242: s/contain/contains

docs/source/asr/speaker_recognition/configs.rst
```diff
@@ -1,32 +1,27 @@
 NeMo ASR Configuration Files
```
redoctopus, line 1: Just noticed that this header is wrong, so links that try to go to the ASR configuration files documentation (rather than for speaker recognition, which is what's actually here) end up here instead.

docs/source/core/core.rst
```diff
@@ -287,9 +287,7 @@ A NeMo configuration file should look similar to the following:
         decoder:
             ...
 
-More specific details about configuration files for each collection can be found on the following pages:
-
-:ref:`NeMo ASR Configuration Files`
+For more information about configuration files for each collection, refer to :ref:`NeMo ASR Configuration Files` section.
```
redoctopus, line 290: We should add additional links to configuration file info pages of the other tasks, as well. We have ASR, speaker diarization, and speaker recogition so far.
(Also, this is the aforementioned wrong link due to an incorrect header on the other page. This link should go to the ASR config files docs but instead leads to the speaker recognition docs, which can be fixed by updating the speaker recognition page title.)

redoctopus (commented):
Thanks for the thorough documentation update! It's very much appreciated.
I've looked through the changes to only the documentation that I'm familiar with, mainly in ASR, and left a few comments.

redoctopus:
Hi Lynsey, could you resolve the merge conflicts with current main? I took a look and they seem to be from others' recent documentation updates.

okuchaiev:
looks like all these changes are already there.

PR #2117, opened 2021-04-26
Modified convert_to_tarred_audio_dataset.py to accept another argument also as input

aahameda:
The newly added argument(dataset_dir) is concatenated with the path obtained from the manifest file. This solves the problem of FileNotFoundError as an absolute path is given for the datasets.

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -23,7 +23,7 @@ from enum import Enum
 from functools import total_ordering
 from pathlib import Path
 from typing import Dict, List, Optional, Union
-
+import os
 import hydra
 import wrapt
 from omegaconf import DictConfig, OmegaConf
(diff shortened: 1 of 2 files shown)


nemo/core/classes/common.py
```diff
@@ -663,12 +663,15 @@ def from_pretrained(
...
         url = location_in_the_cloud.replace(filename, "")
-        cache_dir = Path.joinpath(Path.home(), f'.cache/torch/NeMo/NeMo_{nemo.__version__}/{filename[:-5]}')
+        if os.getenv('NEMO_CACHE_DIR') is not None and os.path.isdir(os.getenv('NEMO_CACHE_DIR')):
+            cache_dir = Path.joinpath(Path(os.getenv('NEMO_CACHE_DIR')), f'.cache/torch/NeMo/NeMo_{nemo.__version__}/{filename[:-5]}')
+        else:
+            cache_dir = Path.joinpath(Path.home(), f'.cache/torch/NeMo/NeMo_{nemo.__version__}/{filename[:-5]}')
         # If either description and location in the cloud changes, this will force re-download
-        cache_subfolder = hashlib.md5((location_in_the_cloud + description).encode('utf-8')).hexdigest()
+        #cache_subfolder = hashlib.md5((location_in_the_cloud + description).encode('utf-8')).hexdigest()
         # if file exists on cache_folder/subfolder, it will be re-used, unless refresh_cache is True
```
titu1994, line 672: This is necessary. Please revert.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -77,6 +78,10 @@
     "--manifest_path", default=None, type=str, required=False, help="Path to the existing dataset's manifest."
 )
 
+parser.add_argument(
+    "--dataset_dir", default=None, type=str, required=True, help="Path to the dataset's directory."
```
titu1994, line 82: Don't make this required, default it to empty string to be compliant with older scripts.

titu1994: Firstly, Tarred datasets don't care about any relative or absolute path - they are all flattened out. All paths are flat relative to the loaded tarfile chunk itself so I don't see the reason for this flag to exist. It has no effect since tarfile should flatten it out.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -473,7 +478,7 @@ def _create_shard(self, entries, target_dir, shard_id):
             base = base.replace('.', '_')
             squashed_filename = f'{base}{ext}'
             if squashed_filename not in count:
-                tar.add(entry['audio_filepath'], arcname=squashed_filename)
+                tar.add(args.dataset_dir+"/"+entry['audio_filepath'], arcname=squashed_filename)
```
titu1994, line 481: Remove the "/" part - that needs to be in the dataset dir, not automatically attached.

titu1994 (changes requested):

titu1994:
Could you sign the commits (instructions are provided in the dco bot output), and explain your reasoning for dataset paths argument?

Tarred datasets flatten all paths, and you can create them from a provided dataset directory and write then to another directory, so this flag is odd.

aahameda:
Created merge request for wrong branch

PR #2138, opened 2021-04-29
Update how artifacts work

okuchaiev:
This PR changes the following:

* how register_artifacts work
* how save_to works

The usage of both register_artifacts and save_to method remains identical.
Key differences in behaiviour:
* register_artifacts always guarantees to return existing path to artifact at the time of call. 
* save_to will replace and hash artifact names in config file before creating .nemo file. This is necessary to prevent name collisions
* when restoring artifacts from .nemo file it will create a temporary folder with artifacts for the lifetime of a model
* temporary folders associated with the model get deleted by destructor

TODOs:

* Some unittests fail
* Model parallel save_to (?)


nemo/core/classes/modelPT.py
```diff
@@ -157,86 +164,79 @@ def register_artifact(self, config_path: str, src: str):
...
+            return_path = os.path.abspath(src)
+        # this is the case when artifact must be retried from the nemo file
+        # we are assuming that the location of the right nemo file is available from _MODEL_RESTORE_PATH
+        elif src.startswith("nemo:"):
+            tmpdir = _CACHE_ROOT + f"_TMPDIRS/{uuid.uuid4().hex}"
+            os.makedirs(tmpdir)
+            # TODO: Write a code to delete this in destructor
+            outfolder = self._unpack_nemo_file(path2file=_MODEL_RESTORE_PATH, out_folder=tmpdir)
+            #print(f" AAA: {os.listdir(tmpdir)}, B: {src}")
+            return_path = os.path.abspath(os.path.join(outfolder, src[5:]))
```
titu1994, line 201: In this case ArtifactPathType should be tarfile instead of local path.

titu1994: Also, I think we should avoid extracting file into temp dir for a long time like this, it would be better to enter a tarfile and check whether it contained the value of not, finally deleting the temp dir immediately if it doesn't. Deferring cleanup can cause a lot of issues in slurm envs + pollute the local storage if there is failure to delete files.

nemo/core/classes/modelPT.py
```diff
@@ -157,86 +164,79 @@ def register_artifact(self, config_path: str, src: str):
...
-                )
+            else:
+                raise ValueError(f"Directly referencing artifacts from other nemo files isn't supported yet")
 
-            self.artifacts[config_path] = archive_item
-            return used_src
-        else:
-            return src
+    def _update_artifact_paths(self, path2yaml_file):
+        if self.artifacts is not None and len(self.artifacts) > 0:
```
titu1994, line 234: This defers updating of config to the moment the file is written, which means at least some training time has already gone. I would rather overwrite self.cfg.* with the artifact paths at the moment the register method is called, so that the actual config is updated. Then the normal save will use the self.cfg to write the config.

nemo/core/classes/modelPT.py
okuchaiev: no, it should be local path because I am going to return and refer to this artifact based on local path.

I added destructor to make sure tmp folders get deleted once model instance is destroyed

nemo/core/classes/modelPT.py
okuchaiev: overwriting self.cfg would be wrong because model will be still in use. save_to hashes artifact names and replaces them with paths starting with "nemo:" specifically for the purposes of .nemo file

nemo/core/classes/modelPT.py
```diff
@@ -152,91 +158,104 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # ModelPT wrappers over subclass implementations
         self.training_step = model_utils.wrap_training_step(self.training_step)
 
+    def __del__(self):
+        """This is a destructor to delete temporary folders associated with this model instance.
+        """
+        if hasattr(self, "to_delete"):
+            for item in self.to_delete:
+                if os.path.isdir(item):
+                    shutil.rmtree(item)
```
titu1994, line 167: Destructors are not guarenteed to be called on clusters since they can terminate the job. Please use only temporary directories and not destructors.

nemo/core/classes/modelPT.py
titu1994: Destructors are not guaranteed to be called, especially in slurm environments. So jobs which run out of time will just keep adding more data to the cache dir. Thats why I had TAR_PATH to denote that at save time, open a temp tarfile and use the artifact inside it.

nemo/core/classes/modelPT.py
```diff
@@ -152,91 +158,104 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+
+        assert os.path.exists(return_path)
+        artifact_item = model_utils.ArtifactItem()
+        artifact_item.path = os.path.abspath(src)
+        artifact_item.path_type = model_utils.ArtifactPathType.LOCAL_PATH
+        self.artifacts[config_path] = artifact_item
+        # we were called by ModelPT
+        if hasattr(self, "_cfg"):
+            with open_dict(self._cfg):
+                self._cfg.update_node(config_path, return_path)
```
titu1994, line 228: Use self.cfg - so that wandb and loggers are also updated.

nemo/core/classes/modelPT.py
titu1994: Ok that makes sense

tests/core/test_save_restore.py
```diff
@@ -38,6 +39,8 @@ def __init__(self, cfg, trainer=None):
         super(MockModel, self).__init__(cfg=cfg, trainer=trainer)
         self.w = torch.nn.Linear(10, 1)
 
+        print(f"ZZZZZZZZ: {self.cfg}")
```
titu1994, line 42: Revert?

titu1994 (commented):
This temporary directory and destructor is not safe for slurm environments.

nemo/core/classes/modelPT.py
okuchaiev: got rid of global cache

nemo/core/classes/modelPT.py
okuchaiev: got rid of it

nemo/core/classes/modelPT.py
okuchaiev: done

okuchaiev:
@titu1994 and @ericharper are we good to merge this?

titu1994:
Yes looks good to me.

nemo/core/classes/modelPT.py
```diff
@@ -152,91 +157,161 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+
+        # this is the case when artifact must be retried from the nemo file
+        # we are assuming that the location of the right nemo file is available from _MODEL_RESTORE_PATH
+        elif src.startswith("nemo:"):
+            # tmpdir = _CACHE_ROOT + f"_TMPDIRS/{uuid.uuid4().hex}"
+            # os.makedirs(tmpdir)
+            # if not hasattr(self, 'to_delete'):
+            #     self.to_delete = set()
+            # self.to_delete.add(tmpdir)
+            # outfolder = self._unpack_nemo_file(path2file=app_state.model_restore_path, out_folder=tmpdir)
```
titu1994, line 225: We can remove the commented code out

tests/core/test_save_restore.py
titu1994: Print can be cleared up.

tests/core/test_save_restore.py
```diff
@@ -198,12 +204,17 @@ def test_mock_restore_from_config_only(self):
...
+            assert os.path.basename(model_copy.temp_file).endswith(os.path.basename(model.temp_file))
+            # assert filecmp.cmp(model.temp_file, model_copy._cfg.temp_file)
+            assert model.cfg.xyz == model_copy.cfg.xyz
 
         # Restore test
-        assert isinstance(model_config_copy, DictConfig)
-        assert model.cfg.temp_file == model_config_copy.temp_file
-        assert model.cfg.xyz == model_config_copy.xyz
+        # assert isinstance(model_config_copy, DictConfig)
+        # assert model.cfg.xyz == model_config_copy.xyz
```
titu1994, line 217: No tests assert ?

titu1994 (commented):
Some of the tests need a bit of cleanup - one has no asserts.

tests/core/test_save_restore.py
okuchaiev: lines 211 and 212 have asserts

titu1994 (approved):
Looks good.

tests/core/test_save_restore.py
okuchaiev: it has few asserts above

ericharper (approved):
LGTM!

PR #2140, opened 2021-04-29
Update bert

yzhang123:


nemo/collections/nlp/models/language_modeling/bert_lm_model.py
```diff
@@ -165,8 +167,8 @@ def validation_step(self, batch, batch_idx):
         mlm_log_probs, nsp_logits = self._parse_forward_outputs(forward_outputs)
         _, _, loss = self._compute_losses(mlm_log_probs, nsp_logits, output_ids, output_mask, labels)
         self.validation_perplexity(logits=mlm_log_probs)
-        tensorboard_logs = {'val_loss': loss}
-        return {'val_loss': loss, 'log': tensorboard_logs}
+        self.log(f'val_loss', loss)
```
ericharper, line 170: I think you should only log in validation_epoch_end. Logging in validation_step will log the value for each validation batch.

ericharper (approved):
LGTM, thanks!

PR #2143, opened 2021-04-30
FastSpeech 2 Test & Docs

redoctopus:
Adding:
- Data loading test for the `FastSpeech2Dataset`
- FS2 data format documentation
- Some TTS class APIs


tests/conftest.py
```diff
@@ -23,7 +23,7 @@
 
 # Those variables probably should go to main NeMo configuration file (config.yaml).
 __TEST_DATA_FILENAME = "test_data.tar.gz"
-__TEST_DATA_URL = "https://github.com/NVIDIA/NeMo/releases/download/v0.11.0/"
+__TEST_DATA_URL = "https://github.com/NVIDIA/NeMo/releases/download/v1.0.0rc1/"
```
blisc, line 26: You should double check with oleksii before making this change

redoctopus: Checked w/ Oleksii during the meeting, should be fine.

blisc (approved):

PR #2144, opened 2021-04-30
Removing graphsurgeon optional dependency, improving import error rep…

borisfom:
…orting
I did not remove graph surgeon install from Dockerfile - as it's nice to have anyway and maybe some users already use it.
But I did remove the extra GS step on export.

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -436,12 +436,11 @@ class Serialization(ABC):
             instance = hydra.utils.instantiate(config=config)
         else:
             instance = None
-
+            imported_cls_tb = None
             # Attempt class path resolution from config `target` class (if it exists)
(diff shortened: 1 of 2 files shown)


okuchaiev (approved):
lgtm

nemo/core/classes/common.py
```diff
@@ -441,11 +441,12 @@ def from_config_dict(cls, config: DictConfig):
             if 'target' in config:
                 target_cls = config.target
                 imported_cls = None
+                imported_cls_tb = None
```
titu1994, line 444: This needs to be moved outside the if `target` in config part,

titu1994 (changes requested):

titu1994 (approved):
LGTM, thanks !

PR #2146, opened 2021-05-01
Fix Nemo eval nondeterminism

PiotrDabkowski:
In the eval mode the model should be deterministic (return the same value for the same input).

I find it strange that this was not fixed before - this makes exports and evaluation non-deterministic.


titu1994 (approved):

titu1994:
Thanks for the PR ! It was an oversight since for papers we use batch size 1 with transcribe() method, and it disables dithering inside.

PR #2150, opened 2021-05-04
[NMT] Multi-validation Patch

ericharper:
We instantiate a TorchMetric for each dataloader. This  needs to happen after the dataloaders are created. Moved the logic for instantiating TorchMetrics to after the dataloaders are created. This is necessary when restoring from .nemo.


MaximumEntropy (approved):
LGTM, thanks!

PR #2153, opened 2021-05-05
fix for electronic

ekmb:

### DIFF
--- a/nemo_text_processing/text_normalization/taggers/electronic.py
+++ b/nemo_text_processing/text_normalization/taggers/electronic.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 
-from nemo_text_processing.text_normalization.graph_utils import NEMO_SIGMA, GraphFst
+from nemo_text_processing.text_normalization.graph_utils import NEMO_ALPHA, NEMO_DIGIT, NEMO_SIGMA, GraphFst
 
 try:
     import pynini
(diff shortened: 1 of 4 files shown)


nemo_text_processing/text_normalization/taggers/electronic.py
```diff
@@ -34,7 +34,12 @@ class ElectronicFst(GraphFst):
     def __init__(self):
         super().__init__(name="electronic", kind="classify")
 
-        username = pynutil.insert("username: \"") + NEMO_SIGMA + pynutil.insert("\"") + pynini.cross('@', ' ')
+        username = (
+            pynutil.insert("username: \"")
+            + pynini.closure(NEMO_ALPHA | NEMO_DIGIT | pynini.accep('.'))
```
yzhang123, line 39: this would also accept ".abc@gmai.com" could you make sure it doesnt start with .?

yzhang123 (changes requested):

okuchaiev:
@ekmb  note that v1.0.0 has been cut. If this is a bugfix, send it to v1.0.0

nemo_text_processing/text_normalization/taggers/electronic.py
ekmb: updated. @yzhang123, note that `test .b@gmail.com~test . b at g mail dot com` is still tagged unless we want to treat punct+email as a special case in the tokenizer_and_classify.py().

nemo_text_processing/text_normalization/taggers/electronic.py
```diff
@@ -34,8 +34,19 @@ class ElectronicFst(GraphFst):
     def __init__(self):
         super().__init__(name="electronic", kind="classify")
 
-        username = pynutil.insert("username: \"") + NEMO_SIGMA + pynutil.insert("\"") + pynini.cross('@', ' ')
-        domain_graph = pynutil.insert("domain: \"") + NEMO_SIGMA + pynutil.insert("\"")
+        username = (
+            pynutil.insert("username: \"")
+            + NEMO_ALPHA
+            + pynini.closure(NEMO_ALPHA | NEMO_DIGIT | NEMO_NOT_SPACE)
```
yzhang123, line 40: could you also use symbols.tsv in here for username and domain ? to prevent the input being in NEMO_NOT_SPACE\symbols

do we need all the symbols in tsv? e.g. can "{" appear in email?

yzhang123 (changes requested):

nemo_text_processing/text_normalization/taggers/electronic.py
ekmb: Based on wiki, `{` could appear username, but domain can only take `.` or `-`

yzhang123 (approved):

PR #2156, opened 2021-05-05
Fix text classification doc

VahidooX:


docs/source/tts/api.rst
```diff
@@ -0,0 +1,43 @@
+NeMo TTS Collection API
+=======================
+
```
ekmb, line 3: @VahidooX did you add this?

PR #2113, opened 2021-04-24
[NMT] Add support for multiple validation and test dataloaders

ericharper:
Specify multiple validation (or test) dataloaders by inputting a list of src (tgt) files.

# Usage
Replace `src_file_name` and `tgt_file_name` with a list of file paths.
```
    model.validation_ds.src_file_name=/data/wmt14-en-de.src \
    model.validation_ds.tgt_file_name=/data/wmt14-en-de.ref \
```
Becomes:
```
    model.validation_ds.src_file_name=[/data/wmt13-en-de.src,/data/wmt14-en-de.src] \
    model.validation_ds.tgt_file_name=[/data/wmt13-en-de.ref,/data/wmt14-en-de.ref] \
```
When using `val_loss` or `val_sacreBLEU` for the `exp_manager.checkpoint_callback_params.monitor` then the 0th indexed dataloader will be used as the monitor. This ensures backwards compatibility.

Additionally, any validation dataloader can now be used as a monitor by appending the index as below:
```
    exp_manager.checkpoint_callback_params.monitor=val_sacreBLEU_dl_index_1
```
Multiple test datasets works exactly the same way as validation datasets, simply replace `validation_ds` by `test_ds` in the above examples.

PR #2160, opened 2021-05-06
Multilingual training for NMT

aklife97:
This PR adds support for multilingual training for NMT.

### Usage ###
To train a model in multilingual mode, simply set the following:

* `model.multilingual=True`
* `model.encoder_tokenizer.library=sentencepiece` and `model.decoder_tokenizer.library=sentencepiece`. 
   Currently we only support sentencepiece for multilingual NMT.
* `model.src_language=['lang1','lang2']` or `model.tgt_language=['lang1','lang2']`
   This PR only supports `many-to-one` or `one-to-many` model, as a result, only either of the above can be a list.
* `model.train_ds.metadata_file = [a,b]` if using tarred dataset
    or
   `model.train_ds.src_file_name = [a.src,b.src]` and `model.train_ds.tgt_file_name = [a.ref,b.ref]`
* MultiValidation from  #2113 can be used to provide val sets from different languages

NOTE: All the lists in the configuration, including src/tgt_languages, train_set, val_set need to be in the same order. 
That is,
if `model.tgt_language=['de','fr']`, the train data should also be provided as `['de_train_data','fr_train_data']`.

### Additional Info ###
Under the hood, the model automatically adds language identifier tokens which are created as `control_symbols` in sentencepiece. 
So for instance, if `model.tgt_language=['de','fr']`, `<de>` is prepended to all German source sentences and `<fr>` is prepended to all French source sentences.

### Edit ###
The PR also does the following changes to the current `TarredTranslationDataset`:
* Adjust `__len__` of the dataset based on the world size when shard strategy is set to `scatter`.
* Change naming convention of fragment pkl files (during tar file creation in `preproc_mt_data`) to make them globally unique in the dataset.
* Remove `load_from_cached_dataset`.
* Small fix for a variable name and file ptr when only tar metadata is passed.


nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -147,6 +179,10 @@ def __init__(self, cfg: MTEncDecModelConfig, trainer: Trainer = None) -> None:
                         raise ValueError(
                             'src_file_name and tgt_file_name needed to create tarred dataset but could not be found.'
                         )
+                    if cfg.get('multilingual'):
+                        raise ValueError(
+                            'Tarred dataset cannot be created with multilingual tag set to True. Pre-process each dataset one-by-one.'
```
ericharper, line 184: Can't we loop through the datasets and process them one-by-one automatically?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -433,21 +479,50 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
+                src_file_list = cfg.src_file_name
+                tgt_file_list = cfg.tgt_file_name
+
+            if len(src_file_list) != len(tgt_file_list):
+                raise ValueError(
+                    'The same number of filepaths must be passed in for source and target while training multilingual.'
+                )
+
+            datasets = []
+            for pos, src_file in enumerate(src_file_list):
```
ericharper, line 495: This is a nitpick but I'd rather use `idx` or `index` over `pos`.

MaximumEntropy (changes requested):
Some initial comments - looks super cool!

ericharper:
Overall this is looking really nice. 

Could you help to explain further how we're getting multi-lingual models (many to one and one to many) by adding multiple datasets and a language identifier token?

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -430,3 +440,109 @@ def __iter__(self):
...
+                yield val
+            except StopIteration:
+                self.iterables[ind] = self.get_iterable(self.datasets[ind])
+                if n < self.N:
+                    n -= 1
+
+    def __len__(self):
+        return self.N
+
+    @staticmethod
```
MaximumEntropy, line 524: Can we also add random sampling with specified sampling ratios irrespective of length and such?

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
aklife97: Actually, we can... Will add this in the next commit.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         cache_data_per_node: bool = False,
         use_cache: bool = False,
         reverse_lang_direction: bool = False,
+        prepend_id: int = None,
```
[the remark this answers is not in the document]
aklife97: The idea for just calling it prepend_ids was that it can hold anything that we need to prepend. So in the future, in case we wish to prepend anything (which may not necessarily be a language token), we can use the same list.
Though, if you feel prepend_lang_id would be a better name for it, I can go ahead and change.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -81,8 +84,31 @@ def __init__(self, cfg: MTEncDecModelConfig, trainer: Trainer = None):
...
-        # After this call, the model will have  self.source_processor and self.target_processor objects
-        self.setup_pre_and_post_processing_utils(source_lang=self.src_language, target_lang=self.tgt_language)
+        if self.multilingual:
+            if isinstance(self.src_language, ListConfig) and isinstance(self.tgt_language, ListConfig):
+                raise ValueError(
+                    "cfg.src_language and cfg.tgt_language cannot both be lists. We only support many-to-one or one-to-many multilingual models."
+                )
+            elif isinstance(self.src_language, ListConfig):
+                self.setup_pre_and_post_processing_utils(
+                    source_lang=self.src_language[0], target_lang=self.tgt_language
```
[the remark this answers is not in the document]
aklife97: Currently, we are assuming that languages are from same families. As a result, an assumption is that they'll have same pre/post processing. We can change this to incorporate different processors if we plan to broaden this assumption.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -433,21 +479,50 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
-            )
-            dataset.batchify(self.encoder_tokenizer, self.decoder_tokenizer)
+            if not self.multilingual:
+                src_file_list = [cfg.src_file_name]
+                tgt_file_list = [cfg.tgt_file_name]
+            else:
+                src_file_list = cfg.src_file_name
+                tgt_file_list = cfg.tgt_file_name
+
+            if len(src_file_list) != len(tgt_file_list):
```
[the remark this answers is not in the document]
aklife97: This will be true as the user is effectively passing multiple parallel datasets. That is, in the case of en->de/fr, the user needs to pass en->de files and en->fr files. Making it 2 src files and 2 tgt files.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
aklife97: Yupp, sure! I can add a random sampler in the next commit. Though, we can keep adding more and more samplers as we keep on needing them. We can simply create them by following the same syntax as current samplers i.e. creating a static method that takes datasets list and kwargs as input.

aklife97:
> Overall this is looking really nice.
> 
> Could you help to explain further how we're getting multi-lingual models (many to one and one to many) by adding multiple datasets and a language identifier token?

Currently, we are only supporting a full-parameter sharing multilingual model. As a result, we use the same MTEncDecModel and feed it data from multiple languages. A multilingual model just ingests data from multiple languages and learns to translate it (much like the standard translation model). Though, in cases like one-to-many, it may be unclear for the model as to what language the user wants to translate to. In order to solve this, we use a language identifier token that guides the translation into the correct language. So in the case of en->de/fr, we add the language identifier (<de> or <fr>) token to the beginning of every English source sentence to tell the model which language the output is going to be in. 

In the future, if we want to create models that allow selective parameter sharing, we can simply create a new model class that accepts configurations from the user telling which parts of the model to share. So we can have the same decoders but different encoder, shared self-attention layers but different FFNs, so on and so forth. The data, in that case, can still be sampled in the same way it is being done currently.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
aklife97: Addressed this in the update.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
```diff
@@ -430,3 +440,109 @@ def __iter__(self):
 
     def __len__(self):
         return self.metadata['num_batches']
+
+
+class MultilingualTranslationDataset(IterableDataset):
```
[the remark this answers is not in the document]
aklife97: Addressed this. Its called ConcatTranslationDataset now.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
aklife97: Added this in the update.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -398,33 +428,49 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
+                        shard_strategy=cfg.get("shard_strategy", "scatter"),
+                        global_rank=self.global_rank,
+                        world_size=self.world_size,
+                        reverse_lang_direction=cfg.get("reverse_lang_direction", False),
+                        prepend_id=self.multilingual_ids[idx],
+                    )
+                    datasets.append(dataset)
+
+                if self.multilingual:
+                    dataset = ConcatTranslationDataset(datasets)
```
MaximumEntropy, line 470: You probably need to pass other args to `ConcatTranslationDataset` from the config like sampling strategy, temperature values etc right?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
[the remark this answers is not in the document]
MaximumEntropy: Ah right, my bad

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -433,21 +479,50 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
+                    cache_data_per_node=cfg.get("cache_data_per_node", False),
+                    use_cache=cfg.get("use_cache", False),
+                    reverse_lang_direction=cfg.get("reverse_lang_direction", False),
+                    prepend_id=self.multilingual_ids[idx],
+                )
+                dataset.batchify(self.encoder_tokenizer, self.decoder_tokenizer)
+                datasets.append(dataset)
+
+            if self.multilingual:
+                dataset = ConcatTranslationDataset(datasets, cfg.shuffle)
```
MaximumEntropy, line 515: Same here to specify other args to the constructor.

aklife97:
The previous few commits bring the following changes to the PR:
* Fix `TarredTranslationDataset` length. Earlier, this length was not reduced based on the world size
* MultiGPU / Multi-worker fixes (especially when provided datasets are of kind `map`)
* Fixes for changes requested earlier (like the addition of `ConcatTranslationDataset` params to config)

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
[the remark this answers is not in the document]
MaximumEntropy: Sounds good.

nemo/collections/nlp/data/machine_translation/machine_translation_dataset.py
MaximumEntropy: Thanks for adding this :)

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -657,6 +701,11 @@ def train_tokenizers(
             if isinstance(decoder_special_tokens, dict):
                 decoder_special_tokens = list(decoder_special_tokens.values())
 
+        if multilingual and encoder_tokenizer_name != 'sentencepiece':
+            raise NotImplementedError(
+                f"Currently we only support training setencepiece tokenizer for multilingual model."
```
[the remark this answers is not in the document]
MaximumEntropy: Yeah, it looks like there is only a hacky way which is to just repeat the special token like 1-10k times somewhere in the file.

okuchaiev (approved):
LGTM, Thanks!

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -389,6 +415,10 @@ def setup_test_data(self, test_data_config: Optional[DictConfig]):
 
     def _setup_dataloader_from_config(self, cfg: DictConfig):
         if cfg.get("load_from_cached_dataset", False):
+            if self.multilingual:
+                raise ValueError(
+                    "Multilingual model does not support dataset caching, it is recommended to use TarredDataset instead."
```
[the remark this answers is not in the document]
aklife97: Done!

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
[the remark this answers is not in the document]
aklife97: Made the appropriate changes!

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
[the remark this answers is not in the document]
aklife97: Yes, and that does not necessarily require multilingual support from the model end. A user can simply create a concatenated file with multiple language sentences and train a multilingual model. So I guess it makes sense to stick to only SPT for now.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: Added these in the config

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: Added!

MaximumEntropy (approved):
LGTM. Thanks for the great PR!

PR #2168, opened 2021-05-06
[DOCS] NLP Model parallel, NMT multi-val, CORE register artifacts 

ericharper:


ericharper:
@titu1994 Could you review the register_artifact docs?

@MaximumEntropy Could you review NLP/NMT docs?

docs/source/core/core.rst
```diff
@@ -556,6 +556,51 @@ checkpoint to be restored. For these models, the user will have to override ``lo
...
+The model config that is packaged with the .nemo file will be updated according to the ``config_path`` key.
+In the above example, the model config will have 
+
+.. code-block:: YAML
+
+    encoder_tokenizer:
+        ...
+        tokenizer_model: nemo:4978b28103264263a03439aaa6560e5e_tokenizer.model
+
+``src`` is the path to the artifact and the basename of the path will be used when packaging the artifcat in the .nemo file.
```
titu1994, line 591:
```suggestion
``src`` is the path to the artifact and the base-name of the path will be used when packaging the artifact in the .nemo file.
```

docs/source/core/core.rst
```diff
@@ -556,6 +556,51 @@ checkpoint to be restored. For these models, the user will have to override ``lo
...
+Each artifact will have a hash prepended to the basename of ``src`` in the .nemo file. This is to prevent collisions with basenames.
```
titu1994, line 592:
```suggestion
Each artifact will have a hash prepended to the basename of ``src`` in the .nemo file. This is to prevent collisions with base-names that are identical (say when there are two or more tokenizers, both called `tokenizer.model`).
```

titu1994 (commented):
Registering artifacts section looks great

docs/source/nlp/megatron_finetuning.rst
```diff
@@ -53,6 +53,159 @@ a large biomedical text corpus, which achieves better performance in biomedical
...
+.. code-block:: bash
+
+    iter_0080000/
+    ├── mp_rank_00
+    │   └── model_optim_rng.pt
+    └── mp_rank_01
+        └── model_optim_rng.pt
+
+
+To start training from a Megatron-LM checkpoint, simply pass the path to the Megatron-LM checkpoint
```
ekmb, line 93: do you mean `fine-tuning` here?

docs/source/nlp/machine_translation.rst
```diff
@@ -414,6 +414,27 @@ the checkpoints that had the top 5 (by default) sacreBLEU scores.
 
 At the end of training, a ``.nemo`` file is written to the result directory which allows to run inference on a test set.
 
+Multi-Validation
```
MaximumEntropy, line 417: Although its kind of obvious from the name, it's probably best to state that this works only with `model.validation_ds`.

MaximumEntropy (dismissed):
LGTM, thanks!

docs/source/nlp/megatron_finetuning.rst
ericharper: Changed it.

docs/source/core/core.rst
ericharper: Done.

docs/source/core/core.rst
ericharper: Done.

docs/source/nlp/machine_translation.rst
ericharper: Updated it.

okuchaiev (approved):
lgtm

PR #2175, opened 2021-05-07
Enabled passing kwargs to export()

borisfom:
Also, removed previous (non–functional) extension for dicts in export - this schema matches what onnx.export expects.
 

okuchaiev (approved):
exortable changes look good to me. @blisc what about tts changes?

borisfom:
> exortable changes look good to me. @blisc what about tts changes?

tts changes are to conform to new extended syntax accepted by onnx.export (optional Dict as the last input tuple element, representing kwargs) instead of former non-standard extended syntax I briefly introduced in previous revision (accepting a Dict instead of a tuple as inputs, not really used except in this test example).

tests/collections/tts/test_waveglow.py
```diff
@@ -54,7 +54,10 @@
 def input_example(sz):
```
blisc, line 54: Can we undo the changes to this file?

borisfom: Why, it would fail to compile then? See my comment above.

PR #2166, opened 2021-05-06
First Revision of TTS Docs and Notebooks Update for 1.0

blisc:

PR #2177, opened 2021-05-07
[docs] variable name error in text to sound example in docs

tarun-bisht:
variable name `spec_gen` was used for parsing text and creating spectogram, which should be `spec_generator`


blisc:
Thanks for noticing this bug. I'll fix this in #2166

blisc:
Oh, it looks like someone already fixed it. We will fix this in the upcoming 1.0 release.

PR #2187, opened 2021-05-08
SDE updates

vsl9:
- added support for external vocabularies
- UI improvements
- refactored code


tools/speech_data_explorer/data_explorer.py
```diff
@@ -81,6 +81,7 @@ def parse_args():
     parser.add_argument(
         'manifest', help='path to JSON manifest file',
     )
+    parser.add_argument('--vocab', help='optional vocabulary to highlight OOV words')
```
titu1994, line 84: Let's note that --vocab points to a tab seperated file or a single word per line

PR #2190, opened 2021-05-10
Fix nodesplitter of webdataset.

VahidooX:
Set the nodessplitter param of webdataset to None as the default is True in their latest code.


MaximumEntropy (dismissed):
LGTM, thanks!

MaximumEntropy:
LGTM.

PR #2194, opened 2021-05-11
NMT fixes

okuchaiev:
* Allow models such as facebook/bart-large in NMT encoder
* Fixes in nmt webapp


MaximumEntropy (approved):
LGTM. Thanks!

PR #2196, opened 2021-05-11
Store mappings file in .nemo for FS2 model

redoctopus:
Update the model code to save the mappings file as an artifact in the checkpoint file.

### DIFF
--- a/nemo/collections/tts/models/fastspeech2.py
+++ b/nemo/collections/tts/models/fastspeech2.py
@@ -75,7 +75,12 @@ class FastSpeech2Model(SpectrogramGenerator):
 
         # Parser and mappings are used for inference only.
         self.parser = parsers.make_parser(name='en')
-        with open(cfg.mappings_filepath, 'r') as f:
+        if 'mappings_filepath' in cfg:
+            mappings_filepath = cfg.get('mappings_filepath')
+        else:
+            mappings_filepath = 'mappings.json'
+        mappings_filepath = self.register_artifact('mappings_filepath', mappings_filepath)
+        with open(mappings_filepath, 'r') as f:
             mappings = json.load(f)
             self.word2phones = mappings['word2phones']
             self.phone2idx = mappings['phone2idx']


nemo/collections/tts/models/fastspeech2.py
```diff
@@ -75,7 +75,12 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
 
         # Parser and mappings are used for inference only.
         self.parser = parsers.make_parser(name='en')
-        with open(cfg.mappings_filepath, 'r') as f:
+        if 'mappings_filepath' in cfg:
+            mappings_filepath = cfg.get('mappings_filepath')
+        else:
+            mappings_filepath = 'mappings.json'
+        mappings_filepath = self.register_artifact('mappings_filepath', mappings_filepath)
+        with open(mappings_filepath, 'r') as f:
```
titu1994, line 83: Extract this step into another function and use the ptl decorator `rank_zero_only` so that during DDP only zero rank writes it (file corruption otherwie).

nemo/collections/tts/models/fastspeech2.py
```diff
@@ -75,7 +75,12 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
 
         # Parser and mappings are used for inference only.
         self.parser = parsers.make_parser(name='en')
-        with open(cfg.mappings_filepath, 'r') as f:
+        if 'mappings_filepath' in cfg:
+            mappings_filepath = cfg.get('mappings_filepath')
+        else:
+            mappings_filepath = 'mappings.json'
+        mappings_filepath = self.register_artifact('mappings_filepath', mappings_filepath)
```
titu1994, line 82: This line will fail if `mapping.json` doesnt exist in the current working directory.

redoctopus: I believe this should never happen, as the filepath should always be passed in by the config at training, and always restored from the .nemo when performing inference.

redoctopus: Will add an error case to the else for clarity to the user

blisc (approved):

PR #2197, opened 2021-05-11
Transformer final norm preln

fayejf:
Fix/Add final layer norm for pre_ln transformer encoder and decoder


nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -168,7 +172,12 @@ def forward(
         cached_mems_list = [memory_states]
 
         for i, layer in enumerate(self.layers):
-            decoder_states = layer(decoder_states, decoder_attn_mask, memory_states, encoder_states, encoder_attn_mask)
+            if i == len(self.layers) - 1:
```
MaximumEntropy, line 175: I think doing it this way will add an extra element to `cached_mems_list` so it will have len(self.layers) + 1 elements in the list. I think the way to do it should be to iterate until len(self.layers) - 1. Then check if i == len(self.layers) - 2 and then do the last transformer block + LN and then add that to cached mems list. My guess is this is why the export test is failing as well?

MaximumEntropy (changes requested):
Thanks for the PR! I had a question regarding the way you add things to `cached_mems_list`.

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
fayejf: hmm yes looks like it gonna have extra hidden states mem from LN. Are you suggesting that we should make **last transformer block + LN** as one component/layer?

MaximumEntropy: I think you can still keep the layers separate like you've done right now, but just apply last the transformer block + LN before adding to the `cached_mems_list`.

fayejf: Got it. On it.

VahidooX (changes requested):

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -138,6 +138,10 @@ def __init__(
             pre_ln,
         )
         self.layers = nn.ModuleList([copy.deepcopy(layer) for _ in range(num_layers)])
+
+        if pre_ln:
+            self.layers.append(nn.LayerNorm(hidden_size, eps=1e-5))
```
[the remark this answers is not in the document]
MaximumEntropy: Yeah this seems like a good idea to avoid adding if statements in the loop below. You can rename `self.layers` to `self.transformer_layers`. Also for backward compatibility for models without this final layernorm, can you please add a `pre_ln_final_form=False` default to the encoder/decoder class? Does that seem like a reasonable solution? @okuchaiev.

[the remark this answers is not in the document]
VahidooX: Changing the name from self.layers to self.transformer_layers may break older checkpoints?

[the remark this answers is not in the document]
MaximumEntropy: Ah yes, my bad.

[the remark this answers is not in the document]
fayejf: > is there a reason eps is set here?

@VahidooX  Authors didn't mention eps in final norm in their paper. Here just follow all the layer norms in blocks.

Thanks for the suggestions. Updated!

nemo/collections/nlp/modules/common/transformer/transformer.py
```diff
@@ -84,11 +85,13 @@ def __init__(
         hidden_act: str = 'relu',
         mask_future: bool = False,
         pre_ln: bool = False,
+        pre_ln_final_layer_norm: bool = True,
```
VahidooX, line 88: Default should be False in order to be able to load the old checkpoints. When we load a model from nemo file, the config file is read form there. You may add this param to the latest config files with value of True so that new models get trained with last layer norm enbaled.

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -124,8 +124,16 @@ def __init__(
         ffn_dropout: float = 0.0,
         hidden_act: str = "relu",
         pre_ln: bool = False,
+        pre_ln_final_layer_norm: bool = True,
     ):
         super().__init__()
+        self.pre_ln = pre_ln
```
VahidooX, line 130: do you need to store these values? you can check the final_layer_norm in the forward to see if it None or not.

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -172,6 +180,11 @@ def forward(
             memory_states = self._get_memory_states(decoder_states, decoder_mems_list, i + 1)
             cached_mems_list.append(memory_states)
 
+        if self.pre_ln and self.pre_ln_final_layer_norm:
```
VahidooX, line 183: You may check it like this:
```
if self.final_layer_norm is not None:
            decoder_states = self.final_layer_norm(decoder_states)
            memory_states = self._get_memory_states(decoder_states, encoder_mems_list, i + 1)
            cached_mems_list.append(memory_states)
```

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -172,6 +180,11 @@ def forward(
...
+            decoder_states = final_layer_norm(decoder_states)
```
VahidooX, line 184: shouldn't it be self.final_layer_norm?

VahidooX (changes requested):

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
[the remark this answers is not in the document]
VahidooX: It is fine as the default of Pytorch is also 1e-5 and they would never change such defaults. You may keep it.

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
fayejf: ohhh good catch...

nemo/collections/nlp/modules/common/transformer/transformer.py
fayejf: I think the default should be pre_ln_final_layer_norm=True because it's the correct way. sandeep will send a patch for NMT setting it to False if the attribute is not in the config

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
fayejf: Yes I was overthinking maybe we gonna add more norm methods. yeah let me simply it

nemo/collections/nlp/modules/common/transformer/transformer.py
VahidooX: I see, then it is fine if Sandeep plan to fix is with another patch. Just a heads up that the current code can not load previous checkpoints which do not have this last layer norm.

VahidooX (approved):
LGTM!

nemo/collections/nlp/modules/common/transformer/transformer.py
okuchaiev: please keep default to True

fayejf: @okuchaiev  Yes. Sandeep has patched it in this PR.

VahidooX (approved):

MaximumEntropy (approved):
LGTM, Thanks :)

PR #2201, opened 2021-05-12
Fix bug in SpecCutout

rbracco:
[Colab notebook demonstrating bug and fix](https://colab.research.google.com/drive/1Tksb-RXAeLSkBW72KDgm1DYUaPAnHXBx?usp=sharing)

There is a bug in the NeMo implementation of SpecCutout that causes the dimensions of masks to be transposed so that `rect_freq` is the size of the mask in the time dimension and `rect_time` is the size of the mask in the frequency dimension. This also causes the placement of the masks to be limited to the lowest frequency bins.

image

Output after fix:
image

### DIFF
--- a/nemo/collections/asr/parts/spectr_augment.py
+++ b/nemo/collections/asr/parts/spectr_augment.py
@@ -55,7 +55,7 @@ class SpecAugment(nn.Module):
             self.adaptive_temporal_width = False
         else:
             if time_width > 1.0 or time_width < 0.0:
-                raise ValueError('If `time_width` is a float value, must be in range [0, 1]')
+                raise ValueError("If `time_width` is a float value, must be in range [0, 1]")
 
             self.adaptive_temporal_width = True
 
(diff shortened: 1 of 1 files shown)


VahidooX:
Thanks Robert for catching the bug! It looks like it has been around for a long time.
Why do you believe that "it causes the placement of the masks to be limited to the lowest frequency bins"?
The starting point for the frequency masking looks correct to me and can be any frequency. The only problem is that maximum length in frequency dimension can be 100 instead of 47 (these are the most common values used with this augmentation technique, F=47 and T=100).

titu1994:
This is an incredible find, thank you @rbracco ! I've verified that your colab (also, very nice way to show the correctness of your fix!) is indeed correct. 

PR approved.

titu1994 (approved):

rbracco:
Thank you both for the quick review and feedback. 

> Thanks Robert for catching the bug! It looks like it has been around for a long time.
> Why do you believe that "it causes the placement of the masks to be limited to the lowest frequency bins"?
> The starting point for the frequency masking looks correct to me and can be any frequency. The only problem is that maximum length in frequency dimension was set to 100 instead of 47 (these are the most common values used with this augmentation technique, F=47 and T=100).

Sorry, that issue only occurs if you have a high F and low T and it's actually the upper frequency bins, but the change fixes the issue anyway.

Also, just a heads up `rect_x` here refers to the frequency mask but the frequency is across the y_axis. Since the mask is applied with `x[idx, rect_x : rect_x + w_x, rect_y : rect_y + w_y] = 0.0` and the 2nd dimension is actually the height/feature dimension, and 3rd is width/time so since x refers to the freq dimension in both places and y refers to the time dimension in both, it still works, but if you want to do a quick name change before merging it may make it more clear. Cheers.

titu1994:
A name change would be good, to reflect which axis they effect. Rather than x and y I'd rather just deference them to width and height too if possible.

blisc (approved):

PR #2206, opened 2021-05-13
Set seed before generating random tensors in NMT test

MaximumEntropy:


titu1994 (approved):
Good seed choice, LGTM

PR #2207, opened 2021-05-14
ASR patches for v1.0.0

titu1994:
# Bugfixes
- Correct WER calculation for RNNT by resetting the metric after each computation.
- Prevent numba from performing a sync across PT cuda stream simply to initialize a cost vector.
- Correctly update the trainer value using `set_trainer`
- Adds .nemo and .ckpt files to dockerignore

# Additions
- Adds RNNT global flags to enable detached gradient normalization of encoder, decoder and joint models
- Adds `weights_init_scale` and `hidden_hidden_bias_scale` values (default to original values) to attempt reduction in variance between runs
- Add `tds_uniform` and `tds_normal` initialization modes for Jasper based models.

### DIFF
--- a/nemo/collections/asr/models/rnnt_models.py
+++ b/nemo/collections/asr/models/rnnt_models.py
@@ -609,6 +609,7 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
             if (sample_id + 1) % log_every_n_steps == 0:
                 self.wer.update(encoded, encoded_len, transcript, transcript_len)
                 _, scores, words = self.wer.compute()
+                self.wer.reset()
                 tensorboard_logs.update({'training_batch_wer': scores.float() / words})
 
         else:
@@ -668,6 +669,7 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
 
             self.wer.update(encoded, encoded_len, transcript, transcript_len)
             wer, wer_num, wer_denom = self.wer.compute()
+            self.wer.reset()
 
             tensorboard_logs['val_wer_num'] = wer_num
             tensorboard_logs['val_wer_denom'] = wer_denom
--- a/nemo/collections/asr/models/rnnt_models.py
+++ b/nemo/collections/asr/models/rnnt_models.py
@@ -112,9 +112,38 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
             self.joint.set_loss(self.loss)
             self.joint.set_wer(self.wer)
 
-        self._setup_rnnt_optional_settings()
+        self.setup_optim_normalization()
 
-    def _setup_rnnt_optional_settings(self):
+    def setup_optim_normalization(self):
+        """
+        Helper method to setup normalization of certain parts of the model prior to the optimization step.
+
+        Supported pre-optimization normalizations are as follows:
+
+        .. code-block:: yaml
+
+            # Variation Noise injection
+            model:
+                variational_noise:
+                    std: 0.0
+                    start_step: 0
+
+            # Joint - Length normalization
+            model:
+                normalize_joint_txu: false
+
+            # Encoder Network - gradient normalization
+            model:
+                normalize_encoder_norm: false
+
+            # Decoder / Prediction Network - gradient normalization
+            model:
+                normalize_decoder_norm: false
+
+            # Joint - gradient normalization
+            model:
+                normalize_joint_norm: false
+        """
         # setting up the variational noise for the decoder
         if hasattr(self.cfg, 'variational_noise'):
             self._optim_variational_noise_std = self.cfg['variational_noise'].get('std', 0)
@@ -123,31 +152,18 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
             self._optim_variational_noise_std = 0
             self._optim_variational_noise_start = 0
 
-        # Setup normalized gradients for model
-        if 'normalize_joint' in self.cfg:
-            self._optim_normalize_joint = self.cfg.normalize_joint
-            self._optim_normalize_txu = None
-        else:
-            self._optim_normalize_joint = False
-            self._optim_normalize_txu = None
+        # Setup normalized gradients for model joint by T x U scaling factor (joint length normalization)
+        self._optim_normalize_joint_txu = self.cfg.get('normalize_joint_txu', False)
+        self._optim_normalize_txu = None
 
         # Setup normalized encoder norm for model
-        if 'normalize_encoder_norm' in self.cfg:
-            self._optim_normalize_encoder_norm = self.cfg.normalize_encoder_norm
-        else:
-            self._optim_normalize_encoder_norm = False
+        self._optim_normalize_encoder_norm = self.cfg.get('normalize_encoder_norm', False)
 
         # Setup normalized decoder norm for model
-        if 'normalize_decoder_norm' in self.cfg:
-            self._optim_normalize_decoder_norm = self.cfg.normalize_decoder_norm
-        else:
-            self._optim_normalize_decoder_norm = False
+        self._optim_normalize_decoder_norm = self.cfg.get('normalize_decoder_norm', False)
 
         # Setup normalized joint norm for model
-        if 'normalize_joint_norm' in self.cfg:
-            self._optim_normalize_joint_norm = self.cfg.normalize_joint_norm
-        else:
-            self._optim_normalize_joint_norm = False
+        self._optim_normalize_joint_norm = self.cfg.get('normalize_joint_norm', False)
 
     def extract_rnnt_loss_cfg(self, cfg: Optional[DictConfig]):
         """
@@ -638,7 +654,7 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
         self.log_dict(tensorboard_logs)
 
         # Preserve batch acoustic model T and language model U parameters if normalizing
-        if self._optim_normalize_joint:
+        if self._optim_normalize_joint_txu:
             self._optim_normalize_txu = [encoded_len.max(), transcript_len.max()]
 
         return {'loss': loss_value}
@@ -779,7 +795,7 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
                     )
                     param.grad.data.add_(noise)
 
-        if self._optim_normalize_joint:
+        if self._optim_normalize_joint_txu:
             T, U = self._optim_normalize_txu
             if T is not None and U is not None:
                 for param_name, param in self.encoder.named_parameters():
--- a/nemo/collections/asr/models/rnnt_models.py
+++ b/nemo/collections/asr/models/rnnt_models.py
@@ -112,6 +112,9 @@ class EncDecRNNTModel(ASRModel, ASRModuleMixin):
             self.joint.set_loss(self.loss)
             self.joint.set_wer(self.wer)
 
+        self._setup_rnnt_optional_settings()
+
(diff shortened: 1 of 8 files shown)


titu1994:
@okuchaiev fyi regarding the dockerfile ignore and correctly updating the self.trainer inside set_trainer.

nemo/collections/asr/parts/jasper.py
```diff
@@ -36,6 +37,28 @@
 jasper_activations = {"hardtanh": nn.Hardtanh, "relu": nn.ReLU, "selu": nn.SELU, "swish": Swish}
 
 
+def tds_uniform_(tensor, mode='fan_in'):
+    """
```
VahidooX, line 41: Should they get dropped?

nemo/collections/asr/parts/jasper.py
```diff
@@ -36,6 +37,28 @@
...
+    fan = _calculate_correct_fan(tensor, mode)
+    gain = 2.0
+    std = gain / math.sqrt(fan)  # sqrt(4.0 / fan_in)
+    bound = std  # Calculate uniform bounds from standard deviation
+    with torch.no_grad():
+        return tensor.uniform_(-bound, bound)
+
+
+def tds_normal_(tensor, mode='fan_in'):
+    """
```
VahidooX, line 52: Should they get dropped?

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -120,6 +123,32 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             self._optim_variational_noise_std = 0
             self._optim_variational_noise_start = 0
 
+        # Setup normalized gradients for model
+        if 'normalize_joint' in self.cfg:
```
VahidooX, line 127: Can they be done like this?
self._normalize_joint = self.cfg.get(normalize_joint, False)

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -120,6 +123,32 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+            self._optim_normalize_joint = self.cfg.normalize_joint
+            self._optim_normalize_txu = None
+        else:
+            self._optim_normalize_joint = False
+            self._optim_normalize_txu = None
+
+        # Setup normalized encoder norm for model
+        if 'normalize_encoder_norm' in self.cfg:
```
VahidooX, line 135: These four parameters, if they are similar and related, how about adding a new section like optim_normalization and add them under that? Just a suggestion to organize them. It can also make the names smaller too.

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -120,6 +123,32 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+            self._optim_normalize_encoder_norm = self.cfg.normalize_encoder_norm
+        else:
+            self._optim_normalize_encoder_norm = False
+
+        # Setup normalized decoder norm for model
+        if 'normalize_decoder_norm' in self.cfg:
```
VahidooX, line 141: Another suggestion is that as all have _norm except one, you may drop that but make the normalize_joint to normalize_joint_txu?

nemo/collections/asr/modules/rnnt.py
```diff
@@ -245,7 +253,18 @@ def predict(
...
 
-    def _predict(self, vocab_size, pred_n_hidden, pred_rnn_layers, forget_gate_bias, t_max, norm, dropout):
+    def _predict(
+        self,
+        vocab_size,
+        pred_n_hidden,
+        pred_rnn_layers,
+        forget_gate_bias,
+        t_max,
+        norm,
```
VahidooX, line 263: Descriptions of the newe params are missing.

nemo/collections/asr/parts/jasper.py
```diff
@@ -36,6 +37,28 @@
...
+    """
+    fan = _calculate_correct_fan(tensor, mode)
+    gain = 2.0
```
VahidooX, line 55: Can these initialization be useful for others? If yes, you may move them somewhere where is shared?

nemo/collections/asr/parts/numba/rnnt_loss/rnnt.py
```diff
@@ -180,7 +180,7 @@ def rnnt_loss_gpu(
     acts, acts_shape = rnnt_helper.flatten_tensor(acts)
 
     ### REPRESENT THE CUDA ARRAY INTERFACE OF COSTS VECTOR ###
-    costs_repr = cuda.as_cuda_array(costs)  # NO COPY OF DATA, JUST CHANGE REPRESENTATION
+    costs_repr = cuda.as_cuda_array(costs, sync=False)  # NO COPY OF DATA, JUST CHANGE REPRESENTATION
```
VahidooX, line 183: Is it OK to not sync it?

nemo/collections/common/parts/rnn.py
```diff
@@ -188,10 +204,15 @@ def __init__(
                     bias.data[hidden_size : 2 * hidden_size].fill_(forget_gate_bias)
                 if "bias_hh" in name:
                     bias = getattr(self.lstm, name)
-                    bias.data[hidden_size : 2 * hidden_size].fill_(0)
+                    # bias.data[hidden_size : 2 * hidden_size].fill_(0)
```
VahidooX, line 207: Drop the comment?

nemo/collections/common/parts/rnn.py
```diff
@@ -188,10 +204,15 @@ def __init__(
...
+                    bias.data[hidden_size : 2 * hidden_size] *= float(hidden_hidden_bias_scale)
 
         self.dropout = torch.nn.Dropout(dropout) if dropout else None
 
+        for name, v in self.named_parameters():
```
VahidooX, line 212: How about to check and if weights_init_scale is 1.0 then skip this part?

nemo/core/classes/modelPT.py
```diff
@@ -1177,6 +1177,7 @@ def set_trainer(self, trainer: Trainer):
         Args:
             trainer: PyTorch Lightning Trainer object.
         """
+        self.trainer = trainer
```
VahidooX, line 1180: Why do we keep both _trianer and trianer?

nemo/collections/common/parts/rnn.py
```diff
@@ -138,6 +152,8 @@ def __init__(
         dropout: Optional[float],
         forget_gate_bias: Optional[float],
         t_max: Optional[int] = None,
+        weights_init_scale: float = 1.0,
```
VahidooX, line 155: Descriptions of the new params are missing.

nemo/collections/common/parts/rnn.py
```diff
@@ -188,10 +204,15 @@ def __init__(
                     bias.data[hidden_size : 2 * hidden_size].fill_(forget_gate_bias)
                 if "bias_hh" in name:
                     bias = getattr(self.lstm, name)
-                    bias.data[hidden_size : 2 * hidden_size].fill_(0)
+                    # bias.data[hidden_size : 2 * hidden_size].fill_(0)
+                    bias.data[hidden_size : 2 * hidden_size] *= float(hidden_hidden_bias_scale)
```
VahidooX, line 208: How about to check and if hidden_hidden_bias_scale is 0.0 then skip this part?

VahidooX: Is it correct that you are multiplying by zero here as the default.?
 Just to make sure it is on purpose.

nemo/collections/asr/parts/jasper.py
VahidooX, line 41: Is there any references for these initialization techniques to mention here?

VahidooX (changes requested):

nemo/collections/asr/models/rnnt_models.py
titu1994: Yes we can rename this function to `_setup_optim_normalization`

nemo/collections/asr/models/rnnt_models.py
titu1994: Good point. We also need docstrings here.

nemo/collections/asr/modules/rnnt.py
titu1994: Good catch

nemo/collections/asr/parts/jasper.py
titu1994: Ah i forgot to fill this.

nemo/collections/asr/parts/jasper.py
titu1994: Hmm they are somewhat specific to Jasper based models right now (ill add a reference to the paper). We can move them to commons if anyone needs it in the future.

nemo/collections/asr/parts/numba/rnnt_loss/rnnt.py
titu1994: Yup tested this. Was the cause of a minor slow down since it would wait for the default pytorch stream to execute and sync before continuing, which was unnecessary since these kernels will just queue more stuff to that same stream.

nemo/collections/common/parts/rnn.py
titu1994: Good catch.

nemo/collections/common/parts/rnn.py
titu1994: It's done just once at init, so its fast enough. The default here is matching what was done in general prior to this commit. Mathematically it is equivalent.

nemo/collections/common/parts/rnn.py
titu1994: Its done only once at init, should not cost much time.

nemo/core/classes/modelPT.py
titu1994: PTL 1.2.8+ refers to self.trainer in order to get values such as self.node_rank, self.global_rank etc. So it needs to be explicitly updated. The `_trainer` is for backward compat with older NeMo models.

nemo/collections/asr/parts/jasper.py
titu1994: Yup I forgot to add it in a hurry.

titu1994 (commented):
Addressed comments

nemo/core/classes/modelPT.py
VahidooX: In that case, why haven't we change our codes to use _trainer instead of trainer? It is just a find and replace?

nemo/collections/asr/models/rnnt_models.py
VahidooX: I meant adding a section to the config file like this:
model:
    optim_normalization:
         normalize_encoder_norm: true
         normalize_joint:false
    ...

the current form is also OK, just a suggestion to make it more organized as they are at least 4 and maybe more in future may be added.

VahidooX (approved):

nemo/core/classes/modelPT.py
titu1994: They both are same thing and are practically the same reference. Newer code should just use self.trainer directly.

nemo/collections/asr/models/rnnt_models.py
titu1994: Yes that can be another option. Other than variational noise, I expect the other 3 to be rarely used. They only really apply to very specific combination of model and optimizer so we might refactor them into their own section if they every become more useful in general case

PR #2209, opened 2021-05-14
fix the versioning name.

VahidooX:
Fix the versioning name to avoid creating the extra folder name "None"  when training is supposed to continue.


titu1994 (commented):
Doesn't this just dump the files into the project dir? @blisc any thoughts if this is correct ?

VahidooX:
> 
> 
> Doesn't this just dump the files into the project dir? @blisc any thoughts if this is correct ?

Actually I made this from "" to None last time to hope it fixes the w&b but it made the issue that now we have a folder name "None" and then there is another folder under that named "checkpoints" where checkpoints are getting stored under. They would not get dumped into the project folder but under a folder named checkpoints.

blisc:
Hmm I see, we need it to be None for future code but doing `Path(str(None))' returns "None" when we need it to be empty string

titu1994:
How about explicitly checking for "None" and then replacing it with ""

VahidooX:
> 
> 
> How about explicitly checking for "None" and then replacing it with ""

Did this this way. Now the version property in the config would be set to None instead of "".

titu1994:
This overall looks like a good solution, @blisc to confirm the approach is correct and merge if its fine.

blisc (approved):

PR #2216, opened 2021-05-15
Support symlinked files

piraka9011:
To preserve disk space on my machine, I generally use symlinks to create different splits of my data.
The tarred audio dataset script doesn't follow symlinks, so simply pass `dereference=True` to the `tarfile.open()` method. 

See more info in the python docs: https://docs.python.org/3/library/tarfile.html#tarfile.TarFile

### DIFF
--- a/scripts/speech_recognition/convert_to_tarred_audio_dataset.py
+++ b/scripts/speech_recognition/convert_to_tarred_audio_dataset.py
@@ -462,7 +462,7 @@ class ASRTarredDatasetBuilder:
         """Creates a tarball containing the audio files from `entries`.
         """
         new_entries = []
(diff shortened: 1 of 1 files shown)


titu1994:
Just a note - The manifest file can (and generally it should) have absolute paths to the dataset files when on a local machine, which the tarfile would then use.

titu1994 (approved):

piraka9011:
> Just a note - The manifest file can (and generally it should) have absolute paths to the dataset files when on a local machine, which the tarfile would then use.

Yup, I noticed this. However, I was splitting based on specific audio/label heuristics unrelated to the train/val/test split.
I'll probably modify my code to make reading these splits better.

PR #2219, opened 2021-05-15
Log audio file on loading errors

piraka9011:
Signed-off-by: Anas Abou Allaban <aabouallaban@pm.me>

It would be helpful to know offending audio files that fail to load if they are corrupt and/or not supported.

### DIFF
--- a/nemo/collections/asr/parts/segment.py
+++ b/nemo/collections/asr/parts/segment.py
@@ -42,6 +42,7 @@ import soundfile as sf
 from kaldiio.matio import read_kaldi
 from kaldiio.utils import open_like_kaldi
 from pydub import AudioSegment as Audio
+from pydub.exceptions import CouldntDecodeError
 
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
LGTM, thanks for the contribution !

titu1994:
Seems theres some formatting errors, could you run `python setup.py style --scope nemo --fix`

titu1994:
This PR will be merged into another PR (https://github.com/NVIDIA/NeMo/pull/2240) due to inactivity. One difference is that it will not raise the error, but instead simply log that the file was not loaded properly.

piraka9011:
Got it, sorry for the delay and thanks for the followup!

titu1994:
Thanks for the PR! Normally I'd wait, but this PR would have merge conflicts with that one so it would be easier to integrate into that one.

PR #2239, opened 2021-05-19
Alias Swish to SiLU 

titu1994:
# Changelog
- Alias `Swish` activation to `nn.SiLU` which is mathematically identical
- Update activations to inplace execution if possible for `ConvASREncoder`

### DIFF
--- a/nemo/collections/asr/parts/activations.py
+++ b/nemo/collections/asr/parts/activations.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import torch
 import torch.nn as nn
 
 __all__ = ['Swish']
--- a/nemo/collections/asr/parts/activations.py
+++ b/nemo/collections/asr/parts/activations.py
@@ -18,10 +18,8 @@ import torch.nn as nn
 __all__ = ['Swish']
 
 
-class Swish(nn.Module):
+class Swish(nn.SiLU):
     """
     Swish activation function introduced in 'https://arxiv.org/abs/1710.05941'
+    Mathematically identical to SiLU. See note in nn.SiLU for references.
     """
-
-    def forward(self, x):
-        return x * torch.sigmoid(x)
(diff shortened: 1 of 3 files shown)


nemo/collections/asr/parts/activations.py
```diff
@@ -18,10 +18,8 @@
 __all__ = ['Swish']
 
 
-class Swish(nn.Module):
+class Swish(nn.SiLU):
```
VahidooX, line 21: Why not drop this class and use nn.SiLU directly?
We may replace all nemo.Swish with nn.SiLU in the code?

VahidooX: In this way we may have the following for backward compatibility:
jasper_activations = {"hardtanh": nn.Hardtanh, "relu": nn.ReLU, "selu": nn.SELU, "swish": nn.SiLU, "silu": nn.SiLU}

VahidooX (changes requested):

nemo/collections/asr/parts/activations.py
```diff
@@ -18,10 +18,8 @@
...
     """
     Swish activation function introduced in 'https://arxiv.org/abs/1710.05941'
+    Mathematically identical to SiLU. See note in nn.SiLU for references.
```
VahidooX, line 24: There is an "import torch" on the top which is not used.

nemo/collections/asr/parts/activations.py
titu1994: Swish is mathematically the same, but its non-obvious at first glance that they are the same (I for one kept searching for swish in PT and never found it until I went through all the activations and found SiLU)..

nemo/collections/asr/parts/activations.py
titu1994: Fixed

VahidooX (approved):
LGTM!

blisc (approved):

PR #2240, opened 2021-05-19
ASR Refactoring

titu1994:
# Changelog
- Significant refactoring of `parts` in ASR to be more modular and systematic
- Move preprocessing steps of ASR text cleaning into `commons` collection.
- Add typing information to SpecAugment and SpecCutout
- Remove NLTK downloading on import and move it to class instantiation time
- Update docs to point to new paths for parts
- Cherry pick PR https://github.com/NVIDIA/NeMo/pull/2219 into this so that audio files that could not be loaded log a warning but do not immediately crash. 
- Make character based ASR models propagate their sample_rate and labels to dataloaders
- Update ASR training scripts to support easy finetuning via pretrained model name or .nemo file

# TODO
- [ ] Update all ASR configs in next PR to latest format


piraka9011:
1 question, is there a reason you don't want training to fail if the audio file could not be loaded?
This might hide underlying issues in the dataset instead of forcing the user to fix the issue with the audio files.

redoctopus (commented):
LGTM. Waiting for others to review also before approving.

We could keep the files that were moved to `parts/utils/` in `parts/`. They generally seem somewhat more "miscellaneous" than the other moved files, and I feel like we wouldn't lose much semantic meaning by having the top-level parts directory be for these files. But just an idle thought, I don't have a strong opinion on this.

titu1994:
@piraka9011 Main reason would be the case where more than one file fails - it would be problematic to await several minutes for data loaders to start up and then immediately crash after the first issue. If we log it (and hopefully it doesnt crash and it continues) there will be a continuous stream of error logs that will notify the user of which files are consistently problematic. Of course, this is under the assumption there is no fatal flaw and the data loader doesnt hard crash anyway.

titu1994:
I would agree on the utils thing @redoctopus - i pushed it into utils cause I saw like 6 files with _utils.py and I went - why not just move them under a utils folder then. Lets see, if theres a strong opinion on this from others we can just move them out again.

blisc (approved):

examples/asr/speech_to_text.py
```diff
@@ -62,16 +55,86 @@
...
+
+def restore_weights_if_required(model: EncDecCTCModel, cfg: OmegaConf):
+    if 'init_from_nemo_model' not in cfg and 'init_from_pretrained_model' not in cfg:
+        # model weights do not need to be restored
+        return
+
+    if 'init_from_nemo_model' in cfg and 'init_from_pretrained_model' in cfg:
+        raise ValueError("Cannot pass both `init_from_nemo_model` and `init_from_pretrained_model` to config!")
+
+    if 'init_from_nemo_model' in cfg and cfg.init_from_nemo_model is not None:
```
VahidooX, line 103: These two if blocks for init_from_nemo_model and init_from_pretrained_model looks exactly the same to me. let's make them one.

examples/asr/speech_to_text.py
```diff
@@ -62,16 +55,86 @@
...
+        with open_dict(cfg):
+            # Restore model
+            model_path = cfg.pop('init_from_nemo_model')
+            restored_model = EncDecCTCModel.restore_from(model_path, map_location='cpu', strict=True)
```
VahidooX, line 107: How about use model.__class__.restore_from or even model.restore_from so that you may use the same code for both CTC, BPE and RNNT?

examples/asr/speech_to_text.py
```diff
@@ -62,16 +55,86 @@
...
+            del restored_model
+
+    if 'init_from_pretrained_model' in cfg and cfg.init_from_pretrained_model is not None:
+        with open_dict(cfg):
+            # Restore model
+            model_name = cfg.pop('init_from_pretrained_model')
+            restored_model = EncDecCTCModel.from_pretrained(model_name, map_location='cpu', strict=True)
+
+            # Restore checkpoint into current model
+            model.load_state_dict(restored_model.state_dict(), strict=False)
```
VahidooX, line 122: If we just need the state of the model, can we add the feature to support PTL checkpoints too?
It happens a lot that the training is not done and we need to download the checkpoint, copy the tokenizer to our local system with the same exact location in order to convert a checkpoint to nemo file.

examples/asr/speech_to_text.py
```diff
@@ -62,16 +55,86 @@
...
 
 @hydra_runner(config_path="conf", config_name="config")
 def main(cfg):
     logging.info(f'Hydra config: {OmegaConf.to_yaml(cfg)}')
+
     trainer = pl.Trainer(**cfg.trainer)
     exp_manager(trainer, cfg.get("exp_manager", None))
     asr_model = EncDecCTCModel(cfg=cfg.model, trainer=trainer)
 
+    restore_weights_if_required(asr_model, cfg)
```
VahidooX, line 136: The word restore conveys the meaning that this model has some weights, it lost them and we want them back. Init may be a more appropriate word here.

examples/asr/speech_to_text_rnnt.py
```diff
@@ -79,15 +59,86 @@
...
+import pytorch_lightning as pl
+from omegaconf import OmegaConf, open_dict
+
+from nemo.collections.asr.models import EncDecRNNTModel
+from nemo.core.config import hydra_runner
+from nemo.utils import logging
+from nemo.utils.exp_manager import exp_manager
+
+
+def restore_weights_if_required(model: EncDecRNNTModel, cfg: OmegaConf):
```
VahidooX, line 99: We are duplicating mostly the same method 8 times in 4 scripts. Let's merge them as much as possible and put it into a common place to be used by all. Maintaining the same method scattered like this can be a burden.

examples/asr/speech_to_text_bpe.py
```diff
@@ -50,25 +50,87 @@
     exp_manager.wandb_logger_kwargs.name="<Name of experiment>" \
     exp_manager.wandb_logger_kwargs.project="<Name of project>"
 ```
+
+Finetune a model
```
VahidooX, line 54: Can we add a section on this topic to the documentations?

examples/asr/speech_to_text.py
```diff
@@ -62,16 +55,86 @@
...
+import pytorch_lightning as pl
+from omegaconf import OmegaConf, open_dict
+
+from nemo.collections.asr.models import EncDecCTCModel
+from nemo.core.config import hydra_runner
+from nemo.utils import logging
+from nemo.utils.exp_manager import exp_manager
+
+
+def restore_weights_if_required(model: EncDecCTCModel, cfg: OmegaConf):
```
VahidooX, line 95: This function runs every time. What happens in multiple runs to continue training?
It is going to download and load the weights every time. Let's add to check if if it is the first step otherwise ignore this weight loading.

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -17,8 +17,11 @@
 import torch
 import torch.nn as nn
 
+from nemo.core.classes import Typing, typecheck
```
VahidooX, line 20: The new organization of the parts looks very good. Nice!

VahidooX (changes requested):

titu1994 (commented):
Address comments

docs/source/asr/configs.rst
```diff
@@ -473,3 +473,56 @@ specify the tokenizer if you want to use sub-word encoding instead of character-
...
 config files and also :doc:`nemo.collections.asr.modules.ConformerEncoder<./api.html#nemo.collections.asr.modules.ConformerEncoder>`.
+
+
+Finetuning Configurations
+-------------------------
+
+All ASR scripts support easy fine tuning by partially/fully loading the pretrained weights from a checkpoint into the currently instantiated model. Pre-trained weights can be provided in multiple ways -
+
+1) Providing a path to a NeMo model (via ``init_from_nemo_model``)
+2) Providing a name of a pretrained NeMo model (which will be downloaded via the cloud) (via ``init_from_pretrained_model``)
```
VahidooX, line 484: How about giving a link to the list of the pretained models so that user understands what we mean by pretrained nemo models?

docs/source/asr/configs.rst
```diff
@@ -473,3 +473,56 @@ specify the tokenizer if you want to use sub-word encoding instead of character-
 
 The encoder section includes the details about the Conformer-CTC encoder architecture. You may find more information in the 
 config files and also :doc:`nemo.collections.asr.modules.ConformerEncoder<./api.html#nemo.collections.asr.modules.ConformerEncoder>`.
+
+
+Finetuning Configurations
+-------------------------
+
+All ASR scripts support easy fine tuning by partially/fully loading the pretrained weights from a checkpoint into the currently instantiated model. Pre-trained weights can be provided in multiple ways -
```
VahidooX, line 481: fine-tuning

VahidooX (approved):

jbalam-nv:
Looks like a better place for the preprocessing directory would be in asr/data.  I always found it difficult to find preprocessing code :), I think it would be much easier to find if it is in asr/data.

titu1994:
asr/data is meant to be only for **datasets**. The two ***_dataset.py** files are wrappers around said datasets so it is still valid. Parts is where everything else goes that doesnt belong to one of 5 main things - dataset/dataloader, models (which inherit ModelPT), neural modules, losses and metrics.

Furthermore that preprocessing is shared between asr and tts so best it goes into commons

titu1994:
Also, hmm we need a high level architecture diagram. Will put that in a todo list.

examples/asr/speech_to_label.py
```diff
@@ -102,20 +102,36 @@
...
 
 @hydra_runner(config_path="conf", config_name="matchboxnet_3x1x64_v1")
 def main(cfg):
+    logging.info(f'Hydra config: {OmegaConf.to_yaml(cfg)}')
+
     trainer = pl.Trainer(**cfg.trainer)
     exp_manager(trainer, cfg.get("exp_manager", None))
     asr_model = EncDecClassificationModel(cfg=cfg.model, trainer=trainer)
 
+    asr_model.maybe_init_from_pretrained_checkpoint(cfg)
```
VahidooX, line 133: How about adding a single line comment here what does it do? something like:
"Initializing the weights of the model from another model if any provided"

VahidooX (approved):

examples/asr/speech_to_label.py
titu1994: Sounds good

redoctopus (approved):

PR #2245, opened 2021-05-21
Support multiple models being instantiated in same execution scope

titu1994:
# Changelog
- Model's are now attached with a one off global unique id (guid) which is registered into the global AppState for maintanence.
- AppState tracks whether the model was restored via a nemo file or via constructor and monitors an ever growing list of model restoration paths (If the model constructor was used, its restoration path is None).
- During registration of artifacts, each model registers its own copy of artifacts (via self)
- During save to, the restoration path is retrieved from AppState and artifacts are copied from the restoration file. If restoration path is None, assume local path artifacts
- During restore from, the restoration path is first registered in an ever growing list. Then the model is constructed (which allocates a GUID and registers this GUID with the restoration path).

# Bugfix
- Fixes a bug which prevents a model constructed via constructor to call save_to multiple times.

### DIFF
--- a/nemo/utils/app_state.py
+++ b/nemo/utils/app_state.py
@@ -29,7 +29,7 @@ class ModelMetadataRegistry:
 class AppState(metaclass=Singleton):
     def __init__(self):
         # method call lock
-        self._lock = Lock()
+        self.__lock = Lock()
 
         # TODO: should we store global config in hydra_runner?
         self._app_cfg = None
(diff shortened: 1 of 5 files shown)


tests/core/test_save_restore.py
```diff
@@ -310,3 +311,138 @@ def test_mock_save_to_restore_from_with_target_class(self):
...
+            model2 = MockModel(cfg=cfg2.model, trainer=None)
+            model2 = model2.to('cpu')
+
+            assert model.temp_file == empty_file.name
+            assert model2.temp_file == empty_file2.name
+
+            # Save test
+            model_copy = self.__test_restore_elsewhere(model, map_location='cpu')
+            model2_copy = self.__test_restore_elsewhere(model2, map_location='cpu')
+            # assert filecmp.cmp(model.temp_file, model_copy.temp_file)
```
ericharper, line 342: Clean up comment?

tests/core/test_save_restore.py
```diff
@@ -310,3 +311,138 @@ def test_mock_save_to_restore_from_with_target_class(self):
...
+            model2 = MockModel(cfg=cfg2.model, trainer=None)
+            model2 = model2.to('cpu')
+
+            assert model.temp_file == empty_file.name
+            assert model2.temp_file == empty_file2.name
+
+            # Save test (inverted order)
+            model2_copy = self.__test_restore_elsewhere(model2, map_location='cpu')
+            model_copy = self.__test_restore_elsewhere(model, map_location='cpu')
+            # assert filecmp.cmp(model.temp_file, model_copy.temp_file)
```
ericharper, line 375: Clean up comment?

nemo/utils/app_state.py
```diff
@@ -12,11 +12,24 @@
...
+from threading import Lock
+from typing import Optional
+
 from nemo.utils.metaclasses import Singleton
 
 
+@dataclass()
+class ModelMetadataRegistry:
+    guid: str
+    gidx: int
```
ericharper, line 25: What's `gidx` used for?

ericharper (approved):
LGTM! Thanks!

Only found a couple of comments to be cleaned.

nemo/utils/app_state.py
titu1994: Unused as of now, just a global index of the model in the scope of AppState.

tests/core/test_save_restore.py
titu1994: Fixed

tests/core/test_save_restore.py
titu1994: Fixed

PR #2250, opened 2021-05-21
[BUGFIX] Only process tarfile artifacts when model was restored from tarfile

ericharper:
NLP models are having tarfile artifacts when calling save_to. This is causing tarfile artifacts to be processed even though the model was not restored from a tarfile.


titu1994 (changes requested):
This is incorrect - a model doesn't need be under restoration state while saving artifacts - in fact that's impossible. We can't save while we are being restored.

titu1994 (approved):
This works

PR #2251, opened 2021-05-21
Log average metrics for Multi-validation in NMT

aklife97:
Add averaged metrics to log when using multiple validation/test sets.


MaximumEntropy (approved):
LGTM!

PR #2269, opened 2021-05-27
Add support for Numba CUDA optimized SpecAugment

titu1994:
# Changelog
- Adds support for numba cuda kernel to perform SpecAugment when following criterion are met:
  - Numba version is >= 0.53.1
  - Input tensor device = CUDA
  - Lengths of the input tensors are provided (handled by NeMo, user's responsibility if called separately)
  - User provided flag `use_numba_spec_augment` is set to True (True by default)
- In such a scenario, if the above conditions are met, SpecAugment() will be replaced by SpecAugmentNumba()
- Users can disable this kernel by setting the flag `use_numba_spec_augment` to False.


jbalam-nv (approved):
Looks great. All good to go!

PR #2271, opened 2021-05-27
Add an NVIDIA DALI implementation of AudioToMelSpectrogramPreprocessor

jantonguirao:
Adds an option `use_dali` to `AudioToMelSpectrogramPreprocessor` to choose between the native torch implementation of the audio preprocessing pipeline or a DALI based implementation.

Adds unit tests to verify that the results produced by both implementations are equivalent.

Removes code related with previous attempts to integrate with NVIDIA DALI.


nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
...
+            m = dali.fn.transforms.scale(scale=[subsampling, 1], center=[0.5, 0])
+            x = dali.fn.reshape(x, rel_shape=[1, 1, -1], layout="HWC")  # Layout required by WarpAffine
+            x = dali.fn.warp_affine(
+                x, matrix=m, size=dali.fn.cat(nfeatures, out_len), interp_type=dali.types.INTERP_NN
+            )
+            x = dali.fn.reshape(x, rel_shape=[1, 1], layout="ft")
+        return x
+
+    def init_dali_pipeline(self, batch_size, device, device_id, num_threads=4):
+        pipe = dali.pipeline.Pipeline(
```
titu1994, line 516: Will 4 threads be sufficient for larger training scenarios (slurm clusters)?

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
...
+            )
+
+            # Spectrogram to Mel Spectrogram
+            spec = dali.fn.mel_filter_bank(
+                spec, sample_rate=self.sample_rate, nfilter=self.nfilt, freq_low=self.lowfreq, freq_high=self.highfreq,
+            )
+
+            # log features if required
+            if self.log:
+                eps = self.log_zero_guard_value_fn(torch.float32)
```
titu1994, line 562: This works under AMP too right?

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
...
+                elif self.normalize == "all_features":
+                    normalization_axes = [0, 1]
+                elif "fixed_mean" in self.normalize and "fixed_std" in self.normalize:
+                    raise ValueError("Normalization with fixed mean/stddev not yet supported.")
+                    # TODO: implement
+                else:
+                    raise ValueError(f"Unknown normalization type: {self.normalize}")
+
+                # Normalization
+                spec = dali.fn.normalize(spec, axes=normalization_axes, epsilon=1e-5 ** 2, ddof=1)
```
titu1994, line 593: 1e-10 eps is too low (maybe) for fp16? Or is this all processed under fp32 first and then finally cast to fp16?

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
...
+        device_id = None
+        if device_type_str == 'gpu':
+            device_id = int(device_str_toks[1])
+            cuda_stream = torch.cuda.current_stream(device=device)
+        else:
+            device_id = None
+            cuda_stream = None
+        batch_size = x.shape[0]
+        if (
+            self.dali_ctx.batch_size != batch_size
```
titu1994, line 620: So this will rebuild the entire pipeline each time batch size changes? Under DDP this should probably be fine, for single GPU might cause a bit of slowdown at the end of each training epoch?

tests/collections/asr/test_asr_modules.py
```diff
@@ -38,13 +52,178 @@ def test_AudioToMelSpectrogramPreprocessor1(self):
...
-                assert len1 == len2
-            assert res1.shape == res2.shape
-            diff = torch.mean(torch.abs(res1 - res2))
-            assert diff <= 1e-3
-            diff = torch.max(torch.abs(res1 - res2))
-            assert diff <= 1e-2
+            self.compare_features(res1, length1, res2, length2)
+
+    @pytest.mark.unit
+    def test_AudioToMelSpectrogramPreprocessor_DALI_spectrogram(self):
```
titu1994, line 58: Could you mark the test as optional (in case Dali is not installed by user) - similar to here https://github.com/NVIDIA/NeMo/blob/main/tests/collections/asr/test_asr_rnnt_encdec_model.py#L95-L97

tests/collections/asr/test_asr_modules.py
```diff
@@ -38,13 +52,178 @@ def test_AudioToMelSpectrogramPreprocessor1(self):
...
+            # Ensure that the two functions behave similarily
+            for _ in range(5):
+                input_signal = torch.randn(size=(4, 512))
+                length = torch.randint(low=161, high=500, size=[4])
+                res1, length1 = instance1(input_signal=input_signal, length=length)
+                res2, length2 = instance2(input_signal=input_signal, length=length)
+                self.compare_features(res1, length1, res2, length2)
+
+    @pytest.mark.unit
+    def test_AudioToMelSpectrogramPreprocessor_DALI_melfilterbank(self):
```
titu1994, line 109: Same as above

tests/collections/asr/test_asr_modules.py
```diff
@@ -38,13 +52,178 @@ def test_AudioToMelSpectrogramPreprocessor1(self):
...
+            # Ensure that the two functions behave similarily
+            for _ in range(5):
+                input_signal = torch.randn(size=(4, 512))
+                length = torch.randint(low=161, high=500, size=[4])
+                res1, length1 = instance1(input_signal=input_signal, length=length)
+                res2, length2 = instance2(input_signal=input_signal, length=length)
+                self.compare_features(res1, length1, res2, length2)
+
+    @pytest.mark.unit
+    def test_AudioToMelSpectrogramPreprocessor_DALI_log(self):
```
titu1994, line 150: Same as above

tests/collections/asr/test_asr_modules.py
```diff
@@ -38,13 +52,178 @@ def test_AudioToMelSpectrogramPreprocessor1(self):
...
+            # Ensure that the two functions behave similarily
+            for _ in range(5):
+                input_signal = torch.randn(size=(4, 512))
+                length = torch.randint(low=161, high=500, size=[4])
+                res1, length1 = instance1(input_signal=input_signal, length=length)
+                res2, length2 = instance2(input_signal=input_signal, length=length)
+                self.compare_features(res1, length1, res2, length2)
+
+    @pytest.mark.unit
+    def test_AudioToMelSpectrogramPreprocessor_DALI_normalization(self):
```
titu1994, line 187: Same as above

tests/collections/asr/test_asr_modules.py
```diff
@@ -38,13 +52,178 @@ def test_AudioToMelSpectrogramPreprocessor1(self):
...
+
+            # Ensure that the two functions behave similarily
+            for _ in range(5):
+                input_signal = torch.randn(size=(4, 512))
+                length = torch.randint(low=161, high=500, size=[4])
+                res1, length1 = instance1(input_signal=input_signal, length=length)
+                res2, length2 = instance2(input_signal=input_signal, length=length)
+                self.compare_features(res1, length1, res2, length2, eps_mean_err=1e-3, eps_max_err=0.5)
+
+    @pytest.mark.unit
```
titu1994, line 209: Same as above

titu1994 (changes requested):
Overall looks fantastic, much cleaner approach than before. One thing is that the update to the DALIOutput etc needs to be replicated in the EncDecRNNTModel as well.

Minor comments about docstrings for readability. 
Tests need to be import guarded (example given in comments). 

Also for user visibility, can you add `use_dali: False` to the configs inside ASR based models? If you'd like I can do that in a separate PR as well.

nemo/collections/asr/parts/preprocessing/features.py
jantonguirao: Those are 4 CPU data-processing threads per pipeline (1 pipeline per GPU instance). In a large training scenario you probably have more than one GPU and end up with ``n_gpus * 4`` threads only for DALI data processing. From our experience, going beyond 4 threads doesn't improve training performance. Also, most of the data processing in this pipeline should happen in the GPU anyway.

nemo/collections/asr/parts/preprocessing/features.py
jantonguirao: this pipeline works with float32. I think it should be no problem with AMP. I am not that familiar with it, so perhaps you can answer: Would the input still be float32 in case of AMP?

nemo/collections/asr/parts/preprocessing/features.py
jantonguirao: The whole pipeline works with fp32. We can cast it to float16 inside DALI, if needed, at the end of the pipeline, but the current pipeline doesn't do it.

nemo/collections/asr/parts/preprocessing/features.py
jantonguirao: I am wondering, why would the batch size change at the end of an epoch?

jantonguirao: You are right. I think we can make it rebuild only if the batch size increases. I'll try that.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -223,8 +249,22 @@ def __init__(
...
+        self.use_dali = use_dali
+        if self.use_dali:
+            if not HAVE_DALI:
+                raise ModuleNotFoundError(
+                    "AudioToMelSpectrogramPreprocessor: use_dali was set to True, but NVIDIA DALI is not installed."
+                    "To install, please follow https://docs.nvidia.com/deeplearning/dali/user-guide/docs/installation.html#nvidia-dali"
+                )
+            else:
+                logging.warning(
+                    f'Using NVIDIA DALI for feature extraction is experimental, not ready for production and is not fully supported. Use at your own risk.'
```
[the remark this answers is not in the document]
jantonguirao: Done.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
...
+
+        pad_end = dali.fn.slice(x, start=(x_len - stft_pad_amount - 1), shape=stft_pad_amount, axes=(0,))
+        pad_end = flip_1d(pad_end)
+        x = dali.fn.cat(pad_start, x, pad_end, axis=0)
+        return x
+
+    def _daligraph_splice_frames(self, x, nfeatures, x_len, stacking=1, subsampling=1):
+        if stacking > 1:
+            seq = [x]
+            for n in range(1, stacking):
```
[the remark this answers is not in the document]
jantonguirao: Documentation added.

tests/collections/asr/test_asr_modules.py
jantonguirao: Done.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -428,5 +476,175 @@ def forward(self, x, seq_len):
             pad_amt = x.size(-1) % pad_to
             if pad_amt != 0:
                 x = nn.functional.pad(x, (0, pad_to - pad_amt), value=self.pad_value)
-
         return x, seq_len
+
+    def _daligraph_reflect_pad(self, x, x_len, stft_pad_amount):
+        def flip_1d(x):
+            # TODO(janton): remove the layout trick when Flip supports arbitrary data layouts
+            x = dali.fn.reshape(x, shape=(-1, 1, 1), layout="HWC")
```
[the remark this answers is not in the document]
jantonguirao: Done.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -48,6 +50,30 @@
...
 
+try:
+    import nvidia.dali as dali
+    from nvidia.dali.plugin.pytorch import feed_ndarray as feed_ndarray
+
+    DALI_VERSION = version.parse(dali.__version__)
+    DALI_VERSION_MIN = version.parse('1.0.0')
+
+    HAVE_DALI = True
+except ModuleNotFoundError:
```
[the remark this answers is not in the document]
jantonguirao: Fixed.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -401,32 +469,232 @@ def forward(self, x, seq_len):
...
+        max_len = x.size(-1)
+        mask = torch.arange(max_len).to(x.device)
+        mask = mask.expand(x.size(0), max_len) >= seq_len.unsqueeze(1)
+        x = x.masked_fill(mask.unsqueeze(1).type(torch.bool).to(device=x.device), self.pad_value)
+        del mask
+
         # frame splicing if required
         if self.frame_splicing > 1:
-            x = splice_frames(x, self.frame_splicing)
+            x = splice_frames(x, stacking=self.frame_splicing)
```
titu1994, line 487: Passing stacking but not subsampling?

nemo/collections/asr/parts/preprocessing/features.py
titu1994: Makes sense, thanks for explanation

nemo/collections/asr/parts/preprocessing/features.py
titu1994: Pytorch lightning would move the input to this class to cuda, then amp would activate if `trainer.precision=16`. Can we add a parallel jenkins test that does the same as the main one but with `trainer.precision=16` set? If the config doesnt have it, it might need `+trainer.precision=16`.

nemo/collections/asr/parts/preprocessing/features.py
titu1994: Ok sounds good.

tests/collections/asr/test_asr_modules.py
titu1994: Thanks

titu1994 (commented):
Thanks for addressing all of the concerns. One thing left is the section regarding masked_fill padding with certain pad_value. It seems incorrect to remove this section - we have certain models that pad with different value

nemo/collections/asr/parts/preprocessing/features.py
jantonguirao: I did not apply subsampling because the previous implementation didn't do it either. I think the particular stacking/subsampling settings might depend on the particular model. It can be easily extended to support arbitrary subsampling when there is a need.

nemo/collections/asr/parts/preprocessing/features.py
```diff
@@ -401,32 +469,232 @@ def forward(self, x, seq_len):
...
 
         # normalize if required
         if self.normalize:
             x = normalize_batch(x, seq_len, normalize_type=self.normalize)
 
-        # mask to zero any values beyond seq_len in batch, pad to multiple of `pad_to` (for efficiency)
```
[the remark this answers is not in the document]
jantonguirao: It's not removed, it was moved to line 478, just before the frame splicing operation. The results are different when you splice before masking to zero the padded part of the batch.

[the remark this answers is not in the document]
titu1994: Oh right, that makes sense. Thanks for clarification.

titu1994 (approved):

titu1994:
Closed for the time being while other approaches are tested

PR #2272, opened 2021-05-27
TTS Doc Fix and Remove TTS Test

### LINKED ISSUE #2262: TTS User Guide
Please note the TTS user guide quick start seems to have two issues (https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/stable/tts/intro.html):

1) More easily solvable: spec_generator = SpectrogramGenerator.from_pretrained("tts_en_tacotron2") <- the variable should be named spec_gen to match the lines that follow

2) The last line:
sf.write("speech.wav", audio.to('cpu').numpy(), 22050)

This gives the error "RuntimeError: Error opening 'speech.wav': Format not recognised." I believe this is an error with soundfile, see: https://github.com/bastibe/python-soundfile/issues/203

EDIT: For issue 2, to fix this in Colab, simply create audio2 = audio.transpose(), which works. May just be an issue with the version of soundfile imported by Colab.

blisc:
Closes #2262

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1383,16 +1383,9 @@ pipeline {
         }
       }
       failFast true
-      parallel {
-        stage('QuartzNet15x5Base-En') {
-          steps {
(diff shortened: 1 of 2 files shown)


titu1994 (approved):
Thanks ! Let's add that test back once it's stable (or with a much larger threshold)

PR #2278, opened 2021-05-28
Added onnxruntime check of exported ONNX, bumped up default ONNX opset

borisfom:


drnikolaev (approved):
Empty line 238 seems to be redundant but overall looks great

PR #2279, opened 2021-05-28
Added unit test for hifigan export, fixed hifigan export

borisfom:
Also - removed runtime test from waveglow test (now done inside export)


ryanleary (approved):
LGTM

PR #2283, opened 2021-05-28
Fix ExpManager Issues and FastPitch

blisc:


titu1994 (approved):
LGTM. Just wondering why typecheck is being disabled there.

blisc:
> LGTM. Just wondering why typecheck is being disabled there.

For some reason, doing `super().forward()` uses it's own input_types but not the super() module's input_types

borisfom:
> For some reason, doing `super().forward()` uses it's own input_types but not the super() module's input_types
That's expected. super().forward() is just calling super()'s method instead of overridden one, but your object is still of derived class and resolution of all method calls from within forward() will not change.

PR #2284, opened 2021-05-28
Organize asr config folders

VahidooX:


titu1994 (approved):
The notebook diff is huge, so I'll trust you on that one. LGTM otherwise

nithinraok (approved):
LGTM

PR #2293, opened 2021-06-01
ASR improvements

titu1994:
# Changelog
- Update Citrinet configs.
- Improved error message when numba is not installed.
- Add tarred dataset support for all supported filetypes.
- Ignore extra parameters for layer-normed LSTM.
- Correct deprecated call signature for OmegaConf.

### DIFF
--- a/examples/asr/conf/citrinet/citrinet_384.yaml
+++ b/examples/asr/conf/citrinet/citrinet_384.yaml
@@ -1,6 +1,15 @@
 # This config contains the default values for training a Citrinet model with CTC loss and BPE-based vocabulary.
 # Default learning parameters in this config are set for effective batch size of 1k on 32 GPUs.
 # To train it with smaller batch sizes, you may need to re-tune the learning parameters or use higher accumulate_grad_batches.
+# If training for a short time, you can also reduce weight decay to 0.
+
+# Training Recipe
+# This model can be trained using the default settings in this config with FP32 precision.
+# When training under AMP, increase `warmup_steps` to 5000 for stable training.
+# In order to create Citrinet-C, find-replace `filters: 384` with `filters: C`.
+# When reducing the receptive field of these models, it is advised to reduce the amount of augmentation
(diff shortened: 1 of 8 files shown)


examples/asr/conf/citrinet/citrinet_384.yaml
```diff
@@ -11,30 +20,34 @@ model:
...
     shuffle: false
+    use_start_end_token: false
 
   model_defaults:
     repeat: 5
     dropout: 0.0
     separable: true
     se: true
     se_context_size: -1
+    kernel_size_factor: 1.0
```
VahidooX, line 50: Do you use these model_defaults in the code directly or they are just some place to keep the defaults?

titu1994: model defaults are used in config directly + used by code to know what is globally shared across the configs. Especially useful for RNNT configs. Here, setting the model_defaults.dropout would enable dropout on most of the blocks of Citrinet so its easier for the user.

VahidooX: It is fine to have it. Just wanted to make sure that our code is reading these values and set the appropriate places.
If we drop all kernel_size_factor in the rest of the config, does it use this value.

titu1994: Oh no - this value is passed by omegaconf resolution only. Later on this value can be changed but the rest of the config will not be updated. 

Hmm that's a good point, guess I should make it forced propagated via code. But it breaks one niche scenario. Hmm will fix that after some consideration in another pr

VahidooX: The option of not propagating is aslo not a bad choice. We would have them as a place holder in the config to get propagated by omegaconf. In this case, no need to worry about conflicts on our side.

titu1994: Yeah, best leave it to omegaconf to handle

PR #2298, opened 2021-06-02
update quantization

Slyne:
Signed-off-by: slyned <slyned@nvidia.com>

Add CER support

But the results of speech_to_text_quant_infer_trt.py and speech_to_text_quant_infer.py are a little different.

I've also tried to use speech_to_text_quant_infer_trt.py to do inference with NeMo models downloaded from ngc without adding --qat,  but the results also are not consistent with speech_to_text_infer.py.


titu1994 (approved):
Looks good to me. Could you comment on if the differences between the scripts are large or just marginal ?

PR #2302, opened 2021-06-03
Fix SqueezeExcite forward for fp16

Oktai15:
We need to have ability of run inference model with SE blocks in fp16 

Signed-off-by: Oktai Tatanov <oktai.tatanov@gmail.com>


titu1994 (commented):
The context changes nothing, since the next conv outside this context block will push it to fp16 again

PR #2310, opened 2021-06-04
get embedding for a single file

nithinraok:

Usage:
`speaker_model = ExtractSpeakerEmbeddingsModel.from_pretrained(model_name="speakerverification_speakernet")`
` embs = speaker_model.get_embedding('audio_path')`

### DIFF
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -341,8 +341,9 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
     @torch.no_grad()
     def get_embedding(self, path2audio_file):
         audio, sr = librosa.load(path2audio_file, sr=None)
-        if sr != 16000:
-            audio = librosa.core.resample(audio, sr, 16000)
+        target_sr = self._cfg.train_ds.get('sample_rate', 16000)
+        if sr != target_sr:
+            audio = librosa.core.resample(audio, sr, target_sr)
         audio_length = audio.shape[0]
         device = self.device
         audio_signal, audio_signal_len = (
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -342,7 +342,7 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
     def get_embedding(self, path2audio_file):
         audio, sr = librosa.load(path2audio_file, sr=None)
         if sr != 16000:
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/models/label_models.py
```diff
@@ -337,6 +338,23 @@ def setup_finetune_model(self, model_config: DictConfig):
 
         logging.info(f"Changed decoder output to # {self.decoder._num_classes} classes.")
 
+    @torch.no_grad()
+    def get_embedding(self, path2audio_file):
+        audio, sr = librosa.load(path2audio_file, sr=None)
+        if sr != 16000:
+            raise ValueError("Make sure input audio has sample rate of 16Khz")
```
jbalam-nv, line 345: Why are we throwing an error here? Shouldn't we just resample the audio to 16 kHz? That is the current behavior in ASR.

nithinraok: we could do that as well

titu1994: Resampling it is better.

titu1994 (commented):
If you don't need gradients, might as well explicitly put model.eval() inside this function. Also make sure that all tensors that are moved to gpu are explicitly deleted at the end of the function.

nemo/collections/asr/models/label_models.py
```diff
@@ -337,6 +338,23 @@ def setup_finetune_model(self, model_config: DictConfig):
 
         logging.info(f"Changed decoder output to # {self.decoder._num_classes} classes.")
 
+    @torch.no_grad()
+    def get_embedding(self, path2audio_file):
+        audio, sr = librosa.load(path2audio_file, sr=None)
+        if sr != 16000:
+            audio = librosa.core.resample(audio, sr, 16000)
```
jbalam-nv, line 345: The target sampling rate should be read from the model config. This will fail for a model that was trained with say 8 kHz data.

nithinraok: I guess the whole idea is to not read from config, hence the quick way to get embeddings for a single audio file and thought about providing the target sample rate, but since all speaker models are 16Khz hence the 16000 number.

titu1994: No that doesn't mean the default should be 16khz. The models SR should be used, whatever it is, from it's own config. You can use self.cfg to access the config that was used to build the model c

titu1994 (approved):
Looks good.

PR #2312, opened 2021-06-05
ddp translate GPU allocation fix

AlexGrinch:
Replaced `model.to(rank)` with `torch.cuda.set_device(rank)` so several model instances do not stuck on 1st GPU.


MaximumEntropy (changes requested):
Thanks for the PR! Do you think setting map_location in restore_from is a little cleaner?

MaximumEntropy (approved):

PR #2316, opened 2021-06-07
Pin OmegaConf version for 1.0.0

titu1994:
# Bugfix
- Update OmageConf signature to correctly save and restore models in OmegaConf 2.1.0 release


blisc (dismissed):

ericharper (approved):
LGTM! Thanks!

Just to summarize, we're adding an upper bound to omegaconf in this PR which is going to 1.0.0. We will then make the omegaconf forward compatibility changes in a separate PR to main.

PR #2318, opened 2021-06-08
Exportable fastpitch

ryanleary:


ryanleary:
@blisc merged in the latest changes from main, so this should have the remaining deltas only now.

PR #2319, opened 2021-06-08
[BUGFIX] OmegaConf forward compatibility

ericharper:
Apply updates where OmegaConf is breaking with 2.1.0.

See #2316 for the original PR to v1.0.0.

Note, we have to upper bound OmegaConf as Hydra 1.0.6 is not compatible with OmegaConf 2.1.0. When the new version of Hydra is released we should remove the upper bound.

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -308,9 +308,9 @@ class ModelPT(LightningModule, Model):
             conf = OmegaConf.load(path2yaml_file)
             for conf_path, item in self.artifacts.items():
                 if item.hashed_path is None:
-                    OmegaConf.update_node(conf, conf_path, item.path)
+                    OmegaConf.update(conf, conf_path, item.path)
                 else:
-                    OmegaConf.update_node(conf, conf_path, item.hashed_path)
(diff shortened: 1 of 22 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -307,10 +307,7 @@ def _update_artifact_paths(self, path2yaml_file):
         if self.artifacts is not None and len(self.artifacts) > 0:
             conf = OmegaConf.load(path2yaml_file)
             for conf_path, item in self.artifacts.items():
-                if item.hashed_path is None:
-                    conf.update_node(conf_path, item.path)
-                else:
-                    conf.update_node(conf_path, item.hashed_path)
+                OmegaConf.update(conf, conf_path, item.path)
             with open(path2yaml_file, 'w') as fout:
```
titu1994, line 311: This is incorrect - needs to have two paths as before just with updated syntax. I think I missed the else path in the previous patch.

titu1994 (changes requested):

nemo/core/classes/modelPT.py
ericharper: Good catch I'll update it.

titu1994 (approved):
Looks good now

PR #2345, opened 2021-06-11
Indic Tokenizer boilerplate

soumye:
Adds support for English - Hindi translation models by adding boilerplate Indic Tokenizer. Can be added with proper indic tokenizers in the future.


nemo/collections/common/tokenizers/indic_tokenizers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def detokenize(self, tokens: List[str]) -> str:
+        """
+        Detokenizes a list of tokens
+        Args:
+            tokens: list of strings as tokens
+        Returns:
+            detokenized string
+        """
+        return self.moses_detokenizer.detokenize(tokens)
```
MaximumEntropy, line 41: Curious - do you need this detokenizer here if you didn't tokenize the sentence, to begin with?

MaximumEntropy (commented):
LGTM, just curious about detokenization

nemo/collections/common/tokenizers/indic_tokenizers.py
soumye: Even if we didn't tokenize using moses, there will be space between period ie '.' and the last word. Detokenizer will get rid of it for example.

MaximumEntropy: I see, okay that sounds good to me.

MaximumEntropy (approved):
LGTM!

PR #2347, opened 2021-06-13
Fixed Hydra-core configuration

parthsrivastava100:
Hydra-core version 1.1.0  was released on 11th June 2021, and it requires OmegaConf == 2.1.\*, while Nemo requires OmegaConf >=2.0.5 & <2.1.*. Therefore, the existing tutorial notebooks( can be found [here](https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/starthere/tutorials.html), on importing the models, reported a configuration error. This pull request changed the specified version of hydra-core, specified in the requirements, from >=1.0.4 to ==1.0.6, which supports OmegaConf <2.1.0, and the tutorial notebooks run correctly, without any error

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -8,7 +8,7 @@ wrapt
 ruamel.yaml
 scikit-learn
 omegaconf>=2.0.5,<2.1.0
-hydra-core>=1.0.4
+hydra-core==1.0.6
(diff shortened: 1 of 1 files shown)


okuchaiev:
thanks for the PR, but it looks like this is now resolved in nemo >=1.1.1

PR #2358, opened 2021-06-14
[BUGFIX] Update jenkins test for tarred dataset creation

ericharper:
Two tests in parallel were outputting to the same directory.


MaximumEntropy (approved):
LGTM, thanks!

PR #2361, opened 2021-06-15
Aymen

amejri:


okuchaiev:
not sure what is "Confnet" - looks like it references Conformer paper which we already have?

PR #2373, opened 2021-06-18
Script for ASR inference on long files

jbalam-nv:
* speech_to_text_buffered_infer.py  supports inference on long audio files by running inference on smaller chunks of audio and them merging the tokens for final transcription.
* This version only supports EncDecCTCModelBPE with AudioMelSpectrogramProcessor as the preprocessor
* Feature normalization is done on each buffer using "per_feature" normalization, future versions will have other methods for experimentation


examples/asr/speech_to_text_buffered_infer.py
```diff
@@ -0,0 +1,159 @@
...
+from nemo.utils import logging
+
+can_gpu = torch.cuda.is_available()
+
+
+def get_wer_feat(mfst, asr, frame_len, tokens_per_chunk, delay, preprocessor_cfg, model_stride_in_secs, device):
+    # Create a preprocessor to convert audio samples into raw features,
+    # Normalization will be done per buffer in frame_bufferer
+    # Do not normalize whatever the model's preprocessor setting is
+    preprocessor_cfg.normalize = "None"
```
titu1994, line 43: String None?

nemo/collections/asr/parts/utils/streaming_utils.py
```diff
@@ -0,0 +1,363 @@
...
+    """
+
+    def __init__(self, asr_model, frame_len=1.6, batch_size=4, total_buffer=4.0):
+        '''
+        Args:
+          frame_len: frame's duration, seconds
+          frame_overlap: duration of overlaps before and after current frame, seconds
+          offset: number of symbols to drop for smooth streaming
+        '''
+        self.ZERO_LEVEL_SPEC_DB_VAL = -16.635
```
titu1994, line 151: Maybe some info can be added on how this was calculated?

titu1994 (approved):
Overall looks great, minor comments

examples/asr/speech_to_text_buffered_infer.py
jbalam-nv: Hack to get through without any normalization. None throws an error here https://github.com/NVIDIA/NeMo/blob/bee43e840837d7e30e17a65b972e9fe89ecb9ca5/nemo/collections/asr/parts/preprocessing/features.py#L78 , as we are looking for a key in a dict. We should probably clean up this normalize_batch function and add  a "no_norm" option.

titu1994: Ah ok. no_norm sounds like a good option

PR #2379, opened 2021-06-21
ListConfig for training, right to left models, noisy channel re-ranking and ensembling for NMT

MaximumEntropy:
This PR implements the following features

1. ListConfig for training NMT datasets that aren't restricted to multilingual training. Implication - users can specify multiple training text/tar files and sample from different sources with probabilities.

For example,

```
python examples/nlp/machine_translation/enc_dec_nmt.py -cn aayn_base  \
    ...
    model.train_ds.tar_files=["/path/to/dataset1_tarfiles_OP_1..990_CL_.tar, /path/to/dataset2_tarfiles_OP_1..200_CL_.tar"] \
    model.train_ds.concat_sampling_technique=random \
    model.train_ds.concat_sampling_probabilities=["0.6, 0.4"] \
    ...
```

2. Right to left factorized models. Implication - users can train NMT models that generate sequences from right to left (YTTM only). This is typically used in offline knowledge distillation.

For example, right to left training can be enabled with the following flags.

```
python examples/nlp/machine_translation/enc_dec_nmt.py -cn aayn_base  \
    ...
    model.decoder_tokenizer.r2l=true \
    model.encoder_tokenizer.r2l=false \
    ...
```

It is also possible to create a tarred dataset that is right to left using:

```
python examples/nlp/machine_translation/create_tarred_parallel_dataset.py \
    ...
    --src_fname /path/to/source.txt \
    --tgt_fname /path/to/target.txt \
    --decoder_tokenizer_r2l \
    ...
```

3. Implements noisy channel re-ranking. Implication - given a reverse direction translation model and a target side language model, this re-ranks beam search outputs based on (a) forward direction translation model scores, (b) reverse direction scores (c) language model scores. This typically improves translation quality. Requirement: All models must be trained with the same tokenizer.

Noisy channel re-ranking involves two steps (a) Generating translations from the forward model (typically with large beam sizes) (b) Re-ranking beam candidates

```
python  examples/nlp/machine_translation/nmt_transformer_infer.py \
    --model /path/to/forward_model.nemo \
    --srctext /path/to/source_test.txt \
    --tgtout /path/to/translations.txt \
    --write_scores \
    --beam_size 15 \
    ...
```

This will generate two files (a) /path/to/translations.txt and (b) /path/to/translations.txt.scores. The second is a TSV file structured as source \t target \t score for every beam hypothesis. We will use this to re-rank beam candidates

```
python examples/nlp/machine_translation/noisy_channel_reranking.py \
      --reverse_model /path/to/reverse_model.nemo \
      --beam_size 15 \
      --tgtout /path/to/reranked_translations.txt \
      --srctext /path/to/translations.txt.scores \
      --lm_model /path/to/target_lm.nemo \
      --forward_model_coef 1.0 \
      --reverse_model_coef 0.5 \
      --target_lm_coef 0.07 \
     --len_pen 0.7 \
      ...
```

4. Ensemble decoding. Implication - given a list of translation models (trained with the same tokenizer), ensemble decoding computes the probability over the next token in the sequence as the average of the probabilities given by each individual model. The averaged probabilities are used to do beam search. This again improves translation quality.

```
python  examples/nlp/machine_translation/nmt_transformer_infer.py \
    --model /path/to/forward_model_1.nemo,/path/to/forward_model_2.nemo,/path/to/forward_model_3.nemo \
    --srctext /path/to/source_test.txt \
    --tgtout /path/to/translations_ensemble.txt \
    --beam_size 4 \
    ...
```

Note: Noisy channel re-ranking also supports an ensemble of reverse NMT models and can also be provided as a comma-separated string.


nemo/collections/nlp/modules/common/transformer/transformer_generators.py
```diff
@@ -372,9 +381,355 @@ def _forward(self, decoder_input_ids=None, encoder_hidden_states=None, encoder_i
...
+        self.freeze()
 
-        return tgt.squeeze(1)
+        try:
+            yield
+        finally:
+            self.unfreeze()
 
 
 class BeamSearchSequenceGeneratorWithLanguageModel(GreedySequenceGenerator):
```
AlexGrinch, line 735: It seems that EnsembleBeamSearchSequenceGenerator has all the functional provided by BSSGWithLanguageModel. Thus, we can deprecate the latter.

examples/nlp/machine_translation/noisy_channel_reranking.py
```diff
@@ -0,0 +1,144 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
AlexGrinch, line 1: 1. What do you think about loading and using both P(Y|X) and P(X|Y) models in this script?

Right now NCR is done on top of the beams generated with `nmt_transformer_infer.py` and it uses incorrect P(Y|X) scores (genuine scores were trainsformed by LM and length penalty). Even though this transformation is linear, I think it would be easier to tune coefficients if all three scores P(Y|X), P(X|Y), P(Y) are taken from the models and lie in the corresponding value ranges.

2. Right now NCR script computes the score for a single value of `noisy_channel_coef`. To do the parameter sweep over it, you have to run the script N times, doing the same inference every time. I would advocate for computing P(Y|X), P(X|Y), P(Y) scores and saving them to file similar to `nmt_transformer_infer.py` with `--write_scores` enabled. Having all these scores at hand, we can do massive parameter sweep very cheap. We can also afford having two different coefficients for LM and reverse model and one additional coefficient for length penalty and do grid search over them.

AlexGrinch (changes requested):
Please, see the comments on the design of NCR script. I think it requires some changes to make it better for hyper parameter sweep.

examples/nlp/machine_translation/noisy_channel_reranking.py
```diff
@@ -0,0 +1,144 @@
...
+# limitations under the License.
+
+"""
+Given NMT model's .nemo file, this script can be used to translate text.
+USAGE Example:
+1. Obtain text file in src language. You can use sacrebleu to obtain standard test sets like so:
+    sacrebleu -t wmt14 -l de-en --echo src > wmt14-de-en.src
+2. Translate:
+    python nmt_transformer_infer.py --model=[Path to .nemo file] --srctext=wmt14-de-en.src --tgtout=wmt14-de-en.pre
+"""
```
okuchaiev, line 22: can we also add a link to the NCR paper in the docstring

examples/nlp/machine_translation/noisy_channel_reranking.py
```diff
@@ -0,0 +1,144 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Given NMT model's .nemo file, this script can be used to translate text.
+USAGE Example:
+1. Obtain text file in src language. You can use sacrebleu to obtain standard test sets like so:
+    sacrebleu -t wmt14 -l de-en --echo src > wmt14-de-en.src
+2. Translate:
+    python nmt_transformer_infer.py --model=[Path to .nemo file] --srctext=wmt14-de-en.src --tgtout=wmt14-de-en.pre
```
okuchaiev, line 21: this docstring needs fixing.

nemo/collections/nlp/models/language_modeling/transformer_lm_model.py
```diff
@@ -74,7 +74,12 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         model_name = encoder_cfg_dict.pop('model_name', None)
         pretrained = encoder_cfg_dict.pop('pretrained', False)
         self.encoder = get_transformer(
-            library=library, model_name=model_name, pretrained=pretrained, config_dict=encoder_cfg_dict, encoder=True,
+            library=library,
+            model_name=model_name,
+            pretrained=pretrained,
+            config_dict=encoder_cfg_dict,
+            encoder=True,
+            pre_ln_final_layer_norm=encoder_cfg_dict.get('pre_ln_final_layer_norm', False),
```
okuchaiev, line 82: This looks like a potential source of bugs, shouldn't it be: 
```
pre_ln_final_layer_norm=encoder_cfg_dict.get('pre_ln_final_layer_norm', encoder_cfg_dict.get('pre_ln', True))
```

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -456,71 +462,62 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
-                        datasets=datasets,
-                        sampling_technique=cfg.get('concat_sampling_technique'),
-                        sampling_temperature=cfg.get('concat_sampling_temperature'),
-                        sampling_probabilities=cfg.get('concat_sampling_probabilities'),
-                        global_rank=self.global_rank,
-                        world_size=self.world_size,
-                    )
+            metadata_file_list = cfg.get('metadata_file')
+            tar_files_list = cfg.get('tar_files', None)
+            if isinstance(metadata_file_list, str):
```
okuchaiev, line 467: Two options here:
1) Drop tar_files_list and tar_files params completely
2) If both ``metadata_file_list`` and ``tar_files_list`` raise an error and say only one needs to be specified.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -693,25 +688,52 @@ def batch_translate(
...
             self.train(mode=mode)
-        return inputs, translations
+        if return_beam_scores:
+            return inputs, all_translations, scores.data.cpu().numpy().tolist(), best_translations
+
+        return inputs, best_translations
+
+    def prepare_inference_batch(self, text, prepend_ids=[], target=False):
+        inputs = []
+        proessor = self.source_processor if not target else self.target_processor
```
okuchaiev, line 713: typo: proessor -> processor

okuchaiev (changes requested):
few small comment (pre-ln stuff and metadatafiles vs tarfiles)

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
okuchaiev: @aklife97 what do you think?

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
MaximumEntropy: Cool, I can move the BSSGWithLanguageModel logic into BeamSearchSequenceGenerator. Does that work?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: I like option 1 here as I don't really find a good use for manually specifying the tar_files. Even if we have older datasets that don't have metadata_files, it still is probably better to generate metadata_files for them and do away with manually specifying every filename.

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
AlexGrinch: I would rather remove it completely and leave BeamSearchSequenceGenerator as it is. Currently it implements clean Beam Search logic for a single model without extra models overhead. Let's have all advanced stuff in EBSSG.

MaximumEntropy: Okay sounds good

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: I don't like 1. for the reason that tarfile paths in the metadata file are only valid on the place where the tarred dataset was created. I want to be able to move tarfiles around from my workstation to different clusters and then point to paths there.

examples/nlp/machine_translation/noisy_channel_reranking.py
MaximumEntropy: Sounds good. I can do that. A couple of questions then (1) Should P(Y|X) scores be generated by `nmt_transformer_infer.py` or by `noisy_channel_reranking.py`? (2) Should I keep P(Y|X) and P(X|Y) length-penalty adjusted but not P(Y) (3) If I understand, `noisy_channel_reranking.py` should work both "from scratch" (using only the input text file) and also from "cached score files" (any subset of P(X|Y), P(Y|X) and P(Y) files and generate missing files if needed).

aklife97 (changes requested):
Looks really good to me overall! Just require some minor changes

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
okuchaiev: then let's just do 2.

nemo/collections/nlp/models/language_modeling/transformer_lm_model.py
MaximumEntropy: Yes, this looks more correct than what I have right now. It is not a problem right now because even if `pre_ln_final_layer_norm` is True and `pre_ln` is False, the extra layer norm won't be created. I'll make this change.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: Even in 2, both need to be specified because the metadata file contains information about num_batches and tar_files_list contains paths to the tarfiles it isn't possible to specify just one unless the metadata file contains valid paths on the env in which it is running.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: Nice catch. Fixed, thanks :)

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -456,71 +462,62 @@ def _setup_dataloader_from_config(self, cfg: DictConfig):
...
+        else:
+            src_file_list = cfg.src_file_name
+            tgt_file_list = cfg.tgt_file_name
+            if isinstance(src_file_list, str):
+                src_file_list = [src_file_list]
+            if isinstance(tgt_file_list, str):
+                tgt_file_list = [tgt_file_list]
 
             if len(src_file_list) != len(tgt_file_list):
                 raise ValueError(
```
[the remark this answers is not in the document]
MaximumEntropy: Ah yeah, thanks!

nemo/collections/nlp/modules/common/transformer/transformer_generators.py
```diff
@@ -372,9 +381,355 @@ def _forward(self, decoder_input_ids=None, encoder_hidden_states=None, encoder_i
...
+
+        # generate initial buffer of beam_size prefixes-hypotheses
+        outputs = [
+            self._one_step_forward(i, tgt, encoder_hidden_states[i], encoder_input_mask, None, 0)
+            for i in range(self.num_models)
+        ]
+        nmt_log_probs = self._average_probs([x[0] for x in outputs])
+        decoder_mems_lists = [x[1] for x in outputs]
+
+        if self.language_model is not None:
```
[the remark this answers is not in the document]
MaximumEntropy: Yeah, I agree with this - there is quite a bit of redundancy between BeamSearch and EnsembleBeamSearch. Probably best to do as part of a separate PR like you suggested.

examples/nlp/machine_translation/noisy_channel_reranking.py
MaximumEntropy: I've made changes that should most of the concerns. The only thing pending is that scores generated with shallow fusion will still have LM scores within them - doing this will probably require quite a few changes to the corresponding shallow fusion class. I'll probably address that in another PR. For now, we just need to take care to adjust --forward_model_coef accordingly.

MaximumEntropy: Main changes - 1) Forward, reverse and lm scores do not contain length penalties 2) Can do sweeps without loading any models from just cached score files that record src, tgt, forward_score, reverse_score, lm_score, src_len, tgt_len 3) Added an arg to do length-normalization of forward/reverse scores like in fairseq and length penalty as well.

examples/nlp/machine_translation/noisy_channel_reranking.py
MaximumEntropy: This has been added, thanks!

examples/nlp/machine_translation/noisy_channel_reranking.py
MaximumEntropy: Fixed, thanks!

nemo/collections/nlp/models/language_modeling/transformer_lm_model.py
MaximumEntropy: This should now be fixed.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: @okuchaiev, @aklife97, and @ericharper hope it is fine with all of you to keep things as they are now to allow pre-processing on a certain machine and training on another.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -693,25 +688,52 @@ def batch_translate(
...
+                encoder_hidden_states=src_hiddens, encoder_input_mask=src_mask, return_beam_scores=return_beam_scores
+            )
+            if return_beam_scores:
+                all_translations, scores, best_translations = best_translations
+                scores = scores.view(-1)
+                all_translations = self.postprocess_beam_results(all_translations)
 
-            translations = [self.decoder_tokenizer.ids_to_text(tr) for tr in beam_results.cpu().numpy()]
+            best_translations = self.postprocess_beam_results(best_translations)
             inputs = [self.encoder_tokenizer.ids_to_text(inp) for inp in src.cpu().numpy()]
```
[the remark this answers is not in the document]
MaximumEntropy: I've refactored this, thanks!

aklife97 (approved):

aklife97:
LGTM! We can merge once the CIs pass

AlexGrinch (approved):

AlexGrinch:
LGTM as well! We can merge once the CIs pass

okuchaiev (approved):

PR #2380, opened 2021-06-21
[BUGFIX] NMT Multi-node was incorrectly computing num_replicas

ericharper:
We need to instantiate the `NLPDDPPlugin` with the `num_nodes` gotten from the trainer config.


okuchaiev (approved):
Thanks!

PR #2396, opened 2021-06-23
Update ModelCheckpoint

blisc:


okuchaiev (approved):
lgtm

titu1994 (commented):
LGTM

titu1994:
We should eventually pop off the params.period too so it doesn't remain in the dictionary

PR #2397, opened 2021-06-23
Mean normalization

nithinraok:
fix voxceleb eval and add embedding normalization

### DIFF
--- a/examples/speaker_recognition/extract_speaker_embeddings.py
+++ b/examples/speaker_recognition/extract_speaker_embeddings.py
@@ -89,7 +89,7 @@ def main():
             labels=None,
             batch_size=1,
             shuffle=False,
-            time_length=8,
+            time_length=20,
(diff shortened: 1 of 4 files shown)


jbalam-nv (approved):
LGTM

titu1994 (commented):
Mean normalization over batch dim? Wont this make it dependent on what batch of samples you get at any time ?

nithinraok:
> Mean normalization over batch dim? Wont this make it dependent on what batch of samples you get at any time ?

yes, this is normalization over batch samples. This is to minimize the source channel effect on embeddings before we compute cosine distance score on enrolled speaker and impostor speaker (verification). There is another normalization called global which doesn't depend on batch dim. Will add that once ready, fully tested, and ready for diarization cases.

titu1994 (approved):
Ok sounds good then. Maybe docs need to be updated later

PR #2398, opened 2021-06-23
Bugfix adaptive spec augment time masking

titu1994:
# Bugfix
- Fix a bug in adaptive time masking where the mask start index could begin in a padded region, therefore making the mask useless (ASR losses dont access masked sections when computing loss).
- In effect, augmentation was reduced.

### DIFF
--- a/nemo/collections/asr/parts/numba/spec_augment/spec_aug_numba.py
+++ b/nemo/collections/asr/parts/numba/spec_augment/spec_aug_numba.py
@@ -267,7 +267,6 @@ class SpecAugmentNumba(nn.Module, Typing):
             else:
                 time_width = (
                     torch.tensor(self.time_width, dtype=torch.int32, device=input_spec.device)
-                    .clamp(min=1)
                     .unsqueeze(0)
                     .repeat(sh[0])
                 )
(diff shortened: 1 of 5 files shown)


jbalam-nv (approved):
LGTM

nemo/collections/asr/parts/numba/spec_augment/spec_aug_numba.py
```diff
@@ -267,10 +262,30 @@ def forward(self, input_spec, length):
...
+            else:
+                time_width = (
+                    torch.tensor(self.time_width, dtype=torch.int32, device=input_spec.device)
+                    .unsqueeze(0)
+                    .repeat(sh[0])
+                )
+
+            time_starts = []
+            time_lengths = []
+            for idx in range(sh[0]):
```
redoctopus, line 276: Just curious--have you profiled the before/after performance of spec augment? Wondering what the potential additional cost of the for-loop is here.

redoctopus (approved):
Thanks!

nemo/collections/asr/parts/numba/spec_augment/spec_aug_numba.py
titu1994: About 300 microseconds added on average, on overall scale its thankfully minimal

PR #2408, opened 2021-06-25
fixed minor errors in docs and filter_langs_nmt.py

Davood-M:
Signed-off-by: David Mosallanezhad <dmosallanezh@nvidia.com>

### DIFF
--- a/scripts/neural_machine_translation/filter_langs_nmt.py
+++ b/scripts/neural_machine_translation/filter_langs_nmt.py
@@ -20,8 +20,16 @@ import shutil
 import warnings
 from pathlib import Path
 from time import sleep
+import os
 
-import fasttext
+try:
+    import fasttext
+    # temp fix for the warning messages
+    fasttext.FastText.eprint = lambda x: None
+except ImportError:
+    # install fasttext
+    from pip._internal import main as pipmain
+    pipmain(['install', 'fasttext'])
 from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


scripts/neural_machine_translation/filter_langs_nmt.py
```diff
@@ -20,8 +20,16 @@
...
 from time import sleep
-
-import fasttext
+import os
+
+try:
+    import fasttext
+    # temp fix for the warning messages
+    fasttext.FastText.eprint = lambda x: None
+except ImportError:
```
MaximumEntropy, line 29: I think this should be fixed by adding fastText to the nlp specific requirements.txt file rather than installing here.

scripts/neural_machine_translation/filter_langs_nmt.py
```diff
@@ -20,8 +20,16 @@
...
 from pathlib import Path
 from time import sleep
-
-import fasttext
+import os
+
+try:
+    import fasttext
+    # temp fix for the warning messages
+    fasttext.FastText.eprint = lambda x: None
```
MaximumEntropy, line 28: Curious what warning messages you are trying to suppress here?

MaximumEntropy (changes requested):
Thanks for the PR! It looks good to me once we address the way in which we install fastText here.

PR #2417, opened 2021-06-28
[BUGFIX] Megatron in NMT was setting vocab_file to None

ericharper:
Makes vocab_file configurable with Megatron encoders in NeMo NMT.


okuchaiev (approved):
lgtm

PR #2421, opened 2021-06-29
Nmt byte level tokenizer model fix

michalivne:
Handles correctly byte-level tokenizer when tokenizer models are missing.


michalivne:
@okuchaiev @MaximumEntropy @ericharper 
Can any of you review the PR?
It is a small fix that is required for the byte-level tokenizer.

ericharper (approved):
LGTM. Thanks!

okuchaiev (approved):
lgtm

PR #2439, opened 2021-07-02
Changes for ONNX conversion of Conformer models

jbalam-nv:
* Brought in the needed changes from Vahid's earlier PR: https://github.com/NVIDIA/NeMo/pull/2359/files
* Tested conversion with scripts/export/convasr_to_single_onnx.py
* Tested the converted large, medium and small models with dev-clean and WER is very close (within 0.01%) of correspondiong .nemo models


titu1994 (commented):
Swish to SiLU is necessary and should not be reverted.

jbalam-nv:
> Swish to SiLU is necessary and should not be reverted.

Reverted these changes as onnx conversion of nn.silu is now supported in pytorch

titu1994 (approved):
LGTM

PR #2451, opened 2021-07-08
Add ckpt to nemo script

piraka9011:
Sometimes I have ckpt files that I want to use/test while my model is still training.
This script simply converts those ckpt's to nemo files to enable various use cases with Jarvix/Onnx/etc.


okuchaiev:
Thank you for the PR! But this is not going to handle artifacts such as tokenizers correctly because .ckpt does not contain any of those.
It might work for you locally because PTL checkpoint saves config, but if you give .ckpt file to someone else this script will not work for them.

piraka9011:
I see, thank you for the clarification!

PR #2456, opened 2021-07-09
Make parallel version of ConvASREncoder

mousebaiker:
The change introduces an encoder that can stack and execute several similar Jasper blocks side by side. 

This implementation can be used to define and train CarneliNet.

Underlying ParallelBlock supports tower dropout to train models that can be reconfigured during inference time.

# Usage

## Config 
Define proper target for encoder.  Use multiple values of kernel sizes to define the towers. The rest of the parameters will be shared across all of them.
```
          ...
  encoder:
    _target_: nemo.collections.asr.modules.ParallelConvASREncoder

          ...
    jasper:
      - filters: 384
        kernel: [11, 13, 15, 17]
        aggregation_mode: "dropout"
        block_dropout: 0.1
        ...
        
```

## Code 

Just call necessary model with correct config, as usual.
```
cfg = OmegaConf.load(cfg_path)
asr_model = nemo_asr.models.EncDecCTCModelBPE(cfg=cgf, trainer=trainer)
```


nemo/collections/asr/parts/submodules/jasper.py
```diff
@@ -958,3 +958,80 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]):
             return xs + [out], lens
 
         return [out], lens
+
+
+class ParallelBlock(nn.Module):
+    """
+    Computational module that computes several `blocks` independently from each other and aggregates the outputs.
+    Due to a particular way of structuring inputs with sequence masks in Jasper-like blocks, this module focuses
+    specifically on the Jasper family.
```
titu1994, line 967: Most asr models take length as input and emit length as output, can this be generalized to other model types?

nemo/collections/asr/parts/submodules/jasper.py
```diff
@@ -958,3 +958,80 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]):
...
+        super().__init__()
+        self.blocks = nn.ModuleList(blocks)
+        self.aggregation_mode = aggregation_mode
+        if aggregation_mode == "dropout":
+            self.weights = nn.Parameter(torch.ones(1, len(blocks), 1, 1), requires_grad=False)
+            self.dropout = nn.Dropout(block_dropout_prob)
+
+    def get_dropout_mask(self):
+        weights = self.dropout(self.weights)
+        while torch.sum(weights) == 0 and self.dropout.p < 1.0:
```
titu1994, line 989: Why is this while loop here? It is fine if dropout(weights) doesnt dropout anything for a few steps

nemo/collections/asr/parts/submodules/jasper.py
```diff
@@ -958,3 +958,80 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]):
...
+        if self.aggregation_mode == "dropout":
+            scaling_weights = self.get_dropout_mask()
+
+        for i, block in enumerate(self.blocks):
+            output, mask = block(x)
+
+            weighted_output = output[-1]
+            out_channels = weighted_output.shape[-2]
+            if self.aggregation_mode == "dropout":
+                weighted_output = scaling_weights[:, i, :, :] * output[-1]
```
titu1994, line 1024: This is just pointwise scaling, is it necessary to use a rank 4 tensor to store this weights?

titu1994 (commented):
Overall looks great, minor comments

nemo/collections/asr/parts/submodules/jasper.py
mousebaiker: Changed comment to be more general about the model types and more specific about the requirements imposed.

nemo/collections/asr/parts/submodules/jasper.py
mousebaiker: This loop prevents the inverse situation: dropout discarding all towers. Since the number of towers is low, this can happen quite often. 
Since the residual is there anyway, the model will learn even with all towers discarded, emulating stochastic depth behaviour, but we want at least one tower to learn something at each iteration.

nemo/collections/asr/parts/submodules/jasper.py
mousebaiker: This was legacy from the time where we also had per-element scaling. 

I changed it to 1D tensor. This will break some of the previous tower dropout checkpoints but that's not an issue right now, since we're not actively using them.

mousebaiker:
Som, thank you for the review. 

I resolved the issues. 
Per our discussion, I've also added a flag for residual connections with convolutions. 
PTAL.

nemo/collections/asr/parts/submodules/jasper.py
titu1994: Ah that makes sense. Alright let's leave it as is.

titu1994 (approved):
Looks great !

titu1994:
Merge when tests pass

PR #2478, opened 2021-07-13
remove redundant tests that fail during multiple ci tests

yzhang123:


ericharper (approved):
Thanks!

PR #2481, opened 2021-07-14
Replace deprecated imp with newer importlib machinery

mousebaiker:
`imp` module has been deprecated since 3.4. Since we do not support older versions of python, we should use supported functions. 

Unfortunately, there is no direct equivalent to `load_source()` function. While `exec_module` loads necessary objects into `package_info` object, it's not possible to use `package_info` in the import statement directly. 

I could not find any way to use this module in the import statements without hacks, but I must admit that I am not very well versed in the python import system. 

The resulting code is not clean, but I want to use this PR a starting point for further discussion.


okuchaiev:
@DEKHTIARJonathan any advice here?

DEKHTIARJonathan:
@okuchaiev it's literally my code in this PR. What kind of opinion you need ?

DEKHTIARJonathan (approved):
LGTM

PR #2482, opened 2021-07-14
Remove multiplication of sample rate for pydub indexing

felixmcgregor:
When loading segments, specified with offset, files that are not supported by SoundFile (i.e. filetypes that are non-seekable such as GSM encoded wavs) the script resorts to using [pydub](https://github.com/jiaaro/pydub). 

Pydub uses millisecond indexing. Fix to remove incorrect multiplication with sampling rate at offset.


titu1994:
Thanks for the fix !

titu1994 (approved):

titu1994 (approved):

PR #2505, opened 2021-07-19
Updated ctc-segmentation Python package version

lumaku:
This PR updates the `ctc-segmentation` Python package used for the CTC segmentation tool from version 1.1.0 to 1.4.0. The update provides a bugfix, as well as sane defaults, which results in better alignments.

I would also recommend the latest version, however, the interface changed at two points:

1. As not every ASR architecture uses subsampling, I replaced the `subsampling` and `frame_duration` parameters with a new parameter `index_duration`. This parameter can simply be calculated as `frame_duration_ms * subsampling_factor / 1000`.
2. The text can now be given as a token ID sequence obtained from the tokenizer (which is already integrated into Nemo). The text input was chosen for our initial implementation, because our ASR model had long tokens and it made sense to scan for partial tokens. As the pre-trained Nemo models are already character-based, scanning for partial tokens is not necessary; using this method will improve alignment speed and probably also generate better alignments (because the pre-processing of the text may be better with a tokenizer).

@ekmb


ekmb:
@lumaku thanks for the PR. Unfortunately, out of the box the upgrade to 1.4.0 produces worse alignments, you could see it if you run the tutorial: most of the resulting audio segments are cut prematurely (the final letters are missing) or too late (the start of the next segment is heard). I'd have to look into what exactly was changed for the updated version. Why do you recommend 1.4.0 and not the latest 1.6.7 version?

lumaku:
True, with the 1.4.0 version, the alignments got worse in the example - I would have expected otherwise.

My guess: The "older" method replaces the spaces between words with blanks which seems to work better with the Nemo model. The scores are then better because space tokens are not accounted for in the scoring. The blank token also serves as placeholder-delimiter between utterances.

You can see the activations by plotting the CTC log posterior probabilities. Here from the example file from 16.0 until 21.0 seconds:
ctc_activations_16-21
Text: "A carrier on his way to a market..."

Spiky "Blank" activations occur relatively often in the example. See how often the "blank" (blue line) is activated in-between characters? These spikes are there in most cases, but sometimes, tokens are more dense. there are relatively few "space" spikes in-between.

> Why do you recommend 1.4.0 and not the latest 1.6.7 version?

The latest version has more features but would require some changes in the code, and v1.4.0 would be the latest without any changes to the code.... To be compatible with a newer version, the `get_segments()` function needs additional parameters, e.g.:
```python
# change timing
config.index_duration = frame_duration_ms * subsampling_factor / 1000
# new configuration
config.blank_transition_cost_zero = False
config.replace_spaces_with_blanks = True
```

The new version has also a feature to set the transition code of blank to zero that allows to skip unrelated segments. But well, I tested it with the example and it did not improve the early start of the first utterance...

The latest version allows to align from tokenized text - this avoids some complicated text peprocessing. However, in this case, because the dictionary of the example model already contains only characters, using tokenized text would change alignments only slightly.

@ekmb After I saw that the change did not bring improvements, my motivation is low to update. Should we close this PR for now?

PR #2510, opened 2021-07-20
Enable RNNT ONNX Export

titu1994:
# Changelog
- Enable RNNT Export to ONNX (No torchscript yet)
- Add inference script for onnx exported rnnt models
- Update neural types for export
- Refactor Exportable to be less monolithic
- Add `ExportableEncDecJointModel` that is subclass of Exportable


nemo/collections/asr/modules/conv_asr.py
```diff
@@ -62,8 +63,14 @@ def _prepare_for_export(self, **kwargs):
         m_count = 0
         for m in self.modules():
             if isinstance(m, MaskedConv1d):
-                m.use_mask = False
-                m_count += 1
+                if hasattr(self, '_rnnt_export') and self._rnnt_export:
```
borisfom, line 66: Why not overriding this function in _rnnt_export ?

titu1994: That's an option, but that would make framework dependent on what type of encoder was used. This way the Conformer and Citrinet can chose to supply seperate input-output ages for CTC and RNNT exports while CTC and RNNT export framework remains unaware of what encoder type it is

borisfom: Also, I do not see any changes in this PR (nor in main) that would set _rnnt_export attribute - are those part of another PR ?

titu1994: Part of this PR Inside the asr_model.py file (needs to be expanded)

nemo/collections/asr/models/asr_model.py
```diff
@@ -89,3 +91,502 @@ def forward_for_export(self, input, length=None):
...
+            # Disable typechecks
+            typecheck.set_typecheck_enabled(enabled=False)
+
+            # Set module to eval mode
+            self._set_eval(set_eval)
+
+            format = self.get_format(output)
+
+            # Assign special flag for RNNT export of encoder
+            self.input_module._rnnt_export = True
```
titu1994, line 184: @borisfom here is where we set it explicitly to the modules. Needs to be before `prepare_for_export` since samples are already extracted by the time `prepare_for_extract` is called.

borisfom (approved):

PR #2522, opened 2021-07-21
Fix time stamps

nithinraok:
* Update timestamps to calculate only once
* introduce segmentation module
* support multi-batch inference for speaker embedding extraction

### DIFF
--- a/examples/speaker_recognition/conf/speaker_diarization.yaml
+++ b/examples/speaker_recognition/conf/speaker_diarization.yaml
@@ -1,7 +1,8 @@
 name: &name "ClusterDiarizer"
 
 num_workers: 4
-sample_rate: &sample_rate 16000
+sample_rate: 16000
+batch_size: 64
 
 diarizer:
   oracle_num_speakers: ??? # oracle number of speakers or pass null if not known. [int or path to file containing uniq-id with num of speakers of that session]
--- a/nemo/collections/asr/parts/utils/speaker_utils.py
+++ b/nemo/collections/asr/parts/utils/speaker_utils.py
@@ -304,7 +304,7 @@ def get_DER(all_reference, all_hypothesis):
     collar in md-eval.pl, 0.5s should be applied for pyannote.metrics.
 
     """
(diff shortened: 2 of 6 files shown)


okuchaiev:
is this a bugfix? should this go to r1.2.0?

nithinraok:
not a bug fix.

nemo/collections/asr/parts/utils/speaker_utils.py
```diff
@@ -395,8 +394,25 @@ def write_rttm2manifest(paths2audio_files, paths2rttm_files, manifest_file):
...
 ):
+    """
+    Generate subsegments manifest from segments manifest file
+    Args
+    input:
+        segments_manifest file (str): path to segments manifest file, typically from VAD output
+        sub_segments_manifest_file (str): path to output sub segments manifest file (default (None) : writes to current working directory)
+        window (float): window length for segments to sub segments division
+        shift (float): hop length for sub segments shift 
+        min_subsegments_duration (float): exclude sub segments smaller than this duration value
```
tango4j, line 411: This is very minor point, but it's better to be no space between "sub" and "segment"
https://www.dictionary.com/browse/subsegment

"subsegment"
This is for the sake of supporting search of the specific word.

tango4j (commented):
All in all, there seems to be no major issue but it would be better off we use consistent variable name over the whole codebase.

sub_segment, sub segment ---> subsegment 

for both doc string and variables.

examples/speaker_recognition/conf/speaker_diarization.yaml
```diff
@@ -1,7 +1,8 @@
 name: &name "ClusterDiarizer"
 
 num_workers: 4
-sample_rate: &sample_rate 16000
+sample_rate: 16000
```
titu1994, line 4: These two args should be inside model but upto you.

titu1994 (commented):
Overall looks fine, perform the requested changes and then @tango4j for final review

tango4j (approved):
Seems like the required changes are all reflected.

PR #2568, opened 2021-07-28
Reduce CPU unittest frequency

okuchaiev:
This PR changes when CPU unittests are run.
They are now run only if something under "nemo_text_processing" is changed.


ericharper:
Should we split the text processing cpu tests and run the text processing test only when the text processing files are changed?

titu1994:
Yes I think splitting CPU core tests and itn tests, then running itn tests only on change would be better

titu1994:
Closed via https://github.com/NVIDIA/NeMo/pull/2611

PR #2592, opened 2021-08-02
Add save restore connector to ModelPT

ericharper:
The SaveRestoreConnector moves the logic for creating and restoring NeMo models to a connector so that other software libraries can more easily customize the functionality.


nemo/collections/nlp/models/nlp_model.py
```diff
@@ -57,9 +58,17 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         super().__init__(cfg, trainer)
         self.set_world_size(trainer)
 
-    def register_artifact(self, config_path: str, src: str, verify_src_exists: bool = False):
+    def register_artifact(
+        self,
+        config_path: str,
+        src: str,
+        verify_src_exists: bool = False,
+        save_restore_connector: SaveRestoreConnector = SaveRestoreConnector(),
```
titu1994, line 66: All function arguments should generally be immutable, I'd rather this save restore default argument be None, and in the implementation, None be replaced by an object of SaveRestoreConnecter. Please update this throughout all such calls. Such a check and replace can be done inside the update save restore connector method itself 

Reason: function arguments are parsed exactly once - at function definition time, not at function runtime. https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -335,6 +344,7 @@ def restore_from(
         strict: bool = True,
         return_config: bool = False,
         trainer: Trainer = None,
+        save_restore_connector: SaveRestoreConnector = SaveRestoreConnector(),
```
titu1994, line 347: Same comment as before

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -229,3 +230,6 @@ def save_checkpoint(self, filepath, weights_only: bool):
                     )
                     atomic_save(checkpoint, filepath)
         return None
+
```
titu1994, line 233: Not required ? Is it a placeholder ?

titu1994 (changes requested):
Overall it looks fantastic. Minor comments about function arguments

nemo/collections/nlp/models/nlp_model.py
tkornuta-nvidia: +1 for Som's comment on immutable args.

But aside of it I need to go deeper into implementation of ModelPT, but for now not fully understanding the logic here: why do we have to add connector to `register_artifacts`? Do we assume that the connector for saving/restoring of a model can change for many times once a process runs, e.g. each time a different artifact is added?

nemo/core/classes/modelPT.py
```diff
@@ -1350,31 +1120,16 @@ def cfg(self, cfg):
...
-        tar.extractall(path=out_folder)
-        tar.close()
-        return out_folder
-
     @staticmethod
     def _is_model_being_restored() -> bool:
-        global _MODEL_IS_RESTORED
-        return _MODEL_IS_RESTORED
+        app_state = AppState()
+        return app_state.model_is_restored
```
tkornuta-nvidia, line 1126: Just a note: if I understand it correctly, the new solution has the same flaw as the old one, i.e. it won't properly handle cases where more than 1 model object is instantiated

nemo/core/classes/modelPT.py
```diff
@@ -1390,29 +1145,9 @@ def _set_model_guid(self):
...
-        is a tarfile (can be any valid archive)._MODEL_EFF_SAVE
-        """
-        app_state = AppState()
-
-        if app_state.model_restore_path is None:
-            return False
+    @classmethod
+    def update_save_restore_connector(cls, save_restore_connector):
+        if hasattr(cls, '_save_restore_connector'):
+            cls._save_restore_connector = save_restore_connector
```
tkornuta-nvidia, line 1151: in line 113 (constructor) you are setting that as instance (self) attribute.

If you would set it as class attribute, in __init_subclass__ method, then this (and generally logic of handling cls._save_restore_connector) could be simplified...

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+from omegaconf import DictConfig, OmegaConf
+from omegaconf.omegaconf import open_dict
+
+from nemo.utils import app_state, logging, model_utils
+from nemo.utils.app_state import AppState
+from nemo.utils.model_utils import import_class_by_path
+
+
+class SaveRestoreConnector:
+    def _default_save_to(self, model, save_path: str):
```
tkornuta-nvidia, line 35: please simplify and get rid of `_default`, just `save_to`

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+
+			Args:
+				save_path: Path to .nemo file where model instance should be saved
+
+		"""
+        app_state = AppState()
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            config_yaml = path.join(tmpdir, app_state.model_config_yaml)
+            model_weights = path.join(tmpdir, app_state.model_weights_ckpt)
```
tkornuta-nvidia, line 52: I am really not sure if those two "constants" should be attributes of app_state.

Why not make them attributes of the Connector class?

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+            model.to_config_file(path2yaml_file=config_yaml)
+            if hasattr(model, 'artifacts') and model.artifacts is not None:
+                self._handle_artifacts(model, nemo_file_folder=tmpdir)
+                # We should not update self._cfg here - the model can still be in use
+                self._update_artifact_paths(model, path2yaml_file=config_yaml)
+            self._save_state_dict_to_disk(model.state_dict(), model_weights)
+            self._make_nemo_file_from_folder(filename=save_path, source_dir=tmpdir)
+
+    def _default_restore_from(
```
tkornuta-nvidia, line 61: Similarly here: please simplify and get rid of `_default`, just `restore_from`

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                if return_config:
+                    instance = conf
+                else:
+                    app_state = AppState()
+                    if app_state.model_parallel_rank is not None:
+                        model_weights = path.join(
+                            tmpdir, f'mp_rank_{app_state.model_parallel_rank:02}', app_state.model_weights_ckpt
+                        )
+                    else:
+                        model_weights = path.join(tmpdir, app_state.model_weights_ckpt)
```
tkornuta-nvidia, line 132: This seems to be related with model parallel handling. Is there any chance to "abstract" that and move to a separate method, so if TLT will overload `restore_from` it will be able to inherit and leverage that logic?

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                    if override_config_path is not None:
+                        # Resolve the override config
+                        conf = OmegaConf.to_container(conf, resolve=True)
+                        conf = OmegaConf.create(conf)
+                # If override is top level config, extract just `model` from it
+                if 'model' in conf:
+                    conf = conf.model
+
+                if return_config:
+                    instance = conf
```
tkornuta-nvidia, line 124: I still think that `restore_from` should always return instance of ModelPT derived object.

Hence there should be a separate method for "restoring configuration" only. Please refer to EFF `NeMoCookbook.retrieve_configuration`:
https://gitlab-master.nvidia.com/tlt/eff/-/blob/main/src/eff/cookbooks/nemo_cookbook.py#L450

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+            cwd = os.getcwd()
+            try:
+                # Step into the nemo archive to try and find the file
+                with tempfile.TemporaryDirectory() as archive_dir:
+                    self._unpack_nemo_file(path2file=model_metadata.restoration_path, out_folder=archive_dir)
+                    os.chdir(archive_dir)
+                    for conf_path, artiitem in tarfile_artifacts:
+                        # Get basename and copy it to nemo_file_folder
+                        if 'nemo:' in artiitem.path:
+                            artifact_base_name = artiitem.path.split('nemo:')[1]
```
tkornuta-nvidia, line 278: hmmm, I haven't followed latest changes, but why you have this "nemo:" prefix  in artifact filename? seems you are getting rid of it anyway...

nemo/core/connectors/save_restore_connector.py
titu1994: This is required by Nemo, EFF can chose to not expose this argument but it's mandatory requirement for Nemo

nemo/core/connectors/save_restore_connector.py
titu1994: This is infected into the saved model config, and reused to restore the hashed artifact from the tarfiles. The slice off is to get the hashed value, the key we look for is Nemo:

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                        filepath = os.path.join(save_dir, f"{primary_key}.ckpt")
+                        self._save_state_dict_to_disk(state_dict_subset, filepath)
+
+                logging.info(f'Checkpoints from {restore_path} were successfully extracted into {save_dir}.')
+            finally:
+                os.chdir(cwd)
+
+        return state_dict
+
+    def _register_artifact(self, model, config_path: str, src: str, verify_src_exists: bool = True):
```
zeyuznv, line 189: can we get rid of `_` prefix here?

nemo/collections/nlp/models/nlp_model.py
ericharper: Re: the connector for register_artifact, it's only there because in NLPModel we don't call super.__init__ until after setting up tokenizers. If we can fix that, then we can remove connector from register_artifact.

nemo/core/classes/modelPT.py
ericharper: @titu1994 was able to add multi-model instantiation so I think it works.

nemo/core/classes/modelPT.py
ericharper: I'll look into this.

nemo/core/connectors/save_restore_connector.py
ericharper: That should work now that we have the connector class :)

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+import torch
+from omegaconf import DictConfig, OmegaConf
+from omegaconf.omegaconf import open_dict
+
+from nemo.utils import app_state, logging, model_utils
+from nemo.utils.app_state import AppState
+from nemo.utils.model_utils import import_class_by_path
+
+
+class SaveRestoreConnector:
```
zeyuznv, line 34: I feel this class has so many methods starting with `_`. If they should be used outside the class, they should not have `_` prefix

nemo/core/classes/modelPT.py
titu1994: Ah missed this comment. Multimodal restoration works now that apostate manages model IDs at a global level. Any number of Nemo models can be instantiated - either via restore or from constructor

nemo/collections/nlp/models/nlp_model.py
ericharper: Defaulted all save_restore_connector args to None.

nemo/collections/nlp/models/nlp_model.py
ericharper: Done.

nemo/collections/nlp/parts/nlp_overrides.py
ericharper: Were you referring to the comment below this?

nemo/core/classes/common.py
```diff
@@ -622,6 +623,7 @@ def from_pretrained(
         map_location: Optional['torch.device'] = None,
         strict: bool = True,
         return_config: bool = False,
+        save_restore_connector: SaveRestoreConnector = SaveRestoreConnector(),
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/classes/modelPT.py
```diff
@@ -157,7 +152,13 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # ModelPT wrappers over subclass implementations
         self.training_step = model_utils.wrap_training_step(self.training_step)
 
-    def register_artifact(self, config_path: str, src: str, verify_src_exists: bool = True):
+    def register_artifact(
+        self,
+        config_path: str,
+        src: str,
+        verify_src_exists: bool = True,
+        save_restore_connector: SaveRestoreConnector = SaveRestoreConnector(),
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/classes/modelPT.py
```diff
@@ -450,6 +234,7 @@ def restore_from(
         map_location: Optional[torch.device] = None,
         strict: bool = True,
         return_config: bool = False,
+        save_restore_connector: SaveRestoreConnector = SaveRestoreConnector(),
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/classes/modelPT.py
```diff
@@ -1143,7 +936,13 @@ def teardown(self, stage: str):
         super().teardown(stage)
 
     @classmethod
-    def extract_state_dict_from(cls, restore_path: str, save_dir: str, split_by_module: bool = False):
+    def extract_state_dict_from(
+        cls,
+        restore_path: str,
+        save_dir: str,
+        split_by_module: bool = False,
+        save_restore_connector=SaveRestoreConnector(),
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/collections/common/models/mock.py
```diff
@@ -0,0 +1,50 @@
...
+
+from typing import Dict, Union
+
+import torch
+from omegaconf import DictConfig
+
+from nemo.core.classes import ModelPT
+
+
+class MockModel(ModelPT):
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+    def _default_save_to(self, model, save_path: str):
+        """
+			Saves model instance (weights and configuration) into .nemo file.
+			You can use "restore_from" method to fully restore instance from .nemo file.
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+        self,
+        restore_path: str,
+        override_config_path: Optional[Union[OmegaConf, str]] = None,
+        map_location: Optional[torch.device] = None,
+        strict: bool = True,
+        return_config: bool = False,
+    ):
+        """
+		Restores model instance (weights and configuration) into .nemo file
+		Args:
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                        self._save_state_dict_to_disk(state_dict_subset, filepath)
+
+                logging.info(f'Checkpoints from {restore_path} were successfully extracted into {save_dir}.')
+            finally:
+                os.chdir(cwd)
+
+        return state_dict
+
+    def _register_artifact(self, model, config_path: str, src: str, verify_src_exists: bool = True):
+
```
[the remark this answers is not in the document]
ericharper: This docstring is nearly identical to the ModelPT one. Should we duplicate it here?

nemo/core/connectors/save_restore_connector.py
```diff
@@ -0,0 +1,325 @@
...
+                        # Update artifacts registry
+                        new_artiitem = model_utils.ArtifactItem()
+                        new_artiitem.path = "nemo:" + artifact_uniq_name
+                        new_artiitem.path_type = model_utils.ArtifactPathType.TAR_PATH
+                        model.artifacts[conf_path] = new_artiitem
+            finally:
+                # change back working directory
+                os.chdir(cwd)
+
+    def _update_artifact_paths(self, model, path2yaml_file):
```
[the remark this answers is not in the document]
ericharper: Does this need a docstring? There are many methods like this in NeMo that do not have one.

nemo/utils/app_state.py
```diff
@@ -62,6 +62,8 @@ def __init__(self):
         self._tmpdir_name = None
         self._model_config_yaml = "model_config.yaml"
         self._model_weights_ckpt = "model_weights.ckpt"
+        self._model_is_restored = False
```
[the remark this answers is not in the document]
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
ericharper: Done.

titu1994 (changes requested):
Minor changes necessary but it's close to mergeable

nemo/core/connectors/save_restore_connector.py
ericharper: Done.

nemo/core/connectors/save_restore_connector.py
ericharper: Added a method. Please let me know if was what you had in mind.

tests/core/test_save_restore.py
```diff
@@ -25,6 +25,7 @@
 from nemo.collections.nlp.models import PunctuationCapitalizationModel, TransformerLMModel
 from nemo.core.classes import ModelPT
 from nemo.utils.app_state import AppState
+from nemo.utils.models import MockModel
```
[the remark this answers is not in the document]
ericharper: It didn't work when it was in tests. I think we can keep core detached from ModelPT but still have a MockModel for testing.

nemo/collections/nlp/parts/nlp_overrides.py
titu1994: Yes

ericharper: Removed it for now. I want to do the NLPModel refactor in a seperate PR as this PR is getting too big already.

zeyuznv (changes requested):
Will unblock this MR after I finish testing it

nemo/core/connectors/save_restore_connector.py
tkornuta-nvidia: Som, I understand that you need to retrieve config - this is ok. I am just saying that I suggest moving config retrieval to a separate method. This will increase the modularity of Connector.

tkornuta-nvidia:
hey @ericharper , I like this PR, starting to look good!

Still,  I don't understand why register_artifacts need to set the connector? can we imagine a situation when a model is partially loaded using one connector and along that path, during artifacts registration, you change the way you load the model?

My mental model or "minimum code that works" is:
a) a class attribute of ModelPT storing connector
b) a class setter of ModelPT for setting the connector
Do we need more?

nemo/core/connectors/save_restore_connector.py
titu1994: Not really. The connectors job is to save and restore. Restoring the config only is also it's job.

tests/core/test_save_restore.py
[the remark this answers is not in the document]
titu1994: Why didn't it work? All tests passed before. I'm saying copy the model code info whatever file needs it in tests and duplicate it several times. That should work. 

I don't want model stuff inside utils - model imports a lot of utils too and it's circular. + It complicates detaching model pf

zeyuznv (approved):

nemo/core/classes/modelPT.py
ericharper: I added this and we were able to remove the connector arg from register_artifact.

nemo/core/classes/modelPT.py
```diff
@@ -157,7 +149,12 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # ModelPT wrappers over subclass implementations
         self.training_step = model_utils.wrap_training_step(self.training_step)
 
-    def register_artifact(self, config_path: str, src: str, verify_src_exists: bool = True):
+    def __init_subclass__(cls) -> None:
+        cls._save_restore_connector = SaveRestoreConnector()
```
titu1994, line 153: Don't you need a super call here ?

tests/core/test_save_restore.py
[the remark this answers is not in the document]
ericharper: Moved MockModel back to tests.

titu1994 (approved):
LGTM, ready to merge

PR #2597, opened 2021-08-03
import fix

ekmb:
fyi, @laituan245 


titu1994 (approved):
Has this patch been tested to make sure circular dependency is removed ?

ekmb:
> Has this patch been tested to make sure circular dependency is removed ?

yes, reproduced the user's issues and these changes resolve it

PR #2602, opened 2021-08-03
Allow using covering grammars for neural English TN model

laituan245:
- Added an option to allow users to use covering grammars to handle unrecoverable errors.
- Added a script for generating visualization files.


nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
```diff
@@ -50,6 +51,22 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # Language
         self.lang = cfg.get('lang', None)
 
+        # Covering Grammars
+        self.cg_normalizer = None  # Default
+        # We only support integrating with English TN covering grammars at the moment
+        self.use_cg = cfg.get('use_cg', False) and self.lang == constants.ENGLISH
+        if self.use_cg:
+            self.setup_cg(cfg.get('neural_confidence_threshold', 0.99))
```
yzhang123, line 59: setup_cgs

yzhang123: why setting up the grammars twice? first here and also in helpers?

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
```diff
@@ -50,6 +51,22 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+
+    # Setup covering grammars
+    def setup_cgs(self, neural_confidence_threshold):
+        self.use_cg = True
```
yzhang123, line 63: this should be always true before setup_cgs is called? so is this redundant?

yzhang123: add doc strings

yzhang123 (changes requested):

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
```diff
@@ -175,9 +192,36 @@ def _infer(
...
+                selected_toks_probs = torch.tensor(selected_toks_probs).to(self.device)
+                sequence_probs *= selected_toks_probs
+
+            # For TN cases where the neural model is not confident, use CGs
+            neural_confidence_threshold = self.neural_confidence_threshold
+            for ix, (_dir, _input, _prob) in enumerate(zip(input_dirs, input_centers, sequence_probs)):
+                if _dir == constants.INST_FORWARD and _prob < neural_confidence_threshold:
+                    if is_url(_input):
+                        _input = _input.replace(' ', '')  # Remove spaces in URLs
+                    cg_outputs = self.cg_normalizer.normalize(text=_input, verbose=False, n_tagged=1)
```
ekmb, line 222: Please add `n_tagged` to the argument list users can set

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
```diff
@@ -175,9 +192,36 @@ def _infer(
...
+                    generated_texts[ix] = list(cg_outputs)[0]
```
yzhang123, line 223: Did we discussed to add also the fall back to the input? Could you please add this?

yzhang123 (changes requested):

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: Yes. I will add that.

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: So models that were trained before this change do not have this field (and so its default value will always be `false` no matter what is set later in the config file). This is a hacky way to allow previously trained models to be able to use CGs. 

I will think about a cleaner way.

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: I guess the cg grammar will always return something (e.g. the original input if it can handle the input) ? I will double-check that. If that is true, I think the fall back to the input is implicitly handled here.

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
yzhang123: ok, thanks for explaining. That is ok then. We will delete later

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: I added that.

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: Added the doc strings.

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: I added code for fallback.

examples/nlp/duplex_text_normalization/conf/duplex_tn_config.yaml
```diff
@@ -75,6 +75,11 @@ decoder_model:
   nemo_path: ${decoder_exp_manager.exp_dir}/decoder_model.nemo # exported .nemo path
   lang: ${lang}
 
+  # Options related to covering grammars for TN
+  use_cg: false # Use covering grammars to avoid catastrophic errors
+  neural_confidence_threshold: 0.99 # If the neural model is not confident, then use the covering grammars
+  n_tagged: 1
```
ekmb, line 81: please add a comment on what this arg means

examples/nlp/duplex_text_normalization/analyze_errors.py
```diff
@@ -0,0 +1,233 @@
...
+        for errorcase in itn_error_cases:
+            f.write(errorcase.get_html())
+
+
+if __name__ == '__main__':
+    # Parse argument
+    parser = ArgumentParser()
+    parser.add_argument('--errors_log_fp', required=True)
+    parser.add_argument('--visualization_fp', required=True)
+    args = parser.parse_args()
```
ekmb, line 231: add help and types to the args + docstrings across the file

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
```diff
@@ -50,6 +51,27 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+        Setup covering grammars (if enabled).
+        :param cfg: Configs of the decoder model.
+        """
+        self.use_cg = True
+        self.neural_confidence_threshold = cfg.get('neural_confidence_threshold', 0.99)
+        self.n_tagged = cfg.get('n_tagged', 1)
+        input_case = 'cased'  # input_case is cased by default
+        if hasattr(self._tokenizer, 'do_lower_case') and self._tokenizer.do_lower_case:
+            input_case = 'lower_cased'
+        self.cg_normalizer = NormalizerWithAudio(input_case=input_case, lang=self.lang)
```
ekmb, line 73: add a PYNINI_AVAILABLE check like [here](https://github.com/NVIDIA/NeMo/blob/main/tests/nemo_text_processing/en/test_address.py#L25)

if it's not available and use_cg==True -> raise an error

ekmb (changes requested):

nemo/collections/nlp/models/duplex_text_normalization/duplex_decoder.py
laituan245: Yes. I added that.

examples/nlp/duplex_text_normalization/analyze_errors.py
laituan245: I added comments and docstrings.

examples/nlp/duplex_text_normalization/conf/duplex_tn_config.yaml
laituan245: I added the comment.

yzhang123 (approved):

PR #2616, opened 2021-08-05
 resolved MisconfigurationException

Qiliang:
end of epoch will throw execption:

pytorch_lightning.utilities.exceptions.MisconfigurationException: ModelCheckpoint(monitor='val_loss') not found in the returned metrics: ['loss', 'l_mle', 'l_length', 'logdet']. HINT: Did you call self.log('val_loss', value) in the LightningModule?

Signed-off-by: Qiliang <xiaoqlster@gmail.com>


blisc (approved):
Thanks for the fix!

PR #2620, opened 2021-08-05
CitriNet export fix

borisfom:


titu1994 (changes requested):

titu1994 (approved):
Nevermind, precheck takes care of that case. LGTM

PR #2629, opened 2021-08-09
Force fastpitch output to fp32

ryanleary:
This enables a much simpler deployment when doing mixed precision inference.

Signed-off-by: Ryan Leary <rleary@nvidia.com>


ryanleary:
Would be good to add this to the 1.2 branch as well if possible.

blisc (approved):

PR #2635, opened 2021-08-10
patch quantization

Slyne:
Update the code to be compatible with latest nemo.
With batchsize=1, now the WER/CER is normal.

But still, the exported onnx model doesn't contain seq_length, which affects models with ```conv_mask:True``` and when batch size > 1 and the samples length is different in a batch, the WER/CER will be degraded if tested by generated trt engine.


examples/asr/speech_to_text_infer.py
```diff
@@ -100,7 +100,7 @@ def main():
             reference = ''.join([labels_map[c] for c in seq_ids[0:seq_len]])
             references.append(reference)
         del test_batch
-
+    print("Ground Truth", references, "Predicted", hypotheses)
```
titu1994, line 103: This can be reverted, no need for batch of samples to be shown.

titu1994 (commented):
At first glance this seems excessive details are leaking about quantization compared to before. Is all of this because the model is not exporting length information ?

examples/asr/speech_to_text_infer.py
Slyne: Sorry. Just removed it.

Slyne:
> At first glance this seems excessive details are leaking about quantization compared to before. Is all of this because the model is not exporting length information ?

Yes. In addition to the previous problem, the generated model should not quantize its decoder but the original ```speech_to_text_quant_infer.py```  will make the decoder be quantized by calling ```quant_modules.initialize() ```.

PR #2643, opened 2021-08-12
Refactor and Minimize Dependencies

blisc:
Duplicate of #2518 to pass DCO bot


Jenkinsfile
```diff
@@ -72,7 +89,7 @@ pipeline {
         }
       }
       steps {
-        sh 'CUDA_VISIBLE_DEVICES="" pytest -m "not pleasefixme" --cpu --with_downloads --relax_numba_compat'
+        sh 'CUDA_VISIBLE_DEVICES="" pytest -m "not pleasefixme and not torch_tts" --cpu --with_downloads --relax_numba_compat'
```
titu1994, line 93: Is this `not torch_tts` really required?

nemo/collections/asr/models/__init__.py
```diff
@@ -12,11 +12,27 @@
...
-from nemo.collections.asr.models.asr_model import ASRModel
-from nemo.collections.asr.models.classification_models import EncDecClassificationModel
-from nemo.collections.asr.models.clustering_diarizer import ClusteringDiarizer
-from nemo.collections.asr.models.ctc_bpe_models import EncDecCTCModelBPE
-from nemo.collections.asr.models.ctc_models import EncDecCTCModel
-from nemo.collections.asr.models.label_models import EncDecSpeakerLabelModel, ExtractSpeakerEmbeddingsModel
-from nemo.collections.asr.models.rnnt_bpe_models import EncDecRNNTBPEModel
-from nemo.collections.asr.models.rnnt_models import EncDecRNNTModel
+# TODO @blisc: Perhaps refactor instead of import guarding
+try:
```
titu1994, line 16: Why is all of this required? I dont want this kind of guard over all domains, it makes NeMo un-maintanable, having to put shim classes in the except. This comment applies to every single place where such import guard is placed with shim in except case. Ordinary import guards are fine but not this shim style.

titu1994: Original prototype showed none of this was required and the torch tts block could be imported with no modifications to imports in other domains.

titu1994 (changes requested):
This PR has two major issues 

1) Import guard over ASR and NLP classes with Shims. This pattern is just unmaintanable over many classes and is bound to break either those domains or torch tts over time. Another thing is if TTS is undergoing refactor why should other domains even be refactored at all. They have no desire to decouple from pytorchlightning as of now, they should not be affected by this tts refactor.

2) This removal of tts imports from the tts `__init__` is going to make it much harder for NeMo users (those who want to use NeMo models with PTL) to actually import their models datasets etc. It enables torch_tts package at significant cost to usability of NeMo users, please think of some other way if possible to a avoid this.

nemo/collections/asr/models/__init__.py
blisc: Because tts depends on code from asr. So there are a few options:
1) What is done in the above file
2) Copy and pasting code and deleting all files that reference lightning, hydra, omegaconf
3) Delete all of the offending imports from __init__.py. IE just have empty __init__.py's.

If you can think of anything else, let me know. IMO, all options are bad in their own way.

nemo/collections/tts/__init__.py
```diff
@@ -11,7 +11,3 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-import nemo.collections.tts.data
```
blisc, line 15: Un-delete and put inside a try, catch

titu1994 (approved):
Discusses and no perfect solution exists at the moment. Plan is to refactor out the common dependencies to a path that doesnt interact with ASR and NMT

nemo/collections/asr/models/__init__.py
titu1994: Agreed, lets refactor in a future PR

ericharper (approved):
LGTM. Thanks!

PR #2647, opened 2021-08-12
[BUGFIX] Revert NMT encoder name change

ericharper:
self._encoder -> self.encoder


MaximumEntropy (approved):
LGTM. Thanks for the fix :)

PR #2648, opened 2021-08-12
Fix incorrect masking that always included all samples.

mousebaiker:
The current nemo version of the tdnn masking performs comparison as follows:
```
    lens_mat = torch.arange(max_len).to(device)
    lens = lens * max_len
    mask = lens_mat[:max_len].unsqueeze(0) < lens.unsqueeze(1)
```

If we expand the comparison and simplify device and unsqueeze calls , we get

```
torch.arange(max_len)[:max_len] < lens * max_len
```

This will be always True as long as each element of `lens` is larger than 1.  The multiplication by `max_len` is not needed. 

The bug can also be confirmed by running the test present in the PR after undoing the fix. 


nemo/collections/asr/parts/submodules/tdnn_attention.py
```diff
@@ -68,7 +68,6 @@ def lens_to_mask(lens: List[int], max_len: int, device: str = None):
         num_values: sum of mask values for each feature (useful for computing statistics later)
     """
     lens_mat = torch.arange(max_len).to(device)
-    lens = lens * max_len
```
mousebaiker, line 71: This is an offending line. 

Sorry for the noise from the formatter.

nithinraok:
Thanks for the fix Aleksei. 
Can you also change `length=None` at this [line](https://github.com/NVIDIA/NeMo/blob/b10bb28c5165d57ba18fd4a9e3807e6832f993d7/nemo/collections/asr/models/label_models.py#L196). When I upload the retrained model, I will revert this back. Thanks

nithinraok:
fixed with https://github.com/NVIDIA/NeMo/pull/2768

PR #2656, opened 2021-08-13
Fix nmt grpc server, concatdataset for raw text files

MaximumEntropy:
This PR fixes

1) NMT grpc server not setting provided source and target langs correctly
2) Issue with using a concatenation of multiple raw text files for training NMT models


ericharper (dismissed):
LGTM. Thanks!

ericharper (approved):
LGTM. Thanks!

PR #2659, opened 2021-08-14
CI Speedups for NMT

MaximumEntropy:
This PR attempts to speed up CI for NMT models.


ericharper (dismissed):
LGTM! Thanks!

Do you know how big the toy validation data is? We could probably truncate it if it's taking some time to tokenize.

titu1994 (dismissed):
There are other NLP ones, and Megatron ones. Will @ericharper  be looking at those ?

MaximumEntropy:
@ericharper  Yup! This is a still draft PR I still have a few more things to do 1) Our pre-LN CI model is 500M params for no good reason I'll create a really small one 2) Post-LN model is also ~200M params so I'll make all of those just 2 layer skinny models 3) Like you pointed out, our data is too big (I think the infer script runs on ~2k lines, I'll change it to like 8-10 lines only)

@titu1994 I spoke to @vadam5 about cutting time for the entity linking model as well (I think that takes about 1min 30secs right now).

titu1994:
That is amazing ! Yes please take the time so that tests cover the cases without super massive models (500 M !?)

okuchaiev:
is this still WIP?

MaximumEntropy:
Hey @okuchaiev sorry I haven't worked on it in a while but yes it is, I'll get back to it next week.

PR #2667, opened 2021-08-16
Dataset changes to allow mixed representation training for TTS

jfsantos:
This PR adds a parameter `phoneme_prob` to the `vocab` section of TTS datasets. That parameter sets the probability of using phonemes vs. graphemes during training. When using graphemes, the whole G2P pipeline is skipped and the model is presented with graphemes only. Functionality when using phonemes is not affected, and the default value of `phoneme_prob` is 1 so older config files will not be affected by this change.


blisc:
> This PR adds a parameter `phoneme_prob` to the `vocab` section of TTS datasets. That parameter sets the probability of using phonemes vs. graphemes during training. When using graphemes, the whole G2P pipeline is skipped and the model is presented with graphemes only. Functionality when using phonemes is not affected, and the default value of `phoneme_prob` is 1 so older config files will not be affected by this change.

Are you sure this will do what you want? This dataset caches the results of the G2P in memory. For each sample, you are only sampling once and then caching whether to store the text as phones or graphemes.

jfsantos:
> > This PR adds a parameter `phoneme_prob` to the `vocab` section of TTS datasets. That parameter sets the probability of using phonemes vs. graphemes during training. When using graphemes, the whole G2P pipeline is skipped and the model is presented with graphemes only. Functionality when using phonemes is not affected, and the default value of `phoneme_prob` is 1 so older config files will not be affected by this change.
> 
> Are you sure this will do what you want? This dataset caches the results of the G2P in memory. For each sample, you are only sampling once and then caching whether to store the text as phones or graphemes.

I was aware of the caching. Although it's not ideal, I did not see a clean way of disabling caching during training, but I assumed that if the dataset is large enough, the model will still get a reasonable amount of samples in either representation. I trained a FastPitch model using it and its performance using characters-only is way better than one trained using the current approach with G2P.

blisc (changes requested):
Minor comment

blisc (approved):

okuchaiev:
@jfsantos the PR fails on code formatting

blisc:
Closing this PR, we are deprecating nemo/collections/asr/data/vocabs.py in favour of moving it to the `torch_tts` collection in #2792. Please reopen and build on top on #2792 if desired.

PR #2672, opened 2021-08-16
Fix max symbol per step updating for RNNT

titu1994:
# Bugfix 
- Fix an issue when changing the decoding config from the dataclass instead of the yaml config
- Fix notebook comment style 

### DIFF
--- a/nemo/collections/asr/metrics/rnnt_wer.py
+++ b/nemo/collections/asr/metrics/rnnt_wer.py
@@ -106,7 +106,9 @@ class AbstractRNNTDecoding(ABC):
                 decoder_model=decoder,
                 joint_model=joint,
                 blank_index=self.blank_id,
-                max_symbols_per_step=self.cfg.greedy.get('max_symbols', None),
+                max_symbols_per_step=(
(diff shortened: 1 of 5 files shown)


nemo/collections/asr/metrics/rnnt_wer.py
```diff
@@ -106,7 +106,9 @@ def __init__(self, decoding_cfg, decoder, joint, blank_id: int):
                 decoder_model=decoder,
                 joint_model=joint,
                 blank_index=self.blank_id,
-                max_symbols_per_step=self.cfg.greedy.get('max_symbols', None),
+                max_symbols_per_step=(
+                    self.cfg.greedy.get('max_symbols', None) or self.cfg.greedy.get('max_symbols_per_step', None)
```
VahidooX, line 110: +What is the difference between max_symbols and max_symbols_per_step?

VahidooX (changes requested):

nemo/collections/asr/metrics/rnnt_wer.py
titu1994: No difference. dataclass requires `max_symbors_per_step` and the config requires `max_symbols`.

VahidooX (approved):

PR #2675, opened 2021-08-17
fix for emojis

ekmb:

### DIFF
--- a/nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
+++ b/nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Any, Dict, Optional
+from typing import Dict, Optional
 
 import numpy as np
 
@@ -59,6 +59,11 @@ def get_features(
 
         for j, word in enumerate(words):
             word_tokens = tokenizer.text_to_tokens(word)
+
+            # to handle emojis that could be neglected during tokenization
+            if len(word.strip()) > 0 and len(word_tokens) == 0:
+                word_tokens = [tokenizer.ids_to_tokens(tokenizer.unk_id)]
+
             subtokens.extend(word_tokens)
 
             loss_mask.append(1)
--- a/nemo/collections/nlp/data/token_classification/token_classification_dataset.py
+++ b/nemo/collections/nlp/data/token_classification/token_classification_dataset.py
@@ -87,6 +87,11 @@ def get_features(
 
         for j, word in enumerate(words):
             word_tokens = tokenizer.text_to_tokens(word)
+
+            # to handle emojis that could be neglected during tokenization
+            if len(word.strip()) > 0 and len(word_tokens) == 0:
+                word_tokens = [tokenizer.ids_to_tokens(tokenizer.unk_id)]
+
             subtokens.extend(word_tokens)
 
             loss_mask.append(1)
--- a/nemo/collections/nlp/models/nlp_model.py
+++ b/nemo/collections/nlp/models/nlp_model.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import glob
 import hashlib
 import json
 import os
@@ -40,7 +39,6 @@ from nemo.core.classes import ModelPT
 from nemo.core.classes.exportable import Exportable
 from nemo.core.connectors.save_restore_connector import SaveRestoreConnector
 from nemo.utils import AppState, logging
-from nemo.utils.exp_manager import configure_checkpointing
 from nemo.utils.get_rank import is_global_rank_zero
 
 __all__ = ['NLPModel']
--- a/nemo/collections/nlp/modules/common/lm_utils.py
+++ b/nemo/collections/nlp/modules/common/lm_utils.py
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
-from dataclasses import dataclass
 from typing import List, Optional, Union
 
 from attr import asdict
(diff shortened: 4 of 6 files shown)


nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
```diff
@@ -59,6 +59,11 @@ def get_features(
 
         for j, word in enumerate(words):
             word_tokens = tokenizer.text_to_tokens(word)
+
+            # to handle emojis that could be neglected during tokenization
+            if len(word.strip()) > 0 and len(word_tokens) == 0:
+                word_tokens = [tokenizer.ids_to_tokens(tokenizer.unk_id)]
```
titu1994, line 65: Does this work for say consecutive multiple emojis (without space or something to seperate as "word")

nemo/collections/nlp/data/token_classification/token_classification_dataset.py
```diff
@@ -87,6 +87,11 @@ def get_features(
 
         for j, word in enumerate(words):
             word_tokens = tokenizer.text_to_tokens(word)
+
+            # to handle emojis that could be neglected during tokenization
+            if len(word.strip()) > 0 and len(word_tokens) == 0:
+                word_tokens = [tokenizer.ids_to_tokens(tokenizer.unk_id)]
```
titu1994, line 93: Same as above. Basically we need multiple unk tokens - 1 for every emoji (even if they are consecutive)

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -40,7 +39,6 @@
 from nemo.core.classes.exportable import Exportable
 from nemo.core.connectors.save_restore_connector import SaveRestoreConnector
 from nemo.utils import AppState, logging
-from nemo.utils.exp_manager import configure_checkpointing
```
titu1994, line 43: Why are these being removed? Not  needed?

nemo_text_processing/inverse_text_normalization/ru/taggers/tokenize_and_classify.py
```diff
@@ -114,7 +114,6 @@ def __init__(self, cache_dir: str = None, overwrite_cache: bool = False):
 
             graph = delete_space + graph + delete_space
             self.fst = graph.optimize()
-            generator_main(far_file, {"tokenize_and_classify": self.fst})
```
titu1994, line 117: Intentionally removed?

nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
ekmb: to extract each emoji we would need to use a package like [ this ](https://pypi.org/project/emoji/)
I do not think it's general enough to add in the processing given the fact that emojies are missing from HF LM vocabs.

ekmb: @carolmanderson any opinion on that?

nemo/collections/nlp/data/token_classification/token_classification_dataset.py
ekmb: see the above

nemo/collections/nlp/models/nlp_model.py
ekmb: unused imports

nemo_text_processing/inverse_text_normalization/ru/taggers/tokenize_and_classify.py
ekmb: yes this lines is duplicated a few lines below

nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
titu1994: Ok so for now, if we dont use that package, a string of consecutive emojis would flatten out to a single unk token. As long as that doesnt hurt the model during training or inference, I am ok with that

carolmanderson: I think it's good as written. If we have a "word" consisting of multiple emojis, and it fails to generate any subtokens,  I think it's fine to replace that whole word with UNK. If users want to split up runs of multiple emojis, they can do that as a preprocessing step.

titu1994 (dismissed):

nemo/collections/nlp/data/intent_slot_classification/intent_slot_classification_dataset.py
titu1994: Sounds good. Approved

nemo/collections/nlp/modules/common/lm_utils.py
```diff
@@ -58,6 +58,7 @@ def get_lm_model(
     config_file: Optional[str] = None,
     checkpoint_file: Optional[str] = None,
     vocab_file: Optional[str] = None,
+    model_is_being_restored: Optional[bool] = False,
```
ericharper, line 61: I don't think we need this. `AppState` has the restore flag and you check for it there.

ericharper (changes requested):
Can you try to use `AppState` instead of passing an arg?

nemo/collections/nlp/modules/common/lm_utils.py
ericharper: https://github.com/NVIDIA/NeMo/blob/1d262c725df150ba837146e7899d8167adf524f6/nemo/utils/app_state.py#L381

titu1994: Agreed, I'd rather not pass this flag in every single method call, appstate is better for this inside the actual method in one place only.

titu1994 (changes requested):

ericharper (approved):
LGTM. Thanks!

titu1994 (approved):
LGTM

PR #2681, opened 2021-08-17
Add CarneliNet training configs

mousebaiker:
The config is the final part of the addition of [CarneliNet](https://arxiv.org/abs/2107.10708) model to NeMo repository. 

The base model is chosen to be 386, providing a good balance between size and quality. The description on what has to be changed to obtain models of different sizes should be enough to reproduce the results.


examples/asr/conf/carnelinet/carnelinet_384.yaml
```diff
@@ -0,0 +1,259 @@
...
+
+  encoder:
+    _target_: nemo.collections.asr.modules.ParallelConvASREncoder
+    feat_in: *n_mels
+    activation: swish
+    init_mode: tds_uniform
+    conv_mask: true
+
+    jasper:
+      - filters: 384
```
titu1994, line 96: Can you make filters a model_defaults variable too ?

examples/asr/conf/carnelinet/carnelinet_384.yaml
```diff
@@ -0,0 +1,259 @@
...
+        stride: [1]
+        dilation: [1]
+        dropout: ${model.model_defaults.dropout}
+        residual: true
+        separable: ${model.model_defaults.separable}
+        se: ${model.model_defaults.se}
+        se_context_size: ${model.model_defaults.se_context_size}
+        se_repeat: ${model.model_defaults.se_repeat}
+
+      - filters: &enc_final 640
```
titu1994, line 186: This can be in model_defaulfs.enc_hidden

examples/asr/conf/carnelinet/carnelinet_384.yaml
```diff
@@ -0,0 +1,259 @@
...
+  max_epochs: 100
+  max_steps: null # computed at runtime if not set
+  num_nodes: 1
+  accelerator: ddp
+  accumulate_grad_batches: 1
+  checkpoint_callback: false  # Provided by exp_manager
+  logger: false  # Provided by exp_manager
+  log_every_n_steps: 50  # Interval of logging.
+  val_check_interval: 1.0 # Set to 0.25 to check 4 times per epoch, or an int for number of iterations
+  precision: 16 # Train with AMP to speed and memory efficiency
```
titu1994, line 234: Let's put precision = 32 as default but add comment that precision is 16 is preferred if available. I need to do that for CongextNet config too

examples/asr/conf/carnelinet/carnelinet_384.yaml
```diff
@@ -0,0 +1,259 @@
...
+  benchmark: false
+
+exp_manager:
+  exp_dir: null
+  name: *name
+  create_tensorboard_logger: true
+  create_checkpoint_callback: true
+  checkpoint_callback_params:
+    monitor: "val_wer"
+    mode: "min"
```
titu1994, line 245: Can you add always_save_nemo = True flag here. We should make it the default.

titu1994 (commented):
Overall looks great ! Can be slightly refactored for even more global level flexibility

examples/asr/conf/carnelinet/carnelinet_384.yaml
mousebaiker: Done.

examples/asr/conf/carnelinet/carnelinet_384.yaml
mousebaiker: Chose a different name.

examples/asr/conf/carnelinet/carnelinet_384.yaml
mousebaiker: Done.

examples/asr/conf/carnelinet/carnelinet_384.yaml
mousebaiker: Done.

mousebaiker:
Thanks for the review. PTAL.

titu1994 (approved):
Looks Awesome !

Issue #2544, opened 2021-07-23
Error in create_tarred_transformer_lm_dataset.py


pehonnet:
**Describe the bug**

When trying to prepare data for ASR neural LM training, using the script `scripts/asr_language_modeling/neural_rescorer/create_tarred_transformer_lm_dataset.py` getting the following error (and a few preceding log lines):
```
[Parallel(n_jobs=-2)]: Done  68 out of 128 | elapsed:   10.6s remaining:    9.3s
[Parallel(n_jobs=-2)]: Done  81 out of 128 | elapsed:   11.4s remaining:    6.6s
[Parallel(n_jobs=-2)]: Done  94 out of 128 | elapsed:   12.0s remaining:    4.3s
[Parallel(n_jobs=-2)]: Done 107 out of 128 | elapsed:   12.6s remaining:    2.5s
[Parallel(n_jobs=-2)]: Done 120 out of 128 | elapsed:   13.5s remaining:    0.9s
[Parallel(n_jobs=-2)]: Done 128 out of 128 | elapsed:   14.1s finished
Traceback (most recent call last):
  File "./scripts/asr_language_modeling/neural_rescorer/create_tarred_transformer_lm_dataset.py", line 318, in <module>
    main()
  File "./scripts/asr_language_modeling/neural_rescorer/create_tarred_transformer_lm_dataset.py", line 299, in main
    chunk_paths, chunk_lens = __tokenize_text(
  File "./scripts/asr_language_modeling/neural_rescorer/create_tarred_transformer_lm_dataset.py", line 174, in __tokenize_text
    chunk = np.asarray(chunk, dtype=np.int64)
  File "/home/pe.honnet/Tools/miniconda3/envs/nemo/lib/python3.8/site-packages/numpy/core/_asarray.py", line 102, in asarray
    return array(a, dtype, copy=False, order=order)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
Read 1000000 lines: : 1048576 lines [00:18, 57331.13 lines/s]
```

**Steps/Code to reproduce bug**
Command run as follows
```bash
python ./scripts/asr_language_modeling/neural_rescorer/create_tarred_transformer_lm_dataset.py \
  --text_path=training_data.txt \
  --data_root=./data/prep_nn_lm \
  --tokenizer_model=conformer_ctc_large/tokenizer/tokenizer.model \
  --tokenizer_vocab_file=bert_uncased_vocab.txt \
[...]

PR #2690, opened 2021-08-18
LM training, support for SentencePiece and HF tokenizers 

AlexGrinch:
1. Added support for SentencePiece and HF tokenizers when training LMs to fix the bug (https://github.com/NVIDIA/NeMo/issues/2544#issuecomment-899922327)
2. Removed obsolete text post-processing in AutoTokenizer


examples/nlp/machine_translation/create_tarred_monolingual_dataset.py
```diff
@@ -19,8 +19,10 @@
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='NMT dataset pre-processing')
-    parser.add_argument('--tokenizer_model', type=str, required=True, help='Path to tokenizer model')
-    parser.add_argument('--tokenizer_name', type=str, default='yttm', help='BPE Tokenizer Name, Options: [yttm]')
+    parser.add_argument(
+        '--tokenizer_name', type=str, default='yttm', help='Supports yttm, sentencepiece and HuggingFace tokenizers',
+    )
+    parser.add_argument('--tokenizer_model', type=str, default=None, help='Path to tokenizer model')
```
MaximumEntropy, line 25: I think having the tokenizer_model is necessary and it isn't possible to train a tokenizer on the provided data right now correct? If so, I think we should make this a required argparse field.

nemo/collections/common/tokenizers/huggingface/auto_tokenizer.py
```diff
@@ -203,7 +172,7 @@ def text_to_tokens(self, text):
 
     def tokens_to_text(self, tokens):
         text = self.tokenizer.convert_tokens_to_string(tokens)
-        return remove_spaces(handle_quotes(text.strip()))
+        return text
```
MaximumEntropy, line 175: High-level question about changes to this entire file - are we sure it doesn't break things for others using HF tokenizers?

MaximumEntropy (approved):
LGTM. Just a couple of minor comments/questions.

nemo/collections/common/tokenizers/huggingface/auto_tokenizer.py
AlexGrinch: The initial reason for all this post-processing was to correctly process all the punctuation after bert-base-uncased detokenization 2 years ago when we tried to fine-tune it on En-De NMT. After that the tokenization/detokenization in HF improved significantly.

examples/nlp/machine_translation/create_tarred_monolingual_dataset.py
AlexGrinch: The reason for removing `required` flag is to support HF tokenisers out of the box. They are usually created by passing corresponding HF model name to `tokenizer_name`.

Alternative design choice is to pass something like `hf` or `hugging_face` as `tokenizer_name` and expecting `tokenizer_model` to be the name of the corresponding HF model.

nemo/collections/common/tokenizers/huggingface/auto_tokenizer.py
MaximumEntropy: I see, makes sense. Thanks for clarifying!

PR #2701, opened 2021-08-20
Rnnt greedy_batch decoding fix

titu1994:
# Bugfix
- Fixes a bug with an extreme edge case thereby bringing 1:1 parity with sample wise greedy decoding.

# Case
- Timestep T = 0, Label ID U = 0
- Batch size > 1 and greedy_batch decoding is used
- Some samples of the first batch emit a label but other samples do not. Therefore for some samples T=0, U=1 and other samples T=1, U=0 step should occur.
- To accommodate T=0, U=1 case, the model steps forward along label axes, but forgets to reset the state of the samples who should have followed the T=1, U=0 case. 
- As such, the state of the T=1, U=0 samples degrades to an incorrect value, and ruins the first subword predicted by the samples of that batch.
- This has significant impact on WER but minute effect on CER. 

# Fix notes
- Cover the edge case where T= 0, U= 0 explicitly
- Check if there is a mix of label forward and time forward actions and explicitly reset the states (to 0) of the samples where time forward action should take place.

### DIFF
--- a/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
@@ -535,15 +535,20 @@ class GreedyBatchedRNNTInfer(_GreedyRNNTInfer):
                                     alignments[batch_idx].append([])  # blank buffer for next timestep
                     else:
                         # Collect batch indices where blanks occurred now/past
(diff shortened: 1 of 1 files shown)


VahidooX (approved):

VahidooX (approved):
LGTM. There are just some LGTM warnings.

PR #2705, opened 2021-08-21
Add mAES beam search for Transducer models

titu1994:
# Changelog
- Add mAES Transducer algorithm for further improved WER at the cost of speed of execution. 
- Update configs with new decoding scheme
- Begin adding hooks for external LM rescoring

# References
- [Accelerating RNN Transducer Inference via Adaptive Expansion Search](https://ieeexplore.ieee.org/document/9250505)

Ported from ESPNet v1

### DIFF
--- a/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
@@ -179,6 +179,7 @@ class BeamRNNTInfer(Typing):
         self.score_norm = score_norm
 
         if self.beam_size == 1:
+            logging.info("Beam size of 1 was used, switching to sample level `greedy_search`")
             self.search_algorithm = self.greedy_search
         elif search_type == "default":
             self.search_algorithm = self.default_beam_search
@@ -268,10 +269,16 @@ class BeamRNNTInfer(Typing):
                 # during the beam loop.
                 with self.decoder.as_frozen(), self.joint.as_frozen():
 
+                    _p = next(self.joint.parameters())
+                    dtype = _p.dtype
+
                     # Decode every sample in the batch independently.
                     for batch_idx in idx_gen:
                         inseq = encoder_output[batch_idx : batch_idx + 1, :, :]  # [1, T, D]
                         logitlen = encoded_lengths[batch_idx]
+
+                        if inseq.dtype != dtype:
+                            inseq = inseq.to(dtype=dtype)
 
                         # Execute the specific search strategy
                         nbest_hyps = self.search_algorithm(inseq, logitlen)  # sorted list of hypothesis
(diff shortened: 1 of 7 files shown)


VahidooX (approved):

b-flo:
Hi @titu1994!

I'm the ESPnet maintainer for Transducer models. For mAES, could you reference the original [AES](https://ieeexplore.ieee.org/document/9250505) and also NSC (related paper soon-to-be-released) please? The modifications we made to AES are minor but motivated by some limitations we observed in OSC and were carried to the new algorithm:

- The original AES method relies on a 2-step expansion procedure which is done in cascade. Allowing a higher number of expansion steps for some tasks (e.g.: VIVOS, Voxforge IT, or ESTER2) is important to allow SOTA performance.
- If the algorithm is modified to allow N expansion steps, we need to deal with some edge cases when Nstep=1. The same condition we used for NSC can be added to mAES to reduce performance degradation but allowing Nstep>=2 instead is clearly debatable. Contrary to OSC/NSC, expansion in mAES is adaptive, meaning the number of expansions can vary at each step between {1, …, Nstep} and the algorithm can naturally deal with most edge cases. Thus, we decided to keep it like that but we may support both behaviors as the second one could lead to further improvement in terms of decoding speed.
- The modified N-step version is slightly more compact and memory-efficient compared to the original one as we keep track to only 2 sets instead of 4 and related combinations are removed.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -198,7 +239,27 @@ def __init__(
...
+        self.nsc_prefix_alpha = int(nsc_prefix_alpha)
+        self.maes_prefix_alpha = int(maes_prefix_alpha)
+        self.maes_num_steps = int(maes_num_steps)
+        self.maes_expansion_gamma = float(maes_expansion_gamma)
+        self.maes_expansion_beta = int(maes_expansion_beta)
+
+        if self.maes_prefix_alpha < 0:
+            raise ValueError("`maes_prefix_alpha` must be a positive integer.")
+
+        if self.maes_num_steps < 1:
```
b-flo, line 251: I would suggest changing the condition to only allow `maes_num_steps >=2` if you don't apply any expansion regularization during training. Otherwise, you'll need to allow incomplete batch hypotheses to be carried onto the next step in order to tackle performance degradation for some edge cases when `maes_num_steps = 1`. Additionally, you may want to rely on a counting method to compute an `auto_nstep` during the last validation phase to improve the condition, i.e. only allows incomplete hypothesis when `maes_num_steps = 1` and `auto_nstep = 1`

titu1994: Sounds good, will set the limit ! Thanks for the great implementation !

PR #2715, opened 2021-08-24
Merge 1.3 bugfixes into main

ericharper:


nemo/collections/tts/models/uniglow.py
```diff
@@ -21,7 +21,7 @@
 from pystoi import stoi
 from pytorch_lightning.loggers import LoggerCollection, TensorBoardLogger
 
-from nemo.collections.tts.helpers.helpers import OperationMode, waveglow_log_to_tb_func
+from nemo.collections.tts.helpers.helpers import OperationMode, split_view, waveglow_log_to_tb_func
```
blisc, line 24: I don't see `split_view` in `main`'s helpers.py. @ericharper do you know why helpers.py isn't being updated?

blisc (changes requested):

nemo/collections/tts/models/uniglow.py
ericharper: It was lost when merging, We can add back it in.

ekmb (approved):
The TN changes LGTM.

blisc (approved):

PR #2737, opened 2021-08-27
1. Updated perceiver architecture.

michalivne:
This PR updates the architecture of Perceiver-encoder.
The upgraded architecture:

1. Matches performance of non-bottleneck NMT models.
2. More stable to train.
3. All available model names are defined in a single place.


ericharper (approved):
LGTM. Thanks!

PR #2768, opened 2021-09-02
update masking and warning for old models

nithinraok:


nemo/collections/asr/models/label_models.py
```diff
@@ -193,7 +193,7 @@ def forward(self, input_signal, input_signal_length):
         )
 
         encoded, length = self.encoder(audio_signal=processed_signal, length=processed_signal_len)
-        logits, embs = self.decoder(encoder_output=encoded, length=length)
+        logits, embs = self.decoder(encoder_output=encoded, length=None)
```
tango4j, line 196: Is there any other use of length variable for decoder?
If there is an expected future use, you may leave it
otherwise setting it with default length=None would make it neater.

tango4j (commented):
A comment was added to self.decoder() part

tango4j (approved):

PR #2772, opened 2021-09-03
fix for HF model restoration

ekmb:

### DIFF
--- a/nemo/collections/nlp/modules/common/huggingface/huggingface_utils.py
+++ b/nemo/collections/nlp/modules/common/huggingface/huggingface_utils.py
@@ -34,7 +34,7 @@ from nemo.collections.nlp.modules.common.huggingface.distilbert import DistilBer
 from nemo.collections.nlp.modules.common.huggingface.roberta import RobertaEncoder
 from nemo.utils import logging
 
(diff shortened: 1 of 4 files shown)


nemo/collections/nlp/modules/common/huggingface/huggingface_utils.py
```diff
@@ -64,6 +64,13 @@
     },
 }
 
+VOCAB_FILE_NAME = {
```
ericharper, line 67: Can this be avoided? is it possible to return the path with `save_vocabulary`?

ekmb: The different models mean different things by "vocab_file", compare: https://github.com/huggingface/transformers/blob/master/src/transformers/models/albert/tokenization_albert.py#L33 vs https://github.com/huggingface/transformers/blob/master/src/transformers/models/bert/tokenization_bert.py#L32. I think it's better to preserve this info for the users

ericharper: It's fine to preserve it, but can we add the functionality to save_vocabulary. Also, what happens if the tokenizer_name is not in this dictionary?

ericharper (approved):
LGTM. Thanks!

PR #2773, opened 2021-09-03
Provide NMT gRPC models via a directory instead of individual files

MaximumEntropy:
Signed-off-by: MaximumEntropy <sandeep.subramanian.1@umontreal.ca>


okuchaiev (approved):
lgtm

PR #2776, opened 2021-09-03
Relative paths in tarred dataset

michalivne:
This PR Adds on-the-fly validation of tarred dataset files. 
Changes:
1. If file is missing, looking for the file in same path as metadata file.
2. If file cannot be found, raises an exception.


MaximumEntropy (approved):
LGTM. Thanks!

PR #2787, opened 2021-09-07
non-deterministic norm update

ekmb:


nemo_text_processing/text_normalization/en/data/whitelist.tsv
```diff
@@ -10017,7 +10017,6 @@ NNSS	n n s s
 NNSU	n n s u
 NNSW	n n s w
 NNW-	n n w
-No.	number
```
yzhang123, line 10020: why removing it from whitelist and moving to alternatives?

yzhang123 (approved):

yzhang123 (approved):

yzhang123 (approved):

PR #2798, opened 2021-09-09
NMT input ID validity test for negative values

michalivne:
This PR extends validity checks of src/tgt input IDs:

1. Added a test for special tokens being valid (i.e., non-negative) for Sentencepiece tokenizer.
2. Extended MTEncDecModel id validity test to include non-negativity.
3. Added `model.validate_input_ids` to `MTEncDecModel` config to allow disabling id validity (`true` by default).

### DIFF
--- a/nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
+++ b/nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
@@ -43,6 +43,16 @@ class SentencePieceTokenizer(TokenizerSpec):
             raise ValueError(f"model_path: {model_path} is invalid")
         self.tokenizer = sentencepiece.SentencePieceProcessor()
         self.tokenizer.Load(model_path)
+        # validate no token is negative
+        negative_tokens = []
+        for n in ["eos_id", "bos_id", "unk_id", "pad_id"]:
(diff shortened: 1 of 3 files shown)


nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
```diff
@@ -43,6 +43,16 @@ def __init__(
             raise ValueError(f"model_path: {model_path} is invalid")
         self.tokenizer = sentencepiece.SentencePieceProcessor()
         self.tokenizer.Load(model_path)
+        # validate no token is negative
```
aklife97, line 46: Can we move this check inside the NMT model? This SentencePiece wrapper is common and is used by ASR as well and this condition might not be true for them.

aklife97 (changes requested):
Looks good overall! Just that we need to move the test inside NMT

nemo/collections/common/tokenizers/sentencepiece_tokenizer.py
michalivne: Done

aklife97 (approved):
LGTM! Feel free to merge once CIs pass.

PR #2811, opened 2021-09-13
Increased the precision of validation metric to be saved in the file names.

VahidooX:
Currently we save the value of the validation metric into the name of the checkpoints upto 2 numbers after the decimal point. For some models like ASR where we have WER and numbers are less than 1, it can cause an issue. The experiment manager cannot find the right topk checkpoints.


blisc:
> Currently we save the value of the validation metric into the name of the checkpoints upto 2 numbers after the decimal point. For some models like ASR where we have WER and numbers are less than 1, it can cause an issue. The experiment manager cannot find the right topk checkpoints.

If you are measuring something like WER, doesn't it make more sense to just multiply your metric by 100?

nemo/utils/exp_manager.py
```diff
@@ -798,7 +798,7 @@ def configure_checkpointing(
     if params.dirpath is None:
         params.dirpath = Path(log_dir / 'checkpoints')
     if params.filename is None:
-        params.filename = f'{name}--{{{params.monitor}:.2f}}-{{epoch}}'
+        params.filename = f'{name}--{{{params.monitor}:.4f}}-{{epoch}}'
```
blisc, line 801: This is fine with me. @titu1994 , any thoughts?

VahidooX:
> 
> 
> > Currently we save the value of the validation metric into the name of the checkpoints upto 2 numbers after the decimal point. For some models like ASR where we have WER and numbers are less than 1, it can cause an issue. The experiment manager cannot find the right topk checkpoints.
> 
> If you are measuring something like WER, doesn't it make more sense to just multiply your metric by 100?

You are right, this can be another solution. I just thought maybe this way can be useful for other tasks too?

nemo/utils/exp_manager.py
titu1994: This is good. Multiplying by 100 is problematic cause you have to remember the scale up in downstream stuff.

titu1994 (approved):
I like this solution. It's more generic and doesn't depend on individual metrics to scale up and down.

blisc (approved):

PR #2818, opened 2021-09-14
Megatron GPT training with NeMo tokenizers

MaximumEntropy:


nemo/collections/nlp/data/language_modeling/megatron/autoaugment.py
```diff
@@ -1,284 +0,0 @@
-"""AutoAugment data augmentation policy for ImageNet.
```
ericharper, line 1: Do we need this file?

ericharper (approved):

PR #2839, opened 2021-09-17
HiFiGAN and Tacotron Finetuning

paarthneekhara:
Also has configurable duration and pitch loss coefficients for FastPitch


examples/tts/conf/hifigan/model/generator/v4.yaml
```diff
@@ -0,0 +1,8 @@
+# @package _group_
```
blisc, line 1: Isn't this file the same as v1?

nemo/collections/tts/models/hifigan.py
```diff
@@ -76,26 +76,29 @@ def configure_optimizers(self):
...
-        self.scheduler_d = CosineAnnealing(
-            optimizer=self.optim_d, max_steps=self._cfg.max_steps, min_lr=self._cfg.sched.min_lr,
-        )
-        sch2_dict = {
-            'scheduler': self.scheduler_d,
-            'interval': 'step',
-        }
-
-        return [self.optim_g, self.optim_d], [sch1_dict, sch2_dict]
+        if hasattr(self._cfg, 'sched'):
```
blisc, line 79: Doesn't your new _44100 yaml still have sched?

blisc (commented):
Where is the hifigan notebook?

nemo/collections/tts/models/hifigan.py
paarthneekhara: Oh we remove it by using ~optim.sched like you suggested from command line.

examples/tts/conf/hifigan/model/generator/v4.yaml
paarthneekhara: upsampling rates here are [8,8,4,2] vs [8,8,2,2] of v1. In the ADLR checkpoints it was [8,8,4,2], so I created a new one to keep it consistent. But we can just remove it and use v1..

blisc: got it

nemo/collections/tts/models/hifigan.py
blisc: ok

blisc:
Can you add your hifigan finetuning notebook?

PR #2846, opened 2021-09-17
Linear warmup with cosine annealing and constant holding

MaximumEntropy:


nemo/core/optim/lr_scheduler.py
```diff
@@ -170,7 +170,19 @@ class WarmupAnnealHoldPolicy(_LRScheduler):
...
+    def __init__(
+        self,
+        optimizer,
+        *,
+        warmup_steps=None,
+        warmup_ratio=None,
+        max_steps=None,
+        min_lr=0.0,
+        last_epoch=-1,
+        constant_steps=None,
```
titu1994, line 182: constant_* should go above max_steps in order of arguments.

nemo/core/config/schedulers.py
```diff
@@ -51,6 +51,20 @@ class WarmupHoldSchedulerParams(WarmupSchedulerParams):
     min_lr: float = 0.0
 
 
+@dataclass
+class WarmupAnnealingHoldSchedulerParams(SchedulerParams):
```
titu1994, line 55: Why not just extend `WarmupSchedulerParams(SchedulerParams)` and drop max_steps, warmup_steps and warmup_ratio

nemo/core/config/schedulers.py
```diff
@@ -72,13 +86,14 @@ class SquareRootAnnealingParams(WarmupSchedulerParams):
...
 @dataclass
-class CosineAnnealingParams(WarmupSchedulerParams):
+class CosineAnnealingParams(WarmupAnnealingHoldSchedulerParams):
     """
     Cosine Annealing parameter config
     It is not derived from Config as it is not a NeMo object (and in particular it doesn't need a name).
     """
 
     min_lr: float = 0.0
+    max_lr: float = 2.5e-4
```
titu1994, line 96: Revert. This should be set by the yaml config, not the default of the data class itself since it's domain dependent

nemo/core/optim/lr_scheduler.py
```diff
@@ -160,6 +160,87 @@ def get_lr(self):
...
+        constant_ratio=None,
+        max_lr=None,
+    ):
+        assert not (
```
titu1994, line 186: Perform these two assertions for constant_* too.

nemo/core/optim/lr_scheduler.py
```diff
@@ -232,15 +337,27 @@ def _get_lr(self, step):
                     f"{self} received an initial learning rate that was lower than the minimum learning rate."
                 )
 
-        new_lrs = [
-            _cosine_annealing(
```
titu1994, line 236: CosineAnnealing needs to extend WarmupAnnealHoldPolicy(_LRScheduler) + add the two new args and pass them to constructor

titu1994 (changes requested):
This looks great! Overall some refactor required with inheritance but looks great otherwise

ericharper (approved):
@MaximumEntropy to address @titu1994 comments when merging to main

PR #2847, opened 2021-09-18
Fix TransformerDecoder export

aklife97:
Fixes export for TransformerDecoder.


nemo/collections/nlp/modules/common/transformer/transformer.py
```diff
@@ -183,6 +184,10 @@ def __init__(
         self._vocab_size = vocab_size
         self._hidden_size = hidden_size
         self._max_sequence_length = max_sequence_length
+        self.num_states = num_layers + 1
```
MaximumEntropy, line 187: Is this something that needs to go in the TransformerConfig class or something that can be computed on the fly during export?

nemo/collections/nlp/modules/common/transformer/transformer.py
```diff
@@ -246,8 +264,18 @@ def input_example(self):
...
 
     def _prepare_for_export(self, **kwargs):
         self._decoder.diagonal = None
+        self.return_mems = True
         super()._prepare_for_export(**kwargs)
+
+    @property
+    def output_types(self) -> Optional[Dict[str, NeuralType]]:
+        if self.return_mems:
+            return {"last_hidden_states": NeuralType(('B', 'D', 'T', 'D'), ChannelType())}
```
MaximumEntropy, line 279: Shouldn't this be batch x num_layers x time x dim? Why is the dim axis present twice?

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
```diff
@@ -148,13 +148,23 @@ def __init__(
 
     def _get_memory_states(self, decoder_states, decoder_mems_list=None, i=0):
```
MaximumEntropy, line 149: Curious why decoder mems is still a list here?

MaximumEntropy (changes requested):
A few questions and nitpicks.

nemo/collections/nlp/modules/common/transformer/transformer.py
aklife97: NeMo does not have a AxisKind to represent num_layers specifically. While there is height, export considers that as a dynamic axis which might be worse to optimize. So AxisKind.Dimension looked like the best fit.

nemo/collections/nlp/modules/common/transformer/transformer_decoders.py
aklife97: I can change the var name to just decoder_mems if you feel that's better. In general, I didn't change any variable names to ensure none of the existing code breaks. But I don't think decoder_mems_list for this particular function (get_memory_states) is being used as a named parameter anywhere - so can change it if you wish.

okuchaiev:
I am wondering if we should send this to r1.4.0 branch? @ericharper

nemo/collections/nlp/modules/common/transformer/transformer.py
```diff
@@ -207,14 +212,28 @@ def __init__(
...
 
     @typecheck()
-    def forward(self, input_ids, decoder_mask, encoder_embeddings, encoder_mask):
-        decoder_embeddings = self._embedding(input_ids=input_ids)
+    def forward(
+        self, input_ids, decoder_mask, encoder_embeddings, encoder_mask, decoder_mems=None, return_mems=False,
+    ):
+        self.return_mems = return_mems
+        start_pos = 0
+        if decoder_mems != None:
```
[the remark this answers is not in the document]
aklife97: Done!

MaximumEntropy (approved):
LGTM. Thanks for making the changes.

PR #2856, opened 2021-09-20
Nmt encoder decoder hidden size fix

michalivne:
1. Enabled encoder/decoder with different size in bottleneck architecture.
2. Validating encoder/decoder are with the same hidden_size in non-bottleneck parent class.


MaximumEntropy (dismissed):
LGTM! Thanks :)

ericharper (approved):
LGTM. Thanks!

PR #2861, opened 2021-09-21
add zero shot intent model

carolmanderson:
- New model class ZeroShotIntentModel is a subclass of TextClassificationModel
- Allows training/inference with either three classes ("entailment", "contradiction", and "neutral") or two classes ("entailment" and "not_entailment")
- ZeroShotIntentDataset and ZeroShotIntentInferenceDataset are subclasses of GLUEDataset
- For inference, users can specify multi_class = True or False and can supply their own hypothesis template
- `list_available_models` is not yet implemented. Tutorial notebook also needs to be updated with the name(s) of pretrained models users can get from NGC. These do not exist yet.


okuchaiev (approved):
looks good to me. Thanks!

PR #2882, opened 2021-09-23
Remove numbers from TTS tutorial names

redoctopus:
No additional references in the repository that I could find with grep.


blisc (changes requested):
Can you also update this page: https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/starthere/tutorials.html

redoctopus:
Ah, looks like the documentation links weren't updated the last time the tutorials were renamed. Updated--can you check if the order is fine?

blisc (approved):

PR #2885, opened 2021-09-23
add fix to not add dot everywhere

yzhang123:


ekmb (approved):

nemo_text_processing/inverse_text_normalization/en/data/whitelist.tsv
```diff
@@ -1,5 +1,4 @@
 e.g.	for example
-i.e.	that is
```
Oktai15, line 2: I would suggest to leave it unchanged, because it is not case for all datasets. @ekmb liked idea with argument to Normalizer class as path to file with whitelist.tsv, so anyone can change it for specific dataset

yzhang123: this is only for ITN, not TN

PR #2894, opened 2021-09-25
(1) nvfuser, (2) amp-casting decoration

erhoo82:
1. Remove legacy fuser setting
2. Add amp precision casting decorator (use input precision as is)


ericharper (approved):
LGTM. Thanks!

PR #2904, opened 2021-09-27
Add DALI dataset unit test

jantonguirao:
Adds a new unit tests that verifies that using DALI for feature extraction produces the same results as the default implementation.


titu1994 (dismissed):
Looks great, thanks !

titu1994:
Could you check out the style fixes and dco bot instructions?

PR #2918, opened 2021-09-28
Merge r1.4 bugfixes to main

ericharper:


requirements/requirements_lightning.txt
```diff
@@ -1,4 +1,4 @@
-pytorch-lightning>=1.4.0
+pytorch-lightning>=1.4.8,<1.5
```
blisc, line 1: Let's remove the `<1.5` for main

blisc (commented):
TTS looks good to go

requirements/requirements_lightning.txt
ericharper: We know that 1.5 has breaking API changes. We should probably relax that after we merge those fixes into main?

aklife97:
Transformer Decoder changes look good!

requirements/requirements_lightning.txt
blisc: But how will I know when lightning gets updated to 1.5 if there are no complaints of bugs in our slack channel?

titu1994 (approved):
LGTM

ericharper:
Removing upper bound on PTL after api fixes are in main.

PR #2923, opened 2021-09-29
Adding simple, greedy-search based text generation to MegatronGPTModel

okuchaiev:
This PR adds the following method ```response=MegatronGPTModel.complete(request)``` and example how to use it.
Example usage:
```
python examples/nlp/language_modeling/megatron_gpt_eval.py --model_file megatron_gpt.nemo \
--prompt "How to fix gpu memory? A: " --tokens_to_generate 256
```
With a somewhat good converged model you should be getting completion like this:
```
How to fix gpu memory? A: 

You can use the gpu_mem_info() function to get the information about the memory.
```


okuchaiev:
will open an new one

PR #2929, opened 2021-09-29
TN infer 

ekmb:
- fixes inference for utterances longer than max_seq_len
- adds additional post-processing to match punct/spaces of the original input 
- adds `whitelist` arg to WFST normalize.py

@Oktai15 for viz


nemo/collections/nlp/data/text_normalization/utils.py
```diff
@@ -161,3 +163,46 @@ def basic_tokenize(input_str, lang):
     if lang == constants.ENGLISH:
         return word_tokenize(input_str)
     return input_str.strip().split(' ')
+
+
+def post_process_punct(input: str, output: str):
+    """
+    Post-process punctuation,
```
yzhang123, line 170: could you describe more what this function is doing? that is working with spaces? Also it looks like the output and "-> yesterday's agenda (april seventh two thousand one) disappointments of the same ol' routine" is the same? what is the difference here?

yzhang123 (changes requested):

yzhang123 (approved):

PR #2936, opened 2021-10-01
Experiment manager step timing

michalivne:
This PR adds the ability to log the timing of a step for train/validation/test modes during training.
This feature is added to all NeMo models.
Time is measured in seconds.

# Configuration

The following config allow to control logging of step timing:

```
exp_manager:
  log_step_timing: True # logging is enabled by default
  step_timing_kwargs:
    reduction: "mean" # reduction for multiple measurements
    sync_cuda: False # if True calls torch.cuda.synchronize() for start/stop of measurement. Disabled by default
```

By default the logging is turned on. In order to turn off the feature, the following config can be used:

```
exp_manager.log_step_timing=False
```

By default cuda sync is turned off. In order to turn it on, the following config can be used:
```
exp_manager.step_timing_kwargs.sync_cuda=True
```

# Logging

The following keys will be logged in the corresponding mode (when available): `train_step_timing`,   `train_backward_timing`, `validation_step_timing`,   `test_step_timing`.

## Weights & Biases Output Example

image


nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        Starts measuring a named timer.
+        """
+        timer_data = self.timers.get(name, {})
+
+        if "start" in timer_data:
+            raise RuntimeError(f"Cannot start timer = '{name}' since it is already active")
+
+        # synchronize pytorch cuda execution if supported
+        if torch.cuda.is_initialized():
```
titu1994, line 64: Why is this required ? This will dramatically slow down training of all models. Revert as it's not particularly needed - while individual kernels are indeed asynchronous, a pytorch forward call will generally deal with this as needed. Also, models total forward cost will generally be significant enough to require repeated syncs for such accurate timing info.

nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+    def stop(self, name=""):
+        """
+        Stops measuring a named timer.
+        """
+        timer_data = self.timers.get(name, None)
+        if (timer_data is None) or ("start" not in timer_data):
+            raise RuntimeError(f"Cannot end timer = '{name}' since it is not active")
+
+        # synchronize pytorch cuda execution if supported
+        if torch.cuda.is_initialized():
```
titu1994, line 80: Same comment as above, please revert to hurt training time.

nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # compute dt and make timer inactive
+        dt = time.time() - timer_data.pop("start")
+
+        # store dt
+        timer_data["dt"] = timer_data.get("dt", []) + [dt]
+
+        self.timers[name] = timer_data
+
+    def active(self):
```
titu1994, line 91: Function name doesn't indicate returned value - consider renaming to something more indicative of the fact if returns all active timer names

nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+    def export(self, mean=None):
+        """
+        Exports a dictionary with average/all dt per named timer
+
+        mean - if True return the mean per name of all measures,
+               else return a list per name.
+        """
+        if mean is None:
+            mean = self.mean
+
```
titu1994, line 122: Lots of overlap with the reduction code below - consider pulling it out into it's own method and simply call it in various places.

titu1994 (changes requested):
Some comments. Overall looks great, however there is significant concern about training slowdown due to per batch call to torch.cuda.synchronize()

nemo/utils/timers.py
michalivne: I was following the timer from megatron code, which was used as an example for the timing functionality. If you think that it is not required in order to correctly measure the step runtime I will remove it. @erhoo82 do you have any thoughts on the matter?

nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    A timer class that supports multiple named timers.
+    A named timer can be used multiple times, in which case the average
+    dt will be returned.
+    A named timer cannot be started if it is already currently running.
+    Use case: measuring execution of multiple code blocks.
+    """
+
+    def __init__(self, mean=True):
+        """
```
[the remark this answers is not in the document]
michalivne: Done.

nemo/utils/timers.py
michalivne: Done. Renamed to `active_timers`

nemo/utils/timers.py
michalivne: Done.

nemo/utils/timers.py
```diff
@@ -0,0 +1,130 @@
...
+        mean - default behaviour. If True mean dt is returned, else a list.
```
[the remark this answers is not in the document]
michalivne: Done.

nemo/utils/timers.py
michalivne: Will be resolved as above.

nemo/utils/timers.py
michalivne: Following a discussion with @erhoo82, it seems that the cuda synchronization might be required by the megatron team. I propose to add a flag to enable cuda synchronization when needed, and to skip it by default. Will that solution address your performance concerns?

michalivne: Added `exp_manager.step_timing_sync_cuda=False # default is False`  config, which disables cuda sync by default.

nemo/utils/exp_manager.py
```diff
@@ -101,6 +101,48 @@ class ExpManagerConfig:
...
+
+class TimingCallback(Callback):
+    """
+    Logs execution time of train/val/test steps
+    """
+
+    def __init__(self, timer=None):
+        # support external timer
+        if timer is None:
+            timer = timers.NamedTimer(sync_cuda=step_timing_sync_cuda)
```
titu1994: No argument called step_timimg_sync_cuda passed to this init

nemo/utils/exp_manager.py
```diff
@@ -238,6 +280,11 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
             cfg.wandb_logger_kwargs,
         )
 
+    # add loggers timing callbacks
+    if cfg.log_step_timing:
+        timing_callback = TimingCallback()
```
titu1994: No argument sync_cuda passed to the NamedTimer

nemo/utils/exp_manager.py
```diff
@@ -101,6 +101,48 @@ class ExpManagerConfig:
     checkpoint_callback_params: Optional[CallbackParams] = CallbackParams()
     # Additional exp_manager arguments
     files_to_copy: Optional[List[str]] = None
+    # logs timing of train/val/test steps
+    log_step_timing: Optional[bool] = True
+    # if True torch.cuda.synchronize() is called on start/stop
```
titu1994, line 106: This should possibly be a nested dict with following defaults (represented as a Data class, modifiable via Hydra) -

step_timing_kwargs:
  reduction: "mean" 
  sync_cuda: False

Then the instantiation of the timer insider expmanager can be TimerCallback(**cfg.step_timing_kwargs)

This further allows more aarguments + keep exp manager cleaner

titu1994 (changes requested):
This is better, minor refactoring needed

nemo/utils/exp_manager.py
michalivne: Fixed.

nemo/utils/exp_manager.py
michalivne: Fixed.

nemo/utils/exp_manager.py
michalivne: Great suggestion! Done.

nemo/utils/exp_manager.py
```diff
@@ -238,6 +289,11 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
             cfg.wandb_logger_kwargs,
         )
 
+    # add loggers timing callbacks
+    if cfg.log_step_timing:
+        timing_callback = TimingCallback(timer_kwargs=cfg.step_timing_kwargs)
```
titu1994, line 294: Since kwargs are optional in the data class, check for none here and replace with {}

titu1994 (commented):
Looks good ready to merge. One last comment about optional kwargs

nemo/utils/exp_manager.py
michalivne: Good point. Replaced with `cfg.step_timing_kwargs or {}`.

titu1994 (approved):
LGTM, merge when tests pass

PR #2947, opened 2021-10-04
Itn fr

tbartley94:
Added README for French ITN to explain possible ambiguity cases. Fixed currency data so there is no longer overlap between cardinals an minor denominations.


yzhang123 (approved):
Thanks!

PR #2951, opened 2021-10-05
NVFuser fixes

MaximumEntropy:


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
             seed=self.cfg.get('seed', 1234),
         )
 
+        if not self.cfg.get('fused_bf16'):
+            set_jit_fusion_options()
+
         if isinstance(self.trainer.accelerator_connector.precision_plugin, NLPNativeBfloat16PrecisionPlugin):
```
ericharper, line 66: We can remove this?

nemo/collections/nlp/modules/common/megatron/utils.py
```diff
@@ -164,11 +164,3 @@ def autocast_forward(self, *args):
...
             else:
                 return self.func(*args)
-
-
-@contextmanager
-def dummy_handler():
-    try:
-        yield
-    finally:
-        pass
```
ericharper, line 174: Don't we use this?

MaximumEntropy: No longer, I had missed removing it in some other places and just fixed it.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
MaximumEntropy: Remove it entirely? I think we want to disable jit fusion for bf16 right?

ericharper (approved):
LGTM. Thanks!

PR #2954, opened 2021-10-06
fix/machine translation/minor bug in passing variables

PeganovAnton:
Signed-off-by: PeganovAnton <peganoff2@mail.ru>


nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -481,10 +481,10 @@ def setup_training_data(self, train_data_config: Optional[DictConfig]):
         self._train_dl = self._setup_dataloader_from_config(cfg=train_data_config)
 
     def setup_multiple_validation_data(self, val_data_config: Union[DictConfig, Dict]):
-        self.setup_validation_data(self._cfg.get('validation_ds'))
+        self.setup_validation_data(val_data_config)
```
michalivne, line 484: Do we need setup_multiple_validation_data at all? It seems redundant.
@titu1994 what do you think?

michalivne (commented):
See comment in code. Also, could you please add a one-line description of what was done?(perhaps wait until we finalize changes though)

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
PeganovAnton: @michalivne Method `MTEncDecModel.setup_multiple_validation_data` overrides `ModelPT.setup_multiple_validation_data`, so if we remove  `MTEncDecModel.setup_multiple_validation_data` we will need to rewrite all dataloader setup.

michalivne (approved):
LGTM!

PR #2976, opened 2021-10-08
Always save last checkpoint on train end even if folder does not exist

blisc:


titu1994:
If changing the signature for save_to, change it in all the places - the SaveRestoreConnector, the super class of ModelPT which defines this abstract method etc. 

Also, why is there a book flag to do this anyway, it seems reasonable to do it by default without changing the signature

PR #3017, opened 2021-10-17
Adding parallel transcribe for ASR models -  suppports multi-gpu/multi-node

VahidooX:
+Added transcribe_speech_parallel.py which supports multi-gpu/multi-node inference for ASR models
+Supports both tarred and non-tarred datasets
+works for both CTC and Transducer models
+Added extra sample_id output to the samples returned by the ASR datasets. It enables us  to track down the samples. Would be useful for other tasks in future.
+Added the config description for ASR datasets/datalaoders

Needs to be done:
+Adding support for changing the decoder
+Adding fused_batch support?
+Retrieving some configs from the model itself instead of trusting the use to set them like use_start_end_token

Example for non-tarred datasets:

```
python transcribe_speech_parallel.py \
model=stt_en_conformer_ctc_small \
predict_ds.manifest_filepath=/dataset/manifest_file.json \
predict_ds.batch_size=16 \
output_path=/tmp/
```

Example for tarred datasets:

```
python transcribe_speech_parallel.py \
model=stt_en_conformer_ctc_small \
predict_ds.is_tarred=true \
predict_ds.manifest_filepath=/tarred_dataset/tarred_audio_manifest.json \
predict_ds.tarred_audio_filepaths=/tarred_dataset/audio__OP_0..127_CL_.tar \
predict_ds.batch_size=16 \
output_path=/tmp/
```

By default the trainer uses all GPUs available and precision is fp32. By setting the trainer config user may control these configs. For example to do the predictions with AMP on just two GPUs:

```
python transcribe_speech_parallel.py \
trainer.precision=16 \
trainer.gpus=2 \
model=stt_en_conformer_ctc_small \
predict_ds.manifest_filepath=/dataset/manifest_file.json \
predict_ds.batch_size=16 \
output_path=/tmp/
```

Or control the dataloader's config by setting the predict_ds:

```
python transcribe_speech_parallel.py \
predict_ds.num_workers=8 \
predict_ds.min_duration=2 \
predict_ds.sample_rate=16000 \
model=stt_en_conformer_ctc_small \
predict_ds.manifest_filepath=/dataset/manifest_file.json \
predict_ds.batch_size=16 \
output_path=/tmp/
```


nemo/collections/asr/data/audio_to_text.py
```diff
@@ -116,7 +130,7 @@ def __init__(
         self.parser = parser
 
         self.collection = collections.ASRAudioText(
-            manifests_files=manifest_filepath.split(','),
+            manifests_files=manifest_filepath,
```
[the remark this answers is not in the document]
VahidooX: It is addressed in another PR for fixing the bucketing feature.

examples/asr/transcribe_speech_parallel.py
```diff
@@ -0,0 +1,174 @@
...
+        "unk_index",
+        "normalize",
+        "parser",
+        "eos_id",
+        "bos_id",
+        "pad_id",
+    ]
+    for cfg_name in cfg_name_list:
+        if hasattr(train_ds, cfg_name):
+            setattr(predict_ds, cfg_name, getattr(train_ds, cfg_name))
```
titu1994, line 113: Cast predict_ds to OmegaConf in here otherwise it wont work if predict_ds remains a dataclass

examples/asr/transcribe_speech_parallel.py
```diff
@@ -0,0 +1,174 @@
...
+from nemo.utils import logging
+from nemo.utils.get_rank import is_global_rank_zero
+
+
+@dataclass
+class ParallelTranscriptionConfig:
+    model: Optional[str] = None  # name
+    predict_ds: ASRDatasetConfig = ASRDatasetConfig(return_sample_id=True, num_workers=4)
+    output_path: Optional[str] = None
+    return_predictions: bool = False
```
titu1994, line 85: Maybe a docstring on what this flag will do

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -78,8 +89,11 @@ def _speech_collate_fn(batch, pad_id):
         audio_signal, audio_lengths = None, None
     tokens = torch.stack(tokens)
     tokens_lengths = torch.stack(tokens_lengths)
-
-    return audio_signal, audio_lengths, tokens, tokens_lengths
+    if sample_ids is None:
+        return audio_signal, audio_lengths, tokens, tokens_lengths
+    else:
+        sample_ids = torch.IntTensor(sample_ids)
```
titu1994, line 95: Use torch.tensor() as recommened in torch instead of IntTensor

nemo/collections/asr/data/audio_to_text.py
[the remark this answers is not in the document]
titu1994: Ah ok.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -204,6 +221,10 @@ def __init__(
         )
         self.featurizer = WaveformFeaturizer(sample_rate=sample_rate, int_values=int_values, augmentor=augmentor)
         self.trim = trim
+        self.return_sample_id = return_sample_id
+
+    def get_sample(self, sample_id):
```
titu1994, line 226: Rename to get_manifest_sample - first read it seems like get audio sample which  is wrong here.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1118,7 +1153,13 @@ def _build_sample(self, tup):
             t = t + [self.eos_id]
             tl += 1
 
-        return f, fl, torch.tensor(t).long(), torch.tensor(tl).long()
+        if self.return_sample_id:
+            return f, fl, torch.tensor(t).long(), torch.tensor(tl).long(), manifest_idx
+        else:
+            return f, fl, torch.tensor(t).long(), torch.tensor(tl).long()
+
+    def get_sample(self, sample_id):
```
titu1994, line 1161: rename

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -299,6 +307,47 @@ def get_dali_bpe_dataset(
     return dataset
 
 
+class ASRPredictionWriter(BasePredictionWriter):
+    def __init__(self, dataset, output_file: str, use_cer=False):
```
titu1994, line 311: If we are writing WER, might as well write CER too. Too expensive to do it separately. Also, will this write global WER / CER? Cause sample level metric will behave like average of averages and give wrong scores.

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -299,6 +307,47 @@ def get_dali_bpe_dataset(
...
+        dataloader_idx: int,
+    ):
+        for sample_id, transcribed_text in prediction:
+            item = {}
+            sample = self.dataset.get_sample(sample_id)
+            item["audio_filepath"] = sample.audio_file
+            item["duration"] = sample.duration
+            item["text"] = sample.text_raw
+            item["pred_text"] = transcribed_text
+            wer_cer = word_error_rate(
```
titu1994, line 335: This is both correct in one sense and wrong in another- this will cause "average of averages" behaviour when calculating metrics if you take the average of all the sample WERs across number of samples in the dataset. 

It is fine to calculate just the single sample level WER/CER though. 
If you need to use it to calculate the "global dataset WER/CER" then you need to create the appropriate CTC/RNNT WER/WERBPE metric object, call it on the predictions of the model, and globally preserve the scores and words tensor (sum across batches) and then do a final division at the end. Make a note of this or remove these two metrics entirely from the logs. It causes more confusion than solving the problem and given a text manifest output it is easy to calculate sample or global metrics anyway.

nemo/collections/asr/models/ctc_models.py
```diff
@@ -518,7 +519,7 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
 
     @typecheck()
     def forward(
-        self, input_signal=None, input_signal_length=None, processed_signal=None, processed_signal_length=None
+        self, input_signal=None, input_signal_length=None, processed_signal=None, processed_signal_length=None,
```
titu1994, line 522: , can be reverted

nemo/collections/asr/models/ctc_models.py
```diff
@@ -596,6 +597,22 @@ def training_step(self, batch, batch_nb):
 
         return {'loss': loss_value, 'log': tensorboard_logs}
 
+    def predict_step(self, batch, batch_idx, dataloader_idx=0):
+        signal, signal_len, transcript, transcript_len, sample_id = batch
+        if isinstance(batch, DALIOutputs) and batch.has_processed_signal:
```
titu1994, line 602: I see we dont support dali for returning index. Could you take a look to add it? Or note that we wont get DALIOutputs here.

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -580,7 +580,7 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
 
     @typecheck()
     def forward(
-        self, input_signal=None, input_signal_length=None, processed_signal=None, processed_signal_length=None
+        self, input_signal=None, input_signal_length=None, processed_signal=None, processed_signal_length=None,
```
titu1994, line 583: Comma can be reverted

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -699,6 +698,23 @@ def training_step(self, batch, batch_nb):
 
         return {'loss': loss_value}
 
+    def predict_step(self, batch, batch_idx, dataloader_idx=0):
+        signal, signal_len, transcript, transcript_len, sample_id = batch
+
+        # forward() only performs encoder forward
+        if isinstance(batch, DALIOutputs) and batch.has_processed_signal:
```
titu1994, line 705: Same as above re: DALIOutputs

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -699,6 +698,23 @@ def training_step(self, batch, batch_nb):
...
+            encoded, encoded_len = self.forward(processed_signal=signal, processed_signal_length=signal_len)
+        else:
+            encoded, encoded_len = self.forward(input_signal=signal, input_signal_length=signal_len)
+        del signal
+
+        best_hyp_text, all_hyp_text = self.decoding.rnnt_decoder_predictions_tensor(
```
titu1994, line 711: For RNNTs, you can have beam decoding (i think its too slow but still an option). So better to return hypotheses = True and return the hypotheses, all_hypotheses as output.

tests/collections/asr/test_asr_datasets.py
```diff
@@ -341,7 +341,6 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
 
         num_samples = 10
         batch_size = 1
-        device = 'gpu' if torch.cuda.is_available() else 'cpu'
```
titu1994, line 344: Why this is reverted?

tests/collections/asr/test_asr_datasets.py
```diff
@@ -394,7 +393,6 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
             )
             ref_preprocessor = EncDecCTCModel.from_config_dict(preprocessor_cfg)
 
-            count = 0
```
titu1994, line 397: Maybe a mistake?

titu1994 (changes requested):
Overall a fantastic improvement, some comments here and there.

examples/asr/transcribe_speech_parallel.py
VahidooX: Added.

nemo/collections/asr/data/audio_to_text.py
VahidooX: done.

nemo/collections/asr/data/audio_to_text.py
VahidooX: done.

nemo/collections/asr/data/audio_to_text.py
VahidooX: done.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: Currently we do not log any wer, this one just calculates the wer per sample and it would be written into the output file. We can add a global wer calculation but let's add it in another PR.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: This code does not support global wer/cer calculation. These are per sample and will be written per sample in the output file. We can add that support in another PR.

nemo/collections/asr/models/ctc_models.py
VahidooX: done.

nemo/collections/asr/models/rnnt_models.py
VahidooX: done.

tests/collections/asr/test_asr_datasets.py
VahidooX: That device was not used in the code at all.

tests/collections/asr/test_asr_datasets.py
VahidooX: This count was not used in the code.

nemo/collections/asr/models/rnnt_models.py
VahidooX: Added error message in the dali layers that they do not support return_sample_id. Here is a general predict which works with or without return_sample_id and may be used in other cases without returning the index.

nemo/collections/asr/data/audio_to_text_dataset.py
titu1994: WER per sample is "wrong" though. People will try to calculate WER global by averaging WER per sample and get wrong results. Please revert or compute global in this PR only.

nemo/collections/asr/models/ctc_models.py
titu1994: Ok sounds good, lets leave Dali for later.

nemo/collections/asr/models/rnnt_models.py
titu1994: Cool thanks.

tests/collections/asr/test_asr_datasets.py
titu1994: Lol ok then

tests/collections/asr/test_asr_datasets.py
titu1994: Cool thanks for the cleanup

titu1994 (commented):
Refer to comment about global vs sample WER. Either remove WER calc entirely or only measure global WER/CER. Sample WER invites incorrect calculation of global WER

titu1994 (dismissed):
Awesome, ready to merge !

titu1994 (approved):
Perfect, looks great !

PR #3046, opened 2021-10-22
Fix the tarred and bucketing support for speech label model

VahidooX:
+ Adding tarred dataset support to the speech label model.
+ Adding the support of bucketing tarred datasets


examples/asr/conf/matchboxnet/matchboxnet_3x1x64_v1.yaml
```diff
@@ -168,7 +168,7 @@ trainer:
   max_epochs: 200
   max_steps: null # computed at runtime if not set
   num_nodes: 1
-  accelerator: ddp
+  accelerator: gpu
```
nithinraok, line 171: should we revert this back to ddp?

nithinraok (approved):
Looks good. Thank you Vahid

examples/asr/conf/matchboxnet/matchboxnet_3x1x64_v1.yaml
VahidooX: reverted it back.

nithinraok (approved):

PR #3059, opened 2021-10-26
Ptl upgrade r1.5.0 fixes

nithinraok:
This PR updates deprecated variables in PTL
* distributed_backened is removed and replaced with strategy inside accelerator connector
* accelerator now only accepts the type of device input


okuchaiev:
@nithinraok should this be closed since 1.5.0 is already out?

PR #3072, opened 2021-10-28
CTC Conformer fixes for ONNX/TS export 

borisfom:


borisfom:
With those changes, we are actually able to convert Conformer to TRT!
https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3410215

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -119,7 +119,7 @@ def __init__(
         att_context_size=None,
         xscaling=True,
         untie_biases=True,
-        pos_emb_max_len=5000,
+        pos_emb_max_len: int = 5000,
```
VahidooX, line 122: Would you please add types for other inputs to be consistent?

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,38 @@ def __init__(
         else:
             self.out_proj = None
             self._feat_out = d_model
+        self.xmax = pos_emb_max_len
```
VahidooX, line 208: let's use pos_emb_max_len for the name of the variable here instead of xmax.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,38 @@ def __init__(
...
+        device = next(self.parameters()).device
+        seq_range = torch.arange(0, self.xmax, device=device)
+        self.register_buffer('seq_range', seq_range, persistent=False)
 
     @typecheck()
     def forward(self, audio_signal, length=None):
+        audio_size: int = audio_signal.size(-1)
+        if audio_size > self.xmax:
+            logging.warning(
+                f"Audio Signal is too long : {audio_size}, please increase pos_emb_max_len (={pos_emb_max_len}) config parameter!"
```
VahidooX, line 218: They can not change that parameter from the config easily. We need to add a method to update this parameter for them.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,38 @@ def __init__(
         else:
             self.out_proj = None
             self._feat_out = d_model
+        self.xmax = pos_emb_max_len
+        device = next(self.parameters()).device
+        seq_range = torch.arange(0, self.xmax, device=device)
```
VahidooX, line 210: Settingthe device in the init means that user may not move the model after loading the model.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,38 @@ def __init__(
...
-
+        # adjust size
+        audio_size = audio_signal.size(1)
         # Create the self-attention and padding masks
-        pad_mask = self.make_pad_mask(length, max_time=xmax, device=audio_signal.device)
-        att_mask = pad_mask.unsqueeze(1).repeat([1, xmax, 1])
-        att_mask = att_mask & att_mask.transpose(1, 2)
+        pad_mask = self.make_pad_mask(audio_size, length)
+        att_mask = pad_mask.unsqueeze(1).repeat([1, audio_size, 1])
+        print(pad_mask.dtype)
```
VahidooX, line 238: print needs to get dropped.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -285,7 +273,13 @@ class RelPositionalEncoding(PositionalEncoding):
     """
 
     def __init__(self, d_model, dropout_rate, max_len=5000, xscale=None, dropout_rate_emb=0.0):
-        super().__init__(d_model, dropout_rate, max_len, xscale=xscale)
+        super().__init__(
+            d_model,
+            dropout_rate,
+            max_len,
+            xscale=xscale,
+            positions=torch.arange(max_len - 1, -max_len, -1, dtype=torch.float32),
```
VahidooX, line 281: positions does not exist in the class this one inherited from.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -315,19 +296,15 @@ def forward(self, x):
...
         if self.xscale:
             x = x * self.xscale
 
         # center_pos would be the index of position 0
         # negative positions would be used for right and positive for left tokens
         # for input of length L, 2*L-1 positions are needed, positions from (L-1) to -(L-1)
-        center_pos = self.pe.size(1) // 2
-        start_pos = center_pos - x.size(1) + 1
-        end_pos = center_pos + x.size(1)
+        start_pos = self.center_pos - x.size(1)
```
VahidooX, line 306: Why center_pos is not calculated directly here?

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -91,48 +93,28 @@ def __init__(self, subsampling, subsampling_factor, feat_in, feat_out, conv_chan
...
-                padding=self._padding,
-                kernel_size=self._kernel_size,
-                stride=self._stride,
-                ceil_mode=self._ceil_mode,
-            )
-            in_length = out_length
-
-        out_length = int(out_length)
-        self.out = torch.nn.Linear(conv_channels * out_length, feat_out)
+        self.add_pad = torch.tensor((2 * self._padding) - self._kernel_size, dtype=torch.float)
```
VahidooX, line 96: These parameters are the same in the normal example but may be different in the same model for different modules. Can we keep the method of calc_length with those inputs?

VahidooX (changes requested):

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Agreed. Look into how this supported for ConvASREncoder inside https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/mixins/mixins.py#L153
And
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/utils/asr_module_utils.py

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: Why, it is there in PositionalEncoding

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: self.center_pos is precalculated in init()

nemo/collections/asr/parts/submodules/subsampling.py
borisfom: Done.

nemo/core/classes/common.py
```diff
@@ -486,7 +486,7 @@ def from_config_dict(cls, config: 'DictConfig', trainer: Optional['Trainer'] = N
             # target class resolution was unsuccessful, fall back to current `cls`
             if instance is None:
                 if imported_cls_tb is not None:
-                    logging.debug(
+                    logging.info(
```
titu1994, line 489: This should remain debug, it prints a lot of information otherwise which just confused users, even though model is properly instantiated.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -240,16 +258,7 @@ def forward(self, audio_signal, length=None):
         audio_signal = torch.transpose(audio_signal, 1, 2)
         return audio_signal, length
 
-    @staticmethod
-    def make_pad_mask(seq_lens, max_time, device=None):
+    def make_pad_mask(self, audio_size, seq_lens):
```
titu1994, line 261: Refer to response in thread. Why is this being made static ? For export ? This is problematic because I'm using this in a subsequent PR for masked SE block in Citrinets and it can't be made static there.

titu1994 (commented):
Overall changes look good. Revert of logging from info to debug for instantiation cause it's unnecessary for many cases. 

A major concern is why is masking not dynamic anymore - is export harmed by it? I plan on using this type of dynamic mask creation for Citrinet SE block soon as well.

nemo/collections/asr/modules/conformer_encoder.py
borisfom: Why, it's the other way around - I removed static.

nemo/core/classes/common.py
borisfom: That was for debug - I will revert that.

borisfom:
@titu1994 : yes dynamic mask creation in forward() brings in a set of operations that are not supported by TRT. So I changed it to pre-computing the mask and slicing it at runtime.

nemo/collections/asr/modules/conformer_encoder.py
borisfom: I have added set method in this PR.

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Now the mask size is limited at init time - previously it was dependent on input sequence length and didn't depend on some notion of max seq length at train time that had to be set. 

For example it is very common to take pretrained Conformer trained on 16s max length and infer on 2 mins of audio in offline. Same for Citrinet, train on 16 s eval on 2 hours audio. Unlike positional embeddings, these masks don't need to be learned either so it makes more sense to construct them at runtime. 

Basically, Init should not be used to define max length,but should adjust to current provided max length.

titu1994:
In this case, that is highly problematic for Nemo use cases. What we can do is (1) keep dynamic shape construction as before for normal case (2) switch to static explicit mask slice inside of prepare for export method by switching the called function pointer to this one (which just slices). In such case the max seq length can be set at export time as well.

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -91,48 +93,40 @@ def __init__(self, subsampling, subsampling_factor, feat_in, feat_out, conv_chan
...
-
-
-def calc_length(lengths, padding, kernel_size, stride, ceil_mode):
-    """ Calculates the output length of a Tensor passed through a convolution or max pooling layer"""
-    if ceil_mode:
-        lengths = torch.ceil((lengths + (2 * padding) - (kernel_size - 1) - 1) / float(stride) + 1)
-    else:
-        lengths = torch.floor(torch.div(lengths + (2 * padding) - (kernel_size - 1) - 1, stride) + 1)
-    return lengths
+            lengths = torch.div(lengths + add_pad, stride) + self.one
```
VahidooX, line 127: We need the calc_length not to be dependent on the self._sampling_num as it is going to be used as a general length calculator in our next releases. In current code, it may work but it would broke some of other experimental codes. Would it be a problem to move the loop outside of calc_length?

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -91,48 +93,40 @@ def __init__(self, subsampling, subsampling_factor, feat_in, feat_out, conv_chan
...
-                padding=self._padding,
-                kernel_size=self._kernel_size,
-                stride=self._stride,
-                ceil_mode=self._ceil_mode,
-            )
-            in_length = out_length
-
-        out_length = int(out_length)
-        self.out = torch.nn.Linear(conv_channels * out_length, feat_out)
+        self.one = torch.tensor(1, dtype=torch.float)
```
VahidooX, line 96: Do we need this self.one? Why not use 1.0 or torch.tensor(1, dtype=torch.float) in the place we need explicitly?

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,43 @@ def __init__(
...
         if isinstance(self.pre_encode, ConvSubsampling):
             audio_signal, length = self.pre_encode(audio_signal, length)
         else:
             audio_signal = self.pre_encode(audio_signal)
-
         audio_signal, pos_emb = self.pos_enc(audio_signal)
-        bs, xmax, idim = audio_signal.size()
-
+        # adjust size
+        audio_size = audio_signal.size(1)
```
VahidooX, line 240: Can we rename audio_size to max_audio_length?

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: It looks to me that this function is not static in sense of the maximum audio lengths. self.seq_range is fixed but it is dependent on the max_len=5000 which is a large number already. I would add a method to give the user the ability to extend it after training is done.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -219,19 +216,24 @@ class PositionalEncoding(torch.nn.Module):
...
         """Construct an PositionalEncoding object."""
         super(PositionalEncoding, self).__init__()
         self.d_model = d_model
         self.xscale = xscale
         self.dropout = torch.nn.Dropout(p=dropout_rate)
-        self.extend_pe(length=max_len)
+        if positions is None:
+            positions = torch.arange(0, max_len, dtype=torch.float32)
+        self.create_pe(positions=positions)
+        self.center_pos = torch.tensor(self.pe.size(1) // 2 + 1, dtype=torch.int32)
```
VahidooX, line 228: self.center_pos is not used in PositionalEncoding. It is a concept which is needed for RelPositionalEncoding. We may need to move it to the init of RelPositionalEncoding or calculate it on the fly.

nemo/collections/asr/parts/submodules/multi_head_attention.py
VahidooX: Oh I see you added that to PositionalEncoding too.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -315,19 +296,15 @@ def forward(self, x):
...
             x = x * self.xscale
 
         # center_pos would be the index of position 0
         # negative positions would be used for right and positive for left tokens
         # for input of length L, 2*L-1 positions are needed, positions from (L-1) to -(L-1)
-        center_pos = self.pe.size(1) // 2
-        start_pos = center_pos - x.size(1) + 1
-        end_pos = center_pos + x.size(1)
+        start_pos = self.center_pos - x.size(1)
+        end_pos = self.center_pos + x.size(1)
```
VahidooX, line 307: You changed the center from "self.pe.size(1) // 2" to "self.pe.size(1) // 2 + 1" but end_pos is still the same as before. How come the code is still working?

VahidooX (changes requested):

nemo/collections/asr/parts/submodules/subsampling.py
borisfom: It won't be a problem except for code duplication. Why moving the loop out though - we can just make it a parameter.

nemo/collections/asr/parts/submodules/subsampling.py
borisfom: That was result of some tweaking for ONNX and TS export - let me double check how it renders with current code.

nemo/collections/asr/modules/conformer_encoder.py
borisfom: Sure.

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: That beats me ...

nemo/collections/asr/modules/conformer_encoder.py
borisfom: Yes, I have made that functionality available in set_pos_emb_max_len. So it's called in init() for config value and can also be called explicitly to increase max len for inference. For regular pytorch inference, I can also provide separate forward() and forward_for_export() so that forward() still has dynamic length - would you like that, too ?
Mask can't be fully dynamic with TRT, unfortunately - that does raise a requirement for possible 'config adjustment' during export() - for RIVA, we can handle that with direct config modification though.

titu1994: Right for pytorch only inference and training let's make it dynamic and for export force fixed mask

borisfom: Done.

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: Still working after I fixed end_pos to be like before.

titu1994 (commented):
Looks good now, so it caches and updates max length based on pos_max_emb_length which is close to 5000 timesteps anyway. A bit of waste of memory considering the layer will never see such long sentences during training but ok for now. 

I'll use similar technique for SE block as well. Can you add a small Conformer (1-2 block) export unit test similar to what we have for Citrinet ? Just to make sure everything stays well over time. 

Also, @VahidooX take this branch and perform inference on say a ASR set model to see if numbers remain good

nemo/collections/asr/parts/submodules/subsampling.py
VahidooX: I know that it is a code duplicate but the issue is that I am now working on some code changes and in those changes the parameters of each covn layer can be different from the other one like the padding. For those cases, I need to be able to pass slightly different parameters to the conv layers. For instance for downsampling of 6, we need one with striding of 3 and one with 2. Having the loop inside reduces our flexibility for future extensions.

nemo/collections/asr/parts/submodules/multi_head_attention.py
VahidooX: The code looks fine generally except the loop of the calc_length. Let me also test the code on a real evaluation set tomorrow to make sure it is working as before.

VahidooX: Thanks Boris!

borisfom:
@titu1994 : can we have this test added in a separate PR (and better by someone more familiar with the model)? 
The pattern we use in test_asr_exportables.py is to create a fake config, not to pull pretrained model.

titu1994:
Oh yeah I meant create a fake Conformer model for the exportable test. Not a pretrained one. Should be close to copy paste of Citrinet code replacing encoder with conformer encoder. @VahidooX you'll need to add that after this PR gets merged then for safety

VahidooX (commented):
LGTM, thanks Boris! Just left some minor comments. I would test it on an evaluation set before merging the PR.

borisfom:
@VahidooX : one more commit to propagate dynamic creation down to pos_enc objects.
Please take a close look : it seems to me that in original code, base PositionalEncoding.extend_pe() was called during RelPositionalEncoding.init(). If large max_len was configured, that probably have led to RelPositionalEncoding always using the mask initialized by PositionalEncoding. I have fixed that - please check if that actually works correctly. I feel like it should - at least it's working the same way my static initialization was working.

titu1994 (commented):
Looks correct to me. @VahidooX for final checks on pretrained model inference scores are same or not, and then let's merge

VahidooX (approved):
Tested the code and it is fine. Just a minor comment about the seq_range.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,49 @@ def __init__(
...
+    def set_pos_emb_max_len(self, pos_emb_max_len):
+        """ Sets maximum input length.
+            Pre-calculates internal seq_range mask.
+        """
+        self.pos_emb_max_len = pos_emb_max_len
+        device = next(self.parameters()).device
+        seq_range = torch.arange(0, self.pos_emb_max_len, device=device)
+        if not hasattr(self, 'seq_range'):
+            self.register_buffer('seq_range', seq_range, persistent=False)
+        self.seq_range = seq_range
```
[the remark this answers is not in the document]
VahidooX: Please also change the name of the input argument of calc_length from sampling_num to repeat_num. Then looks like we good to go and merge it!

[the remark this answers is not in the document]
borisfom: Well, register_buffer() creates volatile data member with a copy of seq_range;  = reassigns it. Yes I could have put it under else, but that does not matter really.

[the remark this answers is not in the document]
borisfom: Added 'else'.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -205,25 +205,50 @@ def __init__(
...
+        seq_range = torch.arange(0, self.pos_emb_max_len, device=device)
+        if hasattr(self, 'seq_range'):
+            self.seq_range = seq_range
+        else:
+            self.register_buffer('seq_range', seq_range, persistent=False)
+        self.pos_enc.extend_pe(pos_emb_max_len)
 
     @typecheck()
     def forward(self, audio_signal, length=None):
+        max_audio_length: int = audio_signal.size(-1)
```
titu1994, line 225: Before this is merged, shouldnt this be audio_signal.max() instead of audio_signal.size(-1) - this is a tensor of shape [B] so size(-1) just gives 32 always? 
Found this while testing the equivalent code for Citrinets. @borisfom @VahidooX

borisfom: 1. This is what we used before my changes. 2. This is just to be used in mask creation etc - I don't think it changes any logic there, maybe just makes them bigger than they have to be.  3. Changing it to length.max() would introduce data-dependent shapes (as opposed to shape-dependant shape) and that would almost certainly break TRT.

titu1994: This is in forward() not forward_for_export() so it won't break shapes with TRT. If it was used before I support no need to change it.

VahidooX (dismissed):

VahidooX (dismissed):

nemo/core/classes/common.py
VahidooX: It is still in debug mode.

VahidooX (approved):
LGMT, thanks!

PR #3083, opened 2021-10-28
pytorch-lightning version requirement fix

darkydash:
The latest version of pytorch-lightning is 1.4.9 though currently the version higher that is required. The commit fixes this by allowing the 1.4.9 version to use


okuchaiev:
closing this because 1.5.0 is out. We were using 1.5.0rc on main which is why it was higher than 1.4.9

PR #3089, opened 2021-10-29
Fix the asr bucketing parsing the manifest.

VahidooX:


nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -155,7 +155,9 @@ def get_tarred_dataset(
     manifest_filepaths = convert_to_config_list(manifest_filepaths)
 
     if len(manifest_filepaths) != len(tarred_audio_filepaths):
-        raise ValueError(f"manifest_filepaths and tarred_audio_filepaths need to have the same number of buckets.")
+        raise ValueError(
```
titu1994, line 158: What if you dont use buckets. This wont work then. You have to account for 4x2 scenarios - 
{loose file dataset, concat loose file dataset, tarred dataset, tarred bucketed dataset} x {char, subword}

titu1994 (changes requested):
Could you elaborate how this fixes scenario when tarred (non-bucketed) dataset is used.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: Tested all those combinations. That convert_to_config_list standardizes the input user gives us. When we reach to this line,  manifest_filepaths and tarred_audio_filepaths would have the same format, a list of lists.
When non-bucketed formats are used (like just passing a list or a string with comma separated for concatenation), this convert_to_config_list makes those into a list of one single list and then we just pass that inside list in line 166 of audio_to_text_dataset.py.

titu1994: Hmm it would be nice to have a CI test such things but lets do it later. This will break in the long term and will break silently when we try to add concat support

titu1994 (approved):

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: Yeah, agree we need to add more tests generally for tarred datasets now that it is used more frequently.

PR #3092, opened 2021-10-29
TalkNet Fix

stasbel:


nemo/core/connectors/save_restore_connector.py
```diff
@@ -122,9 +122,11 @@ def restore_from(
                         # Resolve the override config
                         conf = OmegaConf.to_container(conf, resolve=True)
                         conf = OmegaConf.create(conf)
+
                 # If override is top level config, extract just `model` from it
-                if 'model' in conf:
```
stasbel, line 126: @ericharper 
Hi Eric,
Can you please comment on necessity of this two lines.

The problem for me right now is following:
When we create ModelPT class, we pass only model part of it. So, when doing automatic saving while checkpointing, only this part is saved within .nemo file.
So, if your ModelPT has inner object with the same name 'model' (as well as config part), current code doens't distinct it from outer config and make `conf = conf.model` assignment.
If I comment this two lines, some tests become invalid, i.e. tests/core/test_save_restore.py::TestSaveRestore::test_mock_restore_from_config_override_with_yaml

stasbel: Basically, current NeMo inner implementation forbids to create nn.Modules with plain 'model' naming, which is quite common to use in practice.

ericharper: We use `self.model` here: https://github.com/NVIDIA/NeMo/blob/d5bd93675b82045d1ca7541e45981fe0440631a6/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L86

ericharper: @titu1994

ericharper: So this problem is coming from the config having `model.model` ? https://github.com/NVIDIA/NeMo/blob/d5bd93675b82045d1ca7541e45981fe0440631a6/examples/tts/conf/talknet-durs.yaml#L66

Instead of commenting those lines, can you just check for a top level key or something like that?

stasbel: I really can't come up with definite way to separate config from being top-level vs inner node.
1) There is no parent link in OmegaConf.
2) A check like `'name' in conf` or `'sample_rate' in conf` should work for almost all cases, but there is always test case against it.

I suggest maybe we should't able to load models and overriding restoration with top-level configs. Model is something that have access to only 'model' part of initial config, so when loading it should only operate with `conf.model` part of it.

titu1994: Woops sorry misclick. 
Yes only cfg.model will be saved in Nemo config. Yes, you cannot have cfg.model.model in your config name. It is because save restore does not know if config is from user in constrictor of from file inside .Nemo so it needs to bring "cfg.model" to just model level so that OmegaConf resolution works properly. 

It won't be reverted so I suggest changing cfg.model.model to cfg.model.<some other name>. It is anyway awkward having cfg.model.model. The outer model represents the whole model, therefore inner model name makes no sense. You can call if module if you absolutely want to, or pick another name.

titu1994: > I suggest maybe we should't able to load models and overriding restoration with top-level configs.

This is required in useful cases, so it will be maintained.

nemo/collections/tts/models/talknet.py
```diff
@@ -134,20 +134,23 @@ class TalkNetPitchModel(ModelPT):
 
     def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
         super().__init__(cfg=cfg, trainer=trainer)
+
+        print(cfg)
```
blisc, line 138: Remove please

nemo/collections/tts/models/talknet.py
```diff
@@ -134,20 +134,23 @@ class TalkNetPitchModel(ModelPT):
...
+
         typecheck.set_typecheck_enabled(enabled=False)
```
blisc, line 140: Remove all `typecheck.set_typecheck_enabled(enabled=False)` from your model.

nemo/core/classes/modelPT.py
```diff
@@ -91,6 +91,11 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # Convert config to support Hydra 1.0+ instantiation
         cfg = model_utils.maybe_update_config_version(cfg)
 
+        if 'model' in cfg:
+            raise ValueError(
+                "Creating model config node is forbidden due to collision problem when loading from checkpoint."
+            )
+
```
blisc, line 98: Remove please.
What exactly is the collision?

tutorials/tts/Inference_ModelSelect.ipynb
```diff
@@ -180,7 +178,7 @@
     "        override_conf = conf\n",
     "    elif spectrogram_generator == \"talknet\":\n",
     "        from nemo.collections.tts.models import TalkNetSpectModel\n",
-    "        pretrained_model = \"tts_en_talknet\" \n",
+    "        return TalkNetSpectModel.from_pretrained_all(\"tts_en_talknet\", override_config_path=override_conf)\n",
```
blisc, line 181: Revert

nemo/collections/tts/models/talknet.py
```diff
@@ -370,23 +373,27 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
+        list_of_models.append(model)
+        return list_of_models
+
+    @staticmethod
+    def from_pretrained_all(model_name: str, *args, **kwargs):
+        """Custom TalkNet's three-part load logic."""
+        model = TalkNetSpectModel.from_pretrained(model_name, *args, **kwargs)
+        model.add_module('_pitch_model', TalkNetPitchModel.from_pretrained(model_name, *args, **kwargs))
+        model.add_module('_durs_model', TalkNetDursModel.from_pretrained(model_name, *args, **kwargs))
+        return model
```
blisc, line 399: Just redefine from_pretrained in your class, don't make a new function.

blisc (changes requested):

tutorials/tts/Inference_ModelSelect.ipynb
blisc: See https://github.com/NVIDIA/NeMo/pull/3092#discussion_r753295648

nemo/core/classes/modelPT.py
blisc: If this works correctly, please add a test that checks that conf.model passes this init but conf.model.model does not

nemo/collections/tts/models/talknet.py
stasbel: That what I have in mind in the first place, but I just can't do this by design.
Since the method is static, we need `TalkNetSpectModel.from_pretrained` to be available during loading, otherwise it's gonna be an infinite cycle of calling from_pretrained func.

README.rst
```diff
@@ -70,7 +70,7 @@ Key Features
     * `Neural Duplex Text Normalization <https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/nlp/text_normalization.html>`_
     * `NGC collection of pre-trained NLP models. <https://ngc.nvidia.com/catalog/collections/nvidia:nemo_nlp>`_
 * `Speech synthesis (TTS) <https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/tts/intro.html#>`_
-    * Spectrogram generation: Tacotron2, GlowTTS, FastSpeech2, FastPitch, FastSpeech2
+    * Spectrogram generation: Tacotron2, GlowTTS, FastSpeech2, TalkNet, FastPitch, FastSpeech2
```
Oktai15, line 73: Remove double FastSpeech2

Oktai15 (changes requested):
This notebook is broken: https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/TalkNet_Training.ipynb

Fix it too, please

README.rst
stasbel: yep, thx

nemo/collections/tts/models/talknet.py
blisc: - call `super().from_pretrained` instead of `TalkNetSpectModel.from_pretrained()`?
- Make a new model? `TalkNetFullModel` that defines all three as submodules and use that model for inference?
- Why is it a staticmethod? You have it defined as a classmethod and the base class's from_pretrained is also defined as a classmethod.

nemo/collections/tts/models/talknet.py
stasbel: If I do remove enabled=False line, then I require to type all forward methods.
Just to clarify: is this a requirement now again?

blisc: It always was. I must have missed it when I first approved TalkNet

stasbel: fixed

nemo/collections/tts/models/talknet.py
stasbel: great, thx
I am actually didn't know super() works for staticmethods

blisc:
> This notebook is broken: https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/TalkNet_Training.ipynb
> 
> Fix it too, please

@Oktai15 @stasbel: Has this been fixed?

stasbel:
@blisc @Oktai15 ready 4 review

blisc (dismissed):
LGTM

Oktai15 (approved):

stasbel:
CI seems to be broken on the main branch (as well as for this PR)
waiting to be fixed

PR #3096, opened 2021-10-30
NMT checkpoint averaging

michalivne:
This PR adds convenience scripts to checkpoint average any `.nemo` file (regardless of the class).
Python script is heavily based on code by @MaximumEntropy and @titu1994.

# Usage

The script `checkpoint_averaging.py` builds a `.nemo` file with average weights over multiple `.ckpt` files (assumes `.ckpt` files in same folder as `.nemo` file).
The input is a list of folders that contains a .nemo file (or a list of .nemo files).
The output of the script (per input folder / `.nemo` file) is a file named `*-averaged.nemo` where `*.nemo` is the original `.nemo` file.

# Example

Usage example for building `*-averaged.nemo` for a given `.nemo` file:

```
NeMo/scripts/checkpoint_averaging/checkpoint_averaging.py my_model.nemo
```

Usage example for building *-averaged.nemo files for all results in sub-directories under current path:

```
find . -name '*.nemo' | grep -v -- "-averaged.nemo" | xargs NeMo/scripts/checkpoint_averaging/checkpoint_averaging.py
```


titu1994:
This already exists in a majority form here - https://github.com/NVIDIA/NeMo/blob/main/scripts/average_model_checkpoints.py

MaximumEntropy:
Thanks @michalivne :) Maybe like Som suggested, we can move https://github.com/NVIDIA/NeMo/blob/main/scripts/average_model_checkpoints.py#L120:L167 to a separate function that gets called across both ways of using checkpoint averaging?

michalivne:
@MaximumEntropy @titu1994 script is finished based on all comments. It is now very easy to. use, supposed to work with all nemo classes, and does not rely on a particular internal structure of `.nemo` files. Can you take a quick look?

scripts/checkpoint_averaging/checkpoint_averaging.py
```diff
@@ -0,0 +1,110 @@
...
+import torch
+
+from nemo.core import ModelPT
+from nemo.utils import logging, model_utils
+
+
+def main():
+    device = torch.device("cpu")
+
+    model_fname_list = sys.argv[1:]
```
titu1994, line 40: Why not just use argparse for future changes?

michalivne: I typically use it, but found it an overkill here.
I can use it here as well if you believe it to be useful.

michalivne: For the sake of uniformity I switched to argsparse.

titu1994 (dismissed):
Looks great

titu1994 (dismissed):

titu1994 (approved):

PR #3097, opened 2021-10-30
Fixing unneeded warmup batch when no timing is recored

michalivne:
Fixed continuous warmup with no write_timing is selected.


michalivne:
PR has been moved to r1.5.0

PR #3115, opened 2021-11-01
`WER` and `WERBPE` metrics are computed wrongly if method `update` is called more than once before method `compute`.

### LINKED ISSUE #3114: `WER` and `WERBPE` metrics are computed wrongly if method `update` is called more than once before method `compute`.
**Describe the bug**

`WER` and `WERBPE` metrics are computed wrongly if method `update` is called more than once before method `compute`.

```python
import string
from functools import partial
from itertools import chain
from typing import List, Tuple

import torch
from torch.nn.utils.rnn import pad_sequence

from nemo.collections.asr.metrics.wer import WER


VOCABULARY = [' '] + list(string.ascii_lowercase) + ["'"]
predictions=[
    ["aaa aaa", "b b b b b b"],
    ["sky sea", "s k y s e a"],
]
targets=[
    ["ccc ccc", "d d d d d d"],
    ["sky sea", "s k y s e a"],
]

class Encoder:
    def __init__(self, vocabulary: List[str]):
        self.labels_map = vocabulary.copy()
        self.blank_id = len(self.labels_map)
        self.inv_vocabulary = {v: i for i, v in enumerate(self.labels_map)}

    def string_to_ctc_tensor(self, txt: str) -> torch.Tensor:
        # This function emulates how CTC output could like for txt
        string_in_id_form = self.text_to_ids(txt)
        ctc_list = []
        prev_id = -1
        for c in string_in_id_form:
            # when character is repeated we need to insert CTC blank symbol
            if c != prev_id:
                ctc_list.append(c)
            else:
                ctc_list.append(self.blank_id)
                ctc_list.append(c)
            prev_id = c
        return torch.Tensor(ctc_list)

    def text_batch_to_tensor(self, batch: List[str], ctc: bool) -> Tuple[torch.Tensor, torch.Tensor]:
        batch = [self.string_to_ctc_tensor(s) if ctc else self.string_to_tensor(s) for s in batch]
        lengths = torch.tensor([len(s) for s in batch])
        return pad_sequence(batch, batch_first=True, padding_value=0), lengths

    def batch_of_text_batches_to_tensor(
        self,
        batch_of_text_batches: List[List[str]],
        ctc: bool,
    ) -> Tuple[torch.Tensor, torch.Tensor]:
        huge_batch, lengths = self.text_batch_to_tensor(list(chain(*batch_of_text_batches)), ctc)
        first_dims = [len(batch_of_text_batches), len(batch_of_text_batches[0])]
        return huge_batch.reshape(first_dims + [-1]), lengths.reshape(first_dims)

    def string_to_tensor(self, txt: str) -> torch.Tensor:
        return torch.Tensor(self.text_to_ids(txt))

    def text_to_ids(self, txt: str) -> List[int]:
        return [self.inv_vocabulary[c] for c in txt]

    
def test_wer(
    predictions: torch.Tensor,
    targets: torch.Tensor,
    target_lengths: torch.Tensor,
    predictions_lengths: torch.Tensor,
):
    num_batches = predictions.shape[0]
    wer_metric = WER(vocabulary=VOCABULARY)
    for i in range(num_batches):
        wer_metric(predictions[i], targets[i], target_lengths[i], predictions_lengths[i])

    # check on all batches on all ranks
    result = wer_metric.compute()[0]
    print("Result:", result)

encoder_decoder = Encoder(vocabulary=VOCABULARY)
predictions, predictions_lengths = encoder_decoder.batch_of_text_batches_to_tensor(predictions, ctc=True)
targets, target_lengths = encoder_decoder.batch_of_text_batches_to_tensor(targets, ctc=False)

test_wer(predictions, targets, target_lengths, predictions_lengths)
```

Output:
```
[NeMo I 2021-11-01 08:03:28 wer:225] 
    
[NeMo I 2021-11-01 08:03:28 wer:226] reference:dog cat
[NeMo I 2021-11-01 08:03:28 wer:227] predicted:cat dog
[NeMo I 2021-11-01 08:03:28 wer:225] 
    
[NeMo I 2021-11-01 08:03:28 wer:226] reference:sky sea
[NeMo I 2021-11-01 08:03:28 wer:227] predicted:sky sea
Result: tensor(0.)
```

**Expected behavior**

The model should output `0.5` instead of `0` since the first prediction-target pair has maximum possible Levenstein distance and the same number of words as the second pair.

 - Environment location: Bare-metal
 - Method of NeMo install: pip install

**Environment details**

If NVIDIA docker image is used you don't need to specify these.
Otherwise, please provide:
- OS version Ubuntu 20.04.2 LTS
- PyTorch version 1.10.0
- Python version 3.8

**Additional context**

The problem is in updating metric states [here](https://github.com/NVIDIA/NeMo/blob/25c61f2e6f68af7fe90853b11bd073e6b2625c72/nemo/collections/asr/metrics/wer.py#L259) and [here](https://github.com/NVIDIA/NeMo/blob/25c61f2e6f68af7fe90853b11bd073e6b2625c72/nemo/collections/asr/metrics/wer_bpe.py#L218). There should be `+=` instead of `=`.

PeganovAnton:
Closes #3114


titu1994:
This is done intentionally - the use of these two metrics is - update, compute, reset cycle per step. Not accumulated across more than one step. 

The reason for this is that PTL calls metrics in a weird order, and doesn't reset them automatically, so when doing multi data loader (common in ASR dev and test sets), the metrics of one dataset roll over into another. 

As a solution we keep the metric itself stateless and manually track the metrics per dataset.

titu1994 (changes requested):
This is by requirement. We won't be making the metrics stateful

titu1994:
Closing for now, lets reopen later if needed.

PR #3117, opened 2021-11-01
add offsets calculation to MegatronGPTModel.complete method

dimapihtar:
Signed-off-by: dimapihtar <dpykhtar@nvidia.com>


okuchaiev (approved):
lgtm

examples/nlp/language_modeling/megatron_gpt_eval.py
```diff
@@ -33,7 +33,7 @@ def main():
     parser = ArgumentParser()
     parser.add_argument("--model_file", type=str, default="", required=True, help="Pass path to model's .nemo file")
     parser.add_argument(
-        "--prompt", type=str, default="", required=True, help="Prompt for the model (a text to complete)"
+        "--path_to_file", type=str, default="", required=True, help="Path to file with prompts (a text to complete)"
```
ericharper, line 36: Do we have to remove `--prompt` ? It's nice to be able to feed in some text from the CLI.

ericharper (changes requested):
There's a binary file that needs to be removed from the PR: examples/nlp/language_modeling/.megatron_gpt_eval.py.swp

examples/nlp/language_modeling/megatron_gpt_eval.py
dimapihtar: Should I create a separate `.py` example for evaluation prompts from file?

okuchaiev: @dimapihtar  let's just have both "--prompt" and "--path_to_file" options and require at least one to be specified. If both are specified, use "--prompt" and ignore "--path_to_file"

ericharper (approved):
LGTM. Thanks!

PR #3118, opened 2021-11-01
Enable param groups for ModelPT optimizer

ericharper:
Adds `.setup_optimizer_param_groups` method to `ModelPT`. The `.setup_optimization` method will give optimizers `self._optimizer_param_groups` instead of `self.parameters()`.

The `.setup_optimizer_param_groups` method will by default set `self._optimizer_param_groups=self.parameters()`.

This enables NeMo models to specify custom param groups so that, for example, per layer learning rates can be used.

In particular, this enables Megatron-LM style weight decay where LayerNorm and Bias weights are not decayed.

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -215,6 +215,14 @@ class MegatronGPTModel(NLPModel):
     def setup_optimizer_param_groups(self):
         self._optimizer_param_groups = self._get_params_for_weight_decay_optimization()
 
+        assert len(list(self.parameters())) == len(self._optimizer_param_groups[0]['params']) + len(
+            self._optimizer_param_groups[1]['params']
+        )
+
+        assert sum([p.sum() for p in self.parameters()]) == sum(
+            [p.sum() for p in self._optimizer_param_groups[0]['params']]
+        ) + sum([p.sum() for p in self._optimizer_param_groups[1]['params']])
+
     def process_batch(self, batch):
 
         # Items and their type.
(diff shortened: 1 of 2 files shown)


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 import re
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 import torch
+from apex.normalization.fused_layer_norm import MixedFusedLayerNorm as LayerNorm
```
titu1994, line 19: Doesn't Apex need to be import guarded ?

nemo/core/classes/modelPT.py
```diff
@@ -421,6 +422,14 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
                 The list of "arg_value" will be parsed and a dictionary of optimizer kwargs \
                 will be built and supplied to instantiate the optimizer.
         """
+        if self._optimizer_param_groups is None:
+            logging.info(
```
titu1994, line 426: Why not just call self.setup_optimizer_param_groups() here instead of raising a warning and exiting ?
If it's None after calling that, directly raise error

titu1994: Plus this way setup param groups gets called by used just calling setup_optimization() in one go.

nemo/core/classes/modelPT.py
```diff
@@ -571,7 +580,26 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
...
+            By default, ModelPT will use self.parameters().
+            Override this method to add custom param groups.
+    """
+        param_groups = None
+        if hasattr(self, 'parameters'):
+            param_groups = self.parameters()
+        self._optimizer_param_groups = param_groups
+
     def configure_optimizers(self):
+        self.setup_optimizer_param_groups()
```
titu1994, line 602: Let's remove this and call it inside setup_optimization. Cleaner that way

titu1994 (commented):
Overall it's fine but it can be made cleaner by falling new setup method inside setup_optimization

ericharper:
Closing temporarily to save CI bandwidth.

PR #3119, opened 2021-11-01
Multiple ASR Fixes to SPE tokenization

titu1994:
# Changelog
- Fix issue with transcribe building more workers than samples
- Fix SPE tokenizer vocab being built incorrectly for tokenizers with BOS EOS and PAD
- Update tokenizer script to designate tokenizer dir with BOS EOS and PAD token.

# Breaking change
- SPE tokenizer based models, when requested for their vocab via model.tokenizer.tokenizer.get_vocab() will return special tokens + order of tokens will be different than previous result.

### DIFF
--- a/nemo/collections/asr/parts/mixins/mixins.py
+++ b/nemo/collections/asr/parts/mixins/mixins.py
@@ -76,13 +76,12 @@ class ASRBPEMixin(ABC):
 
             if 'special_tokens' in self.tokenizer_cfg:
                 special_tokens = self.tokenizer_cfg['special_tokens']
-            else:
-                special_tokens = None
+
+                if special_tokens is not None:
+                    raise ValueError("`special_tokens` are no longer supported for SentencePiece based tokenizers.")
 
             # Update special tokens
-            self.tokenizer = tokenizers.SentencePieceTokenizer(
-                model_path=model_path, special_tokens=special_tokens, legacy=True
-            )
+            self.tokenizer = tokenizers.SentencePieceTokenizer(model_path=model_path)
 
             if 'vocab_path' in self.tokenizer_cfg:
                 vocab_path = self.tokenizer_cfg.get('vocab_path')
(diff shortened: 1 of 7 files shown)


nemo/collections/asr/parts/mixins/mixins.py
```diff
@@ -102,11 +101,11 @@ def _setup_tokenizer(self, tokenizer_cfg: DictConfig):
                 # fallback case for older checkpoints that did not preserve the tokenizer.vocab
                 self.spe_vocab_path = None
 
-            vocabulary = {'<unk>': 0}
-            with open(vocab_path) as f:
-                for i, piece in enumerate(f):
-                    piece = piece.replace('\n', '')
-                    vocabulary[piece] = i + 1
+            vocabulary = {}
```
VahidooX, line 104: Does it work with all types of the tokenizers?

nemo/collections/asr/parts/mixins/mixins.py
```diff
@@ -76,13 +76,12 @@ def _setup_tokenizer(self, tokenizer_cfg: DictConfig):
...
-                special_tokens = None
+
+                if special_tokens is not None:
+                    raise ValueError("`special_tokens` are no longer supported for SentencePiece based tokenizers.")
 
             # Update special tokens
-            self.tokenizer = tokenizers.SentencePieceTokenizer(
-                model_path=model_path, special_tokens=special_tokens, legacy=True
-            )
+            self.tokenizer = tokenizers.SentencePieceTokenizer(model_path=model_path)
```
VahidooX, line 84: What there a reason "legacy=True" is dropped?

VahidooX (approved):

nemo/collections/asr/parts/mixins/mixins.py
titu1994: All SPE type tokenizers yes. WPE has its vocab file so different handling

nemo/collections/asr/parts/mixins/mixins.py
titu1994: its not necessary any more. Now we use script to embed the EOS, BOS and PAD tokens along with UNK. Previously this was handled in a weird nemo specific manner (legacy mode). 
Since we are supporting it embedded in the tokenizer now, its good to drop legacy support too.

PR #3124, opened 2021-11-02
[BigNLP] Add CI Tests for Megatron GPT

ericharper:


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -64,7 +63,7 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
             seed=self.cfg.get('seed', 1234),
         )
 
-        if not self.cfg.get('fused_bf16'):
+        if self.cfg.get('precision') != 'bf16':
```
MaximumEntropy, line 66: I don't think we have access to the train config here right? Isn't this the model config that doesn't have the precision param?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -100,8 +99,7 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
             fp16_lm_cross_entropy=cfg.get('fp16_lm_cross_entropy', False),
             use_cpu_initialization=cfg.get('use_cpu_initialization', False),
             hidden_dropout=cfg.get('hidden_dropout', 0.1),
-            fused_fp16=cfg.get('fused_fp16', False),
-            fused_bf16=cfg.get('fused_bf16', False),
+            precision=cfg.get('precision', 16),
```
MaximumEntropy, line 102: Same comment here - maybe we want to add precision to the model config via ${trainer.precision}?

MaximumEntropy (commented):
LGTM! Minor comments about adding precision to the model config.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
ericharper: Yup, that's right. I'll make the change.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
ericharper: We add it to the config in the training script.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
ericharper: I ended up not being able to use interpolation as the ${trainer.precision} could not be found when PTL saves the hparams. I added a comment explaining why we add the precision to the model config in the training script.

okuchaiev (approved):

PR #3126, opened 2021-11-02
Change cfg.cuda type to optional int in schema

erastorgueva-nv:
Signed-off-by: Elena Rastorgueva <erastorgueva@nvidia.com>

### DIFF
--- a/examples/asr/transcribe_speech.py
+++ b/examples/asr/transcribe_speech.py
@@ -63,7 +63,7 @@ class TranscriptionConfig:
     batch_size: int = 32
     # Set `cuda` to int to define CUDA device. If 'None', will look for CUDA
     # device anyway, and do inference on CPU only if CUDA device is not found.
-    # If `cuda` is -1, inference will be on CPU only.
+    # If `cuda` is a negative number, inference will be on CPU only.
     cuda: Optional[int] = None
     amp: bool = False
     audio_type: str = "wav"
@@ -90,7 +90,7 @@ def main(cfg: TranscriptionConfig):
     else:
         device_id = cfg.cuda
 
-    device = torch.device(f'cuda:{device_id}' if cfg.cuda != -1 else 'cpu')
+    device = torch.device(f'cuda:{device_id}' if cfg.cuda > 0 else 'cpu')
     print(device)
 
     # setup model
--- a/examples/asr/transcribe_speech.py
+++ b/examples/asr/transcribe_speech.py
@@ -84,9 +84,9 @@ def main(cfg: TranscriptionConfig):
     # setup GPU
     if cfg.cuda is None:
         if torch.cuda.is_available():
-            cfg.cuda = 0 # use 0th CUDA device
+            cfg.cuda = 0  # use 0th CUDA device
         else:
-            cfg.cuda = -1 # use CPU
+            cfg.cuda = -1  # use CPU
     else:
         device_id = cfg.cuda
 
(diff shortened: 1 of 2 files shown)


titu1994 (changes requested):
You need to update the logic below which checks for cuda as a bool and make it work for cuda:0 for device 0 (right now it will think 0 == False don't use cuda). Use negative number as default to enforce disabling cuda

examples/asr/transcribe_speech.py
```diff
@@ -80,14 +83,15 @@ def main(cfg: TranscriptionConfig):
 
     # setup GPU
     if cfg.cuda is None:
-        cfg.cuda = torch.cuda.is_available()
-
-    if type(cfg.cuda) == int:
-        device_id = int(cfg.cuda)
+        if torch.cuda.is_available():
```
titu1994, line 86: Cfg.cuda can't be none for int, check if cfg.cuda < 0

examples/asr/transcribe_speech.py
```diff
@@ -80,14 +83,15 @@ def main(cfg: TranscriptionConfig):
...
+            cfg.cuda = 0  # use 0th CUDA device
+        else:
+            cfg.cuda = -1  # use CPU
     else:
-        device_id = 0
+        device_id = cfg.cuda
 
-    device = torch.device(f'cuda:{device_id}' if cfg.cuda else 'cpu')
+    device = torch.device(f'cuda:{device_id}' if cfg.cuda >= 0 else 'cpu')
+    print(device)
```
titu1994, line 94: Revert print before merge

examples/asr/transcribe_speech.py
```diff
@@ -103,7 +107,7 @@ def main(cfg: TranscriptionConfig):
         asr_model = ASRModel.from_pretrained(model_name=cfg.pretrained_name, map_location=device)  # type: ASRModel
         model_name = cfg.pretrained_name
 
-    trainer = pl.Trainer(gpus=int(cfg.cuda))
+    trainer = pl.Trainer(gpus=1 if cfg.cuda > 0 else 0)
```
titu1994, line 110: No this will always use gpu 0. You have to pass gpus=[cfg.cuda]

titu1994 (changes requested):

examples/asr/transcribe_speech.py
erastorgueva-nv: I'm not sure what the issue was here. cfg.cuda (now cfg.cuda_device) can be None if it is unspecified.
I think I have resolved all issues in the latest commits.

examples/asr/transcribe_speech.py
```diff
@@ -61,7 +61,10 @@ class TranscriptionConfig:
     # General configs
     output_filename: Optional[str] = None
     batch_size: int = 32
-    cuda: Optional[bool] = None  # will switch to cuda if available, defaults to CPU otherwise
+    # Set `cuda_device` to int to define CUDA device. If 'None', will look for CUDA
+    # device anyway, and do inference on CPU only if CUDA device is not found.
+    # If `cuda_device` is a negative number, inference will be on CPU only.
+    cuda_device: Optional[int] = None
```
titu1994, line 67: Pedantic, but can be revert the name cuda_device and just call if cuda.
Based to pytorch - ("cuda:0") in

titu1994 (commented):
This looks fine but potential reversion of name v

titu1994 (approved):

titu1994 (approved):

PR #3129, opened 2021-11-02
pyt21.11 compatibility: (1) enable nvfuser for bf16 (2) change of api…

erhoo82:
The current NeMo code will crash at PyTorch21.11. 
This update will crash at PyTorch21.10 so should be made sure this is packaged with PyTorch21.11.

Signed-off-by: Sangkug Lym <slym@nvidia.com>


ericharper:
Closed in favor of #3234

PR #3130, opened 2021-11-03
Add missing tag in ALL_TAGS list

choclatier:
There is a missing tag which prevents the `DuplexTagger` Model from loading correctly


yzhang123:
hi @choclatier, thanks for the feedback. we will be updating the checkpoint, so it will be loaded by the code

PR #3134, opened 2021-11-03
Update a few packages' version

fayejf:


requirements/requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy>=1.18.2
 onnx>=1.7.0
 python-dateutil
-torch>1.7
+torch>=1.10
```
titu1994, line 4: This shouldnt be there, it will override container version of torch

titu1994 (changes requested):
Overall looks great, but revert the PT >= 1.10 for now, we will update it in the future relese after subsequent 21.11 comes out.

titu1994 (dismissed):
LGTM, @ericharper for final approval before merge.

ericharper (approved):
LGTM. Thanks!

PR #3136, opened 2021-11-04
Cannot unset both `source_processor` and `target_processor` for NMT model inference.

PeganovAnton:
If I omit both parameters `--target_lang` and `--source_lang` when running script `nmt_transformer_infer.py`, then both `target_processor` and `source_processor` attributes of model `MTEncDecModel` are set as during the model training.

This behaviour appears because 
1) `source_processor` and `target_processor` are set while [restoring from checkpoint](https://github.com/NVIDIA/NeMo/blob/a089791db1d4be51da2cd7cdec2eb822ca6b588c/examples/nlp/machine_translation/nmt_transformer_infer.py#L154),
2) `source_processor` and `target_processor` are [reset](https://github.com/NVIDIA/NeMo/blob/a089791db1d4be51da2cd7cdec2eb822ca6b588c/nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py#L858) only if at least one of parameters `--source_lang`, `--target_lang` are provided.

To produce the problem you need
1) find `.nemo` file for `MTEncDecModel` model which was trained with source or target processor,
2) add some debug logging [here](https://github.com/NVIDIA/NeMo/blob/a089791db1d4be51da2cd7cdec2eb822ca6b588c/nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py#L763) to see if processing is performed,
3) launch script `nmt_transformer_infer.py` without parameters `--source_lang` and `--target_lang`.

### DIFF
--- a/examples/nlp/machine_translation/nmt_transformer_infer.py
+++ b/examples/nlp/machine_translation/nmt_transformer_infer.py
@@ -118,10 +118,24 @@ def main():
         help="Stop generating if target sequence length exceeds source length by this number.",
     )
     parser.add_argument(
-        "--target_lang", type=str, default=None, help="Target language identifier ex: en,de,fr,es etc."
+        "--target_lang",
+        type=str,
+        default=None,
+        help="Target language identifier ex: en,de,fr,es etc. If both `--target_lang` and `--source_lang` are "
+        "no set, then target language processing will be done the same way as during model training. If "
+        "`--target_lang` parameter is not set but `--source_lang` parameter is set, then target language "
+        "processing will not be performed. If `--target_lang` equals 'ignore', then target language processing "
+        "will not be performed regardless of value of `--source_lang` parameter.",
     )
     parser.add_argument(
-        "--source_lang", type=str, default=None, help="Source language identifier ex: en,de,fr,es etc."
+        "--source_lang",
+        type=str,
+        default=None,
+        help="Source language identifier ex: en,de,fr,es etc. If both `--target_lang` and `--source_lang` are "
+        "no set, then source language processing will be done the same way as during model training. If "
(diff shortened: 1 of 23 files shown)


MaximumEntropy (commented):
With this PR, the user always needs to specify the source and target lang correctly all the time. We had discussed earlier that if source and target are left as None, we should use the model's internal src_lang and tgt_lang attributes. I think a better way to ignore source and target processor is to use something like if src_lang == 'ignore' then set it to None when calling `setup_pre_and_post_processing_utils`

PeganovAnton:
@MaximumEntropy thank you! Fixed.

examples/nlp/machine_translation/nmt_transformer_infer.py
```diff
@@ -118,10 +118,24 @@ def main():
         help="Stop generating if target sequence length exceeds source length by this number.",
     )
     parser.add_argument(
-        "--target_lang", type=str, default=None, help="Target language identifier ex: en,de,fr,es etc."
+        "--target_lang",
+        type=str,
+        default=None,
+        help="Target language identifier ex: en,de,fr,es etc. If both `--target_lang` and `--source_lang` are "
+        "no set, then target language processing will be done the same way as during model training. If "
```
MaximumEntropy, line 125: Thanks for this detailed help message! Minor typo no set -> not set.

MaximumEntropy (approved):
LGTM! Thanks for the changes. Just a minor typo.

examples/nlp/machine_translation/nmt_transformer_infer.py
PeganovAnton: @MaximumEntropy , I fixed typo

MaximumEntropy (approved):
Thanks for the typo fixes. Please merge on CI passes.

PR #3140, opened 2021-11-04
Fixing unneeded warmup batch when no timing is recored

michalivne:
Fixed continuous warmup with no write_timing is selected.


MaximumEntropy (approved):
LGTM!

PR #3152, opened 2021-11-09
Patch LR max_step computation check

titu1994:
# Bugfix
- Compute explicit max steps if max_steps = None or negative numbers (new default in PTL 1.5.0+)

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -455,7 +455,7 @@ class ModelPT(LightningModule, Model):
         if 'sched' in optim_config and self._trainer is not None:
             if not isinstance(self._trainer.accumulate_grad_batches, int):
                 raise ValueError("We do not currently support gradient acculumation that is not an integer.")
-            if self._trainer.max_steps is None:
+            if self._trainer.max_steps is None or self.trainer.max_steps <= 0:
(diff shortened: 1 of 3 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -455,7 +455,7 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
         if 'sched' in optim_config and self._trainer is not None:
             if not isinstance(self._trainer.accumulate_grad_batches, int):
                 raise ValueError("We do not currently support gradient acculumation that is not an integer.")
-            if self._trainer.max_steps is None:
+            if self._trainer.max_steps is None or self.trainer.max_steps <= 0:
```
blisc, line 458: Can you comment on why we need <=0: Like # Lightning maps max_steps = None to -1

blisc (approved):

nemo/core/classes/modelPT.py
titu1994: Well my idea was 0 max steps is odd no ? I can revert it to < 0 though

titu1994: Reverted to < 0

PR #3166, opened 2021-11-11
Text2sparql fix 1.5.0

MaximumEntropy:
Fixes

1. Accelerator = None
2. HF API change to outputs. Removed output neural types from the model since HF returns a tuple of tensors now.

### DIFF
--- a/nemo/collections/nlp/models/text2sparql/text2sparql_model.py
+++ b/nemo/collections/nlp/models/text2sparql/text2sparql_model.py
@@ -157,9 +157,6 @@ class Text2SparqlModel(ModelPT):
         Lightning calls this inside the validation loop with the data from the validation dataloader
         passed in as `batch`. Loss calculation from HuggingFace's BartForConditionalGeneration.
         """
-        import ipdb
-
-        ipdb.set_trace()
         input_ids, input_mask, decoder_input_ids, labels = batch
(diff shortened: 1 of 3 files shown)


titu1994:
Neural types support containers - https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/core/core.html#container-types (just use one [] though).

ericharper (approved):
LGTM. Thanks!

nemo/collections/nlp/models/text2sparql/text2sparql_model.py
```diff
@@ -44,14 +44,6 @@ def input_types(self) -> Optional[Dict[str, NeuralType]]:
             "labels": NeuralType(('B', 'T'), ChannelType(), optional=True),
         }
 
-    @property
```
okuchaiev, line 47: Why is removing this necessary for 1.5.0?

MaximumEntropy: Couple of reasons

1) HF changed their return API to return 4 things rather than 3 like before.
2) The extra thing being return is a heavily listed tuple of tuple of tensors corresponding to keys and values for each layer I think. Which is complicated to specify neural types for.

okuchaiev (approved):

PR #3170, opened 2021-11-11
Fix cfg issue for r1.5.0

fayejf:
Fix issues with config saving TypeError: __init__() missing 1 required positional argument: 'cfg' when trying to do model.load_from_checkpoint(XYZ)

related to https://github.com/NVIDIA/NeMo/pull/3058 and https://github.com/PyTorchLightning/pytorch-lightning/discussions/7525

PR #3172, opened 2021-11-12
remove cfg in _load_model_state in nlp_models

fayejf:
We patched cfg issue in modelPT level in https://github.com/NVIDIA/NeMo/pull/3170.
Revert the cfg arg in `_load_model_state` in nlp_models in https://github.com/NVIDIA/NeMo/pull/3149 since it will break the fix and is deprecated in PTL. 


ericharper (approved):
LGTM. Thanks!

PR #3201, opened 2021-11-16
Restore Quick Fix for now

blisc:


titu1994 (approved):
Looks like a good fix for 1.5 release. We need to discuss with PTL team for long term stability fyi - @ericharper @okuchaiev

ericharper (approved):
LGTM. Thanks!

PR #3207, opened 2021-11-17
Add Pitch, Duration Tensors for Riva

blisc:


nemo/collections/tts/models/fastpitch.py
```diff
@@ -386,38 +389,62 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
+        Generates input examples for tracing etc.
+        Returns:
+            A tuple of input examples.
+        """
+        par = next(self.fastpitch.parameters())
+        inp = torch.randint(
+            0, self.fastpitch.encoder.word_emb.num_embeddings, (1, 44), device=par.device, dtype=torch.int64
+        )
+        pitch = torch.randn((1, 44), device=par.device, dtype=torch.float32) * 0.5
+        pace = torch.clamp((torch.randn((1, 44), device=par.device, dtype=torch.float32) + 1) * 0.1, min=0.01)
```
redoctopus, line 404: Shouldn't pace be a float?

nemo/collections/tts/models/fastpitch.py
```diff
@@ -386,38 +389,62 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
+        return self._output_types
+
+    def _prepare_for_export(self, **kwargs):
+        super()._prepare_for_export(**kwargs)
+
+        # Define input_types and output_types as required by export()
+        self._input_types = {
+            "text": NeuralType(('B', 'T'), TokenIndex()),
+            "pitch": NeuralType(('B', 'T'), RegressionValuesType()),
+            "pace": NeuralType(('B', 'T'), optional=True),
```
redoctopus, line 433: Similarly here.

nemo/collections/tts/models/fastpitch.py
blisc: No, I'm not looking for a global pace that gets broadcasted to each predicted duration. I want a pace per text token, hence it will be the same size and shape as the tokens

redoctopus: In that case, should `pace=1.0` be changed in the `infer()` definition?

blisc: Doesn't really matter. In python, it accepts both a float or a [B, T] tensor since it'll broadcast.

redoctopus: Can you add a comment to specify that both are okay to avoid future confusion?

blisc: Both are not ok from an export perspective. I require, in onnx, that pace must be a tensor of [B, T] which is what this function `input_example` does. That is why I only define input and output types during `_prepare_for_export`.
I added a `export_teardown` to delete input and output types post export.

redoctopus (approved):

PR #3213, opened 2021-11-18
text2sparql val loss logging fix

okuchaiev:
Signed-off-by: Oleksii Kuchaiev <okuchaiev@nvidia.com>


ericharper (approved):
Let's merge if the notebook doesn't crash.

Thanks!

ericharper (approved):
Was able to run the whole notebook end-2-end on my workstation.

PR #3215, opened 2021-11-18
WIP: Persistent layer norm and module dependency check

erhoo82:
This adds the option to use the persistent layer norm kernel from APEX.
The kernel is available from PyT container v21.11 so added a dependency check. Other pytorch- or container-dependency modules can have dependency checks in the same method. 

Signed-off-by: Sangkug Lym <slym@nvidia.com>


okuchaiev:
This PR seems to fail on code-formatting checks. Have a look here https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md

PR #3216, opened 2021-11-19
Fix Masked SE for Citrinets + export Limited Context Citrinet

titu1994:
# Changelog
- Fix masking in SqueezeExcite block where padding causes incorrect calculation of global average. Slight improvements in WER found after retraining / fine-tuning. Masking should no longer be affected by batch size.
- Add sequence length tracking to Jasper block.


nemo/collections/asr/modules/conv_asr.py
```diff
@@ -79,35 +80,18 @@ def input_example(self):
...
-        else:
-            return set(["encoded_lengths"])
-
-    def save_to(self, save_path: str):
-        pass
-
-    @classmethod
-    def restore_from(cls, restore_path: str):
-        pass
+        return set([])
```
borisfom, line 94: Better to remove the function altogether ?

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -79,35 +80,18 @@ def input_example(self):
...
+        return tuple([input_example, lens])
 
     @property
     def disabled_deployment_input_names(self):
         """Implement this method to return a set of input names disabled for export"""
-        if self._rnnt_export:
-            return set([])
-        else:
-            return set(["length"])
+        return set([])
```
borisfom, line 89: Better to remove the function altogether ?

borisfom (dismissed):

nemo/collections/asr/modules/conv_asr.py
titu1994: Yup sounds good.

titu1994: On second thought lets keep it, incase future models require it.

PR #3222, opened 2021-11-19
Adding adaptive bucketing for tarred datasets.

VahidooX:
This PR adds the support of adaptive batch size for bucketing.
+It currently just supports tarred datasets
+train_ds.bucketing_batch_size is added which user may use to enable the adaptive batching and specify the batch sizes for the buckets
+when bucketing_batch_size is not set, train_ds.batch_size is going to be used for all buckets (fixed-size bucketing)
+bucketing_batch_size can be set as a list to explicitly specify the batch size for each bucket 
+When bucketing_batch_size is set, traind_ds.batch_size need to be set to 1
+if bucketing_batch_size is set to be an integer, then linear scaling is being used to scale-up the batch sizes for batches with shorted audio size. For example, setting train_ds.bucketing_batch_size=8 for 4 buckets would use these sizes [32,24,16,8] for different buckets.
+Adaptive batching may improve the training speed per epoch more than 2x but may affect the final accuracy of the model


nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1066,13 +1066,14 @@ def __init__(
                 begin_idx = (len(audio_tar_filepaths) // world_size) * global_rank
                 end_idx = begin_idx + (len(audio_tar_filepaths) // world_size)
                 audio_tar_filepaths = audio_tar_filepaths[begin_idx:end_idx]
+                self.length = len(self.collection) // world_size
                 logging.info(
                     "Partitioning tarred dataset: process (%d) taking shards [%d, %d)", global_rank, begin_idx, end_idx
                 )
 
             elif shard_strategy == 'replicate':
+                self.length = len(self.collection) // world_size
```
titu1994, line 1075: In replication mode, full collection is available to all nodes. So why // world_size?

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1444,3 +1445,55 @@ def __call__(self, text):
             world_size=world_size,
             return_sample_id=return_sample_id,
         )
+
+
+class BucketingDataset(IterableDataset):
+    """
+    lodfehlkfeh fqpfjqfpopf fpqojf qpoj!!!
```
titu1994, line 1452: Fill up docstring before merge

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1444,3 +1445,55 @@ def __call__(self, text):
...
+        return _speech_collate_fn(batch[0], self.wrapped_dataset.pad_id)
+
+    def __iter__(self):
+        return BucketingIterator(
+            wrapped_iter=self.wrapped_dataset._dataset.__iter__(), bucketing_batch_size=self.bucketing_batch_size
+        ).__iter__()
+
+    def __len__(self):
+        # print(
+        #     f"dataset size={int(math.ceil(len(self.wrapped_dataset.collection)/float(self.bucketing_batch_size)))}, bucketing_batch_size={self.bucketing_batch_size}"
```
titu1994, line 1472: Remove print and add docstring

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1444,3 +1445,55 @@ def __call__(self, text):
...
+
+    def __next__(self):
+        batches = []
+        for idx in range(self.bucketing_batch_size):
+            try:
+                sample = next(self.wrapped_iter)
+            except StopIteration:
+                break
+            except Exception as e:
+                logging.error(e)
```
titu1994, line 1493: Why not raise the error instead of just log it? Below it will append None then into batches.append(..)

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1444,3 +1445,55 @@ def __call__(self, text):
...
+            batches.append(sample)
+        # print(f"idx={idx}, batch_len={len(batches)}, b_size={self.bucketing_batchsize}")
+        if len(batches) == 0:
+            raise StopIteration
+        # print(batches)
```
titu1994, line 1498: Remove prints before merge

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -218,6 +218,42 @@ def get_tarred_dataset(
...
         datasets.append(dataset)
 
+    if len(datasets) > 1:
+        if config.get('bucketing_batch_size', None) is not None and config['bucketing_batch_size'] > 0:
+            if config['batch_size'] != 1:
+                raise ValueError(
+                    f"batch_size should be set to one when bucketing_batch_size is set and adaptive bucketing is enabled (batch_size={config['batch_size']}!"
+                )
+
+            if type(config['bucketing_batch_size']) == int:
```
titu1994, line 228: Pull this logic which determines bucketing batch sizes into another method and call it here. Makes it easier to modularize in the future.

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -218,6 +218,42 @@ def get_tarred_dataset(
...
+                    bucketing_batch_sizes.append(scale_factor * config['bucketing_batch_size'])
+            elif type(config['bucketing_batch_size']):
+                bucketing_batch_sizes = config['bucketing_batch_size']
+            else:
+                raise ValueError(
+                    f"bucketing_batch_size should be an integer or a list (bucketing_batch_size={config['bucketing_batch_size']})!"
+                )
+
+            if len(bucketing_batch_sizes) != len(datasets):
+                logging.warning(
```
titu1994, line 241: Shouldnt this be an error instead?

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -218,6 +218,42 @@ def get_tarred_dataset(
...
+                    f"batch_size should have the same length as the number of buckets ({len(bucketing_batch_sizes)}!={len(datasets)}) "
+                )
+            logging.warning(
```
titu1994, line 244: Why is this a warning? It should be info

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -218,6 +218,42 @@ def get_tarred_dataset(
...
+                f"Batch bucketing is enabled for {len(datasets)} buckets with adaptive batch sizes of {bucketing_batch_sizes}!"
+            )
+        else:
+            logging.warning(
+                f"Batch bucketing is enabled for {len(datasets)} buckets with fixed batch size of {config['batch_size']}!"
```
titu1994, line 249: Should also be info

titu1994 (changes requested):
Requires some cleanup, but overall logic is correct.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: fixed.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: fixed.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: fixed.

nemo/collections/asr/data/audio_to_text.py
VahidooX: dropped.

nemo/collections/asr/data/audio_to_text.py
VahidooX: dropped it.

nemo/collections/asr/data/audio_to_text.py
VahidooX: dropped.

nemo/collections/asr/data/audio_to_text.py
VahidooX: added.

nemo/collections/asr/data/audio_to_text.py
VahidooX: fixed.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: moved it.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1444,3 +1445,52 @@ def __call__(self, text):
...
+        return self
+
+    def __next__(self):
+        batches = []
+        for idx in range(self.bucketing_batch_size):
+            try:
+                sample = next(self.wrapped_iter)
+            except StopIteration:
+                break
+            batches.append(sample)
```
titu1994, line 1493: This would still be None since sample would cause the StopError instead. Handle this properly like reraise StopIteration right here rather than append None

titu1994 (changes requested):
The logic inside __next__ seems incorrect. Reraise stop iteration immediately instead of break

titu1994 (dismissed):
Break is there, seems fine then. LGTM

titu1994 (approved):
Thanks !

PR #3239, opened 2021-11-24
Fix cast type in _se_pool_step_script related functions

Oktai15:
Signed-off-by: Oktai Tatanov <oktai.tatanov@gmail.com>


titu1994 (approved):
Isn't this the same as .type(x.type) though ?

PR #3259, opened 2021-11-29
(1) O2-style mixed precision recipe, (2) Persistent layer-norm, (3) Grade scale hysteresis, (4) gradient_as_bucket_view

erhoo82:
This PR container has many updates.
Please review and suggest any feedback on (1) code structure change, (2) argument names and structure.


examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -1,5 +1,6 @@
 name: megatron_gpt
 restore_from_path: null # used when starting from a .nemo file
+gradient_as_bucket_view: True
```
ericharper, line 3: Why is this at the highest level of the config? Should it be in `model` ?

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -114,3 +117,9 @@ model:
       warmup_steps: 500
       constant_steps: 50000
       min_lr: 2e-5
+
+    # Megatron O2-style half-precision
+    megatron_amp_o2: False # Enable O2-level automatic mixed precision using master parameters
```
ericharper, line 122: A small nitpick but should this be named `megatron_amp_O2` ?

nemo/collections/nlp/modules/common/megatron/module.py
```diff
@@ -90,3 +98,82 @@ def initialize_word_embeddings(self, init_method, vocab_size, hidden_size, pipel
...
+        if precision == 16:
+            self.add_module('module', module.half())
+            def float16_convertor(val):
+                return val.half()
+        elif precision == 'bf16':
+            self.add_module('module', module.bfloat16())
+            def float16_convertor(val):
+                return val.bfloat16()
+        else:
+            raise Exception('should not be here')
```
ericharper, line 152: Maybe we can make this exception more informative?

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -308,14 +333,145 @@ def update(self, new_scale=None):
...
+        self._init_growth_tracker = state_dict["_growth_tracker"]
+        if self._growth_tracker is not None:
+            self._growth_tracker.fill_(state_dict["_growth_tracker"])
+        self._hysteresis_tracker = state_dict["_hysteresis_tracker"]
+
+
+class MegatronHalfPrecisionPlugin(NativeMixedPrecisionPlugin):
+    """
+    Plugin for Half (FP16 and BF16) precision training.
+    This plugin assumes the uase of the optimizer with master parameters (fp32).
```
ericharper, line 406: typo: uase -> use

nemo/core/classes/modelPT.py
```diff
@@ -555,8 +557,22 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
...
             optimizer = optim.get_optimizer(optimizer_name)
             optimizer = optimizer(self.parameters(), **optimizer_args)
+            if megatron_amp_o2:
+                # Wrap the baseline optimizer with an optimizer class with master parameters
+                optimizer = MasterOptimizerWrapper(
+                    optimizer,
+                    fp32_grad_accum=fp32_grad_accum,
+                    contiguous_grad_bucket=contiguous_grad_bucket,
+                    async_grad_allreduce=async_grad_allreduce,
+                )
```
ericharper, line 575: I'm wondering if we can make this less invasive. Can we create a hook that will let NeMo models customize `optimizer`. I did something similar here: https://github.com/NVIDIA/NeMo/commit/d5b2011df6426efd8ee874fa5b299c794f17b02f

ericharper, line 575: I'm wondering if we can make this less invasive. Can we create a hook that will let NeMo models customize `optimizer`. I did something similar here: https://github.com/NVIDIA/NeMo/commit/d5b2011df6426efd8ee874fa5b299c794f17b02f

nemo/core/optim/optimizer_with_master_params.py
```diff
@@ -0,0 +1,399 @@
...
+
+from contextlib import contextmanager
+
+import torch
+
+from apex.multi_tensor_apply import multi_tensor_applier
+import amp_C
+
+from apex.transformer.parallel_state import get_data_parallel_world_size, get_data_parallel_group
+from apex.transformer.tensor_parallel import copy_tensor_model_parallel_attributes
```
ericharper, line 27: Can we add apex import guards here? Similar to here; https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/parts/nlp_overrides.py#L32

nemo/core/optim/optimizer_with_master_params.py
```diff
@@ -0,0 +1,399 @@
+# coding=utf-8
```
ericharper, line 1: Can you remove this line?

nemo/core/optim/optimizer_with_master_params.py
```diff
@@ -0,0 +1,399 @@
...
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
ericharper, line 2: 2021

ericharper (changes requested):
Mostly minor changes requested. Main concern I have is with ModelPT changes. Can we refactor it so that optimizer is customized via hook?

Also, just a question, will this O2 style recipe eventually be something that is used in NeMo rather than implemented in NeMo?

okuchaiev:
@titu1994 this is potentially a more memory efficient recipe - could you please have a look

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
titu1994: +1

nemo/core/classes/modelPT.py
titu1994: This should not be part of ModelPT really. It's not general to even a domain, but to exactly one model - Megatron. The hooks here go all over the Megatron codebase to wrap modules to force cast to fp16 etc which makes this unusable in every other domain unless it supports such operations. 

As Eric said, please move this to some form of pre/post callback function that is called inside configure_optimizers() after finishing setup_optimization(). Then NLP model can override those specific methods there. Of course, these arguments should no longer be part of model.optim either - they aren't for any optimizer but specific to supported optimizer that Megatron uses.

titu1994 (changes requested):

titu1994:
@okuchaiev we can try it, but I'm pretty sure it won't work for anything other than Megatron. The model has hooks linking these O2 instructions all the way into it's individual modules which get wrapped with FloatModule which explicitly casts inputs and outputs to floats. 

I doubt even normal non megatron NLP models can use this O2 safely without facing nan gradients

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
erhoo82: Done

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
erhoo82: Technically O2 should be right. But, I see that only lower case is used in the config? @ericharper  Do you suggest using O2 instead of o2?

nemo/collections/nlp/parts/nlp_overrides.py
erhoo82: Corrected the typo seems like they were there in Megatron repo.

nemo/core/classes/modelPT.py
erhoo82: Thanks for the feecback. I will change the structure as suggested.

nemo/core/optim/optimizer_with_master_params.py
erhoo82: Done

nemo/core/optim/optimizer_with_master_params.py
erhoo82: Done

nemo/core/optim/optimizer_with_master_params.py
erhoo82: Done

erhoo82:
> Mostly minor changes requested. Main concern I have is with ModelPT changes. Can we refactor it so that optimizer is customized via hook?
> 
> Also, just a question, will this O2 style recipe eventually be something that is used in NeMo rather than implemented in NeMo?

FP16-O2 and BF16-O2 uses different approacehs.

erhoo82:
https://github.com/NVIDIA/NeMo/pull/3259#issuecomment-983285415
@okuchaiev for viz as well.

erhoo82:
Working on debugging.
Switched to WIP temporarly.

erhoo82:
I moved the master optimizer wrapper to the Megatron model and fixed a state_dict maching.

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -69,9 +71,15 @@ model:
   # precision
   native_amp_init_scale: 4294967296 # 2 ** 32
   native_amp_growth_interval: 1000
+  hysteresis: 2 # Gradient scale hysteresis
   fp32_residual_connection: False # Move residual connections to fp32
   fp16_lm_cross_entropy: False # Move the cross entropy unreduced loss calculation for lm head to fp16
 
+  # Megatron O2-style half-precision
+  megatron_amp_o2: False # Enable O2-level automatic mixed precision using master parameters
```
ericharper, line 79: Small thing but can we change this to `megatron_amp_O2`?

nemo/core/classes/modelPT.py
```diff
@@ -407,7 +407,7 @@ def setup_multiple_test_data(self, test_data_config: Union[DictConfig, Dict]):
             if self._test_dl is not None and type(self._test_dl) in [list, tuple]:
                 self._test_names = ['test_{}_'.format(idx) for idx in range(len(self._test_dl))]
 
-    def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = None):
+    def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = None, delayed_sched_init=False):
```
ericharper, line 410: Can we remove this: `delayed_sched_init`?

ericharper (changes requested):
Really nice contribution!

Everything is looking good except, I'd like us to try to remove the `delayed_sched_init` arg from `setup_optimization`

ericharper:
More specifically, if we were to remove `delayed_sched_init`. Can't we just overwrite `self._scheduler` in your `configure_optimizers` override?

nemo/core/classes/modelPT.py
erhoo82: I can do that. Then, I guess then we are just re-do the LR schedule init when using master optimizer wrapper.
I guess that should be fine(?).

markelsanz14:
I believe this PR modifies the state_dict of the saved checkpoints, which can break checkpoint loading, the conversion script to convert to .nemo format, and inference pipelines. We need to change this.

erhoo82:
> 

Yes, this works. Let me change the code.

erhoo82:
> I believe this PR modifies the state_dict of the saved checkpoints, which can break checkpoint loading, the conversion script to convert to .nemo format, and inference pipelines. We need to change this.

About this, I asked @ericharper to take a look at how to by-pass the half_precision module in the model naming. I originally didn't include the name in the state_dict when saving it. But, when loading it PYT's load_state_dict methods doesn't allow bypassing it.

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
erhoo82: Done

nemo/core/classes/modelPT.py
erhoo82: Done

ericharper:
PR looks good. I'm just going to leave as change requested until I can investigate this state_dict issue further.

titu1994 (approved):
Looks great. Thanks for moving the code into MasterOptimizerWrapper

ericharper (dismissed):
LGTM! Thanks!

PR #3265, opened 2021-11-30
Rewrite megatron_gpt_ckpt_to_nemo.py

yaoyu-33:
Signed-off-by: Yu Yao <yuya@nvidia.com>


examples/nlp/language_modeling/megatron_gpt_ckpt_to_nemo.py
```diff
@@ -52,35 +51,74 @@ def get_args():
...
+
+    print("****** Conf: ", conf)
+    print("****** Checkpoints processing time: ", time.time() - start_time)
+
+    tar_start_time = time.time()
+    with open(config_yaml, 'w') as fout:
+        OmegaConf.save(config=conf, f=fout, resolve=True)
+
+    print("****** Calling tar...")
+    subprocess.call(f'tar -czvf {args.nemo_file_path} ./*', cwd=tmp_dir, shell=True)
```
okuchaiev, line 115: please use model.save_to(...) API to create .nemo files. Otherwise we are at risk of producing incompatible .nemo files at some point in time.

okuchaiev (changes requested):
1) Can you please refactor to use torch.distributed.barrier instead of torch.multiprocessing when instantiating a model from .ckpt 
2) Then  use .save_to to produce .nemo file?

examples/nlp/language_modeling/megatron_gpt_ckpt_to_nemo.py
yaoyu-33: Hey Oleksii, I agree with you. Its safer to call apis. However, I am not familiar with the apis. I thought Eric discussed with Markel. The reason I filed this PR is that Markel told me we could use this simple converter at the moment. If we need to refactor the original code, use barrier or whatever I would probably hand it back to @ericharper.

ericharper:
Closed in favor of #3277. Will be adding #3277 changes to main

PR #3272, opened 2021-11-30
Initial refactor

borisfom:


titu1994 (approved):
Fantastic ! Didnt even think of the idea to create a wrapper fused module with the decoder-joint.

PR #3276, opened 2021-12-01
batch implementation

dimapihtar:
Signed-off-by: Dmytro Pykhtar <dpykhtar@nvidia.com>


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -314,78 +314,86 @@ def configure_gradient_clipping(self, *args, **kwargs):
         parameters = self.model.parameters()
         clip_grad_norm_fp32(parameters=parameters, max_norm=clip_val)
 
+    def buckets(self, batch):
```
okuchaiev, line 317: it looks like this should be a class method not an instance method. Or maybe it should go to https://github.com/NVIDIA/NeMo/tree/main/nemo/collections/nlp/data/language_modeling/megatron/....  @ericharper  what do you think?

okuchaiev: Also rename it to something like "bucketize_for_gpt_inference"

nemo/collections/nlp/modules/common/megatron/fused_bias_gelu.py
```diff
@@ -25,7 +25,7 @@
 # x * 0.5 * (1.0 + torch.erf(x * 0.70710678))
 
 
-@torch.jit.script
+# @torch.jit.script
```
okuchaiev, line 28: please uncomment this.

okuchaiev (changes requested):
few small changes - please see inline

okuchaiev (changes requested):
there is some weird binary file as part of this PR nemo/collections/nlp/models/language_modeling/.megatron_gpt_model.py.swp --- please remove it

okuchaiev (approved):
lgtm. thanks

PR #3278, opened 2021-12-01
Fix megatron_gpt_ckpt_to_nemo.py with torch distributed 

yaoyu-33:
Fixed megatron_gpt_ckpt_to_nemo.py with torch distributed along with a minor fix in loading from checkpoint in nlp_model.py


nemo/collections/nlp/models/nlp_model.py
```diff
@@ -250,7 +250,12 @@ def load_from_checkpoint(
                 config_kwargs.pop('trainer')
             checkpoint[cls.CHECKPOINT_HYPER_PARAMS_KEY].update(config_kwargs)
 
-            model = cls._load_model_state(checkpoint, strict=strict, **kwargs)
+            if 'cfg' in kwargs:
+                model = cls._load_model_state(checkpoint, strict=strict, **kwargs)
```
titu1994, line 254: This check shouldn't be required but for now it's ok to keep it until we have a better solution.

titu1994 (approved):
LGTM but we need to clean up the restore portion to not access PTL variables. Bound to break in the future c

nemo/collections/nlp/models/nlp_model.py
yaoyu-33: yeah, I just add this to be safe.

PR #3286, opened 2021-12-04
Reduce test time of punctuation and capitalization model

PeganovAnton:
Signed-off-by: PeganovAnton <peganoff2@mail.ru>


Jenkinsfile
```diff
@@ -1141,25 +1141,6 @@ pipeline {
               trainer.strategy=ddp \
               trainer.max_epochs=1 \
               +exp_manager.explicit_log_dir=/home/TestData/nlp/token_classification_punctuation/output && \
```
ekmb, line 1143: If you don't need output, could you use `exp_manager=null` to reduce test time even more?

ekmb (approved):

PR #3293, opened 2021-12-07
Megatron AMP fix for scheduler step counter

titu1994:
# Bugfix
- Attempt to prevent a step counter update of the scheduler when skipping a step in the optimizer
- Add training steps when nan/inf gradients are detected
- Properly support LR scheduler with "constant lr" at end of training. Refactor CosineAnnealing model for such support.

# Details

Since there are no pre-existing hooks in PTL to modify the scheduler step before the optimizer step, ive had to splice the support into the current GradScaler and then use the on_train_batch_end hook and manually monitor when the optimizer step occurs via the flag  optimizer_update_skipped - None means optimizer didnt step this time (due to accumulated gradient), and True means grads were accumulated but optimizer did not perform step() - skipped step. In such case, perform a no-op for the scheduler too.

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -182,7 +182,7 @@ class MegatronGPTModel(NLPModel):
 
                 # If the grad scaler skipped its optimizer step due to infs/nans,
                 # decrement the step of all schedulers.
-                if grad_scaler.optimizer_update_skipped:
+                if grad_scaler.optimizer_update_skipped is not None and grad_scaler.optimizer_update_skipped is True:
                     schedulers = self.trainer.lr_schedulers
 
                     if not schedulers or not self.trainer.lightning_module.automatic_optimization:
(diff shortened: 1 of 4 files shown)


ericharper (commented):
LGTM. @erhoo82, can you verify if we get the same behavior as megatron-lm?

erhoo82:
The PR looks good to me. Just one question, what's the key reason for decrementing the LR scheduler by two steps then executing one?

titu1994:
So i could manually reduce by 1, but then the actual optimizer param_groups['lr'] does not update automatically. So I decrement by two, then manually call scheduler.step() again, so that it resets back to original lr inside of the optimizer param_groups['lr']. 

Decrement by two is because the training_step itself calls scheduler.step() once automatically before calling on_train_batch_end(). 

So in essence : 
After training_step, before on_train_batch_end(): 
forward_model() -> optimizer.step() [checks for nan/inf and skips update, sets the value for check] -> scheduler.step() [unavoidable until PTL gives a hook here]

Inside on_train_batch_end:
Check if nan/inf occurred, revert scheduler index by 2, recall scheduler.step() [updates sched + optim with the value before the scheduler step() that occurred in between training_step() and on_train_batch_end().

erhoo82:
@titu1994 @ericharper

ericharper (approved):
LGTM. Thanks!

PR #3297, opened 2021-12-08
Wav2vec

tbartley94:
Updated Wav2Vec module in asr collections to integrate with NeMo. Can now be passed usual config file and will process.


yzhang123 (changes requested):

tbartley94:
@titu1994 @sam1373 Would y'all mind reviewing if you have a moment?

sam1373 (changes requested):

nemo/collections/asr/models/wav2vec_models.py
```diff
@@ -0,0 +1,135 @@
...
+import torch
+from omegaconf import DictConfig
+from pytorch_lightning import Trainer
+from torch import nn
+
+from nemo.collections.asr.models.ssl_models import SpeechEncDecSelfSupervisedModel
+from nemo.core.classes.common import PretrainedModelInfo, typecheck
+
+
+class Wav2VecEncoderModel(SpeechEncDecSelfSupervisedModel):
```
sam1373, line 31: Do we actually need this class? Maybe we could just use the parent class with additional optional features added to it, and/or to contrastive loss?

tbartley94: Good point. Any reason we don't keep a simple linear decoder lying around though?

tbartley94: It depends, the main differences are that the parent class detatches spectrograms, the child class has some extra dropouts here and there, and there's an additional penalty added to the loss. It's easy to add but would be only for the sake of only one model.

nemo/collections/asr/modules/wav2vec_modules.py
```diff
@@ -17,290 +17,403 @@
...
-    padding_mask: Optional[torch.Tensor],
-    mask_prob: float,
-    mask_length: int,
-    mask_type: Wav2VecMaskType = Wav2VecMaskType.static,
-    mask_other: float = 0.0,
-    min_masks: int = 0,
-    no_overlap: bool = False,
-    min_space: int = 0,
-) -> np.ndarray:
+class Wav2VecLinearDecoder(NeuralModule, Exportable):
```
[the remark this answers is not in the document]
tbartley94: Fair point. Though, any reason we don't just keep a simple linear decoder lying around?

nemo/collections/asr/models/wav2vec_models.py
sam1373: I think these are minor enough to be options in that class, and we will probably experiment with them for other models later on anyway.

tbartley94: Alrighty, I'll add them in

sam1373 (approved):

yzhang123 (approved):

nemo/collections/asr/modules/wav2vec_modules.py
```diff
@@ -17,290 +17,312 @@
...
+        return audio_lengths
 
-    def set_num_updates(self, num_updates):
-        self.curr_temp = max(self.max_temp * self.temp_decay ** num_updates, self.min_temp)
 
-    def get_codebook_indices(self):
-        if self.codebook_indices is None:
-            from itertools import product
+class Wav2VecTransformerEncoder(TransformerEncoder):
+    def __init__(self, pos_embed: DictConfig, transformer: DictConfig, layer_drop: float = 0.0):
```
titu1994, line 215: Document what are valid values for `transformer` dictconfig here. Similar to jasper block

nemo/collections/asr/parts/submodules/ssl_quantizers.py
```diff
@@ -0,0 +1,185 @@
...
+    def __init__(
+        self,
+        dim,
+        num_vars,
+        temp,
+        groups,
+        combine_groups,
+        vq_dim,
+        time_first,
+        activation=nn.GELU(),
```
titu1994, line 38: take str here and resolve from dictionary for activation

titu1994 (commented):
Minor comments here and there, but the code looks overall correct to me. I leave final review / approval to @sam1373 

One question is - is this model generated here compatible In weights to the Wav2Vec checkpoints released by FairSeq? If so, is it possible to have a weight porting mechanism from their checkpoints to a nemo one? If not, why is it incompatible, and does it affect training?

tbartley94:
@titu1994 

> 



> One question is - is this model generated here compatible In weights to the Wav2Vec checkpoints released by FairSeq? 

I'm doing a few pretraining runs so we can have weights for comparison but from what I've gathered from their reported numbers from finetuning, it does look within the ballpark. For comparison, their reported numbers from finetuning the pretrained model for LS were:

dev-clean 3.2 
dev-other 8.9 
test-clean 3.4 
test-other 8.5

The preliminary results from CTC training (w/o pretraining, and with reduced dropout, no weight decay) are: 
dev-clean: 3.98
dev-other: 11.14
test-clean: 4.16
test-other: 11.51

I still need to fiddle with how to mirror their masking with SpecAug, and am comparing the normalization schemes they report. But the difference seems to suggests the weights are close (especially considering their results had benefit of pretraining). 

> If so, is it possible to have a weight porting mechanism from their checkpoints to a nemo one? If not, why is it incompatible, and does it affect training?

I'll double check the Fairseq codebase again, but to my knowledge: mostly.

There are some small differences to make it easier to integrate with NeMo, but in general: should be able to get most of the weights transferred. The convolutions for feature embedding are simple enough so those weights should transfer. Their attention layers are a tad different so may not port over to ours as easily. Beyond that, there's only a couple of minor transformations that would just be purely incompatible (e.g. learned parameters for normalization layers).  

The major concern that I can imagine (and not even quite sure how significant this would be) is that we vary on the masking size. They use blocks of 10 time steps, we use single block time steps. It's the same learned task but I suppose there's an argument to be made that predicting a larger block of audio data requires different weighting than smaller blocks. But that's just a guess, and it would probably be fixed by finetuning on the NeMo ASR set anyhow.

titu1994:
> Beyond that, there's only a couple of minor transformations that would just be purely incompatible (e.g. learned parameters for normalization layers).

Why would normalization layers be incompatible ? They possibly use layer norm just as us no ?

> The major concern that I can imagine (and not even quite sure how significant this would be) is that we vary on the masking size. They use blocks of 10 time steps, we use single block time steps.

Could you elaborate ? We do have fixed timestep masking - use a int for time_width instead of a float. Freq_width is also an int which is the maximum contiguously freq band that is masked out.

titu1994:
For the group norm vs layer norm, lets default to always on layer norm. We will offer flags to disable the first norm or set it to group to port weights but defaults will be the recommended version. When porting, the Nemo copy just replicate the FairSeq model exactly or conversion should stop and error out. 

We can leave combine timesteps for future - as long as we can port their large or base model (preferably with combine = 1) it's fine. The effect of >1 is probably not significant enough so we can reduce it's priority

tbartley94:
> For the group norm vs layer norm, lets default to always on layer norm. We will offer flags to disable the first norm or set it to group to port weights but defaults will be the recommended version. When porting, the Nemo copy just replicate the FairSeq model exactly or conversion should stop and error out.
> 
Currently it involves a couple of flags to toggle the different parts of the normalization scheme. (one for normalizing audio, one to toggle layer or group norm, one for an l2 penalty). Would it be better to combine all that into a single flag for ease or leave them as individual flags for experimentation?

titu1994:
Individual flags, with defaults to have the normalization on.

sam1373 (approved):

sam1373 (approved):

yzhang123 (commented):
thx

PR #3303, opened 2021-12-09
Adding  Megatron NeMo Bert support

yidong72:
I tested with small dataset with binary_head True/False and model_parallel 2 and 1. They all seem working fine.


nemo/collections/nlp/data/language_modeling/megatron/bert_dataset.py
```diff
@@ -29,17 +29,17 @@
...
+        indexed_dataset=None,
+        data_prefix=None,
+        num_epochs=None,
+        max_num_samples=None,
+        masked_lm_prob=None,
+        max_seq_length=None,
+        short_seq_prob=None,
+        seed=None,
+        binary_head=None,
+        tokenizer=None,
```
ericharper, line 42: Why set these to None? Are they all optional? Could you also add type hints?

ericharper (changes requested):
Overall LGTM. Could you add a CI test for megatron bert training? We have one for megatron gpt in Jenkinsfile. You can use that as an example. Make sure the model is small so the test runs fast.

nemo/collections/nlp/data/language_modeling/megatron/bert_dataset.py
yidong72: I added the type hints. I added "=None" to convert them from positional arguments to keyword arguments.

yidong72:
> Overall LGTM. Could you add a CI test for megatron bert training? We have one for megatron gpt in Jenkinsfile. You can use that as an example. Make sure the model is small so the test runs fast.

I added the CI test into the Jenkinsfile for Megatron Bert model

ericharper (approved):
LGTM. Thanks!

PR #3305, opened 2021-12-09
Refactored prepare_for_export calls to ensure input size of example i…

borisfom:
…s handled properly

See also https://jirasw.nvidia.com/browse/JARS-1682


nemo/collections/asr/modules/rnnt.py
```diff
@@ -162,22 +162,14 @@ def __init__(
...
 
         # state maintenance is unnecessary during training forward call
         # to get state, use .predict() method.
-        if self._rnnt_export:
-            add_sos = False
-        else:
-            add_sos = True
-
-        g, states = self.predict(y, state=states, add_sos=add_sos)  # (B, U, D)
+        g, states = self.predict(y, state=states, add_sos=self.training)  # (B, U, D)
```
titu1994, line 172: This is incorrect - even during eval mode, we need an additional sos token in the forward. Only during export it should be one token (whatever it may be).

titu1994 (changes requested):
Overall looks ok - but the rnn prediction network forward change is incorrect. Once that's reverted, should be good to merge

nemo/collections/asr/modules/rnnt.py
borisfom: Oh I got confused by the comment above. Will put that back. They really should have had it tri-state (train/eval/infer) rather than boolean.

titu1994 (approved):
LGTM

VahidooX (approved):

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -76,7 +76,6 @@ def _prepare_for_export(self, **kwargs):
 
             if isinstance(m, SqueezeExcite):
                 m.set_max_len(int(one_hour // stride))  # One hour divided by current stride level
-                m.forward = m.forward_for_export
```
titu1994, line 79: With the recent fix for DDP, will this line change cause problems ?

titu1994 (approved):
Overall looks good.

PR #3311, opened 2021-12-10
Replacing outdated exports scripts

borisfom:
https://jirasw.nvidia.com/browse/AIA-2166

Also: 
https://jirasw.nvidia.com/browse/AIA-2168


titu1994 (dismissed):
Very nice ! LGTM. There are some unused imports which could be removed

titu1994 (approved):
LGTM

aklife97:
@borisfom I see that MTEncDecModel's export exports the encoder and decoder only. We probably need to export the classifier as well (named self.log_softmax), right?

PR #3314, opened 2021-12-10
Fixed future masking in cross attention of Perceiver

michalivne:
This PR fixed auto-regressive future masking of cross attention in Perceiver encoder.
**NOTE:** Default is kept auto-regressive since it shows better results.

This PR also updated `yaml` config file to default to `aayn_base.yaml` instead of copying the entries.


okuchaiev (approved):
lgtm perciver changes. but why so much stuff removed from examples/nlp/machine_translation/conf/aayn_bottleneck.yaml ?

PR #3317, opened 2021-12-11
Asr fr

tbartley94:
Uploaded new version of Conformer CTC to NGC (now with unhyphenated token set). Updating ctc_bpe_models to reflect new imports.


titu1994 (approved):
Can you also add these to the Nemo docs - we have per language model checkpoints listed there, please update the one for fr.

PR #3334, opened 2021-12-15
Add uniform ASR evaluation script for all models

titu1994:
# Changelog
- Add speech_to_text_eval.py which utilizes `transcribe_speech.py` to first perform transcription (for all asr models) and then uses its output to compute scores against some ground truth (via manifest).
- Remove speech_to_text_infer.py since it was hand tuned for just 1 type of model and was not an inference script but an evaluation script that required ground truth.
- Improve docstrings for transcribe_speech.py and transcribe_speech_parallel.py


titu1994:
FYI @vsl9

VahidooX (approved):

Issue #3220, opened 2021-11-19
Machine translation shared embeddings


1-800-BAD-CODE:
**Describe the bug**

If my understanding is correct, an [MTEncDecModel](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py#L54) exposes no options for sharing embeddings between the encoder and decoder, but this seems to be the de facto standard. E.g. from the AAYN white paper:

> ...we share the same weight matrix between the two embedding layers and the pre-softmax linear transformation

So MT models with shared vocab will learn separate embeddings for each token on both sides.

**Expected behavior**

An option like `cfg.share_embeddings = True` to tie the encoder/decoder embedding weights.

PR #3340, opened 2021-12-15
NMT Shared Embeddings Weights

michalivne:
This PR closes https://github.com/NVIDIA/NeMo/issues/3220 by adding support to share encoder and decoder embeddings and pre-softmax projection layers when the same vocabulary is used in the encoder and decoder.

**NOTE:** Only the shape of the embedding matrix is validated during runtime (i.e., vocabulary size, vocabulary dimension). It is the user's responsibility to make sure the same vocabulary is being used in the encoder and decoder tokenizers.

# Usage:

Set the following flag in `yaml` config to `true` (default is `false`):

```
model:
  shared_embeddings: true
```


nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -193,6 +194,21 @@ def __init__(self, cfg: MTEncDecModelConfig, trainer: Trainer = None):
             max_delta_length=cfg.max_generation_delta,
         )
 
+        # tie embedding weights
+        if self.shared_embeddings:
+            # validate vocabulary size and embedding dimension
+            if (
+                self.encoder.embedding.token_embedding.weight.shape
```
aklife97, line 201: Maybe better to also check the encoder and decoder use the same tokenizer (i.e. the `shared_tokenizer` is set)?

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -193,6 +194,21 @@ def __init__(self, cfg: MTEncDecModelConfig, trainer: Trainer = None):
...
+                != self.decoder.embedding.token_embedding.weight.shape
+            ):
+                raise ValueError(
+                    f"Cannot tie encoder and decoder embeddings due to mismatch in embedding sizes "
+                    f"(num_embeddings, embedding_dim): {self.encoder.embedding.token_embedding.weight.shape} (encoder) "
+                    f"{self.decoder.embedding.token_embedding.weight.shape} (decoder)"
+                )
+
+            self.encoder.embedding.token_embedding.weight = self.decoder.embedding.token_embedding.weight
```
aklife97, line 210: This seems to only tie token_embeddings, what happens in case `learn_positional_encodings` is set?

aklife97 (commented):
Thanks for the PR! Looks good, just a few comments

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
michalivne: Do we have a reliable method to compare tokenizers?
I was considering it but I am not not sure the "==" operator is working as we would like (i.e., comparing the class, all flags/options, and all tokens).
Even if we can compare the class correctly, not all classes allow comparison of the tokens.
Instead, I opted to leave it to the user to make sure the same tokenizers are being used.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
michalivne: Good point. Let me check that.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: Oh, I just meant checking if the `shared_tokenizer` param in the config is set

michalivne: Isn't `shared_tokenizer` only affect learning a new tokenizer? I never changed the value of this parameter (defaults to true), even when using independent tokenizers.

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: Cool

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
aklife97: I guess its a fair assumption that all params are set correctly. In the worst case, the user can correct the param if they see an error of shared_tokenizers.
But I don't mind this staying as is either.

michalivne: Added an error if shared_tokenizer is False but shared_embeddings is True.

aklife97 (approved):
LGTM, thanks!

PR #3354, opened 2021-12-16
fix speaker label models training convergence

nithinraok:


VahidooX (approved):

titu1994 (commented):
Just so you're informed, without index by file I'd set, tarred datasets will no longer work.

PR #3361, opened 2021-12-18
Moved shebangs to the first line

davidalami:
The shebang must be the first line because it is interpreted by the kernel, which looks at the two bytes at the start of an executable file.
@okuchaiev , could you review please?


okuchaiev (approved):
Thanks! Looks good to me.

PR #3368, opened 2021-12-21
Byte-level Multilingual NMT

aklife97:
Add support for Byte-level multilingual NMT + Byte-level preproc fixes.

Other change to note:
* `train_tokenizers::preproc_mt_data` Instead of throwing an error if the tokenizer is not trainable, train_tokenizer now returns None. This is done to keep the pipeline simple and same for tokenizers like byte-level which do not require any explicit training.


michalivne (dismissed):
LGTM! Thanks a lot for the fix!

nemo/collections/common/tokenizers/bytelevel_tokenizers.py
```diff
@@ -38,9 +36,20 @@ def normalize(self, text) -> str:
...
         self.vocab_size = 259
-        self.special_tokens = [self.bos_id, self.eos_id, self.pad_id]
+        self.special_start = 256
+        self.special_token_to_id = {
+            self.pad_id: self.pad_id,
+            self.bos_id: self.bos_id,
+            self.eos_id: self.eos_id,
+        }
+        special_tokens = {} if special_tokens is None else special_tokens
+        for tok in special_tokens:
```
michalivne, line 48: How can special tokens be either dictionary or list?

nemo/collections/common/tokenizers/bytelevel_tokenizers.py
```diff
@@ -38,9 +36,20 @@ def normalize(self, text) -> str:
...
+            self.special_start -= 1
```
michalivne, line 49: Why do we count backwards here? Shouldn't it be forward?

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -756,63 +765,51 @@ def train_tokenizers(
...
+                            os.system('cat %s %s > %s' % (src_fname, tgt_fname, concat_data_path))
+                            if encoder_tokenizer_name == "yttm":
+                                yttm.BPE.train(
+                                    data=concat_data_path,
+                                    vocab_size=encoder_tokenizer_vocab_size,
+                                    model=os.path.join(out_dir, encoder_tokenizer_model),
+                                    coverage=encoder_tokenizer_coverage,
+                                    n_threads=-1,
+                                )
+                            else:
```
michalivne, line 794: Do we have only yttm and sentencepiece supported here?
(why not if yttp: elif spt: else: raise Error ?

nemo/collections/common/tokenizers/bytelevel_tokenizers.py
aklife97: Its a standard thing across tokenizerspec tokenizers that support special tokens (like sentencepiece). Need to support both dict and list to keep usage similar. In case it is dict, the expectation specifically for byte-level is that both key and value will be the same.

nemo/collections/common/tokenizers/bytelevel_tokenizers.py
aklife97: This is done to akin to how ByT5 handles sentinels. So we don't increase the vocab, but resue starting from the last

michalivne: I could not find it in the paper. Is that how they implemented it?
Wouldn't that mean that some input bytes will be treated as a special character?

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
aklife97: This is required to support some of our other scripts like the one to create tarred dataset. Changing other scripts is a larger workload, changing this function's behavior to return None when things can't be trained handles everything without changing other scripts

michalivne: Ok let's keep it as it was before.

nemo/collections/common/tokenizers/bytelevel_tokenizers.py
michalivne: ok let's keep it as the paper, for sake of compatibility.

michalivne (approved):

PR #3374, opened 2021-12-24
[fix] for resume training on SLURM multi-node multi-gpu

itzsimpl:
On SLURM resuming training in a multi-node multi-gpu settings fails, as when `LOCAL_RANK` is undefined `is_globa_rank_zero()` returns true on all processes that run on node 0. In this case `exp_manager.py` https://github.com/NVIDIA/NeMo/blob/f83b2c5524a787be21ffea170850c4b5486eac2b/nemo/utils/exp_manager.py#L446, creates multiple `run_*` folders, and eventually leads to failure (missing files because other processes have moved them already). 

Checking also for `SLURM_PROCID` solves this issue, as the environment variable contains the global rank id.


okuchaiev:
thank you for the PR. Can you please sign your commits (see DCO's bot output https://github.com/NVIDIA/NeMo/pull/3374/checks?check_run_id=4628229084 ) and make sure your code is formatted properly (see https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md on how to autoformat)

nemo/utils/get_rank.py
```diff
@@ -24,25 +24,10 @@ def is_global_rank_zero():
...
-                node_rank = group_rank
-            # Take from NODE_RANK whenever available
-            if node_rank_env:
-                node_rank = node_rank_env
-        return node_rank
-
-    node_rank = get_node_rank()
-    local_rank = get_envint("LOCAL_RANK", 0)
+    # on SLURM variables SLURM_NODEID and SLURM_PROCID will be defined
+    # SLURM_PROCID holds the global rank and computing the node_rank is
```
blisc, line 28: As stated in this comment and looking at lightning's SLURM checks (https://github.com/PyTorchLightning/pytorch-lightning/blob/46f718d2ba31fdbb8f2abbef03471fec66204b66/pytorch_lightning/plugins/environments/slurm_environment.py)
"SLURM_PROCID" holds the global_rank, whereas "SLURM_LOCALID" holds the local rank. Shouldn't localrank be "SLURM_LOCALID"?
I guess at the end of the day, this function checks for global_rank==0 in which case the logic is correct but the variable names are not.

blisc: Let's just change this section of code to be check if we are in a SLURM environment. If so just return get_envint("SLURM_PROCID") == 0 or in that line.

itzsimpl: OK.

blisc (dismissed):

blisc:
Thanks @itzsimpl!

Jesteinbe:
I know this is closed but I recently started using nemo to train asr models and i have not been able to successfully restart training from an existing checkpoint yet. I'm following https://github.com/NVIDIA/NeMo/blob/main/examples/asr/asr_ctc/speech_to_text_ctc_bpe.py and using the conf from https://github.com/NVIDIA/NeMo/blob/main/examples/asr/conf/conformer/conformer_ctc_bpe.yaml.

I stopped training at one point and then when I restarted it again I observed that multiple `run_*` directories were getting created and the files from the previous run weren't getting moved properly (every time i've tried to restart training different files get moved to different run_* directories). I tried this multiple times thinking it might be a slurm "hiccup" but the problem persisted. As for versions, I'm using slurm-wlm 19.05.5 (from `sinfo -V`), python 3.8.10, and nemo v 1.17.0. I'm not totally sure what the best way to debug is or if other people have recently seen similar behavior but wanted to mention that the prior fix might still have some issues.

If you have any recommendations on how to debug/address this, please let me know. Thanks!

PR #3385, opened 2022-01-03
NMT MIM mean variance fix

michalivne:
This PR fixes the mean and variance prediction to use independent weights.
In addition, the default value of encoder future masking is set to be false (i.e., joint prediction instead of auto-regressive).


MaximumEntropy (approved):
LGTM!

PR #3400, opened 2022-01-07
Fix linear annealing not annealing lr to min_lr

MaximumEntropy:
The `WarmupAnnealing` currently anneals the LR to 0 instead of the provided `min_lr` arg.


titu1994 (approved):
Lgtm

PR #3422, opened 2022-01-12
Working around Pytorch exporter issue with expand()

borisfom:
WAR changes to prevent issues described in https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3429305
"ONNX model exported with PyTorch container > 21.06 fails when data of different size is provided to dynamic axes"


titu1994 (commented):
This is a good enough workaround, but does it occur even on 21.12?

aklife97 (approved):
Seems to fix the issue, thanks!
@titu1994 this issue has consistently been present in containers >21.06. Even occurs on 21.12 with the current main code.

titu1994:
Got it, looks good to merge then. Ill merge since tests already passed.

PR #3425, opened 2022-01-12
Preprocess an entire folder of .json or .json.gz files into a single .bin and .idx file.

MaximumEntropy:


scripts/nlp_language_modeling/preprocess_data_for_megatron.py
```diff
@@ -18,9 +18,11 @@
 import gzip
 import json
 import multiprocessing
+import os
 import sys
 import time
 
+import ftfy
```
ericharper, line 25: Do we need to add this to requirements_nlp.txt?

scripts/nlp_language_modeling/preprocess_data_for_megatron.py
```diff
@@ -132,6 +136,12 @@ def get_args():
     group = parser.add_argument_group(title='runtime')
     group.add_argument('--workers', type=int, default=1, help='Number of worker processes to launch')
     group.add_argument('--log-interval', type=int, default=100, help='Interval between progress updates')
+    group.add_argument(
+        '--preproc-folder',
+        action='store_true',
+        help='If set, will preprocess all .json or .json.gz files into a single .bin and .idx file. Folder path provided via the --input arg',
```
ericharper, line 142: Can you add help text to the --input arg explaining that input can be a single file or folder?

scripts/nlp_language_modeling/preprocess_data_for_megatron.py
```diff
@@ -152,12 +162,19 @@ def get_args():
...
+        json_files = [
+            os.path.join(args.input, f) for f in files_in_folder if f.endswith('.json') or f.endswith('.json.gz')
+        ]
+        if len(json_files) == 0:
+            raise FileNotFoundError('No .json or .json.gz files found in folder.')
+        else:
+            print(f'Found {len(json_files)} .json or .json.gz files.')
     else:
-        fin = open(args.input, 'r', encoding='utf-8')
+        json_files = [args.input]
```
ericharper, line 177: might as well check the path exists here?

ericharper (changes requested):
Great idea! Just a few small comments.

scripts/nlp_language_modeling/preprocess_data_for_megatron.py
MaximumEntropy: Done!

scripts/nlp_language_modeling/preprocess_data_for_megatron.py
MaximumEntropy: Good point! Done.

scripts/nlp_language_modeling/preprocess_data_for_megatron.py
MaximumEntropy: Added a check here as well. Good catch.

ericharper (approved):
LGTM. Thanks!

PR #3427, opened 2022-01-12
[BigNLP] Make saving .nemo during on_train_end configurable

ericharper:


nemo/utils/exp_manager.py
```diff
@@ -678,6 +679,7 @@ def __init__(
     ):
         # Parse and store "extended" parameters: save_best model and postfix.
         self.always_save_nemo = always_save_nemo
+        self.save_nemo_on_train_end = save_nemo_on_train_end
```
blisc, line 682: Let's add a warning that if this is False. save_best_model has no effect

blisc: Actually never mind. It's used elsewhere

blisc (approved):

blisc (approved):

PR #3429, opened 2022-01-12
Extending input_example() to take max batch and dimension arguments

borisfom:
This is to implement https://jirasw.nvidia.com/browse/JARS-1682.
It is important to provide user ability to generate input example of particular dimensions - as certain networks would reconfigure themselves according to input dimensions - thus export needs to be performed with real life size example.


borisfom:
"max_dim" argument for existing Nemo networks is the maximum sequence length. 
May be slightly different for vision (2-dim size) etc.

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -80,15 +80,15 @@ def _prepare_for_export(self, **kwargs):
         Exportable._prepare_for_export(self, **kwargs)
         logging.warning(f"Turned off {m_count} masked convolutions")
 
-    def input_example(self):
+    def input_example(self, max_batch=1, max_dim=256):
```
titu1994, line 83: max_dim should be 8192

nemo/collections/asr/modules/rnnt.py
```diff
@@ -699,13 +701,13 @@ def _prepare_for_export(self, **kwargs):
         self.log_softmax = False
         super()._prepare_for_export(**kwargs)
 
-    def input_example(self):
+    def input_example(self, max_batch=1, max_dim=256):
```
titu1994, line 704: max_dim should be 8192

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1101,8 +1103,8 @@ def input_types(self):
 
         return mytypes
 
-    def input_example(self):
-        decoder_example = self.decoder.input_example()
+    def input_example(self, max_batch=1, max_dim=256):
```
titu1994, line 1106: 8192

titu1994 (commented):
Overall seems fine, but some of the RNNT ones need higher max seq lengths.

nemo/collections/asr/modules/rnnt.py
borisfom: This one actually should be 1, as per previous implementation ...

titu1994 (approved):
LGTM

PR #3435, opened 2022-01-13
BioMegatron token classification tutorial fix to be compatible with current Megatron BERT

yidong72:
The token classification model uses the Megatron BERT model loaded from nemo file, which is converted from the Megatron-LM checkpoints. 

Currently it fails to reproduce the same accuracy. Need to investigate the bug.


yidong72:
I figure out the problem.  There is some extra logic in the Megatron-LM ParallelSelfAttention class
```python
        checkpoint_version = get_checkpoint_version()
        if checkpoint_version is not None:
           if checkpoint_version == 0:
               # [s, b, (3 * np * hn)] --> [s, b, (np * 3 * hn)]
               mixed_x_layer = self._transpose_last_dim(mixed_x_layer, 3, True)
           elif checkpoint_version == 1.0:
               # [s, b, (np * hn * 3)] --> [s, b, (np * 3 * hn)]
               mixed_x_layer = self._transpose_last_dim(mixed_x_layer, 3, False)
```
It transpose the tensor based on the checkpoint version. While in MeMo Megatron, there is no such check and transpose.

I fixed this  by doing the tensor transpose inside the `megatron_ckpt_to_nemo.py`
I can reproduce the old result as in the notebook.
```
[NeMo I 2022-01-13 22:54:46 token_classification_model:187] 
    label                                                precision    recall       f1           support   
    O (label_id: 0)                                         99.39      99.52      99.45      22092
    B-Disease (label_id: 1)                                 86.72      90.47      88.56        787
    I-Disease (label_id: 2)                                 94.93      89.27      92.01       1090
    -------------------
    micro avg                                               98.76      98.76      98.76      23969
    macro avg                                               93.68      93.09      93.34      23969
    weighted avg                                            98.77      98.76      98.76      23969
```
It is ready to review now.

nemo/collections/nlp/models/text_classification/text_classification_model.py
```diff
@@ -62,11 +62,18 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             config_file=self.register_artifact('language_model.config_file', cfg.language_model.config_file),
             config_dict=cfg.language_model.config,
             checkpoint_file=cfg.language_model.lm_checkpoint,
+            nemo_file=self.register_artifact('language_model.nemo_file', cfg.language_model.get('nemo_file', None)),
```
ericharper, line 65: It seems strange to me that we're passing a .nemo file as an arg here. Why are we doing this?

ericharper: This also has the effect that the nemo file for the text classification model will have a nemo file inside of it.

ericharper:
Does this work with model parallel models? It's fine if it does not since the BioMegatron models are not model parallel (yet). But maybe add a check in case a user tries to use a model parallel model?

yidong72:
> Does this work with model parallel models? It's fine if it does not since the BioMegatron models are not model parallel (yet). But maybe add a check in case a user tries to use a model parallel model?

It should work. Each model parallel model will have the same copy of the extra layers for downstream tasks. E.g. in a case of 2 tensor parallel worker, each worker will do its share of LM work and get the same hidden layer result which is fed to the classification layers. In the end, each worker will report the same classification prediction.

nemo/collections/nlp/models/text_classification/text_classification_model.py
yidong72: I tested saving the fine tune the model to nemo file and load it to run inference. It works fine.

```python
model_ner.save_to('test.nemo')
tt = nemo_nlp.models.TokenClassificationModel.restore_from('./test.nemo', trainer=trainer)
tt.half().evaluate_from_file(
    text_file=os.path.join(NER_DATA_DIR, 'sample_text_dev.txt'),
    labels_file=os.path.join(NER_DATA_DIR, 'sample_labels_dev.txt'),
    output_dir=exp_dir,
    add_confusion_matrix=False,
    normalize_confusion_matrix=True,
    batch_size=1
)
```

ericharper (approved):
LGTM. Thanks!

ericharper (approved):

PR #3445, opened 2022-01-14
Add randomized bucketing

VahidooX:
+ This PR Adds different bucketing strategies to improve the performance of ASR bucketing training accuracy-wise. A new parameter added to the data loaders (train_ds.bucketing_strategy) to specify one of the following modes:
1-fixed_order: the same order of buckets are used for all epochs
2-syned_randomized: each epoch would have a different order of buckets
3-fully_randomized: similar to syned_randomized but each GPU has its own random order. So GPUs would not be synced.

+Also added the support of layer norm for Conformer layers instead of batch norm.


nemo/collections/asr/data/audio_to_text.py
```diff
@@ -1492,3 +1492,31 @@ def __next__(self):
...
+    def __iter__(self):
+        shuffled_order = self.rnd_gen.permutation(len(self.datasets))
+        for dataset_idx in shuffled_order:
+            d = self.datasets[dataset_idx]
+            assert isinstance(d, IterableDataset), "ChainDataset only supports IterableDataset"
+            for x in d:
+                yield x
+
+
+class RandomizedChainDataset(ChainDataset):
```
titu1994, line 1511: Duplicate class ?

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -22,6 +23,7 @@
 from torch.utils.data import ChainDataset
 
 from nemo.collections.asr.data import audio_to_text, audio_to_text_dali
+from nemo.collections.asr.data.audio_to_text import RandomizedChainDataset
 from nemo.utils import logging
```
titu1994, line 27: Don't import like this, use the above two imports.

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -403,3 +363,59 @@ def convert_to_config_list(initial_list):
...
+                    dataset=dataset, bucketing_batch_size=bucketing_batch_sizes[idx]
+                )
+        else:
+            logging.info(
+                f"Batch bucketing is enabled for {len(datasets)} buckets with fixed batch size of {ds_config['batch_size']}!"
+            )
+
+    if len(datasets) == 1:
+        return datasets[0]
+    bucketing_strategy = ds_config.get('bucketing_strategy', 'synced_randomized')
```
titu1994, line 386: Let's add this config param to all of the major ASR model configs so it's visible. QuartzNet, Citrinet, ContextNet, Conformer (CTC and Transducer)

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -403,3 +363,59 @@ def convert_to_config_list(initial_list):
...
+    elif bucketing_strategy == 'fully_randomized':
+        return RandomizedChainDataset(datasets=datasets, rnd_seed=random.randint(0, 30000))
+    else:
+        raise ValueError(
+            f'bucketing_strategy={bucketing_strategy} is not supported! Supported strategies are randomzied and fixed_order.'
+        )
+
+
+def calc_bucketing_batch_sizes(ds_config, datasets_len):
+    bucketing_batch_size = ds_config['bucketing_batch_size']
```
titu1994, line 400: Same with this argument as above

nemo/collections/asr/models/label_models.py
```diff
@@ -170,10 +169,10 @@ def __setup_dataloader_from_config(self, config: Optional[Dict]):
                 normalize_audio=config.get('normalize_audio', False),
             )
 
-        if type(dataset) is ChainDataset:
-            collate_ds = dataset.datasets[0]
-        else:
+        if hasattr(dataset, 'fixed_seq_collate_fn'):
```
titu1994, line 172: The others above were fine, but this is very brittle. Do a better check.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -123,6 +125,7 @@ def __init__(
         untie_biases=True,
         pos_emb_max_len=5000,
         conv_kernel_size=31,
+        conv_norm_type='batch_norm',
         dropout=0.1,
```
titu1994, line 129: So this is backwards compatible ?

titu1994 (changes requested):
Minor comments. Basically let's add these bucket config args to all of the ASR models since they are relatively important + very useful for training.

nemo/collections/asr/data/audio_to_text.py
VahidooX: dropped it.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: fixed.

nemo/collections/asr/models/label_models.py
VahidooX: Made it consistent with others.

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: Yes, the default would be batch_norm which has been used for older models.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: Added.

nemo/collections/asr/data/audio_to_text_dataset.py
VahidooX: Added.

titu1994 (dismissed):
Thanks !

docs/source/asr/datasets.rst
```diff
@@ -266,3 +266,57 @@ Note that file structures are flattened such that all audio files are at the top
...
+When bucketing_batch_size is set, traind_ds.batch_size need to be set to 1.
+
+Training an ASR model on audios sorted based on length may affect the accuracy of the model. We introduced some strategies to mitigate it.
+We support three types of bucketing strategies:
+
+*   fixed_order: the same order of buckets are used for all epochs
+*   syned_randomized (default): each epoch would have a different order of buckets. Order of the buckets is shuffled every epoch.
+*   fully_randomized: similar to syned_randomized but each GPU has its own random order. So GPUs would not be synced.
+
+Tha parameter train_ds.bucketing_strategy can be set to specify one of these strategies. The recommended strategy is syned_randomized which gives the highest training speedup.
```
titu1994, line 319: Synced_randomized incorrectly spelled.

titu1994 (dismissed):
Looks fantastic. Can be merged.

docs/source/asr/datasets.rst
VahidooX: fixed.

titu1994 (approved):
Fantastic. Ready to merge

PR #3457, opened 2022-01-19
Enforce legacy tokenizer for sentencepiece to add special tokens for T5

MaximumEntropy:


ericharper (approved):
LGTM. Thanks!

titu1994 (commented):
It looks correct to me, and Citrinets work without breaking so its ok from my side. Leaving another reviewer for NMT side review.

PR #3465, opened 2022-01-19
Fix the tutorial notebooks bug

yidong72:
`Joint_Intent_and_Slot_Classification.ipynb`
`Text_Classification_Sentiment_Analysis.ipynb`
problem are caused by extra `cfg` key in the checkpoint file 

`Relation_Extraction-BioMegatron.ipynb`
problem is caused by forgetting to set the correct `model_config` variable.


ericharper (approved):
LGTM. Thanks!

PR #3467, opened 2022-01-19
Add Apex import guard

ericharper:


nemo/collections/nlp/modules/common/megatron/module.py
```diff
@@ -34,6 +45,10 @@ class MegatronModule(torch.nn.Module):
 
     def __init__(self, share_word_embeddings=True):
         super(MegatronModule, self).__init__()
+
+        if not HAVE_APEX:
+            logging.warning("Apex was not found. Using model parallel or megatron models will error out.")
```
MaximumEntropy, line 50: Should we leave this as a warning or just ask the user to install Apex here?

nemo/collections/nlp/modules/common/megatron/module.py
```diff
@@ -140,6 +155,8 @@ def float_conversion(val):
 class Float16Module(MegatronModule):
     def __init__(self, module, precision):
         super().__init__()
+        if not HAVE_APEX:
```
MaximumEntropy, line 158: Same as above.

MaximumEntropy (commented):
Just a minor comment on warnings vs raising an error.

nemo/collections/nlp/modules/common/megatron/module.py
```diff
@@ -15,10 +15,21 @@
 """Megatron Module"""
 
 import torch
-from apex.transformer import parallel_state, tensor_parallel
 from torch.autograd import Variable
 from torch.nn.parameter import Parameter
 
+try:
```
ekmb, line 21: @ericharper also in https://github.com/NVIDIA/NeMo/blob/r1.6.0/nemo/collections/nlp/modules/common/megatron/megatron_bert.py#L20

nemo/collections/nlp/modules/common/megatron/module.py
ericharper: Updated the message.

nemo/collections/nlp/modules/common/megatron/module.py
ericharper: Updated the message.

nemo/collections/nlp/modules/common/megatron/module.py
ericharper: Was actually able to remove that file all together.

MaximumEntropy (approved):
LGTM!

PR #3477, opened 2022-01-20
Remove wordtokenizer example from NLP tokenizer notebook

aklife97:
WordTokenizer seems to be erroring out, removing it from the tutorial for now.


okuchaiev (approved):
lgtm

PR #3478, opened 2022-01-20
NLPDDPPlugin find_unused_parameters is configurable

mlgill:
This PR makes `find_unused_parameters` user configurable in `NLPDDPPlugin` by setting a default dictionary with the current value (`False`) and updating that dictionary with user provided kwargs. This ensures that the value is always `False`, even if it changes from its current default (also `False`) in PyTorch's `DistributedDataParallel`.

An alternative option would be to remove it from the initialization call to `DistributedDataParallel` since it is already `False` by default in PyTorch. 

The reason for this request is that `find_unused_parameters` must be `True` for MegaMolBART.

I have read the contribution guidelines. All tests that are currently passing in `main` pass with this commit. (Note that two tests fail in NeMo `main`.) The code has been formatted with black and the commit is signed. This change was discussed with @ericharper several months ago.

Signed-off-by: Michelle Gill <mgill@nvidia.com>


mlgill:
cc @ericharper for vis

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -102,8 +106,7 @@ def configure_ddp(self):
                 device_ids=device_ids,
                 output_device=device_ids[0],
                 process_group=app_state.data_parallel_group,
-                find_unused_parameters=False,
```
ericharper, line 105: I think we just need to remove this line? I don't think we want to set our own default.

mlgill: OK, that's fine too. It's the easier option but the default will then change in NeMo if it ever changes in PTL. I'll assume that's OK and will update right now.

mlgill: Done

ericharper (approved):
LGTM. Thanks!

PR #3485, opened 2022-01-21
Enable Tarred Dataset Support for NVIDIA DALI

titu1994:
# Changelog
- Add support for tarred dataset support for Nvidia DALI 1.11+


nemo/collections/asr/data/audio_to_text_dali.py
```diff
@@ -320,22 +324,69 @@ def __init__(
...
+                    read_idxs=True,
+                    random_shuffle=shuffle,
+                    shard_id=self.shard_id,
+                    num_shards=self.num_shards,
+                    pad_last_batch=True,
+                )
+
+                self.is_tarred_dataset = False
+
+            elif audio_tar_filepaths is not None and audio_tar_index_filepaths is not None:
```
JanuszL, line 347: I would make `audio_tar_index_filepaths` optional as DALI can build index on the fly.
You can add a check if it is empty or length is equal to audio_tar_filepaths.

nemo/collections/asr/data/audio_to_text_dali.py
```diff
@@ -320,22 +324,69 @@ def __init__(
...
+                    audio_tar_filepaths, shard_strategy=shard_strategy, world_size=world_size, global_rank=global_rank
+                )
+                audio_tar_index_filepaths = expand_audio_filepaths(
+                    audio_tar_index_filepaths,
+                    shard_strategy=shard_strategy,
+                    world_size=world_size,
+                    global_rank=global_rank,
+                )
+
+                if len(audio_tar_filepaths) != len(audio_tar_index_filepaths):
```
JanuszL, line 358:
```suggestion
                if len(audio_tar_filepaths) != len(audio_tar_index_filepaths) and len(audio_tar_index_filepaths) != 0:
```

nemo/collections/asr/data/audio_to_text_dali.py
```diff
@@ -320,22 +324,69 @@ def __init__(
...
+                    raise ValueError(
+                        f"Number of filepaths provided for `audio_tar_filepaths` must match "
+                        f"`audio_tar_index_filepaths`. Got {len(audio_tar_filepaths)} audio_tar_filepaths and "
```
JanuszL, line 361:
```suggestion
                        f"`audio_tar_index_filepaths` or f"`audio_tar_index_filepaths` should be empty. Got {len(audio_tar_filepaths)} audio_tar_filepaths and "
```

nemo/collections/asr/data/audio_to_text_dali.py
```diff
@@ -320,22 +324,69 @@ def __init__(
...
+                )[0]
+                indices = dali.fn.get_property(tar_file, key="source_info")
+                indices = dali.fn.pad(indices)
+
+                self.is_tarred_dataset = True
+
+            else:
+                raise RuntimeError(
+                    "When using DALI datasets, either `audio_tar_filepaths` "
+                    "and `audio_tar_index_filepaths` should either both be None (sequential dataset)"
```
JanuszL, line 387: I would adjust this as well.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -62,10 +68,21 @@
...
-    --sort_in_shards
+    --workers=-1 \
+    --sort_in_shards \
     --shuffle_seed=1 \
     --write_metadata
 
+
+python convert_to_tarred_audio_dataset.py \
+    --manifest_path="/home/smajumdar/PycharmProjects/nemo-eval/nemo_beta_eval/librispeech/manifests/dev_test_all.json" \
+    --target_dir="/media/smajumdar/data/Datasets/Librispeech_Dev_Test_Tarred/" \
```
JanuszL, line 79: I would replace it with a generic place holder.

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Hmm I actually want to avoid the scenario where Dali silently builds index. Since tarred dataset is for large datasets, and index building on millions of files takes non trivial time, I'd rather enforce the user to supply preprocessed indices.

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Accepted, patched manually as otherwise DCO complains.

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Not selecting this change, as Nemo will enforce precomputed index for efficiency reasons on HPC. We don't want users to silently compute indices on the fly on millions of files in the tarred datasets.

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Reasons above.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
titu1994: Ah my bad. Haha this should not be part of the PR but for my testing only.

titu1994 (commented):
Thank you @JanuszL for the review !

nemo/collections/asr/data/audio_to_text_dali.py
JanuszL: Sure. DALI shows a warning if you don't provide the index but it is not a technical blocker.
Up to you.

nemo/collections/asr/data/audio_to_text_dali.py
JanuszL: Sure. Up to you.

nemo/collections/asr/data/audio_to_text_dali.py
JanuszL: Ack

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
JanuszL: 👍

nemo/collections/asr/data/audio_to_text_dali.py
JanuszL: For the reason stated before it is not needed after all.

tests/collections/asr/test_asr_datasets.py
```diff
@@ -406,3 +406,168 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
...
+
+        batch_size = 8
+        device = 'gpu' if torch.cuda.is_available() else 'cpu'
+        texts = []
+
+        with tempfile.NamedTemporaryFile(mode='w', encoding='utf-8') as f:
+            num_samples = 0
+            with open(manifest_path, 'r') as m:
+                for line in m:
+                    num_samples += 1
```
JanuszL, line 430: How about just:
```suggestion
            num_samples = sum(1 for line in open(manifest_path, 'r'))
```

tests/collections/asr/test_asr_datasets.py
```diff
@@ -406,3 +406,168 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
...
+                ]
+                shuffled_transcripts.extend(transcripts)
+                count += len(transcripts)
+            assert count == num_samples
+
+            samples_changed = 0
+            for orig, shuffled in zip(original_transcripts, shuffled_transcripts):
+                if orig != shuffled:
+                    samples_changed += 1
+            assert samples_changed > 1  # assume after shuffling at least 1 sample was displaced
```
JanuszL, line 492: Maybe you can add also:
```
for og_transcript, shuffled_transcript in zip(sorted(original_transcripts), sorted(shuffled_transcripts)):
                og_transcript== shuffled_transcript

File: tests/collections/asr/test_asr_datasets.py [hunk 8]
```diff
@@ -406,3 +406,168 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
...
+        batch_size = 8
+        texts = []
+
+        with tempfile.NamedTemporaryFile(mode='w', encoding='utf-8') as f:
+            num_samples = 0
+            with open(manifest_path, 'r') as m:
+                for ix, line in enumerate(m):
+                    # line = line.replace("tests/data/", "tests/.data/").replace("\n", "")
+                    # f.write(f"{line}\n")
+
```
JanuszL, line 515: To remove?

tests/collections/asr/test_asr_datasets.py
```diff
@@ -406,3 +406,168 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
...
+                    data = json.loads(line)
+                    texts.append(data['text'])
+                    num_samples += 1
```
JanuszL, line 518:
```suggestion
                    num_samples = ix
```
Or you can remove `enumerate`.

JanuszL (dismissed):
Just small nitpicks here and there

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -81,6 +88,13 @@
 from joblib import Parallel, delayed
 from omegaconf import DictConfig, OmegaConf, open_dict
 
+try:
+    import create_dali_tarred_dataset_index as dali_index
+
+    DALI_INDEX_SCRIPT_AVAILABLE = True
```
jantonguirao, line 94: why would `create_dali_tarred_dataset_index` be not available? It is part of this repro. Shouldn't  `dali_index.INDEX_CREATOR_AVAILABLE` be enough?

nemo/collections/asr/data/audio_to_text_dali.py
```diff
@@ -320,22 +324,69 @@ def __init__(
...
+                    ext=["wav"],
+                    missing_component_behavior="error",
+                    random_shuffle=shuffle,
+                    shard_id=self.shard_id,
+                    num_shards=self.num_shards,
+                    pad_last_batch=True,
+                )
+                audio = dali.fn.decoders.audio(
+                    tar_file, dtype=dali.types.FLOAT, downmix=True, sample_rate=float(self.sample_rate),
+                )[0]
```
jantonguirao, line 378: I am not quite sure what this `[0]` is meant to do. When downmixing, the shape should be `[audio_len,]` (as opposed to `[audio_len, nchannels]` when not downmixing. So `[0]` would mean getting the first audio sample (a scalar) from the sequence.

jantonguirao, line 378: I originally got very confused with that `[0]` as I read it as a slice operation (forgot that audio decoder returns two outputs and you were getting the first). I suggest this instead:
```suggestion
                audio, _ = dali.fn.decoders.audio(
                    tar_file, dtype=dali.types.FLOAT, downmix=True, sample_rate=float(self.sample_rate),
                )
```

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -156,6 +165,54 @@ def process_text(self, index) -> (List[int], int):
...
+    valid_shard_strategies = ['scatter', 'replicate']
+    if shard_strategy not in valid_shard_strategies:
+        raise ValueError(f"`shard_strategy` must be one of {valid_shard_strategies}")
+
+    if isinstance(audio_tar_filepaths, str):
+        # Replace '(' and '[' with '{'
+        brace_keys_open = ['(', '[', '<', '_OP_']
+        for bkey in brace_keys_open:
+            if bkey in audio_tar_filepaths:
+                audio_tar_filepaths = audio_tar_filepaths.replace(bkey, "{")
```
hainan-xv, line 178: Can this be achieved with a single regex replace instead of checking individual cases?

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -156,6 +165,54 @@ def process_text(self, index) -> (List[int], int):
...
+            logging.info("All tarred dataset shards will be scattered evenly across all nodes.")
+
+            if len(audio_tar_filepaths) % world_size != 0:
+                logging.warning(
+                    f"Number of shards in tarred dataset ({len(audio_tar_filepaths)}) is not divisible "
+                    f"by number of distributed workers ({world_size})."
+                )
+
+            begin_idx = (len(audio_tar_filepaths) // world_size) * global_rank
+            end_idx = begin_idx + (len(audio_tar_filepaths) // world_size)
```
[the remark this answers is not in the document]
hainan-xv: Is this suggestion missing a ')'?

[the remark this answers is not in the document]
jantonguirao: Yes, my bad. I meant:
```end_idx = begin_idx + len(audio_tar_filepaths) // world_size```

File: nemo/collections/asr/data/audio_to_text_dali.py [hunk 14]
```diff
@@ -550,6 +624,7 @@ def __init__(
         shuffle: bool = False,
         drop_last: bool = False,
         parser: Union[str, Callable] = 'en',
+        shard_strategy: str = "scatter",
```
hainan-xv, line 627: The use of " seems to be inconsistent with one line above. Should we try to be consistent in terms of use of quotation marks? I notice some inconsistencies in other parts of the file as well...

tests/collections/asr/test_asr_datasets.py
```diff
@@ -406,3 +406,168 @@ def test_dali_char_vs_ref_dataset(self, test_data_dir):
...
+                original_transcripts.extend(transcripts)
+                count += len(transcripts)
+            assert count == num_samples
+
+            # Assert transcripts are correct
+            for text, og_transcript in zip(texts, original_transcripts):
+                assert text == og_transcript
+
+            dataset = AudioToCharDALIDataset(
+                manifest_filepath=manifest_path,  # f.name,
```
hainan-xv, line 463: remove the comment here?

hainan-xv (commented):
Left some minor comments.

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Well strings should almost always be "" but python allows both, it's fine to try to be consistent but it's fine not follow it to the letter too. There's no pep against such mixing, not even black style comments about it.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
titu1994: Dali is an optional install that the user has to actively do, whereas the tarred dataset building script can be used even without even installing Nemo itself. So we test and then build if Dali is available (even if Nemo is not)

scripts/speech_recognition/create_dali_tarred_dataset_index.py
```diff
@@ -0,0 +1,95 @@
...
+
+def process_index_path(tar_paths, index_dir):
+    """
+    Appends the folder `{index_dir}` to the filepath of all tarfiles.
+    Example:
+         /X/Y/Z/audio_0.tar -> /X/Y/Z/{index_dir}/audio_0.index
+    """
+    index_paths = []
+    for path in tar_paths:
+        path = path.replace('.tar', '.index')
```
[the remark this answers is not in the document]
titu1994: True I'll split the path, update last then merge back. Very good catch

tests/collections/asr/test_asr_datasets.py
titu1994: I thought of doing @jantonguirao style, but both above options don't explicitly close the file afterwords. So I used the explicit form instead.

tests/collections/asr/test_asr_datasets.py
titu1994: Cool.

tests/collections/asr/test_asr_datasets.py
titu1994: Good point will add.

tests/collections/asr/test_asr_datasets.py
titu1994: Hmm I forgot why those are there in the first place, well no harm done I guess since tests pass locally either way.

titu1994 (commented):
Sorry for delay, some interesting RNNT stuff is taking priority :P

tests/collections/asr/test_asr_datasets.py
jantonguirao: so how about:
```
with open(manifest_path, 'r') as m:
    num_samples = m.count('\n')
```
?

nemo/collections/asr/data/audio_to_text.py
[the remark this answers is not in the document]
titu1994: Done

nemo/collections/asr/data/audio_to_text_dali.py
titu1994: Good point. Done.

tests/collections/asr/test_asr_datasets.py
titu1994: Nice and simple. Cool.

PR #3487, opened 2022-01-21
Properly support -1 for labels in ctc char models

titu1994:
# Changelog
- Add support for -1 in num_classes if vocabulary is provided for CTC char models.

### DIFF
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -171,6 +171,14 @@ class EncDecCTCModel(ASRModel, ExportableEncDecModel, ASRModuleMixin):
             if "feat_in" not in self._cfg.decoder or not self._cfg.decoder.feat_in:
                 raise ValueError("param feat_in of the decoder's config is not set!")
 
+            if self.cfg.decoder.num_classes < 1 and self.cfg.decoder.vocabulary is not None:
+                logging.info(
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/models/ctc_models.py
```diff
@@ -171,6 +171,14 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             if "feat_in" not in self._cfg.decoder or not self._cfg.decoder.feat_in:
                 raise ValueError("param feat_in of the decoder's config is not set!")
 
+            if self.cfg.decoder.num_classes < 1 and self.cfg.decoder.vocabulary is not None:
```
VahidooX, line 174: How about covering if self.cfg.decoder.num_classes is not set (like having None value)?
How about raising error when num_classes is not set and vocabulary is also None?
Do we actually need num_classes to be specified by user when we have and need vocabulary?
What happens when they do not match?

VahidooX (changes requested):

nemo/collections/asr/models/ctc_models.py
titu1994: All of those cases should error out since they are invalid. Only valid cases like -1 is provided + vocabulary is provided is handled. It is the users responsibility to provide correct number of classes or -1 and correct vocabulary. Yes the num classes and vocab size must always match.

VahidooX (approved):
LGTM!

PR #3513, opened 2022-01-25
Add HuggingFace Datasets to NeMo ASR Dataset script

titu1994:
# Changelog
- Add script to convert HuggingFace ASR datasets to preprocessed NeMo compatible datasets.
- These datasets can then be converted to tarred datasets as always.

# NOTE:
- The converted datasets DO NOT PERFORM TEXT PREPROCESING. That should be handled by the user at manifest level.
- The offline mode dataset transformation is fast but requires 3 copies of the dataset in its entirety at any given time - so it needs a lot of disk space.


scripts/speech_recognition/convert_hf_dataset_to_nemo.py
```diff
@@ -0,0 +1,392 @@
...
+        manifest_filename = f"{cfg.path.replace('/', '_')}_manifest.json"
+    else:
+        basedir = cfg.split_output_dir
+        split = os.path.split(cfg.split_output_dir)[-1]
+        manifest_filename = f"{split}_{cfg.path.replace('/', '_')}_manifest.json"
+
+        if not os.path.exists(cfg.split_output_dir):
+            os.makedirs(cfg.split_output_dir, exist_ok=True)
+
+        cfg.split = split
```
vsl9, line 316: The line should be moved out of `else` scope (delete extra tab).

scripts/speech_recognition/convert_hf_dataset_to_nemo.py
```diff
@@ -0,0 +1,392 @@
...
+from dataclasses import dataclass, is_dataclass
+from typing import Optional
+
+import hydra
+import librosa
+import soundfile
+import tqdm
+from datasets import Audio, IterableDataset, load_dataset
+from hydra.core.config_store import ConfigStore
+from omegaconf import OmegaConf, open_dict
```
vsl9, line 96: `open_dict` is unused (according to LGTM check)

vsl9 (approved):
Thank you!

PR #3524, opened 2022-01-26
Conformer onnx fix

borisfom:


nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -163,7 +163,11 @@ def forward(self, x, pad_mask=None):
             x.masked_fill_(pad_mask.unsqueeze(1), 0.0)
 
         x = self.depthwise_conv(x)
-        x = self.batch_norm(x)
+        if torch.onnx.is_in_onnx_export():
+            with torch.cuda.amp.autocast(enabled=False):
```
titu1994, line 167: I thought BN was the primary focus of autocast. Does it fail for some reason under onnx context?

nemo/core/classes/exportable.py
```diff
@@ -134,7 +134,7 @@ def export(
         dynamic_axes=None,
         check_tolerance=0.01,
     ):
-        my_args = locals()
+        my_args = locals().copy()
```
titu1994, line 137: Shouldn't this be a deep copy ?

nemo/core/classes/exportable.py
```diff
@@ -254,6 +253,7 @@ def _verify_runtime(
             import onnxruntime
         except (ImportError, ModuleNotFoundError):
             logging.warning(f"ONNX generated at {output}, not verified - please install onnxruntime.\n")
+            onnx.checker.check_model(onnx_model, full_check=True)
```
titu1994, line 256: This will fail if import fails

titu1994 (commented):
Some minor comments.

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: I am putting it under autocast(False) because otherwise it generates bad ONNX code

nemo/core/classes/exportable.py
borisfom: No its just to shut top warnings about locals list itself

nemo/core/classes/exportable.py
borisfom: onnxruntime and onyx are different packages

nemo/core/classes/exportable.py
titu1994: Ah ok

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -554,7 +555,7 @@ def __init__(
                 )
             )
             self.decoder_layers.append(nn.Conv1d(self.feat_hidden, self.feat_hidden, kernel_size=1, bias=True))
-            self.decoder_layers.append(nn.BatchNorm1d(self.feat_hidden, eps=1e-3, momentum=0.1))
+            self.decoder_layers.append(BatchNorm1dNoAutoCast(self.feat_hidden, eps=1e-3, momentum=0.1))
```
titu1994, line 558: Won't this affect pretrained checkpoints ?

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -157,9 +158,8 @@ def rel_shift(self, x):
             x (torch.Tensor): (batch, nheads, time, 2*time-1)
         """
         b, h, qlen, pos_len = x.size()  # (b, h, t1, t2)
-
-        # need to add a column of zeros on the left side of last two dimensions to perform the relative shifting
-        x = torch.nn.functional.pad(x, pad=(1, 0, 0, 0))  # (b, h, t1, t2+1)
+        # need to add a column of zeros on the left side of last dimension to perform the relative shifting
+        x = torch.nn.functional.pad(x, pad=(1, 0))  # (b, h, t1, t2+1)
```
titu1994, line 162: @VahidooX  please check this formula

nemo/core/classes/exportable.py
```diff
@@ -217,26 +145,9 @@ def export(
                         verbose=verbose,
                         export_params=export_params,
                         do_constant_folding=do_constant_folding,
-                        keep_initializers_as_inputs=keep_initializers_as_inputs,
                         dynamic_axes=dynamic_axes,
                         opset_version=onnx_opset_version,
                     )
-
-                    # Verify the model can be read, and is valid
-                    onnx_model = onnx.load(output)
```
titu1994, line 226: No more verification ?

nemo/core/classes/exportable.py
```diff
@@ -217,26 +145,9 @@ def export(
...
-                    onnx.checker.check_model(onnx_model, full_check=True)
-
-                    if check_trace:
```
titu1994, line 229: Or trace check ?

nemo/core/classes/exportable.py
```diff
@@ -285,88 +172,23 @@ def supported_export_formats(self):
...
-        replace_for_export(self, replace_1D_2D)
+        replace_for_export(self)
 
     def _export_teardown(self):
         """
         Override this method for any teardown code after export.
         """
         pass
 
-    def _wrap_forward_method(self):
```
titu1994, line 329: Without these methods RNNT export will fail. Also why such a heavy refactor back to monolithic design of export? It will make future models harder to deal with

nemo/utils/neural_type_utils.py
```diff
@@ -0,0 +1,63 @@
+from collections import defaultdict
```
titu1994, line 1: This can go into the Nemo/ core / utils directory instead of this one. Well it's just semantics it's fine here too but more appropriate there.

scripts/export.py
```diff
@@ -141,6 +150,11 @@ def nemo_export(argv):
 
     logging.info("Successfully exported to {}".format(out))
 
+    del model
+
+    if args.runtime_check:
+        verify_runtime(out, input_list, input_dict, input_names, output_names, output_example)
+
```
titu1994, line 157: This should really be an implicit part of model export, not an external function. Most people will forget to verify their export in the code when they do model.export().

titu1994 (changes requested):
This refactor breaks RNNT export totally. Even if that's fixed, the refactor to remove testing and runtime trace check as a part of export is necessary.

nemo/collections/asr/modules/conv_asr.py
borisfom: No they are fine - submodule has same name and parameters.

scripts/export.py
borisfom: Well I agree in principle, but it should be available externally to be able to check models that won't allow 2 copies on GPU at the same time.

borisfom:
Nothing is broken - my previous RNNT refactoring removed all those overridden methods. So this is cleanup part to remove unused hooks and keep code more readable.

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: We discussed that yesterday on slack

borisfom: (1,0,0,0) is equivalent to (1,0) - just not misleading. Intent is to pad a single dimension

titu1994 (dismissed):
RNNT seems fine for now.

nemo/utils/neural_type_utils.py
borisfom: done.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -262,7 +264,11 @@ def forward_for_export(self, audio_signal, length=None):
         if self.att_context_size[1] >= 0:
             att_mask = att_mask.tril(diagonal=self.att_context_size[1])
         att_mask = ~att_mask
-        pad_mask = ~pad_mask
+
+        if self.use_pad_mask:
+            pad_mask = ~pad_mask
+        else:
+            pad_mask = None
```
titu1994, line 271: Just a note, None mode isn't used anywhere, is that ok ?

tests/collections/asr/test_asr_exportables.py
```diff
@@ -119,13 +119,12 @@ def test_EncDecCitrinetModel_limited_SE_export_to_onnx(self, citrinet_model):
...
     def test_EncDecRNNTModel_export_to_onnx(self, citrinet_rnnt_model):
-        citrinet_rnnt_model.freeze()
-        model = citrinet_rnnt_model.train()
+        model = citrinet_rnnt_model.cuda()
 
         with tempfile.TemporaryDirectory() as tmpdir:
             fn = 'citri_rnnt.onnx'
             filename = os.path.join(tmpdir, fn)
-            model.export(output=filename, verbose=False, check_trace=True)
+            model.export(output=filename, verbose=False)
```
titu1994, line 127: Just wondering, why aren't we checking the trace is correct similar to before ?

titu1994 (approved):
LGTM, minor comments

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: What is the purpose of this use_pad_mask? It looks to be always True, right?

nemo/utils/export_utils.py
```diff
@@ -20,6 +22,134 @@
...
+
+
+_EXT_DICT = {
+    ".pt": ExportFormat.TORCHSCRIPT,
+    ".ts": ExportFormat.TORCHSCRIPT,
+    ".onnx": ExportFormat.ONNX,
+}
+
+
+class BatchNorm1dNoAutoCast(nn.BatchNorm1d):
```
VahidooX, line 40: How about giving a more general/flexible name for this batch norm as it may get updated and changed over the time. For instance how about naming it something like BatchNorm1dExportable?

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -95,33 +95,31 @@ def forward(self, x, att_mask=None, pos_emb=None, pad_mask=None):
         Returns:
             x (torch.Tensor): (B, T, d_model)
         """
+        dtype = x.dtype
         residual = x
         x = self.norm_feed_forward1(x)
         x = self.feed_forward1(x)
-        x = self.fc_factor * self.dropout(x) + residual
+        residual = residual + self.dropout(x) * self.fc_factor
```
VahidooX, line 102: Was there any specific reason naming of x and residual are changed here?

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -95,33 +95,31 @@ def forward(self, x, att_mask=None, pos_emb=None, pad_mask=None):
...
-        x = self.norm_feed_forward2(x)
+        x = self.norm_feed_forward2(residual)
         x = self.feed_forward2(x)
-        x = self.fc_factor * self.dropout(x) + residual
+        residual = residual + self.dropout(x) * self.fc_factor
 
-        x = self.norm_out(x)
-        return x
+        x = self.norm_out(residual)
+        return x.to(dtype=dtype)
```
VahidooX, line 122: Can x here have a different type other than the type of the input?

nemo/utils/export_utils.py
VahidooX: Many other models in nemo are using BatchNorm1d. Do we know if they have encountered any issues or not or why it just happened for Conformer and not the other models?

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -160,24 +158,31 @@ def __init__(self, d_model, kernel_size, norm_type='batch_norm'):
...
+    def forward(self, x, pad_mask=None):
+        x = x.transpose(1, 2)
+        x = self.pointwise_conv1(x)
+        x = nn.functional.glu(x, dim=1)
+
+        if pad_mask is not None:
+            x.masked_fill_(pad_mask.unsqueeze(1), 0.0)
+
+            if torch.onnx.is_in_onnx_export():
+                with torch.cuda.amp.autocast(enabled=False):
```
VahidooX, line 180: Is this part necessary as we already have these autocasts and converisons already in the BatchNorm1dNoAutoCast?
BatchNorm1dNoAutoCast just misses the last .to(dtype=x.dtype) which can be added.

VahidooX (changes requested):

nemo/collections/asr/modules/conformer_encoder.py
borisfom: I have prepared 'mask disabled' mode - I think you have suggested on Slack it could be skipped for inference - but numeric results were rather different so I left it disabled for now.

tests/collections/asr/test_asr_exportables.py
borisfom: onnxruntime had some intermittent CUDA issues on my box so I disabled it just in case to avoid breaking CI randomly.

nemo/utils/export_utils.py
borisfom: I had to also use it for CitriNet, yes. 
Regarding name - I think this way it's more descriptive.

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: Yes - with autocast, result of norm_out is fp32 when residual is x. I traced that. This is because autocast (correctly) leaves those weights in fp32. But we want to cast back to original type on return.

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: With ONNX opset 15, I have encountered some errors related to aliasing. Those changes fixed that.

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: We are not using BatchNorm1dNoAutoCast here - I had to use that helper class in Citrinet because there batch conv's forward() was not called explicitly. Here it is explicit, and I wanted to keep it fused with previous conv, so I did not use helper class.

borisfom:
> While it does use the new change with "x = self.do_conv_norm(x.to(dtype=torch.float)).to(dtype=x.dtype)" it still finds half and not float. Am I doing something wrong?
@naor2013: Try with updated version. Are you using  .half() on the model? You should use torch.cuda.amp.autocast().

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: (changed to BatchNorm1dNoAutoCast for clarity)

naor2013:
@borisfom Thank you! After your new commit, it worked!

naor2013:
@borisfom Do you know why when I exported it without autocast, the onnx model had 3994 nodes but with autocast it had 4579? Are there more nodes to convert between the sizes or something?

borisfom:
> @borisfom Do you know why when I exported it without autocast, the onnx model had 3994 nodes but with autocast it had 4579? Are there more nodes to convert between the sizes or something?

Yes. Many of those nodes will go away if you run 'polygraphy surgeon sanitize --fold-constants ...'

nemo/utils/export_utils.py
```diff
@@ -219,7 +277,4 @@ def replace_for_export(model: nn.Module, replace_1D_2D: bool = False) -> nn.Modu
     Returns:
         model, possibly modified in-place
     """
-    replace_modules(model, default_Apex_replacements)
-    if replace_1D_2D:
-        # TODO: add squeeze/unsqueeze
-        replace_modules(model, default_1D_2D_replacements)
+    replace_modules(model, default_replacements)
```
titu1994, line 280: This will replace every instance of BN1d - even for citrinet, ContextNet, Jasper QN etc - will they be ok with this chance ?
Is this some regression in onnx itself for BN1d?

titu1994 (approved):
Overall it's good, but new change will now replace conv 1d in all older models - is that fine ?

VahidooX (approved):

naor2013:
> > @borisfom Do you know why when I exported it without autocast, the onnx model had 3994 nodes but with autocast it had 4579? Are there more nodes to convert between the sizes or something?
> 
> Yes. Many of those nodes will go away if you run 'polygraphy surgeon sanitize --fold-constants ...'

@borisfom after running that it removed some nodes but still the autocast version had more nodes than the "regular" version. Why is that?

nemo/utils/export_utils.py
```diff
@@ -20,6 +22,134 @@
...
+        import onnxruntime
+    except (ImportError, ModuleNotFoundError):
+        logging.warning(f"ONNX generated at {output}, not verified - please install onnxruntime.\n")
+        onnx.checker.check_model(onnx_model, full_check=True)
+        return
+
+    onnx_session_opt = onnxruntime.SessionOptions()
+    onnx_session_opt.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_ALL
+
+    providers = [
```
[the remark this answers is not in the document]
borisfom: Why, they are.

nemo/utils/export_utils.py
borisfom: Yes they will be OK and that's what we want. BN1d weights are always in fp32, so those casts are no-ops, really - just helping to generate proper ONNX code.

nemo/collections/asr/parts/submodules/conformer_modules.py
borisfom: Yes, that's what I see if I print out the type. not happening after calling other layers - maybe autocast treated the return from forward() differently.

PR #3525, opened 2022-01-26
fixed the num_classes bug of conv decoder.

VahidooX:


nemo/collections/asr/modules/conv_asr.py
```diff
@@ -420,6 +420,14 @@ def output_types(self):
     def __init__(self, feat_in, num_classes, init_mode="xavier_uniform", vocabulary=None):
         super().__init__()
 
+        if vocabulary is None and num_classes < 0:
+            raise ValueError(
+                f"Neither of the vocabulary and num_classes are set! At least one of them need to be set."
+            )
+
+        if num_classes <= 0:
+            num_classes = len(vocabulary)
```
titu1994, line 429: Set a logging message here telling the user that num_classes was updated

titu1994 (approved):

nemo/collections/asr/modules/conv_asr.py
VahidooX: Added.

titu1994 (approved):

PR #3527, opened 2022-01-27
Add Change Log builder to NeMo

titu1994:
# Changelog
- Build a raw changelog automatically for every new NeMo release based on a comparison from PRs between newest version and previous version. 
- Raw changelog is then tidied to display on final page. 

# Usage

Call the builder and visit the GitHub Action that builds the changelog. It's "Print Changelog" section will have the raw changelog, which needs to be tidied up. 

```sh
git tag vX.YZ
git push origin vX.Y.Z
```

## Redo steps
If something should go wrong tag can be deleted using 

```sh
git tag vX.YZ -d
git push origin vX.Y.Z -d
```


ericharper (approved):
LGTM. Thanks!
How do we use it though?

nithinraok:
This is great, but are speaker-related ones grabbed for asr?

titu1994:
You can add more regex rules to group them into ASR later. For now a good chunk gets moved into ASR but some are lost into buffixes of untagged. Regex classified is only so accurate after all.

PR #3540, opened 2022-01-27
Bugfix, need to freeze the model

yidong72:
The model needs to be in the eval mode by calling `freeze` so the dropout performs correctly.


ericharper (approved):
LGTM. Thanks!

PR #3548, opened 2022-01-28
Fix restoring from checkpoint for case when `model.common_dataset_parameters.label_vocab_dir` is provided

PeganovAnton:
If model is being restored from checkpoint, then `model.common_dataset_parameters.label_vocab_dir` should be prepended to paths to vocab files.


ekmb (approved):

okuchaiev:
@PeganovAnton looks like this slipped. Could you please send it to main instead?

Issue #3559, opened 2022-01-29
No module named 'apex' in colab notebook 00_NeMo_Primer.ipynb


amrzv:
**Describe the bug**

When running `00_NeMo_Primer.ipynb` notebook during `import nemo.collections.nlp as nemo_nlp` raises an error:
`ModuleNotFoundError: No module named 'apex'`

**Steps/Code to reproduce bug**

Opened in colab tutorial notebook [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/NVIDIA/NeMo/blob/main/tutorials/00_NeMo_Primer.ipynb) and "Run all cells". Then after installation section at importing stage error raises.

**Expected behavior**

End-to-end notebook executions without errors.

**Environment overview (please complete the following information)**

 - Environment location: colab
 - Method of NeMo install: as in notebook
 ```
BRANCH = 'main'
!python -m pip install git+https://github.com/NVIDIA/NeMo.git@$BRANCH#egg=nemo_toolkit[all]
```

**Environment details**
- OS version: linux
- PyTorch version: 1.10.0+cu111
- Python version: 3.7.12

**Additional context**
Error occurs in both kernel-types: CPU and GPU.
![apex1](https://user-images.githubusercontent.com/36787333/151659365-9b5a7593-a343-4a8f-8f41-83ff04dfcfc4.jpg)
![apex2](https://user-images.githubusercontent.com/36787333/151659369-7af2fed5-83e7-4ff6-8917-d1960f473453.jpg)

conceptofmind:
You likely will have to install Nvidia Apex into the environment. Nvidia Apex can be found here: https://github.com/NVIDIA/apex. I had just done a clean install on Ubuntu yesterday. Make sure the version of Pytorch matches the version of CUDA or apex will not install correctly.

titu1994:
Is this with Nemo 1.6? @ericharper  we need a patch release with a fix for this. Apex is NOT a hard requirement other than for Megatron, and core notebooks like this should not require apex either until you try to train a model.

titu1994:
@yidong72 add import guards on all ptuning imports and ensure than NLP can be imported in an apex-less environment

amrzv:
> Is this with Nemo 1.6?
> 
> @ericharper we may need a patch release with a fix for this if it's also in r1.6.0 (is ptuning in 1.6?). If not we need to patch main branch.
> 
[...]

PR #3492, opened 2022-01-22
Refactor megatron apex guard

itzsimpl:
Starting of a nemo:1.5.0 container and using `./reinstall.sh` to upgrade to nemo:1.6.0rc0 the `import nemo.collections.nlp` will crash:
```
Python 3.8.12 | packaged by conda-forge | (default, Sep 29 2021, 19:52:28) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import nemo.collections.nlp
NOTE! Installing ujson may make loading annotations faster.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/workspace/nemo2/nemo/collections/nlp/__init__.py", line 15, in <module>
    from nemo.collections.nlp import data, losses, models, modules
  File "/workspace/nemo2/nemo/collections/nlp/models/__init__.py", line 15, in <module>
    from nemo.collections.nlp.models.dialogue_state_tracking.sgdqa_model import SGDQAModel
  File "/workspace/nemo2/nemo/collections/nlp/models/dialogue_state_tracking/__init__.py", line 15, in <module>
    from nemo.collections.nlp.models.dialogue_state_tracking.sgdqa_model import SGDQAModel
  File "/workspace/nemo2/nemo/collections/nlp/models/dialogue_state_tracking/sgdqa_model.py", line 33, in <module>
    from nemo.collections.nlp.models.nlp_model import NLPModel
  File "/workspace/nemo2/nemo/collections/nlp/models/nlp_model.py", line 29, in <module>
    from nemo.collections.nlp.modules import BertModule
  File "/workspace/nemo2/nemo/collections/nlp/modules/__init__.py", line 16, in <module>
    from nemo.collections.nlp.modules.common import (
  File "/workspace/nemo2/nemo/collections/nlp/modules/common/__init__.py", line 29, in <module>
    from nemo.collections.nlp.modules.common.tokenizer_utils import get_tokenizer, get_tokenizer_list
  File "/workspace/nemo2/nemo/collections/nlp/modules/common/tokenizer_utils.py", line 27, in <module>
    from nemo.collections.nlp.parts.nlp_overrides import HAVE_APEX
  File "/workspace/nemo2/nemo/collections/nlp/parts/nlp_overrides.py", line 33, in <module>
    from nemo.collections.nlp.modules.common.megatron.module import Float16Module
  File "/workspace/nemo2/nemo/collections/nlp/modules/common/megatron/__init__.py", line 15, in <module>
[...]

Issue #3472, opened 2022-01-20
nemo2riva of Punctuation_and_Capitalization with NeMo:1.6.0rc0 crashes


itzsimpl:
I have a prebuilt Punctuation_and_Capitalization model (NeMo:1.4.0). Running nemo2riva with NeMo:1.5.1 performs the conversion as it should, however, running nemo2riva with master branch of NeMo (1.6.0rc0) crashes with the following error
```
INFO: Logging level set to 20                                                                                                                                                                                                                   
INFO: Restoring NeMo model from 'data/nemo/Punctuation_and_Capitalization.nemo'                                                                                                                                                                 
[NeMo I 2022-01-20 15:59:39 common:489] Model instantiation from target class nemo.collections.nlp.models.token_classification.punctuation_capitalization_model.PunctuationCapitalizationModel failed with following error.                     
    Falling back to `cls`.                                                                                                                                                                                                                      
    Traceback (most recent call last):                                                                                                                                                                                                          
      File "/data/NeMo/nemo/core/classes/common.py", line 463, in from_config_dict                                                                                                                                                              
        imported_cls = import_class_by_path(target_cls)                                                                                                                                                                                         
      File "/data/NeMo/nemo/utils/model_utils.py", line 462, in import_class_by_path                                                                                                                                                            
        mod = __import__(path, fromlist=[class_name])                                                                                                                                                                                           
      File "/data/NeMo/nemo/collections/nlp/__init__.py", line 15, in <module>                                                                                                                                                                  
        from nemo.collections.nlp import data, losses, models, modules                                                                                                                                                                          
      File "/data/NeMo/nemo/collections/nlp/models/__init__.py", line 15, in <module>                                                                                                                                                           
        from nemo.collections.nlp.models.dialogue_state_tracking.sgdqa_model import SGDQAModel                                                                                                                                                  
      File "/data/NeMo/nemo/collections/nlp/models/dialogue_state_tracking/__init__.py", line 15, in <module>                                                                                                                                   
        from nemo.collections.nlp.models.dialogue_state_tracking.sgdqa_model import SGDQAModel                                                                                                                                                  
      File "/data/NeMo/nemo/collections/nlp/models/dialogue_state_tracking/sgdqa_model.py", line 33, in <module>                                                                                                                                
        from nemo.collections.nlp.models.nlp_model import NLPModel                                                                                                                                                                              
[...]

PR #3579, opened 2022-02-01
Add guards

ericharper:
Signed-off-by: ericharper <complex451@gmail.com>

# What does this PR do?
Adds apex import guards.
Please have one of the following:

* Fixes # (issue)
* New feature: (short feature description)

## Before your PR is "Ready for review"
- [x] Did you add or updated any necessary documentation?

## Who can review?

### DIFF
--- a/nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
@@ -39,6 +39,11 @@ import time
 import numpy as np
 import torch
 
+from nemo.collections.nlp.data.language_modeling.megatron.blendable_dataset import BlendableDataset
+from nemo.collections.nlp.data.language_modeling.megatron.indexed_dataset import make_dataset as make_indexed_dataset
+from nemo.utils import logging
+from nemo.utils.get_rank import is_global_rank_zero
+
 try:
     from apex.transformer import parallel_state
 
@@ -48,11 +53,6 @@ except (ImportError, ModuleNotFoundError):
 
     HAVE_APEX = False
 
-
-from nemo.collections.nlp.data.language_modeling.megatron.blendable_dataset import BlendableDataset
-from nemo.collections.nlp.data.language_modeling.megatron.indexed_dataset import make_dataset as make_indexed_dataset
-from nemo.utils import logging
-from nemo.utils.get_rank import is_global_rank_zero
 
 DSET_TYPE_BERT = 'standard_bert'
 DSET_TYPE_ICT = 'ict'
--- a/nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
@@ -38,7 +38,16 @@ import time
 
 import numpy as np
 import torch
-from apex.transformer import parallel_state
+
+try:
+    from apex.transformer import parallel_state
+
+    HAVE_APEX = True
+
+except (ImportError, ModuleNotFoundError):
+
+    HAVE_APEX = False
+
 
 from nemo.collections.nlp.data.language_modeling.megatron.blendable_dataset import BlendableDataset
 from nemo.collections.nlp.data.language_modeling.megatron.indexed_dataset import make_dataset as make_indexed_dataset
--- a/nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
@@ -19,7 +19,16 @@ import time
 
 import numpy as np
 import torch
-from apex.transformer import parallel_state
+
+
+try:
+    from apex.transformer import parallel_state
+
+    HAVE_APEX = True
+
+except (ImportError, ModuleNotFoundError):
+
+    HAVE_APEX = False
 
 from nemo.collections.nlp.data.language_modeling.megatron.blendable_dataset import BlendableDataset
 from nemo.collections.nlp.data.language_modeling.megatron.dataset_utils import (
@@ -155,6 +164,10 @@ def get_indexed_dataset_(data_prefix, data_impl, skip_warmup):
 
 class GPTDataset(MegatronDataset):
     def __init__(self, cfg, trainer, name, data_prefix, documents, indexed_dataset, num_samples, seq_length, seed):
+        if not HAVE_APEX:
+            logging.warning(
+                "Apex was not found. Please see the NeMo README for installation instructions: https://github.com/NVIDIA/NeMo#megatron-gpt."
+            )
 
         super().__init__(cfg, trainer=trainer)
         self.name = name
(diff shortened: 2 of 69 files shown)


nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
```diff
@@ -38,7 +38,16 @@
...
-from apex.transformer import parallel_state
+
+try:
+    from apex.transformer import parallel_state
+
+    HAVE_APEX = True
+
+except (ImportError, ModuleNotFoundError):
+
+    HAVE_APEX = False
```
titu1994, line 54: Just a note, no need for so many new lines

titu1994: Also, anything that calls apex.*  Needs to first read this value of Have_APEX and then only use parallel_state. Otherwise if will fail with an import error (say in a class init)

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
```diff
@@ -155,6 +164,10 @@ def get_indexed_dataset_(data_prefix, data_impl, skip_warmup):
 
 class GPTDataset(MegatronDataset):
     def __init__(self, cfg, trainer, name, data_prefix, documents, indexed_dataset, num_samples, seq_length, seed):
+        if not HAVE_APEX:
```
titu1994, line 166: Ok so this is done here, what about dataset utils, does it not have functions that use parallel_stare?

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
```diff
@@ -155,6 +164,10 @@ def get_indexed_dataset_(data_prefix, data_impl, skip_warmup):
...
+            logging.warning(
+                "Apex was not found. Please see the NeMo README for installation instructions: https://github.com/NVIDIA/NeMo#megatron-gpt."
+            )
```
titu1994, line 169: Not a warning, raise an import error. Code below this won't work without apex anyway no ?

nemo/collections/nlp/models/language_modeling/megatron/bert_model.py
```diff
@@ -72,6 +81,11 @@ def __init__(
         self, mpu_vocab_size, hidden_size, init_method, layernorm_epsilon, parallel_output, use_openai_gelu, onnx_safe
     ):
 
+        if not HAVE_APEX:
+            logging.warning(
+                "Apex was not found. Please see the NeMo README for installation instructions: https://github.com/NVIDIA/NeMo#megatron-gpt."
```
titu1994, line 84: Raise error everywhere.

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
ericharper: it doesn't have `__init__`

nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
ericharper: This is insane we can't wrap every call to apex in an if statement. If they don't have apex, installed they'll get an appropriate error from python.

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
ericharper: I don't understand this, python will you when there's an import error.

nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
titu1994: Then you'll still get errors when your import failed but some class calls some function which depends on apex. Sure, you can leave if out for now but it will have issues downstream

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
titu1994: Ok. Hopefully anything using those functions will hard crash if apex is not found then.

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
titu1994: Python will give you a generic error without information of how to install apex. And the stack traces are large, so your comment about installing apex as a warning will not be seen unless someone actively looks for it. It is better to raise an import error explicitly with the exact message of how to install it. But it's upto you

nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
titu1994: Btw all the optional code does exactly this - check the import and then raise appropriate error for every single call in every single file. It is nothing special for apex but it's upto the NMT team to decide if they are ok with partial checks like this.

titu1994: Well python will raise an error anyway - so it's fine. But if there are dependent calls mixed into places where apex is actually optional, then the code may fail again. For now it's fine

itzsimpl:
Note also that checking for `apex.transformer` (first mentioned October 2, 2021 https://github.com/NVIDIA/apex/blob/365fdc180f02d56a92c943f72485f3514eafed9a/apex/transformer) may not be enough, since it exists longer than `apex.transformer.utils` (introduced October 27, 2021 https://github.com/NVIDIA/apex/blob/63d5dd638df56a230b35588050bda49365efa345/apex/transformer), and it may happen that the user has apex installed, but not the correct version. This is at least what I had issues with. See also  #3472, #3492, #3559.

ericharper:
> Note also that checking for `apex.transformer` (first mentioned October 2, 2021 https://github.com/NVIDIA/apex/blob/365fdc180f02d56a92c943f72485f3514eafed9a/apex/transformer) may not be enough, since it exists longer than `apex.transformer.utils` (introduced October 27, 2021 https://github.com/NVIDIA/apex/blob/63d5dd638df56a230b35588050bda49365efa345/apex/transformer), and it may happen that the user has apex installed, but not the correct version. This is at least what I had issues with. See also #3472, #3492, #3559.

Thanks for your reply and PR. This PR is going to patch r1.6.1. We're still considering options on how we want to guard these imports (including your #3472) for the main branch and future releases of NeMo.

nemo/collections/nlp/models/language_modeling/megatron/bert_model.py
ericharper: Done.

nemo/collections/nlp/data/language_modeling/megatron/gpt_dataset.py
ericharper: Done.

titu1994 (approved):
Looks good for now. Static checks may fail In the future for complicated dependent function calls, so we may need to revisit this with dynamic import checking in the future.

titu1994:
We will cherry pick this fix onto main though right @ericharper ?

ericharper:
> We will cherry pick this fix onto main though right @ericharper ?

That's correct.

PR #3605, opened 2022-02-04
Add NeMo version to all new .nemo files

titu1994:

# What does this PR do ?

Adds current nemo version to all nemo files generated. 

# Note

The version will point to the *earliest* known version of NeMo that was compatible with this model. This does not preclude later versions being compatible/incompatible with the checkpoint, so it is up to the user to determine which version would work best, but it can be assumed that no version prior to the noted version will support the checkpoint.

**Collection**: [Core]

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/core/classes/modelPT.py
```diff
@@ -109,6 +110,10 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             cfg.target = "{0}.{1}".format(self.__class__.__module__, self.__class__.__name__)
             OmegaConf.set_struct(cfg, True)
 
+        if 'nemo_version' not in cfg:
```
ryanleary, line 113: Will old versions of nemo complain about this living in the config? Would it be better to have it as a separate file in the archive?

titu1994: Interesting question. I tried a few tests of from_pretrained and it was just fine (thought the newer models had incorrectly attributed the current version as their origin aka old 1.3 model claimed it was written to in 1.7.0rc0). That.. can probably be ignored? Or how would I handle selectively adding and not adding this tag to older models is the question.

As to just injecting this into older configs, not an issue. Newer versions will just add it anyways since it doesn't exist for them too. However after a newer version adds this, gets saved as nemo, then upload to ngc, and a few months on a newer version pulls this model, it will correctly have the old version tag on it.

titu1994: I think of this as the earliest point the model was saved with save_to (after this change is merged).
So old model load -> save will point to "new" version.
new model load -> save will point to new version.
new model restored after a few months -> with/without save will point to now "old" version. Even saving again will not update this version.

okuchaiev:
looks good to me. @ryanleary please approve or ask for changes.

nemo/core/classes/modelPT.py
itzsimpl: Just a question. Let’s assume a model was initially created with say 1.4.0 then took as a starting point on the code base of NeMo:1.6.0; and assume that no breaking changes were introduced in the code base. Leaving the version tag as 1.4.0, will leave out the info that there may be differences that are introduced by changing to code base to 1.6.0, but swapping the tag to 1.6.0 will leave out the history. Assuming of course that all other things being equal (database and such) there may be a difference between a model started from scratch on 1.4.0 and one started from scratch on 1.6.0 that derives solely from the code base. 

What if instead the tag was potentially a list with the history of tags?

titu1994: The Nemo model stores only the checkpoint itself and any artifacts. The earliest version details when the checkpoint was created and therefore how the artifacts were stored at that point. Preserving a list of histories is inconsistent - what does it matter if it was saved 10 different versions, we only need to know what was the version that it first worked and currently works. Wrt this PR, it tackles the case - what was the first time it was saved as that was what was requested. We normally deprecate checkpoints that are no longer supported by Nemo code anyways so earliest version is more valuable than last version or current version that is supported - the user can open a container and run the model safely then.

titu1994: Not to mention the complexity of maintaining a history (even tuple of past and current) would require more flags and make it tedious to manage in Nemo. ModelPT is not a versioning system for Nemo models.

itzsimpl: Fair point, and I understand that on NGC the models will be constantly upgraded. I would just like to understand the meaning of the tag and how it is related to users that work on their own data. 

So, to be more specific. Let’s assume one is training an ASR model. To continue from a model with tag 1.4.0, the best way is to download a container 1.4.0 and do the additional training with it. Regardless if a new (say 1.8.0) container exists. Assuming they used their own data and would like to use the new container (e.g required for support in Riva) it is safer to start from scratch. Am I correct?

titu1994: Oh not at all. That's not the intended result of this tag. The reason for the tag is diagnosis in case the model can no longer be supported by Nemo, then users should know what was the version that did support the model. 

However, that is only if the latest version no longer supported it. We always recommend using the latest Nemo version with any checkpoint (old and new) since later version have significant bug fixes or improvements. Of course if something goes wrong, now users will have info of how far back to go to properly support the model.

titu1994: Also we almost never recommend starting from scratch. Any level of a pretraining (even other languages) is better than starting from scratch.

itzsimpl: Doesn't then leaving the tag equal to the original (starting) version then point them way too back? I mean assuming that one strarted off with 1.4.0, did some additional training with 1.5.0, and later discovered that with 1.7.0 it doesn't work anymore. The tag, will point them to 1.4.0, which is not the most recent version that still supported the model (that would be 1.5.0), but the oldest one, the one with which the initial model was created.

titu1994: Thats the point - its the *earliest* known compatible version. Its not ModelPTs requirement to keep track of what model breaks at what version, the user can choose to try later versions and if it works - great! But ModelPT will point the user to what's the earliest version where it is guaranteed to work. As mentioned above, keeping track of history is not a task ModelPT should handle.

itzsimpl: Ok. LGTM.

ryanleary:
Looks good. Thank you for the implementation and reviews.

ryanleary (approved):

titu1994:
Thanks all for reviews and discussion

PR #3614, opened 2022-02-05
Update numba if NVIDIA_PYTORCH_VERSION not correct

### LINKED ISSUE #3613: ./reinstall.sh crashes due to not being able to uninstall llvmlite
Starting off of `nemo:1.5.1` container, cloning the NeMo repo to a folder inside of it and calling `./reinstall.sh` fails with
```
ERROR: Cannot uninstall 'llvmlite'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```
`pip install -e` on the other hand succeeds installing `nemo:1.7.0rc0` and `numpy:1.22.2`, the rest of the packages remain untouched.

It seems that `./reinstall.sh` which used to work fine, a week or so ago when following the same procedure to upgrade to `nemo:1.6.0rc` redeveloped issue #841. The solution remains the same, first call
```
pip install --ignore-installed llvmlite
```
followed by `./reinstall.sh`. In this case, apart `llvml`, the following packages are updated
```
ftfy-6.0.3 nemo-toolkit-1.7.0rc0 numba-0.55.1 pytorch-lightning-1.5.9 sacrebleu-2.0.0 setuptools-59.5.0
```
Interestingly `numpy` in this case is left at `1.21.5`.

itzsimpl:
Update/install numba if `NVIDIA_PYTORCH_VERSION` is not equal to the one in `Dockerfile`. Resolves #3613  

### DIFF
--- a/reinstall.sh
+++ b/reinstall.sh
@@ -19,6 +19,18 @@ ${PIP} uninstall -y nemo_cv
 
 ${PIP} install -U setuptools
 
+if [ "${NVIDIA_PYTORCH_VERSION}" = "22.01" ]
+then
+  echo 'Installing NeMo in NVIDIA PyTorch container:' ${NVIDIA_PYTORCH_VERSION} 'so will not install numba'
+else
+  if [ -x "$(command -v conda)" ]
+  then
+    NUMBA_VERSION=0.55
+    echo 'Installing numba=='${NUMBA_VERSION}
+    conda install -y -c conda-forge numba==${NUMBA_VERSION}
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
I'm wondering if Numba should be commented out for the time being because of it's incompatibility with Nemo. Pushing it to the top just means it needs to be installed before Nemo itself now. Well, that is fine for now but I will revert this when the next Numba rc version is released.

PR #3626, opened 2022-02-08
Remove compression from .nemo files

okuchaiev:
# What does this PR do ?

This PR removes compression from .nemo files.
The compression is not desirable because:
 * It takes awfully long to compress multi-billion parameter language models
 * It does not actually save much space. For example a large CTC conformer uncompressed is 464Mb vs 431Mb compressed

This is a change to core save_to/restore_from functionality of all models.
**It is backwards compatible - older checkpoints (with compression) will still work**

# Usage
* Usage remains exactly the same.
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


titu1994 (approved):
Approved but it will have downstream effect. The try catch check will also waste a few seconds for large Conformers but that is unavoidable

ericharper (approved):
LGTM. Thanks!

PR #3673, opened 2022-02-15
Fixing the bug in the stateful rnnt decoder.

VahidooX:
# What does this PR do ?

Fixing the bug in the stateful rnnt decoder. The current code does not keep the last token properly from partial hypothesis to the next hypothesis. If The latest decoding does not produce any tokens, it would not propagate the non-blank last token and just passes blank token to the next hypothesis which is wrong.

# Changelog 
- Fixed the bug in the stateful rnnt decoder by keeping the last token and passing that along properly.

**PR Type**:
- [ ] New Feature
- [x ] Bugfix
- [ ] Documentation


titu1994 (dismissed):
Lgtm, thanks for the fix !

titu1994 (approved):

PR #3692, opened 2022-02-16
Megatron_LM checkpoint to NeMo checkpoint support 

yidong72:
# What does this PR do ?

Convert Megatron LM checkpoint to NeMo checkpoint. Copy over the weights, optimizer states, LR scheduler states.
Fix the issue of weight decay parameter group assignment. 
Tested that the script converts the old Megatron GPT and BERT checkpoint to `.nemo` file fine.

Tested with 40B checkpoint file, the NeMo Megatron can resume from it and continue the training. Need to verify the training curve works fine.


ericharper (dismissed):
LGTM. Thanks!

ericharper (changes requested):
Still testing if training is working with this PR.

ericharper (approved):
LGTM.

PR #3693, opened 2022-02-16
XNli benchmark

yidong72:
# What does this PR do ?

Added an experiment to run fine-tuning on the MNLI dataset, use XNLI dev as validation dataset. The score on the XNLI test is reported.


ericharper (approved):
LGTM.

PR #3700, opened 2022-02-17
Fixed Apex guard when imported classes are used for default values

michalivne:
1. Added ApexGuardDefaults class that provides None attributes when an Apex class is missing.
  2. Fixed Apex guard.

Signed-off-by: Micha Livne <mlivne@nvidia.com>

# What does this PR do ?

This PR fixes an Apex guard that leaves constructors with undefined constructor values.

# Changelog 
- Added `ApexGuardDefaults` class to `nemo/collections/nlp/modules/common/megatron/utils.py`
- Used `ApexGuardDefaults` to replace missing classes in `nemo/collections/nlp/modules/common/megatron/transformer.py`

### DIFF
--- a/examples/nlp/language_modeling/megatron_gpt_eval.py
+++ b/examples/nlp/language_modeling/megatron_gpt_eval.py
@@ -194,6 +194,8 @@ def main():
     print("***************************")
     print(response)
     print("***************************")
+    if args.prompt and not args.compute_logprobs:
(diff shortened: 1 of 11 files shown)


ericharper (commented):
LGTM. Thanks!

ericharper (approved):

ericharper (approved):
LGTM.

PR #3711, opened 2022-02-19
add pynini pip

yzhang123:

# What does this PR do ?
include pynini as default pip dependency, remove PYNINI_AVAILBLE guard and conda installation


titu1994 (commented):
I can't reasonably review every single file, it doesn't even load on GitHub. Please add the itn imports to the Check Import CI job and then let's merge if @ekmb approves

yzhang123:
> I can't reasonably review every single file, it doesn't even load on GitHub, but overall it looks current. Please add the itn imports to the Check Import CI job and then let's merge if @ekmb approves

update: as discussed we cannot merge this since this will break on macos

PR #3740, opened 2022-02-23
[TTS] Fix bugs in E2E TTS, Mixer-TTS and FastPitch

Oktai15:
# What does this PR do ?

**Collection**: [TTS]

# Changelog 
- Fix bugs in E2E TTS, Mixer-TTS and FastPitch 

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


nemo/collections/tts/models/fastspeech2_hifigan_e2e.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             cfg = OmegaConf.create(cfg)
         super().__init__(cfg=cfg, trainer=trainer)
 
-        self.audio_to_melspec_preprocessor = instantiate(cfg.preprocessor)
+        self.audio_to_melspec_precessor = instantiate(cfg.preprocessor)
```
blisc, line 57: Is it possible to put this in a try block?

Oktai15: We have this typo in several models. Let’s discuss how to resolve it out of this PR. In my opinion, code will be too hard with try block, because we can't check only config field, we would need to support two variants of attribute of class (because torch load weights function looks for concrete attribute of class).

blisc (approved):

PR #3742, opened 2022-02-23
Multi-Label Joint Intent Slot Classification

chenrichard10:
# What does this PR do ?

Multi-Label Intent Model for Joint Intent Slot Classification Task. 

This PR will involves changes in collections/nlp within NeMo.

# Changelog 
- New loss function added to collections/common/losses (BCEWithLogitsLoss)
- Multi-Label Classification Report added to collections/nlp/metrics/classification_report
- Multi-Label Classification Model, Descriptor, Dataset files added in similar folders to intent_slot_classification
- Data Conversion and Augmentation scripts added to example/nlp/multi_label_intent_slot_classification Multi-Label Training Tutorial added to https://github.com/NVIDIA/NeMo/blob/main/tutorials/nlp/Joint_Intent_and_Slot_Classification.ipynb
- Added Weight Calculation Function for BCEWithLogitsLoss (lines 512-537) and Multi Label Stats function (lines 201-242) in nlp/data/data_utils/data_preprocessing.py 

# Usage

Refer to https://github.com/NVIDIA/NeMo/blob/main/tutorials/nlp/Joint_Intent_and_Slot_Classification.ipynb for training. To run the tutorial you will need to change github string link name from NVIDIA to chenrichard10 if the PR has not been approved. 

A simple script that can be used to train the model locally can be with  
 
```python
from nemo.collections import nlp as nemo_nlp
from nemo.collections.nlp.parts.utils_funcs import tensor2list

import os
import time

import torch
import pytorch_lightning as pl

HOME_DIR=""

# directory with data converted to nemo format
data_dir = os.path.join(HOME_DIR, "new_multiatis")

run_name = "test"
config_file = ""
config = OmegaConf.load(config_file)
config.model.data_dir = data_dir
config.model.validation_ds.prefix = "dev"
config.model.test_ds.prefix = "test"
config.model.class_balancing = "weighted_loss"
config.model.head.num_output_layers = 2
config.trainer.max_epochs = 40


# checks if we have GPU available and uses it
cuda = 1 if torch.cuda.is_available() else 0
config.trainer.gpus = cuda

# config.trainer.precision = 16 if torch.cuda.is_available() else 32
config.trainer.precision = 32

# for mixed precision training, uncomment the line below (precision should be set to 16 and amp_level to O1):
# config.trainer.amp_level = 'O1'

# remove distributed training flags
config.trainer.accelerator = None

# early_stop_callback = EarlyStopping(monitor='intent_f1', min_delta=1e-1, patience=10, verbose=True, mode='max')

trainer = pl.Trainer(**config.trainer)
config.exp_manager.exp_dir = os.path.join(HOME_DIR, "output/" + run_name)
config.exp_manager.create_checkpoint_callback = True
config.exp_manager.version = time.strftime('%Y-%m-%d_%H-%M-%S')

exp_dir = exp_manager(trainer, config.get("exp_manager", None))
model = nemo_nlp.models.MultiLabelIntentSlotClassificationModel(config.model, trainer=trainer)
trainer.fit(model)
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
Available model has yet to be trained and uploaded on NGC. Section will be updated once a model has been trained.

### DIFF
--- a/examples/nlp/intent_slot_classification/conf/multi_label_intent_slot_classification_config.yaml
+++ b/examples/nlp/intent_slot_classification/conf/multi_label_intent_slot_classification_config.yaml
@@ -1,0 +1,110 @@
+# Intent and Slot classification with pretrained BERT models
+
+trainer:
+  gpus: 1 # the number of gpus, 0 for CPU
+  num_nodes: 1
+  max_epochs: 5
+  max_steps: null # precedence over max_epochs
+  accumulate_grad_batches: 1 # accumulates grads every k batches
+  precision: 32 # Should be set to 16 for O1 and O2 amp_level to enable the AMP.
+  accelerator: ddp
+  log_every_n_steps: 1  # Interval of logging.
+  val_check_interval: 1.0  # Set to 0.25 to check 4 times per epoch, or an int for number of iterations
+  resume_from_checkpoint: null # The path to a checkpoint file to continue the training, restores the whole state including the epoch, step, LR schedulers, apex, etc.
+
+  checkpoint_callback: false  # Provided by exp_manager
+  logger: false  # Provided by exp_manager
+
+model:
+  nemo_path: null # filename to save the model and associated artifacts to .nemo file
+  data_dir: ??? # /path/to/data
+  class_labels:
+    intent_labels_file: intent_labels.csv
+    slot_labels_file: slot_labels.csv
+  class_balancing: null # or weighted_loss
+  intent_loss_weight: 0.6 # relation of intent to slot loss in total loss (between 0 to 1)
+  pad_label: -1 # if -1 not slot token will be used
+  ignore_extra_tokens: false
+  ignore_start_end: true # do not use first and last token for slot training
+
+  train_ds:
+    prefix: train
+    batch_size: 32
+    shuffle: true
+    num_samples: -1
+    num_workers: 8
+    drop_last: false
+    pin_memory: false
+
+  validation_ds:
+    prefix: dev
+    batch_size: 32
+    shuffle: false
+    num_samples: -1
+    num_workers: 8
+    drop_last: false
+    pin_memory: false
+
+  test_ds:
+    prefix: dev
+    batch_size: 32
+    shuffle: false
+    num_samples: -1
+    num_workers: 8
+    drop_last: false
+    pin_memory: false
+
+  tokenizer:
+    tokenizer_name: ${model.language_model.pretrained_model_name} # or sentencepiece
+    vocab_file: null # path to vocab file
+    tokenizer_model: null # only used if tokenizer is sentencepiece
+    special_tokens: null
+
+  language_model:
+    max_seq_length: 50
+    pretrained_model_name: bert-base-uncased
+    lm_checkpoint: null
+    config_file: null # json file, precedence over config
+    config: null
+
+  head:
+    num_output_layers: 2
+    fc_dropout: 0.1
+
+  optim:
+    name: adam
+    lr: 2e-5
+    args:
+      name: auto
+      params:
+        weight_decay: 0.01
+
+    sched:
+      name: WarmupAnnealing
+      iters_per_batch: null # computed at runtime
+      max_steps: null # computed at runtime or explicitly set here
+
+      # pytorch lightning args
+      monitor: val_loss
+      reduce_on_plateau: false
+
+      # scheduler config override
+      args:
+        name: auto
+        params:
+          warmup_steps: null
+          warmup_ratio: 0.1
+          last_epoch: -1
+
+language_model:
+  max_seq_length: 50
+  pretrained_model_name: bert-base-uncased
+  lm_checkpoint: null
+  config_file: null # json file, precedence over config
+  config: null
+
+exp_manager:
+  exp_dir: null  # exp_dir for your experiment, if None, defaults to "./nemo_experiments"
+  name: "MultiLabelIntentSlot"  # The name of your model
+  create_tensorboard_logger: False  # Whether you want exp_manger to create a tb logger
+  create_checkpoint_callback: False  # Whether you want exp_manager to create a modelcheckpoint callback
--- a/nemo/collections/common/losses/bce_logits_loss.py
+++ b/nemo/collections/common/losses/bce_logits_loss.py
@@ -1,0 +1,79 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List
+
+import torch
+from torch import nn
+
+from nemo.core.classes import Serialization, Typing, typecheck
+from nemo.core.neural_types import LabelsType, LogitsType, LossType, MaskType, NeuralType
+
+__all__ = ["BCEWithLogitsLoss"]
+
+
+class BCEWithLogitsLoss(nn.BCEWithLogitsLoss, Serialization, Typing):
+    """
+    BCEWithLogitsLoss
+
+    https://pytorch.org/docs/1.9.1/generated/torch.nn.BCEWithLogitsLoss.html
+    """
+
+    @property
+    def input_types(self):
+        """Returns definitions of module input ports.
+        """
+        return {
+            "logits": NeuralType(["B"] + ["ANY"] * (self._logits_dim - 1), LogitsType()),
+            "labels": [NeuralType(["B"] + ["ANY"] * (self._logits_dim - 2), LabelsType())],
+            "loss_mask": NeuralType(["B"] + ["ANY"] * (self._logits_dim - 2), MaskType(), optional=True),
+        }
+
+    @property
+    def output_types(self):
+        """Returns definitions of module output ports.
+        """
+        return {"loss": NeuralType(elements_type=LossType())}
+
+    def __init__(
+        self,
+        logits_ndim: int = 2,
+        weight: torch.Tensor = None,
+        reduction: str = "mean",
+        pos_weight: torch.Tensor = None,
+    ):
+        """
+        Args:
+            logits_ndim: number of dimensions (or rank) of the logits tensor
+            weight: list of rescaling weight given to each class
+            reduction: type of the reduction over the batch
+            pos_weight: weight given to positive samples
+        """
+        if pos_weight is not None and not torch.is_tensor(pos_weight):
+            pos_weight = torch.FloatTensor(pos_weight)
+
+        super().__init__(weight=weight, pos_weight=pos_weight, reduction=reduction)
+        self._logits_dim = logits_ndim
+
+    @typecheck()
+    def forward(self, logits: float, labels: List[int], loss_mask: torch.Tensor = None):
+        """
+        Args:
+            logits: output of the classifier
+            labels: ground truth labels
+        """
+        labels = torch.stack(labels)
+        labels = labels.t().float()
+
+        return super().forward(logits, labels)
--- a/nemo/collections/nlp/data/data_utils/data_preprocessing.py
+++ b/nemo/collections/nlp/data/data_utils/data_preprocessing.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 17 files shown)


nemo/collections/nlp/models/multi_label_intent_slot_classification/multi_label_intent_slot_classification_model.py
```diff
@@ -0,0 +1,406 @@
...
+            parts = [1 if label in parts else 0 for label in range(len(self.cfg.data_desc.intent_labels))]
+            metrics_labels.append(parts)
+
+        # Retrieve class probabilities for each sentence
+        intent_probabilities = self.prediction_probabilities(sentences, test_ds)
+
+        metrics_dict = {}
+        # Find optimal logits rounding threshold for intents
+        for i in np.arange(0.5, 0.96, 0.01):
+            predictions = tensor2list((intent_probabilities >= i).int())
```
carolmanderson, line 289: This line will also need to be modified if you convert the torch tensor to a numpy array in `prediction_probabilities`. You can use numpy `.tolist()`

nemo/collections/nlp/models/multi_label_intent_slot_classification/multi_label_intent_slot_classification_model.py
```diff
@@ -0,0 +1,406 @@
...
+
+        logging.info(
+            f'Maximum Threshold for F1-Score: {max_f1_score}, [Precision, Recall, F1-Score]: {metrics_dict[max_f1_score]}'
+        )
+        logging.info(
+            f'Maximum Threshold for Precision: {max_precision}, [Precision, Recall, F1-Score]: {metrics_dict[max_precision]}'
+        )
+        logging.info(
+            f'Maximum Threshold for Recall: {max_recall}, [Precision, Recall, F1-Score]: {metrics_dict[max_recall]}'
+        )
```
carolmanderson, line 307: I found this logging output a little confusing. I think it would be clearer if it said "Best threshold" instead of "Maximum threshold".

nemo/collections/nlp/models/multi_label_intent_slot_classification/multi_label_intent_slot_classification_model.py
```diff
@@ -0,0 +1,406 @@
...
+
+        if metrics_dict[max_f1_score][2] > self.max_f1:
+            self.max_f1 = metrics_dict[max_f1_score][2]
+            self.threshold = max_f1_score
```
carolmanderson, line 311: Since you set a threshold here, I think you should also add a line of logging to tell the user that. Something like "Setting threshold to <whatever>"

tutorials/nlp/Joint_Intent_and_Slot_Classification.ipynb
```diff
@@ -556,17 +556,251 @@
...
+    "Before using our model, we find the probability rounding threshold that gives us the best results on the validation set. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "eval_model.optimize_threshold(config.model.test_ds)"
```
carolmanderson, line 767: After running `optimize_threshold`, I think it would be good to add a cell where you call `eval_model.threshold` so the user can see that the "best" threshold, based on F1 scores, was set as a model attribute. Also please explain that in the notebook.

examples/nlp/multi_label_intent_slot_classification/conf/multi_label_intent_slot_classification_config.yaml
```diff
@@ -0,0 +1,111 @@
+# Intent and Slot classification with pretrained BERT models
+
+trainer:
+  gpus: 2 # the number of gpus, 0 for CPU
```
carolmanderson, line 4: Set the default number of GPUs to 1 since most people won't have more than one. I don't think Colab is offering more than one either.

carolmanderson (changes requested):
Thanks for this valuable contribution to NeMo! I haven't yet reviewed all scripts, but there's plenty to work on based on my comments so far. 

To my mind it would make more sense to have the `multi_label_intent_slot_classification_model.py` sit inside the `models/intent_slot_classification` folder, instead of in its own folder. And similar for the dataset and data descriptor. I know we discussed this before, and maybe there was some good reason you decided not to do that.

tutorials/nlp/Joint_Intent_and_Slot_Classification.ipynb
```diff
@@ -555,6 +557,253 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Optimizing Threshold\n",
+    "\n",
+    "As mentioned above, when classifiying a given query such as `show all flights and fares from denver to san francisco`, our model checks whether each individual intent would be suitable. Before assigning the final labels for a query, the model assigns a probability an intent matches the query. For example, if our dict.intents had 5 different intents, then the model could output for a given query \\[0.52, 0.38, 0.21, 0.67. 0.80\\] where each value represents the probability that query matches that particular intent. \n",
+    "\n",
```
carolmanderson, line 755: Great explanation added here with the example probabilities.

nemo/collections/nlp/data/data_utils/data_preprocessing.py
```diff
@@ -461,9 +504,36 @@ def get_freq_weights(label_freq):
...
+
+    for label_id, class_values in label_freq.items():
+        positive_class = class_values[1]
+        negative_class = class_values[0]
+
+        if positive_class == 0:
+            weights[label_id] = 0
+
+        else:
+            weights[label_id] = float(negative_class) / float(positive_class)
```
carolmanderson, line 532: I think this should be negative/total   (i.e. negative/(negative + positive)

nemo/collections/nlp/data/data_utils/data_preprocessing.py
```diff
@@ -461,9 +504,36 @@ def get_freq_weights(label_freq):
     for lf in label_freq.values():
         total_size += lf
     weighted_slots = {label: (total_size / (len(label_freq) * freq)) for label, freq in label_freq.items()}
+
     return weighted_slots
 
 
+def get_freq_weights_bce_with_logits_loss(label_freq):
+    """
+    Calculate positive class weights in regards to BCEWithLogitsLoss
```
carolmanderson, line 513:
```suggestion
    Calculate positive class weights to be passed to BCEWithLogitsLoss
```

nemo/collections/common/losses/bce_logits_loss.py
```diff
@@ -0,0 +1,72 @@
...
+        return {"loss": NeuralType(elements_type=LossType())}
+
+    def __init__(
+        self, logits_ndim=2, weight=None, reduction="mean", ignore_index=-100, pos_weight=None,
+    ):
+        """
+        Args:
+            logits_ndim (int): number of dimensions (or rank) of the logits tensor
+            weight (list): list of rescaling weight given to each class
+            reduction (str): type of the reduction over the batch
```
carolmanderson, line 54: `weight` and `ignore_index` are currently not used. Does it make sense to have both `weight` and `pos_weight`?

nemo/collections/common/losses/bce_logits_loss.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        if pos_weight is not None and not torch.is_tensor(pos_weight):
+            pos_weight = torch.FloatTensor(pos_weight)
+        # Think about ignore index
```
carolmanderson, line 58: Funny inline comment....have you thought about it?

It looks like that argument isn't being used. Should it be in the `super` call?

examples/nlp/multi_label_intent_slot_classification/data/convert_datasets.py
```diff
@@ -0,0 +1,119 @@
...
+
+    multi_intent_df = pd.DataFrame(intent_labels)
+    multi_intent_df.to_csv(f'{target_dir}/{mode}.tsv', sep='\t', index=False)
+
+
+def convert_intent_dictionary(source_dir: str, target_dir: str) -> None:
+    """
+    Converts original intent dictionary containing labels that represented multiple labels into 
+    dictionary with only single labels. Example: if index 5 was referring to label "a+b", it is no longer 
+    a label in the new intent dictionary. Only labels "a" and "b" are included within the new dictionary
```
carolmanderson, line 75: Nice explanation

nemo/collections/common/losses/bce_logits_loss.py
chenrichard10: BCEWithLogitsLoss does not have a ignore index argument,  so I will remove that

nemo/collections/common/losses/bce_logits_loss.py
chenrichard10: The original BCEWithLogitsLoss seems to have weight and pos_weight parameters (https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html) I think it would be good to have both and pass them in super().

nemo/collections/nlp/data/data_utils/data_preprocessing.py
chenrichard10: I think from https://pytorch.org/docs/1.9.1/generated/torch.nn.BCEWithLogitsLoss.html it says that the pos_weight for a class should be negative / positive example (The paragraph right above "Examples:"

nemo/collections/nlp/models/intent_slot_classification/multi_label_intent_slot_classification_model.py
```diff
@@ -0,0 +1,444 @@
...
+        )
+
+        if metrics_dict[max_f1_score][2] > self.max_f1:
+            self.max_f1 = metrics_dict[max_f1_score][2]
+
+            logging.info(f'Setting Threshold to: {max_f1_score}')
+
+            self.threshold = max_f1_score
+
+    def predict_from_examples(
```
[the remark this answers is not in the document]
chenrichard10: I don't think you can set self.threshold as a default argument, so I set it as 0.5 at the moment.

tutorials/nlp/Joint_Intent_and_Slot_Classification.ipynb
```diff
@@ -555,6 +557,253 @@
...
+   "metadata": {},
+   "source": [
+    "## Evaluation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To see how the model performs, we can evaluate the performance of the trained model on a test data file. Here we would load the best checkpoint (the one with the lowest validation loss) and create a model (eval_model) from the checkpoint. We will use the same trainer for testing. We will need to use the file specified in the output during the training process."
```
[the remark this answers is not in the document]
carolmanderson: I hadn't noticed this before, but there are actually two locations in the notebook where this needed to be changed -- above for the single-label model, and here for the multi-label model.

nemo/collections/nlp/models/intent_slot_classification/multi_label_intent_slot_classification_model.py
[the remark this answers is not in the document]
carolmanderson: Ah yeah, you're right of course. The usual workaround for this situation is to set the default as None and then replace it with the object attribute inside the method. I think this will be better than using 0.5, which might be unexpected if a user has just gone through the threshold finding and thinks the threshold is now going to be used.

examples/nlp/intent_slot_classification/conf/multi_label_intent_slot_classification_config.yaml
```diff
@@ -0,0 +1,110 @@
...
+    prefix: dev
+    batch_size: 32
+    shuffle: false
+    num_samples: -1
+    num_workers: 8
+    drop_last: false
+    pin_memory: false
+
+  test_ds:
+    prefix: test
```
carolmanderson, line 49: Should this be changed to "dev" so that it will work easily with the multi-atis dataset? Alternatively you can pass this in as a command line arg in your docstring example.

nemo/collections/nlp/data/data_utils/data_preprocessing.py
carolmanderson: Ah yes, you're right, of course.

carolmanderson (changes requested):
Almost done! I reopened a couple of "resolved" conversations because the changes didn't get committed. Added a few more comments/suggestions too, but this is looking good.

nemo/collections/nlp/data/intent_slot_classification/multi_label_intent_slot_classification_dataset.py
```diff
@@ -0,0 +1,120 @@
...
+    and slot classification with pretrained model.
+
+    Converts from raw data to an instance that can be used by
+    NMDataLayer.
+
+    Args:
+        input_file: file containing sentences + labels. The first line is header (sentence [tab] label)
+            each line should be [sentence][tab][label] where label can be multiple labels separated by a comma
+        slot_file: file containing slot labels, each line corresponding to slot labels for a sentence in input_file. No header.
+        max_seq_length: max sequence length minus 2 for [CLS] and [SEP]
```
carolmanderson, line 39: num_intents should be included in the args

nemo/collections/nlp/metrics/classification_report.py
```diff
@@ -105,10 +115,11 @@ def update(self, predictions: torch.Tensor, labels: torch.Tensor):
         self.fp += fp
         self.num_examples_per_class += num_examples_per_class
 
-    def compute(self):
+    def compute(self) -> None:
```
carolmanderson, line 118: I don't think this should be None since there are return values.

carolmanderson (changes requested):
Looks great! I've just requested two tiny changes.

PR #3743, opened 2022-02-23
Mirror AN4 while CMU servers are down

titu1994:
# What does this PR do ?

Add mirror to AN4 dataset while CMU servers are down.

**Collection**: [ASR, Speaker Tasks]


okuchaiev (approved):
lgtm

PR #3828, opened 2022-03-11
Offline VAD+ASR tutorial

fayejf:
# What does this PR do ?

This PR adds functions and a tutorial about how to perform offline VAD before ASR.
Pipeline: audio -> [VAD] -> speech segments -> [ASR] -> segmented ASR output -> stitched ASR output

**Collection**:  ASR

# Changelog 
- Transcribe partial audio (for example speech segments) with manifest.
- Stitch segmented ASR output.
- Construct the manifest that aligns stitched ASR outputs and ground truth text. 
- And a walkthrough tutorial for the whole process. 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/asr/transcribe_speech.py
```diff
@@ -151,10 +151,18 @@ def main(cfg: TranscriptionConfig) -> TranscriptionConfig:
     else:
         # get filenames from manifest
         filepaths = []
+        if os.stat(cfg.dataset_manifest).st_size == 0:
+            logging.warning(f"The input dataset_manifest {cfg.dataset_manifest} is empty. Exiting!")
```
titu1994, line 155: Logging.error is better here. Also, cool check to test for empty manifest.

titu1994 (changes requested):
This is not a good change. the .transcribe() method will only work on python string paths. It is not meant for manifest handling and setting up of lengths etc. That's the point of the transcribe_speech.py script - to provide a higher level API over the transcribe method (supporting manifest etc). 

Please revert these changes to transcribe - especially the requirement of num files etc. Such things are not meant for transcribe. They are meant for higher level script to work on transcribe.

examples/asr/transcribe_speech.py
```diff
@@ -93,6 +94,86 @@ class TranscriptionConfig:
     rnnt_decoding: RNNTDecodingConfig = RNNTDecodingConfig(fused_batch_size=-1)
 
 
+def transcribe_partial_audio(
```
titu1994, line 97: Where's the equivalent for RNNT ? If it's not supported, raise an error please. You can do this by asserting that obj is instance of only CTC classes.

examples/asr/transcribe_speech.py
```diff
@@ -151,10 +232,22 @@ def main(cfg: TranscriptionConfig) -> TranscriptionConfig:
...
         filepaths = []
+        if os.stat(cfg.dataset_manifest).st_size == 0:
+            logging.error(f"The input dataset_manifest {cfg.dataset_manifest} is empty. Exiting!")
+            return None
+
         with open(cfg.dataset_manifest, 'r') as f:
+            partial_audio = True
             for line in f:
                 item = json.loads(line)
+                if "duration" not in item:
```
titu1994, line 243: duration is not the key factor determining limit on audio - it is both duration + offset presence that determines that partial audio is chosen. 

Duration is filled in for nearly all manifests always.

examples/asr/transcribe_speech.py
```diff
@@ -151,10 +232,22 @@ def main(cfg: TranscriptionConfig) -> TranscriptionConfig:
...
+                    partial_audio = False
                 filepaths.append(item['audio_filepath'])
+        if not partial_audio:
```
titu1994, line 246: Do not raise this warning - by default most manifest do not have offset set. We don't want to warm for a general case

examples/asr/transcribe_speech.py
titu1994, line 97: These functions are quite large, better to move the into some util dir and import only here.

nemo/collections/asr/models/ctc_models.py
```diff
@@ -691,9 +691,15 @@ def _setup_transcribe_dataloader(self, config: Dict) -> 'torch.utils.data.DataLo
         Returns:
             A pytorch DataLoader for the given audio file(s).
         """
-        batch_size = min(config['batch_size'], len(config['paths2audio_files']))
+        if 'manifest_filepath' in config:
```
titu1994, line 694: What about RNNT overrides ? Add the support for it's two methods, even if it's not used yet. All transcribe related info should ideally be synchronized among all model implementations.

titu1994 (changes requested):

titu1994:
The notebook needs some grammar fixes + @tango4j  for review of the VAD code (it looks correct to me but second checks). 

Also, the method does not work for RNNT - the notebook needs to be renamed to ASR with VAD for CTC models (or some such equivalent title) + mention in the notebook that this is only for CTC at the moment. 
Also add this notebook link to the ASR tutorials readme https://github.com/NVIDIA/NeMo/tree/main/tutorials/asr under it's own section (ASR with VAD).

examples/asr/transcribe_speech.py
fayejf: Cannot find a good existing place. How about having a transcribe_utils.py so we can put everything about transcribe there.

nemo/collections/asr/models/ctc_models.py
fayejf: good point. I totally forgot rnnt. :(

fayejf:
> Also, the method does not work for RNNT - the notebook needs to be renamed to ASR with VAD for CTC models (or some such equivalent title) + mention in the notebook that this is only for CTC at the moment. Also add this notebook link to the ASR tutorials readme https://github.com/NVIDIA/NeMo/tree/main/tutorials/asr under it's own section (ASR with VAD).

I added it to readme. [11] Is there any particular order for notebooks?
Oh, good point. it's should be Offline_ASR_with_VAD_for_CTC_models for now.

tango4j (changes requested):

nemo/collections/asr/parts/utils/vad_utils.py
```diff
@@ -881,3 +881,121 @@ def init_vad_model(model_path):
...
+) -> str:
+    """
+    Stitch the prediction of speech segments.
+    """
+
+    if not stitched_output_manifest:
+        stitched_output_manifest = "asr_stitched_output_manifest.json"
+
+    if not speech_segments_tensor_dir:
+        speech_segments_tensor_dir = "speech_segments"
```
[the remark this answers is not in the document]
fayejf: good point!

nemo/collections/asr/models/ctc_models.py
fayejf: added support for rnnt and rnnt bpe

fayejf:
@titu1994 @tango4j  Thanks for the careful review and feedback! I've addressed all the feedback above and did a grammar check again. Please review again when you have time. :)

docs/source/starthere/tutorials.rst
```diff
@@ -88,6 +88,9 @@ To run a tutorial:
    * - ASR
      - Buffered Transducer inference with LCS Merge Algorithm
      - `Buffered Transducer inference with LCS Merge <https://colab.research.google.com/github/NVIDIA/NeMo/blob/stable/tutorials/asr/Buffered_Transducer_Inference_with_LCS_Merge.ipynb>`_
+   * - ASR
+     - Offline ASR with VAD for CTC models
+     - `Offline ASR with VAD for CTC models <https://github.com/NVIDIA/NeMo/blob/stable/tutorials/asr/Offline_ASR_with_VAD_for_CTC_models.ipynb>`_
```
titu1994, line 93: Shouldn't this be a colab link ?

examples/asr/transcribe_speech.py
```diff
@@ -151,10 +152,18 @@ def main(cfg: TranscriptionConfig) -> TranscriptionConfig:
...
         filepaths = []
+        if os.stat(cfg.dataset_manifest).st_size == 0:
+            logging.error(f"The input dataset_manifest {cfg.dataset_manifest} is empty. Exiting!")
+            return None
+
         with open(cfg.dataset_manifest, 'r') as f:
+            partial_audio = True
             for line in f:
                 item = json.loads(line)
+                if "offset" not in item or "duration" not in item:
```
titu1994, line 163: This should be an if x in manifesf and y in manifest: situation, not an or. Most manifests will have duration but not offset.

tango4j (approved):

nemo/collections/asr/models/ctc_models.py
titu1994: Cool.

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -0,0 +1,104 @@
...
+    asr_model,
+    path2manifest: str,
+    batch_size: int = 4,
+    logprobs: bool = False,
+    return_hypotheses: bool = False,
+    num_workers: int = 0,
+) -> List[str]:
+
+    assert (
+        type(asr_model) == EncDecCTCModelBPE or type(asr_model) == EncDecCTCModel
```
titu1994, line 33: You can do is instance(asr_model, EncDecCTCModel) - subclassing should work out of the box.

examples/asr/transcribe_speech.py
```diff
@@ -187,15 +196,25 @@ def autocast():
...
         with torch.no_grad():
-            transcriptions = asr_model.transcribe(filepaths, batch_size=cfg.batch_size, num_workers=cfg.num_workers)
+            if partial_audio:
+                transcriptions = transcribe_partial_audio(
+                    asr_model=asr_model,
+                    path2manifest=cfg.dataset_manifest,
+                    batch_size=cfg.batch_size,
+                    num_workers=cfg.num_workers,
+                )
+            else:
```
titu1994, line 206: Hmm we don't want to disallow users who have offset and duration in manifest from running transcribe (at least whole transcribe can work).  

Do one thing. Check for CTC here then if partial audio send to transcribe partial. If not CTC and still partial audio flag is set - log a warning message saying that RNNT models do not support partial audio transcription (for now) and instead perform the else part.

tutorials/asr/README.md
```diff
@@ -28,6 +28,8 @@ In this repository, you will find several tutorials discussing what is Automatic
 
 10) `Self_Supervised_Pre_Training`: It can often be difficult to obtain labeled data for ASR training. In this tutorial, we demonstrate how to pre-train a speech model in an unsupervised manner, and then fine-tune with CTC loss.
 
+11) `Offline_ASR_with_VAD_for_CTC_models`: In this tutorial, we will demonstrate how to use offline VAD to extract speech segments and transcribe the speech segments with CTC models. This will help to exclude some non_speech utterances and could save computation resources by removing unnecessary input to the ASR system.
```
titu1994, line 31: Create another section called ASR with VAD instead of adding to current list. Then the table of content can quickly jump to it. We will add future RNNT too so might as well.

titu1994 (commented):
Looks better. Minor modifications left.

tutorials/asr/README.md
fayejf: Actually, I don't think we need to add another section.  RNNT will go to this tutorial by removing CTC models. And streaming vad asr will go to 

> Streaming / Buffered Automatic Speech Recognition: Streaming_ASR_with_VAD_for_CTC_models

docs/source/starthere/tutorials.rst
fayejf: good catch!

examples/asr/transcribe_speech.py
fayejf: Not really. if anyone is missing, then partial_audio=False and we will use `audio_filepath` and `transcribe` method

fayejf:
@titu1994 Replied and updated with review comments. Also, make necessary changes for smooth running on colab. Note the new link might be broken due to the 'stable' issue. 
We probably need to revert stable to main for main branch after each release.

examples/asr/transcribe_speech.py
titu1994: We generally have the 3 values in the manifest - audio_filepath. text and duration. Here the or for "offset" will fail and set partial_audio = False which is fine.
The logic should be - partial audio is false by default, but if both args are provided, then set to true - its the inverse of this.

titu1994 (dismissed):
Overall looks good. @tango4j  for final review before merge.

tango4j (approved):

PR #3858, opened 2022-03-18
[DRAFT]`generate` API proposal

yidong72:
Let's review the proposed `generate` API for models that can generate text. 
Note, it is for discussion. **DON'T** approve and merge it.

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -22,6 +22,7 @@ from omegaconf.dictconfig import DictConfig
 from omegaconf.omegaconf import open_dict
 from pytorch_lightning.plugins.precision.native_amp import NativeMixedPrecisionPlugin
 from pytorch_lightning.trainer.trainer import Trainer
+from torch import Tensor
 
 from nemo.collections.nlp.data.language_modeling.megatron.gpt_dataset import build_train_valid_test_datasets
 from nemo.collections.nlp.data.language_modeling.megatron.gpt_prompt_tuning_dataset import GPTPromptTuningDataset
@@ -39,6 +40,7 @@ from nemo.collections.nlp.modules.common.megatron.utils import (
     get_ltor_masks_and_position_ids,
 )
 from nemo.collections.nlp.modules.common.tokenizer_utils import get_nmt_tokenizer
+from nemo.collections.nlp.modules.common.transformer.text_generable_model import TextGenerableModelInterface
 from nemo.collections.nlp.parts.nlp_overrides import GradScaler
 from nemo.collections.nlp.parts.utils_funcs import get_last_rank
 from nemo.core.optim import MainParamsOptimizerWrapper, prepare_lr_scheduler
@@ -90,7 +92,7 @@ def _get_params_for_weight_decay_optimization(
     return weight_decay_params, no_weight_decay_params
 
 
-class MegatronGPTModel(NLPModel):
+class MegatronGPTModel(NLPModel, TextGenerableModelInterface):
     """
     Megatron GPT pretraining and prompt tuning
     """
@@ -874,6 +876,10 @@ class MegatronGPTModel(NLPModel):
             for param in param_group['params']:
                 params.append(param)
         return params
+
+    def generate(self, inputs: Union[List[str], Tensor, List[dict]], sampling_params: SamplingParam) -> OutputType:
+        # no implementation
+        return super().generate(inputs, sampling_params)
 
     def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: Optional[int] = None) -> Any:
         request, positions, tokens_to_generate, compute_logprobs = MegatronGPTModel._bucketize_gpt_inference(
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -40,7 +40,7 @@ from nemo.collections.nlp.modules.common.megatron.utils import (
     get_ltor_masks_and_position_ids,
 )
 from nemo.collections.nlp.modules.common.tokenizer_utils import get_nmt_tokenizer
-from nemo.collections.nlp.modules.common.transformer.text_generable_model import TextGenerableModelInterface
+from nemo.collections.nlp.modules.common.transformer.text_generable_model import OutputType, SamplingParam, TextGeneration
 from nemo.collections.nlp.parts.nlp_overrides import GradScaler
 from nemo.collections.nlp.parts.utils_funcs import get_last_rank
 from nemo.core.optim import MainParamsOptimizerWrapper, prepare_lr_scheduler
@@ -92,7 +92,7 @@ def _get_params_for_weight_decay_optimization(
     return weight_decay_params, no_weight_decay_params
 
 
-class MegatronGPTModel(NLPModel, TextGenerableModelInterface):
+class MegatronGPTModel(NLPModel, TextGeneration):
     """
     Megatron GPT pretraining and prompt tuning
     """
--- a/nemo/collections/nlp/modules/common/transformer/text_generable_model.py
+++ b/nemo/collections/nlp/modules/common/transformer/text_generable_model.py
@@ -1,0 +1,59 @@
+from typing import List, TypedDict, Union
+
+from torch import Tensor
+
+
+class SamplingParam(TypedDict):
+    max_length: int  # The maximum length of the sequence to be generated.
+    min_length: int  # The minimum length of the sequence to be generated.
+    do_sample: bool  # Whether or not to use sampling ; use greedy decoding otherwise
+    top_k: int  # The number of highest probability vocabulary tokens to keep for top-k-filtering.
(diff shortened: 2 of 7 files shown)


nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class OutputType(TypedDict):
+    sentences: List[str]  # output sentences
+    tokens: List[List[str]]  # output sentences borken into tokens
+    logits: List[Tensor]  # logits output for all the tokens
+    token_ids: List[Tensor]  # output sentence token ids
+
+
+class TextGenerableModelInterface:
```
okuchaiev, line 23: Class naming conventions were described in our internal dev guide but I moved it into https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md so that anyone can see it. 

Let's rename this class to something like ``TextGeneration``

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Interface for all models that is able to generate text.
```
okuchaiev, line 25: grammar?

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,59 @@
...
+
+    """
+
+    def generate(self, inputs: Union[List[str], Tensor, List[dict]], sampling_params: SamplingParam) -> OutputType:
+        """
+        Public method for all models that is able to generate text.
+        @params
```
okuchaiev, line 32: wrong docstring format, it won't be picked up by Sphynx for autodoc generation. Can you please follow this kind of format for method docstrings: https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/main/_modules/nemo/core/classes/modelPT.html#ModelPT.register_artifact

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,59 @@
...
+            inputs: Union[List[str], Tensor, List[dict]]
+                Can be one of the 3 types: 
+                1. List of strings. Each element of the list provides input prompt. The model will apply tokenizer on it.
+                    E.g [‘sentence’, ‘sentence2’ … ]
+                2. Pytorch Tensor of shape (batch_size, seq_length).  The sequence used as a prompt for the generation or as model inputs to the encoder. It will skip the tokenization step.
```
okuchaiev, line 37: How to pad in case 2? (with which token?)

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,59 @@
...
+                    E.g. torch.tensor([[23,5234,23,35…], [223,323,23,23232,232] …])    
+                3. List of python dict objects. Used for prompt/p-tuning inputs where a set of key-value pairs are converted into input token embeddings for the model.
```
okuchaiev, line 39: In prompt/p-tuning case are we only going to support strings? What about tokenized and/or tensor modes?

yidong72: For Prompt/p-tuning model, once the model is trained, the embeddings for the virtual tokens are fixed. We can add virtual tokens to the vocab and expand the embedding table to include those virtual token embeddings which are generated from the prompt encoder.  This is the step that is needed for preparing the model for fastertransformer so it has the best compatibility with GPT models.

Once we finished the virtual tokens vocab expansion step, we can simply use the Tensor type token id input to generate text.

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
yidong72: changed

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
yidong72: fixed

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
yidong72: fixed

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
yidong72: we assume the input tokens are already padded since it expects a single tensor.

okuchaiev:
what about .complete method? It is now redundant and should be gone, right?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -875,6 +877,10 @@ def get_parameters(self):
                 params.append(param)
         return params
 
+    def generate(self, inputs: Union[List[str], Tensor, List[dict]], sampling_params: SamplingParam) -> OutputType:
+        # no implementation
+        return super().generate(inputs, sampling_params)
+
     def predict_step(self, batch: Any, batch_idx: int, dataloader_idx: Optional[int] = None) -> Any:
```
okuchaiev, line 884: I'm wondering how will we maximize the code re-use between predict/eval/test_step and .generate?
Will you be relying on ._bucketize_gpt_inference under the hood of ".generate"?

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,60 @@
+from builtins import NotImplementedError, bool, int, float, str
+from typing import List, TypedDict, Union
+
+from torch import Tensor
+
+
+class SamplingParam(TypedDict):
+    max_length: int  # The maximum length of the sequence to be generated.
```
okuchaiev, line 8: should we add an option to stop if "." was generated?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
yidong72: I would do the opposite. The `.generate` calls `trainer.predict` as we need it to initialize the DDPPlugin. So `predict_step` and `bucket` For `eval` and `test`, it just evaluate the LM loss which is different from text generation.  
` ._bucketize_gpt_inference` will not be used as the `text_generation_util` padded the tokens into the same length.

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
okuchaiev: @vadam5  what do you think?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
okuchaiev: what if Model was created without passing the trainer instance (non-model parallel)?

ericharper: Even non-model parallel models require Trainer.

okuchaiev: @ericharper I think it can be None for non-model parallel stuff, isn't it https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/nlp_model.py#L48 ?

ericharper: for NLPModels, but megatron layers require ddp

ericharper: https://github.com/NVIDIA/NeMo/blob/c4ec081c2a6d718f8917155d55a006b5f6fcbb27/nemo/collections/nlp/modules/common/megatron/transformer.py#L689

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
ericharper: i think compute_logprobs should be replaced by generate(compute_logprobs=True) or something like that

okuchaiev:
"[text_generable_model.py]" - typo in filename

nemo/collections/nlp/modules/common/transformer/text_generable_model.py
```diff
@@ -0,0 +1,60 @@
...
+    token_ids: List[Tensor]  # output sentence token ids
+
+
+class TextGeneration:
+    """
+    Interface for all text generation models.
+    """
+
+    def generate(self, inputs: Union[List[str], Tensor, List[dict]], sampling_params: SamplingParam) -> OutputType:
+        """
```
okuchaiev, line 30: can we set "sampling_params=None" by default

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -875,6 +877,10 @@ def get_parameters(self):
                 params.append(param)
         return params
 
+    def generate(self, inputs: Union[List[str], Tensor, List[dict]], sampling_params: SamplingParam) -> OutputType:
```
vadam5, line 880: Could the default be set to None in the method signature here?

PR #3877, opened 2022-03-23
Continuous prompt refactor

vadam5:
# What does this PR do ?

Refactor to combine p-tuning and prompt tuning  code as well as integrate it with the new evaluation code/api. 
Prompt tuning and p-tuning are now jointly called 'prompt learning' within NeMo. 

**Collection**: BigNLP

# Changelog 
- Added MegatronGPTPromptLearningModel class
- Added GPTPromptLearningDataset class, merging p-tuning and prompt-tuning dataset classes
- Removed p-tuning dataset and model classes
- Removed prompt tuning code from the MegatronGPTModel, GPTModel, and TransformerLanguageModel classes
- Edited inference/generation code to accommodate p-tuning/prompt-tuning 

# Usage

```python
python megatron_gpt_prompt_learning.py --config-name=megatron_gpt_prompt_learning_config.yaml
python megatron_gpt_eval.py --config-name=megatron_gpt_inference.yaml
```

**Pre checks**:
- [ ] Did you add or update any necessary documentation? (**Will come in a different PR**)
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/nlp/language_modeling/megatron_gpt_ptune.py
```diff
@@ -1,135 +0,0 @@
...
-            # inject model parallel rank
-            checkpoint_path = os.path.join(checkpoint_folder, f'mp_rank_{rank:02d}', checkpoint_name)
-        else:
-            checkpoint_path = os.path.join(checkpoint_folder, checkpoint_name)
-
-        # Load the checkpoint
-        best_eval_model = MegatronGPTPTuneModel.load_from_checkpoint(
-            checkpoint_path=checkpoint_path, strict=False, trainer=trainer
-        )
-        logging.info(f'Best checkpoint path: {checkpoint_path}')
```
yidong72, line 101: Shall we add the logics of 

1. find the best checkpoint 
2.  convert model to nemo file

To the `megatron_gpt_prompt_tuning.py` file?

nemo/collections/nlp/data/language_modeling/megatron/gpt_prompt_learning_dataset.py
```diff
@@ -0,0 +1,237 @@
...
+            ), "The number of '<|VIRTUAL_PROMPT_n|>' markers and the number of prompt token splits must match"
+
+            # Format the input example according to the template
+            for field in prompt_template_fields:
+                if field in doc.keys():
+                    field_text = doc[field]
+                    input_example = input_example.replace('{' + field + '}', field_text)
+
+                # If some fields from the template aren't present, e.g. {answer} during inference
+                # just remove that field from the template, leaving the space blank
```
yidong72, line 121: This is a good to assume `answer` is the label. Maybe show some error if the user accidently enter the wrong `key` word.

nemo/collections/nlp/data/language_modeling/megatron/gpt_prompt_learning_dataset.py
```diff
@@ -0,0 +1,237 @@
...
+
+            input_ids = self.tokenizer.text_to_ids(input_example)
+
+            # Add BOS/EOS if desired, adds EOS by default
+            if self.add_bos:
+                input_ids = [self.tokenizer.bos_id] + input_ids
+            if self.add_eos:
+                input_ids = input_ids + [self.tokenizer.eos_id]
+
+            # Skip example if the final length doesn't fit length requirements
```
yidong72, line 143: can we have the option to truncate some of the input fields instead of skipping it? Otherwise, it might throw away a lot of data records.

nemo/collections/nlp/modules/common/transformer/text_generation.py
```diff
@@ -58,7 +58,7 @@ def generate(
         inputs: Union[List[str], Tuple[Tensor, Tensor], List[dict]],
         length_params: LengthParam,
         sampling_params: SamplingParam = None,
-    ) -> OutputType:
+    ):
```
yidong72, line 61: Why removed the output type?

nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -444,11 +545,19 @@ def switch(val1, val2, boolean):
 
 def forward_step(model, batch, tensor_shape):
 
+    # Should call MegatronGPTPPromptLearningModel's forward method
+    if hasattr(model, 'prompt_table'):
+        forward_model = model
```
yidong72, line 550: Nice trick to make prompt model work in the generation code.
Instead of test it has attribute. It is less hacky to do type check
```python
if isinstance(model,  MegatronGPTPPromptLearningModel):
```

nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -533,8 +643,14 @@ def sample_sequence_batch(
                 [set_inference_key_value_memory] * micro_batch_size, device=torch.cuda.current_device()
             )
             len_array = torch.tensor([maxlen] * micro_batch_size, device=torch.cuda.current_device())
-            batch = [tokens2use, attention_mask_repeat, positions2use, setkey_value_array, len_array]
-            tensor_shape = [tokens2use.shape[1], micro_batch_size, model.cfg.hidden_size]
+
+            # Only prompt learning models will have a prompt table, and require task ids
+            if hasattr(model, 'prompt_table'):
```
yidong72, line 648: same comment as above. Do instance type check

nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -553,7 +669,7 @@ def sample_sequence_batch(
                 # make sure it won't sample outside the vocab_size range
                 logits[:, tokenizer.vocab_size :] = -float('Inf')
```
yidong72, line 670: I have this guard to make sure it won't sample outside the `vocab_size`. I see you also do `Replace special soft prompt token ids with unk token ids` later. Isn't it better to make sure the model never samples those special soft prompt tokens?

examples/nlp/language_modeling/megatron_gpt_ptune.py
```diff
@@ -1,135 +0,0 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
-#
```
yidong72, line 2: Since `megatron_gpt_ptune.py` is removed completely. Could you make sure the notebook `https://github.com/NVIDIA/NeMo/blob/main/tutorials/nlp/PTune_multiple_NLP_tasks.ipynb` is still working? It needs to be updated.

yidong72 (commented):
Nice PR overall. I left some comments. 

Since megatron_gpt_ptune.py is removed completely. Could you make sure the notebook https://github.com/NVIDIA/NeMo/blob/main/tutorials/nlp/PTune_multiple_NLP_tasks.ipynb is still working? It needs to be updated.

nemo/collections/nlp/modules/common/prompt_encoder.py
```diff
@@ -73,13 +73,13 @@ def __init__(self, template: List[int], hidden_size: int, lstm_dropout: float, n
         )
 
     @typecheck()
-    def forward(self, enc_taskname) -> torch.Tensor:
-        input_embeds = self.embedding(self.seq_indices).unsqueeze(0)
-        if enc_taskname is not None:
```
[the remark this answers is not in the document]
vadam5: Yeah, I'm assuming a task name always needs to be given because after p-tuning is over, the virtual token embeddings will be moved to the prompt table, and the prompt table needs a task name to look up the correct embeddings.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -0,0 +1,653 @@
...
+                    super().__init__()
+                    self.sentences = sentences
+
+                def __len__(self):
+                    return len(self.sentences)
+
+                def __getitem__(self, idx):
+                    return self.sentences[idx]
+
+            # TODO, this is a little hacky. need to handle this nicely in the future
```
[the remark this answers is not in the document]
vadam5: Yes, I'll try this way instead.

nemo/collections/nlp/modules/common/transformer/text_generation.py
vadam5: That was a mistake, I'll add it back

examples/nlp/language_modeling/megatron_gpt_ptune.py
vadam5: Yeah, I think that will be good for external users pulling models from NGC. I'll add it.

nemo/collections/nlp/data/language_modeling/megatron/gpt_prompt_learning_dataset.py
vadam5: Sure, I'll add back your truncation logic

nemo/collections/nlp/modules/common/text_generation_utils.py
vadam5: I originally wanted to do what you suggested, but it resulted in a circular import because the `megatron_gpt_prompt_learning_model.py` already imports `text_generation_utils.py`

examples/nlp/language_modeling/megatron_gpt_ptune.py
vadam5: Yes, I am planning to do this in a separate PR though because this one is already pretty large. I am working on a new notebook showing how to use both methods. Do you think we should still maintain the ptuning notebook and have two example notebooks for people to use, or would that be confusing for new users?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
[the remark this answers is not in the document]
yidong72: Let me know if it works. I test it locally and it works.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -0,0 +1,653 @@
...
+        # Find the indicies where virtual tokens should be inserted
+        virtual_token_locations = input_ids >= self.pseudo_token_ids_start
+
+        # If there are no virtual tokens, just return discrete token embeds
+        if not virtual_token_locations.any():
+            return discrete_token_embeds
+
+        # Get virtual token embeddings from the prompt table or prompt encoder
+        if self.virtual_prompt_source == 'prompt-table':
+            virtual_token_embeddings = [self.prompt_table(task_id_num) for task_id_num in taskname_ids]
```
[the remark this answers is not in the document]
vadam5: Oh, good point. I'll try it using torch.tensor operations.

nemo/collections/nlp/modules/common/text_generation_utils.py
vadam5: I added that replacement step to stop out of bounds indexing errors related to virtual prompt id values being higher than the number of logits predicted. The virtual prompt id values were sometimes greater than the size of the last dimension of `logits` and `output`. I only apply the replacement to the new token tensor to address soft prompt input tokens for items in the batch that haven't started generating new tokens yet even though others have.

examples/nlp/language_modeling/megatron_gpt_ptune.py
vadam5: I will make sure converting to a nemo file is still showed in the notebook, btw

yidong72: If we already doing that in the `on_train_end`, it is not necessary. 
Finding best checkpoint and save to nemo is something generic enough for all the models. We need some utility function so every model can use it.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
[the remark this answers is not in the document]
yidong72: One solution is to design the prompt table into a high dimensional tensor.
something like `[task_dim, virtual_token_dim, embedding_dim]`
then you need to encode different task-name/ids into number from [0, task_dim-1]. 
Good thing about this approach is you can look up a list of tasks easily by one tensor operation (gather method or tensor slices). 
Bad thing about this approach is it assumes each task uses same amount of virtual tokens. It may cause some waste.
I will let you to decide it is a good idea or not.

nemo/collections/nlp/modules/common/text_generation_utils.py
yidong72: yeah. cyclic import is an issue. one solution is not importing it at the beginning 
do something like
```python
if isinstance(model, nemo.collections.nlp.language_modeling.megatron_gpt_prompt_learning_model.MegatronGPTPPromptLearningModel):```
or import it locally inside the function

File: nemo/collections/nlp/modules/common/text_generation_utils.py (hunk 7 above)
Comment (vadam5) [reply nemo/collections/nlp/modules/common/text_generation_utils.py:671]: Yes, it does. Though they're still included in the list of tokens and long probs portion of the output. So someone using those will have to filter them out before calculating and metrics. But this is true even for padding or EOS tokens, so shouldn't be an issue.

File: examples/nlp/language_modeling/megatron_gpt_ptune.py (hunk 1 above)
Comment (vadam5) [reply examples/nlp/language_modeling/megatron_gpt_ptune.py:101]: Yes, it happens for all nemo models in `one_train_end`

File: nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (vadam5) [reply nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py:320]: I think I'm going to keep the prompt table as is to allow for different amount of virtual tokens per task during inference. I think any bottle neck using this hash table approach might cause is solved in cases where we really want the model to be most performant since the FasterTransformer team will be concatenating virtual token embeddings with the vocab embeddings.

File: nemo/collections/nlp/modules/common/text_generation_utils.py (hunk 5 above)
Comment (vadam5) [reply nemo/collections/nlp/modules/common/text_generation_utils.py:551]: This is kind of long, but I agree its less hacky. I'll add it.

File: nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (vadam5) [reply nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py:320]: I changed the code a bit and changed the CI test to make sure inference works with virtual prompts that are different lengths.

File: nemo/collections/nlp/modules/common/text_generation_utils.py [hunk 12]
```diff
@@ -444,11 +546,19 @@ def switch(val1, val2, boolean):
 
 def forward_step(model, batch, tensor_shape):
 
+    # Should call MegatronGPTPPromptLearningModel's forward method
+    if hasattr(model, 'prompt_table'):
```
yidong72, line 550: change it to?
  ```python
  if isinstance(model, MegatronGPTPromptLearningModel):
  ```

nemo/collections/nlp/modules/common/text_generation_server.py
```diff
@@ -60,9 +61,21 @@ def put(self):
                 logging.error(f"The request key {key} is not allowed")
 
         sentences = request.get_json()["sentences"]
-        if len(sentences) > 128:
+        if isinstance(sentences, tuple):  # Input can be text or tensor
+            if len(sentences[0]) != len(sentences[1]) or sentences[0] > 128:
+                return "Maximum number of sentences is 128", 400
```
yidong72, line 66: In the future, we can try to let setence to take `json` inputs.

nemo/collections/nlp/models/text_classification/ptune_text_classification_model.py
```diff
@@ -29,7 +29,7 @@
 from nemo.collections.nlp.models.language_modeling.megatron_gpt_model import MegatronGPTModel
 from nemo.collections.nlp.models.nlp_model import NLPModel
 from nemo.collections.nlp.modules.common.megatron.megatron_init import initialize_model_parallel_for_nemo
-from nemo.collections.nlp.modules.common.prompt_encoder import PromptEncoder
+from nemo.collections.nlp.modules.common.t5_prompt_encoder import PromptEncoder
```
yidong72, line 32: Is this model still working? maybe we should delete this one if it is broken. 
In the future, let's add the feature that the model can only choose from specified set of labels for classification tasks in the prompt learning model.

yidong72 (commented):
Looks good. only a few comments.

nemo/collections/nlp/modules/common/text_generation_utils.py
vadam5: Oh, whoops. I thought I changed both of them.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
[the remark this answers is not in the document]
vadam5: Confirmed works

yidong72 (approved):
LGTM. Thanks

PR #3883, opened 2022-03-25
Fixing order of exported inputs in NLP models

borisfom:
Signed-off-by: Boris Fomitchev <bfomitchev@nvidia.com>

# What does this PR do ?

Fixing remaining issues with exported order of inputs of NLP models:
* Except Punct&Capital, exported order of inputs of BERT networks was wrong
* There were discrepancies between Bert and Roberta/CamamBert forward() argument orders in nlp/modules/common/huggingface

Also cleaned up unused input_types()/output_types() overloads

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #3893, opened 2022-03-28
Megatron support

ramanathan831:
# What does this PR do ?

Add MegatronBert support for more NLP models; 
Add support for finetuning a downstream task which used MegatronBert as pretrained model

Fix RIVA and ONNX export issues of Bert and MegatronBert (This is coming from https://github.com/NVIDIA/NeMo/pull/3883 PR)
The change in order of arguments won't break usage of anyone unless they deliberately turned off parameters check and passed arguments in positional fashion. ONNX export is one of the places that does that internally in PyTorch and that's why we need to make that order consistent across HF and Nemo BERT models and NLPModel that uses those modules as input_module.

NLPModel __init__() refactoring ( https://github.com/NVIDIA/NeMo/pull/3887 ) 
This is done to reduce code duplication and make it manageable. Each class has their own minor config differences and registers own artifacts, and those subtle inconsistencies break assumptions downstream (e.g. in Riva)

NLP

# Changelog 

# Usage

```python
```

**Pre checks**:
- [ x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
  - [x ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x ] New Feature
- [x ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


Zhilin123 (commented):
For some of the dialogue models (IntentSlot, DialogueStateTracking), I see that `self.bert_model = get_lm_model (...)` is deleted. Presumably, this function is now subsumed under `super().__init__()` . Can you explain why this is done?

Zhilin123 (commented):
I am also curious why the signature for `def forward(self, input_ids, attention_mask, token_type_ids):` would be changed, as presumably this will break usage for many people who have already built upon these models. Is there a way to keep the signature the same?

borisfom:
> I am also curious why the signature for `def forward(self, input_ids, attention_mask, token_type_ids):` would be changed, as presumably this will break usage for many people who have already built upon these models. Is there a way to keep the signature the same?

This is coming from #3883 PR, fixing the bug with ONNX export. The change in order of arguments won't break usage of anyone except they deliberately turned off parameters check and passed arguments in positional fashion. ONNX export is one of the places that does that internally in PyTorch and that's why we need to make that order consistent across HF and Nemo BERT models and NLPModel that uses those modules as input_module.

borisfom:
> For some of the dialogue models (IntentSlot, DialogueStateTracking), I see that `self.bert_model = get_lm_model (...)` is deleted. Presumably, this function is now subsumed under `super().__init__()` . Can you explain why this is done?

This is done to reduce code duplication and make it manageable. Each class has their own minor config differences and registers own artifacts, and those subtle inconsistencies break assumptions downstream (e.g. in Riva)

Zhilin123:
> This is coming from https://github.com/NVIDIA/NeMo/pull/3883 PR, fixing the bug with ONNX export. The change in order of arguments won't break usage of anyone unless they deliberately turned off parameters check and passed arguments in positional fashion. ONNX export is one of the places that does that internally in PyTorch and that's why we need to make that order consistent across HF and Nemo BERT models and NLPModel that uses those modules as input_module.
> This is done to reduce code duplication and make it manageable. Each class has their own minor config differences and registers own artifacts, and those subtle inconsistencies break assumptions downstream (e.g. in Riva)
Makes sense. I think there would a number of people who turn off parameters check [particularly as attention_mask and token_type_ids seem to be of the same shape] and use the args in the positional manner. Therefore, it might be helpful to raise a warning or be clear about it in the documentation that this is a breaking, non-backward-compatible change.

borisfom (approved):

borisfom (approved):

PR #3898, opened 2022-03-29
Make test_gpt_eval unit test less strict

yidong72:
Make the unit test less strict and make inference configuration immutable.


ericharper (approved):
LGTM. Thanks!

PR #3928, opened 2022-04-04
[Draft] [TTS] Fastpitch restructure

subhankar-ghosh:
# What does this PR do ?

AudioToCharWithDursF0Dataset class will be deprecated so this PR adds code and config to use old FastPitch checkpoint (checkpoint that was trained using AudioToCharWithDursF0Dataset class which is now deprecated) with new code (TTSDataset) to support inference.

**Collection**: tts.models.fastpitch

# Changelog 
- Adds Warning when using old checkpoint
- Adds config file to be used with old checkpoint

# Usage
* In FastPitchModel.from_pretrained use https://github.com/NVIDIA/NeMo/tree/main/examples/tts/conf/fastpitch_align_old.yaml \
   as config_file to avoid using deprecated AudioToCharWithDursF0Dataset class and run into errors. \

```python
# FastPitchModel.from_pretrained("tts_en_fastpitch", override_config_path="/path/to/fastpitch_align_old.yaml") 
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/tts/models/fastpitch.py
```diff
@@ -178,13 +174,12 @@ def parser(self):
             ds_class_name = self._cfg.train_ds.dataset._target_.split(".")[-1]
 
             # TODO(Oktai15): remove it in 1.8.0 version
-            if ds_class_name == "AudioToCharWithPriorAndPitchDataset" or ds_class_name == "TTSDataset":
+            if ds_class_name == "TTSDataset":
                 self._parser = self.vocab.encode
             else:
                 raise ValueError(f"Unknown dataset class: {ds_class_name}")
         else:
             # TODO(Oktai15): remove it in 1.8.0 version
```
blisc, line 182: Let's remove all the todo comments since you did them.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -550,3 +538,32 @@ def input_example(self, max_batch=1, max_dim=256):
...
+            model_name,
+            refresh_cache,
+            override_config_path,
+            map_location,
+            strict,
+            return_config,
+            trainer,
+            save_restore_connector,
+            )
+        return model
```
blisc, line 569: Ok let's remove this code then and move it to init. Do a check like `if self.ds_class_name == "AudioToCharWithPriorAndPitchDataset":` then write an appropriate error message.

examples/tts/conf/fastpitch_align_old.yaml
```diff
@@ -0,0 +1,173 @@
+learn_alignment: true
```
blisc, line 1: Do we need this file?

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Done

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Done.

examples/tts/conf/fastpitch_align_old.yaml
subhankar-ghosh: I would say yes. Because, loading checkpoints need a config file and we cannot use the same config file that is inside the old Fastpitch.nemo file. We can replace the config file inside the old Fastpitch.nemo file with this file, then we would not need this file anymore. Or, the user has to manually make quite a lot of changes to existing fastpitch_align_v1.05.yaml file to use it to load the old checkpoint.

blisc: If we were using the old checkpoint, will it raise the ValueError that you added in the init? Or are there other issues?

subhankar-ghosh: Yes, if we were using the old checkpoint then it will raise the ValueError that I added. The ValueError message points towards this config file and shows how to use it.

subhankar-ghosh: > Can we print this error message

We are printing this error message when the user tries to load the old checkpoint with the `AudioToCharWithPriorAndPitchDataset` class.

> try using a config that creates the dataset properly

Do you mean an existing config? Yes, there are existing configs that use the new TTSDataset but other parts of the config are not compatible with the old checkpoint. 
That is why I added this config `fastpitch_align_old.yaml` which uses TTSDataset (the new dataset) to load the old model. This config is the config used to load the old checkpoint with just the dataset parts replaced by the TTSDataset settings. We can use a different name for this config if you want.

blisc: Let's not add this yaml. Let's change this https://github.com/NVIDIA/NeMo/pull/3928/files#diff-eaf0251e9d829a2b3aa53d9d90c6ca8181cc52d02be139215cc8baffc986e7acR73-R77 to an error message.

Since we are only supporting inference with the old models, the only time that we call into the old dataset is in `parser()` so let's change it so that when a user calls into `parser()` with the old config, we then load a correct config (let's define this as a DictConfig or dictionary somewhere in code of just the dataset class rather than a full yaml config file) with TTSDataset. We can remove the the ValueError being raised in the init with a warning message saying no training/finetuning support and only inference.

PR #3975, opened 2022-04-12
Fix notebook `misconfiguration` error

yidong72:
# What does this PR do ?

The notebook is complaining about Trainer `misconfiguration` error because it tries to spawn processes in the python interactive session. Fixed it by using torch elastic environment.


ericharper (approved):
LGTM. Thanks!

PR #3977, opened 2022-04-12
Megatron NMT Restore from T5/BART and finetune

MaximumEntropy:
# What does this PR do ?

Adds the ability to restore from a T5/BART model and fine-tune it for NMT. With tarred datasets only.

**Collection**: NLP

# Changelog 
- Fixes issues with rebasing to the Encoder-Decoder PP changes.
- Overrides the pre-training config with NMT-related things.
- Reconfigures the microbatch calculator on every training/eval step.

# Usage

Edit the aayn_base_megatron.yaml file with

- pretrained_model_path: /path/to/pretrained/t5.nemo
- pretrained_model_type: t5/bart

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
```diff
@@ -538,24 +546,52 @@ def setup_enc_dec_tokenizers(
...
             use_fast=False,
             r2l=decoder_r2l,
+            legacy=decoder_sentencepiece_legacy,
         )
 
         # validate no token is negative for sentencepiece tokenizers
-        for tok_name, tok_library, tok_model in [
-            ("encoder_tokenizer", encoder_tokenizer_library, encoder_tokenizer),
-            ("decoder_tokenizer", decoder_tokenizer_library, decoder_tokenizer),
+        for tok_name, tok_library, tok_model, legacy in [
```
aklife97, line 553: Is it necessary to have the legacy support here in the non-megatron NMT model? It would be great to keep the legacy code in minimal amount of places, so whenever we switch off from using the legacy tokenizer we don't have to worry about a lot of old models breaking.

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
```diff
@@ -570,6 +570,71 @@ def process_micro_batch(self, micro_batch):
 
         return tokens_enc, tokens_dec, loss_mask, labels, enc_mask, dec_mask
 
+    def _process_global_batch_without_megatron_batch_sampler(self, global_batch, tokenizer=None):
```
aklife97, line 573: Wouldn't it better to do this as part of dataloader as discussed? 
If that's upcoming in an upcoming/broader PR then looks good

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
```diff
@@ -944,7 +958,54 @@ def write_parallel_batches_to_tarfiles(
             decoder_bpe_dropout=decoder_bpe_dropout,
             decoder_model_name=decoder_model_name,
             decoder_r2l=decoder_tokenizer_r2l,
+            encoder_tokenizer_legacy=encoder_tokenizer_legacy,
+            decoder_tokenizer_legacy=decoder_tokenizer_legacy,
         )
+
+        # validate no token is negative for sentencepiece tokenizers and add missing special tokens.
+        for tok_name, tok_library, tok_model, legacy in [
```
aklife97, line 966: I would want to avoid putting legacy support in places like this script as later removal could be difficult if things rely on this parameter being there. 
We should probably just have the legacy support in the megatron NMT model and the tarred dataset creation script for immediate use.

examples/nlp/machine_translation/megatron_nmt_training.py
```diff
@@ -83,7 +92,73 @@ def main(cfg) -> None:
...
+            # Pre-trained models should use the legacy sentencepiece tokenizer ex: mT5
+            pretrained_cfg.encoder_tokenizer.sentencepiece_legacy = True
+            pretrained_cfg.decoder_tokenizer.sentencepiece_legacy = True
+
+            # Override dropout
+            pretrained_cfg.hidden_dropout = cfg.model.hidden_dropout
+            pretrained_cfg.attention_dropout = cfg.model.attention_dropout
+
+            # Override precision
+            pretrained_cfg.precision = cfg.model.precision  # Set above from trainer.precision
```
aklife97, line 136: Wouldn't it be an issue if the precision of a pretrained model is changed? Shouldn't we check that the precision values match?

examples/nlp/machine_translation/megatron_nmt_training.py
```diff
@@ -83,7 +92,73 @@ def main(cfg) -> None:
...
+            # Set tokenizer paths:
+            pretrained_cfg.encoder_tokenizer = pretrained_cfg.tokenizer
+            pretrained_cfg.decoder_tokenizer = pretrained_cfg.tokenizer
+
+            # Pre-trained models should use the legacy sentencepiece tokenizer ex: mT5
+            pretrained_cfg.encoder_tokenizer.sentencepiece_legacy = True
+            pretrained_cfg.decoder_tokenizer.sentencepiece_legacy = True
+
+            # Override dropout
+            pretrained_cfg.hidden_dropout = cfg.model.hidden_dropout
```
aklife97, line 132: Does it make sense to be able to change the dropout rate when using a pretrained model? Wouldn't the changed dropout cause a change in the distribution and needlessly change the input domain for the upcoming layers?

aklife97 (commented):
Thanks for the PR, looks great overall! Just have some questions/suggestions

nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py
MaximumEntropy: All of this is to train any model regardless of Megatron-based or NeMo-based models using the mT5 tokenizer which requires adding special tokens.

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
MaximumEntropy: Yeah that refactor will be part of an upcoming PR. Other fine-tuning models also need to use something like this for now.

nemo/collections/nlp/data/machine_translation/preproc_mt_data.py
MaximumEntropy: Same as the above point. If you want to pre-process a dataset with the mT5 tokenizer, your tarred data will silently contain `-1` everywhere you insert <bos> and you may only catch the error while training, where the model will fail with a not-so-helpful CUDA index assert check.

examples/nlp/machine_translation/megatron_nmt_training.py
MaximumEntropy: This is totally fine since weights are fp32 you can train with bf16 and resume with fp16 and vice-versa.

examples/nlp/machine_translation/megatron_nmt_training.py
MaximumEntropy: This is useful for cases where you want to regularize your model more while fine-tuning than while training. Check out the mT5 paper where they train with dropout = 0 and fine-tune with dropout 0.1 https://arxiv.org/abs/2010.11934.

aklife97: Hmm... interesting, never paid attention to the changing dropout in the paper. Looks good in this case

aklife97 (approved):
Looks good! I am still a bit concerned with the widespread legacy usage, but I guess we don't have a better option as of now.

soumye (commented):
lgtm

examples/nlp/machine_translation/megatron_nmt_training.py
```diff
@@ -83,7 +92,73 @@ def main(cfg) -> None:
...
+            pretrained_cfg.max_generation_delta = cfg.model.max_generation_delta
+
+            # Set label smoothing
+            pretrained_cfg.label_smoothing = cfg.model.label_smoothing
+
+            # Set tokenizer paths:
+            pretrained_cfg.encoder_tokenizer = pretrained_cfg.tokenizer
+            pretrained_cfg.decoder_tokenizer = pretrained_cfg.tokenizer
+
+            # Pre-trained models should use the legacy sentencepiece tokenizer ex: mT5
```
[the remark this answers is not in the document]
MaximumEntropy: They are pre-trained with legacy sentencepiece as well to account for the addition of special tokens.

examples/nlp/machine_translation/conf/aayn_base_megatron.yaml
```diff
@@ -44,7 +44,10 @@ model:
   src_language: 'en'
   tgt_language: 'de'
   max_generation_delta: 20 # Maximum decoder sequence length is encoder sequence length + this parameter.
-  micro_batch_size: 0 # This is just here to keep compute_consumed_samples() happy.
+  pretrained_model_path: null # Path to a pretrained model 
+  pretrained_model_type: T5
+  global_batch_size: 32 # This is just here to keep compute_consumed_samples() happy.
+  micro_batch_size: 32 # This is just here to keep compute_consumed_samples() happy.
```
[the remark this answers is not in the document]
MaximumEntropy: For now this is just a place-holder with tarred datasets, but once we move to megatron-based dataloaders, this will actually start to do something and is how you will configure grad acc etc.

examples/nlp/machine_translation/megatron_nmt_training.py
```diff
@@ -83,7 +92,73 @@ def main(cfg) -> None:
...
+            # Override data and global/micro batch size.
+            pretrained_cfg.train_ds = cfg.model.train_ds
+            pretrained_cfg.validation_ds = cfg.model.validation_ds
+            pretrained_cfg.test_ds = cfg.model.test_ds
+
+            pretrained_cfg.micro_batch_size = cfg.model.micro_batch_size
+            pretrained_cfg.global_batch_size = cfg.model.global_batch_size
+
+            # Class target for the new class being restored.
+            pretrained_cfg.target = (
```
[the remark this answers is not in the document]
MaximumEntropy: This isn't something I can change, it is an attribute that is part of NeMo's ModelPT class that needs to be set to decide which class is being restored.

examples/nlp/machine_translation/conf/aayn_base_megatron.yaml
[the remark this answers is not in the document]
soumye: ok

examples/nlp/machine_translation/megatron_nmt_training.py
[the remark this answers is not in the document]
soumye: I see.

PR #3978, opened 2022-04-12
Compatability override to load_state_dict for old TTS checkpoints

redoctopus:

# What does this PR do ?

Overrides `load_state_dict` for four TTS models (to be deprecated in 1.9.0) such that they work with old NGC checkpoints, plus minor tutorial fixes/updates.

**Collection**: TTS

# Changelog 
- Removes `forward_basis` and `inverse_basis` from the `state_dict` of existing checkpoints, which were trained before NeMo moved away from using torch_stft
- Adds `window` to `state_dict` for the preprocessor
- Semicolon added to one line in the TalkNet training tutorial to avoid printing the whole model
- Added `trainer.gpu` command line argument to Tacotron2 notebook
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/models/glow_tts.py
+++ b/nemo/collections/tts/models/glow_tts.py
@@ -288,6 +288,7 @@ class GlowTTSModel(SpectrogramGenerator):
         switch from torch_stft. It will be removed when this model is deprecated.
         """
         if 'preprocessor.featurizer.stft.forward_basis' in state_dict:
+            logging.warning("Loading old checkpoint, defaulting to hann window.")
             window_dim = state_dict['preprocessor.featurizer.stft.forward_basis'].shape[-1]
             state_dict['preprocessor.featurizer.window'] = torch.hann_window(window_dim)
             del state_dict['preprocessor.featurizer.stft.forward_basis']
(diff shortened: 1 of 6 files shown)


nemo/collections/tts/models/glow_tts.py
```diff
@@ -280,3 +280,16 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
+
+    def load_state_dict(self, state_dict, strict=True):
+        """Stopgap measure to keep old checkpoints working before full model deprecation.
+
+        This override fiddles with the state_dict in order to keep functionality from old checkpoints after the
+        switch from torch_stft. It will be removed when this model is deprecated.
+        """
+        if 'preprocessor.featurizer.stft.forward_basis' in state_dict:
+            window_dim = state_dict['preprocessor.featurizer.stft.forward_basis'].shape[-1]
+            state_dict['preprocessor.featurizer.window'] = torch.hann_window(window_dim)
```
blisc, line 292: Let's add a warning message that we are defaulting to `hann` window as that might not be the case for all models

redoctopus: Good call, added to all four model overrides

blisc (approved):

PR #3984, opened 2022-04-13
LogProb calculation performance fix

yidong72:
# What does this PR do ?

Make log prob calculation faster and handles longer sequences.


ericharper (approved):
LGTM. Thanks!

PR #3986, opened 2022-04-13
[Core] Fix type checking to be compatible with named tuples

artbataev:
Signed-off-by: Vladimir Bataev <vbataev@nvidia.com>

# What does this PR do ?

Fix inner type checking system for outputs to be compatible with named tuples.
This adds the ability to use named tuples (`typing.NamedTuple` or `collections.namedtuple`) for outputs. This is especially useful when the number of output elements is large. 
Named tuples are compatible with simple tuples, there should be no problems with mixing them.

# Usage
```python
class OutputType(NamedTuple):
    y: torch.Tensor
    z: torch.Tensor

class MultipleOutputTypesWithNamedTuple(Typing):
    @property
    def output_types(self):
        return {"y": NeuralType(('B',), ElementType()), "z": NeuralType(('B',), ElementType())}

    @typecheck()
    def __call__(self, x):
        y = x + 1
        z = x + 2
        return OutputType(y=y, z=z)

obj = MultipleOutputTypesWithNamedTuple()
result = obj(x=torch.zeros(10))
result_y, result_z = result  # can unpack or use `result.y`, `result.z`
```
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

@titu1994, @okuchaiev, @ericharper

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -262,7 +262,7 @@ class Typing(ABC):
             mandatory_out_types_list = list(metadata.mandatory_types.items())
 
             # First convert all outputs to list/tuple format to check correct number of outputs
-            if type(out_objects) in (list, tuple):
+            if isinstance(out_objects, (list, tuple)):
                 out_container = out_objects  # can be any rank nested structure
             else:
                 out_container = [out_objects]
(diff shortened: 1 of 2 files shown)


titu1994 (approved):
Fantastic ! Merge when ready

ericharper (approved):
LGTM. Thanks!

PR #3993, opened 2022-04-13
Fix compat override for TalkNet Aligner

redoctopus:
# What does this PR do ?

Moves the override to `load_state_dict()` from TalkNet (which doesn't need it) to EncDecCTCModel for compatibility with the TalkNet Aligner checkpoint.

This should be removed in 1.9.0 when the TalkNet Aligner model is deprecated.

**Collection**: TTS, ASR

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


titu1994 (approved):
Good enough for now, let's add a note to revert this in main branch for 1.9

PR #3998, opened 2022-04-14
Megatron work-arounds

borisfom:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/models/question_answering/qa_model.py
+++ b/nemo/collections/nlp/models/question_answering/qa_model.py
@@ -64,7 +64,7 @@ class QAModel(NLPModel):
         if isinstance(hidden_states, tuple):
             hidden_states = hidden_states[0]
         logits = self.classifier(hidden_states=hidden_states)
-        return logits
+        return logits.float()
 
     def training_step(self, batch, batch_idx):
         input_ids, input_type_ids, input_mask, unique_ids, start_positions, end_positions = batch
--- a/nemo/collections/nlp/models/text_classification/text_classification_model.py
+++ b/nemo/collections/nlp/models/text_classification/text_classification_model.py
@@ -85,7 +85,7 @@ class TextClassificationModel(NLPModel, Exportable):
         if isinstance(hidden_states, tuple):
             hidden_states = hidden_states[0]
         logits = self.classifier(hidden_states=hidden_states)
-        return logits
+        return logits.float()
 
     def training_step(self, batch, batch_idx):
         """
--- a/nemo/collections/nlp/modules/common/megatron/transformer.py
+++ b/nemo/collections/nlp/modules/common/megatron/transformer.py
@@ -56,7 +56,7 @@ class ColumnLinear(tensor_parallel.ColumnParallelLinear):
     def forward(self, input_):
         world_size = get_tensor_model_parallel_world_size()
         if input_.requires_grad or world_size > 1:
-            return tensor_parallel.ColumnParallelLinear.forward(input_)
+            return tensor_parallel.ColumnParallelLinear.forward(self, input_)
 
         bias = self.bias if not self.skip_bias_add else None
         # Matrix multiply.
--- a/nemo/collections/nlp/modules/common/megatron/transformer.py
+++ b/nemo/collections/nlp/modules/common/megatron/transformer.py
@@ -32,7 +32,7 @@ from nemo.collections.nlp.modules.common.megatron.fused_bias_dropout_add import 
 from nemo.collections.nlp.modules.common.megatron.fused_bias_gelu import fused_bias_gelu
 from nemo.collections.nlp.modules.common.megatron.fused_layer_norm import get_layer_norm
 from nemo.collections.nlp.modules.common.megatron.module import MegatronModule
-from nemo.collections.nlp.modules.common.megatron.utils import ApexGuardDefaults, attention_mask_func, erf_gelu
+from nemo.collections.nlp.modules.common.megatron.utils import attention_mask_func, erf_gelu
 
 
 """ We use the following notation throughout this file:
@@ -58,7 +58,6 @@ class ColumnLinear(tensor_parallel.ColumnParallelLinear):
         if input_.requires_grad or world_size > 1:
             return tensor_parallel.ColumnParallelLinear.forward(self, input_)
 
-        bias = self.bias if not self.skip_bias_add else None
         # Matrix multiply.
         output = torch.matmul(input_, self.weight.t())
         if not self.skip_bias_add:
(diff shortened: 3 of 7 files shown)


nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -94,47 +114,28 @@ def __init__(
             use_cpu_initialization=use_cpu_initialization,
         )
 
-        if activation in ['geglu', 'reglu', 'swiglu']:
-            # Separate linear layer for *GLU activations.
+        if activation == 'geglu':
```
MaximumEntropy, line 117: Why are we removing the other activations here?

nemo/collections/nlp/models/question_answering/qa_model.py
```diff
@@ -64,7 +64,7 @@ def forward(self, input_ids, attention_mask, token_type_ids):
         if isinstance(hidden_states, tuple):
             hidden_states = hidden_states[0]
         logits = self.classifier(hidden_states=hidden_states)
-        return logits
+        return logits.float()
```
MaximumEntropy, line 67: Do we really need this conversion to float?

nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -236,11 +228,11 @@ def __init__(
             )
         else:
             assert attention_type == AttnType.cross_attn
-            self.query = tensor_parallel.ColumnParallelLinear(
+            self.query = tensor_parallel.ColumnLinear(
```
MaximumEntropy, line 231: You probably want to remove `tensor_parallel.` here.

MaximumEntropy (changes requested):
Just marking a few changes here so we don't merge yet.

nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -58,6 +60,24 @@
 """
 
 
+class ColumnLinear(tensor_parallel.ColumnParallelLinear):
```
ramanathan831, line 63: if APEX is false, then tensor_parallel import will not happen, resulting in a tensorparallel not found kind of error right?

nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -94,47 +114,28 @@ def __init__(
             use_cpu_initialization=use_cpu_initialization,
         )
 
-        if activation in ['geglu', 'reglu', 'swiglu']:
```
ramanathan831, line 97: any specific reason for taking off reglu and swiglu?

nemo/collections/nlp/models/text_classification/text_classification_model.py
```diff
@@ -85,7 +85,7 @@ def forward(self, input_ids, attention_mask, token_type_ids):
         if isinstance(hidden_states, tuple):
             hidden_states = hidden_states[0]
         logits = self.classifier(hidden_states=hidden_states)
-        return logits
+        return logits.float()
```
ramanathan831, line 88: Should we add for other models too like token_classification?

nemo/collections/nlp/modules/common/megatron/transformer.py
ericharper, line 63: Why is this needed?

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: This looks like a merge issue, I will address that.

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: tensor_parallel.ColumnParallelLinear won't export - it has Autograd Function in execution path during inference - which is wrong and I have reported the issue to Apex - but we need some workaround in place now.

nemo/collections/nlp/models/text_classification/text_classification_model.py
borisfom: This is workaround for uncertain output format when using autocast - I am still exploring how to avoid it by forcing FP32 on export side. I guess, it may be needed for some Pytorch scenarios as well. Probably, should be flagged as AMP issue.

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: Fixed.

nemo/collections/nlp/models/question_answering/qa_model.py
borisfom: Explained below.

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: Done.

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: APEX is now a requirement.

nemo/collections/nlp/modules/common/megatron/transformer.py
borisfom: Done.

nemo/collections/nlp/models/text_classification/text_classification_model.py
borisfom: Added reasonable ones.

MaximumEntropy: Have you tried setting the precision to 32 within the `.nemo` model config before running export? Megatron models have a precision attribute that is not a part of the trainer and lives inside the model config.

nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -97,9 +108,9 @@ def __init__(
         if activation in ['geglu', 'reglu', 'swiglu']:
             # Separate linear layer for *GLU activations.
             # Source: https://github.com/huggingface/transformers/blob/bee361c6f1f7704f8c688895f2f86f6e5ff84727/src/transformers/models/t5/modeling_t5.py#L292
-            self.dense_h_to_4h_2 = tensor_parallel.ColumnParallelLinear(
+            self.dense_h_to_4h_2 = ColumnLinear(
                 hidden_size,
-                ffn_hidden_size,  # NOTE: When using *glu, divide ffn dim by 2/3 to keep overall params the same.
+                ffn_hidden_size,  # NOTE: When using geglu, divide ffn dim by 2/3 to keep overall params the same.
```
ramanathan831, line 113: Replaced during one of the merges, I think this should be for the functions *glu like reglu too, instead of only geglu. Correct me if I am wrong

borisfom: You're right, missed it. Will clean it up.

ramanathan831 (approved):

MaximumEntropy (approved):

PR #4001, opened 2022-04-15
Fix P-Tune T5 model

yidong72:
# What does this PR do ?

Fix the broken P-tune T5 model  after the PP T5 feature is merged.


MaximumEntropy (changes requested):

MaximumEntropy (approved):
LGTM!

PR #4017, opened 2022-04-16
Fix the broadcast shape mismatch

yidong72:
# What does this PR do ?

Fix the shape mismatch when doing broadcast, which causes logprob returns bad numbers.


ericharper (approved):
LGTM.

PR #4033, opened 2022-04-20
Fix Punctuation and Capitalization model batching. An issue with shuffling.

PeganovAnton:
# What does this PR do ?

1. If in DDP mode number of batches is not evenly divisible by number of processes, some batches may be processed several times. This lead to distorted metrics. This PR adjusts number of batches to number of processes.
2. In DDP mode batches in different processes are shuffled independently. This PR sets random seed for this shuffling. Before this fix, shuffling lead to some batches being processed several times and some batches not processed at all. 

**Collection**: nlp

# Changelog 
- [adjusting number of batches](https://github.com/NVIDIA/NeMo/blob/45e9deca8c29b6f3142294352c7350f385ba6b28/nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py#L1354)
- [setting random seed](https://github.com/NVIDIA/NeMo/blob/45e9deca8c29b6f3142294352c7350f385ba6b28/nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py#L787)

# Usage

To notice changes support numbers you can use this [stage](https://github.com/NVIDIA/NeMo/blob/91e6c2552a9ca8d00bf8ce758adea68d9c49f803/Jenkinsfile#L1219) in Jenkins.
  
**PR Type**:
- [x] Bugfix

Closes bug https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3610560&cmtNo=

### DIFF
--- a/nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
+++ b/nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
@@ -889,7 +889,7 @@ class BertPunctuationCapitalizationDataset(Dataset):
         verbose: bool = True,
         n_jobs: Optional[int] = 0,
         number_of_batches_is_multiple_of: int = 1,
-        batch_shuffling_random_state: int = 42,
+        batch_shuffling_random_seed: int = 42,
         tokenization_progress_queue: Optional[mp.Queue] = None,
         batch_mark_up_progress_queue: Optional[mp.Queue] = None,
         batch_building_progress_queue: Optional[mp.Queue] = None,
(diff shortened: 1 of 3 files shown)


nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
```diff
@@ -1170,6 +1191,96 @@ def _read_dataset(
...
+            batch_seq_lengths: lengths of elements in batch after padding created inside :meth:`_mark_up_batches`
+                method.
+
+        Returns:
+            batch_beginnings: a list of indices in ``input_ids`` of first samples of every batch
+            batch_sizes: a list of numbers of samples in batches
+            batch_seq_lengths: a list of sequence lengths after padding for every batch
+        """
+        batch_beginnings, batch_sizes, batch_seq_lengths = batch_beginnings.copy(), batch_sizes.copy()
+        batch_seq_lengths = batch_seq_lengths.copy()
```
carolmanderson, line 1233: Line 1232 here generated an error because there are three items on the left of the `=` sign and only two on the right. This can be fixed by either deleting `batch_seq_lengths` from the left side or adding `batch_seq_lengths.copy()` to the right side (and then deleting line 1233).

carolmanderson (changes requested):
This fixed the issue for me. Thank you! There is a small bug that needs changing.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
PeganovAnton: Thank you! I broke this when fixed code style.

PR #4054, opened 2022-04-25
fix GPT model parallel eval

yidong72:
Signed-off-by: Yi Dong <yidong@nvidia.com>

# What does this PR do ?

Fix the `broadcast` inconsistent type issue which cause the model parallel gpt eval failure.


vadam5 (approved):
LGTM

PR #3589, opened 2022-02-02
upgrade PTL trainer flags

nithinraok:
Signed-off-by: nithinraok <nithinrao.koluguri@gmail.com>

# What does this PR do ?

Upgrade pytorch_lightning trainer flags that are going to be deprecated in future

**Collection**: Core and All collections yaml files 

# Changelog 
- Update pytorch_lightning deprecated variables, namely:

> -  accelerator to the trainer now takes the type of accelerator device than distributed type
> -  distributed type is handled by strategy flag
> -  previously: accelerator='ddp' now accelerator='gpu' strategy='ddp'
> -  trainer flag checkpoint_callback is renamed to enable_checkpointing
> - ModelCheckpoint argument evert_n_val_epochs renamed to every_n_epochs
> - Update `trainer.gpus` to `trainer.devices` ( With this change setting devices=0 doesnt auto select gpu instead need to make accelerator='cpu' and also note accelerator=cpu now will only accept >0 as values for devices) 
> - `examples/asr/transcribe_speech.py` script modified to remove cuda option. Script automatically searches for GPU and if its not present then falls back to CPU  
> - `checkpointing_callback` -> `enable_checkpointing` 
> - `flush_logs_every_n_steps` removed from trainer flags
> - `sync_batchnorm` and `num_nodes` removed from NLPDPPPlugins 

# Usage
- 

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [] Did you write any new necessary tests?
- [] Did you add or update any necessary documentation?
- [] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
-

PR #4056, opened 2022-04-25
Update num worker calculation due to PTL flag changes

redoctopus:

# What does this PR do ?

Removes conditionals in favor of simple calculation of num workers due to changes in PTL flags, as per conversation with @nithinraok and prior discussion in #3589 

# Changelog
- Update calculation of num workers in model class (optimizer setup)
- Update calculation of num workers in TTS helpers
- Bump PTL version requirement from 1.6.0 -> 1.6.1

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


nemo/collections/tts/helpers/helpers.py
```diff
@@ -91,17 +91,7 @@ def get_batch_size(train_dataloader):
...
-    elif trainer.accelerator == "ddp_cpu":
-        return trainer.num_devices * trainer.num_nodes
-    elif trainer.accelerator == "ddp":
-        return trainer.num_devices * trainer.num_nodes
-    else:
-        logging.warning(
-            f"The lightning trainer received accelerator: {trainer.accelerator}. We " "recommend to use 'ddp' instead."
-        )
-        return trainer.num_devices * trainer.num_nodes
+    return trainer.num_devices * trainer.num_nodes
```
titu1994, line 94: Can you check if trainer is none and return 1?

nemo/core/classes/modelPT.py
```diff
@@ -483,21 +483,10 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
...
                 optim_config['sched']['t_accumulate_grad_batches'] = self._trainer.accumulate_grad_batches
                 optim_config['sched']['t_limit_train_batches'] = self._trainer.limit_train_batches
-                if self._trainer.accelerator is None:
-                    optim_config['sched']['t_num_workers'] = self._trainer.num_devices or 1
-                elif self._trainer.accelerator == "ddp_cpu":
-                    optim_config['sched']['t_num_workers'] = self._trainer.num_devices * self._trainer.num_nodes
-                elif self._trainer.accelerator == "ddp":
-                    optim_config['sched']['t_num_workers'] = self._trainer.num_devices * self._trainer.num_nodes
-                elif HAVE_NLPPLUGIN and isinstance(self._trainer.accelerator.training_type_plugin, NLPDDPPlugin):
+                optim_config['sched']['t_num_workers'] = self._trainer.num_devices * self._trainer.num_nodes
```
titu1994, line 486: Check if trainer is None and set to 1

nemo/core/classes/modelPT.py
```diff
@@ -483,21 +483,10 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
...
+                if HAVE_NLPPLUGIN and isinstance(self._trainer.accelerator.training_type_plugin, NLPDDPPlugin):
                     app = AppState()
                     optim_config['sched']['t_num_workers'] = app.data_parallel_size
-                else:
-                    logging.warning(
-                        f"The lightning trainer received accelerator: {self._trainer.accelerator}. We "
-                        "recommend to use 'ddp' instead."
-                    )
-                    optim_config['sched']['t_num_workers'] = self._trainer.num_devices * self._trainer.num_nodes
```
titu1994, line 500: Hmm ok to remove the logging but I'd like a default else case

titu1994 (changes requested):

nemo/collections/tts/helpers/helpers.py
nithinraok: where would we use `get_num_workers` is trainer is None?

nemo/core/classes/modelPT.py
redoctopus: Default case here would be to use the above calculation, do you have another behavior in mind?

nemo/collections/tts/helpers/helpers.py
SeanNaren: Looking at the usage, both assume the trainer exists:
https://github.com/NVIDIA/NeMo/blob/b458394830719c1489261185ef728d0a2a5767a0/nemo/collections/tts/models/hifigan.py
https://github.com/NVIDIA/NeMo/blob/66ac0b3cc10d50a46635717bcfccb12426b8cefc/nemo/collections/tts/models/univnet.py

SeanNaren (approved):
I'm glad the abstractions in Lightning are cleaning up some of this code, was one of the motivations for cleaning up the API a bit :)

nithinraok (approved):

nemo/collections/tts/helpers/helpers.py
titu1994: Alright if trainer is mandatory then this is fine as is.

nemo/core/classes/modelPT.py
titu1994: Oh it moved up. My bad.

titu1994 (approved):
Lgtm, thanks

PR #4061, opened 2022-04-25
[Core] Support pre-extracted nemo checkpoint for restoration

titu1994:

# What does this PR do ?

Adds support to SaveRestoreConnector to employ a pre-extracted nemo directory path as a restoration directory.
This is required for extremely large language model restoration, which cannot be extracted partially during multinode inference.

**Collection**: [Core, Megatron]

# Changelog 

# Note
- During restore_from, the path of the nemo file is ignored in its entirety since the actual pre-extracted directory is passed.
- The filepath passed to restore_from in such a case is not preserved, it is overwritten by the value of `connector.model_extracted_dir` inside AppState.

# Usage

```python
# Extract a nemo file apriori to restoration
# Here we will use the SaveRestoreConnector's method to extract the checkpoint to a given directory 

connector = MySaveRestoreConnector()
MySaveRestoreConnector._unpack_nemo_file(nemo_filepath, extracted_directory)

# Then update SaveRestoreConnector with path of the extracted directory
connector.model_extracted_dir = extracted_directory

# Call restore as always
# note, we pass in the "old" nemo_filepath, stored somewhere other than the extracted directory
restored_model = ModelPT.restore_from(nemo_filepath, save_restore_connector=connector)
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?
@okuchaiev @ericharper


nemo/core/classes/modelPT.py
```diff
@@ -294,7 +294,11 @@ def restore_from(
         if save_restore_connector is None:
             save_restore_connector = SaveRestoreConnector()
 
-        restore_path = os.path.abspath(os.path.expanduser(restore_path))
+        if save_restore_connector.model_extracted_dir is None:
+            restore_path = os.path.abspath(os.path.expanduser(restore_path))
+        else:
+            restore_path = os.path.abspath(os.path.expanduser(save_restore_connector.model_extracted_dir))
```
okuchaiev, line 300: what if the user provided absolute path outside of their homedir?

tests/core/test_save_restore.py
```diff
@@ -523,3 +541,67 @@ def test_mock_model_model_collision(self):
         with pytest.raises(ValueError, match="Creating model config node is forbidden"):
             model = MockModel(cfg=cfg.model, trainer=None)  # type: MockModel
             model = model.to('cpu')
+
+    @pytest.mark.unit
+    def test_restore_from_save_restore_connector_extracted_dir(self):
```
okuchaiev, line 546: thanks for adding unittests!

okuchaiev (approved):

nemo/core/classes/modelPT.py
titu1994: expanduser() only expands `~` in a user provided path, so if absolute path is given then as long as theres no ~ in it, it will be correct expansion.

PR #4092, opened 2022-04-30
Tarred fix label models

nithinraok:
# What does this PR do ?

Fix for an issue where during tarred dataset, max_steps computation fails in label models

**Collection**: ASR

# Changelog 
- Adds it during setting up of training data
- Deleted duplication code for mapping tarred audio paths
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


titu1994 (approved):
Lgtm, weird that these models have not updated tarred Dataset in so long

nithinraok:
> Lgtm, weird that these models have not updated tarred Dataset in so long

Also I found much of duplication code where we set up tarred dataset very similarly for all ASR and NLO=P models, we need to merge all these to use common funcs

PR #4098, opened 2022-05-02
access mixin

sam1373:
Signed-off-by: sam1373 <samuelkriman@gmail.com>

# What does this PR do ?

Adds mixin for accessing intermediate layers of a model.

**Collection**: asr

# Changelog 

# Usage

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/core/classes/mixins/access_mixins.py
```diff
@@ -0,0 +1,85 @@
...
+
+import torch
+from omegaconf import DictConfig
+
+_ACCESS_CFG = DictConfig({"access_all_intermediate": False, "detach": False, "convert_to_cpu": False})
+
+
+def set_access_cfg(cfg: 'DictConfig'):
+    global _ACCESS_CFG
+    _ACCESS_CFG = cfg
```
titu1994, line 28: Check cfg is not None and is instance (dict)

nemo/core/classes/mixins/access_mixins.py
```diff
@@ -0,0 +1,85 @@
...
+
+
```
titu1994, line 30: This seems good, but needs a way to enable/disable tensor registry - just because the code exists and config exists we should not always register tensors. 

Also handling it via a config flag seems bad cause you would need to update the config and call set_access to update the value. 

What I'd suggest is 
1) remove the config value to enable or disable the registry 
2) setup a global variable with class method to access/update whether the registry should be enabled or not. 
3) in the constructor, call the set config and inside forward of module, call the accessor to see if enabled or not. 
4) config should no longer control whether it's enabled or not though.

titu1994 (commented):
Preliminary review

nemo/collections/asr/parts/submodules/jasper.py
```diff
@@ -1044,6 +1045,9 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]):
         if self.res is not None and self.dense_residual:
             return xs + [out], lens
 
+        if self.access_cfg.get('access_all_intermediate', False):
```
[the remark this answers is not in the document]
sam1373: I believe xs is the list of prev. outputs we're keeping for residual connections, so we only want the last out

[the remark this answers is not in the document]
titu1994: Ah ok. Makes sense.

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -126,7 +126,7 @@ def forward(self, x, att_mask=None, pos_emb=None, pad_mask=None):
             # Call the adapters
             x = self.forward_enabled_adapters(x)
 
-        if self.access_cfg.get('access_all_intermediate', False):
+        if self.access_enabled:
```
titu1994, line 129: Use a function not a global flag please.

titu1994 (approved):
This looks great, thanks !

Issue #4100, opened 2022-05-03
Runtime error when exporting FastPitch model to ONNX


godspirit00:
I tried to export a FastPitch model I trained to ONNX with [export.py](https://github.com/NVIDIA/NeMo/blob/main/scripts/export.py).
The command I used was:
`python scripts/export.py "/home/xxx/TTSs/Nemo-models/nancy_fastpitch-44k-new-v3.nemo" "nancy_fastpitch-44k-new-v3.onnx" --runtime-check --device="cpu" --autocast`
But it produced an error:
```
Traceback (most recent call last):
  File "scripts/export.py", line 169, in <module>
    nemo_export(sys.argv[1:])
  File "scripts/export.py", line 158, in nemo_export
    raise e
  File "scripts/export.py", line 139, in nemo_export
    output_example = forward_method(model)(*input_list, **input_dict)
  File "/mnt/d/TTSs/NeMo/nemo/collections/tts/models/fastpitch.py", line 576, in forward_for_export
    return self.fastpitch.infer(text=text, pitch=pitch, pace=pace, speaker=speaker)
  File "/mnt/d/TTSs/NeMo/nemo/collections/tts/modules/fastpitch.py", line 288, in infer
    enc_out, enc_mask = self.encoder(input=text, conditioning=spk_emb)
  File "/home/tony/Envs/nemo/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/mnt/d/TTSs/NeMo/nemo/collections/tts/modules/transformer.py", line 262, in forward
    return self._forward(self.word_emb(input), (input != self.padding_idx).unsqueeze(2), conditioning)  # (B, L, 1)
  File "/home/tony/Envs/nemo/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/tony/Envs/nemo/lib/python3.8/site-packages/torch/nn/modules/sparse.py", line 158, in forward
    return F.embedding(
  File "/home/tony/Envs/nemo/lib/python3.8/site-packages/torch/nn/functional.py", line 2044, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: Inference tensors do not track version counter.
```
when it was `Getting output example`.

What am I missing? Thank you!

itzsimpl:
[...]

Issue #4012, opened 2022-04-15
nemo2riva of tts_en_fastpitch:1.4.0 fails with CUDA out of memory


itzsimpl:
@borisfom As per comment https://github.com/NVIDIA/NeMo/issues/3619#issuecomment-1089551439, I'm re-opening the bug as the issue persists with `nemo:22.01` (1.7.1), `nemo:1.7.2` and `nemo:1.9.0rc0` (e4ee26b601a8e6f907f926019367d3d545c62f11) combined with `Riva:2.0.0`. I tried both by starting off with `nemo:22.01` and upgrading to more recent versions of nemo, as well as building new containers from the specific GitHub repo tag. I tested with T4 (16GB) and A100 (40GB) GPUs. On all occasions the conversion of [tts_en_fastpitch_align.nemo](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_en_fastpitch) fails with `RuntimeError: CUDA out of memory`.

itzsimpl:
A little bit of update. The most recent version of nemo I could convert [tts_en_fastpitch_align.nemo](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_en_fastpitch) was `nemo:1.5.1`. Reviewing the changes `1.6.2` brought, the offending change seems to be in 
https://github.com/NVIDIA/NeMo/blob/7da391634e276df7273b4da1443aa1c5b3221a32/nemo/collections/tts/models/fastpitch.py#L473-L494
Changing L473 to
```python
     def input_example(self, max_batch=1, max_dim=44): 
```
the values that were hardcoded in `1.5.1` resolves the issue for [tts_en_fastpitch_align.nemo](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_en_fastpitch). I can even export a model that was built on `nemo:1.8.1`. However, if I apply the same fix in `nemo:1.8.1` and try to convert a model built with `nemo:1.8.1`, the conversion fails with the following trace
```python
...
[NeMo I 2022-04-25 13:35:26 features:259] PADDING: 1
[NeMo I 2022-04-25 13:35:26 features:276] STFT using torch
[NeMo I 2022-04-25 13:35:28 save_restore_connector:209] Model FastPitchModel was successfully restored from /data/data/nemo/fastpitch.nemo.
sz=(1, 44)
[...]

PR #4107, opened 2022-05-04
[TTS] fix for #4012 and #4100

itzsimpl:

# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/tts/models/fastpitch.py
```diff
@@ -511,45 +509,29 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
-        return self._input_types
+        return self.fastpitch.input_types
 
     @property
     def output_types(self):
-        return self._output_types
+        return self.fastpitch.output_types
 
-    def input_example(self, max_batch=1, max_dim=256):
+    def input_example(self, max_batch=1, max_dim=44):
```
blisc, line 534: @borisfom Does it matter what we set max_dim to?

blisc (changes requested):

nemo/collections/tts/models/fastpitch.py
```diff
@@ -511,45 +509,29 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
         """
```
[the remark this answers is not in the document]
itzsimpl: If you'd prefer, I can make two separate PRs, but the rest of the changes are here because, if left as is, the script `scripts/export.py` will fail whenever one uses `--runtime-check` (see https://github.com/NVIDIA/NeMo/issues/4100#issuecomment-1116494057).

nemo/collections/tts/models/fastpitch.py
ArEnSc: I think you want to set this to a power of 2 I would try 64  instead I think 256 probably has a high memory requirement

itzsimpl: I set the value of 44 as this was the value that was hardcoded prior to `nemo:1.6.2`. Empirically testing (inspired by your comment), I see that the maximal value with which export succeeds on a 16GB T4 GPU is 160, hence if a power of 2 is required then the value should be 128?

Can you, please, explain what does the parameter control in the exported model? I.e. what is the difference if one does export with `max_dim=44` vs `max_dim=64` or `max_dim=128` for that matter?

It would really help if @borisfom joined in on the conversation.

ArEnSc: What I am super confused about is how the 256 dimension doesn't fit on a 16gb gpu however. I would test it with 256 again and double check. Perhaps it's because I am using amp.

itzsimpl: Thanks for the explanation, it does go in line with my assumptions. You mention a "Fastpitch's limit is 2048 according to nvidia fast pitch repo". Which repo? [DLE](https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/SpeechSynthesis/FastPitch), NeMo, [NGC](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_en_fastpitch)?

That you are able to convert with `max_input=256` leaves me perplexed, as I was not able to do so even with a 40GB A100 GPU. What is your setup?  I rechecked and using the following configuration export fails: 
- GPU T4 16GB (at the moment I can not run it on a 40GB GPU)
- nemo container built from pytorch:22.04 and this PR
- RIVA 2.0.0
- tts_en_fastpitch v1.4.0 and v1.8.1, dowloaded from (https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_en_fastpitch/version)
- commands run:  	
   1. `nemo2riva --key test tts_en_fastpitch_align.nemo --out tts_en_fastpitch.riva --max_dim=256`
   2. `python3 scripts/export.py ython3 tts_en_fastpitch_align.nemo tts_en_fastpitch.onnx --max-dim 256`

You mention that you are using amp, where? I don’t see any parameters in `nemo2riva` or `export.py` for that. If you, however, mean during training, the same error occurs on my custom models that were trained with `precision:16` and `bf16`.

ArEnSc: Right I don't use the nemo variant of fastpitch I use the R and D variant so it's likely configured differently. 
https://github.com/NVIDIA/DeepLearningExamples I use FP from this repo

ArEnSc: @itzsimpl yeah this worked for me strange I do have a 3080 but I tried it on cpu and works at 128

PR #4125, opened 2022-05-06
bug_fix_diarization_manifest_creation

yzhang123:

bug fix for paths that contain "." in basename


nithinraok (approved):
Thanks

PR #4127, opened 2022-05-07
Multiprocess improvements

nithinraok:
# What does this PR do ?

Improve multiprocessing speed by reading audio file only once, and also remove type_check decorator for forward_to_export function in label_models.py 

**Collection**: ASR

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -251,7 +251,6 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
             "embs": NeuralType(('B', 'D'), AcousticEncodedRepresentation()),
         }
 
-    @typecheck()
     def forward_for_export(self, processed_signal, processed_signal_len):
         encoded, length = self.encoder(audio_signal=processed_signal, length=processed_signal_len)
         logits, embs = self.decoder(encoder_output=encoded, length=length)
--- a/scripts/speaker_tasks/scp_to_manifest.py
+++ b/scripts/speaker_tasks/scp_to_manifest.py
@@ -48,7 +48,7 @@ Args:
 --min_spkrs_count: min number of samples per speaker to consider and ignore otherwise
 """
 
-DURATIONS = [1, 2, 3, 4]
+DURATIONS = sorted([1, 2, 3, 4], reverse=True)
 MIN_ENERGY = 0.01
 CWD = os.getcwd()
 SAMPLE_RATE = 16000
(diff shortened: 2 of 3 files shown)


scripts/speaker_tasks/scp_to_manifest.py
```diff
@@ -146,52 +173,28 @@ def get_labels(lines):
...
+        out_file = os.path.splitext(scp)[0] + '_manifest.json'
+        write_file(out_file, lines, range(len(lines)))
     else:
         lines = read_manifest(manifest)
 
     lines = process_map(get_duration, lines, chunksize=100)
 
     if create_chunks:
-        print("creating chunk")
+        print("creating and writing chunks to {}".format(CWD))
```
tango4j, line 190: https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md
10. F-strings are preferred to formatted strings.

I recently figured that you have already used lots of str.format() method here and there.  Following the NeMo contributing guide, I suggest you to use f-string. Which makes -> `print(f"creating and writing chunks to {CWD}"`

str.format() appeared in python 2.6.

scripts/speaker_tasks/scp_to_manifest.py
```diff
@@ -31,46 +30,74 @@
 This scipt converts a scp file where each line contains  
 <absolute path of wav file> 
 to a manifest json file. 
+Optionally post processes the manifest file to create dev and train split for speaker embedding 
+training, also optionally chunk an audio file in to segments of random DURATIONS and create those
+wav files in CWD. 
+
+While creating chunks, if audio is not sampled at 16Khz, it resamples to 16Khz and write the wav file.
 Args: 
 --scp: scp file name
```
tango4j, line 39: I think this is a good time for doing "scp-> filelist" refactor, although you need to change the name of the file.
Two problems: 
(1) What you are using is not a SCP file.
(2) Even if it is SCP file, SCP file format is very hard to find the description and there is no explanation about scp file format in NeMo. 

I believe this script is quite important to prepare dataset for training speaker embedding model. I think changing the name is necessary.

tango4j (changes requested):

scripts/speaker_tasks/scp_to_manifest.py
nithinraok: Thanks tj, updated the code to use f-string.

scripts/speaker_tasks/scp_to_manifest.py
nithinraok: Updated

nemo/collections/asr/models/label_models.py
```diff
@@ -251,7 +251,6 @@ def output_types(self) -> Optional[Dict[str, NeuralType]]:
             "embs": NeuralType(('B', 'D'), AcousticEncodedRepresentation()),
         }
 
-    @typecheck()
```
titu1994, line 254: You can just replace your forwarded last two lines with calling forward_for_export() there. Do that in a later pr

titu1994 (commented):
I've only skimmed the PR but it seems fine. Finish resolving all of @tango4j requirements then merge v

tango4j (approved):

PR #4145, opened 2022-05-10
[Add] support for prepending data prefix

stevehuang52:
# Add support for prepending data prefix to manifests paths and `audio_paths`

This feature is to add support for automatically adding prefix to manifest paths as well as the `audio_filepath` field in the manifest. The simple trick is by adding a `data_prefix` in the top level of your yaml config file. The program has a default value of `None` for  `data_prefix`, so that it's compatible with all previous config files.

# Features
- Adding single prefix to all input manifests. Each train/val/test dataset can have their own data prefix.
- Add different prefix to different input manifests, where the number of specified prefix must match the number of input manifests.
- Fully compatible with previous manifest, since `data_prefix=None` by default.

# Why Data Prefix?

Consider the following dataset folder:
```
/home/steve/libri_speech/
----------------------------->: train_manifest.json
----------------------------->: wav/
------------------------------------->: 11111.wav
------------------------------------->: 22222.wav
```
and the content of ` train_manifest.json` is:
```
{"audio_filepath": "/home/steve/libri_speech/wav/11111.wav", duration: "2.3",  "text": "one one one one one"}
```

If say, Kevin, wants to use my dataset on his project by copying my folder : `cp -r /home/steve/libri_speech/  /home/kevin/`, then he must also change all the `audio_filepath` in  the manifest `/home/steve/libri_speech/train_manifest.json`, which is certainly a tedious work.

However, if we allow `audio_filepath` to be relative paths to the data folder, then we can have a more flexible `train_manifest.json`:
```
{"audio_filepath": "wav/11111.wav", duration: "2.3",  "text": "one one one one one"}
```

Also, when added with a `data_prefix`, we can move the dataset to anywhere without modifying the manifest. For example, all Kevin has to do is simply:
```
python train.py \
    --config-path=../conf/conformer --config-name=config \ 
    model.train_ds.data_prefix=/home/kevin/libri_speech \
    model.train_ds.manifest_filepath="train_manifest.json"
```

And the program will automatically resolve the path for manifest as `/home/kevin/libri_speech/train_manifest.json`, and all `audio_filepath` in the manifest will also be added with the prefix: 
```
{"audio_filepath": "wav/11111.wav", duration: "2.3",  "text": "one one one one one"}
becomes
{"audio_filepath": "/home/kevin/libri_speech/wav/11111.wav", duration: "2.3",  "text": "one one one one one"}

```

# Example Usage

Take the `examples/asr/conf/config.yaml` as an example. We just need to add `data_prefix` as follows:
```yaml
name: &name "QuartzNet15x5"
sample_rate: &sample_rate 16000
repeat: &repeat 1
dropout: &dropout 0.0
separable: &separable true
labels: &labels [" ", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m",
         "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z", "'"]

data_prefix: null

model:
  train_ds:
    data_prefix: null
    manifest_filepath: ???
    .......(other params)

  validation_ds:
    data_prefix: null
    manifest_filepath: ???
    ......(other params)
```

Then suppose the data folder is `/home/heh/libri_speech`, and the `audio_filepath` are relative paths to the data folder, we can run the program by:

```bash
python speech_to_text_ctc.py \
    --config-path=../conf --config-name=config \
    model.train_ds.data_prefix="/home/heh/libri_speech" \
    model.train_ds.manifest_filepath="train_manifest.json" \
    model.validation_ds.data_prefix="/home/heh/libri_speech" \
    model.validation_ds.manifest_filepath="train_manifest.json" \
    trainer.devices=2 \
    trainer.max_epochs=2 \
```

# Change Log
- [5/10/2022] Modified config example: remove global `data_prefix`, and let `model.train_ds`, `model.validation_ds` and `model.test_ds` have their own `data_prefix`.
- [5/10/2022] Prioritize absolute path for manifest over relative path resolution. Now the `item_iter()` first try to locate the original manifest file, and will try to prepend `data_prefix` to the manifest path only if the file is not found at the original manifest path.

# Suggested Future Change
- In data downloading scripts, change all absolute `audio_filepath` to relative paths.

### DIFF
--- a/nemo/collections/asr/data/audio_to_text.py
+++ b/nemo/collections/asr/data/audio_to_text.py
@@ -121,11 +121,13 @@ class ASRManifestProcessor:
         eos_id: Optional[int] = None,
         pad_id: int = 0,
         index_by_file_id: bool = False,
+        data_prefix: Optional[str] = None
     ):
         self.parser = parser
 
         self.collection = collections.ASRAudioText(
             manifests_files=manifest_filepath,
+            data_prefix=data_prefix,
             parser=parser,
             min_duration=min_duration,
             max_duration=max_duration,
@@ -259,12 +261,17 @@ class _AudioTextDataset(Dataset):
         eos_id: Optional[int] = None,
         pad_id: int = 0,
         return_sample_id: bool = False,
+        data_prefix: Optional[str] = None
     ):
         if type(manifest_filepath) == str:
             manifest_filepath = manifest_filepath.split(",")
 
+        if isinstance(data_prefix, str):
+            data_prefix = data_prefix.split(",")
+
         self.manifest_processor = ASRManifestProcessor(
             manifest_filepath=manifest_filepath,
+            data_prefix=data_prefix,
             parser=parser,
             max_duration=max_duration,
             min_duration=min_duration,
@@ -370,6 +377,7 @@ class AudioToCharDataset(_AudioTextDataset):
         pad_id: int = 0,
         parser: Union[str, Callable] = 'en',
         return_sample_id: bool = False,
+        data_prefix: Optional[str] = None,
     ):
         self.labels = labels
 
@@ -391,6 +399,7 @@ class AudioToCharDataset(_AudioTextDataset):
             eos_id=eos_id,
             pad_id=pad_id,
             return_sample_id=return_sample_id,
+            data_prefix=data_prefix,
         )
 
 
--- a/nemo/collections/common/parts/preprocessing/collections.py
+++ b/nemo/collections/common/parts/preprocessing/collections.py
@@ -203,7 +203,7 @@ class ASRAudioText(AudioText):
         """
 
         ids, audio_files, durations, texts, offsets, speakers, orig_srs, langs = [], [], [], [], [], [], [], []
-        for item in manifest.item_iter(manifests_files, data_prefix):
+        for item in manifest.item_iter(manifests_files, data_prefix=data_prefix):
             ids.append(item['id'])
             audio_files.append(item['audio_file'])
             durations.append(item['duration'])
--- a/nemo/collections/common/parts/preprocessing/collections.py
+++ b/nemo/collections/common/parts/preprocessing/collections.py
@@ -192,7 +192,7 @@ class AudioText(_Collection):
 class ASRAudioText(AudioText):
     """`AudioText` collector from asr structured json files."""
 
-    def __init__(self, manifests_files: Union[str, List[str]], *args, **kwargs):
+    def __init__(self, manifests_files: Union[str, List[str]], data_prefix: Union[str, List[str]] = None, *args, **kwargs):
         """Parse lists of audio files, durations and transcripts texts.
 
         Args:
@@ -203,7 +203,7 @@ class ASRAudioText(AudioText):
         """
 
         ids, audio_files, durations, texts, offsets, speakers, orig_srs, langs = [], [], [], [], [], [], [], []
-        for item in manifest.item_iter(manifests_files):
+        for item in manifest.item_iter(manifests_files, data_prefix):
             ids.append(item['id'])
             audio_files.append(item['audio_file'])
             durations.append(item['duration'])
--- a/nemo/collections/common/parts/preprocessing/manifest.py
+++ b/nemo/collections/common/parts/preprocessing/manifest.py
@@ -32,8 +32,9 @@ class ManifestEN:
 
 
 def item_iter(
-    manifests_files: Union[str, List[str]], data_prefix: Optional[Union[str, List[str]]] = None, 
-    parse_func: Callable[[str, Optional[str]], Dict[str, Any]] = None
+    manifests_files: Union[str, List[str]], 
+    parse_func: Callable[[str, Optional[str]], Dict[str, Any]] = None,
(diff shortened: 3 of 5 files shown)


titu1994 (changes requested):
If the goal of the PR is to have a "shared" prefix path for manifests with relative paths, then this is not a good solution. 

1) you would need to update dozens of ASR configs for this to be useful 
2) the prefix path is outside the model config, doesn't get saved
3) the implementation is updating item_iter() signature which will break many calls to it. 

Overall it seems redundant. Manifest paths are meant to point to data, relative path resolution can be done in other ways without user action - such as use cwd() is a safe and easy alternative which doesn't cause so many changes. Even that should be optional - check if the filepath exists when merging cwd()/relative file path. Otherwise error out.

stevehuang52:
- The goal is to provide a prefix path for both manifest AND the `audio_filepath` in manifests. Actually the latter is more important. For example, if I downloaded a dataset (e.g., `get_librispeech_data.py`) in one folder, and then I want to move the dataset to another dataset, then the previously generated `audio_filepath` are no longer valid.
- The proposed feature should not break other modules, since the new param is just optional and by default set to none. 
- I'm not sure whether cwd() can solve all the problems, not sure if it can prepend prefix to the `audio_filepath` fields in manifest. It's ok for company users to use absolute path, since we have unified storage, but not the same thing for researchers everywhere.

titu1994:
- Updating manifest can be done in any text editor
- New param may be optional but it involves updating every single config of ASR. Look at conf folder and there's already 2 dozen configs there. If it's not part of the config then people won't know to use it. 
- We have never encouraged use of relative paths in ASR. It's not about company or not, absolute manifest is more useful because you can move around your manifest detached from your dataset - and your absolute path can be a mounted data path for docker containers.

stevehuang52:
I agree that it's better to keep manifest with absolute path. I have updated the code so that it will first try to locate the manifest without adding the prefix. If the manifest is not found, then the program will add the prefix and try to find it with the new path.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -370,6 +377,7 @@ def __init__(
         pad_id: int = 0,
         parser: Union[str, Callable] = 'en',
         return_sample_id: bool = False,
```
titu1994, line 379: What about subword datasets ? And Dali datasets ?

nemo/collections/common/parts/preprocessing/collections.py
```diff
@@ -203,7 +203,7 @@ def __init__(self, manifests_files: Union[str, List[str]], *args, **kwargs):
         """
 
         ids, audio_files, durations, texts, offsets, speakers, orig_srs, langs = [], [], [], [], [], [], [], []
-        for item in manifest.item_iter(manifests_files):
+        for item in manifest.item_iter(manifests_files, data_prefix=data_prefix):
```
titu1994, line 206: Changing this signature requires changes in all data loaders and classes that call this signature. Try to revert or redesign this part

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -32,7 +33,9 @@ def __init__(self, *args, **kwargs):
 
 
 def item_iter(
-    manifests_files: Union[str, List[str]], parse_func: Callable[[str, Optional[str]], Dict[str, Any]] = None
+    manifests_files: Union[str, List[str]], 
+    parse_func: Callable[[str, Optional[str]], Dict[str, Any]] = None,
+    data_prefix: Optional[Union[str, List[str]]] = None,
```
titu1994, line 38: Above comment. Try not to change this signature

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -61,17 +67,41 @@ def item_iter(
...
+             f"Got {num_manifests} manifests with {num_prefix} prefix(es)."
+        )
+
     if parse_func is None:
         parse_func = __parse_item
 
     k = -1
-    for manifest_file in manifests_files:
+    for manifest_file, prefix in zip(manifests_files, data_prefix):
+        if not Path(manifest_file).is_file():
```
titu1994, line 90: This step is slow - it will have an effect when passing ASRSet to do this repeatedly

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -61,17 +67,41 @@ def item_iter(
...
+            else:
+                raise ValueError(f"Manifest not found: {manifest_file}")
+
         with open(expanduser(manifest_file), 'r') as f:
             for line in f:
                 k += 1
                 item = parse_func(line, manifest_file)
                 item['id'] = k
-
+                item['audio_file'] = str(Path(prefix) / Path(item['audio_file']))
```
titu1994, line 104: Don't override value of output of parse_func, it will cause errors if output of func is unexpectedly post processed. See if you can modify the functions itself.

titu1994 (commented):
I'm don't think this is a solid approach, it introduced many corner cards, it's use case is not recommended and it will slow down processing of huge manifests (aka ASRSet).

titu1994:
Closing due to inactivity. Please recreate new pr if required.

PR #4185, opened 2022-05-17
Speakertasks manifest

yzhang123:
adding optional duration to manifest, so we can use it with speech explorer


scripts/speaker_tasks/pathsfiles_to_manifest.py
```diff
@@ -120,11 +132,14 @@ def main(wav_path, text_path=None, rttm_path=None, uem_path=None, ctm_path=None,
         if ctm is not None:
             ctm = ctm.strip()
 
+        duration = None
+        if add_duration and sample_rate:
+            duration = librosa.get_duration(filename=audio_line, sr=sample_rate)
```
nithinraok, line 137: Since, we support only inference of 16kHz files, you may pass a default value here.

nithinraok (changes requested):

scripts/speaker_tasks/pathsfiles_to_manifest.py
yzhang123: done

scripts/speaker_tasks/pathsfiles_to_manifest.py
```diff
@@ -120,11 +125,14 @@ def main(wav_path, text_path=None, rttm_path=None, uem_path=None, ctm_path=None,
         if ctm is not None:
             ctm = ctm.strip()
 
+        duration = None
+        if add_duration:
+            duration = librosa.get_duration(filename=audio_line, sr=16000)
```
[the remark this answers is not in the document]
nithinraok: If it doesn't we can use something like this:
`y,sr = librosa.load(audio,sr=None)
librosa.get_duration(y=y, sr=sr)`

nithinraok (approved):

PR #4189, opened 2022-05-18
Add support for Decoder + Joint Adapters for ASR

titu1994:
# What does this PR do ?

Add support for decoder + joint adapters, alongside encoder adapters. Supports CTC + RNNT decoders, RNNT Joint, RNNT Decoder + Joint adapter.

**Collection**: [ASR]

# Changelog 
- Support CTC Decoder, RNNT Decoder, RNNT Joint adapter support (module level).
- Update tests to check new modules
- Add support for property that defines list of valid modules that support adapters.

# Usage

```python
model.add_adapter("decoder:<adapter_name>", cfg=...)  # decoder only adapter; CTC/RNNT
model.add_adapter("joint:<adapter_name>", cfg=...)  # joint only adapter; RNNT only
model.add_adapter("decoder+joint:<adapter_name>", cfg=...)  # decoder + joint adapter; RNNT only
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


examples/asr/asr_adapters/train_asr_adapter.py
```diff
@@ -71,8 +73,12 @@
 from nemo.utils.exp_manager import exp_manager
 
 
-def update_encoder_config_to_support_adapter(model_cfg):
+def update_model_config_to_support_adapter(model_cfg, current_cfg):
     with open_dict(model_cfg):
+        # Override prediction logging in config
+        model_cfg.log_prediction = current_cfg.model.get('log_prediction', False)
```
VahidooX, line 79: What the purpose of this line?

examples/asr/conf/asr_adapters/asr_adaptation.yaml
```diff
@@ -57,10 +57,12 @@ model:
   pretrained_model: null  # name of a pretrained model
   nemo_model: null  # path to a ASR model file (.nemo)
 
+  log_prediction: false # enables logging sample predictions in the output during training
+
   adapter:
     # Config of the adapter training/eval script.
     adapter_name: ???  # Name of the adapter, used by the script
-    adapter_module_name: null  # Name of the preprocessing module, if required. Defaults to None, equivalent to ''.
+    adapter_module_name: null  # Name of the adapter module. Combine multiple modules with '+' between module names.
```
VahidooX, line 65: Are we going to have more detail about this new support in the docs or tutorials?

nemo/collections/asr/modules/conv_asr.py
```diff
@@ -463,6 +469,31 @@ def _prepare_for_export(self, **kwargs):
...
+            if in_planes != self._feat_in:
+                logging.info(
+                    f"Updating {self.__class__.__name__} Adapter input dim from {in_planes} " f"to {self._feat_in}"
+                )
+                in_planes = self._feat_in
+
+            cfg['in_features'] = in_planes
+            return cfg
+        else:
+            raise ValueError(
```
VahidooX, line 492: What are the cases where in_features is missing in the config? cases like user forgets to have it?

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1127,3 +1183,11 @@ def forward(self, encoder_outputs, decoder_inputs, decoder_lengths, state_h, sta
         state_h, state_c = decoder_outputs[2][0], decoder_outputs[2][1]
         joint_output = self.joint(encoder_outputs, decoder_output)
         return (joint_output, decoder_length, state_h, state_c)
+
+
+# Add the adapter compatible modules to the registry
+if adapter_mixins.get_registered_adapter(RNNTDecoder) is None:
```
VahidooX, line 1189: Is there a way to have the list of these three parts in a list and then a loop over them instead of separate block for each one?

nemo/collections/asr/parts/mixins/asr_adapter_mixins.py
```diff
@@ -56,6 +56,12 @@ def setup_adapters(self):
         if hasattr(self, 'encoder') and isinstance(self.encoder, AdapterModuleMixin):
             supports_adapters |= True
 
+        if hasattr(self, 'decoder') and isinstance(self.decoder, AdapterModuleMixin):
```
VahidooX, line 59: How about merging these three in one if block?

VahidooX: There are a bunch of such if blocks, not sure if it is a good idea to merge all, just a suggestion to consider.

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -834,3 +834,17 @@ def update_adapter_cfg(self, cfg: DictConfig):
...
+        List of valid adapter modules that are supported by the model.
+
+        **Note**: Subclasses should override this property and return a list of str names, of all the modules
+            that they support, which will enable users to determine where to place the adapter modules.
+
+        Returns:
+            A list of str, one for each of the adapter modules that are supported. By default, the subclass
+            should support the "global adapter" ('').
+        """
+        return ['']
```
VahidooX, line 850: it is considered to be an empty list?

VahidooX (changes requested):

examples/asr/conf/asr_adapters/asr_adaptation.yaml
```diff
@@ -82,7 +84,9 @@ model:
     # This can be useful in order to select *which type of adapter* is added, *what adapters to enable*,
     # and further global operations that can decide dynamically how to support the requested adapter.
     global_cfg:
-      encoder_adapter: True  # ASR adapter key, determines whether to add encoder adapter modules
+      check_encoder_adapter: False  # ASR adapter key, determines whether to check if encoder adapter modules is supported
```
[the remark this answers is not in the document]
titu1994: These flags will now only check for the value rather than actually determine if the adapter is added or not. The issue becomes if they test a ctc model for joint adapters (not possible), then the adapter script will crash. During restoration, if adapters are not supported then restore will fail even if all adapters are disabled. So it is tricky.

I can do one thing, which is instead of crashing or raising error, I can do a one time log stating that while checking, the module was found to not support adapters.

nemo/collections/asr/modules/conv_asr.py
titu1994: In case users forget to add it, or if they use an adapter module which doesn't have the keyword arg `in_features`. For example, a user created adapter module that does not use in_features but input_dim as name. It is a little brittle, but covers the most common case.

nemo/collections/asr/modules/rnnt.py
```diff
@@ -48,7 +49,7 @@
 from nemo.utils import logging
 
 
-class RNNTDecoder(rnnt_abstract.AbstractRNNTDecoder, Exportable):
+class RNNTDecoder(rnnt_abstract.AbstractRNNTDecoder, Exportable, adapter_mixins.AdapterModuleMixin):
```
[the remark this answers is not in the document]
titu1994: Sure sounds fine.

nemo/collections/asr/modules/rnnt.py
titu1994: Yes possible.

nemo/collections/asr/parts/mixins/asr_adapter_mixins.py
titu1994: Hmm, seperating them makes it clearer. For example, you could have module adapters which are more specific - pre-decoder or pre-lstm or post-lstm decoder adapters. Covering all cases in one giant list would be problematic. This gives freedom to the dev to catch whatever case they want.

nemo/core/classes/mixins/adapter_mixins.py
titu1994: It is considered the "default" and "global" adapter cases. Each domain can chose to support such "default" or "global" adapter cases. For example, ASR allows default adapters, which default to encoder adapters. They are also global in the sense that you don't provide a module name to them.

examples/asr/conf/asr_adapters/asr_adaptation.yaml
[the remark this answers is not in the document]
titu1994: Done

nemo/collections/asr/modules/conv_asr.py
titu1994: I will simplify this method.

nemo/collections/asr/modules/rnnt.py
titu1994: Done

VahidooX (approved):
LGTM!

PR #4196, opened 2022-05-18
Fix the server key value problem in the notebook

yidong72:
Fix the key error in Tabular Data generation caused by the change of default prompts.


ericharper (approved):
LGTM. Thanks!

PR #4209, opened 2022-05-19
Tacotron2 1.9.0 bugfixes

redoctopus:
# What does this PR do ?

Update Tacotron2 training tutorial and config to work with r.1.9.0

**Collection**: TTS

# Changelog 
- Updated 44kHz config to use TTSDataset (now matches with 22kHz config)
- Added file downloads necessary for training to the Tacotron2 notebook
- Minor bugfix to avoid device mismatch during training for the notebook

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/tts/conf/tacotron2_44100.yaml
+++ b/examples/tts/conf/tacotron2_44100.yaml
@@ -1,116 +1,137 @@
-# TODO(Oktai15): update this config in 1.8.0 version
+# This config contains the default values for training Tacotron2 model on LJSpeech dataset.
+# If you want to train model on other dataset, you can change config values according to your dataset.
+# Most dataset-specific arguments are in the head of the config file, see below.
+# This config is for training with sample rate 44.1kHz.
 
 name: Tacotron2
-sample_rate: 44100
-# <PAD>, <BOS>, <EOS> will be added by the tacotron2.py script
-labels:
-- ' '
-- '!'
-- '"'
-- ''''
-- (
-- )
-- ','
-- '-'
-- .
-- ':'
-- ;
-- '?'
-- a
-- b
-- c
-- d
-- e
-- f
-- g
-- h
-- i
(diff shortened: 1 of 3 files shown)


examples/tts/conf/tacotron2_44100.yaml
```diff
@@ -1,116 +1,137 @@
-# TODO(Oktai15): update this config in 1.8.0 version
+# This config contains the default values for training Tacotron2 model on LJSpeech dataset.
```
blisc, line 1: Why do we still say LJSpeech here?

blisc (approved):

examples/tts/conf/tacotron2_44100.yaml
redoctopus: Specifically because of the pitch min/max statistics that are in there as a default. The other configs also have the same default statistics; do you think we should update them across the board to use the function default or have the user input something?

blisc: The pitch fmin and fmax no longer change per model. Tacotron2 isn't pitch conditioned anyways so why do we have this?

redoctopus: Ah, good point--I'll remove it from both configs.

What do you mean by it no longer changes per model? The TTSDataset takes them in for `add_pitch()`, is this deprecated? https://github.com/NVIDIA/NeMo/blob/71c51ec24d0b2774dcdd7dcbbf91c6875ce5383b/nemo/collections/tts/torch/data.py#L358-L359

redoctopus: Ahh, gotcha. Pushing updates.

At some point we should update the other configs to remove these values as well, I'll make a note of it.

PR #4235, opened 2022-05-23
Fix JSON serialization problem

yidong72:
It fixes the text generation server cannot `jsonify` the torch tensor object.


ericharper (approved):
LGTM.

PR #4260, opened 2022-05-25
Fixing the speed slow-down for speech models.

VahidooX:
Signed-off-by: vahidoox <vnoroozi@nvidia.com>

# What does this PR do ?

There was a slow down in speech models when upgraded to PTL 1.6. They changed the default of trainer.benchmark to True which slows down model with variable input lengths including speech models. This PR set all the trainer.benchmark to false for all speech models.

# Changelog 
Fixed the speed slow-down for speech models by setting benchmark=false for all speech configs.

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


titu1994 (approved):
Looks great, thanks !

PR #4269, opened 2022-05-26
Fix missing validation dataset, whitelist certain keywords for datasets

titu1994:

# What does this PR do ?

Fix adapters training script when there are no validation configs.

**Collection**: [ASR]

# Changelog 
- Fix validation config bug 
- Improve the handling of dataloader speeds in ASR

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/asr/asr_adapters/train_asr_adapter.py
+++ b/examples/asr/asr_adapters/train_asr_adapter.py
@@ -78,8 +78,15 @@ def update_encoder_config_to_support_adapter(model_cfg):
 
 
 def update_model_cfg(original_cfg, new_cfg):
-    with open_dict(new_cfg):
-        # drop keys which dont exist in old config
+    with open_dict(original_cfg), open_dict(new_cfg):
+        # force inject some keys into the config
+        whitelist_keys = ['num_workers', 'pin_memory']
+        for wkey in whitelist_keys:
+            if wkey in new_cfg:
+                original_cfg[wkey] = new_cfg[wkey]
+                print(f"Injecting white listed key `{wkey}` into config")
+
+        # drop keys which dont exist in old config and are not whitelisted
         new_keys = list(new_cfg.keys())
         for key in new_keys:
             if key not in original_cfg:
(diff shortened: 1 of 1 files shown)


examples/asr/asr_adapters/train_asr_adapter.py
```diff
@@ -134,11 +141,11 @@ def main(cfg):
 
     # Setup model for finetuning (train and validation only)
     cfg.model.train_ds = update_model_cfg(model.cfg.train_ds, cfg.model.train_ds)
-    cfg.model.validation_ds = update_model_cfg(model.cfg.validation_ds, cfg.model.validation_ds)
-
-    # Call the dataloaders and optimizer + scheduler
     model.setup_training_data(cfg.model.train_ds)
-    model.setup_multiple_validation_data(cfg.model.validation_ds)
+
```
VahidooX, line 145: How about the test_ds? Does this model supports test?

examples/asr/asr_adapters/train_asr_adapter.py
```diff
@@ -78,8 +78,15 @@ def update_encoder_config_to_support_adapter(model_cfg):
...
-        # drop keys which dont exist in old config
+    with open_dict(original_cfg), open_dict(new_cfg):
+        # force inject some keys into the config
+        whitelist_keys = ['num_workers', 'pin_memory']
+        for wkey in whitelist_keys:
+            if wkey in new_cfg:
+                original_cfg[wkey] = new_cfg[wkey]
+                print(f"Injecting white listed key `{wkey}` into config")
+
+        # drop keys which dont exist in old config and are not whitelisted
```
VahidooX, line 89: dont -> don't

examples/asr/asr_adapters/train_asr_adapter.py
```diff
@@ -78,8 +78,15 @@ def update_encoder_config_to_support_adapter(model_cfg):
...
 
 def update_model_cfg(original_cfg, new_cfg):
-    with open_dict(new_cfg):
-        # drop keys which dont exist in old config
+    with open_dict(original_cfg), open_dict(new_cfg):
+        # force inject some keys into the config
+        whitelist_keys = ['num_workers', 'pin_memory']
+        for wkey in whitelist_keys:
+            if wkey in new_cfg:
+                original_cfg[wkey] = new_cfg[wkey]
```
VahidooX, line 86: Is it a good idea to change the original_cfg as it is just an input to the model?
It can be used in other places after calling this method.
Also it assumes that original_cfg has both the num_workers and pin_memory already in itself. Is this assumption always true?

examples/asr/asr_adapters/train_asr_adapter.py
VahidooX, line 145: Is it possible to init an ASR model without train_ds?

VahidooX (changes requested):

examples/asr/asr_adapters/train_asr_adapter.py
titu1994: No we dont support it in this script.

examples/asr/asr_adapters/train_asr_adapter.py
titu1994: Thats why this white list is used - these two keys may be missing from older models, therefore we check if the new config (that is provided to train an adaper) contained these keys, if so then override the old config value (whether or not it had them).

This is necessary to prevent old citrinet and contextnet models from training with 0 workers.

examples/asr/asr_adapters/train_asr_adapter.py
titu1994: No, this is a train script, at least train_ds must be provided. Otherwise no point calling this script.

VahidooX (approved):

PR #4277, opened 2022-05-27
[Add] automaticly resolving relative audio path

stevehuang52:
# Add Support for Automatically Resolving Relative Audio Path

This simple modification allows using relative audio paths in the manifest, so that there is no need to change the manifest after moving the dataset to different locations. The updated `iterm_iter` function will first check if the `audio_file` exists, and skip further processing if it does exist. Then the function checks whether `audio_file` is a relative path. If it is, the function will add the parent folder of manifest to the audio file path. For tarred datasets, since the `audio_file` in manifest is not a real path, we will not perform any operation.

This feature is based on the assumption that the manifest is under the same directory as the audio directory, and the relative audio paths always begins with an audio directory (e.g., `wavs/`).

Consider the following dataset folder:
```
/home/steve/libri_speech/
----------------------------->: train_manifest.json
----------------------------->: wavs/
------------------------------------->: 11111.wav
------------------------------------->: 22222.wav
```
and the content of ` train_manifest.json` is:
```
{"audio_filepath": "wavs/11111.wav", duration: "2.3",  "text": "one one one one one"}
```

The updated function will first try to read the file `wavs/11111.wav`, but since it's a relative path it won't find it. Then the function will add the parent directory of the manifest to the audio path, which becomes `/home/steve/libri_speech/wavs/11111.wav`.


titu1994 (approved):
This seems reasonable

stevehuang52:
Any ideas on the failure on `continuous-integration/jenkins/pr-head`?  Thanks.

titu1994:
Nope.

stevehuang52:
Fixed code style and add modification to evaluation scripts.

examples/asr/transcribe_speech.py
```diff
@@ -166,7 +168,10 @@ def main(cfg: TranscriptionConfig) -> TranscriptionConfig:
                     has_two_fields.append(True)
                 else:
                     has_two_fields.append(False)
-                filepaths.append(item['audio_filepath'])
+                audio_file = Path(item['audio_filepath'])
+                if not audio_file.is_file() and not audio_file.is_absolute():
+                    audio_file = manifest_dir / audio_file
+                filepaths.append(str(audio_file))
```
titu1994, line 174: Shouldn't this be audio_file.absolute()

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -230,12 +231,16 @@ def main():
...
     with open(args.input_manifest, 'r') as manifest_file:
         audio_file_paths = []
         for line in tqdm(manifest_file, desc=f"Reading Manifest {args.input_manifest} ...", ncols=120):
             data = json.loads(line)
+            audio_file = Path(data['audio_filepath'])
+            if not audio_file.is_file() and not audio_file.is_absolute():
+                audio_file = manifest_dir / audio_file
             target_transcripts.append(data['text'])
-            audio_file_paths.append(data['audio_filepath'])
+            audio_file_paths.append(str(audio_file))
```
titu1994, line 243: Should it be . absolute()

titu1994 (approved):
Lgtm

PR #4233, opened 2022-05-23
Update TTS Configs from LAMB to AdamW

redoctopus:
Signed-off-by: Jocelyn Huang <jocelynh@nvidia.com>

# What does this PR do ?

Replaces LAMB optimizer (unstable, hangs during multi-GPU training) with AdamW for FastPitch and Mixer-TTS configs.
Second attempt based off of `main` instead of `r1.9.0` (see #4219)

**Collection**: TTS

# Changelog 
- Replace LAMB defaults with AdamW
- Update `fastpitch.py` example script to use existing config

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

PR #4308, opened 2022-06-01
Adjust lr for AdamW from LAMB default

redoctopus:

# What does this PR do ?

Adjusts default learning rate down to 1e-3 due to change in default optimizer to AdamW from LAMB (see PR #4233)

**Collection**: TTS

# Changelog 
- Changes default learning rate from 1e-1 to 1e-3 for FastPitch and Mixer-TTS models

### DIFF
--- a/examples/tts/conf/fastpitch_align_44100.yaml
+++ b/examples/tts/conf/fastpitch_align_44100.yaml
@@ -211,7 +211,7 @@ model:
 
   optim:
     name: adamw
-    lr: 1e-1
+    lr: 1e-3
     betas: [0.9, 0.999]
     weight_decay: 1e-6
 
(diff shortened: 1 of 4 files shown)


subhankar-ghosh (approved):
LGTM

XuesongYang:
Thank you for the change. It would be more helpful if provided any evidence about the LR decision. Does 1e-3 help achieve better loss, better stability, or faster convergence?

redoctopus:
> Thank you for the change. It would be more helpful if provided any evidence about the LR decision. Does 1e-3 help achieve better loss, better stability, or faster convergence?

The lower lr should make this significantly more stable; 1e-1 is (from a recent) FastPitch experiment much too high for the model to learn at all.

PR #4317, opened 2022-06-02
Option to disable mp in VAD via num_workers=1

gkucsko:
In certain environments python multiprocessing can deadlock. This adds a convenient version to disable by setting num_workers to 1.


gkucsko:
@tango4j mind having a quick look? thanks!

tango4j:
> @tango4j mind having a quick look? thanks!

Could you change the line to add `if config['num_workers'] is not None and num_worker > 1:` to prevent error case to prevent error from `config['num_workers']=None`?

gkucsko:
done, thanks. anything else?

tango4j:
> 

It is working fine with the change, but not passing CI.  

> done, thanks. anything else?

Can you add line 693 if num_workers is not None and num_workers > 1: too ?

gkucsko:
sure thing!

fayejf (approved):
LGTM! Thank you so much!

fayejf:
Yes seems like python multiprocessing can deadlock in certain environments, and this is an unsolved issue in python now. Thanks for the fix! @gkucsko 

For anyone is interested in using multiprocessing in VAD to accelerate the process. I'm with Python 3.7.11 and it works perfectly fine on my end

PR #4349, opened 2022-06-07
fix wrong requirement

yzhang123:

sacremoses is used and imported in [common/tokenizers](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/common/tokenizers/moses_tokenizers.py#L17)
so need to go into common requirements


MaximumEntropy (approved):
LGTM!

PR #4350, opened 2022-06-07
Fix the post LN bug

yidong72:
Fix a bug about how to do post-ln in the transformer layer. 
image


nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -1319,7 +1320,9 @@ def forward(
 
             layernorm_input = bias_dropout_add_func(attention_output, attention_bias, residual, self.hidden_dropout)
             normalization_output = self.post_inter_attention_layernorm(layernorm_input)
-
+            # Post-LN normalization after residual
+            if self.transformer_block_type == 'post_ln':
+                layernorm_input = normalization_output
```
michalivne, line 1325: Why not set directly `residual` here? why set `layernorm_input`?

michalivne (commented):
See minor comment.

nemo/collections/nlp/modules/common/megatron/transformer.py
yidong72: because later logic set `residual = layernorm_input`

michalivne (approved):
LGTM!

ericharper (approved):
LGTM! Thanks!

PR #4377, opened 2022-06-15
Fix no attribute 'pad_id' bug when pre-processing

yidong72:
Fix the no attribute 'pad_id' when pre-processing the data in the Tabular Data generation notebook.


MaximumEntropy (approved):
LGTM!

PR #4210, opened 2022-05-19
MegaMolBART compatibility

michalivne:
# What does this PR do ?

1. Fixes RegExTokenizer missing attributes.
2. Fixes broken Megatron encoder-decoder LM model ability to receive directly encoder hiddens.

```
    MegatronLMEncoderDecoderModel.forward(
        self,
        encoder_input_ids,
        decoder_input_ids,
        encoder_attn_mask,
        decoder_attn_mask,
        token_type_ids=None,
        lm_labels=None,
        # NEW API
        enc_output=None,
        enc_output_attn_mask=None,
        # NEW API
        output_enc_hidden_only=False,
        enc_input=None,
    ):
```

3. Added MegatronLMEncoderDecoderModel.encode method.
4. Improved decoding speed.

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #4385, opened 2022-06-16
fix a bug in T5 decode function

HeyyyyyyG:
# What does this PR do ?

Fix the bug of T5 decoding given input embeddings. 

**Collection**: [NLP]

# Changelog 
- Updating the way of passing arguments to avoid argument mismatching.
- Delete an argument which is not being utilized.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


MaximumEntropy (changes requested):
Thanks for the PR! We do not want to pass things as kwargs because if the model is wrapped in a Float16Module for `megatron_amp_o2`, it expects positional args to cast to half-precision if needed. @michalivne is working on a decoding refactor to avoid re-computing encoder hidden on each decode step, we can incorporate your changes of not passing `enc_input` in that PR if that works?

michalivne:
The following PR fixes both incorrect kwargs, and more efficient decoding (among other issues).
This PR does not support both tensor parallel and pipeline parallel at the time (only one of those do work)

https://github.com/NVIDIA/NeMo/pull/4210

Issue #4455, opened 2022-06-27
Tar file datasets don't allow for "/"in key names for ASR training.


galv:
**Is your feature request related to a problem? Please describe.**

Tar file datasets have two problems today that I am aware of:

1. The filenames inside tar files are *required* not to have "/" or "." in them. Disallowing "." is fine, given that it has special meaning in webdataset. Disallowing "/" causes some difficulties. First, if your key is larger than 255 elements and you convert "/" to "_" (as is done here https://github.com/NVIDIA/NeMo/blob/daf08760dde35433f934a5c38ab293bad3c22825/scripts/speech_recognition/convert_to_tarred_audio_dataset.py#L560, then it is not possible to untar the created tar file. You will get an error like this when running `untar`:

```
File name too long: 'gov_DOT_uscourts_DOT_illappct_DOT_1-07-3566-1-08-0702-1-08-0746-1-08-1373-1-08-1982-1-08-1983-1-08-2151-1-08-2152_gov_DOT_uscourts_DOT_illappct_DOT_1-07-3566-1-08-0702-1-08-0746-1-08-1373-1-08-1982-1-08-1983-1-08-2151-1-08-2152_DOT_2009-11-12_DOT_mp3_00000.flac'
```

This is because linux file names are at most 255 characters: https://serverfault.com/a/9548

2. Suppose that your tar dataset contains the following 4 entries:

```
speaker1/utterance1.wav
speaker1/utterance2.wav
speaker2/utterance1.wav
speaker2/utterance2.wav
```

Today, NeMo will convert the key `speaker1/utterance1.wav` into `${parent_dir_of_manifest}/speaker1/utterance1.wav`. That is to say that it believes that this makes. The code which does that is here:

https://github.com/NVIDIA/NeMo/blob/d8785e022e2ccaf8b02a27ca084aea5acd257b18/nemo/collections/common/parts/preprocessing/manifest.py#L92-L104

This code assumes that the keys of tar files have no "/" in them. A more straightforward way is to pass another argument "is_tar_dataset" into this function to accept tar files appropriately. I will show this below.

**Describe the solution you'd like**

I will link to a work-in-progress PR for "A more straightforward way is to pass another argument "is_tar_dataset" into this function to accept tar files appropriately." shortly after creating this issue. Edit: https://github.com/NVIDIA/NeMo/pull/4456

**Describe alternatives you've considered**

[...]

PR #4470, opened 2022-06-29
[Fix] Relative Audio Paths

stevehuang52:
Signed-off-by: stevehuang52 <heh@nvidia.com>

This PR fixes the second problem in this [issue 4455](https://github.com/NVIDIA/NeMo/issues/4455).

Previously, the code will fail when there are "/" in the filepaths of tarred datasets. Now we only add the manifest directory to the audio filepath if the resulting file exists, thus fixing the problem.

### DIFF
--- a/nemo/collections/common/parts/preprocessing/manifest.py
+++ b/nemo/collections/common/parts/preprocessing/manifest.py
@@ -89,14 +89,13 @@ def __parse_item(line: str, manifest_file: str) -> Dict[str, Any]:
             f"Manifest file {manifest_file} has invalid json line structure: {line} without proper audio file key."
         )
 
-    # If the audio path is relative, and not using tarred dataset,
-    # attach the parent directory of manifest to the audio path.
-    # Assume "audio_file" starts with a dir, such as "wavs/xxxxx.wav".
-    # If using a tarred dataset, the "audio_path" is like "_home_data_tarred_wavs_xxxx.wav",
-    # so we will just ignore it.
+    # If the audio path is a relative path and does not exists,
+    # try to attach the parent directory of manifest to the audio path.
+    # Revert to the oiginal path if the new path still doesn't exist.
+    # Assume that the audio paths are like "wavs/xxxxxx.wav".
     manifest_dir = Path(manifest_file).parent
     audio_file = Path(item['audio_file'])
-    if not audio_file.is_file() and not audio_file.is_absolute() and audio_file.parent != Path("."):
+    if not audio_file.is_file() and not audio_file.is_absolute():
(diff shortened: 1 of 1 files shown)


nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -89,17 +89,19 @@ def __parse_item(line: str, manifest_file: str) -> Dict[str, Any]:
...
-    # so we will just ignore it.
+    # If the audio path is a relative path and does not exist,
+    # try to attach the parent directory of manifest to the audio path.
+    # Revert to the original path if the new path still doesn't exist.
+    # Assume that the audio path is like "wavs/xxxxxx.wav".
     manifest_dir = Path(manifest_file).parent
     audio_file = Path(item['audio_file'])
-    if not audio_file.is_file() and not audio_file.is_absolute() and audio_file.parent != Path("."):
-        # assume the wavs/ dir and manifest are under the same parent dir
+    if not audio_file.is_file() and not audio_file.is_absolute():
```
titu1994, line 98: To note, I think even this is_file() check will fail for too long file names. Any idea ?

titu1994 (approved):
Overall looks good. Minor comments v

nemo/collections/common/parts/preprocessing/manifest.py
stevehuang52: I think we can try to avoid too long file names. When creating tarred datasets, we can trim the common path of all audio files and use the trimmed paths as the new file names.  

For example,
```
/a/b/c/d/e/xxxxxx.wav
/a/b/c/d/f/yyyyyy.wav
```
can be trimmed as 
```
e/xxxxxx.wav
f/yyyyyy.wav
```

titu1994: No let's not do that, there are many datasets of ASR in mcc and MLS where there is a lot of common folders.

stevehuang52: How about changing the code to the following? We just keep the original path if the input file is 255 chars or longer.
```python
 if (len(str(audio_file) < 255) and not audio_file.is_file() and not audio_file.is_absolute():
```

galv: There might be some confusion here (or maybe I am the confused one!). The 255 character limit refers to individual "nodes" in the file system tree. You can certainly have a path to a file with a  length greater than 255, assuming there are directories leading to it. It's just that each file name and directory name must individually not exceed 255 characters.

galv: Yes, I understand.

PR #4480, opened 2022-06-30
Add support for ASR Adapter Auxiliary Losses

titu1994:
# What does this PR do ?

Add support for auxiliary regularization losses to ASR adapters.

**Collection**: [ASR]

# Changelog 
- Add support for access mixin to all adapter modules
- Add utility methods for gathering the loss from auxiliary models
- Add tests for adapter modules

# Usage

```python

# Currently, setting the config value of `l2_lambda` for the Adapter strategy adds auxiliary losses.
python train_adapters.py \
  ... \
  model.adapter.adapter_strategy.l2_lambda=0.001
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/models/asr_model.py
+++ b/nemo/collections/asr/models/asr_model.py
@@ -66,13 +66,14 @@ class ASRModel(ModelPT, ABC):
 
     def add_auxiliary_losses(self, loss: torch.Tensor, reset_registry: bool = True) -> torch.Tensor:
         """
+        Utility method to enable calculation of auxiliary losses for ASR training.
 
         Args:
-            loss:
-            reset_registry:
+            loss: The output loss value prior to addition with auxiliary losses.
+            reset_registry: Bool, whether to reset the AccessMixin registry after adding auxiliary losses.
 
         Returns:
-
+            Loss tensor used for back propagation.
         """
         # Add adapter auxiliary losses, if registered
         if AccessMixin.is_access_enabled():
--- a/nemo/collections/asr/models/asr_model.py
+++ b/nemo/collections/asr/models/asr_model.py
@@ -20,7 +20,7 @@ import torch
 from nemo.core.classes import ModelPT
 from nemo.core.classes.exportable import Exportable
 from nemo.core.classes.mixins import AccessMixin
-from nemo.utils import model_utils
+from nemo.utils import model_utils, logging
 from nemo.utils.export_utils import cast_all
 
 __all__ = ['ASRModel']
@@ -79,16 +79,17 @@ class ASRModel(ModelPT, ABC):
             registry = AccessMixin.get_module_registry(self)
             log_dict = {}
 
-            for loss_key, loss_list in registry.items():
-                if 'adapter' in loss_key:
-                    # Add auxiliary loss to total loss
-                    loss_value = sum(loss_list)
-                    loss += loss_value
+            for loss_key, loss_registry in registry.items():
+                # Add auxiliary loss to total loss
+                loss_list = loss_registry['adapter_loss']
+                loss_value = sum(loss_list)
+                loss += loss_value
 
-                    # Log current loss name and value
-                    keys = loss_key.split(".")
-                    key = "/".join(keys)
-                    log_dict[key] = loss_value.detach()
+                # Log current loss name and value
+                keys = loss_key.split(".")
+                key = "/".join(keys)
+                key = "adapter_loss/" + key
+                log_dict[key] = loss_value.detach()
 
             if len(log_dict) > 0:
                 self.log_dict(log_dict)
--- a/nemo/collections/asr/models/asr_model.py
+++ b/nemo/collections/asr/models/asr_model.py
@@ -19,6 +19,7 @@ import torch
 
 from nemo.core.classes import ModelPT
 from nemo.core.classes.exportable import Exportable
+from nemo.core.classes.mixins import AccessMixin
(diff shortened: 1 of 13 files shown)


nemo/core/classes/mixins/access_mixins.py
```diff
@@ -48,12 +48,16 @@ def register_accessible_tensor(self, tensor):
...
 
         if not hasattr(self, '_registry'):
-            self._registry = []
+            self._registry = {}
 
-        if len(self._registry) > 0:
-            self._registry.clear()
+        # incorrect logic
+        # if len(self._registry) > 0:
+        #     self._registry.clear()
```
sam1373, line 55: The reason this was here before was to avoid a situation where we are instantiating an ssl model with access enabled from a nemo file in order to then fine-tune a model which doesn't use this mixin, in which case we'll end up adding modules to registry but never resetting and run out of memory. Since reset is now present in ctc and rnnt models it won't be an issue for them, but could be for other classes.

nemo/collections/asr/models/asr_model.py
```diff
@@ -62,6 +63,43 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
+                keys = loss_key.split(".")
+                key = "/".join(keys)
+                key = "adapter_loss/" + key
+                log_dict[key] = loss_value.detach()
+
+            if len(log_dict) > 0:
+                self.log_dict(log_dict)
+
+        if reset_registry:
+            AccessMixin.reset_registry(self)
```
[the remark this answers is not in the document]
titu1994: Doable, but risks memory leaks if you aren't aware of the other registries. We should play it safer with global reset - and build the registry back part by part and use it up before the reset. Large chance of memory leaks otherwise.

nemo/core/classes/mixins/access_mixins.py
titu1994: Hmm we can use the internal flag to check if we're being restored, and if so, avoiding activating the registry. Lemme see.

nemo/collections/asr/models/asr_model.py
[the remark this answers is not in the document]
sam1373: Perhaps we can have both global and local reset functions, with global being the one called at start of step

titu1994 (commented):
Addressed comments by 
- Making reset by key possible (defaults to global reset)
- Making reset explicitly disable cache of tensors
- Explicitly reset at the beginning and ending of SSL and ASR models

nemo/collections/asr/models/ssl_models.py
```diff
@@ -460,6 +457,11 @@ def decoder_loss_step(self, spectrograms, spec_masks, encoded, encoded_len, targ
 
     # PTL-specific methods
     def training_step(self, batch, batch_nb):
+        if self.output_from_layer is not None and len(self.output_from_layer) > 0:
```
sam1373, line 460: output_from_layer is a dict of loss name -> None or string, so we should check here that at least one of the value is not None; also need to check that self.decoder_losses is not None, as otherwise we only have one loss so output_from_layer is not defined

nemo/core/classes/mixins/access_mixins.py
```diff
@@ -68,15 +73,37 @@ def get_module_registry(cls, module: torch.nn.Module):
...
+                    if registry_key in self._registry:
+                        self._registry.pop(registry_key)
+                    else:
+                        raise KeyError(
+                            f"Registry key `{registry_key}` provided, but registry does not have this key.\n"
+                            f"Available keys in registry : {list(self._registry.keys())}"
+                        )
+
+        # Explicitly disable registry cache after reset
+        AccessMixin.set_access_enabled(access_enabled=False)
```
sam1373, line 102: should this be always or only for global reset?

nemo/collections/asr/models/asr_model.py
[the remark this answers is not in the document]
sam1373: I think this reset should not be global

nemo/core/classes/mixins/access_mixins.py
titu1994: Always. Resets are first and last call, any call to if should indicate done with the registry.

titu1994: It doent delete the registry itself, just disables adding more things to it.

nemo/collections/asr/models/asr_model.py
```diff
@@ -63,6 +64,43 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
...
```
titu1994, line 99: Hmm what I will do is set this flag to to False by default, but if its set, treat it as global reset. This function doesn't know what keys it should reset anyway.

nemo/collections/asr/models/ssl_models.py
```diff
@@ -310,8 +307,21 @@ def forward(
...
         """
+
         # reset module registry from AccessMixin
-        self.reset_registry()
+        if (
+            self.training
+            and self.decoder_losses is not None
+            and self.output_from_layer is not None
+            and len(self.output_from_layer) > 0
+        ):
```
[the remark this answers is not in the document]
titu1994: Updated with temporary flag.

sam1373 (approved):
lgtm if tests pass

PR #4487, opened 2022-07-01
Add support for Apex distributed Adam optimizer with GPT-3

timmoon10:
# What does this PR do ?

Adds support for training GPT-3 with the [Apex implementation of the ZeRO optimizer](https://github.com/NVIDIA/apex/blob/master/apex/contrib/optimizers/distributed_fused_adam.py).

**Collection**: NLP

# Changelog 
- Add option for `distributed_fused_adam` optimizer and integrate with GPT-3 training

# Usage

Set optimizer to `distributed_fused_adam` in config file:

https://github.com/NVIDIA/NeMo/blob/23f6a95fb1479f5d2ff7111adbc5720ed3c58e48/examples/nlp/language_modeling/conf/megatron_gpt_config.yaml#L122

~Note that this optimizer is incompatible with the `megatron_amp_O2` option. It internally applies similar optimizations (bf16 params, wgrad kernel fusion), so refactoring may be warranted in the future.~ This optimizer requires the `megatron_amp_O2` option.

**Pre checks**:
  - [x] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Depends on https://github.com/NVIDIA/apex/pull/1414.
* Depends on https://github.com/NVIDIA/apex/pull/1432
* This is mostly orthogonal to https://github.com/NVIDIA/NeMo/pull/4380. It just requires changing [this line](https://github.com/NVIDIA/NeMo/blob/5bb47b7638645b6b177638fff4e3b3a82b3fd76c/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L335) to:
```if self.megatron_amp_o2 or self.with_distributed_adam:```

### CODE REVIEW

File: nemo/collections/nlp/modules/common/megatron/clip_grads.py [hunk 1]
```diff
@@ -24,6 +25,7 @@
     from apex.multi_tensor_apply import multi_tensor_applier
     from apex.transformer import parallel_state
     from apex.transformer.tensor_parallel.layers import param_is_not_tensor_parallel_duplicate
+    from apex.contrib.optimizers.distributed_fused_adam import DistributedFusedAdam
```
titu1994, line 29: If it's a contrib in a very recent version of apex, best to put this under it's own try catch block with its own flag for testing

nemo/core/classes/modelPT.py
```diff
@@ -544,6 +544,11 @@ def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = N
         if lr is not None:
             optimizer_args['lr'] = lr
 
+        # Distributed optimizer needs process group for reductions
+        if optimizer_name == 'distributed_fused_adam':
```
titu1994, line 548: Core should not special-case something like this, nor import unguarded optional import like this. Please revert or move into nlp override function.

nemo/core/optim/optimizers.py
```diff
@@ -45,11 +45,13 @@
 try:
     from apex.optimizers import FusedLAMB
     from apex.optimizers import FusedAdam
+    from apex.contrib.optimizers.distributed_fused_adam import DistributedFusedAdam
```
titu1994, line 48: Should have it's own try catch probably.

titu1994 (changes requested):
Minor comments. Please request a rereview regarding core changes once PR is ready by both @ericharper and me.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -588,6 +622,55 @@ def setup_test_data(self, cfg):
             )
             self._test_dl = self.build_pretraining_data_loader(self._test_ds, consumed_samples)
 
+    def configure_optimizers(self):
+        retval = super().configure_optimizers()
+
+        if self.with_distributed_adam:
```
titu1994, line 628: Is this monkey patch really necessary? Can't we ask apex code to support our use case ? This method seems incredibly fragile.

nemo/core/classes/modelPT.py
```diff
@@ -431,9 +431,12 @@ def setup_multiple_test_data(self, test_data_config: Union[DictConfig, Dict]):
             if self._test_dl is not None and type(self._test_dl) in [list, tuple]:
                 self._test_names = ['test_{}_'.format(idx) for idx in range(len(self._test_dl))]
 
-    def setup_optimization(self, optim_config: Optional[Union[DictConfig, Dict]] = None):
-        """
-        Prepares an optimizer from a string name and its optional config parameters.
+    def setup_optimization(
+            self,
+            optim_config: Optional[Union[DictConfig, Dict]] = None,
+            optim_args: Dict[str, Any] = {},
```
titu1994, line 437: Should be optim_kwargs since it's a dict

titu1994: Also do not use {} in function, set default as None and check for not none before update below v

titu1994 (changes requested):
Minor comments about new kwargs in core. I'm not enthusiastic about the monkey patch but it's localized to Megatron so it's fine for now.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
timmoon10: I agree it's not a nice solution. I think there are three ways to support this feature:
- Monkey-patch this function
- Add Megatron-specific code into the Apex class
- Create a sub-class of the optimizer within NeMo and override this function

Now that I think about it, the third option sounds more reasonable.

titu1994: 3rd option definitely seems like the way to go.

timmoon10:
I've rebased to incorporate the sequence parallelism support from https://github.com/NVIDIA/NeMo/pull/4380. Pinging @ericharper.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -87,6 +88,12 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
         # self.setup_optimizer_param_groups()
 
         self.megatron_amp_o2 = cfg.get('megatron_amp_O2', False)
+        self.with_distributed_adam = cfg.optim.get('name') == 'distributed_fused_adam'
+
+        if self.megatron_amp_o2 and self.with_distributed_adam:
+            raise ValueError(
+                "Distributed optimizers are not compatible with O2. Please set megatron_amp_O2 to False in the model config."
```
yaoyu-33, line 95: @timmoon10 Hi, @erhoo82 mentioned that the when `with_distributed_adam=True` it includes `megatron_amp_o2`. Is that the case?
If it is, it might not be good to say they are not compatible, but O2 is encapsulated in distributed optimizer. Set magatron_amp_O2 to False will make some confusions to users.

timmoon10: That's a good point, increasingly I'm coming to the opinion that `megatron_amp_o2` should be required for the distributed optimizer.

ericharper: I like this a lot. This will means can we continue to recommend `megatron_amp_o2` as we do now.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -96,6 +103,10 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
             # Model wrapper to convert both model and inputs to half precision
             self.model = Float16Module(module=self.model, precision=cfg.precision)
 
+        if self.with_distributed_adam:
```
ericharper, line 106: This check could be removed if we require megatron_amp_o2.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -166,6 +177,21 @@ def setup_optimizer_param_groups(self):
...
         self._optimizer_param_groups = get_params_for_weight_decay_optimization([self.model])
 
+    def setup_optimization(
+            self,
+            optim_config: Optional[Union[DictConfig, Dict]] = None,
+            optim_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        optim_kwargs = {} if optim_kwargs is None else optim_kwargs.copy()
+        if self.with_distributed_adam:
+            optim_kwargs['process_group'] = parallel_state.get_data_parallel_group()
```
ericharper, line 187: This is being called after data parallel groups have been created right?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -239,6 +269,9 @@ def training_step(self, batch, batch_idx):
             if self.cfg.get('pipeline_model_parallel_size', 1) > 1 or self.cfg.get('sequence_parallel', False):
                 # main grads are stored in the MainParamsOptimizer wrapper
                 self._optimizer.allreduce_main_grads()
+        elif self.with_distributed_adam:
+            # gradients are reduced internally in distributed optimizer
```
ericharper, line 273: This is great!

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -293,7 +326,7 @@ def allreduce_sequence_parallel_gradients(self):
         grads = []
         for param in self.model.parameters():
             if getattr(param, 'sequence_parallel_enabled', False):
-                if self.megatron_amp_o2:
+                if self.megatron_amp_o2 or self.with_distributed_adam:
```
ericharper, line 329: Can simplify this if we require megatron_amp_o2

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -315,7 +348,7 @@ def allreduce_first_last_embeddings(self):
         ):
             if self.model.share_token_embeddings:
                 word_embeddings_weight = self.model.word_embeddings_weight()
-                if self.megatron_amp_o2:
+                if self.megatron_amp_o2 or self.with_distributed_adam:
```
ericharper, line 351: Can simplify this if we require megatron_amp_o2?

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -387,6 +390,28 @@ def __init__(
     ) -> None:
         super().__init__(precision, device, scaler=scaler)
 
+    def optimizer_step(
```
ericharper, line 393: Why is this being added here?

timmoon10: This is almost identical to the [parent class impl](https://github.com/Lightning-AI/lightning/blob/6a9d626e189038f09a0cf7157f61e75300ea99cc/src/pytorch_lightning/plugins/precision/native_amp.py#L70). The major difference is that it doesn't manually unscale the gradients. The distributed optimizer plugs into a custom interface in `GradScaler` and I previously ran into problems when calling `GradScaler.unscale_`.

That said, I'll remove it. megatron_amp_o2 only supports bf16, which doesn't do grad scaling.

timmoon10:
I've made the distributed optimizer dependent on `megatron_amp_O2` instead of being mutually exclusive. I'm not convinced it simplifies the implementation so much as it shifts around the messiness, but it is definitely more intuitive for users. It also hints that this is an experimental feature.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
timmoon10: The model parallel groups must have been initialized before constructing the model, and [I believe the data parallel groups are created at the same time](https://github.com/NVIDIA/apex/blob/a4f42586d3b4252dc1c8d89bbc693307139959ec/apex/transformer/parallel_state.py#L185).

titu1994 (approved):
Looks a lot better now. @ericharper for final review before merge

ericharper (dismissed):
LGTM. Thanks! Let's merge when CI passes.

ericharper (approved):
Re-approving.

ericharper (approved):
LGTM. Thanks!

PR #4508, opened 2022-07-06
[Add] Support for Different LRs with Param Groups

stevehuang52:
This PR aims to add support for different learning rates for different modules in a model. For example, when finetuning an ASR model on small datasets, we might want the encoder to have smaller learning rate than the decoder. With this PR, we can do this by adding a `optim_param_groups` section under `model` in the config file:
```yaml
model:
    optim_param_groups:
        "encoder": 
            "lr": 1e-4
            "momentum": 0.8
    optim:
        "name": sgd
        "lr": 1e-3   # default lr for non-specified parameters
```
In this example, all parameters in the encoder will have learning rate `1e-4` while all other parameters (e.g., decoder) will have learning rate `1e-3`. 

For logging learning rates, the current models only log the first group of `param_groups`, which in this case is the default `lr`.  To log all learning rates, we need to modify the `training_step()` function in the model.


nemo/core/classes/modelPT.py
```diff
@@ -606,10 +606,44 @@ def setup_optimizer_param_groups(self):
...
+        if not hasattr(self, "parameters"):
+            self._optimizer_param_groups = None
+            return
+
+        known_groups = []
+        param_groups = []
+        if hasattr(self.cfg, "optim_param_groups"):
+            param_groups_cfg = self.cfg.optim_param_groups
+            for group, lr in param_groups_cfg.items():
+                module = getattr(self, group, None)
```
titu1994, line 627: Check for none and raise error that invalid module was provided c

titu1994: Better to be explicit rather than silently be incorrect.

nemo/core/classes/modelPT.py
```diff
@@ -606,10 +606,44 @@ def setup_optimizer_param_groups(self):
...
+                if module and hasattr(module, "parameters"):
+                    params = module.parameters()
+                    known_groups.append(group)
+                    param_groups.append({"params": params, "lr": lr})
+
+            other_params = []
+            for n, p in self.named_parameters():
+                is_unknown = True
+                for group in known_groups:
+                    if n.startswith(group):
```
titu1994, line 637: This makes an assumption that Nemo models are not nested which is incorrect. We can have Nemo model inside Nemo model and so it's param state dict will change (inner model . Group)

Rewrite using better logic to determine params. 
Also note, nested Nemo models can have more than one group - two encoders from two different models.

titu1994 (changes requested):
Needs review from @ericharper due to core change

nemo/core/classes/modelPT.py
stevehuang52: Done

nemo/core/classes/modelPT.py
```diff
@@ -606,10 +606,44 @@ def setup_optimizer_param_groups(self):
...
+                    lr: 3e-3   
+        """
+        if not hasattr(self, "parameters"):
+            self._optimizer_param_groups = None
+            return
+
+        known_groups = []
+        param_groups = []
+        if hasattr(self.cfg, "optim_param_groups"):
+            param_groups_cfg = self.cfg.optim_param_groups
```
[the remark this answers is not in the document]
stevehuang52: Done

nemo/core/classes/modelPT.py
stevehuang52: If it's a nested model, we can use some config like
```
optim_param_groups:
    "module1.encoder": 1e-3
    "module2.encoder": 1e-4
```
Do you think this can handle the situation?

nemo/core/classes/modelPT.py
```diff
@@ -606,10 +606,48 @@ def setup_optimizer_param_groups(self):
...
+                optim_param_groups:
+                    encoder: 1e-4
+                    decoder: 1e-3
+                optim:
+                    lr: 3e-3   
+        """
+        if not hasattr(self, "parameters"):
+            self._optimizer_param_groups = None
+            return
+
```
titu1994, line 621: Btw, just realized, why are we caching the param groups ? Don't do it, it silently causes issues with adapters. Let's return the value instead and update setup optimization to use returned value. 

@ericharper is that fine ? Do we need to make changes in Megatron side ?

titu1994:
Overall looks fine. Let's see if we can remove this caching of the groups and then merge

nemo/core/classes/modelPT.py
titu1994: Yes, let's change it if possible. Caching causes downstream problems with adapters c

ericharper (changes requested):
Let's design this with more generality.

nemo/core/classes/modelPT.py
```diff
@@ -606,10 +606,48 @@ def setup_optimizer_param_groups(self):
...
             By default, ModelPT will use self.parameters().
             Override this method to add custom param groups.
-    """
-        param_groups = None
-        if hasattr(self, 'parameters'):
-            param_groups = [{'params': self.parameters()}]
+            In the config file, add 'optim_param_groups' to support different LRs 
+            for different components (unspecified params will use the default LR):
+            model:
+                optim_param_groups:
```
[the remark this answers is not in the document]
stevehuang52: Yeah let's do that. I'll make some modifications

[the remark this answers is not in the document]
stevehuang52: Just updated the code, do you think that would work?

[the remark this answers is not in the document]
titu1994: Lgtm now. Eric for final review.

titu1994 (approved):

ericharper (approved):
LGTM. Thanks!

PR #4518, opened 2022-07-07
[TTS] bug fix - sample rate was being ignored in vocoder dataset

paarthneekhara:
Signed-off-by: Paarth Neekhara <paarth.n@gmail.com>

# What does this PR do ?

Fixes a bug where sample rate was being ignored in VocoderDataset when loading just the audio and audio lengths and not the precomputed mel.

**Collection**: tts

# Changelog 

# Usage

```python
dataset = VocoderDataset(manifest_filepath=manifest_path, sample_rate=16000)
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


redoctopus (approved):
LGTM

XuesongYang (dismissed):

XuesongYang:
@paarthneekhara, thanks for your PR! please address this error: 
```
  File "/opt/conda/lib/python3.8/site-packages/nemo/collections/tts/losses/hifigan_losses.py", line 69, in forward

    loss += torch.mean(torch.abs(rl - gl))

RuntimeError: The size of tensor a (1882) must match the size of tensor b (1878) at non-singleton dimension 2
```

paarthneekhara:
@XuesongYang @redoctopus I think I fixed the issue. Adjusted the number of segments to get at the original sampling rate so that we get n_segments at our target sampling rate. Let me know if this looks ok.

nemo/collections/asr/parts/preprocessing/segment.py
```diff
@@ -186,19 +187,27 @@ def segment_from_file(cls, audio_file, target_sr=None, n_segments=0, trim=False,
         try:
             with sf.SoundFile(audio_file, 'r') as f:
                 sample_rate = f.samplerate
-                if 0 < n_segments < len(f):
-                    max_audio_start = len(f) - n_segments
+                if target_sr is not None:
+                    n_segments_at_original_sr = math.ceil(n_segments * sample_rate / target_sr)
```
redoctopus, line 191: As mentioned in-person--please double-check the test case for both upsampling and downsampling

nemo/collections/asr/parts/preprocessing/segment.py
```diff
@@ -186,19 +187,27 @@ def segment_from_file(cls, audio_file, target_sr=None, n_segments=0, trim=False,
...
                 else:
                     samples = f.read(dtype='float32')
             samples = samples.transpose()
         except RuntimeError as e:
             logging.error(f"Loading {audio_file} via SoundFile raised RuntimeError: `{e}`.")
 
         samples = samples.transpose()
-        return cls(samples, sample_rate, target_sr=target_sr, trim=trim, orig_sr=orig_sr)
+        features = cls(samples, sample_rate, target_sr=target_sr, trim=trim, orig_sr=orig_sr)
+        features._samples = features._samples[:n_segments]
```
redoctopus, line 208: Add case for `n_segments` = 0

redoctopus (approved):
LGTM

nemo/collections/asr/parts/preprocessing/segment.py
XuesongYang: Out of curiosity, why do we have to handle the case that `sample_rate` mismatches `target_sr` in the func `sefment_from_file`? In the return `cls()`, resampling would be applied as shown in https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/preprocessing/segment.py#L98.

XuesongYang (changes requested):
Please also fix the conflicts.

nemo/collections/asr/parts/preprocessing/segment.py
redoctopus: IIRC Paarth was running into an issue where the mismatch caused an unexpected number of segments to be returned since `n_segments` calculated at the original sampling rate would not be the same as for the target sample rate. So this needed to be corrected for before passing in the segments for resampling.

redoctopus (approved):
Fixed conflict, approving again

nemo/collections/asr/parts/preprocessing/segment.py
XuesongYang: yes. I remembered that (see my above comment with detailed errors). But one thing that confused me: 
* why `n_segments (num of samples)` was independent to `sample_rate`, and why it was not automatically calculated based on `seconds * sample_rate`? 
* If you look at https://github.com/NVIDIA/NeMo/blob/main/examples/tts/conf/hifigan/hifigan_44100.yaml#L20, it applies for some fixed number as `train_n_segments: 16384`. Do we assume this number applies to any sample rate?

XuesongYang:
It would be great if we have a unit test capturing this bug fix.

nemo/collections/asr/parts/preprocessing/segment.py
redoctopus: Honestly, I'm not completely sure. But if I had to guess, maybe something to do with GPU memory during training? Similar to how we have a max audio length cutoff, perhaps...

paarthneekhara: Yes, I think for HiFiGAN the train_n_segments should be 16384 for any sampling rate (Probably because it upsamples the spectrogram by some power of 2 at each layer). Also, I think the expected behavior of the function should be to get n_segments at the specified sampling rate.

XuesongYang: Thank you guys for the answers. It helps me understand well. Approved.

XuesongYang (approved):
LGTM. Thanks for the fix.

PR #4520, opened 2022-07-07
[Fix] Relative Audio Path

stevehuang52:
Fix the too long audio filenames for tarred datasets. If the audio filename is too long, just keep it as it is.


titu1994 (approved):
Thanks !

PR #4532, opened 2022-07-08
Integrating support for GPT/T5/BART for Question Answering

ameyasm1154:
Signed-off-by: Ameya Mahabaleshwarkar <ameyasm1154@gmail.com>

# What does this PR do ?

Integrating support for GPT/T5/BART for Question Answering

**Collection**: nlp

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+            return re.sub(r"\b(a|an|the)\b", " ", text)
+
+        def white_space_fix(text):
+            return " ".join(text.split())
+
+        def remove_punc(text):
+            exclude = set(string.punctuation)
+            return "".join(ch for ch in text if ch not in exclude)
+
+        def lower(text):
```
Zhilin123, line 36: this doesnt need a new method if you're just called .lower()

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+            return text.lower()
+
+        return white_space_fix(remove_articles(remove_punc(lower(s))))
+
+    @classmethod
+    def _get_tokens(cls, s: str):
```
Zhilin123, line 42: how about _get_normalized_tokens?

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+    def get_f1(cls, prediction: str, ground_truth: str):
+        """ Computes f1 score between prediction and ground truth """
+
+        prediction_tokens = cls._get_tokens(prediction)
+        ground_truth_tokens = cls._get_tokens(ground_truth)
+        common = Counter(prediction_tokens) & Counter(ground_truth_tokens)
+        num_same = sum(common.values())
+
+        if len(ground_truth_tokens) == 0 or len(prediction_tokens) == 0:
+            # If either is no-answer, then F1 is 1 if they agree, 0 otherwise
```
Zhilin123, line 58: move comment to line above the if since that's where the logic starts

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+import string
+from collections import Counter
+
+
+class QAMetrics(object):
+    @classmethod
+    def normalize_answer(cls, s: str):
+        """ Lower text and remove punctuation, articles and extra whitespace """
+
+        def remove_articles(text):
```
Zhilin123, line 26: unless there are good reasons (i.e. needing to access local variables), I wouldn't recommend putting methods in a method. This is because they make them harder to test and also if you want to re-use the method somewhere else later on, this is pretty much impossible

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+import string
+from collections import Counter
+
+
+class QAMetrics(object):
+    @classmethod
```
Zhilin123, line 22: I noticed you are using classmethods here - you should have a look at staticmethod vs classmethod and I generally recommend staticmethod if you're not accessing class variables.

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @classmethod
+    def _get_tokens(cls, s: str):
+        """ Get normalized tokens """
+        if not s:
+            return []
+        return cls.normalize_answer(s).split()
+
+    @classmethod
+    def get_f1(cls, prediction: str, ground_truth: str):
```
Zhilin123, line 49: you should make it clear this is for one sample

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+                logging.warning(f"Missing prediction for {qas_id}")
+                continue
+
+            prediction = preds[qas_id]
+            exact_scores[qas_id] = max(QAMetrics.get_exact_match(prediction, a) for a in gold_answers)
+            f1_scores[qas_id] = max(QAMetrics.get_f1(prediction, a) for a in gold_answers)
+
+        return exact_scores, f1_scores
+
+    def _apply_no_ans_threshold(self, scores, na_probs, qid_to_has_ans, na_prob_thresh):
```
Zhilin123, line 705: Can you make the description more descriptive? Currently the desctiption contains no more information than the method name. For instance, you can indicate what each arg is and what the expected output is

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+
+        for example in examples:
+            qas_id = example.qas_id
+            gold_answers = [answer["text"] for answer in example.answers if QAMetrics.normalize_answer(answer["text"])]
+
+            if not gold_answers:
+                # For unanswerable questions, only correct answer is empty string
+                gold_answers = [""]
+
+            if qas_id not in preds:
```
Zhilin123, line 695: I wonder when this will be triggered because this looks like a bug if we ever get here

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+        if orig_end_position is None:
+            if verbose_logging:
+                logging.warning("Couldn't map end position")
+            return orig_text
+
+        output_text = orig_text[orig_start_position : (orig_end_position + 1)]
+
+        return output_text
+
+    def _get_raw_scores(self, examples: List, preds: Dict[str, str]):
```
Zhilin123, line 680: change raw score in method name to exact_match and f_1

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+        logits = self.forward(input_ids=input_ids, token_type_ids=input_type_ids, attention_mask=input_mask)
+        loss, _, _ = self.loss(logits=logits, start_positions=start_positions, end_positions=end_positions)
+        lr = self._optimizer.param_groups[0]['lr']
+        self.log('train_loss', loss)
+        self.log('lr', lr, prog_bar=True)
+
+        return {'loss': loss, 'lr': lr}
+
+    def validation_step(self, batch, batch_idx):
+        if self.trainer.testing:
```
Zhilin123, line 103: nit: in line if else can work here `prefix = 'test' if self.trainer.testing else 'val'`

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+            'start_logits': start_logits,
+            'end_logits': end_logits,
+        }
+        return {f'{prefix}_loss': loss, f'{prefix}_tensors': tensors}
+
+    def test_step(self, batch, batch_idx):
+        return self.validation_step(batch, batch_idx)
+
+    def validation_epoch_end(self, outputs):
+        if self.trainer.testing:
```
Zhilin123, line 125: same as above

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+            
+        Returns:
+            model predictions, model nbest list
+        """
+
+        # store predictions for all queries in a single list
+        all_predictions = []
+        all_nbest = []
+        mode = self.training
+        device = 'cuda' if torch.cuda.is_available() else 'cpu'
```
Zhilin123, line 232: wouldnt recommend this - instead ask the trainer.devices since 'cuda' by default will use the first device (i.e. cuda:0) but sometimes, you can specify devices=[1] which is cuda:1

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+        """
+        Find best threshhold to maximize evaluation metric
+        """
+
+        num_no_ans = sum(1 for k in qid_to_has_ans if not qid_to_has_ans[k])
+        cur_score = num_no_ans
+        best_score = cur_score
+        best_thresh = 0.0
+        qid_list = sorted(na_probs, key=lambda k: na_probs[k])
+        for _, qid in enumerate(qid_list):
```
Zhilin123, line 759: why use enumerate if you don't want the index, for `qid in qid_list` would work

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+                        end_logit=null_end_logit,
+                    )
+                )
+            prelim_predictions = sorted(prelim_predictions, key=lambda x: (x.start_logit + x.end_logit), reverse=True)
+
+            _NbestPrediction = collections.namedtuple("NbestPrediction", ["text", "start_logit", "end_logit"])
+
+            seen_predictions = {}
+            nbest = []
+            for pred in prelim_predictions:
```
Zhilin123, line 453: A lot of code here is quite unclear - I understand this was not something you wrote but if you have extra time, maybe breaking these up into methods will be helpful.

Zhilin123: especially lines 462 to 471 should be easily refactors into a method

Zhilin123: Another advantage to refactoring code that you didnt write yourself is that you get to understand code from the perspective of the reader.

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+            exact_scores[qas_id] = max(QAMetrics.get_exact_match(prediction, a) for a in gold_answers)
+            f1_scores[qas_id] = max(QAMetrics.get_f1(prediction, a) for a in gold_answers)
+
+        return exact_scores, f1_scores
+
+    def _apply_no_ans_threshold(self, scores, na_probs, qid_to_has_ans, na_prob_thresh):
+        """ Applies no answer threshold """
+
+        new_scores = {}
+        for qid, s in scores.items():
```
Zhilin123, line 709: Use more informative names for variables like qid and s (maybe question and score respectively)

nemo/collections/nlp/models/question_answering/qa_bert_model.py
```diff
@@ -0,0 +1,892 @@
...
+        """
+        best_exact, exact_thresh = self._find_best_thresh(preds, exact_raw, na_probs, qid_to_has_ans)
+        best_f1, f1_thresh = self._find_best_thresh(preds, f1_raw, na_probs, qid_to_has_ans)
+
+        main_eval["best_exact"] = best_exact
+        main_eval["best_exact_thresh"] = exact_thresh
+        main_eval["best_f1"] = best_f1
+        main_eval["best_f1_thresh"] = f1_thresh
+
+    def _evaluate_predictions(
```
Zhilin123, line 788: Would like to have some description here

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+        if not train_data_config or not train_data_config.file:
+            logging.info(
+                f"Dataloader config or file_path for the train is missing, so no data loader for test is created!"
+            )
+            self._test_dl = None
+            return
+
+        self._train_dl = self._setup_dataloader_from_config(cfg=train_data_config, mode=TRAINING_MODE)
+
+    def setup_validation_data(self, val_data_config: Optional[DictConfig]):
```
Zhilin123, line 67: Many of these methods seems to be sharable with BERT/S2S?

nemo/collections/nlp/models/question_answering/qa_model.py
```diff
@@ -57,14 +57,15 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
 
     @typecheck()
     def forward(self, input_ids, attention_mask, token_type_ids):
-        hidden_states = self.bert_model(
-            input_ids=input_ids, token_type_ids=token_type_ids, attention_mask=attention_mask
-        )
+        with autocast():
```
Zhilin123, line 60: Are you planning to keep the qa_model? It seems similar to the BERT model

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+                {
+                    "unique_ids": unique_ids,
+                    "per_sample_perplexity": per_sample_perplexity,
+                    "generated_answers": generated_texts,
+                }
+            )
+
+        return outputs
+
+    def _prep_inference_labels(self, input_ids, input_attn_mask, training_mask_end, device):
```
Zhilin123, line 328: Some descriptions please!

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+        labels = GPTQADataset.get_labels_from_inputs(
+            input_ids, training_mask_end, input_attn_mask, self.tokenizer.tokenizer,
+        )
+        if len(labels.shape) == 1:
+            labels = torch.unsqueeze(labels, 0)
+        labels = labels.to(device)
+
+        return input_ids, input_attn_mask, labels, generated_texts
+
+    def _generate_candidates(self, input_ids, input_attn_mask, training_mask_end):
```
Zhilin123, line 358: Some descriptions here too!

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+                    break
+
+            curr_answer = self.tokenizer.tokenizer.decode(
+                token_ids[i, start_point:stop_point], skip_special_tokens=True
+            ).strip()
+            answers.append(curr_answer)
+
+        return answers
+
+    def _get_raw_scores(self, examples, preds):
```
Zhilin123, line 409: This seems exactly like in BERT, maybe we can have a base QA model that does all these? Alternatively this method can be delegate to the QA metrics class

Zhilin123: For instance, the metrics class can have a method called `_get_raw_scores(self, examples, preds):` that various models call

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+                logging.warning(f"Missing prediction for {qas_id}")
+                continue
+
+            pred = preds[qas_id].generated_text
+            exact_scores[qas_id] = max(QAMetrics.get_exact_match(pred, a) for a in gold_answers)
+            f1_scores[qas_id] = max(QAMetrics.get_f1(pred, a) for a in gold_answers)
+
+        return exact_scores, f1_scores
+
+    def _evaluate_predictions(
```
Zhilin123, line 431: Think it should go into the metrics class too

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+
+        return collections.OrderedDict(
+            [
+                ("exact", 100.0 * sum(exact.values()) / total),
+                ("f1", 100.0 * sum(f1.values()) / total),
+                ("total", total),
+            ]
+        )
+
+    def _dump_predictions_to_file(self, output_file, examples, predictions):
```
Zhilin123, line 445: This should go into the metrics class (or a util class if you want one) The reason for this is that dumping predictions has no relation to the model itself

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+                    "question": ex.question_text,
+                    "predicted_answer": predictions[ex.qas_id].generated_text,
+                    "perplexity": predictions[ex.qas_id].perplexity,
+                }
+            )
+
+        with open(output_file, "w") as writer:
+            writer.write(json.dumps(outputs))
+
+    def _transfer_tensors_to_device(self, tensors: Tuple, device):
```
Zhilin123, line 461: no need for a method here - in line will be clearer i.e. for tensor in tensors, tensor.to(device)

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
```diff
@@ -0,0 +1,485 @@
...
+        tensors = (tensor.to(device) for tensor in tensors)
+        return tensors
+
+    def _convert_dict_outputs_to_lists(self, outputs, keys):
```
Zhilin123, line 465: doesn't look like it belongs here - maybe metrics class? Also this need not be a normal method, staticmethod should do

nemo/collections/nlp/models/question_answering/qa_s2s_model.py
```diff
@@ -0,0 +1,447 @@
...
+
+        if self.cfg.library == "huggingface":
+            self.language_model = AutoModelForSeq2SeqLM.from_pretrained(cfg.language_model.pretrained_model_name)
+            self.language_model.resize_token_embeddings(len(self.tokenizer.tokenizer))
+            if self.cfg.language_model.lm_checkpoint:
+                self.language_model.load_state_dict(torch.load(self.cfg.language_model.lm_checkpoint))
+        elif self.cfg.library == "megatron":
+            self.language_model = language_model
+
+    def setup_training_data(self, train_data_config: Optional[DictConfig]):
```
Zhilin123, line 74: Many comments from GPT works here too

nemo/collections/nlp/models/question_answering/qa_s2s_model.py
```diff
@@ -0,0 +1,447 @@
...
+        self.log(f'{prefix}_f1', f1)
+
+    def test_epoch_end(self, outputs):
+        self.validation_epoch_end(outputs)
+
+    @typecheck()
+    def forward(self, input_ids, input_attn_mask, labels):
+        loss, per_sample_perplexity = None, None
+        if self.cfg.library == "huggingface":
+            with autocast(enabled=False):
```
Zhilin123, line 182: not sure about details of autocast implementation, but it looks to me like you're using the autocast context with enabled=False? If that's the case, maybe we should just remove it?

nemo/collections/nlp/models/question_answering/qa_s2s_model.py
```diff
@@ -0,0 +1,447 @@
...
+            drop_last=cfg.drop_last,
+            shuffle=cfg.shuffle,
+            num_workers=cfg.num_workers,
+            pin_memory=cfg.pin_memory,
+        )
+
+        return data_loader
+
+    @torch.no_grad()
+    def _get_per_sample_perplexity(self, logits, labels):
```
Zhilin123, line 284: Seems to be shared with GPT

Zhilin123 (changes requested):
Overall, it's a very good first PR - I'm impressed by the amount of consideration, design and implementation details you put into this so I'm very happy about this as a start. Some particular things that I liked - code organization (into clear folders and classes); methods are generally concise and readable. Some general things that can be improved: Descriptions for methods can be substantiated; some code is duplicated across various modules and the particular methods seem to be owned by classes that are not the most suitable.

nemo/collections/nlp/models/question_answering/qa_s2s_model.py
ameyasm1154: Removing it causes problems when running the model in mixed precision / fp16. I am yet to dig into the details as what this does, but this was pointed out as a bug in HF for T5/BART models along with the fix as above. Removing it currently won't allow operating the model in fp16

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
ameyasm1154: What is dumped in the predictions file apart from just the context, question, text would depend on the model though. Ex. start/end pos for BERT extractive QA, ppl for Generative QA. So, formatting the output would still have to be in the model class, so would it make sense to have a separate method in QAMetrics just for dumping a json file?

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
ameyasm1154: > This seems exactly like in BERT, maybe we can have a base QA model that does all these? Alternatively this method can be delegate to the QA metrics class

Would we need to extend metrics for generative QA when we get to free form answering like Context + Question -> Fluent response?

nemo/collections/nlp/models/question_answering/qa_bert_model.py
Zhilin123: From what I'm seeing in the inference method and the examples script for QA, you can try to call self.trainer.device? The model was initialized with the trainer and hence can call the trainer. If this doesn't work, we can also leave it as it is.

nemo/collections/nlp/models/question_answering/qa_s2s_model.py
Zhilin123: I see - if it's a bug like that, you might want to write a comment and link the bug if possible. As far as I am aware, I think there are some layers in T5 (only for T5-large and above) that are in fp32 with the rest in bf16 so they might not be cast-able to fp16 easily. BART was ok for me.

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
Zhilin123: I think the metrics in this case will remain as F1 and Exact Match? I've also used BLEU before but iirc, they highly correlate with F1 so I think it will be fine to keep using F1 and Exact Match (also easier to understand). BLEU is most helpful when there are multiple acceptable answers, each of which is slightly different from one another (i.e. in Machine Translation).

nemo/collections/nlp/models/question_answering/qa_gpt_model.py
Zhilin123: I see - we can keep formatting the output in the model class in this case. Having a dedicated method to dump a json file also seems a little weird since it's only 3 lines. If this is the case, maybe change the name to _format_and_dump_predictions_to_file and also change `output_file` to `output_filename`

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+    @staticmethod
+    def normalize_answer(s: str):
+        """ Lower text and remove punctuation, articles and extra whitespace """
+
+        return QAMetrics.white_space_fix(QAMetrics.remove_articles(QAMetrics.remove_punc(s.lower())))
+
+    @staticmethod
+    def _get_normalized_tokens(s: str):
+        """ Get normalized tokens """
+        if not s:
```
Zhilin123, line 47: nit: I think your code might handle the base case already? `''.split()` returns []

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        """ 
+        Calculates exact match and f1 scores for all predictions, 
+            questions with answers, and no answer questions
+        """
+
+        qas_id_to_has_answer = {example.qas_id: bool(example.answers) for example in examples[: len(all_predictions)]}
+        has_answer_qids = [qas_id for qas_id, has_answer in qas_id_to_has_answer.items() if has_answer]
+        no_answer_qids = [qas_id for qas_id, has_answer in qas_id_to_has_answer.items() if not has_answer]
+
+        all_exact, all_f1 = QAMetrics.get_exact_match_and_f1(examples, all_predictions, list(qas_id_to_has_answer))
```
Zhilin123, line 154: nit: Looks like you're repeating yourself a few times. if it's fast to write, can you maybe write a for loop instead?

nemo/collections/nlp/metrics/qa_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        
+        no_ans_exact, no_ans_f1 = QAMetrics.get_exact_match_and_f1(examples, all_predictions, no_answer_qids)
+        no_ans_eval_dict = QAMetrics.make_eval_dict(no_ans_exact, no_ans_f1, prefix="NoAns_")
+
+        merged_eval_dict = QAMetrics.merge_eval_dicts([all_eval_dict, has_ans_eval_dict, no_ans_eval_dict])
+
+        return merged_eval_dict
+
+    @staticmethod
+    def dump_predicted_answers_to_file(output_filename, examples, predictions):
```
Zhilin123, line 168: Forgot about this last time, but generally jsonline (.jsonl files) are preferred as they are easier to process line by line  (i.e. more straightforward to load file lazily instead of having to use a dedicated solution like ijson). jsonl can be easier to visualize as well (using jsonify on a huge document can lead to IDEs lagging).

Zhilin123 (changes requested):
Good improvements! Looking forward to the rest!

nemo/collections/nlp/metrics/qa_metrics.py
ameyasm1154: SInce the str passes through the normalize answer method before performing split, checking `not s` condition avoids error in normalize answer method

examples/nlp/question_answering/conf/qa_conf.yaml
```diff
@@ -0,0 +1,157 @@
...
+    drop_last: ${model.dataset.drop_last}
+    pin_memory: ${model.dataset.pin_memory}
+
+  language_model:
+    pretrained_model_name: bert-base-uncased # main config to select model (between bert, gpt2, t5/bart based models)
+    lm_checkpoint: null
+    config_file: null # json file, precedence over config
+    config: null
+
+  token_classifier: # used only by BERTQAModel for defining the extractive QA head
```
Zhilin123, line 109: Be careful to make sure that the default args are kept generally consistent i.e. there might be some people try to call the bert QA model based before this change was made so we would want to make it easy for them to do so

nemo/collections/nlp/data/question_answering/dataset/qa_s2s_dataset.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            self._convert_examples_to_features()
+            if use_cache:
+                QADataset.dump_features_to_cache(self.cached_features_file, self.features)
+
+        logging.info("Converting dict features into object features")
+        for i in trange(len(self.features)):
+            self.features[i] = S2SQAInputExample(**self.features[i])
+
+    def _set_cached_features_filename(self):
```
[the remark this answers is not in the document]
ameyasm1154: The method requires a lot of class `self.` parameters which are different for different model. We tried this earlier but required passing a lot of parameters to the method in base class

Jenkinsfile
```diff
@@ -1247,6 +1247,187 @@ pipeline {
       }
     }
 
+    stage('L2: Parallel BERT Question-Answering SQUAD v1.1 / v2.0') {
```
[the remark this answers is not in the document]
ameyasm1154: Attempted to put stages in parallel. I'll reduce the number of parallel steps if it does not work

docs/source/nlp/question_answering.rst
```diff
@@ -163,139 +136,76 @@ evaluation:
...
-Fine-tuning procedure and logs look similar to what's described in the Model Training section, with the addition of the model
-that is initially loaded from a previously trained checkpoint, e.g. by specifying :code:`pretrained_model=<PRETRAINED_MODEL_NAME>`.
-
-Inference
----------
-
-An example script on how to run inference can be found at `examples/nlp/question_answering/question_answering_squad.py <https://github.com/NVIDIA/NeMo/blob/stable/examples/nlp/question_answering/question_answering_squad.py>`_.
-
-To run inference with the pre-trained model, run:
+For downloading the MS-MARCO dataset, Terms of Use need to be accepted at https://microsoft.github.io/msmarco/. The dataset files can be downloaded from https://msmarco.blob.core.windows.net/msmarco/train_v2.1.json.gz and https://msmarco.blob.core.windows.net/msmarco/dev_v2.1.json.gz.
```
Zhilin123, line 141: Can you add something on if you have other datasets (give some examples based those in the MultiQA 7 datasets) , you should convert into squad format and in a nested json structure how that squad format looks like?

Zhilin123: This is actually already in the previous document - can you copy that over?

docs/source/nlp/question_answering_squad.rst
```diff
@@ -0,0 +1,301 @@
...
+
+
+Available models
+^^^^^^^^^^^^^^^^
+
+.. list-table:: *Pretrained Models*
+   :widths: 5 10
+   :header-rows: 1
+
+   * - Model
```
Zhilin123, line 58: Keep something like this in your new documentation

docs/source/nlp/question_answering_squad.rst
```diff
@@ -0,0 +1,301 @@
...
+   * - qa_squadv2.0_megatron_uncased
+     - https://ngc.nvidia.com/catalog/models/nvidia:nemo:qa_squadv2_0_megatron_uncased
+
+
+.. _dataset_question_answering:
+
+Data Format
+-----------
+
+This model expects the dataset in `SQuAD format`_ (i.e., a JSON file for each dataset split). The code snippet below shows an example
```
Zhilin123, line 83: Keep something like this in your new documentation

docs/source/nlp/question_answering_squad.rst
```diff
@@ -0,0 +1,301 @@
...
+To perform training of the Question Answering model on the SQuAD dataset, you must first download it from `here
+<https://rajpurkar.github.io/SQuAD-explorer/>`_ or run:
+
+.. code::
+
+    python get_squad.py 
+
+There are two versions: 
+
+- SQuAD version 1.1, which does not contain questions without the answer and has 100,000+ question-answer pairs on 500+ articles.
```
Zhilin123, line 149: Keep something like this in your new documentation - also do the same for MSMarco if it's straightforward

Zhilin123 (changes requested):
Looking great - only minor changes needed

Zhilin123 (dismissed):
LGTM Can you merge main into this? Think we are ready

Zhilin123 (dismissed):
LGTM

Zhilin123 (approved):
LGTM

Issue #4300, opened 2022-06-01
Problems running multi-gpu punctuation capitalization training 


itzsimpl:
**Describe the bug**

Running multi-gpu training without a pre-prepared cache, crashes upon initialization with the following trace
```python3
Traceback (most recent call last):
  File "examples/nlp/token_classification/punctuation_capitalization_train_evaluate.py", line 155, in <module>
    main()
  File "/workspace/nemo/nemo/core/config/hydra_runner.py", line 104, in wrapper
    _run_hydra(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 377, in _run_hydra
    run_and_report(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 214, in run_and_report
    raise ex
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 211, in run_and_report
    return func()
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/utils.py", line 378, in <lambda>
    lambda: hydra.run(
  File "/opt/conda/lib/python3.8/site-packages/hydra/_internal/hydra.py", line 111, in run
    _ = ret.return_value
  File "/opt/conda/lib/python3.8/site-packages/hydra/core/utils.py", line 233, in return_value
    raise self._return_value
  File "/opt/conda/lib/python3.8/site-packages/hydra/core/utils.py", line 160, in run_job
    ret.return_value = task_function(task_cfg)
  File "examples/nlp/token_classification/punctuation_capitalization_train_evaluate.py", line 116, in main
    model = PunctuationCapitalizationModel(cfg.model, trainer=trainer)
  File "/workspace/nemo/nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py", line 101, in __init__
    super().__init__(cfg=cfg, trainer=trainer)
  File "/workspace/nemo/nemo/collections/nlp/models/nlp_model.py", line 98, in __init__
    super().__init__(cfg, trainer)
  File "/workspace/nemo/nemo/core/classes/modelPT.py", line 138, in __init__
    self.setup_training_data(self._cfg.train_ds)
  File "/workspace/nemo/nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py", line 469, in setup_training_data
    self._train_dl = self._setup_dataloader_from_config(cfg=train_data_config, train=True)
[...]

PR #4544, opened 2022-07-13
wait for global_rank_zero() to move old files

itzsimpl:

# What does this PR do ?

In multi-gpu settings there is no guarantee that `global_rank_zero()` will be the first process to start. Assuming a shared storage this means that when resuming training with `log_global_rank_0_only=False`, this may lead to previous run `nemo_log_globalrank-*_localrank-*.log` files being overwritten and current run files being moved to the `run_*` subfolder.  

**Collection**: [nemo/utils]

# Changelog 
Ensure all files from previous run are moved before continuing execution of non `global_rank_zero()` processes.

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/utils/exp_manager.py
```diff
@@ -470,10 +470,18 @@ def check_resume(
...
             new_run_dir.mkdir()
             for _file in files_to_move:
                 move(str(_file), str(new_run_dir))
+            time.sleep(10)
+    else:
+        files_to_move = [child for child in Path(log_dir).iterdir() if child.is_file()]
+        while len(files_to_move) > 0:
+            logging.info(f"Waiting for global_rank_zero() to finish back up of previous run")
+            files_to_move = [child for child in Path(log_dir).iterdir() if child.is_file()]
+            time.sleep(5)
```
titu1994, line 485: Instead of busy waiting, check if in distributed torch mode, then use torch.distributed.barrier()

itzsimpl: That will unfortunately not work, at that point in code `torch.distributed` is not yet initialised (I have just checked), hence `torch.distributed.barrier()` cannot be used. 

This should make no difference, but just for clarity; I am running using Slurm 21.08.5, which was deployed via [nvidia/deepos](https://github.com/NVIDIA/deepops), and support for containers is provided by [nvidia/enroot](https://github.com/NVIDIA/enroot) and [nvidia/pyxis](https://github.com/NVIDIA/pyxis). Logs contain a line `Multiprocessing is handled by SLURM.` At least in this setting it seems that `torch.distributed` gets initialised much later, indeed I mentioned in #4300 that it is not initialised even when NLP/PC DataLoaders get initialised. This may be true for other nemo/collections that make use of `torch.distributed.barrier()` as well. 

At the moment I don’t have a better solution than busy waiting. In fact I’ve added additional sleep cycles to ensure all loops exit before any files are created.

itzsimpl: @titu1994 any comments? Is there anything else that could be used as a barrier/semaphore?

itzsimpl: @titu1994 reviewing PTL code, one can see, that `Trainer.strategy.setup_environment()` is called only after calling `Trainer.fit()`, `Trainer.test()` or `Trainer.predict()`, see https://github.com/Lightning-AI/lightning/blob/8d14554383632f4cdff337b3d8d1b226cabdd1d0/src/pytorch_lightning/trainer/trainer.py#L1114-L1136

`Trainer.strategy.setup_environment()` will call `Strategy.setup_distributed()`, which will call `Strategy.init_dist_connection()`, which finaly does `torch.distributed.init_process_group()`, therefore initialises `torch.distribured` and makes it available, and also makes `barrier()` a possibility. 

Based on https://github.com/Lightning-AI/lightning/blob/8d14554383632f4cdff337b3d8d1b226cabdd1d0/src/pytorch_lightning/strategies/ddp.py#L307-L313 I would also assume that it is better to call `trainer.strategy.barrier()` than `torch.distributed.barier()` directly.

In other words, the only way of making `torch.distributed` available earlier is to add the following two lines immediately after trainer setup in all `nemo/examples`
```python3
    if trainer.strategy.is_distributed:
        trainer.strategy.setup_environment()
```
with all synchronisations can be handled via `trainer.strategy.barrier()`.

If you prefer this solution, let me know and I will update the PR, but it will touch a lot of nemo code, i.e. all examples and all calls to `torch.distributed.barrier()`. I'm no expert in PTL, so if you have someone on the inside, do double check.

titu1994:
Modifying dozens of files is not a solution. I fail to see what is the issue here, our team also uses slurm as well, with shared data storage as well and we do not encounter this issue. Another team uses mpirun based DDP and also do not face this issue. 

Importantly, our team definitely sets that flag to false for certain runs, and does not face any issue with race conditions.

itzsimpl:
I will try to explain the best I can. This PR here is just for making sure that when one starts a resume of a job (resumes training) `nemo_log_globalrank-*_localrank-*.log` files of previous run, are safely moved to a subfolder named `run_xxx` prior to continuing. If not done so, some `nemo_log_globalrank-*_localrank-*.log` may end up missing, because for example `rank_1` (which started first), already created the log file for the new run, but `rank_0` (which started next), moved the file to `run_xxx`. This is all non-critical, but a nice to have, when one needs to debug.

What is more critical to me, if I may say so, is that if you search through nemo for `torch.distributed.barrier()` some hits will show that these are placed in initialisation code, and often the initialisation happens prior to `trainer.fit`, `trainer.predict`, or `trainer.test`. One such example are NLP DataLoaders. Since at the time DataLoaders are being initialised `torch.distributed` has not yet been initialised any call to `torch.distributed.barrier()` will serve no function, i.e. not execute. Assuming barriers have been placed there with a reason, this may lead to errors that are really hard to find or debug.

titu1994:
@MaximumEntropy @ekmb I'd suggest looking into such calls and structuring them properly. As to this PR, it fixes debugging but causes more issues to core  so I'd like to skip it v

itzsimpl:
Just as a thought. To avoid modifying dozens of files, one could simply do `trainer.strategy.setup_environment()` inside `exp_manager()`, if that would be ok. If not, I'll close the PR.

titu1994:
It would be ok, probably. We don't know what is the impact on slurm or horovod clusters etc so if would need to be rigorously tested to make sure nothing is screwed up. So even though it's possible it would be a difficult change

itzsimpl:
The impact could be further reduced by doing initialisation only in case of a detected resume training, but OK, I'll close the PR, as it does not resolve anything critical. 

FYI. Searching through nemo for `trainer.strategy.setup_environment()` returns five hits so it does seem that at least some collections already resort to pre-intialization of `torch.distributed`; see `megatron_gpt_model.py`, `megatron_gpt_prompt_learning_model.py`, `megatron_lm_encoder_decoder_model.py`, `test_retrieval_module_inference.py` and `test_retrieval_module.py`

PR #4553, opened 2022-07-15
fix tarred dataset len when num shards is not divisible by workers

### LINKED ISSUE #4522: [NLP] punctuation_capitalization resume training fails
**Describe the bug**

When training PC from scratch, using the example configs, everything works fine, however if I try to resume an interrupted training from the last checkpoint, I get the following log entry
```
[NeMo W 2022-07-07 21:18:38 nemo_logging:349] /opt/conda/lib/python3.8/site-packages/pytorch_lightning/callbacks/model_checkpoint.py:378: UserWarning: `ModelCheckpoint(monitor='val_loss')` could not find the monitored key in the returned metrics: ['epoch', 'step']. HINT: Did you call `log('val_loss', value)` in the `LightningModule`?
      warning_cache.warn(m)
```
and all new checkpoints get a `val_loss=0.0000`

**Environment details**

- pytorch:22.04-py3
- nemo:1.8.2 installed with ./reinstall.sh, without numba install

itzsimpl:

# What does this PR do ?

Fixes #4522. When using tarred datasets and the number of shards is not divisible by `world_size`, nemo logs a warning, but training continues. However, in this case the computation of `DataLoader.len()` becomes incorrect. This leads to invalid `lr_scheduler.max_steps`, and other issues down the road. One example being `Validation check` not running at end of epoch. 

**Collection**: [nlp]
- language_modelling
- machine_translation
- text_normalization
- token_classification

# Changelog 
In case of `shard_strategy='scatter'` the `len()` is computed based on the amount of taken shards and assuming total `num_baches` are equally distributed over all shards, i.e. that each shard contains the same number of batches. In addition, following other implementations this PR adds `shard_strategy` support to NLP/PC tarred datasets.

There are two additional occasions, where tarred datasets are used  
https://github.com/NVIDIA/NeMo/blob/fa2e55e607f787d129141523c724b29dab7a411c/nemo/collections/asr/data/audio_to_text.py#L186
which based on the implementation does not seem to be affected by this issue, and
https://github.com/NVIDIA/NeMo/blob/fa2e55e607f787d129141523c724b29dab7a411c/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py#L169-L184
which might, but I chose to leave it as is, as the computation of `len()` 
https://github.com/NVIDIA/NeMo/blob/fa2e55e607f787d129141523c724b29dab7a411c/nemo/collections/nlp/data/language_modeling/l2r_lm_dataset.py#L239-L240
greatly differs from others
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [x] Documentation

## Who can review?

note to @PeganovAnton for NLP/C

# Additional Information
* Related to #4522


MaximumEntropy (commented):
Nice find. Thanks for fixing this! Looks good to me for NMT and Language Modeling, will let @ekmb and @PeganovAnton review for Punctuation and Capitalization.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
```diff
@@ -897,8 +897,14 @@ def __init__(
         world_size: int = 1,
         global_rank: int = 0,
         shuffle_n: int = 1,
+        shard_strategy: str = "scatter",
```
PeganovAnton, line 900: Could you please add description of `shard_strategy` to docstring?

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
```diff
@@ -897,8 +897,14 @@ def __init__(
...
     ) -> None:
         super().__init__()
+
+        valid_shard_strategies = ['scatter', 'replicate']
+        if shard_strategy not in valid_shard_strategies:
+            raise ValueError(f"`shard_strategy` must be one of {valid_shard_strategies}")
```
PeganovAnton, line 906:
```suggestion
            raise ValueError(
                f"Invalid shard strategy of type {type(shard_strategy)} "
                f"{repr(shard_strategy) if len(repr(shard_strategy)) < 100 else repr(shard_strategy)[:100] + '...'}! "
                f"Allowed values are: {valid_shard_strategies}."
            )
```

PeganovAnton (changes requested):

PeganovAnton:
Great fix! Reviewed punctuation part.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
itzsimpl: Done. I've added the docstring, and also modified the `@dataclass` so as to accept an optional `shard_strategy` and added the docstring there as well.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
itzsimpl: Ok. I've propagated the change to other tarred datasets, so as to keep the error the same on all occasions.

itzsimpl:
While running NLP/PC on 8 GPUs I've that the `lr_scheduler.max_steps` computation is still off. Indeed, the implementation of `compute_max_steps` assumes `num_samples` is the entire dataset size, not per worker, see L866. 
https://github.com/NVIDIA/NeMo/blob/fea3775c00adfacfe0a414dea15544abc96db8dc/nemo/core/optim/lr_scheduler.py#L861-L882

But, `DataLoader.len()` will in case of tarred datasets return the `num_samples` per worker, which will as per line https://github.com/NVIDIA/NeMo/blob/fea3775c00adfacfe0a414dea15544abc96db8dc/nemo/core/optim/lr_scheduler.py#L792
lead to a faulty computation of `max_steps`, i.e. instead of `max_epocs` it will optimise for `max_epocs/num_workers`.

Since prior to this PR `DataLoader.len()`, returned per worker, but the computation of it was faulty, I would like your advice on what is the correct solution. 

Multiplying the `num_samples` per `num_workers`, will probably lead to faulty behaviour with non tarred datasets, I guess. I can make `DataLoader.len()` return the correct total number of samples (not per worker), but I do not know if anywhere in the code `DataLoader.len()` is expected to return the number of samples per worker.

itzsimpl:
@titu1994 I would like your thoughts on what `DataLoader.len()` should return in case of `scattered` tarred datasets.

titu1994:
ASR tarred datasets always return the total number of samples in the manifest - https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/data/audio_to_text.py#L765

For tarred dataset we then limit the train dataset to have a fixed number of steps per epoch explicitly - https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/models/ctc_models.py#L420-L433

itzsimpl:
Thanks @titu1994. 

Note though that when the number of shards is not divisible by `world_size` the `scatter` shard strategy will not use all data, some shards (the non divisible portion, i.e. at max the last `world_size-1` shards, will be dropped), hence the actual number of samples will not be equal to the total number of samples in the manifest, i.e. `num_shards_per_rank * shard_size * world_size != len(self.manifest_processor.collection)`, and the reported length will be invalid even in the case of ASR tarred datasets. This could lead to the same problems as I've observed in NLP/PC. 

https://github.com/NVIDIA/NeMo/blob/fea3775c00adfacfe0a414dea15544abc96db8dc/nemo/collections/asr/data/audio_to_text.py#L186-L200

Assume you have 146 tarred files, each containing 1000 samples, and a `world_size` of 8, then
- `len(self.manifest_processor.collection) = 146000`
- each rank works with `floor(146/8) = 18` shards or `18000` samples, thus the total actual number of samples equals `18*8*1000 = 144000`
- assuming `self._trainer.limit_train_batches=1` and `train_data_config['batch_size']=1` https://github.com/NVIDIA/NeMo/blob/fea3775c00adfacfe0a414dea15544abc96db8dc/nemo/collections/asr/models/ctc_models.py#L427 will give you `18250`, where it should be `18000`.

titu1994:
In practice it doesn't matter, we shuffle tarfiles before each draw of samples anyway.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
```diff
@@ -142,6 +142,17 @@ class PunctuationCapitalizationDataConfigBase:
     tar_shuffle_n: int = 1
     """The size of shuffle buffer of `webdataset`. The number of batches which are permuted."""
 
+    shard_strategy: Optional[str] = 'scatter'
+    """Tarred dataset shard distribution strategy chosen as a str value during ddp. Accepted values are `scatter` and `replicate`.
+    `scatter`: The default shard strategy applied by WebDataset, where each node gets a unique set of shards, which are permanently
+    pre-allocated and never changed at runtime. `replicate` is an optional shard strategy, where each node gets all of the set of shards
+    available in the tarred dataset, which are permanently pre-allocated and never changed at runtime. The benefit of replication is that
+    it allows each node to sample data points from the entire dataset independently of other nodes, and reduces dependence on value of
+    `tar_shuffle_n`. Note: Replicated strategy allows every node to sample the entire set of available tarfiles, and therefore more than
```
PeganovAnton, line 151: Single backquote is for italic in sphinx. In this case, perforated font suits more.
```suggestion
    ``tar_shuffle_n``. Note: Replicated strategy allows every node to sample the entire set of available tarfiles, and therefore more than
```

nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
```diff
@@ -142,6 +142,17 @@ class PunctuationCapitalizationDataConfigBase:
     tar_shuffle_n: int = 1
     """The size of shuffle buffer of `webdataset`. The number of batches which are permuted."""
 
+    shard_strategy: Optional[str] = 'scatter'
```
PeganovAnton, line 145: This one is a great finding! We also need same description in https://github.com/NVIDIA/NeMo/blob/main/docs/source/nlp/punctuation_and_capitalization.rst in table "Parameters for tarred dataset".

PeganovAnton (changes requested):

nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py
```diff
@@ -741,6 +741,7 @@ def _setup_dataloader_from_config(self, cfg: DictConfig, train: bool) -> torch.u
                 world_size=self.world_size,
                 global_rank=self.global_rank,
                 shuffle_n=cfg.tar_shuffle_n,
+                shard_strategy=cfg.shard_strategy,
```
PeganovAnton, line 744: Great fix!

itzsimpl:
@titu1994 Correct me if I'm wrong, but based on the way it is implemented even in ASR shuffling occurs after the WebDataset is built from `scattered` tar files, i.e. a from a subset of all files. I don’t see any shuffling of `audio_tar_filepaths` prior to `expand_audio_filepaths`, which does the scattering.   

Hence, if my understanding is correct, in the above case of 8GPUs, 146 tar files with 1000 samples each, only the first 144 of 146 tar files will ever be used. A larger number of samples in `limit_train_batches` (due to total number of samples / world_size) will result in oversampling the shuffled tar files that were assigned to a specific rank. The ‘replicate’ strategy will on the other hand take samples from all 146 tar files.

  What I am seeing in NLP/PC is that in the above case:
- Prior to this PR, validation at the end of epoch was not run, due to an invalid number of samples returned by `DataLoader.len()`
- If `DataLoader.len()` returns the correct number of samples a rank has, the number of steps per epoch will be correct, but `lr_scheduler` will compute an invalid number of effective maximum steps; with `num_epochs` set, a simple linear `lr_scheduler` will end sooner than expected, at `num_epocs/world_size`, more complex schedulers may hide this problem.
- If I return the actual total length of the dataset, `lr_scheduler` will compute the correct number of effective maximum steps, but the progress bars will have a wrong number of samples. Assuming only 1 validation per epoch, validation will be run at approximately `1/world_size` and the epoch will end immediately afterwords.

I can:

1.  Make `DataLoader.len()` return the total number of samples and modify `limit_train_batches` in `setup_training_data` when the data is tarred, like ASR does, but this will in the case when the number of shards is non divisible by `world_size` and shard strategy is `scattered`, as mentioned earlier, lead to oversampling of a subset of all data. The only mention that not all data was used will be the shard strategy warning in the logs.
2. Make `DataLoader.len()` return the actual total number of samples and modify `limit_train_batches` in `setup_training_data` when the data is tarred, like ASR does, which will avoid oversampling, but since in the case when the number of shards is non divisible by `world_size` and shard strategy is `scattered` not all data is used, the total number of steps will be correct, however, also different from the case when shard strategy is `replicate` or non tarred data is used. The fact that not all data was used will, beside the shard strategy warning in the logs, reflect also in the total number of steps.

  
Let me know which solution do you prefer. Personally, I would opt for the second, especially since some collections use tarred datasets also for validation, where exclusion of data may be acceptable, but oversampling may be less so.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
```diff
@@ -871,6 +871,17 @@ class BertPunctuationCapitalizationTarredDataset(IterableDataset):
             be used in the current process.
         shuffle_n (:obj:`int`, `optional`, defaults to :obj:`1`): a number of shuffled batches in a buffer.
             ``shuffle_n`` batches are loaded into memory, shuffled, and then yielded by a dataset instance.
+        shard_strategy (str): Tarred dataset shard distribution strategy chosen as a str value during ddp.
```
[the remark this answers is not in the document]
itzsimpl: Done.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
```diff
@@ -871,6 +871,17 @@ class BertPunctuationCapitalizationTarredDataset(IterableDataset):
...
+            -   `scatter`: The default shard strategy applied by WebDataset, where each node gets
+                a unique set of shards, which are permanently pre-allocated and never changed at runtime.
+            -   `replicate`: Optional shard strategy, where each node gets all of the set of shards
+                available in the tarred dataset, which are permanently pre-allocated and never changed at runtime.
+                The benefit of replication is that it allows each node to sample data points from the entire
+                dataset independently of other nodes, and reduces dependence on value of `shuffle_n`.
```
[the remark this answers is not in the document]
itzsimpl: Done.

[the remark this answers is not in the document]
itzsimpl: Done.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
itzsimpl: Done. I've also updated docs.

nemo/collections/nlp/models/token_classification/punctuation_capitalization_model.py
itzsimpl: Thanx!

PeganovAnton (approved):

titu1994:
> Hence, if my understanding is correct, in the above case of 8GPUs, 146 tar files with 1000 samples each, only the first 144 of 146 tar files will ever be used. 

No, Webdataset shuffles tarfile list every time iter is called, which is after every epoch - https://github.com/webdataset/webdataset/blob/0.1.65/webdataset/dataset.py#L118

titu1994:
Option 1 is good enough and has been tested in ASR. Every epoch may not sample the whole dataset but over entire training run yes the full dataset is sampled.

titu1994:
You also should not use tarred dataset for validation and test sets. If you need truly global sampling use replication strategy, that's why it is there. But scattered does nearly the same thing and achieves nearly the same scores for ASR.

itzsimpl:
> > Hence, if my understanding is correct, in the above case of 8GPUs, 146 tar files with 1000 samples each, only the first 144 of 146 tar files will ever be used.
> 
> No, Webdataset shuffles tarfile list every time iter is called, which is after every epoch - https://github.com/webdataset/webdataset/blob/0.1.65/webdataset/dataset.py#L118

Sorry if I’m beeing pedantic here, but I would really like to get my head around this and understand the logic and operation. Going step by step over ASR data loader, for shard strategy `scatter`
- each rank creates its onwn dataloder only once
- the init of all ranks receives the entire set as `audio_tar_filepaths`
- `expand_audio_filepaths` reduces the dataset based on `world_size` and `rank_id` this becomes the new `audio_tar_filepaths`
- for each rank a webdataset is constructed from its own reduced set of `audio_tar_filepaths`
- the webdataset is shuffled on each epoch, independently on each rank
- during the epoch samples are taken from the webdataset, independently on each rank

Where is this description wrong? How can the webdataset shuffle the entire dataset if it was created from a subset? The link you provided shuffles only the data that the webdataset was created from.

titu1994:
You're right Webdataset does not shuffle entire dataset but only the portion it has on every rank for scatter mode. 

Then in that case, you're also right that it will drop tarfiles for scatter isn't globally divisible in the last rank (loosing num tarfiles % world size tarfiles). 

We usually do powers of two when building tarfiles so this case never occured for ASR domain. For PC where your number of samples is much larger, I would suggest avoiding scatter option altogether.

titu1994 (approved):
Seems good to me for ASR, thank you for the doc updated !

titu1994:
You also need style fixes

titu1994:
Hmm dco bot requires you to sign your commits - please follow the instructions on the DCO button. 

@PeganovAnton for final review.

titu1994 (approved):

MaximumEntropy:
DCO Rebase always messes things up. I think you need to `git checkout upstream/main -- untouched_file.py` to reset all the files that you didn't touch.

itzsimpl:
I'm a bit lost here, don't have much experience with rebasing. 

What I'm seeing is that under the `commits` section here there are a bunch of commits from the main branch, that are now signed of also by me. I don't think doing checkouts of files will unsigne those commits. I assume this happened because while working on this PR at some point, or multiple, main was merged in to keep the PR updated, and doing `git rebase HEAD~14 --signoff` followed by `git push --force-with-lease origin fix/tarred_dataset_len` as instruced by the DCO, signed those commits. 

I have a backup of the repo prior to doing the rebase, if this helps.

Please advise to what should I do, or let me know the status is acceptable the way it is.

titu1994:
Skip dco for now, we will override it.

titu1994:
To fix this, ignore dco bot and rebase with main branch using an ide in interactive mode, there shouldn't be any conflicts, and finally force push. Dco messed up here and made you sign commits you didn't touch.

MaximumEntropy:
The way it is right now is definitely not what its supposed to be before merging, reverting back to what you had before the DCO related commits might be the best thing to do and we can set DCO to pass.

titu1994:
Agreed let's get the Pr back to original state and override dco

itzsimpl:
OK. This should be it then.

docs/source/asr/datasets.rst
```diff
@@ -210,16 +210,22 @@ of filepaths, e.g. ``['/data/shard1.tar', '/data/shard2.tar']``, or in a single
...
 
-As with non-tarred datasets, the manifest file should be passed in ``manifest_filepath``. The dataloader assumes that the length 
+As with non-tarred datasets, the manifest file should be passed in ``manifest_filepath``. The dataloader assumes that the length
 of the manifest after filtering is the correct size of the dataset for reporting training progress.
 
-The ``tarred_shard_strategy`` field of the config file can be set if you have multiple shards and are running an experiment with 
+The ``tarred_shard_strategy`` field of the config file can be set if you have multiple shards and are running an experiment with
 multiple workers. It defaults to ``scatter``, which preallocates a set of shards per worker which do not change during runtime.
+Note that this strategy, on specific occasions (when the number of shards is not divisible with ``world_size``), will not sample
+the entire dataset. As an alternative the ``replicate`` strategy, will prealocat the entire set of shards to every worker and not
```
PeganovAnton, line 224:
```suggestion
the entire dataset. As an alternative the ``replicate`` strategy, will pre-allocate the entire set of shards to every worker and not
```

nemo/collections/nlp/data/token_classification/punctuation_capitalization_dataset.py
```diff
@@ -142,6 +142,18 @@ class PunctuationCapitalizationDataConfigBase:
...
+    """Tarred dataset shard distribution strategy chosen as a str value during ddp. Accepted values are `scatter` and `replicate`.
+    `scatter`: The default shard strategy applied by WebDataset, where each node gets a unique set of shards, which are permanently
+    pre-allocated and never changed at runtime. `replicate` is an optional shard strategy, where each node gets the entire set of shards
+    available in the tarred dataset, which are permanently pre-allocated and never changed at runtime. The benefit of replication is that
+    it allows each node to sample data points from the entire dataset independently of other nodes, and reduces dependence on value of
+    ``tar_shuffle_n``. Note: Replicated strategy allows every node to sample the entire set of available tarfiles, and therefore more than
+    one node may sample the same tarfile, and even sample the same data points! As such, there is no assured guarantee that all samples
+    in the dataset will be sampled at least once during 1 epoch. Scattered strategy, on the other hand, on specific occasions (when the
+    number of shards is not divisible with ``world_size``), will not sample the entire dataset.
```
PeganovAnton, line 154:
```suggestion
    ``tar_shuffle_n``.
    
    .. warning::
        Replicated strategy allows every node to sample the entire set of available tarfiles, and therefore more
        than one node may sample the same tarfile, and even sample the same data points! As such, there is 
        no assured guarantee that all samples in the dataset will be sampled at least once during 1 epoch. 
        Scattered strategy, on the other hand, on specific occasions (when the number of shards is not divisible
        with ``world_size``), will not sample the entire dataset.
```

itzsimpl: Done. Added to warning also that use of tarred datasets as validation or test dataset is not advised.

nemo/collections/nlp/data/token_classification/punctuation_capitalization_tarred_dataset.py
```diff
@@ -871,6 +871,21 @@ class BertPunctuationCapitalizationTarredDataset(IterableDataset):
...
+                available in the tarred dataset, which are permanently pre-allocated and never changed at runtime.
+                The benefit of replication is that it allows each node to sample data points from the entire
+                dataset independently of other nodes, and reduces dependence on value of :param:`shuffle_n`.
+
+                Note: Replicated strategy allows every node to sample the entire set of available tarfiles,
+                and therefore more than one node may sample the same tarfile, and even sample the same
+                data points! As such, there is no assured guarantee that all samples in the dataset will be
+                sampled at least once during 1 epoch. Scattered strategy, on the other hand, on specific
+                occasions (when the number of shards is not divisible with ``world_size``), will not sample
+                the entire dataset.
```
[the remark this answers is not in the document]
itzsimpl: Done. Added to warning also that use of tarred datasets as validation or test dataset is not advised.

docs/source/nlp/punctuation_and_capitalization.rst
```diff
@@ -588,6 +588,19 @@ For convenience, items of data config are described in 4 tables:
...
+       ``scatter``: Each node gets a unique set of shards, which are permanently pre-allocated and never changed at runtime, when the total
+       number of shards is not divisible with ``world_size``, some shards (at max ``world_size-1``) will not be used.
+       ``replicate``: Each node gets the entire set of shards available in the tarred dataset, which are permanently pre-allocated and never
+       changed at runtime. The benefit of replication is that it allows each node to sample data points from the entire dataset independently
+       of other nodes, and reduces dependence on value of ``tar_shuffle_n``.
+       Note: Replicated strategy allows every node to sample the entire set of available tarfiles, and therefore more than one node may sample
+       the same tarfile, and even sample the same data points! As such, there is no assured guarantee that all samples in the dataset will be
+       sampled at least once during 1 epoch. Scattered strategy, on the other hand, on specific occasions (when the number of shards is not
+       divisible with ``world_size``), will not sample the entire dataset.
```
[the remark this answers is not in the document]
itzsimpl: Done. Added to warning also that use of tarred datasets as validation or test dataset is not advised. Made also docstrings consistent throughout.

itzsimpl:
@MaximumEntropy you'll probably have to once again set DCO to pass.

PeganovAnton (approved):

itzsimpl:
@titu1994 @MaximumEntropy I guess one of you will need to set DCO to pass. If I follow its instructions and rebase it will mess things up.

PR #4564, opened 2022-07-19
[Fix] schedulers with no max_steps param

stevehuang52:
All LR schedulers in PyTorch do not have the `max_steps` parameter, so we should not add `max_steps` to their `scheduler_args`. Previous code tackle the problem in case-by-case manner, while here we try to solve the problem in a more general way. 

To do so, we define a set of schedulers that don't have the `max_steps` parameter, and set the `add_max_args_flag` accordingly. 

Similarly, we define a set of epoch-based schedulers to set the `interval`, while previous code only checks for `ReduceLROnPlateau` scheduler.

Besides, the previous code also has bugs, since it doesn't set the correct `add_max_args_flag` and `interval` when `if 'args' in scheduler_config` is `True`.


nemo/core/optim/lr_scheduler.py
```diff
@@ -681,18 +681,19 @@ def prepare_lr_scheduler(
...
-                interval = 'epoch'
-
             scheduler_args.pop('name', None)
             scheduler_args.pop('t_max_epochs', None)
             scheduler_args.pop('t_accumulate_grad_batches', None)
             scheduler_args.pop('t_limit_train_batches', None)
             scheduler_args.pop('t_num_workers', None)
             scheduler_args.pop('monitor', None)
             scheduler_args.pop('reduce_on_plateau', None)
+
```
titu1994, line 691: Avoid this and just do use python inspect module to check if the scheduler needs max_steps or not inside of get scheduler. Much simpler logic.

titu1994 (changes requested):

nemo/core/optim/lr_scheduler.py
stevehuang52: Thanks for the suggestion, I have changed to using `inspect.signature()` instead. By the way, why do we pass the `scheduler_args` twice, in `scheduler_cls = get_scheduler(scheduler_name, **scheduler_args)` and `schedule = scheduler_cls(optimizer, **scheduler_args)`?

titu1994: One is to get a class, another is to instantiate the class with those kwargs

titu1994 (approved):
Lgtm

titu1994 (approved):

titu1994:
@ericharper  for approval

evilc3:
Hey so with this pr the I don't need to define max_step param for any scheduler ?

stevehuang52:
> Hey so with this pr the I don't need to define max_step param for any scheduler ?

Not really, this PR aims to fix the bug that current code will add a max_steps param to some pytorch schedulers that don't require max_steps. 

For those schedulers that do have a max_steps param, current NeMo code will calculate a max_steps if it's not explicitly specified by the user.

PR #4565, opened 2022-07-19
Make RETRO SP compatible

yidong72:
# What does this PR do ?

Fix all the unit tests and CI tests to make RETRO model works with the current transformer model that supports the SP.


ericharper (approved):
LGTM. Thanks!

PR #4582, opened 2022-07-22
[TTS][ASR] customize arguments for trimming the leading/trailing silence

XuesongYang:
# What does this PR do ?

It supports overriding arguments of `librosa.effects.trim` when trimming silence. By default, we assign `trim: false` in YAML config. But when enabling it by `trim: true`, you have options to override original arguments of `librosa.effects.trim`, for example,

     trim: true
      trim_top_db: 50,
      trim_frame_length: ${model.n_window_size},
      trim_hop_length: ${model.n_window_stride},

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/tts/conf/de/fastpitch_align_44100.yaml
```diff
@@ -90,7 +90,10 @@ model:
       max_duration: 15  # change to null to include longer audios.
       min_duration: 0.1
       ignore_file: null
-      trim: false
+      trim: true
+      trim_top_db: 50,
```
rlangman, line 94: I am not sure what the rationale is for the default librosa config, other than it being very conservative. From my testing it seemed like trim_top_db = 30.0 worked well if you set trim_ref=np.max, and trim_top_db=50.0 if trim_ref=1.0.

Basically using trim_ref=np.max means it sets the threshold based on the maximum power in the audio sample, while using trim_ref=1.0 sets a constant threshold. So if trim_ref=np.max then the top_db needs to be smaller as it will lower the threshold based on how (loud?) the speaker is talking.

I am not sure what is best for a default, but we may as well make trim_ref a configurable value too if we want to use this.

examples/tts/conf/de/fastpitch_align_44100.yaml
```diff
@@ -90,7 +90,10 @@ model:
       max_duration: 15  # change to null to include longer audios.
       min_duration: 0.1
       ignore_file: null
-      trim: false
+      trim: true
```
rlangman, line 93: Are we sure we want to have this as a default configuration? I would think we should have the default trimming off, but let people know they can turn it on based on their particular dataset.

examples/tts/conf/de/fastpitch_align_44100.yaml
XuesongYang: The default value for `trim_ref` is `np.max`, which actually take the max of amplitude of audio signal, instead of power spectrum (https://github.com/librosa/librosa/blob/f39ef0ed4931da25a35b47aa194f5d6b9605053c/librosa/effects.py#L439). For general purpose trimming, it would be better to access the dynamic reference, i.e. `np.max`. Choosing `trim_ref=1.0` requires the PCM normalized within [-1, 1] ahead I guess.

XuesongYang: This YAML config serves for general purpose, so it would better to enable trim option, but we should mention potential tweaks in the tutorial.

nemo/collections/tts/torch/data.py
```diff
@@ -79,6 +79,10 @@ def __init__(
         min_duration: Optional[float] = None,
         ignore_file: Optional[Union[str, Path]] = None,
         trim: bool = False,
+        trim_ref: Optional[float] = None,
```
XuesongYang, line 82: @rlangman see here, `trim_ref` is an optional float type argument. We can override it in the YAML config. Given current use cases NeMo supported, it may be good to constrain it with a float number. We can extend it with a callable func in future if there is new use case available.

rlangman: Yea, you would only need to be able to override it with a float (not another numpy function).

examples/tts/conf/de/fastpitch_align_44100.yaml
rlangman: I am ok with enabling the trimming by default. But before that, can we somehow confirm that trimming the audio for that particular dataset/tutorial does not negatively impact the final model?

XuesongYang: that's what I am doing now. Training with top_db=50. Note that this yaml is for German only.

titu1994: Default should be off so as not to hurt later projects with unexpected changes

examples/tts/conf/de/fastpitch_align_44100.yaml
XuesongYang: since both @rlangman and @titu1994 agreed to assign `false` by default. I will revise it in next version. I will add params in the description.

titu1994 (commented):
Overall looks fine, I will leave to Ryan for final review

rlangman (approved):
Looks good to me. Thanks for adding this.

redoctopus (approved):
LGTM

PR #4594, opened 2022-07-24
A new fused softmax kernel function

yidong72:
# What does this PR do ?
Implemented the fused softmax GPU kernel that can be a drop-in replacement of the `apex`'s FusedScaleMaskSoftmax`. It addresses the following issues:
1. Removed the sequence length limitation. It works with tensor of any length.
2. Fix the `grad_input` modification issue https://github.com/NVIDIA/apex/issues/1430
3. Fix the problem when all keys are masked https://github.com/NVIDIA/apex/issues/1390

It uses the torch.cppextension to build the kernel dynamically like MegatronLM repo does. The kernels will be saved in the `/tmp` directory so it can be used in slurm with enroot clusters too.
It includes a unit test to cover a few cases.
The performance is comparable to the apex one. Here is a training step time graph when it is tested with training a RETRO model.
image
The blue curve is pytorch softmax. The bottom red curve is apex fused kernel and the middle red curve the fused kernel in this PR.

### DIFF
--- a/tests/collections/nlp/test_fused_softmax_kernel.py
+++ b/tests/collections/nlp/test_fused_softmax_kernel.py
@@ -33,13 +33,9 @@ def forward_torch_softmax(input, mask, scale):
 
 @pytest.mark.run_only_on('GPU')
 class TestFusedSoftmaxKernel:
-    @classmethod
-    def setup_class(cls):
-        # this line will trigger building the kernels
-        import nemo.collections.nlp.modules.common.megatron.fused_kernels
-
     @pytest.mark.unit
     def test_forward(self):
+        import nemo.collections.nlp.modules.common.megatron.fused_kernels
         import scaled_masked_softmax_cuda_new
 
         batch = 2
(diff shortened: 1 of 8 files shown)


tests/collections/nlp/test_fused_softmax_kernel.py
```diff
@@ -0,0 +1,98 @@
...
+    probs = torch.nn.Softmax(dim=-1)(mask_output)
+    all_k_masked = mask.all(axis=-1)
+    zero_attention_mask = (1.0 - all_k_masked.float())[:, :, :, None]
+    probs = probs * zero_attention_mask
+    return probs
+
+
+@pytest.mark.run_only_on('GPU')
+class TestFusedSoftmaxKernel:
+    @classmethod
```
titu1994, line 36: Run unittests only after checking apex is available, otherwise import of apex utils will crash in env that doesn't have apex

yidong72: fixed it

PR #4616, opened 2022-07-26
[TTS] Fix off-by-1 bug in Beta Binomial Prior

rlangman:
# What does this PR do ?

Fix a bug in the prior calculation used in the TTS aligner

**Collection**: [TTS]

# Changelog 
- Fix off-by-1 bug in prior calculation

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information

The prior here is supposed to help the TTS model align by teaching it to place more weight on phonemes along the diagonal, with it always starting on the first phoneme and ending on the last phoneme.

The current logic produces a prior that looks like this, which awkwardly ends earlier than expected with maximum weight on the last phoneme before the end of the sequence:

image

After this change, the prior looks like this:

image

I think the intuition is that we model the attention as a binomial distribution (eg. coin flips with head being success, tail failure) where each success results in the attention moving on to the next phoneme. To reach the end of a sequence of N phonemes, you only need (N - 1) successes.

### DIFF
--- a/nemo/collections/tts/torch/helpers.py
+++ b/nemo/collections/tts/torch/helpers.py
@@ -55,7 +55,7 @@ def beta_binomial_prior_distribution(phoneme_count, mel_count, scaling_factor=1.
     mel_text_probs = []
     for i in range(1, mel_count + 1):
         a, b = scaling_factor * i, scaling_factor * (mel_count + 1 - i)
-        mel_i_prob = betabinom(phoneme_count, a, b).pmf(x)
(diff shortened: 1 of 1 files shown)


blisc:
FYI @alancucki @rafaelvalle

rafaelvalle:
Very nice catch! Thank you!

redoctopus (approved):
Dang, thanks for catching this! Approved.

PR #4619, opened 2022-07-26
Support listing Hugging Face model info

titu1994:

# What does this PR do ?

Add new apis to filter and obtain NeMo hugging face models via code.

**Collection**: [Core]

# Changelog 
- Adds two methods `get_hf_model_filter()` and `search_huggingface_models(model_filter=None)` to ModelPT.
- First method returns a basic ModelFilter object, can be overridden by the user. Used as default to next method if nothing is provided.
- Add support for metadata attributes to the filter to modify the result list of ModelInfo.
- `list_available_models_on_hf()` lists the nemo models that satisfy the filter.

# Usage

```python
from nemo.core import ModelPT

# You can replace <DomainSubclass> with any subclass of ModelPT.
# Get default ModelFilter
filt = <DomainSubclass>.get_hf_model_filter()

# Make any modifications to the filter as necessary
filt.language = [...]
filt.task = ...
filt.tags = [...]

# Add any metadata to the filter as needed
filt.limit_results = 5

# Obtain model info
model_infos = <DomainSubclass>.search_huggingface_models(model_filter=filt)

# Browse through cards and select an appropriate one
card = model_infos[0]

# Restore model using `modelId` of the card.
model = ModelPT.from_pretrained(card.modelId)

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -23,12 +23,12 @@ from dataclasses import dataclass, field
 from enum import Enum
 from functools import total_ordering
 from pathlib import Path
-from typing import Dict, List, Optional, Union
+from typing import Dict, Iterable, List, Optional, Union
 
 import hydra
 import wrapt
-from huggingface_hub import hf_hub_download
-from huggingface_hub.hf_api import HfFolder
+from huggingface_hub import HfApi, HfFolder, ModelFilter, hf_hub_download
+from huggingface_hub.hf_api import ModelInfo
 from omegaconf import DictConfig, OmegaConf
 
 import nemo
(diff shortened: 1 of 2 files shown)


nemo/core/classes/common.py
```diff
@@ -680,6 +738,17 @@ def get_available_model_names(cls) -> List[str]:
             model_names = [model.pretrained_model_name for model in cls.list_available_models()]
         return model_names
 
+    @classmethod
+    def get_hf_model_filter(cls) -> ModelFilter:
```
okuchaiev, line 742: why this method needs to be public?

okuchaiev (changes requested):
Let's extend ``list_available_models`` method instead to accept optional ModelFilter instance. By default should be None and should list both NGC and HF models (even if duplicated). Then make ``list_available_models_on_hf`` private and call it under the hood

ericharper:
> Let's extend `list_available_models` method instead to accept optional ModelFilter instance. By default should be None and should list both NGC and HF models (even if duplicated). Then make `list_available_models_on_hf` private and call it under the hood

I agree with this. We don't want users to have to use two different methods to list models.

titu1994:
Reasons states above : it would require updating every single class that implements list_available_models which is impractical

okuchaiev (dismissed):

titu1994:
Updated PR with different name for method to remove confusion against list_available_models

ericharper (approved):
LGTM. Thanks for the changes!

PR #4675, opened 2022-08-04
Fix for incorrect batch size issue while decoding

rilango:
Fix to incorrect batch size during decoding.


michalivne (approved):
LGTM! Thanks!

PR #4638, opened 2022-08-01
 [chinese text normalization] PR for chinese TN part in text_normalization 

mzxcpp:
# What does this PR do ?

PR for chinese TN part in text_normalization 

**Collection**: NeMo/norm_text_processing/text_normalization

Chinese Text Normalization
1. How To Use
python normalize.py --lang "zh" --text "text to be normalized"
2. TN Pipeline
There are 3 components in TN pipeline:

pre-processing (before tagger)
non-standard word normalization
post-processing (after verbalizer)
2.1 Pre-Processing
Char Width Conversion (全角 -> 半角)
苹果ＣＥＯ宣布发布新ＩＰＨＯＮＥ -> 苹果CEO宣布发布新IPHONE
他说：“我们已经吃过了！”。 -> 他说:"我们已经吃过了!".
covers English letters, digits, punctuations and some symbols
the complete mapping table data/char/fullwidth_to_halfwidth.tsv
Blacklist (Removal)
Sometime you may want to remove certain things like interjections/fillers "啊", "呃" etc

呃这个呃啊我不知道 -> 这个我不知道
customizable via data/blacklist/interjections.tsv
2.2 Non-Standard-Words(NSW) normalization
Numbers
  共465篇，约315万字 -> 共四百六十五篇，约三百一十五万字
  共计6.42万人 -> 共计六点四二万人
  同比升高0.6个百分点 -> 同比升高零点六个百分点
Fraction
  总量的1/5以上 -> 总量的五分之一以上
  相当于头发丝的1/16 -> 相当于头发丝的十六分之一
  3/2是一个假分数 -> 二分之三是一个假分数
Percentage
  同比增长6.3% -> 同比增长百分之六点三
  增幅0.4% -> 增幅百分之零点四
Date
  2002/01/28 -> 二零零二年一月二十八日
  2002-01-28 -> 二零零二年一月二十八日
  2002.01.28 -> 二零零二年一月二十八日
  2002/01 -> 二零零二年一月
Time
  8月16号12:00之前 -> 八月十六号十二点之前
  我是5:02开始的 -> 我是五点零二分开始的
  于5:35:36发射 -> 于五点三十五分三十六秒发射
  8:00 a.m. 准时开会 -> 八点 a m 准时开会
Math
  比分定格在78:96 -> 比分定格在七十八比九十六
  计算-2的绝对值是2 -> 计算负二的绝对值是二
  ±2的平方都是4 -> 正负二的平方都是四
Money
[...]

Issue #4543, opened 2022-07-13
[text processing] Chinese text normalization PR


dophist:
This is a preliminary inquiry if NeMo team is open to accept PR for Chinese text normalization, or you guys have internal development plans for it ?

[SpeechColab](https://github.com/SpeechColab) is assigning 1 intern student to work on Chinese text normalization(based on Pynini), we'd like to contribute this work to NeMo, hopefully in the next couple of weeks.  Basically what we will have is a set of WFST rewrite rules just like existing [NeMo en TN](https://github.com/NVIDIA/NeMo/tree/main/nemo_text_processing/text_normalization/en), reproducing TN behaviors in [this project](https://github.com/speechio/chinese_text_normalization), which handles many typical TN scenarios for Chinese. 

**Known Issue**
The intern student will no longer be maintaining the grammars after the internship, but the resulting PR is guaranteed to establish a foundation, enabling Chinese TN in NeMo.

Would like to hear your thoughts on this. @yzhang123  @ekmb

ekmb:
We would appreciate the proposed PR! Looking forward to the contribution!

dophist:
Cool. I'll close this issue when the PR is ready.

PR #4683, opened 2022-08-05
[Chinese text normalization]PR for chinese TN part in text_normalization

mzxcpp:
# What does this PR do ?

new PR for previous chinese TN

**Collection**: [NeMo/norm_text_processing/text_normalization]
I think have covered the problems mentioned previously.And I make a new pr to make it clear.

### DIFF
--- a/nemo/collections/asr/models/ctc_bpe_models.py
+++ b/nemo/collections/asr/models/ctc_bpe_models.py
@@ -534,4 +534,11 @@ class EncDecCTCModelBPE(EncDecCTCModel, ASRBPEMixin):
         )
         results.append(model)
 
+        model = PretrainedModelInfo(
+            pretrained_model_name="stt_enes_conformer_ctc_large_codesw",
+            description="For details about this model, please visit https://ngc.nvidia.com/catalog/models/nvidia:nemo:stt_enes_conformer_ctc_large_codesw",
+            location="https://api.ngc.nvidia.com/v2/models/nvidia/nemo/stt_enes_conformer_ctc_large_codesw/versions/1.0.0/files/stt_enes_conformer_ctc_large_codesw.nemo",
+        )
+        results.append(model)
+
         return results
--- a/nemo/collections/asr/models/rnnt_bpe_models.py
+++ b/nemo/collections/asr/models/rnnt_bpe_models.py
@@ -197,6 +197,13 @@ class EncDecRNNTBPEModel(EncDecRNNTModel, ASRBPEMixin):
         )
         results.append(model)
 
+        model = PretrainedModelInfo(
+            pretrained_model_name="stt_enes_conformer_transducer_large_codesw",
+            description="For details about this model, please visit https://ngc.nvidia.com/catalog/models/nvidia:nemo:stt_enes_conformer_transducer_large_codesw",
+            location="https://api.ngc.nvidia.com/v2/models/nvidia/nemo/stt_enes_conformer_transducer_large_codesw/versions/1.0.0/files/stt_enes_conformer_transducer_large_codesw.nemo",
+        )
+        results.append(model)
+
         return results
 
     def __init__(self, cfg: DictConfig, trainer: Trainer = None):
--- a/scripts/speech_recognition/code_switching/README.md
+++ b/scripts/speech_recognition/code_switching/README.md
@@ -1,0 +1,15 @@
+# Scripts for creation of synthetic code-switched data from monolingual sources
+Follow the 2 steps listed below in order - 
+
+
+1. Create the (intermediate) manifest file using `code_switching_manifest_creation.py`.  It's usage is as follows:
+
+    `python code_switching_manifest_creation.py --manifest_language1 <absolute path of Language 1's manifest file> --manifest_language2 <absolute path of Language 2's manifest file> --manifest_save_path <absolute path to save the created manifest> --id_language1 <language code for language 1 (e.g. en)> --id_language2 <language code for language 2 (e.g. es)> --max_sample_duration_sec <maximum duration of generated sample in seconds> --min_sample_duration_sec <maximum duration of generated sample in seconds> --dataset_size_required_hrs <size of generated synthetic dataset required in hrs>`
+
+    Estimated runtime for dataset_size_required_hrs=10,000 is ~2 mins
+
+2. Create the synthetic audio data and the corresponding manifest file using `code_switching_audio_data_creation.py` It's usage is as follows: 
+
+    `python code_switching_audio_data_creation.py --manifest_path <absolute path to intermediate CS manifest generated in step 1> --audio_save_folder_path <absolute path to directory where you want to save the synthesized audios> --manifest_save_path <absolute path to save the created manifest> --audio_normalized_amplitude <scaled normalized amplitude desired> --cs_data_sampling_rate <desired sampling rate for generated audios> --sample_beginning_pause_msec <pause to be added to the beginning of the generated sample in milli seconds> --sample_joining_pause_msec <pause to be added between segments while joining, in milli seconds> --sample_end_pause_msec <pause to be added to the end of the generated sample in milli seconds> --workers <number of worker processes>`
+
+    Estimated runtime for generating a 10,000 hour corpus is ~40 hrs with a single worker
--- a/scripts/speech_recognition/code_switching/code_switching_audio_data_creation.py
+++ b/scripts/speech_recognition/code_switching/code_switching_audio_data_creation.py
@@ -1,0 +1,283 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import json
+import logging
+import os
+
+import librosa
+import numpy as np
+from joblib import Parallel, delayed
+from scipy.io import wavfile
+from tqdm import tqdm
+
+parser = argparse.ArgumentParser(description='Create synthetic code-switching data audio data from monolingual data')
+parser.add_argument("--manifest_path", default=None, type=str, help='Path to CS indermediate manifest', required=True)
+parser.add_argument(
+    "--audio_save_folder_path",
+    default=None,
+    type=str,
+    help='Path to directory where created synthetic set would be saved',
+    required=True,
+)
+parser.add_argument(
+    "--manifest_save_path", default=None, type=str, help='Path to save the created manifest', required=True
+)
+parser.add_argument(
+    "--audio_normalized_amplitude", default=15000, type=int, help='Normalized amplitdue of audio samples'
+)
+parser.add_argument(
+    "--cs_data_sampling_rate",
+    default=16000,
+    type=int,
+    help='Desired sampling rate for the audios in the generated dataset',
+)
+parser.add_argument(
+    "--sample_beginning_pause_msec",
+    default=20,
+    type=int,
+    help='Pause to be added at the beginning of the sample (msec)',
+)
+parser.add_argument(
+    "--sample_joining_pause_msec",
+    default=100,
+    type=int,
+    help='Pause to be added between different phrases of the sample (msec)',
+)
+parser.add_argument(
+    "--sample_end_pause_msec", default=20, type=int, help='Pause to be added at the end of the sample (msec)'
+)
+parser.add_argument("--workers", default=1, type=int, help='Number of worker processes')
+
+args = parser.parse_args()
+
+
+def read_manifest(manifest_path: str):
+    """
+    Args:
+        manifest_path: absolute path of the manifest file
+
+    Returns:
+        List with manifest entires as elements
+
+    """
+    data = []
+
+    for line in open(manifest_path, 'r'):
+        data.append(json.loads(line))
+
+    return data
+
+
+def split_list(input_list: list, num_splits: int):
+    """
+    Args:
+        input_list: the input list to split
+        num_splits: number of splits required
+
+    Returns:
+        iterator of split lists
+
+    """
+    k, m = divmod(len(input_list), num_splits)
+    return (input_list[i * k + min(i, m) : (i + 1) * k + min(i + 1, m)] for i in range(num_splits))
+
+
+def combine_manifests(manifest_save_path: str, num_split: int):
+    """
+    Args:
+        manifest_save_path: absolute path to save the combined manifest
+        num_splits: number of splits of manifest
+
+    Returns:
+        num_samples_combined: the total number of samples in the generated dataset
+    """
+    num_samples_combined = 0
+    base_directory = os.path.dirname(manifest_save_path)
+
+    with open(manifest_save_path, 'w') as outfile:
+        for i in range(num_split):
+            split_manifest_path = base_directory + '/temp_' + str(i) + '.json'
+            data_split = read_manifest(split_manifest_path)
+
+            for elem in data_split:
+                s = json.dumps(elem)
+                outfile.write(s + '\n')
+                num_samples_combined += 1
+
+            # removing the intermediate file
+            os.remove(split_manifest_path)
+
+    return num_samples_combined
+
+
+def create_cs_data(
+    intermediate_cs_manifest_list: list,
+    audio_save_folder: str,
+    manfest_save_path: str,
+    audio_amplitude_normalization: int,
+    pause_beg_msec: int,
+    pause_join_msec: int,
+    pause_end_msec: int,
+    cs_data_sampling_rate: int,
+):
+
+    """
+    Args:
+        intermediate_cs_manifest_list: the intermediate cs manifest obtained from code_switching_manifest_creation.py as a list
+        audio_save_folder: Absolute path to save the generated audio samples
+        manfest_save_path: Absolute path to save the corresponding manifest
+        audio_amplitude_normalization: The amplitude to scale to after normalization
+        pause_beg_msec: Pause to be added at the beginning of the sample (msec)
+        pause_join_msec: Pause to be added between different phrases of the sample (msec)
+        pause_end_msec: Pause to be added at the end of the sample (msec)
+        cs_data_sampling_rate: Desired sampling rate of the generated samples
+
+    Returns:
+
+    """
+
+    fs = cs_data_sampling_rate
+    incorrect_sample_flag = 0
+
+    with open(manfest_save_path, 'w') as outfile:
+        for data in tqdm(intermediate_cs_manifest_list):
+
+            combined_audio = []
+
+            staring_pause = np.zeros(int(pause_beg_msec * fs / 1000))
+            combined_audio += list(staring_pause)
+
+            for index in range(len(data['lang_ids'])):
+
+                data_sample, fs_sample = librosa.load(data['paths'][index], sr=fs)
+                # Alternative-  fs_sample, data_sample = wavfile.read(data['paths'][index])
+
+                if fs_sample != fs:
+                    logging.error('Sampling rate error inside create_cs_data function')
+                    exit
+
+                # Remove leading and trailing zeros
+                data_sample = np.trim_zeros(data_sample)
+
+                # take care of empty arrays: rare
+                if data_sample.size == 0:
+                    incorrect_sample_flag = 1
+                    continue
+
+                # normalizing data
+                data_sample_norm = (
+                    data_sample
+                    / np.maximum(np.abs(data_sample.max()), np.abs(data_sample.min()))
+                    * audio_amplitude_normalization
+                )
+
+                combined_audio += list(data_sample_norm)
+
+                # adding small pause between gemgments
+                if index != (len(data['lang_ids']) - 1):
+                    pause = np.zeros(int(pause_join_msec * fs / 1000))
+                    combined_audio += list(pause)
+
+            if incorrect_sample_flag == 1:
+                incorrect_sample_flag = 0
+                continue
+
+            ending_pause = np.zeros(int(pause_end_msec * fs / 1000))
+            combined_audio += list(ending_pause)
+
+            sample_id = data['uid']
+            audio_file_path = audio_save_folder + '/' + str(sample_id) + ".wav"
+
+            # saving audio
+            wavfile.write(audio_file_path, fs, np.array(combined_audio).astype(np.int16))
+            # Alternative-  librosa.output.write_wav(audio_file_path, combined_audio, fs)
+
+            metadata_json = {}
+            metadata_json['audio_filepath'] = audio_file_path
+            metadata_json['duration'] = float(len(combined_audio) / fs)
+            metadata_json['text'] = ' '.join(data['texts'])
+
+            metadata_json['language_ids'] = data['lang_ids']
+            metadata_json['original_texts'] = data['texts']
+            metadata_json['original_paths'] = data['paths']
+            metadata_json['original_durations'] = data['durations']
+
+            s = json.dumps(metadata_json)
+            outfile.write(s + '\n')
+
+
+def main():
+
+    cs_intermediate_manifest_path = args.manifest_path
+    audio_save_folder = args.audio_save_folder_path
+    manifest_save_path = args.manifest_save_path
+    audio_amplitude_normalization = args.audio_normalized_amplitude
+    pause_beg_msec = args.sample_beginning_pause_msec
+    pause_join_msec = args.sample_joining_pause_msec
+    pause_end_msec = args.sample_end_pause_msec
+    cs_data_sampling_rate = args.cs_data_sampling_rate
+    num_process = args.workers
+
+    # Sanity Checks
+    if (cs_intermediate_manifest_path is None) or (not os.path.exists(cs_intermediate_manifest_path)):
+        logging.error('Please provide correct CS manifest (obtained from code_switching_manifest_creation.py)')
+        exit
+
+    if (audio_save_folder is None) or (not os.path.exists(audio_save_folder)):
+        logging.error('audio_save_folder_path is incorrect or does not exist')
+        exit
+
+    if manifest_save_path is None:
+        logging.error('Please provide valid manifest_save_path')
+        exit
+
+    # Reading data
+    logging.info('Reading manifests')
+    intermediate_cs_manifest = read_manifest(cs_intermediate_manifest_path)
+
+    # Spliting the data
+    data_split = split_list(intermediate_cs_manifest, num_process)
+
+    # Creating Audio data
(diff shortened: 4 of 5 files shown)


dophist:
@BuyuanCui @ekmb  This PR continues https://github.com/NVIDIA/NeMo/issues/4543 & https://github.com/NVIDIA/NeMo/pull/4638 , ready for another round of reviews.

nemo_text_processing/text_normalization/zh/data/char/lower_to_upper.tsv
```diff
@@ -0,0 +1,26 @@
...
+q	Q
+r	R
+s	S
+t	T
+u	U
+v	V
+w	W
+x	X
+y	Y
+z	Z
```
XuesongYang, line 26: Thanks for your PR. Is there any reason we need this mapping? It seems a trivia op in Python.

nemo_text_processing/text_normalization/zh/data/char/upper_to_lower.tsv
```diff
@@ -0,0 +1,26 @@
...
+Q	q
+R	r
+S	s
+T	t
+U	u
+V	v
+W	w
+X	x
+Y	y
+Z	z
```
XuesongYang, line 26: Same question as the above. Is there any reason we need this mapping as a standalone file? There are lower() and upper() functions in Python.

nemo_text_processing/text_normalization/zh/data/money/currency.tsv
```diff
@@ -0,0 +1,21 @@
...
+AUD	澳元
+A$	澳元
+SUR	卢布
+₽	卢布
+DEM	马克
+DM	马克
+FRF	法郎
+CAD	加元
+CAD$	加元
+FRF	法郎
```
XuesongYang, line 21: nit: is it possible to store currency codes (e.g. USD) and symbols (e.g. $) in separate files?

XuesongYang: It seems `currency.tsv` file only consider a bunch of currencies. My guess is that the selected currency set occurs most frequently in the Chinese datasets used here. Could you please share any thoughts? Given https://www.eurochange.co.uk/travel-money/world-currency-abbreviations-symbols-and-codes-travel-money, it seems many more currencies worth to add in this file.

nemo_text_processing/text_normalization/zh/verbalizers/time.py
```diff
@@ -0,0 +1,98 @@
...
+class Time(GraphFst):
+    '''
+        tokens { time { h: "1" m: "02" s: "36" } } -> 一点零二分三十六秒
+        tokens { time { suffix "am"  hours: "1" minutes: "02" seconds: "36" } } -> 上午一点零二分三十六秒
+    '''
+
+    def __init__(self, deterministic: bool = True, lm: bool = False):
+        super().__init__(name="time", kind="verbalize", deterministic=deterministic)
+        graph_digit = pynini.string_file(get_abs_path("data/cardinal/digit.tsv"))
+        graph_ten = pynini.string_file(get_abs_path("data/cardinal/digit_teen.tsv"))
```
XuesongYang, line 36: I observed there are three places loaded `"data/cardinal/digit_teen.tsv"`, but they declared two different variables below, please unify the name space. 
* `graph_ten`: line 36 in `text_normalization/zh/verbalizers/time.py`; line 30 in `text_normalization/zh/verbalizers/date.py`.
* `graph_teen`: line 48 in `text_normalization/zh/taggers/cardinal.py`.

nemo_text_processing/text_normalization/zh/data/char/lower_to_upper.tsv
mzxcpp: In this case,we want to process in graph and it could be set as a part of postprocessor instead of applying to string in public normalize.py.I have the plan to add chinese upper and lower forms in this tsv when I got engough tsv data.just like trans "壹“ to ”一“

nemo_text_processing/text_normalization/zh/taggers/tokenize_and_classify.py
```diff
@@ -0,0 +1,92 @@
...
+            char = Char(deterministic=deterministic)
+            fraction = Fraction(deterministic=deterministic)
+            math_symbol = MathSymbol(deterministic=deterministic)
+            money = Money(deterministic=deterministic)
+            measure = Measure(deterministic=deterministic)
+            time = Time(deterministic=deterministic)
+            whitelist = Whitelist(deterministic=deterministic)
+
+            classify = pynini.union(
+                pynutil.add_weight(date.fst, 0.4),
```
yzhang123, line 77: could you please keep class weights between [1.0, 1.1] and default chars 100, @fayejf, Alex Cui for help

tests/nemo_text_processing/zh/data_text_normalization/test_cases_time.txt
```diff
@@ -0,0 +1,4 @@
+8月16号12:00之前 -> 八月十六号十二点之前
```
yzhang123, line 1: ~ as delimiter

tests/nemo_text_processing/zh/test_char.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class TestChar:
+    normalizer_zh = Normalizer(lang='zh', cache_dir=CACHE_DIR, overwrite_cache=False)
+
+    @parameterized.expand(parse_test_case_file('zh/data_text_normalization/test_cases_char.txt'))
+    @pytest.mark.run_only_on('CPU')
+    @pytest.mark.unit
+    def test_norm_money(self, expected, test_input):
+        preds = self.normalizer_zh.normalize(test_input, n_tagged=50)
```
yzhang123, line 29: change `preds= self.normalizer_zh.normalize(test_input, n_tagged=50) ` to pred= self.normalizer_zh.normalize(test_input,)` this is deterministic normalization

and assert expected in preds -> assert expected ==  pred

tests/nemo_text_processing/zh/test_char.py
```diff
@@ -0,0 +1,30 @@
...
+from ..utils import CACHE_DIR, parse_test_case_file
+
+
+class TestChar:
+    normalizer_zh = Normalizer(lang='zh', cache_dir=CACHE_DIR, overwrite_cache=False)
+
+    @parameterized.expand(parse_test_case_file('zh/data_text_normalization/test_cases_char.txt'))
+    @pytest.mark.run_only_on('CPU')
+    @pytest.mark.unit
+    def test_norm_money(self, expected, test_input):
```
yzhang123, line 28: fix name

tests/nemo_text_processing/zh/test_sparrowhawk_normalization.sh
```diff
@@ -0,0 +1,60 @@
...
+
+testTNMoneyText() {
+  input=$PROJECT_DIR/zh/data_text_normalization/test_cases_money_text.txt
+  runtest $input
+}
+testTNCharText() {
+  input=$PROJECT_DIR/zh/data_text_normalization/test_cases_char_text.txt
+  runtest $input
+}
+testTNClockText() {
```
yzhang123, line 32: change to time

tests/nemo_text_processing/zh/test_time.py
```diff
@@ -0,0 +1,30 @@
...
+from ..utils import CACHE_DIR, parse_test_case_file
+
+
+class TestClock:
+    normalizer_zh = Normalizer(lang='zh', cache_dir=CACHE_DIR, overwrite_cache=False)
+
+    @parameterized.expand(parse_test_case_file('zh/data_text_normalization/test_cases_time.txt'))
+    @pytest.mark.run_only_on('CPU')
+    @pytest.mark.unit
+    def test_norm_clock(self, expected, test_input):
```
yzhang123, line 28: change to time

tests/nemo_text_processing/zh/test_time.py
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+import pytest
+from nemo_text_processing.text_normalization.normalize import Normalizer
+from parameterized import parameterized
+
+from ..utils import CACHE_DIR, parse_test_case_file
+
+
+class TestClock:
```
yzhang123, line 22: change to time

yzhang123 (changes requested):
Thank you so much! 
A few comments: 
1. could you please add more test cases for every class? i noticed there are no cardinal tests, decimals tests etc.
2. Also our pipeline does not run our code, so could you please make sure the pytests run? you do this by cd NeMo/tests; pytest nemo_text_processing/zh/test_*.py --cpu --tn_cache_dir cache, only 2 out of all tests passed
3. also make sure the sparrowhawk tests are passing, by running cd NeMo/tools/text_processing_deployment; bash export_grammars.sh --GRAMMARS=tn_grammars  --LANGUAGE=zh --MODE=test --OVERWRITE_CACHE=True
4. could you adjust the weights of each semiotic class to be between [1, 1.01] see https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/text_normalization/en/taggers/tokenize_and_classify.py#L165, 

The rest is ok, for language specific details like coverage of test cases, @fayejf and AlexCui could you add your review?

nemo_text_processing/text_normalization/zh/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+### 2.3 Post-Processing
+#### Punctuation Removal
+If enabled, punctuations are removed.
+
+#### Uppercase or Lowercase Conversion
+If enabled, English letters are converted to uppercases / lowercases
+
+#### Out-Of-Vocabulary(OOV) Tagger
+If enabled, OOV chars are tagged with `<oov>` and `</oov>`, e.g.:
```
yzhang123, line 123: why do we need this? can't we pass through everything as plain text? this is at most preprocessor,  but i would keep this optional and let the user set it

yzhang123 (changes requested):

nemo_text_processing/text_normalization/zh/taggers/cardinal.py
```diff
@@ -0,0 +1,149 @@
...
+        5       -> cardinal{ integer: "五"}
+        12      -> cardinal{ integer: "十二"}
+        213     -> cardinal{ integer: "二百一十三"}
+        3123    -> cardinal{ integer: "三千一百二十三"}
+        3,123   -> cardinal{ integer: "三千一百二十三"}
+        51234   -> cardinal{ integer: "五万一千二百三十四"}
+        51,234  -> cardinal{ integer: "五万一千二百三十四"}
+        0.125   -> cardinal{ integer: "零点一二五"}
+        self.fst:
+        123     -> cardinal{ integer: "一二三" }
```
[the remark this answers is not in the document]
mzxcpp: Thank you for your advice！The opinion I  handle with number is:①when there is a unit,we read it into cardinal forms,just like "20头牛、125个人、3000元",so the measure part is used to normalize number with unit both in English and Chinese.②When it comes to a single number, just like  "747 or 10086"only, although we prefer reading it to"七百四十七 or 一万零八十六", in real cases,the correct forms may be “波音七四七、编号七四七人员、拨打一二三零六".
To put it simply，the previous work just transport number into cardinal forms and I just found it has a lower accuracy in lots of real cases we tested.Therefore,to handle with more detailed cases,I set it more detailed,just like now.I think we could not deal with all the cases by one form,and the present forms may lead to more accuracy in Chinese Text Normalization.

dophist:
@mzxcpp please give clear signals here once you have addressed all issues from last review, so that we know when to move forward, and everyone won't get distracted by intermediate development activities.

mzxcpp:
Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with  sparrowhawk related.

nemo_text_processing/text_normalization/zh/taggers/cardinal.py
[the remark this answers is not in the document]
yzhang123: Thank you @mzxcpp  for the explanation. Very interesting observation. @BuyuanCui, did you see similar things like this and can confirm this?
@mzxcpp is it possible to share the real cases you tested on? that would be highly interesting.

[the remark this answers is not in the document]
mzxcpp: Thank you for taking my idea into consideration!I just base on real cases in Leaderboard.
https://github.com/SpeechColab/Leaderboard
and
https://github.com/dophist/textnorm/blob/main/zh/misc/speechio_0_15.tsv
and I have just gone through some cases and find the following cases:

原知识产权5条热线和平台的整合，实现12315一个号码对外一个平台受理，方便消费者。
针对禁止报出的波音737ng型飞机被发现了机身出现裂缝一事。美国波音公司10月31日宣布。
在5个月内连续发生两次空难后，目前波音公司已暂停向客户交付t37-8及t37-9新飞机。
当天在伊朗首都德黑兰附近机场失事的波音707货机的黑匣子已经找到。
春节期间，各界群众若发现涉黄涉非违法行为，可通过全国扫黄打非办公室举报电话12390。
鲁豫探访奇虎360，带您走进这位互联网传奇老炮周弘毅。
2002年一度领先市场份额的3721逐渐迎来强劲对手。
感谢3158创业信息网的大力支持。
您这个是号是000~008。
身高1米84的克洛泽似乎是在有意避开人们对他只会投球的质疑。
最近呢有个小朋友跟我说，呢他们公司准备要开展996工作制，就是早上9点工作到晚上9点。
1234这4个队员会选择进攻。这样一来先防守他们。
510有点有一点点的偏局红调唉510我试一下吧510应该还可以。我试一下510好不好？
DC37烟熏玫瑰色，所有女生都可以用的颜色。

These cases will go wrong if we set each number into cardinal forms.And it's difficult to try to cover all the special forms with number string.I think it may be easier to cover all the common units rather than all this special forms.

Admittedly,there are cases with wrong answer:

一年每年呢给你免24,000，如果你是非独生子女。
9999995这么大。

In these cases,speaker ignores the unit because in some situation,others can know the correct meaning in Chinese,just like "元" may be left out.
Just my experimental idea and I will follow your ideas whether to set it as cardinal forms or keep the current forms.

yzhang123:
> Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.


Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box

mzxcpp:
> > Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.
> 
> Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box

Yes,the TN output is deterministic and I think it will work sucessfully in this case.However,when i test this,there is an error as "ModuleNotFoundError: No module named 'nemo_text_processing'".Should I install an another version of nemo in my env instead of running the sparrowhawk test in local nemo files?

yzhang123:
> > > Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.
> > 
> > 
> > Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box
> 
> Yes,the TN output is deterministic and I think it will work sucessfully in this case.However,when i test this,there is an error as "ModuleNotFoundError: No module named 'nemo_text_processing'".Should I install an another version of nemo in my env instead of running the sparrowhawk test in local nemo files?

you need to install nemo and nemo_text_processing is installed using https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/install_pynini.sh

mzxcpp:
> > > > Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.
> > > 
> > > 
> > > Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box
> > 
> > 
> > Yes,the TN output is deterministic and I think it will work sucessfully in this case.However,when i test this,there is an error as "ModuleNotFoundError: No module named 'nemo_text_processing'".Should I install an another version of nemo in my env instead of running the sparrowhawk test in local nemo files?
> 
> you need to install nemo and nemo_text_processing is installed using https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/install_pynini.sh

Thank you!But what can I do if a tagger is not included in sparrowawk?Should I remove them all or keep them without test_sparrowhawk?  I find that if a tagger's form is not included in the semiotic_classes.proto,it will not work successfully.

yzhang123:
> > > > > Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.
> > > > 
> > > > 
> > > > Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box
> > > 
> > > 
> > > Yes,the TN output is deterministic and I think it will work sucessfully in this case.However,when i test this,there is an error as "ModuleNotFoundError: No module named 'nemo_text_processing'".Should I install an another version of nemo in my env instead of running the sparrowhawk test in local nemo files?
> > 
> > 
> > you need to install nemo and nemo_text_processing is installed using https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/install_pynini.sh
> 
> Thank you!But what can I do if a tagger is not included in sparrowawk?Should I remove them all or keep them without test_sparrowhawk? I find that if a tagger's form is not included in the semiotic_classes.proto,it will not work successfully. The other part just work successfully like time or currency in this TN.

which classes' tagger is not included? optimally we want to have a  semiotic_classes.proto or expand it so it supports all classes. you can add the pytest still, just don't include them in test_sparrowhawk until we figured out how to expand the .proto file

mzxcpp:
> > > > > > Thank you for your detailed review again!I have fixed some naming problems and now make sure the pytest will work.However,with only TN part I have designed,it may be difficult for me to deal with work with sparrowhawk related.
> > > > > 
> > > > > 
> > > > > Is your TN output non-deterministic? if not, sparrowhawk should run Out of the box
> > > > 
> > > > 
> > > > Yes,the TN output is deterministic and I think it will work sucessfully in this case.However,when i test this,there is an error as "ModuleNotFoundError: No module named 'nemo_text_processing'".Should I install an another version of nemo in my env instead of running the sparrowhawk test in local nemo files?
> > > 
> > > 
> > > you need to install nemo and nemo_text_processing is installed using https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/install_pynini.sh
> > 
> > 
> > Thank you!But what can I do if a tagger is not included in sparrowawk?Should I remove them all or keep them without test_sparrowhawk? I find that if a tagger's form is not included in the semiotic_classes.proto,it will not work successfully. The other part just work successfully like time or currency in this TN.
> 
> which classes' tagger is not included? optimally we want to have a semiotic_classes.proto or expand it so it supports all classes. you can add the pytest still, just don't include them in test_sparrowhawk until we figured out how to expand the .proto file

Well,I checked my code again and reorganize it.
I think math in tagger in which I designed detailed graph for cases like "25:26" into"二十五比二十六"，may be we call this "score"
All in all,I think these are just small cases and I have work succeessfully with "pytest" and "sparrowhawk" test.I upload the latest version.
Although there may more detailed cases I don't cover,I think it now can handle with most common cases in Chinese and I hope this can be a base of Chinese TN so that we can add more details to get a perfect TN tool for Chinese in further works.

mzxcpp:
The local setup fix nothing for current code but I can't pass CI test this time.
Is it because I changed something in tools/text_processing_deployment?

pengzhendong:
`他说：“我们已经吃过了！”。` => `他说比“我们已经吃过了!”。`
`这儿有只鸟儿` => `normalize.py", line 363, in _permute raise ValueError()`

mzxcpp:
> `他说：“我们已经吃过了！”。` => `他说比“我们已经吃过了!”。` `这儿有只鸟儿` => `normalize.py", line 363, in _permute raise ValueError()`

Thank you!I have fixed it.

XuesongYang:
Could you please try to fix the alerts as described here: https://lgtm.com/projects/g/NVIDIA/NeMo/rev/pr-f0d82e6d82662b60dc1c597bc403b4ac6dc1615a? Thanks.

mzxcpp:
> Could you please try to fix the alerts as described here: https://lgtm.com/projects/g/NVIDIA/NeMo/rev/pr-f0d82e6d82662b60dc1c597bc403b4ac6dc1615a? Thanks.

Ok,I have deleted the useless lines in the lastest commit

nemo_text_processing/text_normalization/zh/graph_utils.py
```diff
@@ -0,0 +1,134 @@
...
+    Args:
+        name: name of grammar class
+        kind: either 'classify' or 'verbalize'
+        deterministic: if True will provide a single transduction option,
+            for False multiple transduction are generated (used for audio-based normalization)
+    """
+
+    def __init__(self, name: str, kind: str, deterministic: bool = True):
+        self.name = name
+        self.kind = str
```
pengzhendong, line 81: `self.kind = kind`

pengzhendong (changes requested):

nemo_text_processing/text_normalization/zh/graph_utils.py
pengzhendong: There are some useless variables in `graph_utils.py`, such as `HEX`, `NON_BREAKING_SPACE`, `GRAPH`, and so on.

mzxcpp: > `self.kind = kind`

Thank you.I just copy that file from en and I will remove useless things.But self.kind = str is also in en.Is there also a bug in that?

nemo_text_processing/text_normalization/zh/taggers/cardinal.py
```diff
@@ -0,0 +1,141 @@
...
+# limitations under the License.
+import pynini
+from nemo_text_processing.text_normalization.zh.graph_utils import GraphFst
+from nemo_text_processing.text_normalization.zh.utils import UNIT_1e01, UNIT_1e02, UNIT_1e03, UNIT_1e04, get_abs_path
+from pynini.lib import pynutil
+
+
+class Cardinal(GraphFst):
+    '''
+        self.graph_cardinal:
```
pengzhendong, line 22: `self.graph_cardinal` does not contain tokens. It should be

```
5  -> 五
12 -> 十二
```

pengzhendong (changes requested):

nemo_text_processing/text_normalization/zh/graph_utils.py
pengzhendong: I think it is a bug.

nemo_text_processing/text_normalization/zh/taggers/cardinal.py
```diff
@@ -0,0 +1,141 @@
...
+                + pynutil.insert(UNIT_1e03)
+                + delete_punct
+                + graph_zero
+                + graph_digit
+                + pynutil.insert(UNIT_1e01)
+                + graph_digit_no_zero
+            )
+            | (graph_digit + pynutil.insert(UNIT_1e03) + delete_punct + graph_zero + graph_no_zero + graph_digit)
+            | (graph_digit + pynutil.insert(UNIT_1e03) + delete_punct + graph_no_zero ** 3)
+        )
```
pengzhendong, line 85:
```
|  closure(...)
 |      closure(self, lower)
 |      closure(self, lower, upper)
 |      
 |      Computes concatenative closure.
 |      
 |      This operation destructively converts the FST to its concatenative closure.
 |      If A transduces string x to y with weight w, then the zero-argument form
 |      `A.closure()` mutates A so that it transduces between empty strings with
 |      weight 1, transduces string x to y with weight w, transduces xx to yy with
 |      weight w \otimes w, string xxx to yyy with weight w \otimes w \otimes w
 |      (and so on).
 |      
 |      When called with two optional positive integer arguments, these act as
 |      lower and upper bounds, respectively, for the number of cycles through the
 |      original FST that the mutated FST permits. Therefore, `A.closure(0, 1)`
 |      mutates A so that it permits 0 or 1 cycles; i.e., the mutated A transduces
 |      between empty strings or transduces x to y.
 |      
 |      When called with one optional positive integer argument, this argument
 |      acts as the lower bound, with the upper bound implicitly set to infinity.
 |      Therefore, `A.closure(1)` performs a mutation roughly equivalent to
 |      `A.closure()` except that the former does not transduce between empty
 |      strings.
 |      
 |      The following are the equivalents for the closure-style syntax used in
 |      Perl-style regular expressions:
 |      
 |      Regexp:             This method:            Copy shortcuts:
 |      
 |      /x?/                x.closure(0, 1)         x.ques
 |      /x*/                x.closure()             x.star
 |      /x+/                x.closure(1)            x.plus
 |      /x{N}/              x.closure(N, N)         x ** N
 |      /x{M,N}/            x.closure(M, N)
 |      /x{N,}/             x.closure(N)
 |      /x{,N}/             x.closure(0, N)
```

Please try to use `x.ques` to avoid the redundant code.

pengzhendong (changes requested):

nemo_text_processing/text_normalization/zh/taggers/cardinal.py
pengzhendong, line 85:
```
|  Regexp:  This method:     Copy shortcuts:
|  /x?/     x.closure(0, 1)  x.ques
```

I think it's better to use `delete_punct.ques` to avoid the redundant code.

pengzhendong (changes requested):

nemo_text_processing/text_normalization/zh/taggers/char.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Char(GraphFst):
+    '''
+        你 -> tokens { char: "你" }
+    '''
+
+    def __init__(self, deterministic: bool = True, lm: bool = False):
+        super().__init__(name="char", kind="classify", deterministic=deterministic)
+
+        graph = pynutil.insert("name: \"") + NEMO_CHAR + pynutil.insert("\"")
```
pengzhendong, line 26: `pynutil.insert("char: \"")`

pengzhendong (changes requested):

nemo_text_processing/text_normalization/zh/taggers/char.py
mzxcpp: It used to be "char" and I followed the previous suggestions from ekmb to change it into name as the basic char.Should I change it to char again?

pengzhendong: OK, may be you should chang this line, too.

```
你 -> tokens { char: "你" }
```

mzxcpp: Ok.I have solved the above suggestions.

yzhang123:
@mzxcpp could you please remove the unrelated files from your PR:

image

After that, i think we can merge it for now. Verified that pytests and SH tests are passing.

nemo_text_processing/text_normalization/zh/graph_utils.py
yzhang123: yes it is, thanks for catching it @pengzhendong , @mzxcpp could you please fix it in this file and all other files?

yzhang123 (changes requested):

PR #4685, opened 2022-08-05
Move trainer calls for ssl models to training and validation steps only

sam1373:
# What does this PR do ?

Move trainer calls for ssl models to training and validation steps only

**Collection**: ASR


titu1994 (approved):
Looks good !

PR #4700, opened 2022-08-08
Port Huggingface T5v1_1 weights to NeMo-Megatron

MaximumEntropy:
# What does this PR do ?

Adds a conversion script and related compatibility args to port Huggingface T5v1_1 weights to NeMo-Megatron.

**Collection**: NLP

# Changelog 
- Adds a state dict conversation script.
- Adds necessary args to yaml configs to make sure the forward passes in HF and NeMo result in identical results.

# Usage

```bash
python scripts/nlp_language_modeling/hf_t5_to_nemo_coverter.py \
    --hf_model_name google/ul2 \
    --nemo_state_dict /path/to/nemo_ul2_state_dict.pt
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -400,7 +406,7 @@ def forward(
             query_layer.transpose(0, 1),  # [b * np, sq, hn]
             key_layer.transpose(0, 1).transpose(1, 2),  # [b * np, hn, sk]
             beta=0.0,
-            alpha=(1.0 / self.norm_factor),
+            alpha=(1.0 / self.norm_factor) if self.normalize_attention_scores else 1.0,
```
yidong72, line 409: why we don't want to normalize attention scores? And if `apply_query_key_layer_scaling` is set to true, this norm factor is need to counter the effect of `coeff` in the softmax.
```python
 self.scale_mask_softmax = FusedScaleMaskSoftmax(
            self.fp16,
            self.bf16,
            self.attn_mask_type,
            masked_softmax_fusion,
            attention_mask_func,
            self.attention_softmax_in_fp32,
            coeff,
        )
```
If it is just a speical thing we want to turn it off, maybe we can do something similar to my muTransfer code, so we don't have to add this argument throughout different models?
https://github.com/NVIDIA/NeMo/blob/f921ebe0436e55f7547b183ca83a623f6678422d/examples/nlp/language_modeling/megatron_retro_mutransfer_pretrain.py#L114

MaximumEntropy: Looks like Google didn't use this norm factor in their released T5 checkpoints, they accounted for it in the Q,K,V projection initialization. Check out these lines in their code - https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L796. Will take a look at what you did for MuTransfer.

yidong72: Yeah. so it is very specific to this google model. Let's handle it as a special case then.

MaximumEntropy: How do you suggest we address it as a special case? I added the special arg of `normalize_attention_scores` which is False only for Google models.

PR #4713, opened 2022-08-09
Adding RETRO model Faiss sharding index and KNN sharding index

yidong72:
# What does this PR do ?
To handle very large dataset, e.g. hundreds of gigabyte to terrabyte compressed raw data, we need multiple nodes to create sharding index  and combine them together.
In this PR, it creates GPU Faiss index in 3 stages. 
  1. stage 0, train Faiss index structure. 
  2. stage 1, create Faiss sharding index
  3. stage 2, merge sharding index into one.
It creates KNN index in 2 stages.
  1. stage 1, create KNN sharding index
  2. stage 2, merge KNN sharding index into one.


nemo/collections/nlp/models/language_modeling/megatron_retrieval_model.py
```diff
@@ -86,7 +89,43 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
         else:
             raise ValueError('precision must be in [32, 16, "bf16"]')
         self.model.model_type = ModelType.encoder_and_decoder
-        # self.grad_clip_pl_default = True
+
+        if hasattr(self.cfg, "shape_file"):
+            # self.grad_clip_pl_default = True
```
michalivne, line 94: Is this comment required?

nemo/collections/nlp/models/language_modeling/megatron_retrieval_model.py
```diff
@@ -86,7 +89,43 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
...
+            set_base_shapes(self, self.cfg.shape_file, rescale_params=False)
+            # add shape file
+            self.register_artifact("model.shape_file", self.cfg.shape_file),
+
+            for name, tensor in self.named_parameters():
```
michalivne, line 99: Can you add a comment explaining this initialization scheme?

nemo/collections/nlp/modules/common/megatron/transformer.py
```diff
@@ -636,12 +636,24 @@ def _checkpointed_attention_forward(
         """Forward method with activation checkpointing."""
 
         def custom_forward(*inputs):
-            query_layer = inputs[0]
-            key_layer = inputs[1]
-            value_layer = inputs[2]
-            attention_mask = inputs[3]
-            rotary_pos_emb = inputs[4]
-            relative_position_bias = inputs[5]
+            if len(inputs) == 7:
```
michalivne, line 639: This looks very brittle.Perhaps we can use here the same solution we used in `nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py` with `self._get_forward_output_only_func` ?

scripts/nlp_language_modeling/build_knn_map_index.py
```diff
@@ -110,19 +160,55 @@ def get_tokenizer(args):
...
+        total_chunks = splits[shard_id + 1]
+    elif shard_id == total_shards - 1:
+        start = splits[shard_id]
+        total_chunks = total_chunks
+    else:
+        raise ValueError(f'{shard_id} bigger than {total_shards}')
+    return start, total_chunks
+
+
+def process_sentence_chunks(
```
michalivne, line 177: Can you add a docsring describing what the function?

scripts/nlp_language_modeling/build_knn_map_index.py
```diff
@@ -201,22 +292,82 @@ def get_emb():
...
+        path = input_file.parent
+        fname = input_file.name
+        all_files = [str(i) for i in pathlib.Path(path).glob(fname + '*')]
+        merge_knn_files(all_files, args.output_file)
+        f = KNNIndex(args.output_file)
+        logging.info(f'Write to {args.output_file},  Size of Index : {f.len}')
+        logging.info(f'Index neighbors: {f.K}')
+        logging.info(f'Index chunk start id: {f.chunk_start_id}')
+        logging.info(f'Index chunk end id: {f.chunk_end_id}')
+        sys.exit(0)
```
michalivne, line 337: Can you add a few more high level comments in the code below explaining the purpose of code blocks?

scripts/nlp_language_modeling/exam_knn_map_quality.py
```diff
@@ -105,7 +105,7 @@ def get_tokenizer(args):
     logging.info(f'Data index has {data_ds.chunks} chunks')
     logging.info(f'Retrieval Data index has {retrieval_ds.chunks} chunks')
     logging.info(f'KNN index has {knn_index.K} neighbors')
-    assert knn_index.knn_map.max() < retrieval_ds.chunks
+    # assert knn_index.knn_map.max() < retrieval_ds.chunks
```
michalivne, line 108: Remove comment if not needed

scripts/nlp_language_modeling/build_retrieval_index.py
```diff
@@ -68,7 +125,15 @@ def get_tokenizer(args):
 
 
 def process_sentence_chunks(
-    ds: MMapRetrievalIndexedDataset, tokenizer, chunk_size: int, warm_up_size: int, percent: float
+    ds: MMapRetrievalIndexedDataset,
```
michalivne, line 128: Can you add a docstring?

michalivne (approved):
LGTM! Great PR. See minor comments regarding missing docstrings/comments

nemo/collections/nlp/models/language_modeling/megatron_retrieval_model.py
yidong72: no. removed it

nemo/collections/nlp/models/language_modeling/megatron_retrieval_model.py
yidong72: added

scripts/nlp_language_modeling/build_knn_map_index.py
yidong72: added

scripts/nlp_language_modeling/build_knn_map_index.py
yidong72: I do have a high level comment about the script at the beginning of the file with some examples of how to use it.

scripts/nlp_language_modeling/build_retrieval_index.py
yidong72: added

scripts/nlp_language_modeling/exam_knn_map_quality.py
yidong72: removed

scripts/nlp_language_modeling/build_knn_map_index.py
```diff
@@ -61,37 +100,48 @@
...
+            if beg <= target_chunk_id < end:
+                # target chunk is from the same document
+                continue
+            tmp_neighbors[cid, position] = target_chunk_id
+            position += 1
+
+
+def dedup(chunk_id_to_range, I, tmp_neighbors, chunk_id_start, offset):
     """
     deduplicate the KNN who are from the same document as the data chunks.
```
soumye, line 133: Why do we want to do this? Can we not have multiple neighbors from a given document? Or am I misunderstanding something here?

soumye, line 133: Why do we want to do this? Can we not have multiple neighbors from a given document? Or am I misunderstanding something here?

scripts/nlp_language_modeling/build_knn_map_index.py
```diff
@@ -175,6 +264,8 @@ def get_emb():
         default='bert-base-nli-mean-tokens',
```
soumye, line 264: @yidong72 I thought you were now using `all-mpnet-base-v2`

scripts/nlp_language_modeling/build_retrieval_index.py
```diff
@@ -34,11 +35,67 @@
 It creates a index.sav which can be loaded by Faiss. It can look up the KNN chunk ids of the 
 DB dataset given the input embedding vector. 
 
+To use it in multiple stages, it follows the example as shown in  
+https://github.com/facebookresearch/faiss/blob/main/demos/demo_ondisk_ivf.py
+
+stage-0: train on the dataset, example,
+
+```python
+python scripts/nlp_language_modeling/build_retrieval_index.py \
```
soumye, line 44: Training uses the entirety of the dataset right?  Can you tell a bit what `train_index_size` & `train_chunk_size` do here?

scripts/nlp_language_modeling/build_retrieval_index.py
```diff
@@ -117,10 +217,10 @@ def get_emb():
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="build Faiss index",)
     parser.add_argument(
-        '--input_file', type=str, required=True, help='Input file',
+        '--input_file', type=str, required=False, help='Input file',
     )
     parser.add_argument(
-        '--train_index_size', type=int, required=True, help='The number of sentences that is used to train the index',
+        '--train_index_size', type=int, required=False, help='The number of sentences that is used to train the index',
```
soumye, line 223: So do you use the entirety or only a subset of data to train the index?

scripts/nlp_language_modeling/build_retrieval_index.py
```diff
@@ -191,31 +350,63 @@ def get_emb():
...
-    index.add(emb)
+    if args.stage is None or args.stage == 0:
+        emb, slice_id = get_emb()
+        # initialize the Faiss index
+        # m is number of subquantizers. So vector of size D is broken into m sub-vectors of size D/m
+        m = args.subquantizers
+        k = 4  # num_nearest neighbors to get
+        quantizer = faiss.IndexFlatIP(emb.shape[1])
+        # 8 specifies that each sub-vector is encoded as 8 bits
+        index = faiss.IndexIVFPQ(quantizer, emb.shape[1], nlist, m, 8)
```
soumye, line 361: @yidong72 I think it's worth looking into other indices particularly hnsw as well. In NMT it didn't make much of a difference but I am not sure in RETRO how much impact the quality of NNs would have. My guess would be quite a lot so some analysis on improving nearest neighbor quality we can look at.

soumye (commented):
The PR looks great to me. I have added some comments.

nemo/collections/nlp/modules/common/megatron/transformer.py
yidong72: I checked the `self._get_forward_output_only_func` method, which has a help function to convert dictionary arguments to positional arguments. Here I have to translate positional arguments into dictionary arguments. Not sure what's a better solution.

scripts/nlp_language_modeling/build_knn_map_index.py
yidong72: This is RETRO's trick to train on the same data that is used as retrieval data. We simply remove the neighbors from the same document as the input data chunk.

yidong72: see above

scripts/nlp_language_modeling/build_knn_map_index.py
yidong72: yes. this is just the default argument. When using this script, I set it to `all-mpnet-base-v2`

scripts/nlp_language_modeling/build_retrieval_index.py
yidong72: use a subset of the data. It is impractical to use all the data as my dataset can be as large as 200billion tokens.

scripts/nlp_language_modeling/build_knn_map_index.py
soumye: sounds good

scripts/nlp_language_modeling/build_retrieval_index.py
soumye: I see. Makes sense. I am guessing it shouldn't even matter that much for performance.

scripts/nlp_language_modeling/build_retrieval_index.py
soumye: Yeah fair point. GPU acceleration and sharding are important.

michalivne (approved):
LGTM!

PR #4718, opened 2022-08-10
[TTS] FastPitch training: speed up align_prior_matrix calculation

racoiaws:
# What does this PR do ?

Speed up align_prior_matrix calculation and don't save it to disk. They take lots of space (tens of gigabytes per experiment) and it is perfectly fine to compute them on-the-fly. Matrix calculation speed-up is significant.

```python
%timeit beta_binomial_prior_distribution_numpy_old(80, 300)
226 ms ± 5.86 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit beta_binomial_prior_distribution_new_torch(80, 300)
586 µs ± 22 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

**Collection**: TTS

# Changelog 

# Usage

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/tts/torch/data.py
+++ b/nemo/collections/tts/torch/data.py
@@ -521,7 +521,7 @@ class TTSDataset(Dataset):
             if self.use_beta_binomial_interpolator:
                 align_prior_matrix = torch.from_numpy(self.beta_binomial_interpolator(mel_len, text_length.item()))
             else:
-                align_prior_matrix = beta_binomial_prior_distribution(text_length, mel_len)
+                align_prior_matrix = torch.from_numpy(beta_binomial_prior_distribution(text_length, mel_len))
 
         non_exist_voiced_index = []
         my_var = locals()
--- a/nemo/collections/tts/torch/data.py
+++ b/nemo/collections/tts/torch/data.py
@@ -153,7 +153,6 @@ class TTSDataset(Dataset):
             highfreq (Optional[int]): The highfreq input to the mel filter calculation. Defaults to None.
         Keyword Args:
             log_mel_folder (Optional[Union[Path, str]]): The folder that contains or will contain log mel spectrograms.
-            align_prior_matrix_folder (Optional[Union[Path, str]]): The folder that contains or will contain align prior matrices.
             pitch_folder (Optional[Union[Path, str]]): The folder that contains or will contain pitch.
             voiced_mask_folder (Optional[Union[Path, str]]): The folder that contains or will contain voiced mask of the pitch
             p_voiced_folder (Optional[Union[Path, str]]): The folder that contains or will contain p_voiced(probability) of the pitch
@@ -390,15 +389,6 @@ class TTSDataset(Dataset):
                 )
 
     def add_align_prior_matrix(self, **kwargs):
-        self.align_prior_matrix_folder = kwargs.pop('align_prior_matrix_folder', None)
-
-        if self.align_prior_matrix_folder is None:
-            self.align_prior_matrix_folder = Path(self.sup_data_path) / AlignPriorMatrix.name
-        elif isinstance(self.align_prior_matrix_folder, str):
-            self.align_prior_matrix_folder = Path(self.align_prior_matrix_folder)
-
-        self.align_prior_matrix_folder.mkdir(exist_ok=True, parents=True)
-
         self.use_beta_binomial_interpolator = kwargs.pop('use_beta_binomial_interpolator', False)
         if not self.cache_text:
             if 'use_beta_binomial_interpolator' in kwargs and not self.use_beta_binomial_interpolator:
@@ -527,19 +517,11 @@ class TTSDataset(Dataset):
         # Load alignment prior matrix if needed
         align_prior_matrix = None
         if AlignPriorMatrix in self.sup_data_types_set:
+            mel_len = self.get_log_mel(audio).shape[2]
             if self.use_beta_binomial_interpolator:
-                mel_len = self.get_log_mel(audio).shape[2]
                 align_prior_matrix = torch.from_numpy(self.beta_binomial_interpolator(mel_len, text_length.item()))
             else:
-                prior_path = self.align_prior_matrix_folder / f"{rel_audio_path_as_text_id}.pt"
-
-                if prior_path.exists():
-                    align_prior_matrix = torch.load(prior_path)
-                else:
-                    mel_len = self.get_log_mel(audio).shape[2]
-                    align_prior_matrix = beta_binomial_prior_distribution(text_length, mel_len)
-                    align_prior_matrix = torch.from_numpy(align_prior_matrix)
-                    torch.save(align_prior_matrix, prior_path)
+                align_prior_matrix = beta_binomial_prior_distribution(text_length, mel_len)
 
         non_exist_voiced_index = []
         my_var = locals()
--- a/nemo/collections/tts/torch/helpers.py
+++ b/nemo/collections/tts/torch/helpers.py
@@ -16,14 +16,15 @@ from pathlib import Path
 
 import numpy as np
 import torch
+from einops import rearrange
 from scipy import ndimage
-from scipy.stats import betabinom
+from torch.special import gammaln
 
 
 class BetaBinomialInterpolator:
     """
-        This module calculates alignment prior matrices (based on beta-binomial distribution) using cached popular sizes and image interpolation.
-        The implementation is taken from https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/FastPitch/fastpitch/data_function.py
+    This module calculates alignment prior matrices (based on beta-binomial distribution) using cached popular sizes and image interpolation.
+    The implementation is taken from https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/FastPitch/fastpitch/data_function.py
     """
 
     def __init__(self, round_mel_len_to=50, round_text_len_to=10, cache_size=500):
@@ -38,7 +39,7 @@ class BetaBinomialInterpolator:
     def __call__(self, w, h):
         bw = BetaBinomialInterpolator.round(w, to=self.round_mel_len_to)
         bh = BetaBinomialInterpolator.round(h, to=self.round_text_len_to)
-        ret = ndimage.zoom(self.bank(bw, bh).T, zoom=(w / bw, h / bh), order=1)
+        ret = ndimage.zoom(self.bank(bw, bh).numpy().T, zoom=(w / bw, h / bh), order=1)
         assert ret.shape[0] == w, ret.shape
         assert ret.shape[1] == h, ret.shape
         return ret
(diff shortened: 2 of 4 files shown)


racoiaws:
I'd also remove the `BetaBinomialInterpolator` class, apparently it exists just for caching `beta_binomial_prior_distribution`

nemo/collections/tts/torch/data.py
```diff
@@ -516,19 +506,13 @@ def __getitem__(self, index):
         # Load alignment prior matrix if needed
         align_prior_matrix = None
         if AlignPriorMatrix in self.sup_data_types_set:
+            # formula derived after investigating depths of librosa
+            mel_len = audio.flatten().shape[-1] + 2 * int(self.n_fft / 2) - (self.n_fft - 1)
+            mel_len = math.ceil(mel_len / self.hop_length)
```
XuesongYang, line 511: is there any reason why to replace `mel_len = self.get_log_mel(audio).shape[2]` with manual operations?

XuesongYang:
add @alancucki and @rafaelvalle for vis.

nemo/collections/tts/torch/data.py
racoiaws: Now that `beta_binomial_prior_distribution` is optimized, `get_log_mel` might take the majority for time needed for computing alignment prior matrices (stft + matmul inside). I transfered this from a scratch notebook where this was a concern. Not entirely sure if it makes any difference in routine data loading but is still nice to have.

Could benchmark if needed.

redoctopus: Can you make a note that this is equivalent to `self.get_log_mel(audio).shape[2]` in the comment?

nemo/collections/tts/torch/helpers.py
```diff
@@ -50,14 +51,28 @@ def general_padding(item, item_len, max_len, pad_value=0):
...
-    for i in range(1, mel_count + 1):
-        a, b = scaling_factor * i, scaling_factor * (mel_count + 1 - i)
-        mel_i_prob = betabinom(phoneme_count - 1, a, b).pmf(x)
-        mel_text_probs.append(mel_i_prob)
-    return np.array(mel_text_probs)
+def logbeta(x, y):
+    return gammaln(x) + gammaln(y) - gammaln(x + y)
+
+
+def logbetabinom(n, a, b, x):
```
rlangman, line 58: Is there a way to implement this using out-of-the-box logic in Pytorch distributions?

https://pytorch.org/docs/stable/distributions.html

nemo/collections/tts/torch/helpers.py
```diff
@@ -50,14 +51,28 @@ def general_padding(item, item_len, max_len, pad_value=0):
...
+def beta_binomial_prior_distribution(
+    phoneme_count: int, mel_count: int, scaling_factor: float = 1.0
+) -> torch.FloatTensor:
+    x = rearrange(torch.arange(0, phoneme_count), "b -> 1 b")
+    y = rearrange(torch.arange(1, mel_count + 1), "b -> b 1")
+    a = scaling_factor * y
+    b = scaling_factor * (mel_count + 1 - y)
+    n = torch.FloatTensor([phoneme_count])
+
+    return logbetabinom(n, a, b, x).exp()
```
rlangman, line 75: (1) How are we testing this to ensure that the results from the previous and new implementation are identical?

(2) In general, can we add unit tests for these relatively complicated algorithmic modules?

The previous implementation already had 2 bugs in it due to insufficient testing (using phoneme_count instead of phoneme_count - 1, and the BetaBinomialInterpolator class still uses its phoneme_count and mel_count parameters backwards)

racoiaws:
I did a closer-to-real benchmark: iterate through LibriTTS train-clean-100 dataset (one epoch), `batch size=32`, `num_workers=16`, no training

Baseline: 
- 800s 1st epoch, 77s per epoch onwards, +14GB for a 9G dataset
- 800s per epoch when not saving to SSD

BetaBinomialInterpolator: 
- 114s per epoch

PyTorch:
- 72s per epoch, same with no mel spec calculation



So, PyTorch implementation is much faster than baseline; on par with caching but no extra space and I/O. Also, `self.get_log_mel(audio).shape[2]` is good enough, no need to replace it with explicit formula

Will be updating the PR shortly

nemo/collections/tts/torch/helpers.py
racoiaws: (1) I've checked results, numerically differences are under 1e-5. Same thing but vectorized.

(2) Could you elaborate on what we would need to test? Tensor shapes? Small distance to precomputed answer?

nemo/collections/tts/torch/helpers.py
racoiaws: I don't immediately see a neat way to implement this using PyTorch distributions.

PyTorch doesn't have beta-binomial distribution, and to implement its pdf we need gamma and beta functions rather than corresponding distribution likelihoods.

nemo/collections/tts/torch/helpers.py
racoiaws: Fixed phoneme_count - 1

nemo/collections/tts/torch/helpers.py
rlangman: I also don't see any neat way to implement it. PyTorch distributions looks very limited compared to Scipy and Tensorflow probability, both in terms of functionality and interfaces.

nemo/collections/tts/torch/helpers.py
rlangman: (1) It would be great if you could show this data in the PR (but its unnecessary if you do (2)). You could do that by posting an example of the prior graph for the output for the old and new function, like I did in https://github.com/NVIDIA/NeMo/pull/4616.

The code I used for generating the graph is:

```
fig, ax = plt.subplots(figsize=(12,5))
im = ax.imshow(prior.transpose(), origin='lower', aspect='auto', interpolation='None')
fig.colorbar(im, ax=ax)
ax.set_ylabel('Phonemes')
ax.set_xlabel('Mel Frames')
plt.savefig("prior.png")
```

(2) I think small distance to precomputed answer is the main one. Tensor shape a nice to have. Other nice-to-have tests could be:

- If this supports batch compute (I am guessing not since inputs are int rather than tensor), then validate that the batch compute produces same results as calling individually on multiple inputs.
- Validate that the logbetabinom() function itself produces the same output as scipy.statsbetabinom().pmf()
- Could validate some high level characteristics, eg. first/last frame puts max weight on first/fast phoneme, values near edges are close to 0.

racoiaws: If there's a need for smoke tests we can just log the prior matrix right next to alignment matrix and take a look there if something is wrong; I myself wouldn't want to spend time writing tests for a straightforward implementation of a simple formula.

On batch compute: `collate_fn` is the one responsible for composing batches out of individual samples, no need to think about that here

requirements/requirements_common.txt
```diff
@@ -2,3 +2,4 @@ sentencepiece<1.0.0
 youtokentome>=1.0.5
 pandas
 sacremoses>=0.0.43
+einops
```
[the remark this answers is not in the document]
racoiaws: Done

blisc (dismissed):

nemo/collections/tts/torch/helpers.py
rlangman: _I myself wouldn't want to spend time writing tests for a straightforward implementation of a simple formula._

Why rely on manually logging a matrix (with the alignment matrix which are not even logging today either), when one can spend 10 minutes writing a test to automate it which will catch any unintentional regressions in the future?

_On batch compute: collate_fn is the one responsible for composing batches out of individual samples, no need to think about that here_

`collate_fn `processes every element separately and then concatenates the results, which is usually several times slower than batching the computation and masking it. Though if this function is fast enough, then its an unnecessary optimization since the prior is not computed at inference time.

racoiaws: I believe such tests won't take 10 minutes to write and are more prone to errors than the function we are discussing.

Is this a requirement for merging the PR?

rlangman: If it takes much longer than that, then that is a problem with us lacking templates/documentation on how to write tests. And then if it that slow, then I suppose it might not be worth the time.

All that should be needed is:

1. Take your current code and serialize the matrix for 1 output in a numpy file.
2. Copy the numpy file into the test directory.
3. Have the test run the expected inputs through the function and assert the output is the same shape and value (within 1E-5 tolerance) as the matrix loaded from the numpy file.

This will make it so that if someone wants to modify this code in the future, its safe and there is already easily runnable code that that person can use to test/benchmark the updates.

Despite being a "straightforward implementation of a simple formula" the original code before you modified it as mentioned already had 2 bugs, yours presumably had the same off-by-1 bug before it was mentioned, and its more complicated and bug-prone than 95% of other software that typically requires automated tests.

Personally if the PR is not urgent to merge then I would like for it to be a requirement until we discuss the expectations around automated testing with the team. Sorry if it would be unfair to use this particular CR as the starting point for such a discussion.

racoiaws: Thank you for your explanations, I am aware what automated software testing is. I believe in this case it is not worth time and effort, both coding and further discussion. I also disagree with the suggestion to put a serialized matrix into a git repo.

This PR is a non-urgent, nice-to-have improvement, and the amount of friction is surprising. I’d really rather not spend any more attention on it than I already did. Correctness was demonstrated above visually, code is reasonably clean, and the approval’s been already received.

If you wish to start a greater discussion on testing standards within the team, please do so, I would like to participate. I also appreciate you taking the time to write elaborate responses in this thread.

rlangman: Approving after discussing the topics offline. Thanks for PR and the constructive comments.

rlangman (approved):

nemo/collections/tts/torch/helpers.py
```diff
@@ -16,14 +16,15 @@
 
 import numpy as np
 import torch
+from einops import rearrange
 from scipy import ndimage
-from scipy.stats import betabinom
+from torch.special import gammaln
 
 
 class BetaBinomialInterpolator:
```
kevjshih, line 24: I think there are two small changes missing in the BetaBinomialInterpolator. 

1. The default border setting for the zoom function is to use constant padding, which can result in a column of zero probabilities. I think this is easily resolved using mode='nearest'.
2. The probabilities for phoneme_token given mel frame do not exactly add up to 1 after interpolation (just slightly off), so a renormalization would be necessary.

racoiaws: Hi! I agree with both points, that would be a more correct implementation.

However, I'd rather deprecate the whole `BetaBinomialInterpolator`. It's is a cache that is not needed anymore precisely due to changes in this PR.

kevjshih: Right! I totally missed that part :)

PR #4764, opened 2022-08-18
Add EMA support to NeMo

SeanNaren:
# What does this PR do ?

Adds Exponential Moving Average (EMA) support. We've seen promising results in reducing training times to reach convergence.

Things to do:

- [x] Capture speed degradation using the EMA callback.

There is practically none when the EMA weights are already on CUDA. Below are convergence results using LibriSpeech (SP) and the Conformer RNNT config for ASR:

**Precision 32**

WER vs Global Step

Screenshot 2022-09-14 at 15 33 17

WER vs Process Time

Screenshot 2022-09-14 at 15 34 40

**Precision 16**

WER vs Global Step

Screenshot 2022-09-14 at 15 33 43

WER vs Process Time

Screenshot 2022-09-14 at 15 34 17

- [x] Explore keeping EMA weights on the CPU (transfer to GPU just for the multi-tensor apply, then bring it back to CPU)
- [x] Test using Quartznet/Conformer to ensure we see expected improvements using the EMA callback
- [ ] Add multi-node/gpu correctness test (EMA weights should be the same across processes) to jenkinsfile

**Collection**:
- Common

# Changelog 
- Adds Exponential Moving Average (EMA) support.

# Usage

The callback can be added to pure lightning, however it requires Apex to be installed (for the multi-tensor apply).

```python
...
lenet5 = MNISTLeNet5(MNISTLeNet5Config())
lenet5.setup_training_data()
lenet5.setup_optimization()

trainer = Trainer(
    accelerator='gpu',
    devices=1,
    callbacks=EMA(ema=0.999)
)
trainer.fit(model=lenet5)
```

Currently I've exposed this through the exp manager (questionable if this is the right approach or not) to use through the hydra configs.
```bash
python examples/asr/asr_ctc/speech_to_text_ctc.py \
    model.train_ds.manifest_filepath=... \
    model.validation_ds.manifest_filepath=... \
    trainer.devices=1 \ 
    +exp_manager.enable_ema=True
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

cc @titu1994


nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,124 @@
...
+    When training a model, this callback will maintain moving averages of the trained parameters.
+    When evaluating, we use the moving averages copy of the trained parameters.
+
+    Args:
+        ema: The exponential decay used when calculating the moving average. Has to be between 0-1.
+        apply_ema_every_n_steps: Apply EMA every n global steps.
+        start_step: Start applying EMA from ``start_step`` global step onwards.
+    """
+
+    def __init__(self, ema: float, apply_ema_every_n_steps: int = 1, start_step: int = 0):
```
SeanNaren, line 46: Part of me wants to change `ema` to `decay` as that's closer to what it really is.

Also it seems a default of `0.999` works well, but hopefully with experimentation we can set this as default.

nemo/utils/exp_manager.py
```diff
@@ -119,6 +121,9 @@ class ExpManagerConfig:
     # Configures creation of log files for different ranks
     log_local_rank_0_only: Optional[bool] = False
     log_global_rank_0_only: Optional[bool] = False
+    # todo (sean): might not be the best place to add this, but allows us to enable it fast
```
MaximumEntropy, line 124: Maybe we can put this as part of our optimizer's configs?

titu1994: Not part of optim/sched. Those pertain only the the constructors of the optimizer and scheduler, not logging.

SeanNaren: It's a bit tricky to feel where exactly this should go. @titu1994 is there a registry of callbacks somewhere that may be this can be registered to add to the callbacks list?

titu1994: Exp manager technically has no registry, but that's a very good idea to implement. We already have the timer callback that we inject into ExpManager maybe lets register that as well as this one

SeanNaren:
@titu1994 would I be able to get a review here when you get time?

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class EMA(Callback):
+    """
+    Implements Exponential Moving Averaging (EMA).
+
+    When training a model, this callback will maintain moving averages of the trained parameters.
+    When evaluating, we use the moving averages copy of the trained parameters.
+    When saving, we save an additional set of parameters with the prefix `ema`.
+
```
titu1994, line 41: Maybe link to the PT EMA page for explanation of what we want to achieve.

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+        self,
+        decay: float,
+        apply_ema_every_n_steps: int = 1,
+        start_step: int = 0,
+        save_ema_weights_in_callback_state: bool = False,
+        evaluate_ema_weights_instead: bool = False,
+    ):
+        if not apex_available:
+            raise MisconfigurationException(
+                "EMA requires Apex to be installed: https://github.com/NVIDIA/apex#installation."
```
titu1994, line 62: Huh ? Why would apex be required for ema ?
We can make it optional - use apex codepath if available, use basic Python loop and torch ops if not. No need to limit ema to apex

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+        self._cur_step: Optional[int] = None
+        self._weights_buffer: Optional[List[torch.Tensor]] = None
+        self.apply_ema_every_n_steps = apply_ema_every_n_steps
+        self.start_step = start_step
+        self.save_ema_weights_in_callback_state = save_ema_weights_in_callback_state
+        self.evaluate_ema_weights_instead = evaluate_ema_weights_instead
+        self.decay = decay
+
+    def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        if pl_module.device.type != "cuda":
```
titu1994, line 77: Isn't this an artificial restriction? It should work with CPU just fine ?

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+        # when loading using NeMo, ema weights will be loaded by the experiment manager separately.
+        if self._ema_model_weights is None:
+            self._ema_model_weights = state_dict.get('ema_weights')
+
+    def on_load_checkpoint(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", checkpoint: Dict[str, Any]
+    ) -> None:
+        checkpoint_callback = trainer.checkpoint_callback
+        from nemo.utils.exp_manager import NeMoModelCheckpoint
+
```
titu1994, line 123: Won't this cause circular dependency ? I see it's being used just for isinstance check, so maybe just see if NeMo in `class.__name__`

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+                    "training will start with new EMA weights.",
+                    UserWarning,
+                )
+
+    def replace_model_weights(self, pl_module: "pl.LightningModule") -> None:
+        self._weights_buffer = [p.detach().clone().to('cpu') for p in pl_module.state_dict().values()]
+        new_state_dict = {k: v for k, v in zip(pl_module.state_dict().keys(), self._ema_model_weights)}
+        pl_module.load_state_dict(new_state_dict)
+
+    def restore_model_weights(self, pl_module: "pl.LightningModule") -> None:
```
titu1994, line 155: Two methods with same name ? If you want to overload you'll need the decorator + different args list

tests/collections/common/test_ema.py
```diff
@@ -0,0 +1,400 @@
...
+            def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+                ema_callback = [x for x in trainer.callbacks if isinstance(x, EMA)][0]
+                weights = list(pl_module.state_dict().values())
+                for x, y in zip(weights, terminate_callback.pl_module_weights):
+                    assert torch.allclose(x.cpu(), y.cpu())
+                for x, y in zip(ema_callback._ema_model_weights, terminate_callback.saved_ema_weights):
+                    assert torch.allclose(x.cpu(), y.cpu())
+                assert ema_callback._cur_step == 8
+
+        trainer = Trainer(
```
titu1994, line 167: This test might be a bit expensive. Well let's measure it's runtime and if it's over 30 seconds reconsider the data set size

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,179 @@
...
+        logging.info('Creating EMA weights copy.')
+        if self._ema_model_weights is None:
+            self._ema_model_weights = [p.detach().clone() for p in pl_module.state_dict().values()]
+        # ensure that all the weights are on the correct device
+        self._ema_model_weights = [p.to(pl_module.device) for p in self._ema_model_weights]
+        self._overflow_buf = torch.IntTensor([0]).to(pl_module.device)
+
+    def apply_multi_tensor_ema(self, pl_module: "pl.LightningModule") -> None:
+        model_weights = list(pl_module.state_dict().values())
+        amp_C.multi_tensor_axpby(
```
titu1994, line 88: In this block, check if apex exists, then use current code, if not exists, switch to basic loop over the state dicts and update with pytorch inplace ops (if possible). Should be relatively fast and support both CPU and GPU (though we will never train on CPU)

titu1994 (changes requested):
Excellent work. Tests are very thorough. I'd like to remove the restriction of cpu only and apex only. 

Inside the Nemo containers apex is already installed and so it will be efficient codepath by default but for notebooks and examples run without apex it would be better to support non apex codepath

nemo/collections/common/callbacks/ema.py
SeanNaren: They are different, once is for `restore` one is for `replace`! I'm going to modify `restore` to `restore_original` to make that clearer :)

tests/collections/common/test_ema.py
SeanNaren: All tests together in this file take around 30s total, since it's all using fake data the longest thing is read/write and CUDA initialisation!

nemo/collections/common/callbacks/ema.py
SeanNaren: Which pytorch EMA page? I don't think there is one but I might be missing something...

nemo/collections/common/callbacks/ema.py
SeanNaren: Seems it runs fine as it would fail during the tests if it's broken. Do you want me to make the change regardless?

nemo/collections/common/callbacks/ema.py
```diff
@@ -0,0 +1,189 @@
...
+            [self._ema_model_weights, model_weights, self._ema_model_weights],
+            self.decay,
+            1 - self.decay,
+            -1,
+        )
+
+    def apply_ema(self, pl_module: "pl.LightningModule") -> None:
+        new_ema_weights = []
+        for orig_weight, ema_weight in zip(list(pl_module.state_dict().values()), self._ema_model_weights):
+            new_ema_weights.append(orig_weight * (1 - self.decay) + ema_weight * self.decay)
```
titu1994, line 104: This creates new memory vs pytorch inplace ops which use old checkpoint. For a moment, you will have 3 copies of the model - original, ema prior to update, and this list.

nemo/collections/common/callbacks/ema.py
titu1994: Hmm if it's just a type check yes id rather do a str check than import.

tests/collections/common/test_ema.py
titu1994: Hmm I suppose that's fine. We can't do much about cuda init.

titu1994 (commented):
Try to do inplace ops on preexisting tensor rather than alloc more memory.

titu1994 (dismissed):
Fantastic work ! Ready to merge. Fyi @okuchaiev @ericharper EMA support will now be in exp manager for all domains, please look into extensions related to model parallel.

SeanNaren:
@titu1994 just added BF16 check to ensure AMP works. As discussed offline we'd need to add separate tests/functionality for BigNLP but I think it makes sense to iterate on top of this PR! should we merge this?

titu1994:
Yes let's merge

titu1994 (approved):

PR #4777, opened 2022-08-19
Fix checkpoint restoring

nithinraok:
# What does this PR do ?

Fix restoring issue with PTL 1.7  

**Collection**: Core

# Changelog 
Move all non-zero rank value checkpoints to rank 0 checkpoints before restoring. 

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


SeanNaren (approved):
Looks good! I assume all custom strategies in NeMo (I think theres just the NLPDDPStrategy?) support broadcasting so should be fine to call the broadcast

PR #4779, opened 2022-08-19
fix handling of the final word for rnnt word timestamps

mgoldey:
# What does this PR do ?

Fixes another observed edge case for the final word in RNNT word timestamps

**Collection**: ASR

# Changelog 
- the last word in rnnt decoding for a segment may contain multiple tokens, but the data model for how these were aggregated 

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

# Additional Information
* Related to # (issue)

Example object from real use case
```python
from torch import tensor
offsets = [{'char': [tensor(456, dtype=torch.int32), tensor(1024, dtype=torch.int32)],
  'start_offset': 0,
  'end_offset': 1},
 {'char': [tensor(28, dtype=torch.int32),
   tensor(291, dtype=torch.int32),
   tensor(1024, dtype=torch.int32)],
  'start_offset': 2,
  'end_offset': 3}]
built_token = [456, 28, 291]
```

If you attempt to index offsets via `offsets[-len(built_token)]]` you get an error because `len(offsets)==2` and `len(built_token)==3`

### DIFF
--- a/nemo/collections/asr/metrics/rnnt_wer.py
+++ b/nemo/collections/asr/metrics/rnnt_wer.py
@@ -652,7 +652,7 @@ class AbstractRNNTDecoding(ABC):
                 if token != token_text:
                     # If there are any partially or fully built sub-word token ids, construct to text.
                     # Note: This is "old" subword, that occurs *after* current sub-word has started.
-                    if len(built_token) > 0:
+                    if built_token:
                         word_offsets.append(
                             {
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
Thank you so much ! I didn't expect this case, very glad it was caught !

PR #4782, opened 2022-08-19
amend rnnt word timestamps

mgoldey:
# What does this PR do ?

This PR amends the start time for the last token in RNNT decoding because the code aggregates (inclusively) since the previous_token_index all tokens in the current word. This has tripped me up in my own implementations of similar code external to NeMo in the past. My apologies for not catching it in #4779. Without these changes, some transcription invocations fail.

**Collection**: ASR

**Pre checks**:
  
**PR Type**:
- [X] Bugfix

@titu1994 I tested this code against an ASR corpus of 163,708 words, and it's looking stable with no more errors thrown. My tests were in WER rather than time stamps. There's a jitter of about 0.1%pt WER on clean English and 0.5%pts on noisy English, which is similar to what I've seen before on repeated invocations of custom models, though perhaps a bit larger for noisy audio. I don't know where to enumerate more test cases in this codebase, but this logic is a bit difficult to exhaustively test. I've seen 200-300 millisecond drifts in word timestamps across different models, so I don't know how to formalize a test for these outputs more specifically.

### DIFF
--- a/nemo/collections/asr/metrics/rnnt_wer.py
+++ b/nemo/collections/asr/metrics/rnnt_wer.py
@@ -685,10 +685,9 @@ class AbstractRNNTDecoding(ABC):
         # may have different dimensionality.
         # As such, we can't rely on the length of the list of built_token to index offsets.
         if built_token:
-            # start from the last processed token + (1 if anything was processed else 0)
-            start_offset = offsets[
-                previous_token_index + (1 if word_offsets else 0)
-            ]["start_offset"]
+            # start from the previous token index as this hasn't been committed to word_offsets yet
+            # if we still have content in built_token
+            start_offset = offsets[previous_token_index]["start_offset"]
             word_offsets.append(
                 {
                     "word": decode_tokens_to_str(built_token),


titu1994:
I faced the same issue, it's hard to formalize test for this cause each model has its own delay for token emission. None of the models were trained via fast emit either so it's somewhat expected that they will delay their predictions. It's fine for now. 

Thanks for catching all these edge cases !

titu1994 (approved):

titu1994:
Could you give an example of how you test this via wer ? Maybe we can mock it as a nightly test on LS dataset corpus

mgoldey:
Re: testing - for incremental changes, I test via transcribing a 60 second file and assert in the tests that all words exist, are time-ordered, and have non-zero duration. For major upgrades, I'm mostly looking for any transcription failures across a few thousand files with a variety of lengths and acoustic environments. These failures often look like throwing python exceptions, mostly edge cases of if statements, but sometimes this might show up as dropping 10-40% of words, so that's why I check the WER too.

PR #4829, opened 2022-08-28
adding loss normalization options to rnnt joint 

bmwshop:
# What does this PR do ?

Adds "mean" and "mean_volume" loss normalization options to RNNT Joint

**Collection**: 
ASR

# Changelog 

# Usage
In the config file, set `rnnt_reduction` to `mean`, `mean_batch` (same as the default today), or `mean_volume` in the `model` section

```python
rnnt_reduction: mean_batch
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


titu1994 (commented):
This should not go into joint but the RNNT loss should handle this.

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1356,7 +1359,15 @@ def forward(
             # Collect sub batch loss results
             if losses is not None:
                 losses = torch.cat(losses, 0)
-                losses = losses.mean()  # global batch size average
+                target_lengths = torch.cat(target_lengths, 0)
```
titu1994, line 1362: Remove all this. The RNNT loss calculates the average loss per minibatch provided to it, this mean is just to average over sub-batches.

nemo/collections/asr/losses/rnnt.py
```diff
@@ -281,19 +281,21 @@ def forward(self, log_probs, targets, input_lengths, target_lengths):
...
 
         # Loss reduction can be dynamic, so reset it after call
-        if self.reduction != 'mean_batch':
+        if self.reduction in ['mean', 'sum']:
             self._loss.reduction = 'none'
 
-        # Loss reduction only for mean_batch mode
+        # Loss reduction only for mean_batch and mean_volume modes
         if self.reduction == 'mean_batch':
             loss = torch.mean(loss)
```
sam1373, line 296: What is the difference between "mean" and "mean_batch" for this loss? In the underlying loss mean reduction seems to already be described as "the output losses will be divided by the target lengths and then the mean over the batch is taken"

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -72,7 +72,10 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         loss_name, loss_kwargs = self.extract_rnnt_loss_cfg(self.cfg.get("loss", None))
 
         self.loss = RNNTLoss(
-            num_classes=self.joint.num_classes_with_blank - 1, loss_name=loss_name, loss_kwargs=loss_kwargs
+            num_classes=self.joint.num_classes_with_blank - 1,
+            loss_name=loss_name,
+            loss_kwargs=loss_kwargs,
+            reduction=self.cfg.get("rnnt_reduction", "mean_batch"),
```
sam1373, line 78: probably should be in the loss config?

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1219,6 +1219,7 @@ def __init__(
 
         # Optional arguments
         dropout = jointnet.get('dropout', 0.0)
+        # self.reduction = jointnet.get('reduction', 'mean_batch')
```
sam1373, line 1222: remove

nemo/collections/asr/losses/rnnt.py
```diff
@@ -281,19 +281,21 @@ def forward(self, log_probs, targets, input_lengths, target_lengths):
...
+        elif self.reduction == 'mean_volume':
+            loss = torch.sum(loss) / torch.sum(target_lengths)
```
sam1373, line 298: Should the underlying loss reduction be disabled for this case (seems to be mean by default)? Also, would this work properly with the fused batches?

nemo/collections/asr/losses/rnnt.py
bmwshop: "mean_batch" - our current default - does not divide the losses by target lengths...

nemo/collections/asr/models/rnnt_models.py
bmwshop: This is meant to be similar to the implementation in the [CTC models](https://github.com/NVIDIA/NeMo/blob/d19146cd5cad07151f94820ff92392a29c3225f6/examples/asr/conf/conformer/conformer_ctc_bpe.yaml#L32)

nemo/collections/asr/losses/rnnt.py
bmwshop: To the best of my understanding, the RNNT Joint [does disable ](https://github.com/NVIDIA/NeMo/blob/61ddd23eba8d88cc08d2f2f7909b75f5333cd4d1/nemo/collections/asr/modules/rnnt.py#L1330)the underlying loss reduction so that it can do its own calculation

nemo/collections/asr/modules/rnnt.py
bmwshop: To the best of my understanding, the RNNT Joint [does disable ](https://github.com/NVIDIA/NeMo/blob/61ddd23eba8d88cc08d2f2f7909b75f5333cd4d1/nemo/collections/asr/modules/rnnt.py#L1330)the underlying loss reduction so that it can do its own calculation

okuchaiev:
is this necessary for r1.12 ?

nemo/collections/asr/modules/rnnt.py
titu1994: No it's not. It calls the RNNTLoss which calls NumbaRNNTLoss. Those two are involved for the reductions as you have implemented in this pr. The average step in the joint is only to average these subbatch losses which have already been r reduced by the loss already.
So this file needs to be reverted

titu1994: I should be clearer - the reductions we define by mean batch and mean volume utilise self.loss._reduction and not self.lpss.reduction. This arg is used only for the sum and mean types of reduction, which we remove cause it would do incorrect mean over sub-batches due to super call. 

So RNNT loss has no need to control the reductoon RNNTLoss handles it, joint just averaged losses over sub-batches

bmwshop: I just took another pass at the code, with debug statements. I see that self.loss._reduction in RNNTLossNumba gets set to none.

titu1994 (approved):
After discussion, seems there was a silent bug for all cases other than "mean_batch" inside the fused joint. This PR fixes the issue. Since the default for RNNT was mean_batch, most model training should not be affected.

nemo/collections/asr/modules/rnnt.py
titu1994: Seems to be silent bug for all cases != mean_batch. Good fix !

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1353,12 +1355,8 @@ def forward(
...
                 del sub_enc, sub_transcripts, sub_enc_lens, sub_transcript_lens
 
-            # Collect sub batch loss results
-            if losses is not None:
-                losses = torch.cat(losses, 0)
-                losses = losses.mean()  # global batch size average
-            else:
-                losses = None
+            # Reduce over sub batches
+            losses = self.loss.reduce(losses, target_lengths)
```
titu1994, line 1359: reduce only when not None. Keep the else block.

nemo/collections/asr/losses/rnnt.py
```diff
@@ -243,13 +247,32 @@ def __init__(self, num_classes, reduction: str = 'mean_batch', loss_name: str =
...
-            raise ValueError('`reduction` must be one of [mean, sum, mean_batch]')
+        if reduction not in [None, 'mean', 'sum', 'mean_batch', 'mean_volume']:
+            raise ValueError('`reduction` must be one of [mean, sum, mean_batch, mean_volume]')
 
         self._blank = num_classes
         self.reduction = reduction
         self._loss = resolve_rnnt_loss(loss_name, blank_idx=self._blank, loss_kwargs=loss_kwargs)
 
+    def reduce(self, losses, target_lengths):
+        if losses is None:
```
titu1994, line 258: Assume losses is not None, not this classes responsibility to do this check.

nemo/collections/asr/losses/rnnt.py
```diff
@@ -243,13 +247,32 @@ def __init__(self, num_classes, reduction: str = 'mean_batch', loss_name: str =
...
-        if reduction not in [None, 'mean', 'sum', 'mean_batch']:
-            raise ValueError('`reduction` must be one of [mean, sum, mean_batch]')
+        if reduction not in [None, 'mean', 'sum', 'mean_batch', 'mean_volume']:
+            raise ValueError('`reduction` must be one of [mean, sum, mean_batch, mean_volume]')
 
         self._blank = num_classes
         self.reduction = reduction
         self._loss = resolve_rnnt_loss(loss_name, blank_idx=self._blank, loss_kwargs=loss_kwargs)
 
+    def reduce(self, losses, target_lengths):
```
titu1994, line 257: Add types to this function signature

titu1994 (approved):

PR #4838, opened 2022-08-30
Incorporate API change in TE

ksivaman:
Moves the `checkpoint_core_attention` argument to the forward pass instead of init to allow more flexibility in SAR, i.e. per layer setting.


ericharper (approved):
LGTM. Thanks!

PR #4859, opened 2022-09-01
Fix mha bug

yzhang123:
# What does this PR do ?

fix bug in multi head attention forward function


nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -129,9 +129,7 @@ def forward(self, query, key, value, mask, pos_emb=None, cache=None, cache_next=
         returns:
             output (torch.Tensor): transformed `value` (batch, time1, d_model) weighted by the query dot key attention
         """
-        key, value, query, cache_next = self.update_cache(
-            key=key, value=value, query=query, cache=cache, cache_next=cache_next
-        )
+        key, value, query = self.update_cache(key=key, value=value, query=query, cache=cache, cache_next=cache_next)
```
SeanNaren, line 132: How comes no test failed due to this? do we have a test that runs this part of the code?

SeanNaren:
Good catch, but hoping there is/we add a test to catch this 😅

SeanNaren (approved):

yzhang123:
@jbalam-nv  @bmwshop  just fyi, @SeanNaren ok i will add test

PR #4860, opened 2022-09-01
Fix uppercasing mismatch for IPA heteronyms

redoctopus:

# What does this PR do ?

Fixes bug where lowercase heteronyms wouldn't match with uppercased graphemes in IPA G2P. (Graphemes were uppercased to avoid conflicts with IPA symbols, but the heteronyms list is lowercased, so they would never match.)

**Collection**: TTS

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


ekmb (approved):
Thanks!

PR #4876, opened 2022-09-05
Add sync for logging calls to ensure aggregation across devices

SeanNaren:
# What does this PR do ?

After speaking to @nithinraok a few places do not have sync calls when logging (some are taken care off by the underlying Metrics, or they are already synced across processes like global steps).

I've been a bit cautious and not touched the return objects from `validation_step` or `test_step`. There seems to be some parsing of logic in the ModelPT file so need to investigate further as to what to do with those metrics, and how they are logged.

# Changelog 
- Add `sync_dist` to logging metric calls to ensure aggregation across all devices.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?


SeanNaren:
@nithinraok I think the returning dicts that look like `return {'log'{...}}` are fine because of the logic that happens here: https://github.com/NVIDIA/NeMo/blob/main/nemo/core/classes/modelPT.py#L723

Both `asr_model.py` and `classification_model.py` implement `multi_validation_epoch_end`  and `multi_test_epoch_end` to do the aggregations before being logged. I think this is fine, it could be cleaned up slightly but I think the code looks acceptable.

nemo/collections/asr/models/label_models.py
```diff
@@ -307,15 +307,15 @@ def training_step(self, batch, batch_idx):
...
+        self.log('loss', loss, sync_dist=True)
         self.log('learning_rate', self._optimizer.param_groups[0]['lr'])
         self.log('global_step', self.trainer.global_step)
 
         self._accuracy(logits=logits, labels=labels)
         top_k = self._accuracy.compute()
         self._accuracy.reset()
         for i, top_i in enumerate(top_k):
-            self.log(f'training_batch_accuracy_top@{i}', top_i)
+            self.log(f'training_batch_accuracy_top@{i}', top_i, sync_dist=True)
```
SeanNaren, line 318:
```suggestion
            self.log(f'training_batch_accuracy_top@{i}', top_i)
```

Removing this as the metric handles aggregation.

titu1994:
Those older multi stuff needs some cleanup, agreed.

titu1994:
Be careful of adding too many syncs, they can cause deadlocks on the clusters (at least on other domains some versions before)

SeanNaren:
Thanks @titu1994! 

Yeah great point, I need to double check the warnings @nithinraok was mentioning around the calls when `sync_dist` is missing. 

A part of me thinks (atleast for training) it's worth omitting sync for training steps (since sync happens every step) to reduce communication overhead + its usually used to track overall convergence. I do think keeping the sync for validation steps is the right thing to do, but curious what people suggest!

nemo/collections/asr/models/classification_models.py
```diff
@@ -488,7 +488,7 @@ def training_step(self, batch, batch_nb):
         logits = self.forward(input_signal=audio_signal, input_signal_length=audio_signal_len)
         loss_value = self.loss(logits=logits, labels=labels)
 
-        self.log('train_loss', loss_value)
+        self.log('train_loss', loss_value, sync_dist=True)
```
SeanNaren, line 491:
```suggestion
        self.log('train_loss', loss_value)
```

nemo/collections/asr/models/classification_models.py
```diff
@@ -690,14 +690,16 @@ def training_step(self, batch, batch_idx):
...
-            'learning_rate': self._optimizer.param_groups[0]['lr'],
-        }
+        self.log_dict(
+            {
+                'train_loss': loss,
+                'train_mse': train_mse,
+                'train_mae': train_mae,
+                'learning_rate': self._optimizer.param_groups[0]['lr'],
+            },
+            sync_dist=True,
```
SeanNaren, line 700:
```suggestion
```

SeanNaren:
Dropped the training sync_dist, let's keep it for eval steps as this could vary the loss quite a bit.

nemo/collections/asr/models/classification_models.py
```diff
@@ -690,14 +690,15 @@ def training_step(self, batch, batch_idx):
...
         train_mae = self._mae(preds=logits, target=targets)
 
-        tensorboard_logs = {
-            'train_loss': loss,
-            'train_mse': train_mse,
-            'train_mae': train_mae,
-            'learning_rate': self._optimizer.param_groups[0]['lr'],
-        }
+        self.log_dict(
+            {
```
titu1994, line 694: You'll lose multi data loader support for classification models with this (GSC has val and test sets)

nithinraok:
@SeanNaren are you keeping or removing `sync_dict` in training step methods?

nemo/collections/asr/models/classification_models.py
SeanNaren: Thanks @titu1994 I might be missing something here, all I've done is assignment of the variable and combined it into the `self.log_dict` call (primarily because the idea of `tensorboard_logs` has been deprecated for a while in Lightning).

SeanNaren:
> @SeanNaren are you keeping or removing `sync_dict` in training step methods?

hey @nithinraok, I'll remove them based on trying to prevent too much communication overhead per step. 

The PossibleUserWarning shouldn't appear for the training step as the metric is logged every step instead of across the entire epoch.

nithinraok:
Thanks Sean, I see `sync_dict=True` in label_model.py training_step, does it cause communication overhead?

nithinraok (approved):

PR #4891, opened 2022-09-07
Update label_models.py

stevehuang52:
# What does this PR do ?

- Removed AUROC metric since it's slow and cost a lot of memory to store the logits&labels. 
- Fixed macro-accuracy metric usage for multiple validation manifests.
- Refactored validation and test function since they share the same code but just different tags in display ("val_" vs. "test_")

### DIFF
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -343,11 +343,11 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
         self._accuracy.correct_counts_k = correct_counts
         self._accuracy.total_counts_k = total_counts
         topk_scores = self._accuracy.compute()
-        
-        self._macro_accuracy.tp = torch.stack([x[f'{tag}_acc_stats'][0] for x in outputs]).sum(axis=0)
-        self._macro_accuracy.fp = torch.stack([x[f'{tag}_acc_stats'][1] for x in outputs]).sum(axis=0)
-        self._macro_accuracy.tn = torch.stack([x[f'{tag}_acc_stats'][2] for x in outputs]).sum(axis=0)
-        self._macro_accuracy.fn = torch.stack([x[f'{tag}_acc_stats'][3] for x in outputs]).sum(axis=0)
+
+        self._macro_accuracy.tp = torch.stack([x[f'{tag}_acc_macro_stats'][0] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.fp = torch.stack([x[f'{tag}_acc_macro_stats'][1] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.tn = torch.stack([x[f'{tag}_acc_macro_stats'][2] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.fn = torch.stack([x[f'{tag}_acc_macro_stats'][3] for x in outputs]).sum(axis=0)
         macro_accuracy_score = self._macro_accuracy.compute()
 
         self._accuracy.reset()
@@ -378,7 +378,7 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
             path2audio_file: path to audio wav file
 
         Returns:
-            embs: speaker embeddings 
+            embs: speaker embeddings
         """
         audio, sr = librosa.load(path2audio_file, sr=None)
         target_sr = self._cfg.train_ds.get('sample_rate', 16000)
@@ -408,11 +408,11 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
         Verify if two audio files are from the same speaker or not.
 
         Args:
-            path2audio_file1: path to audio wav file of speaker 1  
-            path2audio_file2: path to audio wav file of speaker 2 
+            path2audio_file1: path to audio wav file of speaker 1
+            path2audio_file2: path to audio wav file of speaker 2
             threshold: cosine similarity score used as a threshold to distinguish two embeddings (default = 0.7)
 
-        Returns:  
+        Returns:
             True if both audio files are from same speaker, False otherwise
         """
         embs1 = self.get_embedding(path2audio_file1).squeeze()
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -332,7 +332,7 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
             f'{tag}_correct_counts': correct_counts,
             f'{tag}_total_counts': total_counts,
             f'{tag}_acc_micro_top_k': acc_top_k,
-            f'{tag}_acc_macro_stats': stats
+            f'{tag}_acc_macro_stats': stats,
         }
 
     def multi_validation_epoch_end(self, outputs, dataloader_idx: int = 0, tag: str = 'val'):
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/models/label_models.py
```diff
@@ -319,97 +318,56 @@ def training_step(self, batch, batch_idx):
...
+        total_counts = torch.stack([x[f'{tag}_total_counts'] for x in outputs]).sum(axis=0)
 
         self._accuracy.correct_counts_k = correct_counts
         self._accuracy.total_counts_k = total_counts
         topk_scores = self._accuracy.compute()
+        
+        self._macro_accuracy.tp = torch.stack([x[f'{tag}_acc_stats'][0] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.fp = torch.stack([x[f'{tag}_acc_stats'][1] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.tn = torch.stack([x[f'{tag}_acc_stats'][2] for x in outputs]).sum(axis=0)
+        self._macro_accuracy.fn = torch.stack([x[f'{tag}_acc_stats'][3] for x in outputs]).sum(axis=0)
```
nithinraok, line 350: is this val_acc_stats or val_acc_macro_stats?

stevehuang52: Thanks for pointing out, I've fixed that~

nemo/collections/asr/models/label_models.py
```diff
@@ -319,97 +318,56 @@ def training_step(self, batch, batch_idx):
...
-        self.log('val_acc_macro', macro_accuracy_score, sync_dist=True)
-        self.log('val_auroc', auroc_score, sync_dist=True)
-
-        return {
-            'val_loss': val_loss_mean,
-            'val_acc_top_k': topk_scores,
-            'val_acc_macro': macro_accuracy_score,
-            'val_auroc': auroc_score,
-        }
+        return {'log': tensorboard_log}
```
nithinraok, line 364: also remove this tensorboard_logs, refer to https://github.com/NVIDIA/NeMo/pull/4876

stevehuang52: Sure, just fixed.

nemo/collections/asr/models/label_models.py
```diff
@@ -319,97 +318,58 @@ def training_step(self, batch, batch_idx):
...
-        self.log('test_acc_macro', macro_accuracy_score, sync_dist=True)
-        self.log('test_auroc', auroc_score, sync_dist=True)
-
-        return {
-            'test_loss': test_loss_mean,
-            'test_acc_top_k': topk_scores,
-            'test_acc_macro': macro_accuracy_score,
-            'test_auroc': auroc_score,
-        }
+        return self.multi_validation_epoch_end(outputs, dataloader_idx, 'test')
```
fayejf, line 372: Not sure if it's a good idea to merge val and test here

fayejf: what do you think? @nithinraok

nithinraok: we can merge, why do you think its not a good idea?

fayejf: @nithinraok  just worried it would break things such eval on multi gpu and reduce flexibility, but I guess I think too much.  Then let's merge.

nithinraok: I dont think since its re-use of function, but @fayejf or @stevehuang52 , can you run an experiment and see this working as expected before merging? We have a unit test for training but do not check validation and test time.

stevehuang52: I can do it, but do you have any examples of how to do it quickly? I think it will work correctly, and ASR models are also doing it the same way: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/models/ctc_models.py#L650

nithinraok: @fayejf can you provide steve with some lang id examples. Notice the validation time before and after the change.

fayejf: let me run it really quick.

nithinraok:
@stevehuang52 since you are removing auroc, remove it from instantiation and also corresponding import

stevehuang52:
> @stevehuang52 since you are removing auroc, remove it from instantiation and also corresponding import

It's already removed

nemo/collections/asr/models/label_models.py
SeanNaren: What might be cleaner is to separate out into a function if we're worried about using the validation hook within the test hook:

```python
def calculate_metrics(self, outputs, dataloader_idx, tag):
    ...

def multi_test_epoch_end(self, outputs, dataloader_idx):
   return self.calculate_metrics(outputs, dataloader_idx, 'test')

def multi_validation_epoch_end(self, outputs, dataloader_idx):
   return self.calculate_metrics(outputs, dataloader_idx, 'val')

```

nithinraok: yes, this would be neat.

fayejf:
@nithinraok @stevehuang52 I've tested with a toy sample for validation time, but probably my sample is not large enough, I didn't observe time reduction with this fix than main.

nemo/collections/asr/models/label_models.py
```diff
@@ -319,97 +318,58 @@ def training_step(self, batch, batch_idx):
...
         macro_accuracy_score = self._macro_accuracy.compute()
-        auroc_score = self._auroc.compute()
 
         self._accuracy.reset()
         self._macro_accuracy.reset()
-        self._auroc.reset()
 
-        logging.info("val_loss: {:.3f}".format(val_loss_mean))
-        self.log('val_loss', val_loss_mean)
+        logging.info(f'{tag}_loss: {loss_mean:.3f}')
```
fayejf, line 356: could you remove the duplicate line here?

fayejf (changes requested):
Mostly good. could you update with sean's suggestion?

nithinraok:
Did you also test it with r1.10 to compare, can you please do that as well, to find the cause of the delay in the validation time issue?

nemo/collections/asr/models/label_models.py
stevehuang52: will do

fayejf:
> Did you also test it with r1.10 to compare, can you please do that as well, to find the cause of the delay in the validation time issue? @fayejf

I don't think we add auroc and val_macro_acc in r1.10, right?

stevehuang52:
> Mostly good. could you update with sean's suggestion?

Done~

fayejf:
> > > Did you also test it with r1.10 to compare, can you please do that as well, to find the cause of the delay in the validation time issue? @fayejf
> > 
> > 
> > I don't think we add auroc and val_macro_acc in r1.10, right?
> 
> yes, hence want to compare with it and test

@nithinraok So I compared this PR with main and r1.10 (val acc since val acc macro is not implemented) and I didn't reinstall anything, just switched branch.

No much diff in terms of time actually.

per epoch for first 5 epochs. 
r1.10 - total: 12.42-12.44 val: 02:04
main  - total: 12.43-12.45 val : 02.05-02.07
[steve](https://github.com/NVIDIA/NeMo/pull/4891/commits/70a4337568c6c99d32e15bb0c22010787e5be641) - total: 12.43-16.36 val : 02.05-02.05

I do observe some weird 16:36 run above and let me try it again. Besides that, no time diff/reduction in terms of the fix. Possibly because our val data is not huge enough. But I think it's okay to merge it to fix and remove auroc for now.

nithinraok:
> @nithinraok So I compared this PR with main and r1.10 (val acc since val acc macro is not implemented) and I didn't reinstall anything, just switched branch.

I think it didn't run what you might be intended, since r1.10 runs with PTL 1.6, and throws errors with PTL 1.7. So, you might need to reinstall for each test.

fayejf:
> > @nithinraok So I compared this PR with main and r1.10 (val acc since val acc macro is not implemented) and I didn't reinstall anything, just switched branch.
> 
> I think it didn't run what you might be intended, since r1.10 runs with PTL 1.6, and throws errors with PTL 1.7. So, you might need to reinstall for each test.

what kind of error do we expect to see for 1.10 with PTL1.7? I'm with PTL 1.7.3 and r1.10 and it works fine.

nithinraok:
> what kind of error do we expect to see for 1.10 with PTL1.7? I'm with PTL 1.7.3 and r1.10 and it works fine.
It would be from lightningmodule.trainer attribute, But I think you ran correctly I guess, vice versa shouldn't work (latest and PTL 1.6).

fayejf:
> > what kind of error do we expect to see for 1.10 with PTL1.7? I'm with PTL 1.7.3 and r1.10 and it works fine.
> > It would be from lightningmodule.trainer attribute, But I think you ran correctly I guess, vice versa shouldn't work (latest and PTL 1.6).

@nithinraok I run with PTL 1.7.3 for all of the above tests. Though I think I should run r1.10 with PTL 1.6 but the run was fine with 1.7.3

nithinraok (approved):
LGTM, thanks. Some more runs need to be done to find the cause of the increase in validation time.

fayejf:
> LGTM, thanks. Some more runs need to be done to find the cause of the increase in validation time.

@nithinraok I didn't see an increase in validation time, but I didn't see the expected reduction of this commit actually.

fayejf (approved):
LGTM thanks! i found using weighted cross entropy loss in test in `speech_to_label.py` is problematic. let me patch in another PR.

PR #4902, opened 2022-09-08
[TTS] fix wrong g2p path.

XuesongYang:

# What does this PR do ?

fixed wrong g2p path. Previously in https://github.com/NVIDIA/NeMo/pull/4690, g2p modules were moved to `nemo_text_processing.g2p.modules`. 

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/tts/torch/g2ps.py
+++ b/nemo/collections/tts/torch/g2ps.py
@@ -11,4 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# TODO (xueyang): deprecate this file since no other places import modules from here anymore. However,
+#  all checkpoints uploaded in ngc used this path. So it requires to update all ngc checkpoints g2p path as well.
 from nemo_text_processing.g2p.modules import IPAG2P, BaseG2p, EnglishG2p
--- a/nemo/collections/tts/models/fastpitch.py
+++ b/nemo/collections/tts/models/fastpitch.py
@@ -46,7 +46,7 @@ from nemo.utils import logging, model_utils
 
 @dataclass
 class G2PConfig:
-    _target_: str = "nemo.collections.common.tokenizers.text_to_speech.g2ps.EnglishG2p"
+    _target_: str = "nemo_text_processing.g2p.modules.EnglishG2p"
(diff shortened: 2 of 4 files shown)


redoctopus (approved):
Thanks for the fix!

nemo/collections/tts/torch/g2ps.py
```diff
@@ -11,4 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+# TODO (xueyang): deprecate this file since no other places import modules from here anymore. However,
+#  all checkpoints uploaded in ngc used this path. So it requires to update all ngc checkpoints g2p path as well.
```
XuesongYang, line 16: @redoctopus @ekmb what is the best way to deprecate `g2ps.py` and `tts_tokenizers.py` as described in the `TODO`?

redoctopus: We should make a list of models that are affected and that we'll need to replace on NGC. At that point we can figure out how long it will take to retrain everything and set a release for deprecation.

XuesongYang: yes, we need to define a strategy to do this and beyond this. model releases usually tie to the latest versioning of tooling. We may need to update models already in NGC in a monthly cadence aligning with NeMo release cadence. Otherwise, NGC models would fall apart because NeMo is moving so fast.

rlangman: Would it be possible to update the model loading logic itself to alias the old paths to the new paths? Rather than physically update the paths in the checkpoints themselves.

XuesongYang: That's also a legit workaround, but I guess it is not easy to maintain by humans. Retraining would just cost computer time rather than human time.

redoctopus: If we want to change the model updating or checkpoint load/restore setup, that may be a good topic for a sync.

A module move of this scale tends to be relatively rare, and models don't tend to break frequently, so a monthly update schedule might lead to some redundant work. But it might be good to have a process for updating models automatically anyway, for when it does happen.

PR #4904, opened 2022-09-08
Add mixed-representation config to PhonemizerTokenizer

rlangman:

# What does this PR do ?

Extends the IPATokenizer and PhonemizerTokenizer classes so we can do mixed-representation grapheme/phoneme training for non-English languages.

**Collection**: [TTS]

# Changelog 
- Add phoneme_probability argument to PhonemizerTokenizer, with code changes to add the grapheme set for each language. The grapheme set was obtained by taking all graphemes in the words of each language's pronunciation dictionary.
- Removed "use_chars" logic in IPATokenizer, which currently adds hardcoded list of English ASCII chars, in favor of parsing them from the user-provided phoneme dictionary in IPAG2P (similar to how it infers its phoneme set from the dictionary).
- In IPAG2P I added accented characters & locale-agnostic digit to the word/token checker. In an earlier commit, I accidentally added this to the EnglishG2p instead (when refactoring the PR around the tokenizer file directory change). Though I don't see any reason to remove it from EnglishG2p.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

# Additional Information

### DIFF
--- a/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
@@ -20,8 +20,27 @@ DEFAULT_PUNCTUATION = (
     ')', '[', ']', '{', '}',
 )
 
+GRAPHEME_CHARACTER_SETS = {
+    "en-us": (
+        'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
+        'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T',
+        'U', 'V', 'W', 'X', 'Y', 'Z', 'À', 'É'
+    ),
+    "es": (
+        'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
+        'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T',
+        'U', 'V', 'W', 'X', 'Y', 'Z', 'Á', 'É', 'Í', 'Ñ',
+        'Ó', 'Ú', 'Ü'
+    ),
+    "de": (
+        'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
+        'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T',
+        'U', 'V', 'W', 'X', 'Y', 'Z', 'Ä', 'É', 'Ê', 'Ñ',
+        'Ö', 'Ü'
+    ),
+}
 
-CHARACTER_SETS = {
+IPA_CHARACTER_SETS = {
     "en-us": (
         'a', 'b', 'd', 'e', 'f', 'h', 'i', 'j', 'k', 'l',
         'm', 'n', 'o', 'p', 'r', 's', 't', 'u', 'v', 'w',
@@ -46,11 +65,18 @@ CHARACTER_SETS = {
 # fmt: on
 
 
-def get_ipa_character_list(language):
-    if language not in CHARACTER_SETS:
-        raise ValueError(f"Character set not found for language {language}")
-    char_list = list(CHARACTER_SETS[language])
-    return char_list
+def get_grapheme_character_set(language):
+    if language not in GRAPHEME_CHARACTER_SETS:
+        raise ValueError(f"Grapheme character set not found for language {language}")
+    char_set = set(GRAPHEME_CHARACTER_SETS[language])
+    return char_set
+
+
+def get_ipa_character_set(language):
+    if language not in IPA_CHARACTER_SETS:
+        raise ValueError(f"IPA character set not found for language {language}")
+    char_set = set(IPA_CHARACTER_SETS[language])
+    return char_set
 
 
 def get_ipa_punctuation_list(language):
--- a/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
@@ -21,18 +21,18 @@ DEFAULT_PUNCTUATION = (
 )
 
 GRAPHEME_CHARACTER_SETS = {
-    "en-us": (
+    "en-US": (
         'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
         'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T',
-        'U', 'V', 'W', 'X', 'Y', 'Z', 'À', 'É'
+        'U', 'V', 'W', 'X', 'Y', 'Z'
     ),
-    "es": (
+    "es-ES": (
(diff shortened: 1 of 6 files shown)


nemo_text_processing/g2p/modules.py
```diff
@@ -254,6 +254,7 @@ def __init__(
         apply_to_oov_word=None,
         ignore_ambiguous_words=True,
         heteronyms=None,
+        use_chars=False,
```
redoctopus, line 257: Could you update the IPA FastPitch config to match these new changes?

nemo_text_processing/g2p/modules.py
```diff
@@ -286,13 +289,20 @@ def __init__(
         """
         self.use_stresses = use_stresses
         self.set_graphemes_upper = set_graphemes_upper
+        self.use_chars = use_chars
+        self.phoneme_probability = phoneme_probability
+        self._rng = random.Random()
+
+        if not self.use_chars and self.phoneme_probability is not None:
+            logging.warning("use_chars was not set while using phoneme_probability.")
```
redoctopus, line 297: The old default was to override and add chars in the conflicting; if we don't want that behavior I think this should error out instead.

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
```diff
@@ -20,8 +20,27 @@
     ')', '[', ']', '{', '}',
 )
 
+GRAPHEME_CHARACTER_SETS = {
+    "en-us": (
+        'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
+        'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T',
+        'U', 'V', 'W', 'X', 'Y', 'Z', 'À', 'É'
+    ),
+    "es": (
```
XuesongYang, line 29: nit: since we are expanding languages and dialects, shall we start now using standardized locale code (ISO 639-3 for the language name plus ISO 3166 for the country code)? for example, `es-ES`, `de-DE`, `en-US`?

nemo_text_processing/g2p/modules.py
```diff
@@ -371,6 +379,8 @@ def _parse_as_cmu_dict(phoneme_dict_path, use_stresses=False, stress_symbols=[],
                         for stress_symbol in stress_symbols:
                             pronunciation = pronunciation.replace(stress_symbol, '')
                     symbols.update(pronunciation)  # This will insert each char individually
+                    if use_chars:
+                        symbols.update(word)
```
redoctopus, line 383: Hmm, this may add some punctuation in the dictionary. It probably won't mess anything up, but I think it'll override the punctuation list set by the tokenizer, which may lead to some unexpected behavior (might mess with reserved symbols?).

Not sure what the best way of handling this is that's still generic across languages.

redoctopus: Maybe we just strip all `string.punctuation` from the symbol set at the end of dictionary loading... But then that doesn't catch other languages' punctuation marks :T

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
rlangman: Using ISO does make sense. I chose these language tags because they are what eSpeak recognizes. But it makes more sense to use the official language tags, and manually map them to the eSpeak ones as needed.

ekmb: @rlangman, what's the motivation behind adding a few accented chars, like `À` into en-US set?

nemo_text_processing/g2p/modules.py
rlangman: The interface is a bit confusing, but I cannot think of a downside to using the old default and automatically adding the graphemes from the dictionary when the probability is set. Will change.

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
rlangman: In English there are a lot of words adopted from other languages with accented characters. I got this list from parsing the ipa_cmudict on main. However looking at it closer, there is only a single accented word (DÉJÀ, as in "déjà vu") and all other words which should be accented (eg. naïve, café) had their accents stripped before we included them in our dictionary. Most likely the CMU Arpabet dictionary did not have them.

For IPA I don't think there is any reason to remove these characters from English since we definitely need to support them for other languages. But supporting them requires us to add such words to our English dictionary (ideally we would include both the accented and non-accented versions).

In the meantime, I should probably either remove these 2 characters or include other accented characters too.

Happy to here anyone else's thoughts on the matter.

nemo_text_processing/g2p/modules.py
rlangman: It would be safer to strip with the regex r"[a-zA-ZÀ-ÿ\d]".

Currently when we have contractions (eg. don't) and compound phrases (eg. long-term) do we process the words as-is or do we process them after stripping the punctuation?

redoctopus: We may have to revisit this if/when we add non-Latin languages, depending on how they're going to be represented.

I'd say process words as-is in the dictionary (i.e. what the original behavior was) and only strip them out of the symbol set for the G2P class. Some of the punctuation may be valid and will be added back in by the tokenizer, e.g. apostrophes, so we want the dictionary lookup to still match.

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
rlangman: For now I decided to remove those 2 accented chars and leave the option of coming up with a more complete list of chars/words for a different PR.

nemo_text_processing/g2p/modules.py
rlangman: Done

redoctopus (approved):
LGTM, thanks!

redoctopus:
Hmm, I'm not familiar with the `Overly permissive regular expression range` LGTM alert... Can you check that the regex works as expected?

nemo_text_processing/g2p/modules.py
```diff
@@ -286,32 +293,40 @@ def __init__(
...
 
             self.symbols = set()
-            for entries in phoneme_dict.values():
-                for phonemes in entries:
-                    self.symbols.update(phonemes)
+            for word, phonemes in phoneme_dict.items():
+                self.symbols.update(phonemes)
+                if self.use_chars:
+                    word_no_punct = self.PUNCT_REGEX.sub('', word)
+                    self.symbols.update(word_no_punct)
```
blisc, line 321: This doesn't seem ideal to me. Looking through our current ipa dictionary we have entries like: `G6` and `DÉJÀ`. This results in errors with backwards-compatibility since we previously assumed that numbers [0-9] and loan characters (from French) are not consider English graphemes. We should either re-evaluate this logic or not process these entries when looping through the g2p dict.

PR #4905, opened 2022-09-08
Fix saving the last checkpoint when using val check interval

SeanNaren:
# What does this PR do ?

When saving a checkpoint and the validation check interval isn't divisible by the number of batches, we can miss saving the final checkpoint. This ensures we save the final checkpoint.

I've noticed that the epoch saved in the checkpoint is incremented. Not sure if this will be an issue, but considering this is the final step I assume it should be fine.

cc @ericharper @okuchaiev

# Changelog 
- Fix saving the last checkpoint when `val_check_interval`.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

### DIFF
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -810,6 +810,16 @@ class NeMoModelCheckpoint(ModelCheckpoint):
         if trainer.fast_dev_run:
             return None
 
+        # check if we need to save a last checkpoint manually as validation isn't always run based on the interval
+        if self.save_last and trainer.val_check_interval != 0:
+            should_save_last_checkpoint = False
+            if isinstance(trainer.val_check_interval, float) and trainer.val_check_interval % trainer.global_step != 0:
+                should_save_last_checkpoint = True
+            if isinstance(trainer.val_check_interval, int) and trainer.global_step % trainer.val_check_interval != 0:
+                should_save_last_checkpoint = True
+            if should_save_last_checkpoint:
+                monitor_candidates = self._monitor_candidates(trainer)
+                super()._save_last_checkpoint(trainer, monitor_candidates)
         # Call parent on_train_end() to save the -last checkpoint
         super().on_train_end(trainer, pl_module)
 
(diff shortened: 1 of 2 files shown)


okuchaiev (approved):
lgtm. Thx!

yaoyu-33:
@SeanNaren with the change, 
1. the code saves a checkpoint `'megatron_gpt--val_loss=0.00-step=102510-consumed_samples=0-last.ckpt'` with not reasonable `val_loss` and `consumed_samples`, and the next run cannot resume from that. 
2. Also `102510` is the last step from previous run, so it didn't progress anything during last resume, i.e. the steps didn't increase.

SeanNaren:
Thanks @yaoyu-33 so when restarting from the saved checkpoint we do not continue training as expected? Realised the test does not check this, will extend the test.

Regarding the 0 val loss it might be due to this is saving outside of the validation loop. Curious if this is the reason why the loss is 0

MaximumEntropy:
cc: @ericharper who has investigated this a fair bit. 

@SeanNaren for some reason, when resuming training, PTL wants to run a full validation epoch but without any validation batches, so in `validation_epoch_end()` `outputs` is an empty list or an empty list of lists. So we have a condition like here https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py#L646 to just return if outputs is empty. However, this still triggers the checkpointing callback and we save a checkpoint without any logged validation metrics hence it saves a checkpoint like @yaoyu-33 reported with val loss 0 and consumed samples 0. If you know how to disable the validation epoch loop from running on resume, that would be super useful!

invisiblepancake (commented):
Ah, me hearty, grab a pint and let's chat about this commit, eh?

So, ye've gone and tinkered with the ptl_load_state function, eh? Swapping it to use cls in a right few spots. Aye, that's the ticket!

Line 353: Ye've added cls to the ptl_load_state call. Fair play to ya, makes the code cleaner than a freshly poured stout.
Line 355: Same gig here, adding cls to the ptl_load_state call with cfg. Smart move, lad, like adding an extra shot of whiskey to the mix.
Overall, ye've done a grand job. This wee change should keep things shipshape and Bristol fashion. Now, let's raise a toast to bug-free code!

Sláinte! 🍻

PR #4923, opened 2022-09-12
Fix implicit bug in _AudioLabelDataset

stevehuang52:
Add type checking before calling .split(). The old code will break if the input is a list of strings, which happens when using multiple manifests.

### DIFF
--- a/nemo/collections/asr/data/audio_to_label.py
+++ b/nemo/collections/asr/data/audio_to_label.py
@@ -214,7 +214,7 @@ target_label_0, "offset": offset_in_sec_0}
         {"audio_filepath": "/path/to/audio_wav_n.wav", "duration": time_in_sec_n, "label": \
 target_label_n, "offset": offset_in_sec_n}
     Args:
-        manifest_filepath (str): Dataset parameter. Path to JSON containing data.
+        manifest_filepath (Union[str, List[str]]): Dataset parameter. Path to JSON containing data.
         labels (list): Dataset parameter. List of target classes that can be output by the speaker recognition model.
         featurizer
         min_duration (float): Dataset parameter. All training files which have a duration less than min_duration
@@ -261,7 +261,7 @@ target_label_n, "offset": offset_in_sec_n}
     def __init__(
         self,
         *,
-        manifest_filepath: str,
+        manifest_filepath: Union[str, List[str]],
         labels: List[str],
         featurizer,
         min_duration: Optional[float] = 0.1,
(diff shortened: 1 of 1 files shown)


nithinraok:
ASRSpeechLabel expects manifest_files to be iterable list, this is not a bug.

stevehuang52:
> ASRSpeechLabel expects manifest_files to be iterable list, this is not a bug.

But list does not have the .split() function, isn't that a bug? the code will work if manifest_filepath is a string, but sometimes it could be a list of strings, when we use "[train1.json,train2.json]" in config

nithinraok:
If there are multiple manifests, it's still string and are separated by `,` but need not be list.

nithinraok:
But this can be merged. Can you add a test for this? Using the train_ds with multiple manifest files, with list and str and single manifest as well

stevehuang52:
> If there are multiple manifests, it's still string and are separated by `,` but need not be list.

If in config we set train_ds.manifest_filepath="[train1.json,train2.json]" , then in the dataset the manifest_filepath=['train1.json', 'train2.json], which is  a list not a string, so we still need to check if the input is a string before we call split(). Since the proposed fix does not change the behavior of the original code, I believe it will work as expected.

nithinraok:
Yes, it will work expected for prior cases, but can you add a test case, is that possible?

nithinraok (approved):
LGTM

titu1994:
@nithinraok, don't merge if the test wasn't added yet. This affects more than one domain - speech classification, vad and speaker tasks. It needs to be treated more carefully.

nithinraok:
@stevehuang52 can you add relevant test?

stevehuang52:
I will add later, but honestly I don't think that will break anything. It's just to check if the manifest_filepath is string before calling split(), and I have tested the change with my local branch....We can revert the merge if you think it's unsafe.

stevehuang52:
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/data/audio_to_text.py#L267 

Here in ASR we have something almost the same:

```python
      if type(manifest_filepath) == str:
          manifest_filepath = manifest_filepath.split(",")

      self.manifest_processor = ASRManifestProcessor(
          manifest_filepath=manifest_filepath,
          parser=parser,
          max_duration=max_duration,
          min_duration=min_duration,
          max_utts=max_utts,
          bos_id=bos_id,
          eos_id=eos_id,
          pad_id=pad_id,
      )
```

PR #4936, opened 2022-09-14
Update megatron interface to dialogue

Zhilin123:
# What does this PR do ?

1. Increase speed of rank candidates
2. Include support for p tuning for Megatron GPT
3. Include support for TP> 1 model (i.e. 5B Megatron GPT )
4. Update generate_candidate interface for Megatron GPT based on unified text generation method
5. Included custom collate_fn to remove excess padding in batch for Dialogue GPT Classification Dataset

**Collection**: nlp

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/nlp/dialogue/dialogue.py
```diff
@@ -121,8 +124,9 @@ def main(cfg: DictConfig) -> None:
 
     if cfg.do_training:
         trainer.fit(model)
-        if cfg.model.nemo_path:
+        if cfg.model.nemo_path and not os.path.exists(cfg.model.nemo_path):
```
vadam5, line 127: If a new model just finished training, and has the same name as an existing model, instead of just not saving the new model, shouldn't a warning be raised and the new model saved with a slightly different name? e.g. with (1) appended to the end or something?

nemo/collections/nlp/data/dialogue/dataset/dialogue_gpt_classification_dataset.py
```diff
@@ -201,6 +201,30 @@ def format_prompt(self, utterance, few_shot=0, idx=None):
...
+        batch_max_length = 0
+        for candidate_attn_mask in candidate_attn_masks:
+            for one_attn_mask in candidate_attn_mask:
+                batch_max_length = max(batch_max_length, torch.sum(one_attn_mask).item())
+        # padding for tp=2 situation
+        if batch_max_length % 2:
+            batch_max_length += 1
+
+        all_items = []
+        for item in zip(*batch):
```
vadam5, line 216: Can you add some comments stating what's happening with the data when each of these if statements are true?

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -224,47 +267,72 @@ def predict_step(self, batch, batch_idx, dataloader_idx=None):
...
+        prompt_token_labels = prompt_token_labels.to(input_ids.device)
+
+        return prompt_token_labels
+
+    def get_prompt_ids_for_megatron_gpt(self, input_ids):
+        if self.cfg.virtual_prompt_style == 'prompt-tuning' or not self.prompt_learning or self.trainer.testing:
+            prompt_ids = torch.tensor([0] * input_ids.size(0)).to(input_ids.device) if self.prompt_learning else None
+        else:
+            total_virtual_tokens = self.cfg.task_templates[0].total_virtual_tokens
+            init_text = self.cfg.prompt_tuning.new_prompt_init_text[0]
```
vadam5, line 296: Why do you use `new_prompt_init_text` here? You should be using `self.cfg.task_templates[0].taskname` right?

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -224,47 +267,72 @@ def predict_step(self, batch, batch_idx, dataloader_idx=None):
...
+        if self.prompt_learning:
+            prompt_token_labels.data = torch.LongTensor(
+                np.tile(np.array(self.language_model.pseudo_token_ids), (input_ids.size(0), 1))
+            )
+
+        prompt_token_labels = prompt_token_labels.to(input_ids.device)
+
+        return prompt_token_labels
+
+    def get_prompt_ids_for_megatron_gpt(self, input_ids):
```
vadam5, line 291: Can this method name be changed to `get_virtual_prompt_ids_for_megatron_gpt`? You have a method called `get_prompt_token_labels_for_megatron_gpt` where `prompt_token` seems to refer to all prompt tokens, whether they are virtual or discrete prompt tokens. I think making the distinction that this method is just for virtual prompt tokens would add clarity.

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -224,47 +267,72 @@ def predict_step(self, batch, batch_idx, dataloader_idx=None):
...
 
         if self.cfg.library == "huggingface":
             output = self.language_model(input_ids=input_ids, attention_mask=attention_mask, labels=labels)
             loss = output['loss']
 
+            b_logits = output['logits']
+            # Shift so that tokens < n predict n
+            shift_logits = b_logits[..., :-1, :].contiguous()
+            shift_labels = labels[..., 1:].contiguous()
+            loss_fct = torch.nn.CrossEntropyLoss(reduction='none')
```
vadam5, line 312: Why do you reinitialize the loss function during every forward call? The loss function should probably be a class attribute that gets initialized once.

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -276,7 +344,7 @@ def forward(self, input_ids, attention_mask, labels, inference=True):
                     attn_mask,
                     labels=left_shifted_input_ids,
                     taskname_ids=prompt_ids,
-                    inference=inference,
+                    inference=False if len(prompt_ids.size()) == 2 else True,
```
vadam5, line 347: I think its better to have an inference flag that gets set in a way that clearly shows why inference is true or false. In this check, it is not immediately clear why `len(prompt_ids.size()) == 2 ` being true or false should indicate that we are doing inference.

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -294,7 +362,16 @@ def forward(self, input_ids, attention_mask, labels, inference=True):
             labels_mask = labels_mask_0 > 0
 
             loss = self.mask_and_reduce_loss(labels_mask, unmasked_unreduced_loss)
-        return loss
+            loss_per_sample = self.mask_and_reduce_loss_per_sample(labels_mask, unmasked_unreduced_loss)
+
+        return loss, loss_per_sample
+
+    def mask_and_reduce_loss_per_sample(self, loss_mask, output_tensor):
```
vadam5, line 369: How is this different from `mask_and_reduce_loss`?

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
```diff
@@ -612,8 +677,11 @@ def prepare_data(self):
 
     def setup(self, stage=None):
         super().setup()
-        if self.cfg.library == "megatron" and self.prompt_learning:
-            self.language_model.init_new_prompts()
+        if self.cfg.library == "megatron" and self.prompt_learning and stage == "fit":
+            if self.cfg.virtual_prompt_style == 'prompt-tuning':
```
vadam5, line 681: Can you use the enum check here instead of comparing strings? Like how we do here: https://github.com/NVIDIA/NeMo/blob/42c7a671dea3ef767ddeaf6891861ff830106c42/nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py#L179

Same comment for the if statement in line 292

examples/nlp/dialogue/dialogue.py
Zhilin123: Sure makes sense!

nemo/collections/nlp/data/dialogue/dataset/dialogue_gpt_classification_dataset.py
Zhilin123: Sure

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: sure!

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: Makes sense!

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: Good catch!

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: It is for `Mask and reduce loss based on each sample in batch. Useful for ranking candidates with the same prompt in batch based on loss` Also address as a comment in code

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: Thanks for catching this

nemo/collections/nlp/models/dialogue/dialogue_gpt_classification_model.py
Zhilin123: Sure, I agree it would be clearer that way

vadam5 (approved):
LGTM

PR #4942, opened 2022-09-17
[TTS] Condition All Components of FastPitch on Speaker Embeddings

subhankar-ghosh:
# What does this PR do ?

This PR does the following:
1) Adds capability to condition decoder, pitch and duration predictor of FastPitch and Aligner on speaker embeddings.
2) Adds capability to condition aligner with any kind of embedding.

nemo/tts/collection

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/tts/modules/aligner.py
+++ b/nemo/collections/tts/modules/aligner.py
@@ -143,7 +143,7 @@ class AlignmentEncoder(torch.nn.Module):
 
         return dist_sum / total_frames
 
-    def forward(self, queries, keys, mask=None, attn_prior=None):
+    def forward(self, queries, keys, mask=None, attn_prior=None, conditioning=None):
         """Forward pass of the aligner encoder.
 
         Args:
@@ -151,10 +151,13 @@ class AlignmentEncoder(torch.nn.Module):
             keys (torch.tensor): B x C2 x T2 tensor (text data).
             mask (torch.tensor): B x T2 x 1 tensor, binary mask for variable length entries (True = mask element, False = leave unchanged).
             attn_prior (torch.tensor): prior for attention matrix.
+            conditioning (torch.tensor): B x T2 x 1 conditioning embedding
         Output:
             attn (torch.tensor): B x 1 x T1 x T2 attention mask. Final dim T2 should sum to 1.
             attn_logprob (torch.tensor): B x 1 x T1 x T2 log-prob attention mask.
         """
+        if conditioning is not None:
+            keys = keys + conditioning.transpose(1,2)
         keys_enc = self.key_proj(keys)  # B x n_attn_dims x T2
         queries_enc = self.query_proj(queries)  # B x n_attn_dims x T1
 
--- a/nemo/collections/tts/modules/aligner.py
+++ b/nemo/collections/tts/modules/aligner.py
@@ -157,7 +157,7 @@ class AlignmentEncoder(torch.nn.Module):
             attn_logprob (torch.tensor): B x 1 x T1 x T2 log-prob attention mask.
         """
         if conditioning is not None:
-            keys = keys + conditioning.transpose(1,2)
+            keys = keys + conditioning.transpose(1, 2)
         keys_enc = self.key_proj(keys)  # B x n_attn_dims x T2
         queries_enc = self.query_proj(queries)  # B x n_attn_dims x T1
 
--- a/nemo/collections/tts/modules/fastpitch.py
+++ b/nemo/collections/tts/modules/fastpitch.py
@@ -231,7 +231,7 @@ class FastPitchModule(NeuralModule):
         # Input FFT
         enc_out, enc_mask = self.encoder(input=text, conditioning=spk_emb)
 
-        log_durs_predicted = self.duration_predictor(enc_out, enc_mask)
+        log_durs_predicted = self.duration_predictor(enc_out + spk_emb, enc_mask)
         durs_predicted = torch.clamp(torch.exp(log_durs_predicted) - 1, 0, self.max_token_duration)
 
         attn_soft, attn_hard, attn_hard_dur, attn_logprob = None, None, None, None
@@ -242,7 +242,7 @@ class FastPitchModule(NeuralModule):
             attn_hard_dur = attn_hard.sum(2)[:, 0, :]
 
         # Predict pitch
-        pitch_predicted = self.pitch_predictor(enc_out, enc_mask)
+        pitch_predicted = self.pitch_predictor(enc_out + spk_emb, enc_mask)
         if pitch is not None:
             if self.learn_alignment and pitch.shape[-1] != pitch_predicted.shape[-1]:
                 # Pitch during training is per spectrogram frame, but during inference, it should be per character
@@ -262,7 +262,7 @@ class FastPitchModule(NeuralModule):
             len_regulated, dec_lens = regulate_len(durs_predicted, enc_out, pace)
 
         # Output FFT
-        dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens)
+        dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens, conditioning=spk_emb)
         spect = self.proj(dec_out).transpose(1, 2)
         return (
             spect,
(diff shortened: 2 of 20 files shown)


nemo/collections/tts/modules/fastpitch.py
```diff
@@ -230,19 +236,27 @@ def forward(
 
         # Input FFT
         enc_out, enc_mask = self.encoder(input=text, conditioning=spk_emb)
-
-        log_durs_predicted = self.duration_predictor(enc_out, enc_mask)
+        if self.speaker_emb_condition_prosody:
+            log_durs_predicted = self.duration_predictor(enc_out + spk_emb, enc_mask)
+        else:
+            log_durs_predicted = self.duration_predictor(enc_out, enc_mask)
```
rlangman, line 242: We could reduce code duplication by rewriting the if conditions like:

```
if self.speaker_emb_condition_prosody:
  prosody_input = enc_out + spk_emb
else:
  prosody_input = enc_out
log_durs_predicted = self.duration_predictor(prosody_input, enc_mask)
pitch_predicted = self.pitch_predictor(prosody_input, enc_mask)
```

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -230,19 +236,27 @@ def forward(
...
+                attn_soft, attn_logprob = self.aligner(spec, text_emb.permute(0, 2, 1), enc_mask == 0, attn_prior, conditioning=spk_emb)
+            else:
+                attn_soft, attn_logprob = self.aligner(spec, text_emb.permute(0, 2, 1), enc_mask == 0, attn_prior)
             attn_hard = binarize_attention_parallel(attn_soft, input_lens, mel_lens)
             attn_hard_dur = attn_hard.sum(2)[:, 0, :]
 
         # Predict pitch
-        pitch_predicted = self.pitch_predictor(enc_out, enc_mask)
+        if self.speaker_emb_condition_prosody:
+            pitch_predicted = self.pitch_predictor(enc_out + spk_emb, enc_mask)
```
rlangman, line 257: Summing features/embeddings is usually only done with transformers because of the limitation that their input & output dimensions must be equal. For other models like DNN/RNN/CNN the standard is to concatenate the feature vectors together. The latter approach looks like what is done in the RadTTS code as well. Did we ever try it?

To save refactoring time, it may be worth changing these booleans to string/enum and having something like:

```
if not self.speaker_emb_condition_prosody:
  prosody_input = enc_out
elif self.speaker_emb_condition_prosody == "add":
  prosody_input = enc_out + spk_emb
elif self.speaker_emb_condition_prosody == "concatenate":
  spk_emb_tile = spk_emb.unsqueeze(1).tile([1, text.shape[1], 1])
  prosody_input = torch.concat([enc_out, spk_emb_tile], dim=2)
else:
  raise ValueError()
```

nemo/collections/tts/modules/aligner.py
```diff
@@ -143,18 +143,21 @@ def get_mean_distance_for_word(l2_dists, durs, start_token, num_tokens):
...
             keys (torch.tensor): B x C2 x T2 tensor (text data).
             mask (torch.tensor): B x T2 x 1 tensor, binary mask for variable length entries (True = mask element, False = leave unchanged).
             attn_prior (torch.tensor): prior for attention matrix.
+            conditioning (torch.tensor): B x T2 x 1 conditioning embedding
         Output:
             attn (torch.tensor): B x 1 x T1 x T2 attention mask. Final dim T2 should sum to 1.
             attn_logprob (torch.tensor): B x 1 x T1 x T2 log-prob attention mask.
         """
+        if conditioning is not None:
+            keys = keys + conditioning.transpose(1,2)
```
rlangman, line 160: I think it would make more sense to condition/concatenate the speaker embedding to the spectrogram features, rather than the text features. Parsing spectrogram/audio benefits heavily from knowing the speakers vocal characteristics ahead of time.

(Text + speaker) is an ambiguous one to many mapping, while (spectrogram + speaker) to text is a deterministic many to one map. So conditioning the spectrogram parser will be more robust.

Or you could condition both, I guess.

nemo/collections/tts/modules/fastpitch.py
XuesongYang: agreed! tts collections have too many duplicated snippets. If we are not ready to refactor them now, it'd be better to avoid duplicate codes sooner or later.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Cheng-Ping had experimented with this and we did not get very good results or better results than what we have now. We can definitely add this feature in the future but it would need more experimentation.

nemo/collections/tts/modules/aligner.py
subhankar-ghosh: I guess we can do that (add condition vector to spectrogram also) but the reason why I did add to text vector only is because of dimensions. So dimensions of all vectors (text, speaker etc) are of same dimension but spectrogram vector dimension varies. So adding won't be that straight forward.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Cool, I will add this change.

nemo/collections/tts/modules/fastpitch.py
rlangman: My concern would be leaving it as a boolean means that fixing it later will break backwards compatibility. And adding in the functionality seems inevitable because:

(1) RadTTS does it, and we will need to merge/consolidate the common components.
(2) Concatenation creates a non-linearity with the feature vector that is needed if you condition on multiple feature vectors at once, or use a static fixed external feature vector like from eg. a speaker embedding or language model.

Though I guess (1) will be backwards incompatible regardless, so I guess we could leave it for then.

nemo/collections/tts/modules/aligner.py
rlangman, line 160: Out of curiosity, what is the benefit of passing the "conditioning" vector as an argument to the aligner, as opposed to having the method which calls aligner modify the keys?

Right now the FP module is conditioning the input for the pitch/duration predictors while the encoder & aligner take is as an input, which looks like an inconsistent design pattern.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -288,11 +303,18 @@ def infer(self, *, text, pitch=None, speaker=None, pace=1.0, volume=None):
         enc_out, enc_mask = self.encoder(input=text, conditioning=spk_emb)
 
         # Predict duration and pitch
-        log_durs_predicted = self.duration_predictor(enc_out, enc_mask)
+        if self.speaker_emb_condition_prosody:
+            log_durs_predicted = self.duration_predictor(enc_out + spk_emb, enc_mask)
+        else:
+            log_durs_predicted = self.duration_predictor(enc_out, enc_mask)
```
rlangman, line 309: We refactored the code in the forward() method but not the same code in the infer() method.

Not necessary here, but ideally we can avoid this by replacing any duplicated logic with a helper function.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -262,7 +274,10 @@ def forward(
             len_regulated, dec_lens = regulate_len(durs_predicted, enc_out, pace)
 
         # Output FFT
-        dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens)
+        if self.speaker_emb_condition_decoder:
+            dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens, conditioning=spk_emb)
+        else:
+            dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens)
```
rlangman, line 280: Can similarly reduce duplication like:

```
if self.speaker_emb_condition_decoder:
  dur_conditioning = spk_emb
else:
  dur_conditioning  = 0.0
dec_out, _ = self.decoder(input=len_regulated, seq_lens=dec_lens, conditioning=spk_emb)
```

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Makes sense. Done.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: I believe the current implementation is intuitive in terms of readability. Like we can see that there is no conditioning in one case and there is conditioning in another. Whereas it might be a little confusing as in what is 0.0 conditioning. This is what I think, let me know if that makes sense.

rlangman: Setting it to 0.0 is confusing. Leaving it as is would be alright. Some alternatives would be:

(a) Set conditioning = None and then change the transformer logic to handle the None input.
(b) Make the behavior consistent with the pitch/duration predictor by removing "conditioning" argument and setting input = len_regulated or len_regulated + spk_emb

nemo/collections/tts/modules/aligner.py
subhankar-ghosh: I think transformers in our implementation has the conditioning parameter and other CNN/DNNs do not. We can change that for sure, but it should be a separate refactoring PR.

rlangman (approved):
LGTM. Thanks for the changes & comments.

subhankar-ghosh:
Closing. Related PR https://github.com/NVIDIA/NeMo/pull/4986

PR #4972, opened 2022-09-21
Introduce pre-commit with pre-commit suggestions

SeanNaren:
# What does this PR do?

Currently style checking is done within the `setup.py` and enforced via jenkins.

This is problematic for users who do not have jenkins access and builds fail due to simple style fixes. By using pre-commit CI, we can automate style fixes in PRs meaning less work for users (and have a dedicate CI job for this).

As a side effect of this, due to the old version of black running I had to upgrade the Black version from the pre-release (19) to a stable version (22.*). Due to this there has been some additional style changes. A few summarized below:

- Lots of whitespace removed
- Multiple argument functions are expanded based on length. this is primarily for readability (and hence why the number of lines go up):

i.e 
```python
# old
{'label': NeuralType(tuple('B'), LabelsType()), 'label_length': NeuralType(tuple('B'), LengthsType()),}

# new
{
    'label': NeuralType(tuple('B'), LabelsType()),
    'label_length': NeuralType(tuple('B'), LengthsType()),
}
``` 

# Changelog 
- Introduce pre-commit, update Black version.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation


SeanNaren:
I'm seeing quite a lot of code changes from Black, might need to play with the formatter to get it to replicate the current setup.

SeanNaren:
Since we were using a really old version of Black, I was running into this: https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

All suggestions say to upgrade Black, upgrading black also updated the formatting for many files. I think on some quick inspections the suggestions are fine (imo) but @ericharper @titu1994 let me know if it's reasonable.

I think eventually we would drop the style check from jenkins to reduce time.

SeanNaren:
Going to wait till 1.12 is released, so that we can pull this in without causing cherry-picking issues.

PR #4984, opened 2022-09-22
Fix ASR issues

titu1994:

# What does this PR do ?
Bug fixes

**Collection**: [ASR]

# Changelog 

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/asr/conf/marblenet/marblenet_3x2x64.yaml
+++ b/examples/asr/conf/marblenet/marblenet_3x2x64.yaml
@@ -167,7 +167,7 @@ trainer:
   max_epochs: 150
   max_steps: -1 # computed at runtime if not set
   num_nodes: 1
-  accelerator: gpu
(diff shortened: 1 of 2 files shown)


nithinraok (commented):
It should be 'gpu' . Cuda usage for accelarator is deprecated

titu1994:
Hmm. Can you take the nvbug then ? Seems the test crashed with that.

nithinraok:
Seems like cuda is not deprecated yet, tests are passing though. But on the src code it's not available, no idea what's happening. https://github.com/Lightning-AI/lightning/blob/dbb5ca8d436a917fc9c2bdd6e9c00e4fd6187735/src/pytorch_lightning/trainer/trainer.py#L196

nithinraok:
It not the problem with code. You may close this, will be following up this on nvbug

PR #5023, opened 2022-09-27
Bug fix - Limit val batches set to 1.0 

shanmugamr1992:
Signed-off-by: shanmugamr1992 <shanmugamr1992@gmail.com>

# What does this PR do ?
* Limit val batches 1.0 does not work as expected when num validation iterations * global batch size > num samples

Large language model

# Changelog 
* If limit val batches is set to 1.0 it automatically uses only one epoch of validation data for every validation iteration
* Bug fix when adding blendable dataset as values for train / test / validation keys . 

# Usage
limit_val_batches: 1.0

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -510,12 +510,15 @@ def build_train_valid_test_datasets(self):
...
         eval_iters = (max_train_steps // self.trainer.val_check_interval + 1) * self.trainer.limit_val_batches
         test_iters = self.trainer.limit_test_batches
-
+        
         train_valid_test_num_samples = [
             max_train_steps * global_batch_size,
             eval_iters * global_batch_size,
             test_iters * global_batch_size,
         ]
+        if self.trainer.limit_val_batches == 1.0 and type(self.trainer.limit_val_batches) == float:
```
MaximumEntropy, line 519: Can you raise an error here if `limit_val_batches` is not an int or != `1.0`? I think intermediate values like `0.6` etc do not work as expected right?

MaximumEntropy (changes requested):
Looks good, thanks! Just a minor comment

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
shanmugamr1992: Yes. Sure will do that

shanmugamr1992: Actually I think it will work. If the limit val batches is a float value less than 1, it will work as long as we set the train_valid_test_num_samples[1] to 1. . I am going to == 1.0 to <= 1.0 condition. I tested it out on my local box. 
I will raise an error if the limit val batches is set to a value greater than 1.0.

MaximumEntropy (approved):

PR #5028, opened 2022-09-28
patch to fix tao error for bug 3809486

charul15:
# What does this PR do ?

Added patch for tao realted bug 3809486. Included functionality to restore the value of is_model_being_restore to true if user has sent restore_attr_is_model_being_restore=True

# Changelog 
- modified nlp_model.py to restore is_model_being_restore=True as at line 81( bert_model = get_lm_model(..) ) while intializing the model , the states is_model_being_restore is set to false . This  needs to be true to initialize the model (at line super().__init__(cfg, trainer) ). 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/models/nlp_model.py
+++ b/nemo/collections/nlp/models/nlp_model.py
@@ -86,6 +86,8 @@ class NLPModel(ModelPT, Exportable):
                 )
                 # explicitly setting the is_model_being_restored to true becuase in above model
                 # initialization in get_lm_model function these states are being set to false
+                # and to reinitialize the model below at line super().__init__(cfg, trainer)
+                # the state is_model_being_restored needs to be true
                 AppState().is_model_being_restored = True
                 # set the tokenizer if it is not initialized explicitly
                 if (
--- a/nemo/collections/nlp/models/nlp_model.py
+++ b/nemo/collections/nlp/models/nlp_model.py
@@ -84,6 +84,8 @@ class NLPModel(ModelPT, Exportable):
                     trainer=trainer,
                     cfg=copy_cfg,
                 )
+                # explicitly setting the is_model_being_restored to true
+                AppState().is_model_being_restored = True
                 # set the tokenizer if it is not initialized explicitly
                 if (
                     (hasattr(self, 'tokenizer') and self.tokenizer is None) or not hasattr(self, 'tokenizer')
@@ -95,8 +97,6 @@ class NLPModel(ModelPT, Exportable):
                 and cfg.get('tokenizer').get('vocab_file')
             ):
                 vocab_file = self.register_artifact('tokenizer.vocab_file', cfg.tokenizer.vocab_file)
-        if not AppState().is_model_being_restored and AppState().restore_attr_is_model_being_restored:
-            AppState().is_model_being_restored = True
 
         super().__init__(cfg, trainer)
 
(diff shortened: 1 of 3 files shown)


nemo/collections/nlp/models/nlp_model.py
```diff
@@ -95,6 +95,10 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None, no_lm_init=False):
                 and cfg.get('tokenizer').get('vocab_file')
             ):
                 vocab_file = self.register_artifact('tokenizer.vocab_file', cfg.tokenizer.vocab_file)
+        
+        if (not AppState().is_model_being_restored and AppState().restore_attr_is_model_being_restored ):
+            AppState().is_model_being_restored = True
```
ericharper, line 100: I don't understand why you have to create the attribute `restore_attr_is_model_being_restored` . Is the issue that AppState().is_model_being_restored not set properly when restoringthe model?

charul15: yes , before restoring we explicitly set the is_model_being_restored to true using function set_model_restore_state . But while restoring it gets overwritten to false (inside if block at line 70 - if cfg.get('tokenizer'):).
That is why this new attribute is created restore_attr_is_model_being_restored which is by default false but if we set it to true then it will restore true state of is_model_being_restored.

yidong72: why do you want to manually modify `is_model_being_restored`? It seems that this variable is temporally used when loading the checkpoint file. 
```
    @classmethod
    def load_from_checkpoint(
        cls,
        checkpoint_path: str,
        *args,
        map_location: Optional[Union[Dict[str, str], str, torch.device, int, Callable]] = None,
        hparams_file: Optional[str] = None,
        strict: bool = True,
        **kwargs,
    ):
        """
        Loads ModelPT from checkpoint, with some maintenance of restoration.
        For documentation, please refer to LightningModule.load_from_checkpoint() documentation.
        """
        checkpoint = None
        try:
            cls._set_model_restore_state(is_being_restored=True)

            checkpoint = super().load_from_checkpoint(
                checkpoint_path=checkpoint_path,
                *args,
                map_location=map_location,
                hparams_file=hparams_file,
                strict=strict,
                **kwargs,
            )

        finally:
            cls._set_model_restore_state(is_being_restored=False)
        return checkpoint

```

nemo/collections/nlp/models/nlp_model.py
```diff
@@ -84,6 +84,8 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None, no_lm_init=False):
                     trainer=trainer,
                     cfg=copy_cfg,
                 )
+                # explicitly setting the is_model_being_restored to true
+                AppState().is_model_being_restored = True
```
ericharper, line 88: Please explain more in comments what is happening here and why we're doing line 88.

charul15: At line no 81 while intializing the model , the states is_model_being_restore is set to false which is conflicting below when we again initailize the model at line 101 super().__init__(). Hence manually resetting the is_model_being_restore=True.
modified changelog as well

ericharper (approved):
LGTM. Thanks!

Issue #5136, opened 2022-10-11
checkpoint_dir with checkpoint manager is not being set corrector while resuming training


daniel-kukiela:
**Describe the bug**

When I stopped fine-tuning the FastPitch model and wanted to restart training, I noticed it checks for the last checkpoint in the wrong directory (I'm settings the checkpoints directory outside of the logs).
So, in my case the checkpoints are being saved at `models/fastpitch/checkpoints` but it is trying to search in `models/fastpitch/logs/checkpoints`.
This is due to the hard-coded path value here: https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L424, so inside `check_resume()`:
```py
    checkpoint_dir = Path(Path(log_dir) / "checkpoints")
```
While the correct path is being set here: https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L889, so inside `configure_checkpointing()`:
```py
    if params.dirpath is None:
        params.dirpath = Path(log_dir / 'checkpoints')
```
it correctly checks if `dirpath` is set (it is in my case) and only otherwise defaults to the `checkpoints` folder inside of the `logs` folder.

I did not fill in the rest of the template of the bug report as this is a question about a specific code piece, not related to any hardware or scenario.

redoctopus:
Sorry, I'm not exactly sure what you're asking, but if you're having trouble with the `exp_manager` finding the correct directory for checkpoints, can you try adding `+exp_manager.explicit_log_dir=<your_log_path>`? This should ensure that the two runs will always use the same directory.

The two code snippets should give the same result assuming `log_dir` is consistent between them.

daniel-kukiela:
What I mean is you can use `exp_manager.checkpoint_callback_params.dirpath` to set a checkpoint folder and it can be any path, not necessarily `checkpoints` inside of `logs`. If you use this parameter, then (https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L424):
```py
    if params.dirpath is None:
        params.dirpath = Path(log_dir / 'checkpoints')
```
so the basove code inside of the if statement is not being executed and the set paths is being used instead. Checkpoints indeed save in this directory of my choice. However, when I want to resume training, the checkpoint path (dirpath) set in exp_manager is ignored and a hardcoded one is used instead (https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L424):
```py
[...]

PR #5155, opened 2022-10-12
Fix loading checkpoints from dirpath if it's specified

redoctopus:

# What does this PR do ?

Fixes checkpoint loading if `exp_manager.checkpoint_callback_params.dirpath` is specified.

**Collection**: Utils

# Changelog 
- Adds a check for whether `dirpath` has been set, and passes it in to `check_resume()` to use instead of `<log_dir>/checkpoints` for checkpoint restoration
- Unit tests added for restoring from `dirpath`
- Tested load/store for both `dirpath` being set and not (using `<log_dir>/checkpoints`) and both now work

# Usage

The following should now save to AND load from the specified `ckpt_dir`.

```bash
python examples/tts/fastpitch.py \
train_dataset=train.json \
validation_datasets=val.json \
trainer.devices=1 \
trainer.max_epochs=5 \
+exp_manager.checkpoint_callback_params.dirpath=ckpt_dir/ \
+exp_manager.explicit_log_dir=logdir/ \
exp_manager.resume_if_exists=true \
exp_manager.resume_ignore_no_checkpoint=true \
trainer.check_val_every_n_epoch=1
```

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information
* Related to #5136


ericharper:
Could you add a unit test for this?

redoctopus:
> Could you add a unit test for this?

Added one to check failure when the specified directory is empty, and one to check that the directory matches what the model is using to load.

ericharper (approved):
LGTM. Thanks!

PR #5169, opened 2022-10-14
Move to optimizer based EMA implementation

SeanNaren:
Signed-off-by: SeanNaren <snarenthiran@nvidia.com>

# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/common/callbacks/ema.py
+++ b/nemo/collections/common/callbacks/ema.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,25 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import os.path
-import warnings
-from typing import Any, Dict, List, Optional
+import contextlib
+import copy
+import threading
 
 import pytorch_lightning as pl
 import torch
 from pytorch_lightning import Callback
-from pytorch_lightning.utilities import rank_zero_warn
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
-from pytorch_lightning.utilities.types import STEP_OUTPUT
-
-from nemo.utils import logging
-
-try:
-    import amp_C
-
-    apex_available = True
-except Exception:
-    apex_available = False
 
 
 class EMA(Callback):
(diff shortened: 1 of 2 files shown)


nemo/utils/exp_manager.py
```diff
@@ -100,10 +100,7 @@ class StepTimingParams:
 @dataclass
 class EMAParams:
     enable: Optional[bool] = False
-    evaluate_ema_weights_instead: Optional[bool] = False
     decay: Optional[float] = 0.999
-    apply_ema_every_n_steps: Optional[int] = 1
-    start_step: Optional[int] = 0
```
titu1994, line 106: We can support start step and apply_ema_every_n_steps by using the step count inside of an optimizer right? Or is  it dependent on the optimizer whether it maintains the counter?

nemo/collections/common/callbacks/ema.py
```diff
@@ -42,143 +31,188 @@ class EMA(Callback):
...
+        with torch.cuda.stream(self.stream):
+            current_model_state = tuple(
+                param.data.to(self.device, non_blocking=True) for param in self.all_parameters()
+            )
+
+            if self.device.type == 'cuda':
+                ema_update(self.ema_params, current_model_state, self.decay)
+
+        if self.device.type == 'cpu':
+            self.thread = threading.Thread(
```
titu1994, line 156: Does PT encourage thread use for tensors? Id think the cost of marshalling the tensor would be prohibitive?

markelsanz14:
Leaving a few comments for next week:

- We should make the validation configurable, so that the user can decide whether to run val with regular or ema weights.
- We should make the CPU offloading configurable too, for both the ema weights update, and for keeping the ema weights in CPU mem rather than GPU mem.
- We should update the config yaml files of all the models that will support EMA to add the ema parameters to the exp_manager config. We can default it to False, but it should be present in the config.
- Maybe we should set the default decay value to 0.9995 in the config files and in the code. I think that should be a decent default value.
- I found this error when the model tries to save a checkpoint:
```
  File "/opt/conda/lib/python3.8/site-packages/pytorch_lightning/callbacks/model_checkpoint.py", line 713, in _update_best_and_save
    self._save_checkpoint(trainer, filepath)
  File "/opt/bignlp/NeMo/nemo/utils/exp_manager.py", line 904, in _save_checkpoint
    ema_callback.replace_model_weights(trainer.lightning_module)
AttributeError: 'EMA' object has no attribute 'replace_model_weights'
```

SeanNaren:
Have quite a few local changes and improvements, just figuring out how to ensure that the "non" EMA checkpoint can be re-loaded without the optimizer wrapper.

SeanNaren:
With how the checkpoint needs to be overwritten, we require PyTorch Lightning 1.8 for the new updated behaviour where the callback has access to the entire state of the checkpoint. I've started the PR here to fix issues: https://github.com/NVIDIA/NeMo/pull/5353

nemo/collections/common/callbacks/ema.py
szmigacz: The goal of this code is to execute CPU EMA update asynchronously, Thread is the simplest way to achieve that, I didn't find any issues with this code when running tests.

SeanNaren:
@titu1994 ready for review :) fixed the tests and should now be at feature parity with what we had existing!

titu1994 (approved):
PR looks awesome. Minor comments. Also for training test, could you add a batch norm to the model so that it tests what happens if ema is done on the int component (num steps) for batchnorm. I wonder if there's any literature for that case, maybe we should take max instead of ema of int

nemo/collections/common/callbacks/ema.py
```diff
@@ -142,43 +120,226 @@ def on_load_checkpoint(
...
+        device: torch.device,
+        decay: float = 0.9999,
+        every_n_steps: int = 1,
+        current_step: int = 0,
+    ):
+
+        # copy most of the `Optimizer` methods into this instance. `__del__` is skipped in case the optimizer has
+        # implemented custom logic which we would not want to call on destruction of the `EMAOptimizer`.
+        # this allows us to use the underlying optimizer without having to go through the EMAOptimizer wrapper.
+        self.__dict__ = {k: v for k, v in optimizer.__dict__.items() if k not in ("step", "__del__")}
```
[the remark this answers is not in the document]
SeanNaren: aha, I think we can remove this. the primary reason I added this boilerplate code was to try fix O2 support, however currently O2 isn't working with EMA.

nemo/collections/common/callbacks/ema.py
```diff
@@ -142,43 +120,226 @@ def on_load_checkpoint(
...
+        self.current_step += 1
+        return loss
+
+    def _should_update_at_step(self) -> bool:
+        return self.current_step % self.every_n_steps == 0
+
+    @torch.no_grad()
+    def update(self):
+        if self.stream is not None:
+            self.stream.wait_stream(torch.cuda.current_stream())
```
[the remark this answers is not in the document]
szmigacz: Yes, optimizer.step needs to be completed before we start EMA update

SeanNaren:
Just included batch norm into the model, which causes the tests to fail. Will debug!

cc @titu1994 @szmigacz

amorehead:
@malbergo, as an aside, I was successfully able to make this EMA callback work my needs in the context of 3D diffusion generative models. If you'd like to see an example of how I make this work inside my custom LightningModules, feel free to check out the following links to my `bio-diffusion` repository:

- EMA callback: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/src/utils/__init__.py#L71
- Custom Lightning checkpoint callback that wraps the EMA callback to save EMA versions of each model checkpoint
- How I set the arguments for this EMA callback using Hydra config files: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/configs/callbacks/ema.yaml#L5
- How I include the EMA callback as one of many Lightning callbacks for my Trainer to make use of: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/configs/callbacks/default.yaml#L4

SeanNaren:
This is fantastic @amorehead! We do not have documentation explicitly for using this callback with pure LightningModules, but @amorehead code seems to be a good place to start!

Here are the docs for using it with NeMo however: https://docs.nvidia.com/deeplearning/nemo/user-guide/docs/en/stable/common/callbacks.html#exponential-moving-average-ema

flashszn:
> @malbergo, as an aside, I was successfully able to make this EMA callback work for my needs in the context of 3D diffusion generative models. If you'd like to see an example of how I make this work inside my custom LightningModules, feel free to check out the following links to my `bio-diffusion` repository:
> 
> * EMA callback: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/src/utils/__init__.py#L71
> * Custom Lightning checkpoint callback that wraps the EMA callback to save EMA versions of each model checkpoint: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/src/utils/__init__.py#L224
> * How I set the arguments for this EMA callback using Hydra config files: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/configs/callbacks/ema.yaml#L5
> * How I include the EMA callback as one of many Lightning callbacks for my Trainer to make use of: https://github.com/BioinfoMachineLearning/bio-diffusion/blob/e4bad15139815e562a27fb94dab0c31907522bc5/configs/callbacks/default.yaml#L4
> 
> I hope this helps!

Thanks a lot. When I use this callback with top-k ckpt, it won't automatically del the ckpt with -EMA and will save a lot of ckpts with -EMA.

So I modify the `EMAModelCheckpoint`, overriding the `_update_best_and_save`:
```
# only change the last line
def _update_best_and_save(
        self, current: Tensor, trainer: "pl.Trainer", monitor_candidates: Dict[str, Tensor]
) -> None:
    k = len(self.best_k_models) + 1 if self.save_top_k == -1 else self.save_top_k

    del_filepath = None
    if len(self.best_k_models) == k and k > 0:
        del_filepath = self.kth_best_model_path
        self.best_k_models.pop(del_filepath)

    # do not save nan, replace with +/- inf
    if isinstance(current, Tensor) and torch.isnan(current):
        current = torch.tensor(float("inf" if self.mode == "min" else "-inf"), device=current.device)

    filepath = self._get_metric_interpolated_filepath_name(monitor_candidates, trainer, del_filepath)

    # save the current score
    self.current_score = current
    self.best_k_models[filepath] = current

    if len(self.best_k_models) == k:
        # monitor dict has reached k elements
        _op = max if self.mode == "min" else min
        self.kth_best_model_path = _op(self.best_k_models, key=self.best_k_models.get)  # type: ignore[arg-type]
        self.kth_value = self.best_k_models[self.kth_best_model_path]

    _op = min if self.mode == "min" else max
    self.best_model_path = _op(self.best_k_models, key=self.best_k_models.get)  # type: ignore[arg-type]
    self.best_model_score = self.best_k_models[self.best_model_path]

    if self.verbose:
        epoch = monitor_candidates["epoch"]
        step = monitor_candidates["step"]
        rank_zero_info(
            f"Epoch {epoch:d}, global step {step:d}: {self.monitor!r} reached {current:0.5f}"
            f" (best {self.best_model_score:0.5f}), saving model to {filepath!r} as top {k}"
        )
    self._save_checkpoint(trainer, filepath)

    if del_filepath is not None and filepath != del_filepath:
        self._remove_checkpoint(trainer, del_filepath)
        self._remove_checkpoint(trainer, del_filepath.replace(self.FILE_EXTENSION, f"-EMA{self.FILE_EXTENSION}"))
```

PR #5194, opened 2022-10-18
checks for accepted adapter type at module level

arendu:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/modules/common/megatron/adapters/adapter_constants.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class AdapterType(str, enum.Enum):
+    # Adapters requires in the IA3 schema
+    KEY_INFUSED = "key_infused_adapter"
+    VALUE_INFUSED = "value_infused_adapter"
+    MLP_INFUSED = "mlp_infused_adapter"
+
+    # Standard Adapters
+    ADAPTER_ONE = "adapter_1"
```
titu1994, line 26: Very poor names, please chose something more meaningful.

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -150,6 +150,20 @@ class AdapterModuleMixin(ABC):
     adapter_global_cfg_key = "global_cfg"
     adapter_metadata_cfg_key = "adapter_meta_cfg"
 
+    def set_accepted_adapters(self, adapters):
```
titu1994, line 153: Add type + docstring. Is ti a string, list of string, enum, list of enum ?

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -150,6 +150,20 @@ class AdapterModuleMixin(ABC):
...
+        self.accepted_adapters = adapters
+
+    def get_accepted_adapters(self,):
+        if hasattr(self, 'accepted_adapters'):
```
titu1994, line 157: Docstring

nemo/core/classes/mixins/adapter_mixins.py
titu1994: Also, this should probably be recursive, setting the value for all children of self.

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -150,6 +150,20 @@ class AdapterModuleMixin(ABC):
...
+            return self.accepted_adapters
+        else:
+            return []
+
+    def get_from_adapter_layer(self, name: str):
```
titu1994, line 162: No need for this method. ModuleDict has get(str, default) semantics, use that.

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -159,6 +173,11 @@ def add_adapter(self, name: str, cfg: DictConfig):
             cfg: A DictConfig or Dataclass that contains at the bare minimum `__target__` to instantiate a
                 new Adapter module.
         """
+        if self.get_accepted_adapters():
+            assert (
```
titu1994, line 177: Dont assert, check with if and raise appropriate error.

titu1994: This check is also incorrect. Name should be name of the class, not some random name.
Adapters can have any name, including module adapter style names ("module:adapter_name").

You can check class via class inflation (look at model_utils.import_class_from_path) or via str match of class, but not of individual adapter names.

titu1994 (changes requested):
Requires major changes.

nemo/core/classes/mixins/adapter_mixins.py
arendu: :) I thought the same, but turns out ModuleDict does not have `get(str, default)`

nemo/collections/nlp/modules/common/megatron/adapters/adapter_constants.py
arendu: The name change would break backward compatibility. For now, I just want this PR to implement the checking feature. And I'll have another for dealing with backward compatability.

nemo/core/classes/mixins/adapter_mixins.py
titu1994: Hmm does using model_utils.import_class_from_path() here cause circular dependency? If not, might as well load the class and do an issubclass() check rather than explicit str check. 

Subclasses can then be supported, plus if paths change in the future due to refactor, older checkpoints won't break (since we can do import in old dir init files).

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -159,6 +182,11 @@ def add_adapter(self, name: str, cfg: DictConfig):
             cfg: A DictConfig or Dataclass that contains at the bare minimum `__target__` to instantiate a
                 new Adapter module.
         """
+        if self.get_accepted_adapter_types():
```
titu1994, line 185: Check returned value is not none, and isinstance (Iterable) and finally len(output) > 0 before entering. This current check is very risky for older models

titu1994 (changes requested):
Minor comments, overall looks much better

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -159,6 +182,14 @@ def add_adapter(self, name: str, cfg: DictConfig):
             cfg: A DictConfig or Dataclass that contains at the bare minimum `__target__` to instantiate a
                 new Adapter module.
         """
+        if hasattr(self, '_accepted_adapter_types'):
```
titu1994, line 185: Instead of directly accessing the variables l, use `get_accepted_adapter_types(self,)` to get the list and then check length.

nemo/core/classes/mixins/adapter_mixins.py
```diff
@@ -159,6 +182,14 @@ def add_adapter(self, name: str, cfg: DictConfig):
...
+            if len(self._accepted_adapter_types) > 0:
+                test = model_utils.import_class_by_path(cfg._target_)
+                if test not in self._accepted_adapter_types:
+                    raise ValueError(
```
titu1994, line 189: Don't do a test like this. Create a bool, loop through the classes, do issubclass(test, cls_in_loop) and then if it's true, set the flag and break, check that the book is true, else raise value error.

titu1994 (changes requested):

nemo/core/classes/mixins/adapter_mixins.py
arendu: Whats the benefit of subclass checks? I'd rather check the exact class that is suppose to be added. I can imagine cases where the subclass behavior is very different from the parent, e.g. suppose we have MoE-Adapters subclassing Adapters.

arendu: How about leaving it as exact check for now and if subclass is needed in the future we can revisit? @titu1994

arendu: nvm, done as you suggested.

titu1994 (dismissed):
LGTM, thanks !

titu1994 (approved):

PR #5196, opened 2022-10-18
[ASR] AudioToAudio datasets and related test

anteju:
# What does this PR do ?

This is a draft PR of datasets for different audio-to-audio tasks.

**Datasets**

- `BaseAudioDataset`: Abstract base class.
- `AudioToTargetDataset`: A dataset for audio-to-audio tasks where the goal is to use an input signal to recover the corresponding target signal.
- `AudioToTargetWithReferenceDataset`: A dataset for audio-to-audio tasks where the goal is to use an input signal to recover the corresponding target signal and an additional reference signal is available.
- `AudioToTargetWithEmbeddingDataset`: A dataset for audio-to-audio tasks where the goal is to use an input signal to recover the corresponding target signal and an additional embedding signal. It is assumed that the embedding is in a form of a vector.

**Tests**

Multiple tests are implemented in `test_asr_datasets.py` in class `TestAudioDatasets`. These tests include

- `test_audio_to_target_dataset`: multiple tests for `AudioToTargetDataset`
- `test_audio_to_target_dataset_with_target_list`: tests specifically a scenario where target is provides as a list of files
- `test_audio_to_target_with_reference_dataset`: tests for `AudioToTargetWithReferenceDataset`
- `test_audio_to_target_with_embedding_dataset`: tests for `AudioToTargetWithEmbeddingDataset`

Tests can be started using the following command
```
pytest tests/collections/asr/test_asr_datasets.py::TestAudioDatasets
```

**Collection**: ASR

# Changelog 
- Added datasets in `audio_to_audio.py`
- Added unit tests in `test_asr_datasets.py` and `test_audio_utils.py`
- Added an option in `AudioSegment. segment_from_file` to use a fixed offset

# Usage

Usage is illustrated in unit tests, which can be executed using

```python
pytest tests/collections/asr/test_asr_datasets.py::TestAudioDatasets
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/data/audio_to_audio.py
+++ b/nemo/collections/asr/data/audio_to_audio.py
@@ -16,7 +16,7 @@ import abc
 import math
 import random
 from collections import OrderedDict
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import librosa
 import numpy as np
@@ -24,7 +24,7 @@ import torch
 
 from nemo.collections.asr.parts.preprocessing.segment import AudioSegment
 from nemo.collections.asr.parts.utils.audio_utils import ChannelSelectorType
-from nemo.collections.common.parts.preprocessing.collections import AudioCollection
+from nemo.collections.common.parts.preprocessing import collections
 from nemo.collections.common.parts.utils import flatten
 from nemo.core.classes import Dataset
 from nemo.core.neural_types import AudioSignal, EncodedRepresentation, LengthsType, NeuralType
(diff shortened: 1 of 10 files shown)


nemo/collections/common/parts/preprocessing/collections.py
```diff
@@ -742,3 +744,151 @@ def __parse_item_rttm(self, line: str, manifest_file: str) -> Dict[str, Any]:
...
+        """
+        # Keys from manifest which contain audio
+        self.audio_keys = audio_keys
+
+        # Initialize data
+        audio_files_list, duration_list, offset_list = [], [], []
+
+        # Parse manifest files
+        for item in manifest.item_iter(manifest_files, parse_func=self.__parse_item):
+            audio_files_list.append(item['audio_files'])
```
yzhang123, line 831: what about not just using AudioColleciton for manifest loading but use a general loader that loads the entire manifest? Maybe we have a 'text' key in manifest. And then in the dataset class we can pass certain keys to a more specific collection class?

yzhang123:
Thanks for this draft @anteju !  This is very helpful.
I wonder how the pipeline will look like if you add all audio based augmentations/perturbations. could you add an example?

yzhang123:
> The most straightforward way would be to add `augmentor` object to datasets, similarly to existing datasets.
Yes, good idea!
> > Also, i am not sure if this is out of scope, but it would be great to also see an example where we have (multiple) audios and (multiple) text.
> 
> Could you please clarify `multiple audios and multiple text`?

Maybe have a example dataset class, which in __get_item__ can return two or more audio files, and two and more transcripts? This is a most general class that comes to mind, and is POC for the general design you worked on. This would allow us to then easily adapt the class for TSASR, which returns two audios and 1 transcript

titu1994:
Augnentor should be a detached operation, even if it's part of the config. Ie don't use the way we do it right now where we have no control over augmentation and it's all random inside of Waveform Featurizer. Let's seperate and make the call to augmentation more controllable inside of the new data loaders

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,891 @@
...
+    fixed_offset: float = 0,
+) -> np.ndarray:
+    """Load samples from an audio file.
+
+    Args:
+        audio_file: path to an audio file
+        sample_rate: desired sample rate for output samples
+        duration: Optional desired duration of output samples.
+                  If `None`, complete file will be loaded.
+                  If set, a random segment of `duration` seconds will be
```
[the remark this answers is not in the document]
tango4j: It would be great of the function docsting explains more about optional multichannel aspect.
I think diarization/buffer-ASR codes should use this function for removing duplications

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+from nemo.utils.decorators import experimental
+
+__all__ = [
+    'AudioToTargetDataset',
+    'AudioToTargetWithReferenceDataset',
+    'AudioToTargetWithEmbeddingDataset',
+]
+
+
+def load_samples_synchronized(
```
titu1994, line 41: Could you make this a class and factorize its components inside the function so that people can swap out parts of it. Then create objects of that class / subclass and use it inside of the actual datasets.

Make it similar to ASRManifestProcessor basically.

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+                    )
+                    samples_list.append(segment.samples)
+                output.append(samples_list)
+            else:
+                raise RuntimeError(f'Unexpected audio_file type {type(audio_file)}')
+
+    return output
+
+
+def load_samples(
```
titu1994, line 161: Part of above class ^

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+        duration=duration,
+        channel_selectors=[channel_selector],
+        fixed_offset=fixed_offset,
+        random_offset=random_offset,
+    )
+
+    return output[0]
+
+
+def load_embedding(filepath: str) -> np.ndarray:
```
titu1994, line 202: Part of above class ^

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+    if filepath.endswith('.npy'):
+        with open(filepath, 'rb') as f:
+            embedding = np.load(f)
+    else:
+        raise RuntimeError(f'Unknown embedding file format in file: {filepath}')
+
+    return embedding
+
+
+def list_to_multichannel(signal: Union[np.ndarray, List[np.ndarray]]) -> np.ndarray:
```
titu1994, line 221: Part of above class ^

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+        self,
+        manifest_filepath: str,
+        sample_rate: int,
+        audio_to_manifest_key: Dict[str, Union[str, List[str]]],
+        audio_duration: Optional[float] = None,
+        random_offset: bool = False,
+        max_duration: Optional[float] = None,
+        min_duration: Optional[float] = None,
+        max_utts: int = 0,
+    ):
```
titu1994, line 367: Base  audio dataset will take object of that class as input and set it as variable. Note that this will NOT instantiate the class on its own, it must accept a subclasses value only. 

So basically if ASRAudioProcessor (tentative name) needs to be subclassed by a subclass of BaseAudioDataset then it will look like below

ASRAudioProcessor -> MySpecificAudioProcessor
BaseAudioProcessor __init__ -> MyCustomAudioToAudioDataset __init__ loads an object of MySpecificAudioProcessor and passes it to BaseAudioProcessor.__init)__

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+        
+        Args:
+            signal_key: string, used to select a signal from the dictionary
+                        output by __getitem__
+
+        Returns:
+            Number of channels for the selected signal.
+        """
+        # Assumption: whole dataset has the same number of channels
+        item = self.__getitem__(0)
```
titu1994, line 412: Do not assume it is an indexed dataset. This may be invalid for tarred datasets.

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,936 @@
...
+        super().__init__()
+
+        if type(manifest_filepath) == str:
+            manifest_filepath = manifest_filepath.split(',')
+
+        for audio_key, manifest_key in audio_to_manifest_key.items():
+            if type(manifest_key) == str and ',' in manifest_key:
+                audio_to_manifest_key[audio_key] = manifest_key.split(',')
+
+        self.collection = AudioCollection(
```
titu1994, line 379: Do not auto instantiate collection. Take it as input in __init__ - this allows subclasses from generalizing and extending the collection without depending on base class.

titu1994 (changes requested):
Sorry for the delay, overall it looks very good, just needs some detachment of the audio processor and the base class.

nemo/collections/asr/parts/preprocessing/segment.py
```diff
@@ -257,12 +257,22 @@ def from_file(
...
 
         Note that audio_file can be either the file path, or a file-like object.
+
+        :param audio_file: path to a file or a file-like object
+        :param target_sr: sample rate for the output samples
+        :param n_segments: desired number of samples
+        :param trim: if true, trim leading and trailing silence from an audio signal
+        :param orig_sr: the original sample rate
+        :param channel selector: select a subset of channels. If set to `None`, the original signal will be used.
+        :param offset: fixed offset in seconds
```
XuesongYang, line 274: nit: `n_segments` is in samples, while `offset` is in seconds. Sometimes mixed units may bring confusion. But I am fine with the second unit since it ties to your usage case. Maybe we can rename it as `offset_secs`.

nemo/collections/asr/parts/preprocessing/segment.py
```diff
@@ -257,12 +257,22 @@ def from_file(
 
     @classmethod
     def segment_from_file(
-        cls, audio_file, target_sr=None, n_segments=0, trim=False, orig_sr=None, channel_selector=None,
+        cls, audio_file, target_sr=None, n_segments=0, trim=False, orig_sr=None, channel_selector=None, offset=None
```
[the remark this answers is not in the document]
XuesongYang: nit: `offset` can be enforced as a float type. All negative values mean no offsets. So we could make it default to `-1.0` to specify no offsets. Then benefit is that we can have a cleaner type hint.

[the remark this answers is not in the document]
XuesongYang: it seems an easy add-ons with type hint for this function. Do you mind if enforcing that? Thanks.

XuesongYang (commented):
LGTM for `segment_from_file()` func. Added some comments accordingly. Thanks.

XuesongYang:
discussed offline. Approved for my parts and please hold off merging until other folks approve.

XuesongYang (dismissed):

tango4j (changes requested):

nemo/collections/asr/data/audio_to_audio.py
titu1994: The idea was to make the function itself split into seperate components, each with private methods inside of the class, which can be overriden. Having a class which calls a monolithic function defeats the purpose of extensible code

titu1994 (changes requested):

nemo/collections/asr/data/audio_to_audio.py
```diff
@@ -0,0 +1,1119 @@
...
+            # Load each signal independently
+            async_signals = self.load_async_signals(example)
+            output.update(async_signals)
+
+        # Load embedding vector
+        if self.embedding_setup is not None:
+            embedding = self.load_embedding(example)
+            output.update(embedding)
+
+        if not output:
```
titu1994, line 304: Btw, does not output even work if output is a OrderedDict?

titu1994: Maybe len(output) == 0?

titu1994 (dismissed):

tango4j (dismissed):

tango4j (approved):
The changes look good to me

PR #5218, opened 2022-10-20
[TTS] Fastpitch energy condition and refactoring

subhankar-ghosh:
# What does this PR do ?

Adds energy conditioning to FastPitch

TTS

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


subhankar-ghosh:
carry forward from : https://github.com/NVIDIA/NeMo/pull/5176

nemo/collections/tts/models/fastpitch.py
```diff
@@ -404,13 +438,18 @@ def training_step(self, batch, batch_idx):
...
+        attn_prior, durs, speaker, energy = None, None, None, None
         if self.learn_alignment:
             if self.ds_class_name == "TTSDataset":
-                if SpeakerID in self._train_dl.dataset.sup_data_types_set:
-                    audio, audio_lens, text, text_lens, attn_prior, pitch, _, speaker = batch
-                else:
-                    audio, audio_lens, text, text_lens, attn_prior, pitch, _ = batch
+                batch_dict = process_batch(batch, self._train_dl.dataset.sup_data_types_set)
+                audio = batch_dict.get("audio")
+                audio_lens = batch_dict.get("audio_lens")
```
rlangman, line 446: Would it be better to represent our batch object as an @dataclass (which is already used a lot in NeMo) instead of a dict?

It is fine if we leave the decision on how to pass these long argument lists around for a later PR.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -267,6 +282,25 @@ def forward(
 
         enc_out = enc_out + pitch_emb.transpose(1, 2)
 
+        # Predict energy
+        if self.energy_predictor is not None:
+            energy_pred = self.energy_predictor(enc_out, enc_mask).squeeze(-1)
+
+            if energy is not None:
+                # Average energy over characters
+                energy_tgt = average_pitch(energy.unsqueeze(1), attn_hard_dur)
```
rlangman, line 291: Would supporting the situation when learn_alignment is false only required changing "attn_hard_dur" to "durs"? Or are these variables in different formats?

rlangman (approved):
Looks fine to me.

nemo/collections/tts/losses/fastpitchloss.py
```diff
@@ -120,6 +120,37 @@ def forward(self, pitch_predicted, pitch_tgt, len):
...
+        }
+
+    @property
+    def output_types(self):
+        return {
+            "loss": NeuralType(elements_type=LossType()),
+        }
+
+    @typecheck()
+    def forward(self, energy_predicted, energy_tgt, len):
```
XuesongYang, line 143: nit: pls rename `len` cause it is a build-in func name.

nemo/collections/tts/losses/fastpitchloss.py
```diff
@@ -120,6 +120,37 @@ def forward(self, pitch_predicted, pitch_tgt, len):
...
+        if energy_tgt is None:
+            return 0.0
```
XuesongYang, line 145: nit: `forward` function is expected to return a Tensor. Should we enforce a Tensor here as well? @ryanleary @redoctopus ?

nemo/collections/tts/models/fastpitch.py
```diff
@@ -329,24 +341,43 @@ def generate_spectrogram(
...
+                text = batch_dict.get("text")
+                text_lens = batch_dict.get("text_lens")
+                attn_prior = batch_dict.get("align_prior_matrix", None)
+                pitch = batch_dict.get("pitch", None)
+                energy = batch_dict.get("energy", None)
+                speaker = batch_dict.get("speaker_id", None)
             else:
                 raise ValueError(f"Unknown vocab class: {self.vocab.__class__.__name__}")
         else:
             audio, audio_lens, text, text_lens, durs, pitch, speaker = batch
```
XuesongYang, line 359: I wonder if we can make the logic more concise like below, @rlangman could you pls confirm?

``` python
        ...
        batch_dict = process_batch(batch, self._train_dl.dataset.sup_data_types_set)
        audio = batch_dict.get("audio")
        audio_lens = batch_dict.get("audio_lens")
        text = batch_dict.get("text")
        text_lens = batch_dict.get("text_lens")
        attn_prior = batch_dict.get("align_prior_matrix", None)
        pitch = batch_dict.get("pitch", None)
        energy = batch_dict.get("energy", None)
        speaker = batch_dict.get("speaker_id", None)
        dur = batch_dict.get("dur", None)

        if self.learn_alignment:
            assert self.ds_class_name == "TTSDataset", f"Unknown vocab class: {self.vocab.__class__.__name__}"
```

nemo/collections/tts/models/fastpitch.py
```diff
@@ -404,13 +438,18 @@ def training_step(self, batch, batch_idx):
...
+                audio_lens = batch_dict.get("audio_lens")
+                text = batch_dict.get("text")
+                text_lens = batch_dict.get("text_lens")
+                attn_prior = batch_dict.get("align_prior_matrix", None)
+                pitch = batch_dict.get("pitch", None)
+                energy = batch_dict.get("energy", None)
+                speaker = batch_dict.get("speaker_id", None)
             else:
                 raise ValueError(f"Unknown vocab class: {self.vocab.__class__.__name__}")
         else:
```
XuesongYang, line 455: Same question as the above. I wonder if we can make the logic more concise like below, @rlangman could you pls confirm?

``` python
        ...
        batch_dict = process_batch(batch, self._train_dl.dataset.sup_data_types_set)
        audio = batch_dict.get("audio")
        audio_lens = batch_dict.get("audio_lens")
        text = batch_dict.get("text")
        text_lens = batch_dict.get("text_lens")
        attn_prior = batch_dict.get("align_prior_matrix", None)
        pitch = batch_dict.get("pitch", None)
        energy = batch_dict.get("energy", None)
        speaker = batch_dict.get("speaker_id", None)
        dur = batch_dict.get("dur", None)

        if self.learn_alignment:
            assert self.ds_class_name == "TTSDataset", f"Unknown vocab class: {self.vocab.__class__.__name__}"
```

XuesongYang (changes requested):

nemo/collections/tts/models/fastpitch.py
```diff
@@ -433,16 +473,19 @@ def validation_step(self, batch, batch_idx):
...
 
-        mel_loss = self.mel_loss(spect_predicted=mels_pred, spect_tgt=mels)
-        dur_loss = self.duration_loss(log_durs_predicted=log_durs_pred, durs_tgt=durs, len=text_lens)
-        pitch_loss = self.pitch_loss(pitch_predicted=pitch_pred, pitch_tgt=pitch, len=text_lens)
-        loss = mel_loss + dur_loss + pitch_loss
+        mel_loss = self.mel_loss_fn(spect_predicted=mels_pred, spect_tgt=mels)
+        dur_loss = self.duration_loss_fn(log_durs_predicted=log_durs_pred, durs_tgt=durs, len=text_lens)
+        pitch_loss = self.pitch_loss_fn(pitch_predicted=pitch_pred, pitch_tgt=pitch, len=text_lens)
+        energy_loss = 0.0
+        energy_loss = self.energy_loss_fn(energy_predicted=energy_pred, energy_tgt=energy_tgt, len=text_lens)
```
XuesongYang, line 480: do we really need to initialize `energy_loss`? It seems to me a redundant line in Line 479.

XuesongYang (changes requested):

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Sure, we can but the assert statement has to go before we start extracting elements from the dict. Because if `ds_class_name != "TTSDataset"` then the extraction code would throw error and we will not exit gracefully.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -142,16 +145,22 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         speaker_emb_condition_prosody = cfg.get("speaker_emb_condition_prosody", False)
         speaker_emb_condition_decoder = cfg.get("speaker_emb_condition_decoder", False)
         speaker_emb_condition_aligner = cfg.get("speaker_emb_condition_aligner", False)
+        energy_embedding_kernel_size = cfg.get("energy_embedding_kernel_size", 0)
+        energy_predictor = None
+        if self._cfg.get("energy_predictor", None) is not None:
+            energy_predictor = instantiate(self._cfg.energy_predictor)
```
XuesongYang, line 151: copy and paste my comments from the closed PR:https://github.com/NVIDIA/NeMo/pull/5176

> But energy_predictor needs to be initialized to None. Because energy_predictor in line 150 is variable for the Neural module for predicting energy. Whereas, self._cfg.get("energy_predictor", None) will give us just the config, it is only when we call instantiate in line 152 that energy_predictor gets it's value (the neural module).

It is still not necessary because if the input cfg is `None` then `instantiate` func will directly return `None` (ref: https://github.com/facebookresearch/hydra/blob/main/hydra/_internal/instantiate/_instantiate2.py#L178-L179). So in this case, you only need,

```python
energy_predictor = instantiate(self._cfg.get("energy_predictor", None))
```

XuesongYang (changes requested):

XuesongYang:
Let's hold off on merging before figuring out some requests. Thanks.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Made the change. Although it added one more if...else. I think we should discuss about if we want to support not learning aligner anymore.

nemo/collections/tts/losses/fastpitchloss.py
subhankar-ghosh: Done.

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Done.

subhankar-ghosh: Done.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Done.

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Done.

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: Done.

nemo/collections/tts/models/fastpitch.py
XuesongYang: This is a great idea if you mean the `@dataclass` decor from `from dataclasses import dataclass`. Could you please create a ticket and we will address it later?

subhankar-ghosh:
Does the changes look good, can we merge now @XuesongYang ? Or are we waiting for something else https://github.com/NVIDIA/NeMo/pull/5218#issuecomment-1287355229

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -261,12 +276,37 @@ def forward(
             if self.learn_alignment and pitch.shape[-1] != pitch_predicted.shape[-1]:
                 # Pitch during training is per spectrogram frame, but during inference, it should be per character
                 pitch = average_pitch(pitch.unsqueeze(1), attn_hard_dur).squeeze(1)
+            elif not self.learn_alignment:
+                # If alignment is not learnt attn_hard_dur is None, hence durs_predicted is used in average_pitch
+                pitch = average_pitch(pitch.unsqueeze(1), durs_predicted).squeeze(1)
```
rlangman, line 281: The beginning of the forward method has:

```
if not self.learn_alignment and self.training:
  assert durs is not None
  assert pitch is not None
```
This implies that if alignment is not being learned, then the ground truth duration information will be passed into forward(). This ground truth "durs" is then what would have to be used for all of the training inference. Using the predicted durs would make sense if self.training is not set, though they can use the infer() function for that.

Trying to reverse engineer the intent for the old alignment code and support it without tests is more trouble than it is worth.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -261,12 +276,37 @@ def forward(
...
+        # Predict energy
+        if self.energy_predictor is not None:
+            energy_pred = self.energy_predictor(enc_out, enc_mask).squeeze(-1)
+
+            if energy is not None:
+                # Average energy over characters
+                if self.learn_alignment:
+                    energy_tgt = average_pitch(energy.unsqueeze(1), attn_hard_dur)
+                else:
+                    energy_tgt = average_pitch(energy.unsqueeze(1), durs_predicted)
```
rlangman, line 297: "average_pitch" should be renamed to..."average_features"?

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -261,12 +276,37 @@ def forward(
             if self.learn_alignment and pitch.shape[-1] != pitch_predicted.shape[-1]:
```
rlangman, line 276: Whatever we do between the pitch and energy should be consistent. In which case I would remove the shape check here. Ideally the pitch and energy processing could just use the same helper method.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -261,12 +276,37 @@ def forward(
...
+        if self.energy_predictor is not None:
+            energy_pred = self.energy_predictor(enc_out, enc_mask).squeeze(-1)
+
+            if energy is not None:
+                # Average energy over characters
+                if self.learn_alignment:
+                    energy_tgt = average_pitch(energy.unsqueeze(1), attn_hard_dur)
+                else:
+                    energy_tgt = average_pitch(energy.unsqueeze(1), durs_predicted)
+                energy_tgt = torch.log(1.0 + energy_tgt)
```
rlangman, line 298: I think we should move the log compression into the data loader, where we do all of our other feature preprocessing/normalization. Otherwise there will be a conflict when we later refactor the pitch and energy to use the same code flow.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Ideally we should, but I am sceptical removing a condition in the pitch workflow in an energy related PR (which makes it difficult to track changes), and also that would take some more testing for pitch. Can we take notes of the general refactoring and keep them for a different dedicated PR for refactoring?

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Do you mean we log(energy_tgt) in the dataloader and hence before averaging? But won't that lead to mathematical inconsistencies? like log(a+b) != log(a) + log(b), I think we will run into such scenarios.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: I am not sure. As I mentioned before, I think we should have a separate dedicated PR for dealing with `self.learn_alignment`

nemo/collections/tts/modules/fastpitch.py
rlangman: Yes, the inconsistency is why I bring it up now to avoid backwards incompatibility.

I doubt taking the log before or after the averaging will make a difference in the final model accuracy. Which is why I suggest doing it beforehand as its more flexible and lets us add in other processing/configurations (eg. mean/var normalization), if needed in the future.

nemo/collections/tts/modules/fastpitch.py
subhankar-ghosh: Done

nemo/collections/tts/modules/fastpitch.py
rlangman: I am OK with leaving it to another PR. If we do that, should we revert the pitch change to be safe?

nemo/collections/tts/modules/fastpitch.py
rlangman: Fine by me

XuesongYang (dismissed):
LGTM. Thanks for your PR. Let's discuss some comments offline or in separate PRs if they are necessary.

rlangman (approved):

nemo/collections/tts/models/fastpitch.py
XuesongYang: I created a ticket already. So we can track there.

XuesongYang (dismissed):

XuesongYang (approved):

PR #5223, opened 2022-10-22
fix num_workers to 1 (temp fix)

nithinraok:

# What does this PR do ?

Fix VAD struck issue temporarily with num_workers=1

**Collection**: [SD, VAD]

# Changelog 
- config update

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
+++ b/examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
@@ -6,7 +6,7 @@
 # {"audio_filepath": "/path/to/audio_file", "offset": 0, "duration": null, "label": "infer", "text": "-", "num_speakers": null, "rttm_filepath": "/path/to/rttm/file", "uem_filepath": "/path/to/uem/file"}
 name: &name "ClusterDiarizer"
 
-num_workers: 4
+num_workers: 1
 sample_rate: 16000
 batch_size: 64
 
@@ -24,7 +24,7 @@ diarizer:
     parameters: # Tuned parameters for CH109 (using the 11 multi-speaker sessions as dev set) 
       window_length_in_sec: 0.63  # Window length in sec for VAD context input 
       shift_length_in_sec: 0.01 # Shift length in sec for generate frame level VAD prediction
-      smoothing: False # False or type of smoothing method (eg: median)
+      smoothing: median # False or type of smoothing method (eg: median)
       overlap: 0.5 # Overlap ratio for overlapped mean/median smoothing filter
       onset: 0.9 # Onset threshold for detecting the beginning and end of a speech 
       offset: 0.5 # Offset threshold for detecting the end of a speech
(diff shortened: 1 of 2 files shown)


examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
```diff
@@ -6,7 +6,7 @@
 # {"audio_filepath": "/path/to/audio_file", "offset": 0, "duration": null, "label": "infer", "text": "-", "num_speakers": null, "rttm_filepath": "/path/to/rttm/file", "uem_filepath": "/path/to/uem/file"}
 name: &name "ClusterDiarizer"
 
-num_workers: 4
+num_workers: 1
```
fayejf, line 9: Could you add a comment say they can use larger values ?

fayejf (changes requested):

examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
```diff
@@ -24,7 +24,7 @@ diarizer:
     parameters: # Tuned parameters for CH109 (using the 11 multi-speaker sessions as dev set) 
       window_length_in_sec: 0.63  # Window length in sec for VAD context input 
       shift_length_in_sec: 0.01 # Shift length in sec for generate frame level VAD prediction
-      smoothing: False # False or type of smoothing method (eg: median)
+      smoothing: median # False or type of smoothing method (eg: median)
```
fayejf, line 27: temp fix here is not a good idea in my mind. Could we quick patch it in code?

examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
nithinraok: But that doesn't work now. So its better to add when we fix multiprocessing issue

examples/speaker_tasks/diarization/conf/inference/diar_infer_meeting.yaml
nithinraok: Updated

nemo/collections/asr/parts/utils/diarization_utils.py
```diff
@@ -237,7 +237,10 @@ def run_diarization(
         diar_model = ClusteringDiarizer(cfg=diar_model_config)
         score = diar_model.diarize()
         if diar_model_config.diarizer.vad.model_path is not None and not diar_model_config.diarizer.oracle_vad:
-            self.get_frame_level_VAD(vad_processing_dir=diar_model.vad_pred_dir)
+            self.get_frame_level_VAD(
+                vad_processing_dir=diar_model.vad_pred_dir,
+                smoothing_type=diar_model_config.diarizer.vad.parameters.smoothing,
```
fayejf, line 242: I'm worried the smoothing does not present in  diar_model_config.diarizer.vad and looks like we do not have a check/default value in diar_model.
Approving this but I think we should have general default values for diar_model_config to avoid dict access error.  @tango4j

fayejf (approved):
The temp fix looks good to me. Let's discuss the config check later

PR #5224, opened 2022-10-22
Support for inserting additional subsampling in conformer encoder

shan18:
# What does this PR do?

Added support for specifying an additional subsampling layer in conformer encoder. The user can specify the type and the position of the subsampling layer.

**Collection**: asr

# Changelog 
- New `SubsamplingReductionModule` class in `nemo/collections/asr/parts/submodules/subsampling.py`
- Updated the `ConformerEncoder` in `nemo/collections/asr/modules/conformer_encoder.py` to accept the new reduction parameters for adding a new subsampling layer in the desired position.
- Updated the _transducer-related configs_ with reduction parameters.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

Everyone in the NeMo ASR team.


docs/source/asr/speech_intent_slot/configs.rst
```diff
@@ -88,6 +88,11 @@ An example config for the model can be:
       subsampling_factor: 4 # must be power of 2
       subsampling_conv_channels: -1 # -1 sets it to d_model
 
+      # Reduction params
```
VahidooX, line 91: Maybe adding a comment what is the purpose of reduction?

docs/source/asr/speech_intent_slot/configs.rst
```diff
@@ -88,6 +88,11 @@ An example config for the model can be:
...
+      reduction: null # pooling, subsampling, or null
+      reduction_position: -2 # Encoder block index or -1 for subsampling at the end of encoder
```
VahidooX, line 93: What does -2 does?
How about using null to disable it?

docs/source/asr/speech_intent_slot/configs.rst
```diff
@@ -88,6 +88,11 @@ An example config for the model can be:
       subsampling_factor: 4 # must be power of 2
       subsampling_conv_channels: -1 # -1 sets it to d_model
 
+      # Reduction params
+      reduction: null # pooling, subsampling, or null
```
VahidooX, line 92: What would be the subsampling approach?

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -60,6 +64,12 @@ class ConformerEncoder(NeuralModule, StreamingEncoder, Exportable):
             Defaults to 4.
         subsampling_conv_channels (int): the size of the convolutions in the subsampling module
             Defaults to -1 which would set it to d_model.
+        reduction (str, Optional): the method of reduction, choices=['pooling', 'subsampling']. If no value
```
VahidooX, line 67: What do you mean by not passing any value? do you mean None?

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -252,6 +266,16 @@ def __init__(
         else:
             self.pre_encode = nn.Linear(feat_in, d_model)
 
+        # Reduction
+        if reduction and reduction_factor > 1:
+            self.reduction_subsampling = SubsamplingReductionModule(
+                reduction=reduction, d_model=d_model, reduction_factor=reduction_factor,
+            )
+            self.reduction_position = reduction_position
```
VahidooX, line 274: Let's check that reduction_position has a valid value and raise an error otherwise.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -384,6 +408,46 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
...
+            audio_signal, length = self.reduction_subsampling(x=audio_signal, lengths=length)
+
+        audio_signal = torch.transpose(audio_signal, 1, 2)
+
+        if cache_last_channel is not None:
+            return audio_signal, length, cache_last_channel_next, cache_last_time_next
+        else:
+            return audio_signal, length
+
+    def _create_masks(self, audio_signal, max_audio_length, length, cache_last_channel=None):
```
VahidooX, line 450: First part of the code related to caching should not be inside create_masks. It should just creates the maskings.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -384,6 +408,46 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
         max_audio_length = audio_signal.size(1)
 
         # Create the self-attention and padding masks
+        audio_signal, max_audio_length, pad_mask, att_mask, pos_emb, cache_last_channel_next = self._create_masks(
```
VahidooX, line 411: What is the benefit of creating this method of _create_masks? Is it going to be used anywhere else?

VahidooX (changes requested):

docs/source/asr/speech_intent_slot/configs.rst
bmwshop: Agree with @VahidooX ; it's not clear what -2 means.

docs/source/asr/speech_intent_slot/configs.rst
bmwshop: +1 on explaining the approaches. Suggest using None instead of null

docs/source/asr/speech_intent_slot/configs.rst
shan18: - -2 means no reduction is applied.
- -1 means that reduction is applied at the end of encoder (after the output projection layer)
- Integers 0-17 specify the block index after which the subsampling should be applied.

Having null instead for this value also works. I'll change it.

As for checking the bounds, should I add an assert statement to ensure the value is either None or remains between [-1, len(self.layers)]?

docs/source/asr/speech_intent_slot/configs.rst
shan18: The subsampling approach is the initial 4x subsampling that we apply to the encoder. It was there already.
What I added are the reduction parameters which add another 2x subsampling layer in the specified position.

Regarding the null values, across all the configs in NeMo, I have seen the values null specified in yaml files. When we use these files in python, the null values automatically get converted to None type.

nemo/collections/asr/modules/conformer_encoder.py
shan18: If the reduction parameter is not specified in the config then no such parameter is passed while creating an object of this class so the values are by default None.

nemo/collections/asr/modules/conformer_encoder.py
shan18: does an assert statement to ensure that the value remains between `[-1, len(self.layers)]` sound good?

nemo/collections/asr/modules/conformer_encoder.py
shan18: Yes, if we apply subsampling in between the encoder blocks then we need to calculate the masks again because the length of the audio signal will be reduced. So for this purpose, I put it in a separate function.

See line 434. I call this function there again.

docs/source/asr/speech_intent_slot/configs.rst
VahidooX: Yes, that is the bound to get checked.

docs/source/asr/speech_intent_slot/configs.rst
VahidooX: Looks like the reduction just supports striding. So I suggest to use from ["pooling", "striding"] and raise a "not supported" error if something else is passed.

nemo/collections/asr/modules/conformer_encoder.py
VahidooX: Yes, sounds good.

docs/source/asr/speech_intent_slot/configs.rst
shan18: Added

docs/source/asr/speech_intent_slot/configs.rst
shan18: Added a bound check in the new commit

docs/source/asr/speech_intent_slot/configs.rst
shan18: Have added that check in `subsampling.py`, inside the `SubsamplingReductionModule` class

nemo/collections/asr/modules/conformer_encoder.py
shan18: Pushed the code with caching outside create_masks

VahidooX (approved):

VahidooX (approved):

bmwshop (approved):

PR #5231, opened 2022-10-24
Update perturb.py

stevehuang52:
Add checking for channels mismatch for audio and noise data, throw an exception if they have different number of channels. Also fixed `perturb_with_foreground_noise` as done in `perturb_with_input_noise` 

Signed-off-by: He Huang (Steve) <105218074+stevehuang52@users.noreply.github.com>


stevehuang52:
> Could we add a test for this change, e.g., to verify it passes when channels count matches and fails when there's a mismatch?

Yes we could, but for this simple logic I feel like it's not that necessary? A more necessary test might be to make sure that noise is added to the data in the expected way, so that we can avoid the situation that noise is only added to a few frames as in Sven's case.

stevehuang52:
> It seems CI is failing with the new test.

There was a typo in the code, just fixed it.

PR #5235, opened 2022-10-24
Add NGC links for Aligner and FastPitch

redoctopus:

# What does this PR do ?

Add option to load IPA-based checkpoints for FastPitch and the TTS Aligner.

**Collection**: TTS

# Changelog 
- Added one entry each to the Aligner model and FastPitch model NGC model loading.
- Did not add entries to the checkpoints table in the TTS documentation since they are new versions of existing model cards, and each entry in each checkpoint table is a single model card.

# Usage

```python
from nemo.collections.tts.models import AlignerModel, FastPitchModel

# IPA Aligner
aligner = AlignerModel.from_pretrained("tts_en_radtts_aligner_ipa")
# IPA FastPitch
fastpitch = FastPitchMode.from_pretrained("tts_en_fastpitch_ipa")
```


titu1994 (approved):

XuesongYang (changes requested):
Thanks. Could you also add this IPA aligner into NeMo doc and distinguish wit Arpabet? https://github.com/NVIDIA/NeMo/blob/main/docs/source/tts/data/ngc_models_aligner.csv

redoctopus:
> Thanks. Could you also add this IPA aligner into NeMo doc and distinguish wit Arpabet? https://github.com/NVIDIA/NeMo/blob/main/docs/source/tts/data/ngc_models_aligner.csv

I looked at that table earlier and wasn't quite sure how to distinguish it--they're on the same model card so all the fields would remain the same. Do you have a suggestion for how we could make it clearer in these tables?

XuesongYang:
> I looked at that table earlier and wasn't quite sure how to distinguish it--they're on the same model card so all the fields would remain the same. Do you have a suggestion for how we could make it clearer in these tables?

right, I remembered this. I saw you added a new entry in Versions section in the model card. Then We are good to merge. Let's keep this as an open discussion.

XuesongYang (approved):
LGTM.

PR #5274, opened 2022-10-28
Add SDP documentation

erastorgueva-nv:
Signed-off-by: Elena Rastorgueva <erastorgueva@nvidia.com>

# What does this PR do ?

Adds documentation to Speech Dataset Processor

tools/speech_dataset_processor

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [x] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


tools/speech_dataset_processor/README.md
```diff
@@ -1,7 +1,165 @@
...
+2. share the steps for processing a speech dataset. Sharing processing steps can be as easy as sharing a YAML file.
 
-This toolkit contains many of the most common speech dataset processing operations. To process a new dataset, you simply need to write a YAML file containing the parameters needed for dataset processing. It is also easy to add your own code for various speech dataset processing steps if needed.
+SDP's philosophy is to represent processing operations as 'processor' classes. Many common processing operations are provided, and it is easy to add your own. In some cases, all you will need to do to process a new dataset is simply to write a YAML file containing the parameters needed to process your dataset.
 
-TBD
+SDP is specifically intended for the use case when you have an existing dataset with the audio & text pairs already specified in some form, and you wish to create a JSON manifest suitable for use with NeMo. SDP allows for intermediate cleaning and filtering steps which involve amending the 'ground truth' `"text"` or dropping utterances which are deemed to be too inaccurate for training on.
+
+# Overview of how SDP processes a dataset
```
Kipok, line 11: I think that there are currently too many details in this and follow-up sections. While it's probably ok to keep it as is, I would suggest that we refactor the documentation in more modular pieces that introduce only the information reader needs at the current stage and not too many details. As an example - in this section, which is an "overview" you're telling too many details that I don't think I need to know for the overview, such as main.py calls run_processors.py or that classes are created with `hydra.utils.instantiate` method.

What I would propose is to keep the readme simple and straightforward. E.g., you can have an overview here, but keep it high-level, just mention that we do operations on the data with each processor, and they are easy to modify/add/combine by users. I think it will also be useful to either directly include an example of the config file in the readme (maybe not full, but schematic) or reference included config, so that users know where to look for it. You can maybe also add some description on what kind of high-level operations we support (such as downloading/processing from such and such datasets, running inference, filtering text based on regexs, filtering based on ASR/trancription mismatches).

For the other material that you currently have, I would consider it to be more like a "developers guide". So, if you want to include it, I'd suggest creating other pages in the sphinx documentation and put it there. It's also going to be easier to manage, since instead of writing description of available classes, you can directly import the docs from the code and then the description will only stay in one place.

Another consideration is that there are likely going to be 2 types of users. One type - people who would just be creating new yaml files for their data (maybe only adding a single processor for initial processing or even not adding that, if using new language for existing dataset). Those users don't need to know any details about internal structure of the code/api, but would be very interested in what kind of classes we already support and what parameters those classes have. And then there is another type of people who would want to implement new  processors for some cases. Those people would need to know the details of what base class to inherit from or what kind of return value should be in specific functions. I guess in many cases users will start in the first category and then gradually move to the second one if they require different operations. But the general advise stays the same - I think we should have separate sections, one with description of "what's available already and how to use it" and another with "how you extend the code to add new things".

Sorry for too many words and please don't consider the above suggestions blocking for the merge. We can still merge the current version and then work on improving it later if you think it makes sense.

tools/speech_dataset_processor/sdp/processors/modify_manifest/modify_manifest.py
```diff
@@ -54,10 +62,10 @@ def test(self):
                 )
 
     @abstractmethod
-    def _process_dataset_entry(self, data_entry):
+    def _process_dataset_entry(self, data_entry: Type[DataEntry]):
```
Kipok, line 65: As mentioned above, this is not really of type DataEntry (most often it's a simple dict, but can be anything in principle)

Kipok (approved):
Thanks @erastorgueva-nv, looks good to me!

PR #5298, opened 2022-11-01
[TTS] Replace IPA lambda arguments with locale string

rlangman:
# What does this PR do ?

Currently you cannot configure the IPATokenizer in .yaml config because there is no way to modify the lambda arguments. This PR replaces the lambda arguments with a locale string that maps to the correct behavior, which now just maps to English vs Non-English.

An alternative approach would be to replace the lambda arguments with a new class/interface, but we thought that was too complicated for now. And that having default configurations available that someone can control with just a language/locale string would be nice to have regardless.

**Collection**: [TTS]

# Changelog 
- Replace IPA tokenizer lambda arguments with locale string that maps to English vs non-English IPA functions.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
@@ -71,8 +71,7 @@ IPA_CHARACTER_SETS = {
 
 def validate_locale(locale):
     if locale not in SUPPORTED_LOCALES:
-        raise ValueError(f"Unsupported locale '{locale}'. "
-                         f"Supported locales {SUPPORTED_LOCALES}")
+        raise ValueError(f"Unsupported locale '{locale}'. " f"Supported locales {SUPPORTED_LOCALES}")
 
 
 def get_grapheme_character_set(locale):
@@ -88,8 +87,7 @@ def get_grapheme_character_set(locale):
 def get_ipa_character_set(locale):
     if locale not in IPA_CHARACTER_SETS:
         raise ValueError(
-            f"IPA character set not found for locale '{locale}'. "
-            f"Supported locales {IPA_CHARACTER_SETS.keys()}"
+            f"IPA character set not found for locale '{locale}'. " f"Supported locales {IPA_CHARACTER_SETS.keys()}"
         )
     char_set = set(IPA_CHARACTER_SETS[locale])
     return char_set
--- a/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
@@ -14,6 +14,9 @@
 
 
 # fmt: off
+
+SUPPORTED_LOCALES = ["en-US", "de-DE", "es-ES"]
+
 DEFAULT_PUNCTUATION = (
     ',', '.', '!', '?', '-',
     ':', ';', '/', '"', '(',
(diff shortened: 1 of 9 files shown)


nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -561,8 +551,11 @@ def __init__(
 
         if punct:
             if non_default_punct_list is not None:
-                self.PUNCT_LIST = non_default_punct_list
-            tokens.update(self.PUNCT_LIST)
+                self.punct_list = non_default_punct_list
+            else:
+                self.punct_list = get_ipa_punctuation_list(locale)
```
redoctopus, line 556: Do we have a check that the locale given is valid? We should make sure to catch if a user has typoed instead of silently returning a default punct list.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -521,12 +515,12 @@ def __init__(
         sep='|',  # To be able to distinguish between symbols
         add_blank_at=None,
         pad_with_space=False,
-        text_preprocessing_func=lambda text: english_text_preprocessing(text, lower=False),
     ):
         """General-purpose IPA-based tokenizer.
         Args:
             g2p: Grapheme to phoneme module, should be IPAG2P or some subclass thereof.
             punct: Whether to reserve grapheme for basic punctuation or not.
+            locale: Locale used to determine default text processing logic and punctuation.
```
redoctopus, line 523: We should have a list of supported locales either here or in the documentation.

nemo_text_processing/g2p/modules.py
```diff
@@ -280,13 +280,7 @@ def __init__(
         Args:
             phoneme_dict (str, Path, Dict): Path to file in CMUdict format or dictionary of CMUdict-like entries.
                 Must be given for IPA G2P. (Consider using scripts/tts_dataset_files/ipa_cmudict-0.7b_nv22.06.txt.)
-            word_tokenize_func: Function for tokenizing text to words.
-                It has to return List[Tuple[Union[str, List[str]], bool]] where every tuple denotes word
-                representation and flag whether to leave unchanged or not.
-                It is expected that unchangeable word representation will be represented as List[str], other
-                cases are represented as str.
-                It is useful to mark word as unchangeable which is already in phoneme representation.
-                Defaults to the English word tokenizer.
```
redoctopus, line 289: Worth moving this info down to the added section?

redoctopus (commented):
Thanks for implementing this!

nemo_text_processing/g2p/data/data_utils.py
```diff
@@ -146,6 +146,10 @@ def ipa_word_tokenize(text):
     return _word_tokenize(words)
 
 
+def ipa_text_preprocessing(text):
+    return text.lower()
```
XuesongYang, line 150: [Discussion] Do we need such strong requirements for any languages using IPA? Letter cases may lead to different pronunciations for German.

redoctopus: This is just for tokenization, right?

Either way, I'd err on the side of having all graphemes be uppercased due to IPA symbol collisions with letters (e.g. "j").

rlangman: It would make sense to leave capitalization in grapheme only training, but for training with phonemes/IPA it would not be worth the complexity unless you have a G2P system that can infer different pronunciations based on the capitalization.

Or if we do want to use it, we may need to rethink the convention of grapheme words being represented as all UPPERCASE, both in the tokenizer and in the pronunciation dictionaries.

XuesongYang: for German dict, I chose to keep the uppercase and the word surface mixes with upper and lower cases.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
rlangman: I added the supported locales to the error message if someone provides a wrong one.

Should I provide a link in the docstring to the line or class path (nemo.collections.common.tokenizers.text_to_speech.ipa_lexicon) where the locale list is defined, even if that risks going out of date?

The best approach would be for the classes to have a "list_supported_locales()" method which a string that describes exactly what each locale string refers to, but that is probably overkill right now.

nemo_text_processing/g2p/modules.py
rlangman: This is very similar to the comment in the data_utils._word_tokenize() method itself. Can we assume that anyone who wants to override the behavior will look at the original function as a reference?

Though it would make sense to me to add at least the first line (describing what the return value is) somewhere in the tokenizer.

redoctopus: Yeah, makes sense.

> Though it would make sense to me to add at least the first line (describing what the return value is) somewhere in the tokenizer.

This would be nice!

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
rlangman: Added

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
rlangman: Done

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -241,7 +242,7 @@ def __init__(
 
 class SpanishCharsTokenizer(BaseCharsTokenizer):
 
-    PUNCT_LIST = get_ipa_punctuation_list("es-ES")
+    PUNCT_LIST = get_ipa_punctuation_set("es-ES")
```
redoctopus, line 245: Should this (and the following `self.punct_list`s) be updated to indicate they are sets? Alternatively, change `get_ipa_punctuation_set` back to `list` and convert to list before returning.

rlangman: I think the safest thing is to have it return a sorted list.

The original PUNCT_LIST variables are actually tuples, which is a bit confusing.

redoctopus (approved):
LGTM, thanks!

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
```diff
@@ -62,31 +65,48 @@
         'ʊ', 'ʌ', 'ʒ', '̃', 'θ'
     )
 }
+
 # fmt: on
 
 
+def validate_locale(locale):
+    if locale not in SUPPORTED_LOCALES:
+        raise ValueError(f"Unsupported locale '{locale}'. " f"Supported locales {SUPPORTED_LOCALES}")
```
XuesongYang, line 74: nit: remove `" f"`?

nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
```diff
@@ -62,31 +65,48 @@
...
+        )
     char_set = set(GRAPHEME_CHARACTER_SETS[locale])
     return char_set
 
 
 def get_ipa_character_set(locale):
     if locale not in IPA_CHARACTER_SETS:
-        raise ValueError(f"IPA character set not found for locale {locale}")
+        raise ValueError(
+            f"IPA character set not found for locale '{locale}'. " f"Supported locales {IPA_CHARACTER_SETS.keys()}"
```
XuesongYang, line 90: nit: remove " f"?

nemo_text_processing/g2p/data/data_utils.py
XuesongYang: we can simplify that with the current solution now, and maybe we could iterate if we have more datapoints.

XuesongYang (approved):
Thanks for the PR! LGTM

PR #5304, opened 2022-11-01
[TTS] bugfix IPAG2P and refactor to remove duplicate process.

XuesongYang:

# What does this PR do ?

1. fixed a bug for IPAG2P: the returned self.phoneme_dict mismatches between the one obtained from [self.parse_as_cmu_dict](https://github.com/NVIDIA/NeMo/blob/44de3950f0bf2f3e85d7df1a962eb24674f4fa4e/nemo_text_processing/g2p/modules.py#L320) and the one directly updated from a [dict obj](https://github.com/NVIDIA/NeMo/blob/44de3950f0bf2f3e85d7df1a962eb24674f4fa4e/nemo_text_processing/g2p/modules.py#L351). The former will return something like `{..., "WIRE": [["ˈ", "w", "a", "ɪ", "ɚ"], ["ˈ", "w", "a", "ɪ", "ɹ"]], ...}`, while the latter will return like `{..., "WIRE": ["ˈ", "w", "a", "ɪ", "ɚ"], ...}`.
2. refactored to remove duplicate processes for the case when the phoneme dict input is a dict object and for the case when the phoneme dict is a file.
3. added type hints and docstrings for the whole class `IPAG2P`.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo_text_processing/g2p/modules.py
+++ b/nemo_text_processing/g2p/modules.py
@@ -252,7 +252,7 @@ class EnglishG2p(BaseG2p):
 
 
 @experimental
-class IpaG2p(BaseG2p):
+class IPAG2P(BaseG2p):
     # fmt: off
     STRESS_SYMBOLS = ["ˈ", "ˌ"]
     # Regex for roman characters, accented characters, and locale-agnostic numbers/digits
--- a/nemo_text_processing/g2p/modules.py
+++ b/nemo_text_processing/g2p/modules.py
@@ -13,13 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import abc
 import pathlib
 import random
 import re
 import time
+from abc import ABC, abstractmethod
 from collections import defaultdict
-from typing import Optional
+from typing import DefaultDict, List, Optional, Tuple
 
 import nltk
 import torch
@@ -30,7 +30,7 @@ from nemo.utils.decorators import experimental
 from nemo.utils.get_rank import is_global_rank_zero
 
 
-class BaseG2p(abc.ABC):
+class BaseG2p(ABC):
     def __init__(
         self,
         phoneme_dict=None,
@@ -50,7 +50,7 @@ class BaseG2p(abc.ABC):
         self.apply_to_oov_word = apply_to_oov_word
         self.mapping_file = mapping_file
 
-    @abc.abstractmethod
+    @abstractmethod
     def __call__(self, text: str) -> str:
         pass
 
@@ -192,7 +192,7 @@ class EnglishG2p(BaseG2p):
         if re.search(r"[a-zA-ZÀ-ÿ\d]", word) is None:
             return list(word), True
 
-        # heteronym
+        # heteronyms
         if self.heteronyms is not None and word in self.heteronyms:
             return word, True
 
@@ -252,7 +252,7 @@ class EnglishG2p(BaseG2p):
 
 
 @experimental
-class IPAG2P(BaseG2p):
+class IpaG2p(BaseG2p):
     # fmt: off
     STRESS_SYMBOLS = ["ˈ", "ˌ"]
     # Regex for roman characters, accented characters, and locale-agnostic numbers/digits
@@ -278,8 +278,9 @@ class IPAG2P(BaseG2p):
         Ignored words are left unchanged or passed through apply_to_oov_word for handling.
 
         Args:
-            phoneme_dict (str, Path, Dict): Path to file in CMUdict format or dictionary of CMUdict-like entries.
-                Must be given for IPA G2P. (Consider using scripts/tts_dataset_files/ipa_cmudict-0.7b_nv22.10.txt.)
+            phoneme_dict (str, Path, Dict): Path to file in CMUdict format or a IPA dict object with CMUdict-like entries.
+                a dictionary file example: scripts/tts_dataset_files/ipa_cmudict-0.7b_nv22.06.txt;
+                a dictionary object example: {..., "WIRE": [["ˈ", "w", "a", "ɪ", "ɚ"], ["ˈ", "w", "a", "ɪ", "ɹ"]], ...}
             word_tokenize_func: Function for tokenizing text to words.
                 It has to return List[Tuple[Union[str, List[str]], bool]] where every tuple denotes word
                 representation and flag whether to leave unchanged or not.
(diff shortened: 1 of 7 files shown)


nemo_text_processing/g2p/modules.py
```diff
@@ -315,40 +316,29 @@ def __init__(
...
-            self.phoneme_dict, self.symbols = self._parse_as_cmu_dict(phoneme_dict)
+            # load the dictionary file where there may exist a digit suffix after a word, which
+            # represents the pronunciation variant of that word.
+            phoneme_dict_obj = defaultdict(list)
+            _alt_re = re.compile(r'\([0-9]+\)')
+            with open(phoneme_dict, "r") as fdict:
+                for line in fdict:
+                    if len(line) and ('A' <= line[0] <= 'Z' or line[0] == "'"):
+                        parts = line.strip().split()
+                        assert len(parts) == 2, f"Wrong format for the entry: {line.strip()}."
```
redoctopus, line 329: I think this is equivalent to the previous implementation (which splits on the double space between an entry and its pronunciation) for the CMUdicts we have, but splitting on any whitespace is looser and that worries me a bit. We probably won't encounter an entry with a space in it, but is there any other case in which this might break something?

Should we keep it as a split on double space?

nemo_text_processing/g2p/modules.py
```diff
@@ -315,40 +316,29 @@ def __init__(
...
-                for k, vs, in self.phoneme_dict.items():
-                    updated_phoneme_dict[k.upper()] = vs
-                self.phoneme_dict = updated_phoneme_dict
+            phoneme_dict_obj = phoneme_dict
+
+        # verify if phoneme dict obj is empty
+        if phoneme_dict_obj:
+            self.phoneme_dict, self.symbols = self._parse_as_cmu_dict(phoneme_dict_obj)
+        else:
+            raise ValueError(f"Found empty entries in {phoneme_dict}.")
```
redoctopus, line 341: Would it be more accurate to say "phoneme_dict is empty" in this error message?

nemo_text_processing/g2p/modules.py
```diff
@@ -379,31 +369,48 @@ def _parse_file_by_lines(p):
         with open(p, 'r') as f:
             return [l.rstrip() for l in f.readlines()]
 
-    def _parse_as_cmu_dict(self, phoneme_dict_path):
-        """Loads IPA CMUdict, and generates a set of all valid symbols."""
+    def _parse_as_cmu_dict(self, phoneme_dict_obj: dict) -> Tuple[DefaultDict[str, List], set]:
```
redoctopus, line 372: We can rename this to remove the cmudict reference.

nemo_text_processing/g2p/modules.py
```diff
@@ -379,31 +369,48 @@ def _parse_file_by_lines(p):
...
+        """
+        parse a python dict object according to the decision on word cases and removal of lexical stress markers.
```
redoctopus, line 374: Nitpick: s/parse/Parse

nemo_text_processing/g2p/modules.py
```diff
@@ -252,7 +252,7 @@ def __call__(self, text):
 
 
 @experimental
-class IPAG2P(BaseG2p):
+class IpaG2p(BaseG2p):
```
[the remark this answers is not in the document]
XuesongYang: ok. I will revert it. But let's make a strict naming convention when reviewing PRs to avoid similar issues.

nemo_text_processing/g2p/modules.py
XuesongYang: Not really. `phoneme_dict` could be a `dict` or a file. it is fine to say that if input is a dict, but not good if the input is a file because the dict file could contain no entries but contain comments or license or something else. So it is not empty.

nemo_text_processing/g2p/modules.py
XuesongYang: common entries in a dictionary file usually look like the below. I was thinking to support any formats below (trying to keep away from CMUlike dict constaints) so I applied `.split()`, but added an assertion to make sure an equal functioning as the previous implementation. As shown in `_parse_as_cmu_dict`, new implementation supports any below formats already. The assertion actually can be removed. If keeping splitting on double space, then `m.split("  ")` becomes `['word_4', 'b c d']`. 
```
x = "word_1  bcd"
y = "word_2 bcd"
z = "word_3\tbcd"
m = "word_4  b c d"
n = "word_5\tb c d"
```

nemo_text_processing/g2p/modules.py
XuesongYang: agreed!

XuesongYang: will change as `_normalize_dict()`.

nemo_text_processing/g2p/modules.py
redoctopus: Then should we say "phoneme_dict contains no entries"? The current text reads to me like there are word entries without pronunciations rather than no entries at all.

nemo_text_processing/g2p/modules.py
redoctopus: I just remembered that the double-space requirement was a holdover from the ARPABET-based dictionary, where the phonemes were separated by single spaces. IPA is all mashed together so I think this should be fine (and it'll be good to support other formats), but I do like having the assertion there just to catch any accidental user error.

nemo_text_processing/g2p/modules.py
```diff
@@ -315,40 +316,29 @@ def __init__(
...
+            # represents the pronunciation variant of that word.
+            phoneme_dict_obj = defaultdict(list)
+            _alt_re = re.compile(r'\([0-9]+\)')
+            with open(phoneme_dict, "r") as fdict:
+                for line in fdict:
+                    if len(line) and ('A' <= line[0] <= 'Z' or line[0] == "'"):
+                        parts = line.strip().split()
+                        assert len(parts) == 2, f"Wrong format for the entry: {line.strip()}."
+                        word = re.sub(_alt_re, '', parts[0])
+                        phoneme_dict_obj[word].append(parts[1:])
```
redoctopus, line 331: `parts` should only have two entries as per the previous assertion--change to `parts[1]`

nemo_text_processing/g2p/modules.py
XuesongYang: Yes, that sounds much better! will revise.

nemo_text_processing/g2p/modules.py
XuesongYang: ok then, I will match the previous implementation with double whitespace for this PR and will make another PR to support the flexible entry formats above. 

again, for language support expansion purposes, it would be great to make the dict format flexible.

nemo_text_processing/g2p/modules.py
XuesongYang: sure. will revise.

nemo_text_processing/g2p/modules.py
redoctopus: Wait, before you change it back, I'm agreeing--please leave as-is. It should be fine with the assertion.

redoctopus (approved):
Discussed on Slack, will have a separate PR for extending support for different dict formats.

PR #5309, opened 2022-11-02
Fixes for Conformer-xl export

borisfom:
# What does this PR do ?
Fixes for Conformer-xl model export:
1) Fixes ONNXRT error: Type parameter (Tind) of Optype (Slice) bound to different types (tensor(int32) and tensor(int64) in node (Slice_63).
2) Fixes 'runtime-check' crash on ONNX > 2GB

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/parts/submodules/multi_head_attention.py
+++ b/nemo/collections/asr/parts/submodules/multi_head_attention.py
@@ -331,7 +331,6 @@ class RelPositionalEncoding(PositionalEncoding):
         # positive positions would be used for left positions and negative for right positions
         positions = torch.arange(length - 1, -length, -1, dtype=torch.float32, device=device).unsqueeze(1)
         self.create_pe(positions=positions)
-        self.center_pos = torch.tensor(self.pe.size(1) // 2 + 1, dtype=torch.int32, device=device)
 
     def forward(self, x, cache_len=0):
         """Compute positional encoding.
@@ -350,8 +349,9 @@ class RelPositionalEncoding(PositionalEncoding):
         # negative positions would be used for right and positive for left tokens
         # for input of length L, 2*L-1 positions are needed, positions from (L-1) to -(L-1)
         input_len = x.size(1) + cache_len
-        start_pos = self.center_pos - input_len
-        end_pos = self.center_pos + input_len - 1
+        center_pos = self.pe.size(1) // 2 + 1
+        start_pos = center_pos - input_len
+        end_pos = center_pos + input_len - 1
         pos_emb = self.pe[:, start_pos:end_pos]
         if self.dropout_emb:
             pos_emb = self.dropout_emb(pos_emb)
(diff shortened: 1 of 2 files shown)


nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -350,8 +349,9 @@ def forward(self, x, cache_len=0):
         # negative positions would be used for right and positive for left tokens
         # for input of length L, 2*L-1 positions are needed, positions from (L-1) to -(L-1)
         input_len = x.size(1) + cache_len
-        start_pos = self.center_pos - input_len
-        end_pos = self.center_pos + input_len - 1
+        center_pos = self.pe.size(1) // 2 + 1
```
VahidooX, line 352: Was center_pos causing any issue?

VahidooX (approved):

nemo/collections/asr/parts/submodules/multi_head_attention.py
titu1994: How about registering it as a buffer ?

nemo/utils/export_utils.py
```diff
@@ -186,12 +186,10 @@ def verify_runtime(model, output, input_examples, input_names, check_tolerance=0
...
         onnx.checker.check_model(onnx_model, full_check=True)
         return
-
+    del onnx_model
     onnx_session_opt = onnxruntime.SessionOptions()
     onnx_session_opt.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_ALL
-    sess = onnxruntime.InferenceSession(
-        onnx_model.SerializeToString(), sess_options=onnx_session_opt, providers=['CUDAExecutionProvider']
-    )
+    sess = onnxruntime.InferenceSession(output, sess_options=onnx_session_opt, providers=['CUDAExecutionProvider'])
```
titu1994, line 192: Hmm this works inside the export engine, but what about when you want to restore a onnx model ? You'll still need to pass SerializeToString() no ?

titu1994 (approved):
It looks good but what was the issue with center_pos?

nemo/utils/export_utils.py
borisfom: no, you cam always use a file. SerializeToString() is just not compatible with external weights when ONNX size is > 2GB.

titu1994: Oh interesting. I need to update the RNNT script to support this type of loading

nemo/collections/asr/parts/submodules/multi_head_attention.py
borisfom: The problem was that this constant should not have been a tensor - and being a tensor of int32, it also generated ONNXRT error that wanted all slice args be same int type (and used int64 for other calculated expr). 
registering it as a buffer does not make any sense - it's a scalar and should not be flipped to and from GPU, also it's fine for ONNX exporter to treat it as constant. If we want to have dynamic resize in ONNX - my fix is already fine at low cost of 2 scalar ops - but it would also require more changes on resize side.

PR #5353, opened 2022-11-08
Fix issues with PL 1.8

SeanNaren:
# What does this PR do ?

Attempts to fix all issues after many deprecations/lightning lite becoming the backbone for a lot of the underlying pytorch lightning code.

cc @titu1994 

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/utils/exp_manager.py
```diff
@@ -658,24 +657,6 @@ def get_git_diff():
         return "{}\n".format(err.output.decode("utf-8"))
 
 
-class LoggerList(_LoggerCollection):
```
SeanNaren, line 661: This is not available in Lightning any more (it was moved to the `_grave_yard` which is pretty aggressive). I've removed it for now, but I need to check what the disadvantage of this is.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -228,11 +228,10 @@ def tb_logger(self):
...
                 return None
             tb_logger = self.logger.experiment
-            if isinstance(self.logger, LoggerCollection):
-                for logger in self.logger:
-                    if isinstance(logger, TensorBoardLogger):
-                        tb_logger = logger.experiment
-                        break
+            for logger in self.trainer.loggers:
+                if isinstance(logger, TensorBoardLogger):
+                    tb_logger = logger.experiment
```
titu1994, line 233: I think they deprecated LoggerCollection - should we upgrade and stop using it too ? Maybe just pass list of callback to trainer ?

nemo/core/config/hydra_runner.py
```diff
@@ -67,6 +67,9 @@ def wrapper(cfg_passthrough: Optional[DictConfig] = None) -> Any:
                 # Set run.dir ONLY for ExpManager "compatibility" - to be removed.
                 overrides.append("hydra.run.dir=.")
 
+                # Set working directory to the job's output directory
+                overrides.append("hydra.job.chdir=True")
```
titu1994, line 71: Wait why do we want this ? We Don't want to chdir

nemo/utils/exp_manager.py
titu1994: That's painful, though I wonder what we needed Nemo name and version for

SeanNaren:
In this PR: https://github.com/Lightning-AI/lightning/pull/11617

Lightning changed behaviour on how we start the other sub-processes with Hydra. This conflicts with how our hydra runner operates, need to investigate further on a solution cc @titu1994

SeanNaren:
This PR should hopefully unblock the move to 1.8: https://github.com/Lightning-AI/lightning/pull/15737#pullrequestreview-1188782551

We'll just wait till it gets into a minor release version (which hopefully would be tomorrow).

nemo/utils/exp_manager.py
SeanNaren: From my look around, they were not being used

Jenkinsfile
```diff
@@ -225,9 +225,11 @@ pipeline {
         stage('Speech to Text EMA') {
           steps {
             sh 'python examples/asr/asr_ctc/speech_to_text_ctc.py \
+            --config-path="../conf/" --config-name="config" \
```
SeanNaren, line 228:
```suggestion
```

nemo/core/classes/exportable.py
```diff
@@ -128,7 +129,7 @@ def _export(
             # Set module mode
             with torch.onnx.select_model_mode_for_export(
                 self, training
-            ), torch.inference_mode(), torch.no_grad(), torch.jit.optimized_execution(True):
+            ), torch.inference_mode(), torch.no_grad(), torch.jit.optimized_execution(True), _jit_is_scripting():
```
titu1994, line 132: Hmm we don't use lightning during export, is this method safe ?

tests/hydra/test_hydra_runner.py
```diff
@@ -41,8 +41,6 @@ def test_config1(self):
         # Run the call as subprocess.
         subprocess.check_call(call, shell=True, stdout=sys.stdout, stderr=sys.stdout)
 
-        # Make sure that .hydra dir is not present.
-        assert not path.exists(f".hydra")
```
titu1994, line 45: Why is assert being removed ?

titu1994 (commented):
Overall, incredible work @SeanNaren ! Minor comments and needs one other reviewer since it's a core change

tests/hydra/test_hydra_runner.py
SeanNaren: Ah forgot to revert, have done thanks!

nemo/core/classes/exportable.py
SeanNaren: Would the tests have confirmed if this is fine? The underlying code is simple to what is happening:

In Lightning:
```python
@contextmanager
def _jit_is_scripting() -> Generator:
    """Workaround for https://github.com/pytorch/pytorch/issues/67146."""
    LightningModule._jit_is_scripting = True
    try:
        yield
    finally:
        LightningModule._jit_is_scripting = False
```

titu1994: Hmm i dont think it would be caught in tests, lets keep it for now

PR #5358, opened 2022-11-08
[TTS] Fixing RADTTS training - removing view buffer and fixing accuracy issue

borisfom:

A view to the bias was cached as a buffer - that led to cpu/gpu mismatch on backward pass when training - also the caching was wrong - removed. That led to script() breaking for export - had to change script routine to do trace on PartialConv first.
Training had NaNs when using partial convolutions in mixed precision mode - had to move from epsilon to masking to fit in float16 range without damping.
Reverting submodules.py back to original and re-doing script optimizations in a cleaner manner, fixed the issues.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/tts/modules/common.py
```diff
@@ -219,7 +219,7 @@ def conv_to_padded_tensor(self, context: Tensor, lens: Tensor) -> Tensor:
         ret = torch.nn.utils.rnn.pad_sequence(context_embedded, batch_first=True)
         return ret
 
-    @torch.jit.export
+    @torch.jit.ignore
     def masked_conv_to_sequence(self, context: Tensor, lens: Tensor, enforce_sorted: bool = False) -> PackedSequence:
```
titu1994, line 223: Ignore ? Mask calculation will be incorrect then in export mode

nemo/collections/tts/modules/common.py
```diff
@@ -232,24 +232,35 @@ def masked_conv_to_sequence(self, context: Tensor, lens: Tensor, enforce_sorted:
...
-                context = self.dropout(F.relu(conv(context)))
-            context = context.transpose(1, 2)
-            context, _ = self.bilstm(context)
-        else:
-            # borisf : does not match ADLR (values, lengths)
-            # seq = self.masked_conv_to_sequence(context, lens, enforce_sorted=False)
-            # borisf : does match ADLR
-            seq = self.conv_to_sequence(context, lens, enforce_sorted=False)
-            context, _ = self.lstm_sequence(seq)
+    def forward(self, context: Tensor, lens: Tensor) -> Tensor:
```
titu1994, line 235: Let's is optional apparently in types, maybe TTS folks need careful review of this pr - @redoctopus @XuesongYang

titu1994: What about making it non optional in forward_for_export()

nemo/collections/tts/modules/common.py
```diff
@@ -232,24 +232,35 @@ def masked_conv_to_sequence(self, context: Tensor, lens: Tensor, enforce_sorted:
...
+        # borisf : does match ADLR
+        seq = self.conv_to_sequence(context, lens, enforce_sorted=False)
+        context, _ = self.lstm_sequence(seq)
 
         if self.dense is not None:
             context = self.dense(context).permute(0, 2, 1)
 
         return context
 
+    def script(self):
```
titu1994, line 247: maybe should be _prepare_for_export()?

titu1994 (commented):
The script() function is a bit weird, the rest seem fine. Still needs a review from @redoctopus / @XuesongYang for the TTS part, many significant changes

nemo/collections/tts/modules/common.py
borisfom: This function is never used (see ADLR comment below). I do plan to make it work - but with current scripting changes, it breaks as it gets scripted anyway. Let me remove the qualifier altogether to avoid confusion.

nemo/collections/tts/modules/common.py
borisfom: Why? This module is only used with lengths. I can restore Optional though, just in case.
Those changes were made in collaboration with @MikyasDesta and Kevin Shih  - RADTTS authors from ADLR, and tested on training and export.

redoctopus: If the `lens is None` handling is being removed, I think we should get rid of `Optional` just to minimize confusion. As long as we're certain that `lens` will always be passed in this is fine.

nemo/collections/tts/modules/common.py
titu1994: Ok

nemo/collections/tts/modules/common.py
```diff
@@ -232,24 +231,36 @@ def masked_conv_to_sequence(self, context: Tensor, lens: Tensor, enforce_sorted:
         )
         return seq
 
-    def forward(self, context: Tensor, lens: Optional[Tensor] = None) -> Tensor:
-        if lens is None:
-            for conv in self.convolutions:
-                context = self.dropout(F.relu(conv(context)))
-            context = context.transpose(1, 2)
-            context, _ = self.bilstm(context)
+    def forward(self, context: Tensor, in_lens: Optional[Tensor]) -> Tensor:
```
titu1994, line 234: typechecking probably expects "lens" not "in_lens", plus it should be Optional[Tensor]

nemo/collections/tts/modules/common.py
borisfom: I just added it back just in case.

nemo/collections/tts/modules/common.py
titu1994: Hmm its just a bit odd but if it works i guess that's fine

titu1994 (commented):
Looks better, tts folks need to review and sign off

nemo/collections/tts/modules/common.py
XuesongYang: > Why? This module is only used with lengths. I can restore Optional though, just in case. Those changes were made in collaboration with @MikyasDesta and Kevin Shih - RADTTS authors from ADLR, and tested on training and export.

a side note: there are not any unit tests and CI tests for RADTTS training and export in NeMo repo, which may have risks. I suggested adding tests when porting it from ADLR (https://github.com/NVIDIA/NeMo/pull/4538#pullrequestreview-1061119710), but they are still missing.

nemo/collections/tts/modules/common.py
borisfom: What type checking? This is not a Nemo Module. And it is Optional.

XuesongYang:
added one data point of this issue. when I added a CI for radtts on top of NeMo main, I hit error `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
`. See details errors in http://10.110.40.131:8080/blue/rest/organizations/jenkins/pipelines/NeMo/branches/PR-5386/runs/1/nodes/1235/log/?start=0

I am running my CI on top of this PR, and will report results here later.

tests/collections/tts/test_tts_exportables.py
```diff
@@ -15,8 +15,10 @@
 import tempfile
 
 import pytest
+from omegaconf import DictConfig, OmegaConf
```
XuesongYang, line 18: nit: `DictConfig` is an unused import. Please remove.

borisfom:
Now fixed all training issues. Also, improved speed by eliminating loop by batch element using masked instance norm.
Results are verified agains original ADLR.

tests/collections/tts/test_tts_exportables.py
borisfom: Done.

nemo/utils/export_utils.py
```diff
@@ -134,20 +134,6 @@ def to_onnxrt_input(ort_input_names, input_names, input_dict, input_list):
     return odict
 
 
-def verify_torchscript(model, output, input_examples, input_names, check_tolerance=0.01):
```
titu1994, line 137: Why is this being removed?

XuesongYang (changes requested):
Could you please cherry-pick my CI test commit (04b319cd21656cd51af59b2ccdb9d24517f8de21) for RADTTS? It would provide a safeguard for your changes.

XuesongYang:
please also rebase to the latest main and make sure all CI passed. Thanks!

examples/tts/conf/rad-tts_feature_pred.yaml
```diff
@@ -12,8 +12,8 @@ whitelist_path: "nemo_text_processing/text_normalization/en/data/whitelist/tts.t
 
 # these frame-wise values depend on pitch_fmin and pitch_fmax, you can get values
 # by running `scripts/dataset_processing/tts/extract_sup_data.py`
-pitch_mean: ???  # e.g. 212.35873413085938 for LJSpeech
-pitch_std:  ???  # e.g.  68.52806091308594 for LJSpeech
+pitch_mean: 145.0  # e.g. 212.35873413085938 for LJSpeech
+pitch_std:  30.0  # e.g.  68.52806091308594 for LJSpeech
```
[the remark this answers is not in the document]
borisfom: That led to errors in my test -- I guess I can reassign those.

tests/collections/tts/test_tts_exportables.py
```diff
@@ -50,7 +70,15 @@ def test_HifiGanModel_export_to_onnx(self, hifigan_model):
...
+        model = radtts_model.cuda()
+        with tempfile.TemporaryDirectory() as tmpdir:
+            filename = os.path.join(tmpdir, 'rad.ts')
+            model.export(output=filename, verbose=True, check_trace=True)
+
 
 if __name__ == "__main__":
     t = TestExportable()
-    t.test_FastPitchModel_export_to_onnx(fastpitch_model())
+    t.test_RadTTSModel_export_to_torchscript(radtts_model())
```
XuesongYang, line 84: i mean these three lines.

examples/tts/conf/rad-tts_feature_pred.yaml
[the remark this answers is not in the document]
XuesongYang: yes for sure. Could you pls explicitly assign values to these params?

nemo/collections/tts/modules/submodules.py
```diff
@@ -12,45 +12,83 @@
...
 
 
+def masked_instance_norm(
+    input: Tensor, mask: Tensor, weight: Tensor, bias: Tensor, momentum: float, eps: float = 1e-5,
+) -> Tensor:
+    r"""Applies Masked Instance Normalization for each channel in each data sample in a batch.
+
+    See :class:`~MaskedInstanceNorm1d` for details.
+    """
+    shape = input.shape
```
XuesongYang, line 30: remove line 30 since `shape` is not used anywhere.

examples/tts/radtts.py
```diff
@@ -21,6 +26,12 @@
 from nemo.utils.exp_manager import exp_manager
 
 
+def handle_pdb(sig, frame):
+    import pdb
+
+    pdb.Pdb().set_trace(frame)
```
titu1994, line 32: ?

titu1994 (approved):
Overall it looks good from export point of view, tts for final review

examples/tts/radtts.py
borisfom: Debugging hook in case the training gets hung.

XuesongYang, line 32: it would be better to convey clear thread error information to users instead of calling pdb here.

examples/tts/radtts.py
```diff
@@ -59,6 +70,7 @@ def prepare_model_weights(model, unfreeze_modules):
 
 @hydra_runner(config_path="conf", config_name="rad-tts_dec")
 def main(cfg):
+    signal.signal(signal.SIGUSR1, handle_pdb)
```
XuesongYang, line 73: is it possible to register a signal interruption with a handler that triggers exception rather than debugger?

borisfom: There is no point to do that in signal handler - it won't be able to trigger exception properly

examples/tts/radtts.py
borisfom: This is to debug hung process - there is no error information here.

nemo/collections/tts/modules/submodules.py
```diff
@@ -296,7 +305,7 @@ def __init__(self, c):
         self.conv = torch.nn.Conv1d(c, c, kernel_size=1, stride=1, padding=0, bias=False)
 
         # Sample a random orthonormal matrix to initialize weights
-        W = torch.linalg.qr(torch.FloatTensor(c, c).normal_())[0]
+        W = torch.qr(torch.FloatTensor(c, c).normal_())[0]
```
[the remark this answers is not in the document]
borisfom: Odd, I thought I was fixing it the other way ... Anyway - restored.

examples/tts/radtts.py
borisfom: Removed debugging hooks.

borisfom: Removed just in case.

examples/tts/radtts.py
borisfom: Removed just in case.

nemo/collections/tts/modules/radtts.py
```diff
@@ -60,11 +59,11 @@ def pad_energy_avg_and_f0(energy_avg, f0, max_out_len):
 
 def adjust_f0(f0, f0_mean, f0_std, vmask_bool):
     if f0_mean > 0.0:
-        f0_mu, f0_sigma = f0[vmask_bool].mean(), f0[vmask_bool].std()
-        f0[vmask_bool] = (f0[vmask_bool] - f0_mu) / f0_sigma
+        f0_sigma, f0_mu = torch.std_mean(f0[vmask_bool])
+        f0 = ((f0 - f0_mu) / f0_sigma).to(dtype=f0.dtype)
         f0_std = f0_std if f0_std > 0 else f0_sigma
```
XuesongYang, line 64: is it possible the `f0_sigma` tends to zero? If so, better to take care of zero-division.

XuesongYang: nit: not quite sure if we need to bother `.to(dtype=f0.dtype)` because all `f0_sigma` and `f0_mu` are computed directly from `f0`. So same dtype should be automatically cast.

nemo/collections/tts/modules/common.py
```diff
@@ -255,7 +216,7 @@ def getRadTTSEncoder(
     encoder_n_convolutions=3,
     encoder_embedding_dim=512,
     encoder_kernel_size=5,
-    norm_fn=nn.BatchNorm1d,
+    norm_fn=MaskedInstanceNorm1d,
```
XuesongYang, line 219: Line 215: we should follow the naming convention for functions over class names. `getRadTTSEncoder` -> `get_radtts_encoder`.

nemo/collections/tts/modules/radtts.py
```diff
@@ -183,9 +182,7 @@ def __init__(
         self.speaker_embedding = torch.nn.Embedding(n_speakers, self.n_speaker_dim)
```
XuesongYang, line 182: Line 146 and 147: `n_hidden` and `mel_encoder_n_hidden` are not used anymore. Please remove these two params.
```
        n_hidden,
        mel_encoder_n_hidden,
```

nemo/collections/tts/modules/radtts.py
```diff
@@ -199,6 +196,8 @@ def __init__(
         self.decoder_use_unvoiced_bias = kwargs['decoder_use_unvoiced_bias']
         self.ap_pred_log_f0 = ap_pred_log_f0
         self.ap_use_unvoiced_bias = kwargs['ap_use_unvoiced_bias']
+        self.prepared_for_export = False
```
XuesongYang, line 199: line 194: remove the comment.
```
        # self.length_regulator = LengthRegulator()
```

nemo/collections/tts/modules/radtts.py
```diff
@@ -356,7 +355,7 @@ def preprocess_context(self, context, speaker_vecs, out_lens, f0, energy_avg):
                     context_w_spkvec = torch.cat((context_w_spkvec, energy_avg), 1)
```
XuesongYang, line 355: Line 219 - 222: remove because there are not any ops calling `n_context_hidden`. Am I missing something?
```
                    n_context_hidden = n_f0_dims + n_energy_avg_dims + n_text_dim
                    n_context_hidden = n_context_hidden * n_group_size / 2
                    n_context_hidden = self.n_speaker_dim + n_context_hidden
                    n_context_hidden = int(n_context_hidden)
```

nemo/collections/tts/modules/radtts.py
```diff
@@ -613,7 +613,7 @@ def infer(
         spk_vec_text = self.encode_speaker(speaker_id_text)
         spk_vec_attributes = self.encode_speaker(speaker_id_attributes)
         txt_enc, _ = self.encode_text(text, in_lens)
-        print("txt_enc: ", txt_enc.shape)
+        # print("txt_enc: ", txt_enc.shape)
```
XuesongYang, line 616: remove the comment.

nemo/collections/tts/modules/radtts.py
```diff
@@ -667,29 +668,30 @@ def infer(
         # replication pad, because ungrouping with different group sizes
         # may lead to mismatched lengths
         # FIXME: use replication pad
-        print("V mask, energy_avg, f0, f0_bias: ", voiced_mask.shape, energy_avg.shape, f0.shape, f0_bias.shape)
         (energy_avg, f0) = pad_energy_avg_and_f0(energy_avg, f0, max_out_len)
-        print("V mask, energy_avg, f0, f0_bias: ", voiced_mask.shape, energy_avg.shape, f0.shape, f0_bias.shape)
+        # print("V mask, energy_avg, f0, f0_bias: ", voiced_mask.shape, energy_avg.shape, f0.shape, f0_bias.shape)
```
XuesongYang, line 672: remove the comment.

nemo/collections/tts/modules/radtts.py
```diff
@@ -667,29 +668,30 @@ def infer(
...
+            txt_enc_time_expanded, spk_vec, out_lens, (f0 + f0_bias) * voiced_mask, energy_avg
         )
 
         residual = torch.normal(txt_enc.new_zeros(batch_size, 80 * self.n_group_size, torch.max(n_groups))) * sigma
 
         # map from z sample to data
         num_steps_to_exit = len(self.exit_steps)
-        mel = residual[:, num_steps_to_exit * self.n_early_size :]
-        remaining_residual = residual[:, : num_steps_to_exit * self.n_early_size]
+        # mel = residual[:, num_steps_to_exit * self.n_early_size :]
```
XuesongYang, line 682: remove the comment.

nemo/collections/tts/modules/radtts.py
```diff
@@ -667,29 +668,30 @@ def infer(
...
 
         for i, flow_step in enumerate(reversed(self.flows)):
             curr_step = self.n_flows - i - 1
             mel = flow_step(mel, context_w_spkvec, inverse=True, seq_lens=n_groups)
             if num_steps_to_exit > 0 and curr_step == self.exit_steps[num_steps_to_exit - 1]:
                 # concatenate the next chunk of z
                 num_steps_to_exit = num_steps_to_exit - 1
-                residual_to_add = remaining_residual[:, num_steps_to_exit * self.n_early_size :]
-                remaining_residual = remaining_residual[:, : num_steps_to_exit * self.n_early_size]
+                # residual_to_add = remaining_residual[:, num_steps_to_exit * self.n_early_size :]
```
XuesongYang, line 691: remove the comment.

nemo/collections/tts/modules/radtts.py
```diff
@@ -464,10 +463,11 @@ def forward(
 
         f0_bias = 0
         # unvoiced bias forward pass
+        voiced_mask_bool = voiced_mask.bool()
```
XuesongYang, line 466: Line 385 and 400: functions `binarize_attention` and `get_first_order_features` should add decor `@staticmethod`, or perhaps there are better places to host such generic functions. 

Besides, `out_lens` in `get_first_order_features` is not used anywhere, please remove.

please also update docstring for the two functions.

XuesongYang: line 439: `p_voiced=None` is not used in the `forward()`. pls remove.

nemo/collections/tts/modules/radtts.py
```diff
@@ -464,10 +463,11 @@ def forward(
...
         if self.use_unvoiced_bias:
             f0_bias = self.unvoiced_bias_module(context.permute(0, 2, 1))
             f0_bias = -f0_bias[..., 0]
-            f0_bias = f0_bias * (~voiced_mask.bool()).float()
+            f0_bias.masked_fill_(voiced_mask_bool, 0.0)
 
         # mel decoder forward pass
         if 'dec' in self.include_modules:
```
XuesongYang, line 473: Line 479: `z_out = []` is duplicated. pls remove.

nemo/collections/tts/modules/radtts.py
```diff
@@ -635,8 +635,10 @@ def infer(
             if self.use_vpred_module:
                 # get logits
                 voiced_mask = self.v_pred_module.infer(None, txt_enc_time_expanded, spk_vec_attributes, lens=out_lens)
-                voiced_mask = torch.sigmoid(voiced_mask[:, 0]) > 0.5
-                voiced_mask = voiced_mask.float()
+                voiced_mask_bool = torch.sigmoid(voiced_mask[:, 0]) > 0.5
```
XuesongYang, line 638: lINE 629: remove 
```
        # print ("txt_enc_time_expanded, out_lens, dur: ", txt_enc_time_expanded.shape, out_lens, dur)
```

nemo/collections/tts/modules/radtts.py
```diff
@@ -667,29 +668,30 @@ def infer(
...
             mel = flow_step(mel, context_w_spkvec, inverse=True, seq_lens=n_groups)
             if num_steps_to_exit > 0 and curr_step == self.exit_steps[num_steps_to_exit - 1]:
                 # concatenate the next chunk of z
                 num_steps_to_exit = num_steps_to_exit - 1
-                residual_to_add = remaining_residual[:, num_steps_to_exit * self.n_early_size :]
-                remaining_residual = remaining_residual[:, : num_steps_to_exit * self.n_early_size]
+                # residual_to_add = remaining_residual[:, num_steps_to_exit * self.n_early_size :]
+                remaining_residual, residual_to_add = torch.tensor_split(
+                    remaining_residual, [num_steps_to_exit * self.n_early_size,], dim=1
+                )
```
XuesongYang, line 694: remove line 700: `# print ("mel=", mel.shape, "out_lens=", out_lens, "dur=", dur.shape)`

XuesongYang (changes requested):
Thanks for your updates. I tried my best to understand all changes, and added comments. Please have a look. 

In general, please remove all your debugging funcs or prints or loggings; please also consider to add type hints if time allows, but it is not necessary for this PR.

nemo/collections/tts/modules/common.py
```diff
@@ -44,7 +44,7 @@ def get_mask_from_lengths_and_val(lengths, val):
     max_len = val.shape[-1]
     ids = torch.arange(0, max_len, device=lengths.device)
     mask = ids < lengths.unsqueeze(1)
-    return mask.float()
+    return mask
```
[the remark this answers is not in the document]
borisfom: Well, max(lengths) is way more expensive than shape[-1]. Also, if for any reasons, max(lengths)!=val.shape[-1] - can be valid use case in inference - we'll end up with invalid mask.

nemo/collections/tts/modules/common.py
```diff
@@ -137,16 +137,11 @@ def lstm_sequence(self, seq: PackedSequence) -> Tuple[Tensor, Tensor]:
```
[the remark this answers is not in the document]
borisfom: I am not getting any warnings from Torch so I assume it does conversion internally. Adding item() risks warnings and errors on export side.

nemo/collections/tts/modules/radtts.py
borisfom: Well I had errors reported by exporter, that's why I had to add those casts. Division is fp32 when operands are fp16.

borisfom: f0_std was not a problem before, don't see why I should take care of it now.

nemo/collections/tts/modules/radtts.py
borisfom: Those cleanups are not in any way related to the changes I was doing - I would rather avoid any changes that are not necessary.

XuesongYang: ok. We can have a separate PR for clean up.

XuesongYang (approved):
Thank you for the fix and help cleaning up legacy. We are good to go!

PR #5383, opened 2022-11-10
[TTS] Add Spanish FastPitch training configs

rlangman:

# What does this PR do ?

Add configurations for training FastPitch with Spanish tokenizers

**Collection**: [TTS]

# Changelog 
- Add .yaml configs with Spanish grapheme and IPA tokenizers

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/tts/conf/es/fastpitch_align_44100.yaml
+++ b/examples/tts/conf/es/fastpitch_align_44100.yaml
@@ -1,0 +1,224 @@
+# This config contains the default values for training grapheme Spanish FastPitch model with aligner using
+# 44.1KHz sampling rate. If you want to train model on other dataset, you can change config values according
+# to your dataset. Most dataset-specific arguments are in the head of the config file, see below.
+
+name: FastPitch
+
+train_dataset: ???
+validation_datasets: ???
+sup_data_path: ???
+sup_data_types: ["align_prior_matrix", "pitch", "speaker_id"]
+
+# Default values from librosa.pyin
+pitch_fmin: 65.40639132514966
+pitch_fmax: 2093.004522404789
+
+# Stats (per frame), train (these values depend on pitch_fmin and pitch_fmax)
+pitch_mean: 212.35873413085938
+pitch_std: 68.52806091308594
+
+sample_rate: 44100
+n_mel_channels: 80
+n_window_size: 2048
+n_window_stride: 512
+n_fft: 2048
+lowfreq: 0
+highfreq: null
+window: hann
+
+model:
+  learn_alignment: true
+  bin_loss_warmup_epochs: 100
+
+  n_speakers: 1
+  max_token_duration: 75
+  symbols_embedding_dim: 384
+  pitch_embedding_kernel_size: 3
+  speaker_emb_condition_prosody: true
+  speaker_emb_condition_aligner: true
+
+  pitch_fmin: ${pitch_fmin}
+  pitch_fmax: ${pitch_fmax}
+
+  pitch_mean: ${pitch_mean}
+  pitch_std: ${pitch_std}
+
+  sample_rate: ${sample_rate}
+  n_mel_channels: ${n_mel_channels}
+  n_window_size: ${n_window_size}
+  n_window_stride: ${n_window_stride}
+  n_fft: ${n_fft}
+  lowfreq: ${lowfreq}
+  highfreq: ${highfreq}
+  window: ${window}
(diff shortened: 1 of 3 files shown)


examples/tts/conf/es/fastpitch_align_44100.yaml
```diff
@@ -0,0 +1,224 @@
+# This config contains the default values for training grapheme Spanish FastPitch model with aligner using
+# 44.1KHz sampling rate. If you want to train model on other dataset, you can change config values according
+# to your dataset. Most dataset-specific arguments are in the head of the config file, see below.
+
+name: FastPitch
+
+train_dataset: ???
+validation_datasets: ???
+sup_data_path: ???
+sup_data_types: ["align_prior_matrix", "pitch", "speaker_id"]
```
redoctopus, line 10: `speaker_id` is not in the default English FastPitch config. Do we expect Spanish models to mainly be multi-speaker?

examples/tts/conf/es/fastpitch_align_44100_ipa.yaml
```diff
@@ -0,0 +1,237 @@
+# This config contains the default values for training IPA Spanish FastPitch model with aligner using
+# 44.1KHz sampling rate. If you want to train model on other dataset, you can change config values according
+# to your dataset. Most dataset-specific arguments are in the head of the config file, see below.
+
+name: FastPitch
+
+train_dataset: ???
+validation_datasets: ???
+sup_data_path: ???
+sup_data_types: ["align_prior_matrix", "pitch", "speaker_id"]
```
redoctopus, line 10: Same as above.

examples/tts/conf/es/fastpitch_align_44100.yaml
rlangman: I suppose fine tuning model on a single speaker could be the most common use case. But I do not have any good single speaker dataset yet (the 15 hour one I scraped from LibriVox was pretty bad). For the multi-speaker it also has the additional configs "speaker_emb_condition_prosody" and "speaker_emb_condition_aligner".

English is missing a multi speaker config example. In the German configs the 22.05kHz config is single speaker and the 44.1kHz is multi speaker, but the difference is not called out in the config name or comments.

Should I add a single speaker IPA config while adding a "_multi" suffix to these? Assuming we will need it later.

rlangman: To be more consistent, I made both configs single speaker and added a separate multi-speaker IPA config.

redoctopus: Gotcha. That sounds like a good compromise. Can you add a bit of text in the header comment just to distinguish them and point out the differences for single vs. multi speaker configs?

examples/tts/conf/es/fastpitch_align_44100.yaml
```diff
@@ -0,0 +1,222 @@
...
+sup_data_path: ???
+sup_data_types: ["align_prior_matrix", "pitch"]
+
+# Default values from librosa.pyin
+pitch_fmin: 65.40639132514966
+pitch_fmax: 2093.004522404789
+
+# Stats (per frame), train (these values depend on pitch_fmin and pitch_fmax)
+pitch_mean: 212.35873413085938
+pitch_std: 68.52806091308594
```
XuesongYang, line 18: [discussion] no need to change for this PR. we provided default values for pitch norm params in all yaml configs. I wonder if we should replace them with `???` just in case users forget to update with dataset-specific values?

rlangman: I do think it would be safer to leave them (and any other dataset specific configs like n_speakers) as ???

redoctopus: Yeah, that's a good idea. Let's have a separate PR to set them to `???` across the board.

redoctopus (approved):

examples/tts/conf/es/fastpitch_align_44100.yaml
XuesongYang: ok. I will file a separate PR to replace all pitch norm params with `???`

XuesongYang (approved):
LGTM. I will follow up with another PR to replace pitch norms with `???`.

PR #5415, opened 2022-11-14
Draft: Add support for modeling words in kenLM for subword models.

galv:
# What does this PR do ?

Add support for modeling words in kenLM for subword models.

**Collection**: ASR

# Changelog 
- Add support for modeling words in kenLM for subword models.
- Add missing dependency to beam search installation.

# Usage
NA

**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

Finally, this is a draft. I am going to check with the effected party that these changes will work for them. Then maybe someone like @VahidooX can take a look.

# Additional Information
One commit message describes this in detail. Pasting here:

Previously, NeMo did not support beam search with a KenLM language
model trained on words when using a subword acoustic model. Rather, it
forces KenLM to model subpiece units instead. This caused problems for
a customer using https://github.com/nvidia-riva/riva-asrlib-decoder,
which does support this out of the box.

Adding --encoding_level_override=char enables this behavior. It is an
opt-in behavior.

There are very other additions here to make training KenLM models more
flexible:

- I add "--skip_symbols" to the "lmplz" call. Sometimes training data
can have "<s>" or "</s>" in it, which can cause a crash, unless you
trun this flag on. Always turning on the flag has no bad effects and
is backwards compatible.

- I add support for lmplz's "-T" option. This is useful if you are
working on a machine where "/tmp", the default directory, is small.

- I added "--limit_vocab_file". It speeds up training when you happen
to have a training file that has a lot of out of your vocabulary of interest
(e.g., scraped from webpages).

I disable forcibly deleting the arpa file created by lmplz. You need
to keep this if you are using a WFST decoder.

### DIFF
--- a/scripts/asr_language_modeling/ngram_lm/train_kenlm.py
+++ b/scripts/asr_language_modeling/ngram_lm/train_kenlm.py
@@ -81,6 +81,16 @@ def main():
     parser.add_argument(
         "--do_lowercase", action='store_true', help="Whether to apply lower case conversion on the training text"
     )
+    parser.add_argument(
+        "-T", default=None, help="Provide non-default temporary directory for KenLM's lmplz. Useful if /tmp runs out of space during training."
+    )
+    parser.add_argument(
+        "--limit_vocab_file", default=None, help="see lmplz --limit_vocab_file documentation. https://github.com/kpu/kenlm/blob/bcd4af619a2fa45f5876d8855f7876cc09f663af/lm/builder/lmplz_main.cc#L107"
+    )
+    parser.add_argument(
+        "--encoding_level_override", default=None,
+        help="NeMo's ctc beam search implementation does not support KenLM modeling words (the so-called 'char' encoding) for acoustic models that predict subpiece units. Rather, it makes KenLM model the individual subword units instead. However, https://github.com/nvidia-riva/riva-asrlib-decoder does support this. If you happen to be using that decoder, set --encoding_level_override=char when using a subword model."
+    )
     args = parser.parse_args()
 
     """ TOKENIZER SETUP """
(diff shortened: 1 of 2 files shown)


titu1994:
@galv is this ready for review? Could you rebase and push please ?

galv:
@titu1994 I thought adding "Draft:" to the title would make it a draft, but it doesn't apparently. This is a draft for now. (I just changed it.) I needed to share the change with the person who's affected. Once that person gets back to me with feedback that this works, we can mark it as a non-draft.

scripts/asr_language_modeling/ngram_lm/train_kenlm.py
```diff
@@ -81,6 +81,21 @@ def main():
...
+        "-T",
+        default=None,
+        help="Provide non-default temporary directory for KenLM's lmplz. Useful if /tmp runs out of space during training.",
+    )
+    parser.add_argument(
+        "--limit_vocab_file",
+        default=None,
+        help="see lmplz --limit_vocab_file documentation. https://github.com/kpu/kenlm/blob/bcd4af619a2fa45f5876d8855f7876cc09f663af/lm/builder/lmplz_main.cc#L107",
+    )
+    parser.add_argument(
```
VahidooX, line 94: As NeMo currently does not support word-level KenLM, what is the purpose of training such a model with nemo?
The Riva users may use the training scripts provided by the Riva right?

scripts/asr_language_modeling/ngram_lm/train_kenlm.py
```diff
@@ -81,6 +81,21 @@ def main():
...
+        "--encoding_level_override",
```
VahidooX, line 95: Maybe use no_encoding or skip_encoding as the parameter's name to be shorter?

VahidooX (changes requested):

scripts/asr_language_modeling/ngram_lm/train_kenlm.py
galv: The short answer is that there are some ASR users that aren't using Riva.

With more elaboration, CTC model decoding is fastest when using a WFST decoder on the GPU and people want this outside of Riva. In addition, whenever non-CTC models come (in particular, from K2 experiments), it is smart to use a WFST decoder because it is generic. There will be no need to manually create a decoder for each topology you happen to train with.

Of course, I'm not super happy about just adding on options to existing code in general; this was just a first draft to unblock someone.

titu1994: Hmm it seems like it will invite confusion. Also word LMs occupy way too much storage space + training time. We should have a seperate arg to make it a users explicit choice to train such model. Defaulting to char or subword LM otherwise.

PR #5424, opened 2022-11-15
Fixing bug in Megatron BERT when loss mask is all zeros

shanmugamr1992:
Signed-off-by: Shanmugam Ramasamy <111910568+shanmugamr1992@users.noreply.github.com>

# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


yidong72 (approved):
LGTM

MaximumEntropy (changes requested):
Hey! Sorry to block merging, but I think this might be a symptom of a bug at the data level and we should try to get to the bottom of that and not handle it this way.

yidong72:
> Hey! Sorry to block merging, but I think this might be a symptom of a bug at the data level and we should try to get to the bottom of that and not handle it this way.

It related to the dataset @shanmugamr1992 is using. I didn't see this behavior in my biomegatron dataset before.

shanmugamr1992:
@yidong72  @MaximumEntropy  So this is not about my dataset. The number of tokens to be padded is calculated as (total_tokens. * masking_probability) . So if total tokens is 200 and masking probability is 0.15, then 30 random tokens are padded. However if the number of tokens is small (Say 4 or 5) meaning remaining of the 512 are just zero padded, we have this case that there is nothing to be predicted. (Loss mask is all zeros). 
So this can be handled in two ways : 
1. Change the code that says minimum 1 token needs to be masked (However I think it might throw some weird behaviour when we only have 1 token in a batch)
2. We can assign the loss for that batch to be zero , which seems quite right, since the actual loss for that batch, when there is nothing to predict is zero.
I tried both approaches. Approach 1 threw Nan's after 20k or something. I can run it locally and debug, or we can go ahead with approach 2 , which is not that hacky. 
NOTE : This is a very weird corner case that will happen only when you set the microbatchsize to be 1. So wont really affect users as much.

nemo/collections/nlp/data/language_modeling/megatron/dataset_utils.py
```diff
@@ -234,6 +234,10 @@ def create_masked_lm_predictions(
         return (output_tokens, masked_lm_positions, masked_lm_labels, token_boundary)
 
     num_to_predict = min(max_predictions_per_seq, max(1, int(round(len(tokens) * masked_lm_prob))))
+    if num_to_predict < 1:
+        logging.info(
```
yidong72, line 238: logging.warn?

MaximumEntropy (dismissed):

MaximumEntropy (approved):

PR #5455, opened 2022-11-17
[TTS] Support speaker level pitch normalization

rlangman:
# What does this PR do ?

Add script for computing speaker specific pitch mean & standard deviation, and logic in data loader to use it at training time.

Global pitch normalization can be done using the previous configs, or by providing a JSON with only a "default" key.

There is a bit of awkward typing logic due to our speaker IDs being integers but JSON only supporting string keys.

**Collection**: [TTS]

# Changelog
- Add script for computing the mean and standard deviation of pitch for each speaker in a dataset and storing it in a JSON dictionary
- Add data loading logic to support parsing these pitch statistics from the JSON.
- Register stats file in FastPitch model
- Add pitch stats for Spanish openSLR data and modify .yaml config example.
- Added map from speaker name to speaker ID (integer) used as the keys in the pitch statistics, and as the possible inputs to the NGC model (https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan). The script for creating this speaker mapping will not be merged until I add the script/tutorial for downloading and processing the openSLR data.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


scripts/dataset_processing/tts/compute_speaker_stats.py
```diff
@@ -0,0 +1,108 @@
...
+
+from nemo.collections.tts.data.data_utils import get_sup_data_file_path, read_manifest
+from nemo.collections.tts.torch.helpers import get_base_dir
+from nemo.collections.tts.torch.tts_data_types import Pitch
+from nemo.utils import logging
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter, description="Compute speaker level pitch statistics .",
```
redoctopus, line 45: Nitpick: extra space before period

examples/tts/conf/es/fastpitch_align_44100_ipa_multi.yaml
```diff
@@ -13,10 +13,8 @@ sup_data_types: ["align_prior_matrix", "pitch", "speaker_id"]
 pitch_fmin: 65.40639132514966
 pitch_fmax: 2093.004522404789
 
-# these frame-wise values depend on pitch_fmin and pitch_fmax, you can get values
-# by running `scripts/dataset_processing/tts/extract_sup_data.py`
-pitch_mean: ???  # e.g. 178.86880493164062 for 174 speakers in https://research.google/pubs/pub49150/
```
subhankar-ghosh, line 18: Is this backward compatible? Meaning, with these changes can the checkpoints published in NGC be loaded and finetuned/inferred on?

scripts/dataset_processing/tts/compute_speaker_stats.py
```diff
@@ -0,0 +1,108 @@
...
+
+    global_pitch_values = []
+    speaker_pitch_values = defaultdict(list)
+    for entry in tqdm(entries):
+        pitch_path = get_sup_data_file_path(entry, base_dir, sup_data_path, Pitch.name)
+        if not os.path.exists(pitch_path):
+            logging.warning(f"Unable to find pitch file for {entry}")
+            continue
+
+        pitch = torch.load(pitch_path)
```
subhankar-ghosh, line 89: So, this pitch calculation script will only work in case the pitch data is already calculated. It might be a good idea to provide a message to the users about this requirement, also pointing to any code/tutorial that calculates pitch data.

scripts/dataset_processing/tts/compute_speaker_stats.py
```diff
@@ -0,0 +1,108 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This script is to compute speaker-level statistics, such as pitch mean & standard deviation, for a given
+TTS training manifest.
+
+This script should be run after extract_sup_data.py as it uses the precomputed supplemental features.
```
subhankar-ghosh, line 19: It may be a good idea to put this in the warning as well, so that if a user does not read this line and still runs the script he/she will be shown this warning.

nemo/collections/tts/torch/data.py
```diff
@@ -165,6 +165,7 @@ def __init__(
             pitch_mean (Optional[float]): The mean that we use to normalize the pitch.
             pitch_std (Optional[float]): The std that we use to normalize the pitch.
             pitch_norm (Optional[bool]): Whether to normalize pitch (via pitch_mean and pitch_std) or not.
+            pitch_stats_path (Optional[Path, str]): Path to file containing speaker level pitch statistics.
```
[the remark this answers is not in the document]
rlangman: I will update the pitch_norm docstring to clarify the requirement.

examples/tts/conf/es/fastpitch_align_44100_ipa_multi.yaml
rlangman: Yes. To confirm, I will test it with one of the existing English FastPitch models on NGC.

scripts/dataset_processing/tts/compute_speaker_stats.py
rlangman: Sure, I will provide a more helpful warning message if the sup_data/pitch directory does not exist.

nemo/collections/tts/torch/data.py
```diff
@@ -416,6 +417,23 @@ def add_pitch(self, **kwargs):
         self.pitch_mean = kwargs.pop("pitch_mean", None)
         self.pitch_std = kwargs.pop("pitch_std", None)
         self.pitch_norm = kwargs.pop("pitch_norm", False)
+        pitch_stats_path = kwargs.pop("pitch_stats_path", None)
+
+        if self.pitch_norm:
+            # XOR to validate that both or neither pitch mean and std are provided
+            if (self.pitch_mean is None) != (self.pitch_std is None):
```
subhankar-ghosh, line 424: may be use assert, to reduce the number of if statements.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -131,6 +131,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         self.duration_loss_fn = DurationLoss(loss_scale=dur_loss_scale)
         self.energy_loss_fn = EnergyLoss(loss_scale=energy_loss_scale)
 
+        if "pitch_stats_path" in cfg:
```
subhankar-ghosh, line 134: It is not clear to me why do we need to register pitch_stats_path as an artifact of the model? Pitch stats is a feature derived from the data and not the model. Also, if we do register it as an artifact, won't it throw errors when we try to load checkpoints which were trained using `pitch_mean`, `pitch_std` ... variables?

scripts/dataset_processing/tts/compute_speaker_stats.py
```diff
@@ -0,0 +1,108 @@
...
+        global_pitch_values.append(pitch)
+        if "speaker" in entry:
+            speaker_id = entry["speaker"]
+            speaker_pitch_values[speaker_id].append(pitch)
+
+    global_pitch_mean, global_pitch_std = _compute_stats(global_pitch_values)
+    pitch_stats = {"default": {"pitch_mean": global_pitch_mean, "pitch_std": global_pitch_std}}
+    for speaker_id, pitch_values in speaker_pitch_values.items():
+        pitch_mean, pitch_std = _compute_stats(pitch_values)
+        pitch_stats[speaker_id] = {"pitch_mean": pitch_mean, "pitch_std": pitch_std}
```
subhankar-ghosh, line 101: why not use the same stats to calculate `pitch_min` and `pitch_max` ?

nemo/collections/tts/torch/data.py
```diff
@@ -554,12 +572,26 @@ def __getitem__(self, index):
...
+                    sample_pitch_std = self.pitch_std
+                elif self.pitch_stats:
+                    if "speaker_id" in sample and str(sample["speaker_id"]) in self.pitch_stats:
+                        pitch_stats = self.pitch_stats[str(sample["speaker_id"])]
+                    elif "default" in self.pitch_stats:
+                        pitch_stats = self.pitch_stats["default"]
+                    else:
+                        raise ValueError(f"Could not find pitch stats for {sample}.")
+                    sample_pitch_mean = pitch_stats["pitch_mean"]
+                    sample_pitch_std = pitch_stats["pitch_std"]
```
subhankar-ghosh, line 588: We should get `pitch_min` and `pitch_max` as well from the same dict?

examples/tts/conf/es/fastpitch_align_44100_ipa_multi.yaml
subhankar-ghosh: Great.

scripts/dataset_processing/tts/compute_speaker_stats.py
rlangman: Do you mean include pitch_min and pitch_max just for informational purposes? Or to replace the "pitch_fmin" and "pitch_fmax" values in the config? It cannot do the latter because those are not used at train time, but as inputs to librosa.pyin() to compute the F0 to begin with.

nemo/collections/tts/models/fastpitch.py
rlangman: I was trying to copy the convention of the pitch_mean/pitch_std in which in the .yaml it configures it at the top level, then saves it in the model, and then passes it from model to data loader. It then saves some value in the module, but I cannot tell if that actually does anything. https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/modules/fastpitch.py#L189-L190

Register artifact copies the artifact into the .nemo file and then replaces the path on a user's local machine with the relative path in the checkpoint. It will have no impact if that config is not provided. If you do not register it, then loading the checkpoint breaks whenever you load it on a different machine (or move the path the file as at). So I think registering it somewhere is necessary.

I think this does imply that you cannot change the convention from the original training. So if you train with the hardcoded mean/std, you cannot fine tune with the dictionary based config. Or vice versa.

rlangman: After testing it, I don't think register artifact matters for training or fine-tuning. It is only used when loading the model for inference, which does not need the data loader configs.

So, removing the call to register_artifact.

scripts/dataset_processing/tts/compute_speaker_stats.py
subhankar-ghosh: Yes I mean pitch_fmin and pitch_fmax. They are used in the `__get_item__` method here https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/torch/data.py#L538
I don't follow why can't we get these also from your dict?

subhankar-ghosh:
Did you test the existing FastPitch training/finetuning code in the tutorials? All of them use pitch inputs, will they break or work with these changes?

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: sure.

scripts/dataset_processing/tts/compute_speaker_stats.py
rlangman: The fmin and fmax are not dataset statistics, they are bounds (hardcoded to 65Hz-2093Hz) to make the F0 estimation more computationally efficient.

That line which calls "librosa.pyin()" is only called if you compute the pitch on the first epoch instead of running "extract_sup_data.py". This script computes the pitch mean/std using the outputs of librosa.pyin(), assuming it was done beforehand using "extract_sup_data.py".

Though that also means no one should use this branch of the data loader you linked, as that would mean they begin training before knowing the pitch norm statistics. Some of our past tutorials suggested people do this without realizing the bug (causing them to train on their dataset using the hardcoded LJSpeech mean/std).

rlangman:
> Did you test the existing FastPitch training/finetuning code in the tutorials? All of them use pitch inputs, will they break or work with these changes?

From my testing, it does not break anything. The config in the checkpoint is only used for inference, and inference does not use any of the data loader parameters.

During training and finetuning it does not reuse any of the config values from the previous training run. You are required to provide all data loader params again; pitch mean/std, phoneme_dict, heteronyms, etc.

scripts/dataset_processing/tts/compute_speaker_stats.py
rlangman: I see now our FP training tutorial suggests tuning fmin/fmax to target speaker, which is very confusing. The only reason to tune it (if I am not mistaken) is if you want to (1) reduce the range to make the compute faster, but risk truncating the pitch values or (2) have training data greater than 2k Hz such as singing or yelling, and are willing to heavily slow down the compute to capture those ranges.

https://github.com/NVIDIA/NeMo/blame/main/tutorials/tts/FastPitch_Finetuning.ipynb#L347-L349

redoctopus: Reaching back in my memory--I recall we had some issues where some speakers we were fine-tuning on had noisier environments, and pyin kept predicting pitches for static/noise as speech. So in the end we went speaker-by-speaker and adjusted fmin (don't think we touched fmax) to avoid "noise-as-speech" with minimal loss of actual speech pitches.

Anyway, that's probably what that line refers to, but yeah... without more explanation it's somewhat unclear.

rlangman: That would make sense. You might be able to eliminate outliers or influence the VAD output by restricting the range of F0 to normal human speech (I have no clue how the VAD in pyin works).

redoctopus (commented):
LGTM. @subhankar-ghosh all good?

subhankar-ghosh:
LGTM

subhankar-ghosh (approved):

PR #5462, opened 2022-11-19
AIStore for ASR datasets

anteju:
Signed-off-by: Ante Jukić <ajukic@nvidia.com>

# What does this PR do ?

This PR adds an option to use ASR datasets from an AIStore cluster.

## NOTE

Blocked by #5569.

**Collection**: [ASR]

# Changelog 
- Added AIS binary to Docker image
- Added env variables for configuring NeMo data store cache
- Added an option to defer dataloader initialization to `fit`, so data download can be synchronized across workers after a distributed environment is initialized
- For tarred datasets, added caching of manifests to store cache
- For non-tarred datasets, added caching of manifests and audio to store cache
- Utilities for handling data from AIStore cluster

# Usage

### AIS cluster config

AIStore endpoint needs to be configured before running the training command

```bash
export AIS_ENDPOINT=host:port
```

### Dataset setup

Manifests and tarred filepaths can point to AIS. For example, manifest can be set as

```bash
manifest_filepath = ‘ais://bucket/train_manifest.json'
```

and tarred filepaths as

```bash
tarred_audio_filepaths = ‘ais://bucket/audio__OP_0..511_CL_.tar'
```

To make sure data is download only on local or global rank zero, we use defer dataloader setup.
This can be achieved by setting `defer_setup = True` for the corresponding dataset or as

```bash
+model.train_ds.defer_setup=true \
+model.validation_ds.defer_setup=true \
+model.test_ds.defer_setup=true \
```

Note: A similar behavior could be achieved by using DataModule and controlling shared/non-shared caching using `datamodule.prepare_data_per_node`. However, that may require a more significant refactor.

### Cache setup

Local node-specific cache can be set up by setting the following env variables:

```bash
export NEMO_DATA_STORE_CACHE_DIR=/tmp
export NEMO_DATA_STORE_CACHE_SHARED=0
```
In this case, each local rank zero will cache data from AIS.

A shared cache location can be setup, for example, as

```bash
export NEMO_DATA_STORE_CACHE_DIR=/path/to/shared/dir
export NEMO_DATA_STORE_CACHE_SHARED=1
```
In this case, only the global rank zero will cache data from AIS.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Blocked by #5569. Currently rebased on top of that branch.

### DIFF
--- a/nemo/collections/asr/data/audio_to_text.py
+++ b/nemo/collections/asr/data/audio_to_text.py
@@ -251,17 +251,22 @@ def cache_datastore_manifests(
         def cache_data(manifest_filepaths, cache_audio, num_workers, max_num_workers):
             """Cache manifests and audio data from object store.
             """
+            # Determine the number of workers to use
             if num_workers is None:
                 num_workers = os.cpu_count() - 1
             num_workers = min(num_workers, max_num_workers)
 
+            # Process each manifest file
             for manifest_file in manifest_filepaths:
+                # If manifest is on a data store, then cache it.
+                # Otherwise, nothing to do.
                 if is_datastore_path(manifest_file):
                     logging.info('Cache manifest file: %s', manifest_file)
                     cached_manifest_file = DataStoreObject(manifest_file).get()
                     logging.info('Cached at: %s', str(cached_manifest_file))
 
                     if cache_audio:
+                        # Each audio file from manifest will be cached.
                         logging.info('Cache audio from manifest file: %s', manifest_file)
                         # Assumes that manifest is using relative paths
                         manifest_dir = os.path.dirname(manifest_file)
@@ -285,13 +290,13 @@ def cache_datastore_manifests(
                             for audio_object in tqdm(audio_objects):
                                 result.append(audio_object.get() is not None)
 
-                        if any(result) == False:
+                        if not all(result):
                             raise RuntimeError('Some files not downloaded successfully')
                         logging.info('Caching complete')
 
                 else:
                     # Nothing to do here
-                    logging.debug('Manifest is not on a store: %s', manifest_file)
+                    logging.debug('Manifest is not on a data store: %s', manifest_file)
 
         if torch.distributed.is_available() and torch.distributed.is_initialized():
             logging.debug('Distributed environment is available and initialized.')
--- a/nemo/utils/data_utils.py
+++ b/nemo/utils/data_utils.py
@@ -152,7 +152,7 @@ def datastore_path_to_local_path(store_path: str) -> str:
     if store_path.startswith('ais://'):
         endpoint = ais_endpoint()
         if endpoint is None:
-            raise RuntimeError(f'AIS endpoint not set, cannot resolve {path}')
+            raise RuntimeError(f'AIS endpoint not set, cannot resolve {store_path}')
 
         local_ais_cache = os.path.join(ais_cache_base(), ais_endpoint_to_dir(endpoint))
         store_bucket, store_object = bucket_and_object_from_uri(store_path)


Dockerfile
```diff
@@ -94,3 +94,10 @@ COPY tutorials /workspace/nemo/tutorials
 
 RUN printf "#!/bin/bash\njupyter lab --no-browser --allow-root --ip=0.0.0.0" >> start-jupyter.sh && \
     chmod +x start-jupyter.sh
+
+# Prepare AIS CLI
+ARG AIS_VERSION=3.11
```
titu1994, line 99: Don't use so many seperate RUN, use just one.

titu1994 (commented):
Hmm can you ask someone else to do a thorough review ? I'll review as well but best to have multiple people know of the internal implementation details of this pr.

nemo/core/classes/modelPT.py
```diff
@@ -128,13 +128,25 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             app_state.device_id = torch.cuda.current_device()
 
         if self._cfg is not None and not self._is_model_being_restored():
-            if 'train_ds' in self._cfg and self._cfg.train_ds is not None:
+            if (
+                'train_ds' in self._cfg
+                and self._cfg.train_ds is not None
+                and not self._cfg.train_ds.get('defer_setup', False)
```
okuchaiev, line 134: what is this "defer_setup"?

nemo/core/classes/modelPT.py
```diff
@@ -676,6 +688,40 @@ def setup_optimizer_param_groups(self):
 
         self._optimizer_param_groups = param_groups
 
+    def setup(self, stage: Optional[str] = None):
```
okuchaiev, line 691: is there a better way to do it? What if user needs to overwrite setup for their model

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +226,125 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+
+                        if num_workers is not None and num_workers > 1:
+                            logging.debug('Using multiprocessing with num_workers: %d.', num_workers)
+                            with multiprocessing.Pool(processes=num_workers) as p:
+                                result = list(tqdm(p.imap(store_object_get, audio_objects), total=len(audio_objects)))
+                        else:
+                            logging.debug('Using a single process.')
+                            result = []
+                            for audio_object in tqdm(audio_objects):
+                                result.append(audio_object.get() is not None)
```
titu1994, line 288: This is appending bools ?

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +226,125 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+        # Handle distributed environment
+        if shared_cache is None:
+            shared_cache = is_store_cache_shared()
+
+        if shared_cache:
+            logging.debug('Cache is shared among nodes.')
+            is_rank_zero = is_global_rank_zero()
+        else:
+            logging.debug('Cache is not shared among nodes.')
+            local_rank = int(os.environ.get("LOCAL_RANK", 0))
```
titu1994, line 307: If multi gpu is setup, but local rank is not set correctly, all ranks will be set to 0. Do we want to catch this behaviour and raise warning?

titu1994: Oh its done below, but all of them will be rank 0. I guess its fine.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +226,125 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+        else:
+            logging.warning(
+                'Torch distributed is not initialized and caching may be prone to data race conditions. '
+                'This process is not %s rank 0, and now will go to sleep for %d seconds. If this '
+                'rank wakes from sleep prior to rank 0 finishing downloading, errors might result.',
+                'global' if shared_cache else 'local',
+                num_store_manifests * sleep_seconds,
+            )
+            time.sleep(num_store_manifests * sleep_seconds)
+
```
titu1994, line 346: Is this the busy waiting part? Doesnt num store manifest * sleep seconds become dozens seconds?

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +226,125 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
     return audio_tar_filepaths
 
 
+def cache_store_manifests(
```
titu1994, line 229: This could potentially be a class method of ASRManifestProcessor. Upto you

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -266,6 +403,9 @@ def __init__(
         if type(manifest_filepath) == str:
             manifest_filepath = manifest_filepath.split(",")
 
+        # If necessary, cache manifests and audio from object store
+        cache_store_manifests(manifest_filepaths=manifest_filepath, cache_audio=True)
```
titu1994, line 407: If it was a class method you could call self.manifest_processor.cache_store_manifest() before below instantiation of ASRManifestProcessor. Just a thought, ignore if its extra work.

nemo/core/classes/modelPT.py
titu1994: Other alternatives are more problematic. As far as I know, super() is a catch all empty method declaration in base PTL. If subclasses need to call ModelPT (this) impl, they can do so with super().

NLPModel overrides it currently without calling super, and that's fine, since they have their own way of deferring data setup. They may choose to call it (and ignore the processing of defer_setup arg) or not call it (and also ignore the defer_setup config argument).

nemo/core/classes/modelPT.py
```diff
@@ -676,6 +688,40 @@ def setup_optimizer_param_groups(self):
...
+        Args:
+            stage: fit, validate, test or predict
+        """
+        if stage == 'fit':
+            train_deferred_setup = (
+                'train_ds' in self._cfg
+                and self._cfg.train_ds is not None
+                and self._cfg.train_ds.get('defer_setup', False)
+            )
+            if self.train_dataloader() is None and train_deferred_setup:
```
titu1994, line 704: One thing im worried about is caching - it generally causes problems. But i suppose for the defer setup flag will mostly be fine since by default it is false and this codepath will not execute unless specified.

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,289 @@
...
+import pathlib
+import subprocess
+from typing import Tuple
+
+import nemo
+from nemo import constants
+from nemo.utils import logging
+
+
+def resolve_cache_dir() -> pathlib.Path:
```
titu1994, line 25: This method should be imported inside of model_utils for backward compat.

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,289 @@
...
+
+    endpoint = endpoint.replace('http://', '')
+    host, port = endpoint.split(':')
+    return os.path.join(host, port)
+
+
+def ais_binary() -> str:
+    """Return location of `ais` binary.
+    """
+    return '/usr/local/bin/ais'
```
titu1994, line 127: Should this be an env variable? Any chance users can have a different path?

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,289 @@
...
+                raise RuntimeError('Download failed: %s', subprocess.list2cmdline(cmd))
+
+        return local_path
+
+    else:
+        # Assume the file is local
+        return path
+
+
+class StoreObject:
```
titu1994, line 210: Make the name explicit by calling it `AIStoreObject` or `AIStoreObjectWeakRef`. Confusing unless the class is more detached from AIStore.

nemo/utils/model_utils.py
```diff
@@ -565,25 +561,6 @@ def check_lib_version(lib_name: str, checked_version: str, operator) -> Tuple[Op
     return None, msg
 
 
-def resolve_cache_dir() -> Path:
```
titu1994, line 568: Import a reference to this method in this file for backward compatibility

titu1994 (commented):
Consider if all debug logging statements are needed. Overall looks good.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +226,125 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+                cache_data(
+                    manifest_filepaths=manifest_filepaths,
+                    cache_audio=cache_audio,
+                    num_workers=num_workers,
+                    max_num_workers=max_num_workers,
+                )
+            logging.debug('Reached barrier')
+            torch.distributed.barrier()
+
+        elif is_rank_zero:
```
Kipok, line 323: Can we fully remove this part and raise an error in torch distributed is not initialized? This seems quite fragile to me with the ad-hoc sleep time, and potentially undefined behavior if multiple processes write to the same file. So I'd prefer to require distributed to be initialized for ais to be used.

And for the rank, would it be safer to use `torch.distributed.get_rank` (if we assume it's always initialized)? It seems that `is_global_rank_zero` checks a bunch of environment variables which (I guess) might not always work correctly

nemo/collections/asr/data/audio_to_text.py
Kipok, line 407: Does it make sense to expose other caching parameters to users?

nemo/collections/asr/parts/utils/manifest_utils.py
```diff
@@ -368,6 +369,8 @@ def read_manifest(manifest: str) -> List[dict]:
     Returns:
         data (list): List of JSON items
     """
+    manifest = StoreObject(manifest).get()
```
Kipok, line 372: This seems a bit strange to me as documentation of `StoreObject` mentions that it's used for AIS objects, but we use it here and in other files for both ais and regular objects. Maybe we should update documentation or maybe change api, such that it's more visible (e.g., adding `get_manifest` or `prepare_manifest` function that will wrap this)?

Kipok: It's also a bit strange that in some cases we use "functional" api, like `store_path_to_local_path` or `is_store_path` and in some cases we use "class" api like here. Would be cleaner in my opinion if it was the same style (e.g., always wrapping with class objects and then using methods or always using top-level functions)

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,289 @@
...
+        path = pathlib.Path.joinpath(pathlib.Path.home(), f'.cache/torch/NeMo/NeMo_{nemo.__version__}')
+    else:
+        path = pathlib.Path(override_dir).resolve()
+    return path
+
+
+def is_store_path(path) -> bool:
+    """Check if a path is from an object store.
+    """
+    if path.startswith('ais://'):
```
Kipok, line 47:
```suggestion
    return path.startswith('ais://')
```

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,289 @@
...
+        return True
+    else:
+        return False
+
+
+def is_store_cache_shared() -> bool:
+    """Check if store cache is shared.
+    """
+    # Assume cache is shared by default, e.g., as in resolve_cache_dir (~/.cache)
+    cache_shared = int(os.environ.get(constants.NEMO_ENV_DATA_STORE_CACHE_SHARED, 1))
```
Kipok, line 57: Not sure how this is setup in nemo, but if that's a normal environment variable, should we add conversion to int? Since normally it would be a string if set?

nemo/utils/data_utils.py
Kipok: Agree with Som on the naming - I'd rename everything from "store" to "aistore", otherwise it's not immediately clear what kind of store we refer to. Or if you want to make API generic so that we can add more things in the future, then maybe "Store" is not a good name and we should rename it to "cluster" or "remote" or something like that?

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +224,122 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+                                result = list(
+                                    tqdm(p.imap(datastore_object_get, audio_objects), total=len(audio_objects))
+                                )
+                        else:
+                            logging.debug('Using a single process.')
+                            result = []
+                            for audio_object in tqdm(audio_objects):
+                                result.append(audio_object.get() is not None)
+
+                        if any(result) == False:
```
Kipok, line 288:
```suggestion
                        if not any(result):
```
of "is False" instead of "== False"

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -208,6 +224,122 @@ def expand_audio_filepaths(audio_tar_filepaths, shard_strategy: str, world_size:
...
+                    cache_audio=cache_audio,
+                    num_workers=num_workers,
+                    max_num_workers=max_num_workers,
+                )
+            logging.debug('Reached barrier')
+            torch.distributed.barrier()
+
+        elif is_global_rank_zero():
+            # Handle non-distributed environment, e.g., if running on a single GPU
+            logging.warning(
```
Kipok, line 324: I'd remove this warning, since the only legit situation for getting here without error is when using single GPU. In other cases, the code will raise exception from other ranks anyway

nemo/utils/data_utils.py
```diff
@@ -0,0 +1,312 @@
...
+    Args:
+        store_path: a path to an object on an object store
+
+    Returns:
+        Path to the same object in local cache.
+    """
+    if store_path.startswith('ais://'):
+        endpoint = ais_endpoint()
+        if endpoint is None:
+            raise RuntimeError(f'AIS endpoint not set, cannot resolve {path}')
```
Kipok, line 155:
```suggestion
            raise RuntimeError(f'AIS endpoint not set, cannot resolve {store_path}')
```
@titu1994 - I strongly suggest revisiting our CI setup to make sure such errors are caught by linter

Kipok (dismissed):
I am approving since the requested changes are minor, but @anteju please make sure to fix undefined variable before merging

nemo/utils/data_utils.py
titu1994: image
Its managed by Github CodeQL - https://github.com/NVIDIA/NeMo/pull/5443 in the autocast() function of eval_beam_search.

nemo/collections/asr/data/audio_to_text.py
titu1994: Lets keep it, at worst its a warning and at best it will indicate where to debug in multinode failure.

nemo/core/classes/modelPT.py
titu1994: One thing with the deffered flag is if you use it, you cannot register any artifact inside of setup_train/val/test methods. Which IMO is fine, we want to allow registration of artifacts only at init time.

nemo/utils/data_utils.py
titu1994: Alternative : try catch cast int, check ints, except, check str, if even that fails, give up and raise error.
Im fine with current impl.

titu1994 (approved):
Excellent work, ready to merge.

Do you need so many reviews? Eric and Oleksii are for core, that's merged already. Merge when ready or after one last reviewer?

Kipok (approved):

Kipok (approved):

PR #5499, opened 2022-11-24
[TTS] Add adapter for new speaker adaptation

subhankar-ghosh:
# What does this PR do ?

Adds adapter support for new speaker adaptation. The following are the contributions of this PR:
1) Adds adapter support to TTS (FastPitch) pipeline for new speaker adaptation.
2) Adds Global Style Token based speaker representation to multi-speaker TTS pipeline.
3) Adds Speaker Verification based speaker representation to multi-speaker TTS pipeline.

**Collection**: [tts]

# Changelog 
- `examples/tts/conf/fastpitch_speaker_adaptation.yaml` - A new config file that is used for adapter training or using GST speaker representation.
- `examples/tts/fastpitch_finetune_adapters.py` - Script to Finetune adapters for new speaker adaptation. This script edits the config to add adapters, freeze the pretrained model and trains just the adapters.
- `nemo/collections/tts/parts/mixins/fastpitch_adapter_mixins.py` - This is a new file that contains all adapter related methods for FastPitch model. `parts/mixins` follows ASR folder structure.
- `nemo/collections/tts/modules/aligner.py, fastpitch.py, transformer.py` - the changes in these files are corresponding to inserting adapters to each of these modules.
- `nemo/collections/tts/modules/submodules.py` - GST based speaker representation module's implementation is in this file.
- `nemo/collections/tts/torch/tts_data_types.py, data.py` - GST related changes made to dataloader. GST needs a reference audio for a particular speaker. So a new datatype `GSTRefAudio` is introduced, this datatype refers to the reference audio.
- `tutorials/tts/FastPitch_Speaker_Adaptation.ipynb` - Tutorial explaining step by step how to use adapters to adapt a new speaker in FastPitch.

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


blisc:
Is there another PR for using Titanet or other SV model as embeddings?

examples/tts/conf/fastpitch_speaker_adaptation.yaml
```diff
@@ -0,0 +1,296 @@
...
+  sample_rate: ${sample_rate}
+  n_mel_channels: ${n_mel_channels}
+  n_window_size: ${n_window_size}
+  n_window_stride: ${n_window_stride}
+  n_fft: ${n_fft}
+  lowfreq: ${lowfreq}
+  highfreq: ${highfreq}
+  window: ${window}
+
+  adapter:
```
blisc, line 62: The comments on the adapter params are really nice. Is the config the right place to put them though?

nemo/collections/tts/models/fastpitch.py
```diff
@@ -142,13 +144,15 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         speaker_emb_condition_prosody = cfg.get("speaker_emb_condition_prosody", False)
         speaker_emb_condition_decoder = cfg.get("speaker_emb_condition_decoder", False)
         speaker_emb_condition_aligner = cfg.get("speaker_emb_condition_aligner", False)
+        global_style_token_encoder = instantiate(self._cfg.get("global_style_token", None))
```
blisc, line 147: Is there any chance that the changes inside of the `FastPitchModel` class with regard to all these GST components can be moved to the `FastPitchAdapterModelMixin` class instead? How will the `FastPitchModel` class work if we end up wanting to add SV embedding instead of GST?

subhankar-ghosh: `FastPitchAdapterModelMixin` class has functionalities related to adapters only. GST component is for speaker representation, so moving this to `FastPitchAdapterModelMixin` class will not be the right thing to do. What I have in mind is to have a speaker representation component (just like other components like pitch predictor, duration predictor etc) which can be initialized from the config. That way we can use GST or SV embedding or lookup embeddings or anything else based on the config. This would mean the config would have a section/struct for speaker representation like `model.speaker_embedding`.

examples/tts/conf/fastpitch_speaker_adaptation.yaml
subhankar-ghosh: I have their description in the tutorial as well, but for users who skip the tutorial I kept them in the config.

subhankar-ghosh:
> Is there another PR for using Titanet or other SV model as embeddings?
I was planning a separate PR for speaker representation using SV embedding, but now I think I will merge it with Adapter + GST PR (this PR). I would create a new PR as this branch seems to have diverged quite a bit from main.

nemo/collections/tts/models/fastpitch.py
subhankar-ghosh: I need to check compatibility of this approach with already published models.

subhankar-ghosh:
> > Is there another PR for using Titanet or other SV model as embeddings?
> 
> I was planning a separate PR for speaker representation using SV embedding, but now I think I will merge it with Adapter + GST PR (this PR). I would create a new PR as this branch seems to have diverged quite a bit from main.

Added Titanet based speaker embeddings to condition FastPitch.

nemo/collections/tts/models/fastpitch.py
blisc: We need a way of separating adapted FastPitch model from non-adapted Fastpitch models. I think we should consider a new class for this PR.

examples/tts/conf/fastpitch_speaker_adaptation.yaml
```diff
@@ -0,0 +1,303 @@
+# This config contains the default values for training FastPitch model with aligner using 44.1KHz sampling
```
blisc, line 1: This PR is quite substantial. We could improve on the following

- Consider creating a new class, subclass, mixin, etc to house this work
- This new object should contain all needed docstrings which are currently in the yaml file
- We should have two yamls: 1 for gst, and 1 for SV embeddings
- The tutorial is ok for NeMo, but it would be nice to make it easier to understand at parts
  - Parts of the tutorial that are function definitions can be moved into NeMo itself with a link to the github code and the corresponding cells that call it can be replaced with a simple function/script call

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -222,6 +273,9 @@ def forward(
         attn_prior=None,
         mel_lens=None,
         input_lens=None,
+        gst_ref_spec=None,
+        gst_ref_spec_lens=None,
```
rlangman, line 277: I would suggest the function take the output of the reference encoder/GST as input, and move the logic of running the GST inference to the model (or elsewhere). At training time a random reference can be used, but at inference it should be possible to provide a centroid or interpolated embedding computed across several references from a speaker.

For this specific implementation, it looks like you might be able to just preprocess it and pass in the final summed "speaker_embedding"

nemo/collections/tts/modules/submodules.py
```diff
@@ -401,3 +413,280 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+            "token_id": NeuralType(('B'), Index(), optional=True),
+        }
+
+    @property
+    def output_types(self):
+        return {
+            "style_emb": NeuralType(('B', 'D'), EncodedRepresentation()),
+        }
+
+    def forward(self, inputs, token_id=None):
```
rlangman, line 567: Please add documentation around the shapes/dimensions of the operations. Especially for things like "permute(2, 0, 1, 3)"

nemo/collections/tts/modules/submodules.py
```diff
@@ -401,3 +413,280 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        return {
+            "out": NeuralType(('B', 'D'), EncodedRepresentation()),
+        }
+
+    def forward(self, inputs, inputs_masks):
+        inputs = inputs.transpose(1,2)
+
+        inputs = inputs * inputs_masks
+        out = inputs.unsqueeze(3)
+        out = self.conv(out)
```
rlangman, line 526: Masking is required in between CNN layers (In theory, maybe not always in practice. Definitely required for final GRU layer). This is relatively complicated with reference encoder due to (2, 2) stride in CNN cutting sequence lengths in half at each step.

I suggest copying implementation from this or other github repo, use it to create mask after each CNN layer, and ideally adding a unit test to it: https://github.com/mozilla/TTS/blob/e9e07844b77a43fb0864354791fb4cf72ffded11/TTS/tts/layers/gst_layers.py#L84

nemo/collections/tts/modules/submodules.py
```diff
@@ -401,3 +413,280 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        out = out.view(out.shape[0], out.shape[1], -1).contiguous()
+        self.gru.flatten_parameters()
+        _, out = self.gru(out)
```
rlangman, line 529: It looks like masking in GRU requires converting its input to torch.nn.utils.rnn.pack_padded_sequence. Otherwise result will be very wrong for all but the longest sequence in the batch.

https://pytorch.org/docs/stable/generated/torch.nn.GRU.html?highlight=gru#torch.nn.GRU

nemo/collections/tts/modules/submodules.py
```diff
@@ -401,3 +413,280 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        # Project speaker embedding from Speaker Verification based 
+        if self.sv_projection_module is not None and speaker_embedding is not None:
+            x += self.sv_projection_module(speaker_embedding)
+        return x
+
+
+class Conv2d(torch.nn.Module):
+    """
+    Convolution 2D Module
+    """
```
rlangman, line 651: Could use a more unique name and clarification/comment that this wrapper is intended to handle the boilerplate transposes needed to convert input dimension X to conv2d input dimensions Y.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -79,7 +82,7 @@ def average_pitch(pitch, durs):
     return pitch_avg
 
 
-class ConvReLUNorm(torch.nn.Module):
+class ConvReLUNorm(torch.nn.Module, adapter_mixins.AdapterModuleMixin):
```
rlangman, line 85: I am wondering if it will be cleaner to move all of the adapter layers into a single adapters.py file, even if requires moving or duplicating some of the generic layers (like this and TemporalPredictor) to the submodules. Then the models definitions themselves will have no explicit adapter logic.

rlangman (commented):
As @blisc mentioned, the PR is quite substantial. I think we should break it into smaller PRs.

(1) The adapters
(2) GST
(3) The tutorial

nemo/collections/tts/models/fastpitch.py
Pydataman: how is the result？ I do this exp，but the result is bad

PR #5525, opened 2022-11-29
Conformer local attention

sam1373:
Signed-off-by: Samuel Kriman <samuelkriman@gmail.com>

# What does this PR do ?

This PR introduces the option of using local attention in Conformer with relative positional encoding, following the approach from Longformer (https://arxiv.org/abs/2004.05150), based on Ankur Kumar's branch https://github.com/iankur/NeMo/tree/longformer_rel_pos. This is useful for running on long audio files, as it can be done faster than with buffered inference. It may also improve performance on audio with long segments of noise that may confuse models with full attention.

This method can be used with checkpoints trained with regular attention. For example, Conformer-Large checkpoints (such as https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/stt_en_conformer_ctc_large and https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/stt_en_conformer_transducer_large) can be run on ~30 minute audio files on a single 32GB GPU after switching to local attention. The duration can be further increased to ~2 hours by switching to stacking subsampling (this won't work without additional training). Please note that the WER may increase slightly on our main evaluation sets when using small context local attention with models trained with full context (see table below for some WER(%) comparisons). This effect can be mitigated by either increasing the context size or fine-tuning the model using local attention.

| Model | LS test-other | LS test-clean | People Speech | SLR |
|----------|----------|----------|----------|----------|
| Conformer-Large-CTC   | 4.53   | 2.19   | 23.04   | 5.94   |
| Conformer-Large-CTC with local attention (64, 64)  | 5.39   | 2.57   | 23.89   | 6.2  |
| Conformer-Large-CTC with local attention (192, 192)  | 4.53   |  2.19 | 23.2   | 5.96  |
| Conformer-Large-RNNT  | 3.73  | 1.8  | 21.69  | 5.4  |
| Conformer-Large-RNNT with local attention (64, 64) | 4.6  | 2.14  | 21.85  | 5.67  |
| Conformer-Large-RNNT with local attention (192, 192) | 3.75  | 1.83 | 21.64  | 5.41 |

**Collection**: ASR

# Changelog 
- adds option to use local attention for Conformer
- adds option for overriding attention parameters when running transcribe or eval scripts, or after loading model

# Usage

With an eval script can be used like this to switch to local attention with any Conformer checkpoint

```
python speech_to_text_eval.py \
  (...other params...)  \
  ++model_change.conformer.self_attention_model="rel_pos_local_attn" \
  ++model_change.conformer.att_context_size=[64, 64]
```

Or after loading model 

```python
asr_model = ASRModel.from_pretrained('stt_en_conformer_ctc_large')
asr_model.change_conformer_attention_model(
                self_attention_model="rel_pos_local_attn",
                att_context_size=[64, 64]
)
```

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [x] Documentation


examples/asr/transcribe_speech.py
```diff
@@ -116,6 +116,9 @@ class TranscriptionConfig:
     # Decoding strategy for RNNT models
     rnnt_decoding: RNNTDecodingConfig = RNNTDecodingConfig(fused_batch_size=-1)
 
+    # This will be merged with the model config, overriding existing parameters.
+    merge_into_model_config: Optional[Any] = None
```
titu1994, line 120: I do not think we should add such an argument. Merging of configs is a very flaky feature, if you try to add keys that didn't exist in the main config it will fail, it requires subconfig knowledge. 

In the name of making it simpler we are adding loads of complexity + chance of failure. 

Users can and see recommended to use the restore_config=True, update config, then override_cfg workflow. 

Please revert the changes wrt partial merge.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,25 +379,26 @@ def set_max_audio_length(self, max_audio_length):
...
+                if self.att_context_size[1] >= 0:
+                    att_mask = att_mask.tril(diagonal=self.att_context_size[1])
+            else:
+                chunk_idx = torch.arange(0, max_audio_length, dtype=torch.int, device=att_mask.device)
+                chunk_idx = torch.div(chunk_idx, self.chunk_size, rounding_mode="trunc")
+                diff_chunks = chunk_idx.unsqueeze(1) - chunk_idx.unsqueeze(0)
+                chunked_limited_mask = torch.logical_and(
+                    torch.le(diff_chunks, self.left_chunks_num), torch.ge(diff_chunks, 0)
+                )
+                att_mask = torch.logical_and(att_mask, chunked_limited_mask.unsqueeze(0))
```
titu1994, line 396: Else section should set local variables to None

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -256,6 +259,250 @@ def forward(self, query, key, value, mask, pos_emb, cache=None, cache_next=None)
         return out
 
 
+class RelPositionMultiHeadAttentionLocal(RelPositionMultiHeadAttention):
```
titu1994, line 262: Add citation for the local attention part

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -256,6 +259,250 @@ def forward(self, query, key, value, mask, pos_emb, cache=None, cache_next=None)
...
+            dropout_rate=dropout_rate,
+            pos_bias_u=pos_bias_u,
+            pos_bias_v=pos_bias_v,
+            max_cache_len=max_cache_len,
+        )
+        self.att_context_size = att_context_size
+
+    # Longformer implementation for no overlap case adapted for arbitrary left and right chunk size
+    # https://github.com/allenai/longformer/blob/master/longformer/sliding_chunks.py
+    def _skew(self, x, direction, padding_value):
```
titu1994, line 289: Need docstring explaining what all methods below are doing
Plus, please note the symbol shapes of each operation how it's done in relpos conformer code, so that at a glance we can tell what shapes are being operated on. (yes it's a lot of work, but it helps a ton for future)

Or refer to parts of the paper eqn for the steps / code in longformer repo

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -256,6 +259,250 @@ def forward(self, query, key, value, mask, pos_emb, cache=None, cache_next=None)
...
+        diagonal_attn = diagonal_attn.view(bsz, num_heads, seqlen, 2 * w + 1)
+
+        self.mask_invalid_locations(diagonal_attn, w, 1, False)
+        return diagonal_attn
+
+    def sliding_chunks_matmul_pv(self, prob: torch.Tensor, v: torch.Tensor, w: int):
+        '''Same as sliding_chunks_matmul_qk but for prob and value tensors. It is expecting the same output
+        format from sliding_chunks_matmul_qk'''
+        bsz, num_heads, seqlen, head_dim = v.size()
+        assert seqlen % (w * 2) == 0
```
titu1994, line 414: Instead of assert, do if check and raise appropriately warning in code

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -256,6 +259,250 @@ def forward(self, query, key, value, mask, pos_emb, cache=None, cache_next=None)
...
+            query, key, value = query.to(torch.float32), key.to(torch.float32), value.to(torch.float32)
+
+        # temporary until we solve this more gracefully
+        with avoid_float16_autocast_context():
+
+            q, k, v = self.forward_qkv(query, key, value)
+            n_batch, _, T, _ = q.size()
+
+            w = max(self.att_context_size[0], self.att_context_size[1])
+            assert w > 0
```
titu1994, line 452: Use if and raise exception

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -362,3 +609,52 @@ def forward(self, x, cache_len=0):
...
+    def extend_pe(self, length, device):
+        """Reset and extend the positional encodings only at the beginning"""
+        if hasattr(self, 'pe'):
+            return
+
+        positions = torch.arange(
+            self.left_context, -self.right_context - 1, -1, dtype=torch.float32, device=device
+        ).unsqueeze(1)
+        self.create_pe(positions=positions)
+        self.center_pos = torch.tensor(self.left_context, dtype=torch.int32, device=device)
```
titu1994, line 641: I believe Boris updated it so this is no longer cached, please use the current Nemo code to see how to compute it on the fly so if can be traced properly

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -184,13 +184,15 @@ def setup_model(cfg: DictConfig, map_location: torch.device) -> Tuple[ASRModel,
         imported_class = model_utils.import_class_by_path(classpath)  # type: ASRModel
         logging.info(f"Restoring model : {imported_class.__name__}")
         asr_model = imported_class.restore_from(
-            restore_path=cfg.model_path, map_location=map_location
+            restore_path=cfg.model_path, map_location=map_location, merge_into_model_config=cfg.merge_into_model_config
         )  # type: ASRModel
```
titu1994, line 188: Revert this arg

titu1994 (changes requested):
The code for local attn looks great, minor comments here and there for docstrings and input argument shapes. 

Please revert all changes related to merging config. We have established workflow for that without adding more flaky version with config merging.

Please add tests for instantiating a Conformer with this config (one layer is fine, check the CTC test suite)

Is it possible to change preexisting conformer to local attention ? If so will that be added in another pr ?

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,25 +379,26 @@ def set_max_audio_length(self, max_audio_length):
...
-                att_mask = att_mask.tril(diagonal=self.att_context_size[1])
-        else:
-            chunk_idx = torch.arange(0, max_audio_length, dtype=torch.int, device=att_mask.device)
-            chunk_idx = torch.div(chunk_idx, self.chunk_size, rounding_mode="trunc")
-            diff_chunks = chunk_idx.unsqueeze(1) - chunk_idx.unsqueeze(0)
-            chunked_limited_mask = torch.logical_and(
-                torch.le(diff_chunks, self.left_chunks_num), torch.ge(diff_chunks, 0)
-            )
-            att_mask = torch.logical_and(att_mask, chunked_limited_mask.unsqueeze(0))
+        if self.use_att_mask:
```
VahidooX, line 382: How about just check the self_attention_model here and skip the part if the it is rel_pos_local_attn to avoid creating a new member?

VahidooX: You may set the att_mask to None when not used and then check if it None or not in other places.

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -89,6 +90,16 @@ def __init__(
                 pos_bias_v=pos_bias_v,
                 max_cache_len=MHA_max_cache_len,
             )
+        elif self_attention_model == 'rel_pos_local_attn':
```
VahidooX, line 93: How about using the longformer instead of local attention in the names as we already had local attention feature? This one is an sufficient implementation of that which is an adaptation of LongFormer for relative positional encoding. What do you think?

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -146,6 +157,16 @@ def forward(
                 cache=cache_last_channel,
                 cache_next=cache_last_channel_next,
             )
+        elif self.self_attention_model == 'rel_pos_local_attn':
```
VahidooX, line 160: What is the different between 'rel_pos' and 'rel_pos_local_attn' when calling the forward? If there is no difference, let's merge them in one case.

nemo/collections/asr/parts/submodules/multi_head_attention.py
VahidooX: And please also add some explanation on the difference between this one and RelPositionMultiHead.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -256,6 +259,250 @@ def forward(self, query, key, value, mask, pos_emb, cache=None, cache_next=None)
...
+            n_head=n_head,
+            n_feat=n_feat,
+            dropout_rate=dropout_rate,
+            pos_bias_u=pos_bias_u,
+            pos_bias_v=pos_bias_v,
+            max_cache_len=max_cache_len,
+        )
+        self.att_context_size = att_context_size
+
+    # Longformer implementation for no overlap case adapted for arbitrary left and right chunk size
```
VahidooX, line 287: Is it the overlapping or the non-overlapping one?
Overlapping one gave better accuracy and the same output with RelPositionMultiHeadedAttention right?
Shouldn't we add the one instead?
Also if we plan to add the other version in future, we may need to mention the overlapping or non-overlapping in the names and docs.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -362,3 +609,52 @@ def forward(self, x, cache_len=0):
         if self.dropout_emb:
             pos_emb = self.dropout_emb(pos_emb)
         return self.dropout(x), pos_emb
+
+
+class ChunkedRelPositionalEncoding(PositionalEncoding):
```
VahidooX, line 614: Would you please add a reference to the paper and some description on what it does?

VahidooX: If this class inherits from the relative positional encoding class, can you drop some of the methods here  if they are the same as the upper class?

VahidooX (changes requested):
We need to add the support of this Longformer feature to the main README of nemo along with somewhere in the docs. Also please add some description of this feature, what it does and when to use it, citation to the longformer code and paper, and numbers and comparisons with regular attention in the PR detail section.

examples/asr/transcribe_speech.py
sam1373: Reverted, added instead method for changing model in utils

nemo/collections/asr/parts/submodules/multi_head_attention.py
sam1373: Added more docstrings and hints

nemo/collections/asr/parts/submodules/multi_head_attention.py
sam1373: For this class this would only be updated once anyway (as length doesn't matter), so I don't think there's a problem

nemo/collections/asr/parts/submodules/conformer_modules.py
sam1373: For now I called this local_attn because it's missing global attention tokens from longformer. I was thinking if we add that later it would be more correct to call that option longformer

nemo/collections/asr/parts/utils/transcribe_utils.py
sam1373: Removed this way of merging

nemo/collections/asr/parts/submodules/conformer_modules.py
sam1373: There is difference in what is passed to mask, merged into one

nemo/collections/asr/parts/submodules/multi_head_attention.py
sam1373: This is overlapping, fixed some of the comments. Since this is default in longformer and I'm not sure if we need to add non-overlapping for now I thought to not add it to names for simplicity

nemo/collections/asr/parts/submodules/multi_head_attention.py
sam1373: The methods aren't the same

VahidooX: Let's give a hint or ref to the longformer and mention it can be used with which self-attention class.

VahidooX (changes requested):

nemo/collections/asr/parts/utils/asr_module_utils.py
```diff
@@ -80,3 +80,124 @@ def _update_se_context_window(model: 'ASRModel', context_window: int, cfg: Optio
             # update config
             if cfg is not None:
                 cfg.jasper[jasper_block_counter].se_context_size = context_window
+
+
+def change_conformer_attention_model(
```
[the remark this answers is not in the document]
sam1373: Do you mean to the ASRModel class?

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -184,13 +184,18 @@ def setup_model(cfg: DictConfig, map_location: torch.device) -> Tuple[ASRModel,
         imported_class = model_utils.import_class_by_path(classpath)  # type: ASRModel
         logging.info(f"Restoring model : {imported_class.__name__}")
         asr_model = imported_class.restore_from(
-            restore_path=cfg.model_path, map_location=map_location
+            restore_path=cfg.model_path, map_location=map_location,
         )  # type: ASRModel
+        if cfg.get("change_self_attention_model", None) is not None:
```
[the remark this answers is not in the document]
sam1373: Sure but it would be pretty inconvenient for the user to have to create a separate script just to call that

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -478,16 +502,20 @@ def _create_masks(self, max_audio_length, padding_length, device):
...
         ) < padding_length.unsqueeze(-1)
 
-        # pad_mask_for_att_mask is the mask which helps to ignore paddings
-        pad_mask_for_att_mask = pad_mask.unsqueeze(1).repeat([1, max_audio_length, 1])
-        pad_mask_for_att_mask = torch.logical_and(pad_mask_for_att_mask, pad_mask_for_att_mask.transpose(1, 2))
-        # att_mask is the masking to be used by the MHA layers to ignore the tokens not supposed to be visible
-        att_mask = self.att_mask[:, :max_audio_length, :max_audio_length]
-        # paddings should also get ignored, so pad_mask_for_att_mask is used to ignore their corresponding scores
-        att_mask = torch.logical_and(pad_mask_for_att_mask, att_mask.to(pad_mask_for_att_mask.device))
+        if self.att_mask is not None:
```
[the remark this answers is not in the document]
sam1373: The attention mask gets automatically updated to match maximum sequence length passed, so this would be an issue when using local attention with long audio. I guess we can have a separate parameter for if we want to keep attention mask, but for now it would be pointless anyway since all layers are either full or local

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -263,7 +270,7 @@ def __init__(
                     feat_in=feat_in,
                     feat_out=d_model,
                     conv_channels=subsampling_conv_channels,
-                    activation=nn.ReLU(),
+                    activation=nn.ReLU(True),
```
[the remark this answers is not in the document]
sam1373: I've not seen any issues, and with local attention this increases max duration by ~30% in some cases

nemo/collections/asr/parts/mixins/mixins.py
```diff
@@ -394,6 +394,27 @@ def change_conv_asr_se_context_window(self, context_window: int, update_config:
...
+            self_attention_model (str): type of the attention layer and positional encoding
+                'rel_pos': relative positional embedding and Transformer-XL
+                'rel_pos_local_attn': relative positional embedding and Transformer-XL with local attention using
+                    overlapping windows. Attention context is determined by att_context_size parameter.
+                'abs_pos': absolute positional embedding and Transformer
+            att_context_size (List[int]): List of 2 ints corresponding to left and right attention context sizes,
+                or None for full context.
+            update_config: Whether to update the config or not with the new attention model
+        """
+        asr_module_utils.change_conformer_attention_model(
```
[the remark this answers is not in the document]
sam1373: I did it this way since this is what we have for change_conv_asr_se_context_window

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,25 +379,28 @@ def set_max_audio_length(self, max_audio_length):
...
 
-        if hasattr(self, 'att_mask'):
-            self.att_mask = att_mask
+            if hasattr(self, 'att_mask'):
+                self.att_mask = att_mask
+            else:
+                self.register_buffer('att_mask', att_mask, persistent=False)
         else:
-            self.register_buffer('att_mask', att_mask, persistent=False)
+            self.att_mask = None
```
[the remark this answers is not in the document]
sam1373: When changing to regular att_mask is created afterwards when calling set_max_length

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -429,14 +449,18 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
             cache_last_channel_next = None
             cache_len = 0
 
-        audio_signal, pos_emb = self.pos_enc(x=audio_signal, cache_len=cache_len)
+        if self.self_attention_model == 'abs_pos':
+            audio_signal, pos_emb = self.pos_enc(x=audio_signal)
```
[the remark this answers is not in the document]
sam1373: It doesn't seem to. We just had an error when switching to abs_pos

nemo/collections/asr/parts/utils/asr_module_utils.py
```diff
@@ -12,12 +12,12 @@
...
 
-from typing import Optional
+from typing import List, Optional
 
 from omegaconf import DictConfig, open_dict
 
-from nemo.collections.asr.modules import conv_asr
-from nemo.collections.asr.parts.submodules import jasper
+from nemo.collections.asr.modules import conformer_encoder, conv_asr
+from nemo.collections.asr.parts.submodules import conformer_modules, jasper, multi_head_attention
```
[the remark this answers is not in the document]
VahidooX: Some imports need to get dropped here.

nemo/collections/asr/parts/utils/transcribe_utils.py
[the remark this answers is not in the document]
VahidooX: Isn't this how a user can use this feature?
```
asr_model = ASRModel.get_pretrained(...)
asr_model.change_attention_model("rel_att_local", [100,100])
asr_model.transcribe(...)
```

Why do they need a script?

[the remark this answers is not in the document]
sam1373: Yes, I just mean they would need to write these lines somewhere themselves

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,25 +384,28 @@ def set_max_audio_length(self, max_audio_length):
...
-                att_mask = att_mask.tril(diagonal=self.att_context_size[1])
-        else:
-            chunk_idx = torch.arange(0, max_audio_length, dtype=torch.int, device=att_mask.device)
-            chunk_idx = torch.div(chunk_idx, self.chunk_size, rounding_mode="trunc")
-            diff_chunks = chunk_idx.unsqueeze(1) - chunk_idx.unsqueeze(0)
-            chunked_limited_mask = torch.logical_and(
-                torch.le(diff_chunks, self.left_chunks_num), torch.ge(diff_chunks, 0)
-            )
-            att_mask = torch.logical_and(att_mask, chunked_limited_mask.unsqueeze(0))
+        if self.self_attention_model != "rel_pos_local_attn":
```
titu1994, line 387: We need to refactor this code @VahidooX - make it functions and separate most of this logic out

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -478,16 +507,20 @@ def _create_masks(self, max_audio_length, padding_length, device):
...
+            att_mask = self.att_mask[:, :max_audio_length, :max_audio_length]
+            # paddings should also get ignored, so pad_mask_for_att_mask is used to ignore their corresponding scores
+            att_mask = torch.logical_and(pad_mask_for_att_mask, att_mask.to(pad_mask_for_att_mask.device))
+
+            att_mask = ~att_mask
+        else:
+            att_mask = None
 
         pad_mask = ~pad_mask
-        att_mask = ~att_mask
```
titu1994, line 490: How can att_mask be none??

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -610,6 +643,123 @@ def get_initial_cache_state(self, batch_size=1, dtype=torch.float32, device=None
...
+            update_config (bool): Whether to update the config or not with the new attention model.
+                Defaults to True.
+            device (torch.device): If provided, new layers will be moved to the device.
+                Defaults to None.
+        """
+
+        if att_context_size:
+            att_context_size = list(att_context_size)
+        else:
+            att_context_size = self._cfg.att_context_size
```
titu1994, line 675: Do NOT access config during forward pass. Cache the config values into self.xyz and only access that.
If this is called outside forward just once, that is fine.

tests/collections/asr/test_asr_local_attn.py
```diff
@@ -0,0 +1,40 @@
...
+import pytest
+import torch
+
+from nemo.collections.asr.models import ASRModel
+
+
+class TestASRLocalAttention:
+    @pytest.mark.with_downloads()
+    @pytest.mark.unit
+    def test_forward(self):
```
titu1994, line 24: Please added another test to change the self attention, then save in first temp dir, move .nemo to another dir, delete first temp dir, and restore in second dir. 
Assert that the changes made were reflected in the model - you may need to update the model.cfg inside of nemo/collections/asr/parts/mixins/mixins.py

titu1994 (changes requested):

nemo/collections/asr/modules/conformer_encoder.py
sam1373: The att_mask isn't used for this approach, and in fact we can't even create it since we want to avoid quadratic memory. We do still have pad_mask

tests/collections/asr/test_asr_local_attn.py
sam1373: Added

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Hmm it might cause problems for the middle downsampling but well deal with it later.

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Ok thats good then, btw why is a module maintaining a config under _cfg? You should call it something else possibly cause its specific to modification cfg

titu1994 (approved):
Awesome work, thanks ! ready to merge

titu1994 (approved):

VahidooX (approved):

PR #5532, opened 2022-12-01
TTS tutorial update: use speaker 9017 instead of 6097

redoctopus:

# What does this PR do ?

Replaces speaker 6097's data with speaker 9017's since the latter has given explicit permission.

**Collection**: TTS
# Changelog 
- Updated link to data and recalculated pitch stats for the new speaker.

# Usage
* Tutorial should run as usual.

### DIFF
--- a/requirements/requirements_lightning.txt
+++ b/requirements/requirements_lightning.txt
@@ -2,7 +2,7 @@ hydra-core>=1.2.0,<1.3
 omegaconf>=2.2,<2.3
 pytorch-lightning>=1.8.3
 pyyaml<6  # Pinned until omegaconf works with pyyaml>=6
-torchmetrics>=0.4.1rc0
+torchmetrics>=0.4.1rc0,<=0.10.3
 transformers>=4.0.1,<=4.21.2
 wandb
 webdataset>=0.1.48,<=0.1.62


tutorials/tts/FastPitch_Finetuning.ipynb
```diff
@@ -326,7 +326,7 @@
                 "`\n",
```
ekmb, line 326: In general, it's better to use this whitelist: `https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/text_normalization/en/data/whitelist/tts.tsv` , not needed for hi-fi tts as it's already normalized

ekmb (dismissed):
Why pitch statistics calculation is outside of the scope of this tutorial?  `model.pitch_mean=152.3 model.pitch_std=64.0 model.pitch_fmin=30 model.pitch_fmax=512`

tutorials/tts/FastPitch_Finetuning.ipynb
redoctopus: Should be a quick change. Would you prefer it in this PR or a new one?

ekmb: up to you. Some checks are still running - hmm, it's been 19+ hrs since the last commit

ekmb (approved):

PR #5551, opened 2022-12-05
Add kwargs to TE full activation recompute path

ksivaman:
Signed-off-by: Kirthi Shankar Sivamani <ksivamani@nvidia.com>

# What does this PR do ?

`te_checkpoint` now has support for kwargs. This is necessary for the `is_first_microbatch` keyword argument during forward pass without which the numerics of the forward pass in the normal vs recompute phase differ.


ericharper (approved):
LGTM. Thanks!

ericharper (changes requested):
We still need to test this change further. When running with 22.11 container there is an error.

ksivaman:
[This line](https://github.com/NVIDIA/NeMo/blob/dd4a041d919dae918244e3ecb0fae15b7c29d5f5/nemo/collections/nlp/modules/common/megatron/transformer.py/#L2299) takes care of passing the correct `is_first_microbatch` parameter during the activation recompute phase. Closing.

PR #5552, opened 2022-12-06
Fix ASR Buffered inference scripts

titu1994:
# What does this PR do ?

**Collection**: [ASR]

# Changelog 
- Deepcopy the alignments before compute of alignments
- Do proper class level check of log level instead of incorrect config checks (older model doesn't have in config because the log=True is set by default inside of the function)

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


fayejf (approved):
LGTM. one thing we can discuss and add later: Could we add more tests regarding buffered inference? Is it necessary?

PR #5565, opened 2022-12-07
[TTS] GAN-based spectrogram enhancer

racoiaws:
# What does this PR do ?

Add a GAN-based spectrogram enhancer to NeMo.

The intended purpose is to add details to blurry spectrograms from TTS models like FastPitch.

**Collection**: TTS

# Changelog 
- Add the spectrogram enhancer NeMo model
- Add example config and training script

# Usage
Given a trained model,
```python
model = SpectrogramEnhancer.restore_from("checkpoint.nemo")
with torch.no_grad():
    # spectrograms :: B x 80 x L
    # lengths  :: B - the usual LongTensor with lengths or every spectrogram in the batch
    enhanced_spectrograms = model.forward(condition=spectrograms, lengths=lengths)
    # enhanced_spectrograms = B x 80 x L
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Based on https://github.com/lucidrains/stylegan2-pytorch (MIT License)


racoiaws:
Example: FastPitch trained on LibriTTS train-clean-100 + enhancer)
1475274b-8f1b-4c4d-a1eb-53a78b14f32a

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# The following is largely based on code from https://github.com/lucidrains/stylegan2-pytorch
+
+import math
+from random import random, randrange
+from typing import Dict, List, Optional, TypedDict, Union
```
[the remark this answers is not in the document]
subhankar-ghosh: Remove unused imports.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        return enhanced_spectrograms
+
+    def prepare_batch(self, batch):
+        if not isinstance(self.spectrogram_model, FastPitchModel):
+            raise NotImplementedError(
+                f"{type(self.spectrogram_model)} is not supported, please implement batch preparation for this model."
+            )
+
+        attn_prior, durs, speaker = None, None, None
+        (audio, audio_lens, text, text_lens, attn_prior, pitch, _, speaker,) = batch
```
subhankar-ghosh, line 240: Use this function instead for uniformity: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/helpers/helpers.py#L564

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        target = nn.utils.rnn.pad_sequence(targets, batch_first=True)
+        condition = nn.utils.rnn.pad_sequence(conditions, batch_first=True)
+        lengths = torch.LongTensor(lengths).unsqueeze(-1)
+
+        target = rearrange(target, "b l c -> b 1 c l")
+        condition = rearrange(condition, "b l c -> b 1 c l")
+
+        return self.move_to_correct_device((target, condition, lengths))
+
+    def training_step(self, batch, batch_idx, optimizer_idx):
```
subhankar-ghosh, line 281: You should add a corresponding `validation_step` as well.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+            network_capacity=cfg["network_capacity"],
+            style_depth=cfg["style_depth"],
+            channels=1,
+            fmap_max=cfg["fmap_max"],
+        )
+        self.D = Discriminator(
+            n_bands=cfg["n_bands"], network_capacity=cfg["network_capacity"], channels=1, fmap_max=cfg["fmap_max"],
+        )
+
+    def init_spectrogram_model(self):
```
subhankar-ghosh, line 147: Do we really need a `spectrogram_model`, can this model not get spectrograms directly from the dataloader? Look at Vocoder dataloader. Or if you want to use TTSDataset class then I think you can use "log_mel". The spectrograms can be precomputed, this should remove dependency on any particular spectrogram generator.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+    """GAN-based model to add details to blurry spectrograms from TTS models like Tacotron or FastPitch."""
+
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None) -> None:
+        self.spectrogram_model = None
+        super().__init__(cfg=cfg, trainer=trainer)
+
+        self.init_spectrogram_model()
+
+        self.G = Generator(
+            n_bands=cfg["n_bands"],
```
subhankar-ghosh, line 136: try to maintain uniformity across all models by using `cfg.n_bands` instead of `cfg["n_bands"]`

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        lengths = []
+
+        for i in range(batch_size):
+            l = mel_lens[i].item()
+            lengths.append(l)
+
+            targets.append(mels[i, :, :l].T)
+            conditions.append(mels_pred[i, :, :l].T)
+
+        target = nn.utils.rnn.pad_sequence(targets, batch_first=True)
```
subhankar-ghosh, line 272: Best to keep these kind of padding and data manipulation logic in dataloader.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+    )[0]
+
+    return (pl_grads ** 2).sum(dim=2).mean(dim=1).sqrt()
+
+
+def gen_hinge_loss(fake):
+    return fake.mean()
+
+
+def hinge_loss(real, fake):
```
subhankar-ghosh, line 110: loss should be part of the class. like `self.loss_fn = hinge_loss`
Also if this is simple hinge loss then why not use predefined hinge loss from torch? If it is your own version (not exactly hinge loss) then name it accordingly. It can be confusing to users.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+    return (F.relu(1 + real) + F.relu(1 - fake)).mean()
+
+
+def consistency_loss(condition, output, lengths):
```
subhankar-ghosh, line 114: same as above. All losses should be part of the class.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        if optimizer_idx == 0:
+            enhanced_spectrograms = self.forward(condition, lengths, mixing=True, normalize=False)
+            fake_logits = self.D(enhanced_spectrograms, condition, lengths)
+
+            real_images = target.requires_grad_()
+            real_logits = self.D(real_images, condition, lengths)
+            d_loss = hinge_loss(real=real_logits, fake=fake_logits)
+            self.log("d_loss", d_loss, prog_bar=True)
+
+            if batch_idx % 4 == 0:
```
subhankar-ghosh, line 301: What is the logic behind this if statement? Have a comment/note explaining this. Or have a method with the proper name. Like:
```
def if_<condition_name>(self, batch_idx):
    return batch_idx % 4 == 0

if self.if_<condition_name>(batch_idx):
   ...
```

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+                return d_loss + gp_loss
+
+            return d_loss
+
+        # train generator
+        if optimizer_idx == 1:
+            enhanced_spectrograms = self.forward(condition, lengths, mixing=True, normalize=False)
+            fake_logits = self.D(enhanced_spectrograms, condition, lengths)
+            g_loss = gen_hinge_loss(fake_logits)
+            c_loss = 10 * consistency_loss(condition, enhanced_spectrograms, lengths)
```
subhankar-ghosh, line 313: reason behind 10? Hardcoding should be avoided. Make it a class variable with proper name.
loss should be part of class.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        grad_outputs=torch.ones(outputs.shape, device=device),
+        create_graph=True,
+        retain_graph=True,
+        only_inputs=True,
+    )[0]
+
+    return (pl_grads ** 2).sum(dim=2).mean(dim=1).sqrt()
+
+
+def gen_hinge_loss(fake):
```
subhankar-ghosh, line 106: name of the method and what it is doing are not the same. can be confusing. You can get rid of the method and replace with just `variable.mean()`

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+
+            self.log("g_loss", g_loss, prog_bar=True)
+            self.log("c_loss", c_loss, prog_bar=True)
+
+            with torch.no_grad():
+                self.log_illustration(target, condition, enhanced_spectrograms, lengths)
+            return g_loss + c_loss
+
+    def configure_optimizers(self):
+        generator_opt = torch.optim.Adam(self.G.parameters(), lr=self._cfg["lr"], betas=(0.5, 0.9),)
```
subhankar-ghosh, line 323: optimizers should be initialized from `cfg` and hardcoding needs to be avoided.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+            self._train_dl = self.spectrogram_model._train_dl
+
+    def setup_validation_data(self, val_data_config):
+        if self.spectrogram_model and val_data_config:
+            self.spectrogram_model.setup_validation_data(val_data_config)
+            self._validation_dl = self.spectrogram_model._validation_dl
+
+    @classmethod
+    def list_available_models(cls):
+        return None
```
subhankar-ghosh, line 339: return an empty list.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+    def pad_spectrogram(self, spectrogram):
+        multiplier = 2 ** sum(1 for block in self.G.blocks if block.upsample)
+        *_, max_length = spectrogram.shape
+        return torch.nn.functional.pad(spectrogram, (0, multiplier - max_length % multiplier))
+
+    def forward(
+        self,
+        condition: Tensor,
+        lengths: Tensor,
+        zs: Optional[List[Tensor]] = None,
```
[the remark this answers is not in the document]
subhankar-ghosh: All parameters need dimension and explanation. Naming can be better - what is zs, ws, etc. needs to be explained.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+        spectrogram = spectrogram - self._cfg["spectrogram_min_value"]
+        spectrogram = spectrogram / (self._cfg["spectrogram_max_value"] - self._cfg["spectrogram_min_value"])
+        return spectrogram
+
+    def unnormalize_spectrograms(self, spectrogram: Tensor) -> Tensor:
+        spectrogram = spectrogram * (self._cfg["spectrogram_max_value"] - self._cfg["spectrogram_min_value"])
+        spectrogram = spectrogram + self._cfg["spectrogram_min_value"]
+        return spectrogram
+
+    def generate_zs(self, batch_size: int = 1, mixing: bool = False):
```
subhankar-ghosh, line 168: naming can be better - zs?

nemo/collections/tts/modules/spectrogram_enhancer.py
```diff
@@ -0,0 +1,355 @@
...
+# SOFTWARE.
+
+# The following is largely based on code from https://github.com/lucidrains/stylegan2-pytorch
+
+import math
+from functools import partial
+from math import log2
+from typing import List
+
+import torch
```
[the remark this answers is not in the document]
subhankar-ghosh: Needs to be taken care of.

requirements/requirements_tts.txt
```diff
@@ -1,6 +1,7 @@
 attrdict
 einops
 inflect
+kornia
```
subhankar-ghosh, line 4: we need tests to verify that this does not affect other dependencies and break any test cases.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,382 @@
...
+class SpectrogramEnhancerModel(ModelPT, Exportable):
+    """GAN-based model to add details to blurry spectrograms from TTS models like Tacotron or FastPitch."""
+
+    def __init__(self, cfg: DictConfig, trainer: Trainer = None) -> None:
+        self.spectrogram_model = None
+        super().__init__(cfg=cfg, trainer=trainer)
+
+        self.init_spectrogram_model()
+
+        self.G = Generator(
```
subhankar-ghosh, line 135: G, D needs to be instantiated from the yaml config.

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: I'm not convinced validation step is necessary here - all of the losses are rather uninformative in deciding which model is "better"

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: This is a major part of the training process, we can't do without a spectrogram model
sg2-enhancer-training-150

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Good catch, this is supposed to be a parameter

requirements/requirements_tts.txt
racoiaws: How would one do that? CI says tests pass

It is used for a single op though, could be rewritten in PyTorch

nemo/collections/tts/modules/spectrogram_enhancer.py
```diff
@@ -0,0 +1,355 @@
...
+        if prev_rgb is not None:
+            x = x + prev_rgb
+
+        if self.upsample is not None:
+            x = self.upsample(x)
+
+        return x
+
+
+class Conv2DMod(nn.Module):
```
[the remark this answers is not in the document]
racoiaws: Modulated convolution - https://arxiv.org/abs/1912.04958

Should I rename it to Conv2DModulated?

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: I initially wrote it as a separate dataloader and gave up after some debugging. 
In this implementation data loading is delegated to underlying spectrogram generator. Which ended up much more robust and clean.

Should I move it to its own class? Still would be a regular class, not a torch/nemo dataset or dataloader or collate_fn

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: You mean passing a DictConfig to them? What is the rationale? Neither of them are ModelPT

nemo/collections/tts/modules/spectrogram_enhancer.py
```diff
@@ -0,0 +1,355 @@
...
+
+        for ind, (in_chan, out_chan) in enumerate(chan_in_out):
+            is_not_last = ind != (len(chan_in_out) - 1)
+
+            block = DiscriminatorBlock(in_chan, out_chan, downsample=is_not_last)
+            blocks.append(block)
+
+        self.blocks = nn.ModuleList(blocks)
+
+        chan_last = filters[-1]
```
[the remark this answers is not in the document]
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Did not find anything named "hinge loss" in the docs

This is copied from https://github.com/lucidrains/stylegan2-pytorch

racoiaws: Moved to a class and made a field inside the model

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: I'm on the fence on this - official and unofficial implementations of StyleGAN use G and D, so it reads easier for people familiar with that code

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,395 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# The following is largely based on code from https://github.com/lucidrains/stylegan2-pytorch
+
+import math
+from random import random, randrange
+from typing import List, Optional
+
+import torch
```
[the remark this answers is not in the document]
racoiaws: I do not understand what is the issue here

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,395 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# The following is largely based on code from https://github.com/lucidrains/stylegan2-pytorch
+
+import math
```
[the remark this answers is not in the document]
racoiaws: Will be fixed in next push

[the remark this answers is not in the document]
racoiaws: Done

tests/collections/tts/test_spectrogram_enhancer.py
```diff
@@ -0,0 +1,186 @@
...
+        "discriminator_opt": {"_target_": "torch.optim.Adam", "lr": 2e-4, "betas": [0.5, 0.9]},
+    }
+
+    return DictConfig(config)
+
+
+@pytest.fixture
+def fastpitch_model():
+    model = FastPitchModel.from_pretrained(model_name="tts_en_fastpitch")
+    return model
```
redoctopus, line 86: I wonder if this should be in a separate test file since this does just test FastPitch NGC checkpoint loading in general.

tests/collections/tts/test_spectrogram_enhancer.py
```diff
@@ -0,0 +1,186 @@
...
+# @pytest.mark.unit
+def enhancer_with_fastpitch_training_step_generator(enhancer_with_fastpitch: SpectrogramEnhancerModel):
+    batch = next(iter(enhancer_with_fastpitch._train_dl))
+    enhancer_with_fastpitch.training_step(batch, batch_idx=0, optimizer_idx=1)
+
+
+@pytest.mark.with_downloads
+@pytest.mark.unit
+def test_enhancer_with_fastpitch(enhancer_with_fastpitch: SpectrogramEnhancerModel):
+    # long setup for enhancer_with_fastpitch, around 1 min
```
redoctopus, line 183: Should this be a unit test if it takes 1m+ to set up and run?

racoiaws: Good question! Do we have a working way to say THIS TAKES LONG in our testing setup? The `.unit` mark here is just copy-paste

tests/collections/tts/test_spectrogram_enhancer.py
racoiaws: This not a test, this is a fixture, a fixture is not supposed to test something by itself

Is designing reusable testing code something worth of effort? I've been taught e.g. copying-pasting tests is an OK thing to do

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+            "normalize": NeuralType(None, BoolType(), optional=True),
+        }
+    )
+    def forward(
+        self, *, condition: Tensor, lengths: Tensor, mixing: bool = False, normalize: bool = True,
+    ):
+        """
+        Generator forward pass. Noise inputs will be generated.
+
+        condition: batch of blurred spectrograms
```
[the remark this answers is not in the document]
racoiaws: Since it came up in the review process, it's probably too abstract. I'll rename it to something more concrete, `blurred_spec` sounds good.

examples/tts/conf/spectrogram-enhancer.yaml
```diff
@@ -0,0 +1,67 @@
...
+  lr: 2e-4
+  n_bands: 80
+  latent_dim: 192
+  style_depth: 4
+  network_capacity: 16
+  mixed_prob: 0.9
+  fmap_max: 192
+
+  consistency_loss_weight: 10.0  # less for clean datasets, higher for noisier
+  gradient_penalty_loss_weight: 10.0
```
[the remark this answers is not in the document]
racoiaws: For consistency loss, 1.0 to 20.0 is probably a good range (eyeballing). Will put it in the comment.

For gradient penalty loss, I honestly have no idea. This is the default value from the original repo and it works

examples/tts/conf/spectrogram-enhancer.yaml
```diff
@@ -0,0 +1,67 @@
...
+  gradient_penalty_loss_every_n_steps: 4
+
+  # Needed for treating spectrograms as images with pixel values around [0, 1].
+  # Given values are for LibriTTS, might still work for your data.
+  # In case of training not converging, calculate min and max values over your TTSDataset and update the config.
+  spectrogram_min_value: -13.18
+  spectrogram_max_value: 4.78
```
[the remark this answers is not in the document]
racoiaws: Will do

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def normalize_spectrograms(self, spectrogram: Tensor) -> Tensor:
+        spectrogram = spectrogram - self._cfg.spectrogram_min_value
+        spectrogram = spectrogram / (self._cfg.spectrogram_max_value - self._cfg.spectrogram_min_value)
+        return spectrogram
+
+    def unnormalize_spectrograms(self, spectrogram: Tensor) -> Tensor:
+        spectrogram = spectrogram * (self._cfg.spectrogram_max_value - self._cfg.spectrogram_min_value)
+        spectrogram = spectrogram + self._cfg.spectrogram_min_value
+        return spectrogram
```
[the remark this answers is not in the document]
racoiaws: Personally not a fan of abstracting one-liners like this. Makes me jump around the project and lose context.

Will surely make the `norm(unnorm(x)) == x` and `unnorm(norm(x)) == x` tests

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+    def __call__(self, fake_logits):
+        return fake_logits.mean()
+
+
+class HingeLoss(nn.Module):
+    def __call__(self, real_logits, fake_logits):
+        return (F.relu(1 + real_logits) + F.relu(1 - fake_logits)).mean()
+
+
+class ConsistencyLoss(nn.Module):
```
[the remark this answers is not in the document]
racoiaws: Will do

tests/collections/tts/test_spectrogram_enhancer.py
racoiaws: @rlangman please take a look at how tests are made, thanks

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
redoctopus: Sure, that's reasonable. Thanks!

tests/collections/tts/test_spectrogram_enhancer.py
redoctopus: Ah, I found this in the development docs:
> Unit tests must be fast: <1s ideally, <10s at worst, and >10s needs to have an exception from NeMo Core team.

So I guess this should be an integration test ("These tests are longer running and are used to confirm a behavior.").

nemo/collections/tts/models/spectrogram_enhancer.py
subhankar-ghosh: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/torch/data.py#L995
Take a look at this part, it can help with padding.

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
subhankar-ghosh: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/models/fastpitch.py#L290
for dim=1

nemo/collections/tts/models/spectrogram_enhancer.py
subhankar-ghosh: Well, essentially what you are doing in this model is mel to mel. The goal is to not have any dependency on the spectrogram generator. Here FastPitch is frozen and is not trained as part of the spectrogram enhancer.

In HiFiGAN training also the input is mel, which are precomputed, which can be done here as well. 
Before training, do inference on FastPitch or Any spectrogram generator model to precompute the mels. Now, these mels can be an input to the spectrogram enhancer. In your PR as well it is part of the `prepare_batch` method, which means it is not part of training but used in preparing data, which in turn means this can be precomputed. See the https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/torch/data.py#L995 part of the VocoderDataset class that loads precomputed spectrograms.

During training your batch can look like this: `(precomputed_spec, groundtruth audio, len, ...)`
You can use `LogMel` ttsdatatype to load your mels and incorporate the padding logic - https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/torch/data.py#L514

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
subhankar-ghosh: This means when you are importing `import torch` you should call all functionalities of torch with full name like use use `torch.Tensor` and `torch.nn` instead of `from torch import Tensor, nn`

nemo/collections/tts/models/spectrogram_enhancer.py
subhankar-ghosh: Not DictConfig object but providing the configs of your module in the yaml file like , I mean users should have the option to change G, D from the yaml file instead of touching the code. In the future, if someone wants to use a different Generator or Discriminator, it should be easier for them to just implement it and change in the yaml file instead of importing the modules in this file and changing the instantiation here. Doesn't really matter if they are inheriting ModelPT or not.

Example: duration predictor is instantiated here https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/models/fastpitch.py#L143
And the corresponding config is provided here https://github.com/NVIDIA/NeMo/blob/main/examples/tts/conf/fastpitch_align_44100.yaml#L198
Now, if I want to use a different module as duration predictor I do not need to change the code, all I will need to do is to change the yaml file. Makes it easy and clean.

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
subhankar-ghosh: Sure, that sounds good. Also in the docstring add the link to the paper.

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: > you should call all functionalities of torch with full name

Is is a NeMo code style requirement? I see nothing wrong with it, my preference would be is "ignore / won't fix".

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: No. Precomputing spectrograms is very inconvenient (and also requires lots of space). I speak from experience.

Current implementation is the most clean and convenient way of doing this from user point of view.

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Sorry, I don't understand your last message about padding, what is your suggestion or what is the issue.

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: No, I'm talking about a tensor of B x 1 x D x L - I did not find a way of expressing the 1 requirement in a NeuralType.

I guess result is good for the end user - no need to care about this extra dimension.

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: `"gradient_penalty_loss_every_n_steps"`

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: `"consistency_loss_weight"`

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: > I mean users should have the option to change G, D from the yaml file instead of touching the code.

> In the future, if someone wants to use a different Generator or Discriminator, it should be easier for them to just implement it and change in the yaml file instead of importing the modules in this file and changing the instantiation here.

We use a specific config of a very specific generative model (StyleGAN 2), modified for working with sequences. Using your own generator or discriminator is discouraged.

Personally to me, this YAML-oriented design caused a lot of frustration when I was starting with NeMo and FastPitch. Some issues off the top of my head:
1) Readability. There is no class name in code, just a bare `instantiate` and a config field. It instantiates _something_, and one has to jump between the config and the code. (Half a year passed, and I'm still doing this)
2) Explorabilily. Which modules exactly can I put there? The ones with the name "Predictor" in it (FP example)?
3) Extensibility is illusionary. Want your new replacement submodule to take some extra input? Modify Python code, YAML can't do this. Or write a Python monster to make this YAML-configurable.

Overall, the "just change it in the yaml file, easy and clean" feature actually comes at cost of spiraling cognitive load when trying to understand the code. It also blurs the border between YAML configuration and Python programming, making both harder to work with.

While there are strong benefits for models like FastPitch, I don't think it is warranted here at all.

racoiaws: Not exposing G and D in the YAML config was a conscious decision from my side. Unless it's an absolute requirement and blocks merging, I'd rather not do it.

rlangman: To not instantiate them with configurable parameters in the YAML would go against the NeMo design pattern. I do not know if it is the optimal approach, and its hard to say because in general we are clearly executing on it poorly (eg. 90% of the config files are copy-pasted making them way too large. NN layers which could be shared between models are instead copy-pasted between model definitions. The interfaces for the modules are not generic or documented).

But I think arguing that is out of the scope for a PR. Any individual PR should have to follow the coding styles and design patterns of the code base as decided by the organization.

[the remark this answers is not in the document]
racoiaws: It is ubiquitously G and D in the official NVIDIA implementations, e.g. https://github.com/NVlabs/stylegan3/blob/main/training/training_loop.py

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Thanks Ryan, both are good points 

I guess at the initial research stage I copied messy code from the precomputing spectrograms script and reassembled tensors just to be absolute sure about data format. These were B x 1 x D x L for 2d convs.

Will remove redundancies

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: > would go against the NeMo design pattern

> But I think arguing that is out of the scope for a PR. Any individual PR should have to follow the coding styles and design patterns of the code base as decided by the organization.

OK, good points.

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: The main goal was to completely repeat data setup from FastPitch training: to use the same texts, audio and ground-truth precomputed features like pitch (and I guess energy, with new revisions). This gets us would get natural-sythesize) spectrogram pairs and minimizes chances of the GAN learning to fix pitch or other features. 
I figured the most easy and natural way to implement this is just to run FastPitch on its underlying TTSDataset, which is the intended way to train this model.

The initial precomputing script I used just initialized the TTSDataset and iterated through it, saving results. Same thing happening here, but inside PTL.

I haven't measured overhead, but I also did not notice any significant slowdowns while doing experiments for our paper
compared to precomputed spectrograms. Slower training is solved by waiting a bit more, but lack of storage is solved only by buying more storage.
Same thing with GPU memory though... I'll look at memory usage and get back to you on this

Note that `generate_mels.py` mirrors some of TTSDataset internals but does not use it directly. Handling of sup data mangled names is a pain, so it would be a full rewrite or just a separate script.

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+from nemo.collections.tts.models import FastPitchModel
+from nemo.collections.tts.models.base import SpectrogramGenerator
+from nemo.collections.tts.modules.spectrogram_enhancer import Discriminator, Generator, mask
+from nemo.core import Exportable, ModelPT, typecheck
+from nemo.core.neural_types import LengthsType, MelSpectrogramType, NeuralType
+from nemo.core.neural_types.elements import BoolType
+from nemo.utils import logging
+
+
+def type_as_recursive(e, source: Tensor):
```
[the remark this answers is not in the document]
racoiaws: Will do

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+        batch = process_batch(batch, self._train_dl.dataset.sup_data_types)
+        mels, mel_lens = self.spectrogram_model.preprocessor(input_signal=batch["audio"], length=batch["audio_lens"])
+
+        mels_pred, *_ = self.spectrogram_model(
+            text=batch["text"],
+            durs=None,
+            pitch=batch["pitch"],
+            speaker=batch.get("speaker"),
+            pace=1.0,
+            spec=mels if self.spectrogram_model.learn_alignment else None,
```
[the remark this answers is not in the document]
racoiaws: We use the model to "resynthesize" input dataset, so the more ground-truth features the better.

`learn_alignment=True` means we FastPitch learns to compute alignment from input spectrogram and encoded text, which happens during training and exactly what we need here, because we need tts output spectrograms to have the same length as the input.

There is a bug here though: a model with `learn_alignment=False` is not handled correctly. `durs=None` always, so we most likely get exception somewhere down the line.

nemo/collections/tts/models/spectrogram_enhancer.py
subhankar-ghosh: Also, one very important aspect of spectrogram enhancer is that it should be independent of spectrogram generator. Why do we want to tie to any particular spectrogram generator (FastPitch)? If we want it to work with FastPitch only then we should make it a feature of FastPitch class. 

Having precomputed mels will enable us to use it for any spectrogram generator model. I understand that the experiments were done using FastPitch as spec generator, but when we release it on NeMo we should make it generic (so that it can be used with any spec gen model) as much as possible in my opinion.

subhankar-ghosh: > Note that generate_mels.py mirrors some of TTSDataset internals but does not use it directly. Handling of sup data mangled names is a pain, so it would be a full rewrite to TTSDataset (questionable) or just a separate script.

`generate_mels.py` is just doing inference to generate the mels. You don't really need any other supplementary data other than `attn_prior` and `speaker` both of which are supplied. Pitch, energy, duration etc are predicted by the model.

racoiaws: > Why do we want to tie to any particular spectrogram generator (FastPitch)? 

We really do not, it's just an example we're discussing. To use any other spectrogram generator just implement the boilerplate for converting a collated batch from TTSDataset to batch of paired real-fake spectrograms and it's good to go. Unfortunately they have different inputs (and interfaces) and thus need to be supported separately.

> generate_mels.py is just doing inference to generate the mels. You don't really need any other supplementary data other than attn_prior and speaker both of which are supplied. Pitch, energy, duration etc are predicted by the model.

Yes, and that's why it is not suitable for training this model. The more ground-truth features (pitch, energy), the better; durations also need to match. We need resynthesis using a TTS model.

rlangman: There should not be any functional difference from the approaches. I guess generate_mels.py is not using ground truth information, but it should have it as a configuration. When doing GAN based fine-tuning with the spectrogram (whether through StyleGan or indirectly HiFi-Gan fine-tune) it is not 100% clear to me which would be better. Using ground truth feature input is definitely safer but can limit the potential quality improvements.

So precomputing and making generate_mels.py more generic is one possible solution. It sounds like the other solution you are proposing is to turn "prepare_batch(self, batch)" into an abstract method or separate class that someone can implement with model-specific logic.

tests/collections/tts/test_spectrogram_enhancer.py
redoctopus: You should be able to mark them with `@pytest.mark.integration`, but I only see a few Hydra tests that use it in NeMo. @okuchaiev can you advise on best practice here? Should we be using this marker more, or is something else more appropriate?

nemo/collections/tts/models/spectrogram_enhancer.py
rlangman: Losses in GAN training are always difficult to interpret, but we should track the losses in wandb if only for debugging and making sure things do not become NaN and such.

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
rlangman: I am just thinking things up on the spot right now. I am not an expert with PyTorch conventions, but I have never seen such layout/structure/usage with any other library.

I would default to just importing those 2 and  providing full paths beyond that, for clarity. The usual alternative, as mentioned in the code style warning, is to import all methods `from torch import Tensor, nn, randn, ...` but that sounds like a bad idea because so many torch functions like randn are the exact same name as equivalent numpy methods.

I see no consistency on this in the rest of NeMo codebase. It probably does not matter.

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done but will be replaced with a preprocessing step

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Done

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: Done

examples/tts/conf/spectrogram-enhancer.yaml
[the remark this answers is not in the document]
racoiaws: Done

examples/tts/conf/spectrogram-enhancer.yaml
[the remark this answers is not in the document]
racoiaws: Comments done

nemo/collections/tts/modules/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: Done

Left only `import torch` and `import torch.nn.functional as F`

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: Done

Also added masking, better user interface

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,428 @@
...
+        }
+    )
+    def forward(
+        self, *, condition: Tensor, lengths: Tensor, mixing: bool = False, normalize: bool = True,
+    ):
+        """
+        Generator forward pass. Noise inputs will be generated.
+
+        condition: batch of blurred spectrograms
+        lenghts: length for every spectrogam in the batch
```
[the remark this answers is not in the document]
racoiaws: Fixed, thanks

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: Renaming done

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Will do after implementing preprocessing

nemo/collections/tts/models/spectrogram_enhancer.py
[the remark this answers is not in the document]
redoctopus: Seems fine to me :)

racoiaws:
Hi folks @rlangman @subhankar-ghosh
Please take a look at data preprocessing & loading update.

- added a script for resynthesis 
- implemented a Dataset for loading pairs of real-resynthesized spectrograms
- removed the spectrogram generator inside the model, switched to the new Dataset
- updated example config and removed tests

Couple of commits got squashed into the last one unfortunately

tests/collections/tts/test_spectrogram_enhancer.py
racoiaws: Removed the tests because the configuration they used was also removed

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -0,0 +1,244 @@
...
+
+import numpy as np
+import torch
+from omegaconf import DictConfig, OmegaConf
+from tqdm import tqdm
+
+from nemo.collections.tts.helpers.helpers import process_batch, to_device_recursive
+from nemo.collections.tts.models import FastPitchModel
+from nemo.collections.tts.models.base import SpectrogramGenerator
+from nemo.utils import logging
```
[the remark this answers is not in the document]
racoiaws: resolved

rlangman (commented):
Overall, looks good to me. Thanks for the effort refactoring the precompute and data loading. Just a few minor comments.

tests/collections/tts/test_spectrogram_enhancer.py
```diff
@@ -0,0 +1,159 @@
...
+
+    assert logits.shape == lengths.shape
+
+
+@pytest.mark.unit
+def test_nemo_save_load(enhancer: SpectrogramEnhancerModel, tmp_path):
+    path = tmp_path / "test-enhancer-save-load.nemo"
+
+    enhancer.save_to(path)
+    return SpectrogramEnhancerModel.restore_from(path)
```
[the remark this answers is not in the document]
racoiaws: I think this return value was used in another test that was deleted

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -0,0 +1,243 @@
...
+        Each returned batch is a dict containing main data, sup data, and model output
+        """
+        self.model.setup_training_data(self.model._cfg["train_ds"])
+
+        for batch_tuple in iter(self.model._train_dl):
+            batch = process_batch(batch_tuple, sup_data_types_set=self.model._train_dl.dataset.sup_data_types)
+            yield self.resynthesize_batch(batch)
+
+
+def read_manifest(path: Path) -> Dict[str, str]:
```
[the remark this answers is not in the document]
racoiaws: Good catch, was not aware of that, will do

nemo/collections/tts/losses/spectrogram_enhancer_losses.py
```diff
@@ -0,0 +1,102 @@
...
+    L1 distance on x0.25 Mel scale (20 bins for typical 80-bin scale)
+    """
+
+    def __init__(self, weight: float = 10):
+        super().__init__()
+        self.weight = weight
+
+    def __call__(self, condition, output, lengths):
+        *_, w, h = condition.shape
+        w, h = w // 4, h
```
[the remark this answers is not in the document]
racoiaws: Nothing, just visual coherence. Even more explicit statement we don't do anything with `h`, I guess.

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -0,0 +1,243 @@
...
+        for i, (manifest_entry, length) in enumerate(zip(batch_manifest, mel_lens.tolist())):
+            print(manifest_entry["audio_filepath"])
+            filename = Path(manifest_entry["audio_filepath"]).stem
+
+            # note that lengths match
+            pred_mel = pred_mels[i, :, :length].clone().numpy()
+            true_mel = true_mels[i, :, :length].clone().numpy()
+
+            pred_mel_path = output_folder / "mels" / f"{filename}.mel.npy"
+            true_mel_path = output_folder / "mels" / f"{filename}.mel.gt.npy"
```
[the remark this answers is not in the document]
racoiaws: Will do

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -0,0 +1,243 @@
...
+        "dataloader_params": {"batch_size": batch_size, "num_workers": num_workers, "shuffle": False},
+    }
+    model._cfg.train_ds = OmegaConf.merge(model._cfg.train_ds, DictConfig(dataset_config_overrides))
+    resynthesizer = TTSDatasetResynthesizer(model, device)
+
+    input_manifest = read_manifest(input_json_manifest)
+
+    output_manifest = []
+    output_json_manifest = output_folder / f"{input_json_manifest.stem}_mel{input_json_manifest.suffix}"
+    (output_folder / "mels").mkdir(exist_ok=True, parents=True)
```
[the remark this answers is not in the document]
racoiaws: Will do

tests/collections/tts/test_spectrogram_enhancer.py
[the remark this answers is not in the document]
racoiaws: Removed

scripts/dataset_processing/tts/resynthesize_dataset.py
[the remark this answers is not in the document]
racoiaws: It's already in a "mels" folder, so changed to just `.npy` and `_gt.npy`

nemo/collections/tts/modules/spectrogram_enhancer.py
```diff
@@ -0,0 +1,372 @@
...
+
+        style2 = self.to_style2(istyle)
+        x = self.conv2(x, style2)
+        x = self.activation(x + noise2)
+
+        rgb = self.to_rgb(x, prev_rgb, istyle)
+        return x, rgb
+
+
+def mask(tensor, lengths):
```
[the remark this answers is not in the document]
racoiaws: Done

rlangman (commented):
Looks good to me. Will let @subhankar-ghosh review again before approving.

examples/tts/conf/spectrogram-enhancer.yaml
```diff
@@ -0,0 +1,86 @@
...
+  # Spectrogram values range, calculated over your dataset with matching STFT parameters.
+  # Needed for treating spectrograms as images with pixel values around [0, 1].
+  # For LibriTTS, you can try [-13.18, 4.78]
+  spectrogram_min_value: ???
+  spectrogram_max_value: ???
+
+  train_ds:
+    dataset:
+      _target_: "nemo.collections.tts.torch.data.PairedRealFakeSpectrogramsDataset"
+      manifest_filepath: "/home/user/train-clean-100.manifest_mel.json"
```
subhankar-ghosh, line 38: Let's not put paths over here. Change this to a mandatory field i.e. ???

nemo/collections/tts/models/spectrogram_enhancer.py
```diff
@@ -0,0 +1,313 @@
...
+        discriminator_opt = instantiate(self._cfg.discriminator_opt, params=self.discriminator.parameters())
+        return [discriminator_opt, generator_opt], []
+
+    def setup_training_data(self, train_data_config):
+        dataset = instantiate(train_data_config.dataset)
+        self._train_dl = torch.utils.data.DataLoader(
+            dataset, collate_fn=dataset.collate_fn, **train_data_config.dataloader_params
+        )
+
+    def setup_validation_data(self, val_data_config):
```
subhankar-ghosh, line 270: We can go two ways here:
1) Remove `setup_validation_data` because the is no `val_ds` in the config, nor is there a validation step implementation.
OR
2) Add `val_ds` section to config as well as a validation_step to the model class.

nemo/collections/tts/modules/spectrogram_enhancer.py
```diff
@@ -0,0 +1,359 @@
...
+        result = (target + condition) / 2
+        result = mask_sequence_tensor(result, (condition_lengths / scale).ceil().long())
+
+        return result
+
+    def forward(self, condition: torch.Tensor, lengths: torch.Tensor, ws: List[torch.Tensor], noise: torch.Tensor):
+        batch_size, _, _, max_length = condition.shape
+
+        x = self.initial_block.expand(batch_size, -1, -1, -1)
+        x = x[:, :, :, : max_length // 16]
```
subhankar-ghosh, line 296: what is 16?

subhankar-ghosh:
Looks much better now. Left a few comments/Nit picks.

examples/tts/conf/spectrogram-enhancer.yaml
racoiaws: Good catch, thank you

This is a spillover from my test docker environment, definitely supposed to be `???`

nemo/collections/tts/modules/spectrogram_enhancer.py
racoiaws: This depends on the fact that we have four 2x upsample steps, so 2**4

Supposed to depend on number of blocks, also a good catch

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Let's say we compute discriminator loss on some held-out set. I am extremely skeptical whether it is a sensible way to choose between two models based on that loss. Until proven on a couple of downstream setups which use output of this GAN, of course.
I don't think validation steps are of value here, even though the generation is conditional. A validation step might be introduced in the future, if (when) we get some positive evidence, but for now let's keep it simple.

Replaced with a `pass` and a comment summing up above. This is an `@abstractmethod` in `ModelPT`, so there has to be _some_ implementation

racoiaws: Basically went with option 1.

nemo/collections/tts/modules/spectrogram_enhancer.py
racoiaws: Done, made a computable property for this

examples/tts/conf/spectrogram-enhancer.yaml
racoiaws: Fixed

nemo/collections/tts/models/spectrogram_enhancer.py
racoiaws: Current decision is not to

subhankar-ghosh:
LGTM.

subhankar-ghosh (approved):
LGTM

subhankar-ghosh:
Future Work @racoiaws :
1) Let's create a tutorial on how to use it OR you can add a section to FastPitch Finetuning tutorial about the spectrogram enhancer as an alternative to fine-tuning hifigan.
2) Let's also show users the comparison of audio produced by spec enhancer + frozen vocoder and fine-tuned vocoder.

racoiaws:
Thanks everyone for your review @redoctopus @subhankar-ghosh @rlangman

> Future Work @racoiaws :
> 
> 1. Let's create a tutorial on how to use it OR you can add a section to FastPitch Finetuning tutorial about the spectrogram enhancer as an alternative to fine-tuning hifigan.
> 2. Let's also show users the comparison of audio produced by spec enhancer + frozen vocoder and fine-tuned vocoder.

OK.

subhankar-ghosh (approved):

PR #5614, opened 2022-12-13
Update Dockerfile to use numba==0.53.1

stevehuang52:
Pinned to use `numba==0.53.1` to avoid crashing in training with `num_workers > 0`. This is just a temporary workaround, still need to fix it in the future.

Signed-off-by: He Huang (Steve) <105218074+stevehuang52@users.noreply.github.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,6 +85,10 @@ RUN python -c "import nemo.collections.nlp as nemo_nlp" && \
 # install pinned numba version
 # RUN conda install -c conda-forge numba==0.54.1
 
+# Pinned to numba==0.53.1 to avoid bug in training with num_workers > 0
+# The bug still exists with PTL 1.8.4, this is just a temporary workaround.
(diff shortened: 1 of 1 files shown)


titu1994 (approved):

titu1994:
@stevehuang52 can you raise an issue in PTL if you know what it causing this issue ? Is it still the same hacky workaround for detecting cuda in notebooks ?

stevehuang52:
> @stevehuang52 can you raise an issue in PTL if you know what it causing this issue ? Is it still the same hacky workaround for detecting cuda in notebooks ?

I'll try to figure out if it's still the same issue with detecting cuda devices, and will raise an issue to PTL if it is still the same problem

titu1994:
Thanks, I'll try debugging this week but it will probably next year when we have a solution. I don't want to in Numba, there's a lot of improvement's needed from newer version

PR #5671, opened 2022-12-19
Support nested NeMo models

artbataev:
# What does this PR do ?

Adds support for nested NeMo models (with resources).

It's possible to instantiate and save models with NeMo submodules. Artifacts are correctly saved and restored.

**Collection**: [core]

# Changelog 

# Usage

Example usage:
- 3 ways to instantiate child models: 
  - use subconfig directly
  - use `child_model_path` with `.nemo` checkpoint path to load the model
  - use pretrained model
- child model can contain artifacts
- child model config can be changed after parent model instantiation. Config will be saved when saving parent model

```python
from nemo.core.classes import ModelPT

class ChildModel(ModelPT):
    ...  # implement necessary methods

class ParentModel(ModelPT):
    def __init__(self, cfg, trainer=None):

        # optionally annotate type for IDE autocompletion and type checking
        self.child_model: Optional[ChildModel]
        if cfg.get("child_model") is not None:
            # load directly from config
            # either if config provided initially, or automatically
            # after model restoration
            self.register_nemo_submodule(
                "child_model",
                config_field="child_model",
                model=ChildModel(self.cfg.child_model),
            )
        elif cfg.get('child_model_path') is not None:
            # load from .nemo model checkpoint
            # while saving, config will be automatically assigned/updated
            # in cfg.child_model
            self.register_nemo_submodule(
                "child_model",
                config_field="child_model",
                model=ChildModel.restore_from(self.cfg.child_model_path),
            )
        elif cfg.get('child_model_name') is not None:
            # load from pretrained model
            # while saving, config will be automatically assigned/updated
            # in cfg.child_model
            self.register_nemo_submodule(
                "child_model",
                config_field="child_model",
                model=ChildModel.from_pretrained(self.cfg.child_model_name),
            )
        else:
            self.child_model = None
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


artbataev:
> This PR is stale because it has been open for 14 days with no activity. Remove stale label or comment or update or this will be closed in 7 days.

The PR is still actual. Need a review / discussion (maybe after 9 Jan)

artbataev:
@titu1994, I reworked the approach to handle nested submodules artifacts automatically, `register_submodule_artifacts` is not required now.
This is because nested models can change their config + artifacts, e.g., `change_vocabulary` for BPE models changes model resources. Since we can't track changes in submodules from the parent module, I check submodules for changes when saving the model.

I added tests for:
- serialization + deserialization
- multi-nested model
- multiple save-restore passes
- cases when the child model changes its config

I'm not sure about model-parallel serialization. Do we have any examples to start from?

Also, I'm not sure about the `is_model_being_restored` flag. 
I can add a simple context manager to preserve this flag when loading the child NeMo model. Is it a good solution? Do we have any tests for this flag?

nemo/core/classes/modelPT.py
```diff
@@ -240,6 +241,22 @@ def register_artifact(
 
         return self._save_restore_connector.register_artifact(self, config_path, src, verify_src_exists)
 
+    def has_artifacts(self) -> bool:
+        """Returns True if model has artifacts registered"""
+        return hasattr(self, 'artifacts') and self.artifacts is not None and len(self.artifacts) > 0
```
artbataev, line 246: What is the reason to check everywhere `hasattr` + `not None` for `artifacts`? Can we just assign `self.artifacts = {}` in `ModelPT` constructor?

nemo/core/connectors/save_restore_connector.py
```diff
@@ -37,7 +39,7 @@ def __init__(self) -> None:
         self._model_weights_ckpt = "model_weights.ckpt"
         self._model_extracted_dir = None
 
-    def save_to(self, model, save_path: str):
+    def save_to(self, model: nemo_classes.ModelPT, save_path: str):
```
titu1994, line 42: Just use string here, do not use the actual class just for typing

nemo/core/classes/modelPT.py
titu1994: Im ok with this, but instead of just setting it, inside modelpt do a hasattr check and only then set to None if it doesn't have the attr. Actually, it might cause issues but we can think about it (maybe another PR)

There is a case where we need to register before calling the constructor (look at ASRBPEMixin)

nemo/core/connectors/save_restore_connector.py
```diff
@@ -400,40 +403,71 @@ def register_artifact(self, model, config_path: str, src: str, verify_src_exists
...
+
+                    # Generate new uniq artifact name and copy it to nemo_file_folder
+                    # Note uuid.uuid4().hex is guaranteed to be 32 character long
+                    artifact_base_name = os.path.basename(artiitem.path)
+                    artifact_uniq_name = f"{uuid.uuid4().hex}_{artifact_base_name}"
+                    shutil.copy2(artiitem.path, os.path.join(nemo_file_folder, artifact_uniq_name))
+
+                    # Update artifacts registry
+                    artiitem.hashed_path = "nemo:" + artifact_uniq_name
+                    model.artifacts[conf_path] = artiitem
```
titu1994, line 433: Ah so we are gathering all the artifacts into the top level model artifact container. Very cool way to handle this.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -400,40 +403,71 @@ def register_artifact(self, model, config_path: str, src: str, verify_src_exists
...
+            restoration_paths.append(model_metadata.restoration_path)
+        # aggregate restoration paths for all submodules recursively
+        for module in model.modules():
+            if isinstance(module, nemo_classes.ModelPT):  # if NeMo model
+                submodule_restoration_path = app_state.get_model_metadata_from_guid(module.model_guid).restoration_path
+                if submodule_restoration_path is not None:
+                    restoration_paths.append(submodule_restoration_path)
+        if tarfile_artifacts and not restoration_paths:
+            logging.warning("Model contains registered artifacts, but no restoration paths found")
+        if tarfile_artifacts and restoration_paths:
```
titu1994, line 459: Can you be explicit here and do length > 0 checks

nemo/core/connectors/save_restore_connector.py
```diff
@@ -400,40 +403,71 @@ def register_artifact(self, model, config_path: str, src: str, verify_src_exists
...
             # Get path where the command is executed - the artifacts will be "retrieved" there
             # (original .nemo behavior)
             cwd = os.getcwd()
             try:
                 # Step into the nemo archive to try and find the file
                 with tempfile.TemporaryDirectory() as archive_dir:
-                    self._unpack_nemo_file(path2file=model_metadata.restoration_path, out_folder=archive_dir)
+                    # unpack all restorations paths (nemo checkpoints)
+                    # in nemo checkpoints all resources contain hash in name, so there should be no collisions
+                    for path in restoration_paths:
```
titu1994, line 469: Restoration paths were tempdirs, do we recreate those tempdirs?

nemo/core/connectors/save_restore_connector.py
```diff
@@ -454,8 +488,25 @@ def _handle_artifacts(self, model, nemo_file_folder):
                 # change back working directory
                 os.chdir(cwd)
 
+    def _update_subconfigs(self, model, path2yaml_file):
```
titu1994, line 491: Do we really need nested configs? Can you make the name closer to _update_subconfig_for_artifacts if you think this wont be used anywhere else? + move this function to SaveRestoreConnector maybe?

tests/core/test_save_restore.py
```diff
@@ -85,12 +101,66 @@ def setup_validation_data(self, val_data_config: Union[DictConfig, Dict]):
...
+            self.child1_model = None
+        else:
+            # if `child1_model_path` is set, model should be restored from nemo checkpoint
+            # otherwise model is constructed from config in `child1_model`
+            if cfg.get('child1_model_path') is not None:
+                self.child1_model = ModelPT.restore_from(cfg.child1_model_path)
+                # path in config should be set to `None` to restore model from config after saving
+                self.cfg.child1_model_path = None
+                # config for child model should be stored in base model config
+                self.cfg.child1_model = self.child1_model.cfg
```
titu1994, line 128: Is this for demonstration or do we need to ask model devs to do this? If so, can we automate it somehow.

tests/core/test_save_restore.py
```diff
@@ -85,12 +101,66 @@ def setup_validation_data(self, val_data_config: Union[DictConfig, Dict]):
...
 
+class MockModelWithChildren(MockModel):
+    """
+    Mock Model, can contain 2 children (other NeMo models)
+    """
+
+    def __init__(self, cfg, trainer=None):
+        super().__init__(cfg=cfg, trainer=trainer)
+
+        # child 1
```
titu1994, line 117: Where do you register the model file? Or is that no longer necessary cause internally it will handle the module traversal ?

tests/core/test_save_restore.py
```diff
@@ -549,6 +619,290 @@ def test_mock_model_model_collision(self):
...
+
+                parent = MockModelWithChildren(cfg_parent.model)
+                if change_child_number:
+                    parent.child2_model.change_stub_number(10)
+                parent.save_to(parent_path)
+            # restore, separate children checkpoints are not available here (tmpdir_child destroyed)
+            parent = ModelPT.restore_from(parent_path)
+            # check model is transparent, child models can be accessed and can be saved/restored separately
+            _ = self.__test_restore_elsewhere(parent.child1_model, map_location='cpu')
+            child2 = self.__test_restore_elsewhere(parent.child2_model, map_location='cpu')
```
titu1994, line 660: Amazing test, incredible work here !

titu1994 (commented):
Minor comments here and there, but overall incredible work !
Would not have thought of this solution, but it works very very well. 
@ericharper after your review, we can merge this for Non-MP models. I think the refactor for MP models would be only minor changes which can be done in a subsequent PR (this PR is a blocker for two others)

tests/core/test_save_restore.py
```diff
@@ -85,12 +101,66 @@ def setup_validation_data(self, val_data_config: Union[DictConfig, Dict]):
...
+    """
+
+    def __init__(self, cfg, trainer=None):
+        super().__init__(cfg=cfg, trainer=trainer)
+
+        # child 1
+        if cfg.get('child1_model_path') is None and cfg.get('child1_model') is None:
+            self.child1_model = None
+        else:
+            # if `child1_model_path` is set, model should be restored from nemo checkpoint
```
ericharper, line 121: here "nemo checkpoint" -> "nemo model"?

tests/core/test_save_restore.py
```diff
@@ -85,12 +101,66 @@ def setup_validation_data(self, val_data_config: Union[DictConfig, Dict]):
...
+                # config for child model should be stored in base model config
+                self.cfg.child1_model = self.child1_model.cfg
+            else:
+                self.child1_model = ModelPT.from_config_dict(self.cfg.child1_model)
+
+        # child 2
+        if cfg.get('child2_model_path') is None and cfg.get('child2_model') is None:
+            self.child2_model = None
+        else:
+            # if `child2_model_path` is set, model should be restored from nemo checkpoint
```
ericharper, line 136: "nemo checkpoint" -> "nemo model" ?

tests/core/test_save_restore.py
```diff
@@ -549,6 +619,290 @@ def test_mock_model_model_collision(self):
...
+        with tempfile.TemporaryDirectory() as tmpdir_parent:
+            parent_path = os.path.join(tmpdir_parent, "parent.nemo")
+            with tempfile.TemporaryDirectory() as tmpdir_child:
+                # save children
+                child1_path = os.path.join(tmpdir_child, 'child1.nemo')
+                child1.save_to(child1_path)
+                child2_path = os.path.join(tmpdir_child, 'child2.nemo')
+                child2.save_to(child2_path)
+
+                # create model with children using saved "nemo" checkpoints
```
ericharper, line 647: "checkpoints" -> "models"

ericharper (commented):
LGTM! Really nice PR and excellent unit tests. I'm good with this once all of @titu1994 's comments are resolved.

nemo/core/connectors/save_restore_connector.py
artbataev: Fixed

nemo/core/connectors/save_restore_connector.py
artbataev: `restoration_paths` are only paths to `.nemo` checkpoints, since we don't use `.nemo` files inside parent `.nemo` file. Using `.nemo` in `.nemo` will still break the code and should be avoided.

I changed it to a `set` to unpack each checkpoint only once

nemo/core/connectors/save_restore_connector.py
artbataev: > Do we really need nested configs?
Yes. The problem is that the child model can change its config (not only artifacts! e.g. you can change number of heads, momentum parameter in BatchNorm and so on). There is no easy way to notify parent about these config changes (some hacks exist, but it will be too complicated and fragile). So when saving we have to update subconfigs.
There is a test for this: see `change_child_number` parametrization (child model changes config, this should be handled correctly).

>  move this function to SaveRestoreConnector maybe?
Didn't understand, it is in `SaveRestoreConnector`. 

I marked it with `@staticmethod`, since it is really static.

tests/core/test_save_restore.py
artbataev: There are 2 ways to instantiate child model. 

I added comments to `MockModelWithChildren`.

**Variant 1** for creating nested NeMo model:
- create a child model from .nemo checkpoint (from `*_model_path`)
- assign config to indicate that this model should be saved/restored
- when model is restored, parent and children will be restored from a signle parent `.nemo` checkpoint

**Variant 2** for creating nested NeMo model: load directly from config (can be instantiated via `MyModel(child_cfg)` or universally `ModelPT.from_config_dict(child_cfg)` (if `_target_` is provided).

In both cases no `.nemo` file inside parent `.nemo` model will be saved (this is still incompatible with NeMo and global `AppState`!). So, after loading submodels from `.nemo` checkpoints, a solid model is constructed.

Keep in mind, that synchronization is only done in `save_to`:
- aggregate and update subconfigs
- aggregate and save artifacts

tests/core/test_save_restore.py
artbataev: This is a checkpoint, so I changed it to ".nemo model checkpoints"

tests/core/test_save_restore.py
artbataev: This is a checkpoint, so I changed it to ".nemo model checkpoints"

tests/core/test_save_restore.py
artbataev: ".nemo model checkpoints"

tests/core/test_save_restore.py
artbataev: If the model is loaded from checkpoint (see a comment about 2 ways to construct child models), the developer should assign a config.

I also thought about doing this automatically (since we can use `__getattr__` on `ModelPT` to do this), but:
- there are cases when the model will be directly constructed from config, so we don't need to assign it once more
- this will break existing code: there is existing NLP code, that uses additional NeMo models as attributes, but doesn't save them (actually I catched errors when tried to do everything recursively)
- if the model is not needed to be saved (assigned as attribute, but then deleted), the developer will have to delete `cfg.child1_model` directly after 'magic' construction of this attribute.

So, I think it's better to ask developers to do it manually if constructing child models from `.nemo` checkpoints.

nemo/core/connectors/save_restore_connector.py
artbataev: I think we should avoid quotes, and use [PEP563](https://peps.python.org/pep-0563/) as a more stable approach, since quotes are intended to use with forward declarations by original [PEP484](https://peps.python.org/pep-0484/#forward-references), and there is no forward declaration here. 

But there is still a problem with partial initialization, so I used `from __future__ import annotations`. This is fully compatible with Python3.7+, widely recommended and adopted.

E.g. see in [mypy](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563), [pydantic](https://docs.pydantic.dev/usage/postponed_annotations/) and so on.

You can't see it in old code (due to Python3.7+ limitations), but in new code it is also widely use it, e.g.:
- [Lightning (app)](https://github.com/Lightning-AI/lightning/blob/10a4b245291c986984885ed50aba830318305674/src/lightning_app/frontend/panel/app_state_watcher.py)
- [TorchVision](https://github.com/pytorch/vision/blob/1353df9e45aae6739c23db4a053c3d9a9fd17b3d/torchvision/prototype/datapoints/_datapoint.py)

To avoid long discussions here, **I added quotes**. 

But I will appreciate if you look at alternative approach and discuss it (maybe separately).

nemo/core/connectors/save_restore_connector.py
titu1994: We need to add a test where we attempt to register a Nemo file. Maybe check and raise an error ?

tests/core/test_save_restore.py
titu1994: Let's not modify getattr. Very annoying to do it correctly and brittle for future. 

However it is also very easy to forget this step and completely break your implementation. 

Add a test that checks this - a model which is nested but "forgets" to set the config. Raise an error. 

What about doing this - inside of save_to, check if any module is ModelPT and if so, check the config is available other wise raise error

titu1994: Or do this - inside save to check if the config is set, if yes don't do anything else otherwise use self.nested_module_i.cfg and set the config automatically inside of self.cfg

tests/core/test_save_restore.py
titu1994: Then add a check inside of register_artifact that checks if it's a .Nemo file and raise an error saying that Nemo file itsel cannot be registered. Add a test case for this

titu1994:
It seems there are more edge cases introduced by the PR. Let's capture those case's in tests, and raise error for users 

1) requirement to manually inject config. This is painful and should be done somehow automatically with checks in save to

2) registering artifact of .Nemo file should be disallowed. 

3) nested model config right now is saved as just the module name - this can conflict with predefined config. 

Use "_" pre-and-appended to the nested modules name in the config to prevent override of config. Or use somethinf like prepend "inner" or some other tag to denote config of inner module

nemo/core/connectors/save_restore_connector.py
titu1994: Please use string name here. Pep is not an enforcement it is a suggestion. It does not make sense to mess around internals of Nemo just for an annotation.

titu1994:
+ can you add documentation in Nemo docs about nested modules supportb- how to use it and what to be careful of

artbataev:
@titu1994, let's agree on edge cases and design

>1. requirement to manually inject config

There are 2 cases, how nested models can be constructed

a) Main option: use (sub)config, which is a natural way in NeMo to instantiate different submodules. There is no need to register this config.

```python
class ChildModel(ModelPT):
    ...  # add necessary implementations

class NestedModel1(ModelPT):
    """Case 1: Construct model from subconfig"""
    def __init__(self, cfg, trainer=None):
        self.child_model = ChildModel(self.cfg.child_model)
        # also possible, if _target_ is in child model config
        # self.child_model = ModelPT.from_config_dict(self.cfg.child_model)
```
Let's focus on this implementation.

b) **Additional** option to load child from `.nemo` checkpoint. Yes, you have manually inject config + set path to null + in `else` instantiate the model from config. This will always work if the option **a** works.

```python
class NestedModel2(ModelPT):
    """Case 2: Construct model from .nemo model checkpoint"""
    def __init__(self, cfg, trainer=None):
        if self.cfg.child_model_path is not None:
            self.child_model = ModelPT.restore_from(self.cfg.child_model_path)
            self.cfg.child_model = self.child_model.cfg  # register config
            self.cfg.child_model_path = None
        else:
            # manually restore the model
            self.child_model = ChildModel(self.cfg.child_model)
            # _target_ will be in child model config, so we can use
            # self.child_model = ModelPT.from_config_dict(self.cfg.child_model)
```

> 2. registering artifact of .Nemo file should be disallowed.

Nothing changed. You can't load ".nemo" model inside `__init__`: it's broken by current serialization design, since serialization uses global AppState. 

But you still can use it as an artifact and load separately **not in `init`**. Since nothing is changed, I don't see reasons to raise an error if registering `.nemo` artifact (**this can break someone's code**).

Generally assigning a model loaded from `.nemo` + registering it as an artifact is a bad idea, since you end up with duplicated weights (PyTorch saves them recursively). But if you don't do both things in the same code - it can be ok.

It is better to redesign serialization to avoid using global AppState properties (e.g. use model-specific properties), but let's postpone it, since it is tricky and can be significantly more controversial than the current simple solution of aggregating subconfigs and artifacts).

> 3. nested model config right now is saved as just the module name - this can conflict with predefined config.

Yes, in current implementation there is a limitation: name in config should be equal to the property name `self.child_model = ChildModel(self.cfg.child_model)`

If this is not a good idea (by design in NeMo config field and object attribute share the same name, so I considered it to be ok), we can **instead** of automatic registration add **manual registration** for submodule (explicit is better then implicit!), and use the following design:

```python
class NestedModel1(ModelPT):
    """Case 1: Construct model from subconfig"""
    def __init__(self, cfg, trainer=None):
        self.child_model_custom = ChildModel(self.cfg.child_model)
        self.register_nemo_submodule(config_field="child_model", property_name="child_model_custom")
```
I't easy to implement, it's explicit, and doesn't rely on same config/model property names.

Registering submodules **automatically** will end up with a lot of more potential edge cases and errors. Also, there is a case, when it **will break existing code**:  in `nemo/collections/nlp/models/nlp_model.py` `self.bert_model` is a NeMo model, but there is a workaround to handle this case, and `bert_model` shouldn't be saved as a child model. 

If you are ok with `self.register_nemo_submodule`, I will implement it (it's easy), and add documentation.
If not, let's discuss a better approach that will not break existing code.

requirements/requirements_lightning.txt
```diff
@@ -1,6 +1,6 @@
 hydra-core>=1.2.0,<1.3
 omegaconf>=2.2,<2.3
-pytorch-lightning>=1.8.6
+pytorch-lightning>=1.8.6,<1.9.0
```
artbataev, line 3: Temporary fix CI :)

nemo/core/classes/modelPT.py
```diff
@@ -240,6 +245,75 @@ def register_artifact(
...
+
+    def has_native_or_submodules_artifacts(self) -> bool:
```
titu1994, line 252: Im wondering if these functions should be inside of SaveRestoreConnector and just forward the calls here. Im ok with this also - @ericharper

nemo/core/classes/modelPT.py
```diff
@@ -240,6 +245,75 @@ def register_artifact(
...
+            raise NeMoBaseException(
+                "You are trying to register a submodule before the model is initialized. This is not allowed. "
+                "Did you forget to call `super().__init__`?"
+            )
+        # assign attribute to self
+        setattr(self, name, model)
+        # add to the submodules mapping
+        self.nemo_submodule_name_to_config_field[name] = config_field
+
+    def named_cfg_nemo_modules(self, prefix: str = "") -> Iterator[Tuple[str, "ModelPT"]]:
```
titu1994, line 291: Rename to `named_nemo_modules`

nemo/core/connectors/save_restore_connector.py
```diff
@@ -400,40 +404,69 @@ def register_artifact(self, model, config_path: str, src: str, verify_src_exists
...
+        if model_metadata.restoration_path is not None:
+            restoration_paths.add(model_metadata.restoration_path)
+        # aggregate restoration paths for all submodules recursively
+        for module in model.modules():
+            if isinstance(module, nemo_classes.ModelPT):  # if NeMo model
+                submodule_restoration_path = app_state.get_model_metadata_from_guid(module.model_guid).restoration_path
+                if submodule_restoration_path is not None:
+                    restoration_paths.add(submodule_restoration_path)
+        if len(tarfile_artifacts) > 0 and len(restoration_paths) == 0:
+            logging.warning("Model contains registered artifacts, but no restoration paths found")
```
titu1994, line 457: This case can occur if you load (from constructor) inner model. It is fine to have it logged, but also note a comment with a TODO to see if we can fix it in the future

requirements/requirements_lightning.txt
titu1994: I think fix got merged, rebase and revert this

tests/core/test_save_restore.py
```diff
@@ -549,6 +670,339 @@ def test_mock_model_model_collision(self):
...
+                assert parent.child2_model.temp_data == child_with_child_data
+                assert parent.child2_model.child2_model.temp_data == child_data
+
+    @pytest.mark.unit
+    @pytest.mark.with_downloads
+    def test_mock_model_nested_child_from_pretrained(self):
+        """
+        Test nested model with child initialized from pretrained model
+        """
+        cfg = _mock_model_with_child_encdecctc_config("QuartzNet15x5Base-En")
```
titu1994, line 991: Quartznet doesnt have any artifact, switch to `stt_en_conformer_ctc_small`

titu1994 (commented):
This looks much better, minor change of name, and maybe move the ModelPT code to SaveRestoreConnector and just forward the calls in ModelPT (if it makes sense to you).

If not, this PR is ready to merge. @ericharper please look at it one last time before merge

nemo/core/classes/modelPT.py
artbataev: I firstly used `named_nemo_modules` name, but I think it's confusing. This function returns **not names** (paths) like `named_modules` in PyTorcgh, but **paths in configs** that can be different. So, I added `cfg` to the name

titu1994: Pytorch has named_modules that return name, module. named_nemo_modules can return name, Nemo model

tests/core/test_save_restore.py
artbataev: Done

requirements/requirements_lightning.txt
artbataev: Fixed

nemo/core/connectors/save_restore_connector.py
artbataev: Added todo

tests/core/test_save_restore.py
artbataev: > Then add a check inside of register_artifact that checks if it's a .Nemo file and raise an error saying that Nemo file itsel cannot be registered. Add a test case for this

Firstly added, but then reverted, since such a check breaks existing code. See
`tests/collections/asr/test_asr_rnnt_encoder_model_bpe.py::TestEncDecRNNTBPEModel::test_save_restore_nested_model`

tests/core/test_save_restore.py
```diff
@@ -19,14 +19,15 @@
...
+from huggingface_hub.hf_api import ModelFilter
 from omegaconf import DictConfig, OmegaConf, open_dict
 
 from nemo.collections.asr.models import EncDecCTCModel, EncDecCTCModelBPE
-from nemo.collections.nlp.models import PunctuationCapitalizationModel, TransformerLMModel
+from nemo.collections.nlp.models import PunctuationCapitalizationModel
 from nemo.core.classes import ModelPT
 from nemo.core.connectors import save_restore_connector
 from nemo.utils.app_state import AppState
+from nemo.utils.exceptions import NeMoBaseException
```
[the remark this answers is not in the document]
artbataev: Fixed

titu1994:
> Disallowing registering .nemo file as an artifact breaks existing code, so I reverted changes related to this check. See tests/collections/asr/test_asr_rnnt_encoder_model_bpe.py::TestEncDecRNNTBPEModel::test_save_restore_nested_model

Enforce this check, remove/update that test. That test is a proxy of the old way I was trying to hack together multi model support.

tests/core/test_save_restore.py
titu1994: Lets add this check, and either remove or update the faulty RNNT test. That one was my initial attempt to hack together nested model support and is no longer in the correct format of this PR

titu1994 (commented):
This is ready to merge (after adding back the nemo file checks).
@ericharper for final review

Amazing work !

docs/source/core/core.rst
```diff
@@ -635,7 +635,64 @@ The resulting .nemo file will then have the following file:
...
+if the ``src`` cannot be found.
+
+Nested NeMo Models
+------------------
+
+In some cases, it may be helpful to use NeMo models inside other NeMo models. For example, we can incorporate language models into ASR models to use in a decoding process to improve accuracy or use hybrid ASR-TTS models to generate audio from the text on the fly to train or finetune the ASR model.
+
+There are 3 ways to instantiate child models inside parent models:
+
+- use subconfig directly
```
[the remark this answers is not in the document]
artbataev: Yes, this will work. There is a parameter in tests `child2_model_from_path` to test this case: if `child2_model_from_path` is False, model is constructed directly from subconfig.

docs/source/core/core.rst
```diff
@@ -635,7 +635,64 @@ The resulting .nemo file will then have the following file:
...
+            super().__init__(cfg=cfg, trainer=trainer)
+
+            # optionally annotate type for IDE autocompletion and type checking
+            self.child_model: Optional[ChildModel]
+            if cfg.get("child_model") is not None:
+                # load directly from config
+                # either if config provided initially, or automatically
+                # after model restoration
+                self.register_nemo_submodule(
+                    "child_model",
```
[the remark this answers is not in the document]
artbataev: done

docs/source/core/core.rst
```diff
@@ -635,7 +635,64 @@ The resulting .nemo file will then have the following file:
...
+                    config_field="child_model",
+                    model=ChildModel(self.cfg.child_model),
```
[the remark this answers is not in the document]
artbataev: done

artbataev:
> > Disallowing registering .nemo file as an artifact breaks existing code, so I reverted changes related to this check. See tests/collections/asr/test_asr_rnnt_encoder_model_bpe.py::TestEncDecRNNTBPEModel::test_save_restore_nested_model
> 
> Enforce this check, remove/update that test. That test is a proxy of the old way I was trying to hack together multi model support.

I added the check and removed the test

ericharper (approved):
LGTM. Thanks for many changes, iterations, and documentation!

nemo/core/classes/modelPT.py
```diff
@@ -240,6 +243,94 @@ def register_artifact(
...
+            if (
+                isinstance(module, ModelPT)
+                and hasattr(module, 'artifacts')
+                and module.artifacts is not None
+                and len(module.artifacts) > 0
+            ):
+                return True
+        return False
+
+    def register_nemo_submodule(self, name: str, config_field: str, model: "ModelPT") -> None:
```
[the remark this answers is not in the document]
artbataev: I checked documentation build, and it's already in API docs:

Screenshot 2023-01-23 at 12 57 43

artbataev:
@titu1994, @ericharper 
Thank you so much for your comments and discussion!

I finally updated the documentation + fixed the faulty test for NestedRNNTModel in `tests/collections/asr/test_asr_rnnt_encoder_model_bpe.py` instead of removing it.

I think now it's ready to be merged)

nemo/core/classes/modelPT.py
```diff
@@ -110,6 +112,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
 
         self._cfg = cfg
 
+        # init mapping submodule attribute -> config_field for nested NeMo models
+        self.nemo_submodule_name_to_config_field = dict()
```
SeanNaren, line 116:
```suggestion
        self._nemo_submodule_name_to_config_field = dict()
```

Can we protect this?

SeanNaren:
LGTM as well, absolutely bonkers you made this work! I hope this promotes new awesome clean multi-model pipelines in NeMo 🚀

nemo/core/classes/modelPT.py
artbataev: Yep, I agree. Fixed 👍

ericharper (approved):
LGTM. Thanks!

PR #5672, opened 2022-12-19
[NLP] Supervised Fine-tuning for GPT

soares-f:
# What does this PR do ?

The goal of this PR is to add a supervised fine-tuning functionality for GPT.

**Collection**:  NLP - language modeling 

# Changelog 
- Created a new class similar to GPT pretraining: MegatronGPTSFTModel
- Create a new dataset class for SFT, allowing answer-only loss and the inclusion of EOS and SEP tokens
- Create a new usage example

**Pre checks**:
- [X] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
+++ b/examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
@@ -1,0 +1,188 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import tempfile
+
+from omegaconf.omegaconf import OmegaConf, open_dict
+from pytorch_lightning import Trainer
+from pytorch_lightning.callbacks.timer import Timer
+from pytorch_lightning.plugins.environments.torchelastic_environment import TorchElasticEnvironment
+from pytorch_lightning.trainer.connectors.checkpoint_connector import CheckpointConnector
+
+from nemo.collections.nlp.models.language_modeling.megatron_gpt_sft_model import MegatronGPTSFTModel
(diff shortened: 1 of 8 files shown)


soares-f:
@MaximumEntropy

examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
```diff
@@ -0,0 +1,186 @@
...
+    if cfg.model.tensor_model_parallel_size > 1 or cfg.model.pipeline_model_parallel_size > 1:
+        app_state.model_parallel_size = cfg.model.tensor_model_parallel_size * cfg.model.pipeline_model_parallel_size
+        app_state.tensor_model_parallel_size = cfg.model.tensor_model_parallel_size
+        app_state.pipeline_model_parallel_size = cfg.model.pipeline_model_parallel_size
+        (
+            app_state.tensor_model_parallel_rank,
+            app_state.pipeline_model_parallel_rank,
+            app_state.model_parallel_size,
+            app_state.data_parallel_size,
+            app_state.pipeline_model_parallel_split_rank,
```
markelsanz14, line 95: This is not set in GPT configs, I think it's only for encoder/decoder architectures. So this will crash if not set. Please use .get() so that it returns None if not found.

examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
```diff
@@ -0,0 +1,186 @@
...
+            app_state.virtual_pipeline_model_parallel_rank,
+        ) = fake_initialize_model_parallel(
+            world_size=app_state.model_parallel_size,
```
markelsanz14, line 98: This is not correct. You are doing world_size=TP*PP, but it should be world_size=TP*PP*DP.

examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
MaximumEntropy: The fake initialization call will just return the split rank as 0 for GPT models so this is harmless for GPT.

examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
MaximumEntropy: @soares-f this bit of code is probably something that is used in GPT inference code because it assumes DP=1, like @markelsanz14 said. Also I don't think you need these bits since calling `trainer.fit()` should set up most things you need. This script should look very similar to the GPT pre-training script.

XuesongYang (changes requested):
please rebase to the latest main branch, otherwise, the unnecessary changes for TTS related codes would crash. Thanks.

PR #5692, opened 2022-12-21
Fix for incorrect computation of batched alignment in transducers

Kipok:
# What does this PR do ?

The original code had a bug because of which whenever at least one batch element predicted non-blank symbol, all blank elements would duplicate the last prediction in their `.alignments` variable.

Additionally, I've unified `_greedy_decode_blank_as_pad` and `_greedy_decode_masked` into a single function with an if check. The two methods had 99% of the same code except 5 lines or so, so imo it does not make sense to have them as fully separate code paths. Please let me know if for some reason we don't want to have this change.

**Collection**: ASR

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
@@ -632,13 +632,13 @@ class GreedyBatchedRNNTInfer(_GreedyRNNTInfer):
             # Initialize Hidden state matrix (shared by entire batch)
             hidden = None
 
-            # If alignments need to be preserved, register a danling list to hold the values
+            # If alignments need to be preserved, register a dangling list to hold the values
             if self.preserve_alignments:
                 # alignments is a 3-dimensional dangling list representing B x T x U
                 for hyp in hypotheses:
                     hyp.alignments = [[]]
 
-            # If confidence scores need to be preserved, register a danling list to hold the values
+            # If confidence scores need to be preserved, register a dangling list to hold the values
             if self.preserve_frame_confidence:
                 # frame_confidence is a 3-dimensional dangling list representing B x T x U
                 for hyp in hypotheses:
@@ -787,6 +787,8 @@ class GreedyBatchedRNNTInfer(_GreedyRNNTInfer):
                                 hypotheses[kidx].y_sequence.append(ki)
                                 hypotheses[kidx].timestep.append(time_idx)
                                 hypotheses[kidx].score += float(v[kidx])
+                            elif self.preserve_alignments:  # creating new alignment list for samples that had blank
+                                hypotheses[kidx].alignments.append([])
 
                         symbols_added += 1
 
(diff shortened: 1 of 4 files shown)


nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -811,211 +819,6 @@ def _greedy_decode_blank_as_pad(
 
         return hypotheses
 
-    def _greedy_decode_masked(
```
titu1994, line 814: Why is this being removed?

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -679,8 +671,21 @@ def _greedy_decode_blank_as_pad(
                     if time_idx == 0 and symbols_added == 0 and hidden is None:
                         g, hidden_prime = self._pred_step(self._SOS, hidden, batch_size=batchsize)
                     else:
-                        # Perform batch step prediction of decoder, getting new states and scores ("g")
-                        g, hidden_prime = self._pred_step(last_label, hidden, batch_size=batchsize)
+                        if self.decoder.blank_as_pad:
```
titu1994, line 674: Please revert, we dont want to merge these two functions.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -650,7 +642,8 @@ def _greedy_decode_blank_as_pad(
             # Last Label buffer + Last Label without blank buffer
             # batch level equivalent of the last_label
             last_label = torch.full([batchsize, 1], fill_value=self._blank_index, dtype=torch.long, device=device)
-
+            if not self.decoder.blank_as_pad:
+                last_label_without_blank = last_label.clone()
```
titu1994, line 646: Add else case setting var to None

titu1994 (changes requested):
Revert merge of the two functions please.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
Kipok: Can you please explain why we don't want to merge them? They are different in just a few lines, which can be handled with an if-else block (or a new function if you'd prefer that way). And I had to apply the same fix to both functions, so I thought it makes sense to unify them. If we unify them and in the future we need to make some changes, it's going to be in a single place and not in two places, which is easy to forget.

titu1994: They share similar logic but not concept. Normally the prednet is not supposed to have the blank token in it's vocab. We abuse the notation by using the pytorch embeddings mechanism for masking one token in NeMo for efficiency sake but the academically correct code is the second one. 

They should not be merged.

Kipok: Rolled-back

Kipok:
I think the same changes would need to go for multi-blank RNNT as well, but it's better if @hainan-xv can confirm. It should be easy to test using the same test code I added, assuming there is some pretrained model we can use

titu1994 (approved):
Looks great, thanks ! @hainan-xv please look into the multi blank rnnt version and send a PR

PR #5724, opened 2023-01-03
Ensure EMA checkpoints are also deleted when normal checkpoints are

### LINKED ISSUE #5631: EMA Doesn't delete previous EMA ckpts when k > 0 for checkpointing
**Describe the bug**

EMA saves a separate EMA based ckpt when normally saving. The problem is when saving a set amount of checkpoints (i.e k=5) the older EMA ckpts are not being deleted. This means there is an excess of EMA based ckpts.

What is expected is that older EMA ckpts are deleted.

**Proposed Solution**

Overriding the code is tricky, as the call to remove ckpts is found within functions, like such:

https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/callbacks/model_checkpoint.py#L652
https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/callbacks/model_checkpoint.py#L671
https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/callbacks/model_checkpoint.py#L713

My suggestion would be to move this code into a separate function that can be overridden in the [NeMoModelCheckpoint](https://github.com/NVIDIA/NeMo/blob/4d3e99f986fd9354b5528a3e9c82fe01ec2df916/nemo/utils/exp_manager.py#L745) class by introducing a `delete_checkpoint` function. In this function we will also be able to delete the EMA weights.

SeanNaren:
# What does this PR do ?

Closes #5631

**Collection**: Common

# Changelog 
- Delete EMA checkpoints when normal checkpoints are deleted.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/requirements/requirements_lightning.txt
+++ b/requirements/requirements_lightning.txt
@@ -1,6 +1,6 @@
 hydra-core>=1.2.0,<1.3
 omegaconf>=2.2,<2.3
-pytorch-lightning>=1.8.3
+pytorch-lightning>=1.8.6
 pyyaml<6  # Pinned until omegaconf works with pyyaml>=6
 torchmetrics
 transformers>=4.0.1,<=4.21.2
--- a/tests/collections/common/test_ema.py
+++ b/tests/collections/common/test_ema.py
@@ -248,6 +248,29 @@ class TestEMAConfig:
         for saved_weight, ema_weight in zip(duplicate_model.state_dict().values(), ema_weights):
             assert torch.allclose(saved_weight.cpu(), ema_weight.cpu())
 
+    @pytest.mark.unit
+    @pytest.mark.run_only_on('GPU')
+    def test_exp_manager_ema_weights_topk(self, tmpdir):
+        """Test to ensure that EMA correctly ensures we only keep topk checkpoints."""
+        tmp_path = tmpdir / "exp_manager_test"
+        model = ExampleModel()
+        save_top_k = 3
+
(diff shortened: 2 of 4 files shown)


requirements/requirements_lightning.txt
```diff
@@ -1,6 +1,6 @@
 hydra-core>=1.2.0,<1.3
 omegaconf>=2.2,<2.3
-pytorch-lightning>=1.8.3
+pytorch-lightning>=1.8.6
```
SeanNaren, line 3: Hope its ok to upgrade the version here @titu1994

tests/collections/common/test_ema.py
```diff
@@ -248,6 +248,27 @@ def test_exp_manager_ema_weights(self, tmpdir):
         for saved_weight, ema_weight in zip(duplicate_model.state_dict().values(), ema_weights):
             assert torch.allclose(saved_weight.cpu(), ema_weight.cpu())
 
+    @pytest.mark.unit
+    def test_exp_manager_ema_weights_topk(self, tmpdir):
```
SeanNaren, line 252: Confirmed this test fails when using pytorch-lightning==1.8.3

requirements/requirements_lightning.txt
titu1994: Should be fine, r1.15 has been cut already so it affects only main.

titu1994 (approved):
Looks great ! Thanks !

PR #5727, opened 2023-01-03
[ASR] Use a subset of manifest when using `scatter` shard strategy

anteju:
# What does this PR do ?

When using `shard_strategy='scatter'` this PR loads only a subset of lines from the manifest file.
This may reduce the time to process a manifest file by an order of magnitude.

Opening as a draft to get feedback if there are any underlying assumptions which may be broken by this change.

**Collection**: ASR

# Changelog 
| File                      | Change |
| ---------------- | -------- |
| `manifest.py::item_iter` | Load only a subset of lines if `shard_strategy == 'scatter'` |
| `collections.py:: ASRAudioText ` | forward `shard_strategy`, `global_rank` and `world_size` to `manifest.item_iter`  |
| `collections.py:: AudioText ` | Use rank and world size to restore the original data list length for `shard_strategy='scatter'` |
|  `audio_to_text.py` | forward `shard_strategy`, `global_rank` and `world_size` to `collections.ASRAudioText` |
| `utils,py` | Added a function to get number of lines from a text file |
| `test_utils.py` | Added unit test for the utility function added above |

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/data/audio_to_text.py
+++ b/nemo/collections/asr/data/audio_to_text.py
@@ -133,6 +133,9 @@ class ASRManifestProcessor:
         eos_id: Optional[int] = None,
         pad_id: int = 0,
         index_by_file_id: bool = False,
+        shard_strategy: Optional[str] = None,
+        global_rank: Optional[int] = None,
+        world_size: Optional[int] = None,
     ):
         self.parser = parser
 
@@ -143,6 +146,9 @@ class ASRManifestProcessor:
             max_duration=max_duration,
             max_number=max_utts,
             index_by_file_id=index_by_file_id,
+            shard_strategy=shard_strategy,
+            global_rank=global_rank,
+            world_size=world_size,
         )
 
         self.eos_id = eos_id
@@ -786,6 +792,9 @@ class _TarredAudioToTextDataset(IterableDataset):
             eos_id=eos_id,
             pad_id=pad_id,
             index_by_file_id=True,  # Must set this so the manifest lines can be indexed by file ID
+            shard_strategy=shard_strategy,
+            global_rank=global_rank,
+            world_size=world_size,
         )
 
         self.featurizer = WaveformFeaturizer(sample_rate=sample_rate, int_values=int_values, augmentor=augmentor)
--- a/nemo/collections/common/parts/preprocessing/collections.py
+++ b/nemo/collections/common/parts/preprocessing/collections.py
@@ -113,6 +113,8 @@ class AudioText(_Collection):
         max_number: Optional[int] = None,
         do_sort_by_duration: bool = False,
         index_by_file_id: bool = False,
+        shard_strategy: Optional[str] = None,
+        world_size: Optional[int] = None,
     ):
         """Instantiates audio-text manifest with filters and preprocessing.
 
@@ -131,6 +133,8 @@ class AudioText(_Collection):
             max_number: Maximum number of samples to collect.
             do_sort_by_duration: True if sort samples list by duration. Not compatible with index_by_file_id.
             index_by_file_id: If True, saves a mapping from filename base (ID) to index in data.
+            shard_strategy: Used to recover the full length of data list if shard_strategy == 'scatter'
+            world_size: Total number of processes, used for partitioning the manifest.
         """
 
         output_type = self.OUTPUT_TYPE
@@ -190,6 +194,12 @@ class AudioText(_Collection):
             else:
                 data.sort(key=lambda entity: entity.duration)
 
+        if shard_strategy == 'scatter' and world_size > 1:
+            # WORKAROUND:
+            # Originally, data list had full length even when the data is scattered
+            # Here, we restore the length by repeating the same sequence, to make sure epoch length is unchanched
+            data = world_size * data
+
         logging.info("Dataset loaded with %d files totalling %.2f hours", len(data), total_duration / 3600)
         logging.info("%d files were filtered totalling %.2f hours", num_filtered, duration_filtered / 3600)
 
@@ -199,7 +209,15 @@ class AudioText(_Collection):
 class ASRAudioText(AudioText):
     """`AudioText` collector from asr structured json files."""
 
-    def __init__(self, manifests_files: Union[str, List[str]], *args, **kwargs):
+    def __init__(
+        self,
+        manifests_files: Union[str, List[str]],
+        shard_strategy: Optional[str] = None,
+        global_rank: Optional[int] = None,
+        world_size: Optional[int] = None,
+        *args,
+        **kwargs,
+    ):
         """Parse lists of audio files, durations and transcripts texts.
 
         Args:
@@ -217,7 +235,9 @@ class ASRAudioText(AudioText):
             [],
         )
         speakers, orig_srs, token_labels, langs = [], [], [], []
-        for item in manifest.item_iter(manifests_files):
+        for item in manifest.item_iter(
+            manifests_files, shard_strategy=shard_strategy, global_rank=global_rank, world_size=world_size
+        ):
             ids.append(item['id'])
             audio_files.append(item['audio_file'])
             durations.append(item['duration'])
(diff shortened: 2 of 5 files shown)


nemo/collections/common/parts/preprocessing/collections.py
```diff
@@ -190,6 +194,12 @@ def __init__(
             else:
                 data.sort(key=lambda entity: entity.duration)
 
+        if shard_strategy == 'scatter' and world_size > 1:
```
titu1994, line 197: Why duplicate data ? It will take multiple times the memory.

nemo/collections/common/parts/preprocessing/collections.py
```diff
@@ -228,7 +248,19 @@ def __init__(self, manifests_files: Union[str, List[str]], *args, **kwargs):
...
+            audio_files,
+            durations,
+            texts,
+            offsets,
+            speakers,
+            orig_srs,
+            token_labels,
+            langs,
+            shard_strategy=shard_strategy,
+            world_size=world_size,
```
titu1994, line 261: Missing global rank

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -75,9 +84,44 @@ def item_iter(
...
         cached_manifest_file = DataStoreObject(manifest_file).get()
         logging.debug('Cached at: %s', str(cached_manifest_file))
+
+        if shard_strategy is not None and world_size > 1:
+            if shard_strategy == 'scatter':
+                # Read only necessary manifest lines
+                logging.debug('Partitioning manifest: global rank = %d, world size = %d', global_rank, world_size)
+
+                # Get number of lines in the current manifest
+                num_lines = get_num_lines(cached_manifest_file)
```
titu1994, line 94: Rather import data_util and then explicitly call data_util.get_num_lines for clarity

nemo/collections/common/parts/utils.py
```diff
@@ -96,3 +96,28 @@ def flatten(list_in: List) -> List:
         A flat list of values.
     """
     return list(flatten_iterable(list_in))
+
+
+def get_num_lines(filepath: str, buffer_size: int = 2 ** 16) -> int:
```
titu1994, line 101: Name it get num lines from file

titu1994 (changes requested):
Since many classes depend on ASR manifest processor please go through where it is used and update them all. 

Also rework logic to avoid duplicating data

nemo/collections/common/parts/preprocessing/collections.py
titu1994: Change the default behavior is better. Such duplication of data just for length calculation is too wasteful

tests/collections/common/test_utils.py
```diff
@@ -31,3 +34,22 @@ def test_flatten(self):
...
+    def test_get_num_lines(self):
+        """Test getting the number of lines from a text file.
+        """
+        test_cases = [0, 1, 10, 42, 97]
+
+        for num_lines in test_cases:
+
+            lines = [f'line {n}\n' for n in range(num_lines)]
+
+            with tempfile.TemporaryDirectory() as tmp_dir:
```
Kipok, line 50: Pytest can handle this natively: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html (so just add tmp_path fixture and use it insteam of tmp_dir). It's a bit more user-friendly, because if test fails, it's not deleted right away, and you can inspect generated files.

tests/collections/common/test_utils.py
```diff
@@ -31,3 +34,22 @@ def test_flatten(self):
...
         for n, test_case in enumerate(test_cases):
             assert flatten(test_case['input']) == test_case['golden'], f'Test case {n} failed!'
+
+
+class TestTextUtils:
+    @pytest.mark.unit
+    def test_get_num_lines(self):
+        """Test getting the number of lines from a text file.
+        """
+        test_cases = [0, 1, 10, 42, 97]
```
Kipok, line 44: Not very important, but why not put as `@pytest.mark.parametrize(num_lines)`?

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -75,9 +84,44 @@ def item_iter(
...
+            for n, line in enumerate(f):
                 k += 1
+                if begin_idx is not None and n < begin_idx:
+                    # Skip lines before begin_idx
+                    continue
+                if end_idx is not None and n >= end_idx:
+                    # Stop parsing lines from end_idx
+                    break
                 item = parse_func(line, manifest_file)
                 item['id'] = k
```
Kipok, line 126: I'm not really sure what this "id" is used for, but if it's important to preserve the original mapping, then you'd need to continue the loop until the end to make sure it's properly counted, I guess? And if it's not important to have it the same, then you can move the `k += 1` to after the ifs, so that it's going from 0 to N on all processes without skips.

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -75,9 +84,44 @@ def item_iter(
...
+                # Read only necessary manifest lines
+                logging.debug('Partitioning manifest: global rank = %d, world size = %d', global_rank, world_size)
+
+                # Get number of lines in the current manifest
+                num_lines = get_num_lines(cached_manifest_file)
+                logging.debug('Found %d lines in manifest %s', num_lines, cached_manifest_file)
+
+                # Determine start and end lines in the manifest.
+                # This corresponds to splitting the manifest in audio_to_text.py::expand_audio_filepaths
+                begin_idx = (num_lines // world_size) * global_rank
```
Kipok, line 99: I don't fully understand how this works. Are we trying to select all the lines that have shards selected for the current rank? If so, is the assumption here that they always go in sequence and each shard has the same number of lines? I'm just not sure if this is always the case (or if we should assume that it's going to always be the case in the future).

Kipok: And in any case, I think it's very important to have some tests for this functionality, as it might be very easy to add hard-to-notice bugs here by ignoring part of the data

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -133,6 +133,9 @@ def __init__(
         eos_id: Optional[int] = None,
         pad_id: int = 0,
         index_by_file_id: bool = False,
+        shard_strategy: Optional[str] = None,
+        global_rank: Optional[int] = None,
```
Kipok, line 137: I guess that's more of a question to @titu1994, but is it possible to use pytorch's `dist.get_rank()` and same for world size? Or a light-weight wrapper function if we need to support some edge cases (and in fact I think we already have such function `utils.get_rank`)? Just feels that this is a global property that never changes and would be much easier to get it from a function whenever it's needed, than pass around through each model and dataloader.

tests/collections/common/test_utils.py
titu1994: Did not know this, very nice given how many places we use tempfile in tests

nemo/collections/asr/data/audio_to_text.py
titu1994: I think AppState is a more valid source of "world size" cause it takes a few more cases into account for model parallel. Ignoring that, you first have to check whether distributed is initialized or not, then you can use a method to get world size and local rank ID. It is definitely possible. 

The problem with the global property is the fact that it's global. Originally tarred datasets used such global variables but some domains had case where training is done in multi gpu, you destroy the distributed context and clean up, then eval on single GPU, all in the same run. This caused clash with such cases cause global setting was still set and not cleaned up automatically.

The other way is not via property but global check via testing if pt is under distributed enabled, whether global rank is correct, etc and then return the rank other wise -1. It's higher cost per call though, but it's fine for a init level thing

titu1994:
If it can work for speech ASR dataset (tarred, bucketed) I think @AlexGrinch would like this. Right now we're loss hacking to skip samples

PR #5729, opened 2023-01-03
Update torch.stft() calls due to deprecation of return_complex=False

redoctopus:

# What does this PR do ?

Fixes errors due to deprecation of `return_complex=False` in `torch.stft()` in version 1.14.

Most of our calls implicitly use `return_complex=False` (was the default), which returns a tensor of shape [B, N, T, 2] where the last dimension captures the real and imaginary parts of the tensor. However, moving forward, pytorch is deprecating this in favor of the behavior of `return_complex=True`, which returns a complex tensor of shape [B, N, T].

This bugfix explicitly sets `return_complex=True` as required by v1.14, and then converts the output of `torch.stft()` back to [B, N, T, 2] using `torch.view_as_real()`.

Similarly, `torch.istft()` now wants a complex tensor input; this is fixed by calling `torch.view_as_complex()` on our original input.

**Collection**: TTS

# Changelog 
- `torch.stft()` calls explicitly set `return_complex=True` and then convert output where necessary
- Added `torch.view_as_complex()` to input of `torch.istft()`

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/losses/stftlosses.py
+++ b/nemo/collections/tts/losses/stftlosses.py
@@ -61,7 +61,7 @@ def stft(x, fft_size, hop_size, win_length, window):
     Returns:
         Tensor: Magnitude spectrogram (B, #frames, fft_size // 2 + 1).
     """
-    x_stft = torch.stft(x, fft_size, hop_size, win_length, window, return_complex=False)
+    x_stft = torch.view_as_real(torch.stft(x, fft_size, hop_size, win_length, window, return_complex=True))
     real = x_stft[..., 0]
     imag = x_stft[..., 1]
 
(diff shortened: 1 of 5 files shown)


ericharper (approved):
LGTM! Thanks!

redoctopus:
> I think there are asr preprocessor so also using this, could you see if they need any similar update? I can send those patches after I return

Just checked, there's just the one in: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/preprocessing/features.py#L286

I can make a quick fix for that, it should be exactly the same!

titu1994:
Thanks, saw the PR, looks good !

PR #5742, opened 2023-01-06
[TTS] Update Spanish TTS model to 1.15

rlangman:

# What does this PR do ?

Update FastPitch Spanish config from 1.14.0 to 1.15.0, which contains IPA support and improvements from speaker level pitch normalization.

https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan/version

**Collection**: [TTS]

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


docs/source/tts/data/ngc_models_am.csv
```diff
@@ -8,5 +8,5 @@ en-US,RAD-TTS,TBD,TBD,TBD,ARPABET,nemo.collections.tts.models.radtts.RadTTSModel
...
-es,tts_es_fastpitch_multispeaker,OpenSLR crowdsourced Latin American Spanish,44100Hz,174,grapheme,nemo.collections.tts.models.fastpitch.FastPitchModel,`tts_es_multispeaker_fastpitchhifigan <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.14.0/files/tts_es_fastpitch_multispeaker.nemo``
+es,tts_es_fastpitch_multispeaker,OpenSLR crowdsourced Latin American Spanish,44100Hz,174,IPA,nemo.collections.tts.models.fastpitch.FastPitchModel,`tts_es_multispeaker_fastpitchhifigan <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.15.0/files/tts_es_fastpitch_multispeaker.nemo``
```
XuesongYang, line 11: Why override this line? We could provide two models, one with grapheme-only and the other wit IPA.

docs/source/tts/data/ngc_models_vocoder.csv
```diff
@@ -9,5 +9,5 @@ en-US,tts_waveglow_88m,librosa.filters.mel,LJSpeech,22050Hz,1,nemo.collections.t
...
-es,tts_es_hifigan_ft_fastpitch_multispeaker,FastPitch,OpenSLR crowdsourced Latin American Spanish,44100Hz,174,nemo.collections.tts.models.hifigan.HifiGanModel,`tts_es_multispeaker_fastpitchhifigan <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.14.0/files/tts_es_hifigan_ft_fastpitch_multispeaker.nemo``
+es,tts_es_hifigan_ft_fastpitch_multispeaker,FastPitch,OpenSLR crowdsourced Latin American Spanish,44100Hz,174,nemo.collections.tts.models.hifigan.HifiGanModel,`tts_es_multispeaker_fastpitchhifigan <https://catalog.ngc.nvidia.com/orgs/nvidia/teams/nemo/models/tts_es_multispeaker_fastpitchhifigan>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.15.0/files/tts_es_hifigan_ft_fastpitch_multispeaker.nemo``
```
XuesongYang, line 12: Same comment as above. It would be great to provide the two models.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -601,10 +601,10 @@ def list_available_models(cls) -> 'List[PretrainedModelInfo]':
         )
         list_of_models.append(model)
 
-        # es, 174 speakers, 44100Hz, OpenSLR
+        # es, 174 speakers, 44100Hz, OpenSLR (IPA)
```
XuesongYang, line 604: append IPA model instead of override.

nemo/collections/tts/models/hifigan.py
```diff
@@ -406,7 +406,7 @@ def list_available_models(cls) -> 'Optional[Dict[str, str]]':
         # Spanish, multi-speaker, 44100 Hz, Latin American Spanish OpenSLR
         model = PretrainedModelInfo(
             pretrained_model_name="tts_es_hifigan_ft_fastpitch_multispeaker",
-            location="https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.14.0/files/tts_es_hifigan_ft_fastpitch_multispeaker.nemo",
+            location="https://api.ngc.nvidia.com/v2/models/nvidia/nemo/tts_es_multispeaker_fastpitchhifigan/versions/1.15.0/files/tts_es_hifigan_ft_fastpitch_multispeaker.nemo",
```
XuesongYang, line 409: Shall we append your IPA models here instead of

XuesongYang (changes requested):

docs/source/tts/data/ngc_models_am.csv
rlangman: We could, but is there is value in maintaining 2 models? The IPA model also fully supports grapheme-only inference.

rlangman: Should we make it clearer that "IPA" means it supports mixed inputs?

redoctopus: Is the quality comparable between the IPA model using graphemes only & the previous model?

If so, then no objections to replacement.

redoctopus: Replacing seems fine to me, then.

> Should we make it clearer that "IPA" means it supports mixed inputs?

In this case I think that's worthwhile--we can tell users they can use any of [all graphemes/all IPA/mix].

XuesongYang: yeah, good question! But I vote for keeping both char and IPA ones. Char-only input is still useful for languages that lack IPA dicts, such as low-resourced African languages. 

we have char/phoneme(to be deprecated and adjusted to ipa)/ipa tokenizers for each language now, and yes we can hack IPA tokenizer to support char-only inputs, but we have to wait until detailed instructions are provided how to do char-only using ipa tokenizer.

@redoctopus what do you think?

redoctopus: Honestly, I'm fine with choosing whether to replace/keep both on a case-by-case basis; if we've already done the work to determine that the new model trained on both IPA/grapheme is still comparable to the grapheme-only model for this model, then I'm not sure there's a great reason to keep the old one.

> we have char/phoneme(to be deprecated and adjusted to ipa)/ipa tokenizers for each language now, and yes we can hack IPA tokenizer to support char-only inputs, but we have to wait until detailed instructions are provided how to do char-only using ipa tokenizer.

As it stands our IPA tokenizer can still deal with grapheme input if user sets `use_chars=True`, and we currently always do this since we don't have any other way to deal with OOV. I would like to keep this functionality even in the future when we have compatibility with neural G2P that _can_ deal with OOV, so that users who still want to use a dict lookup instead can still have an option for training on samples with OOV words.

So assuming we do keep that around, we'll always technically be able to train IPA-based models that _can_ take all grapheme inputs (though without any guarantees of quality--but this one is evaluated so I think it's OK to use as a replacement).

XuesongYang: great discussion. approved. please help to add a sentence for this `In this case I think that's worthwhile--we can tell users they can use any of [all graphemes/all IPA/mix].`

XuesongYang (approved):

docs/source/tts/data/ngc_models_am.csv
rlangman: I do think that grapheme-only models are useful. But similar to @redoctopus I think we should handle it on a per-language or per-model basis.

Even in this case based on the data, the performance might be slightly better if I did grapheme-only + speaker pitch norm. But I would rather invest the time it takes to train and manage that model into further improving the IPA one.

The way to do the grapheme-only inference could use some improvements and documentation (eg. escaping the entire input or manually setting model.tokenizer.phoneme_probability to 0%).

PR #5752, opened 2023-01-08
Set the stream position to 0 for pydub

jonghwanhyeon:
# What does this PR do ?

This PR sets the stream position to 0 to use pydub as fallback if SoundFile fails to read an audio file. 

When using tarred datasets, `audio_file` is a file-like object, not a path-like object. In this situation, to use pydub as fallback, we need to set the stream position of `audio_file` to 0 since SoundFile have advanced the position while reading it.

**Collection**: ASR

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/parts/preprocessing/segment.py
+++ b/nemo/collections/asr/parts/preprocessing/segment.py
@@ -219,6 +219,9 @@ class AudioSegment(object):
                     f"NeMo will fallback to loading via pydub."
                 )
 
+                if hasattr(audio_file, "seek"):
+                    audio_file.seek(0)
+
         if HAVE_PYDUB and samples is None:
             try:
                 samples = Audio.from_file(audio_file)


nemo/collections/asr/parts/preprocessing/segment.py
```diff
@@ -219,6 +219,9 @@ def from_file(
                     f"NeMo will fallback to loading via pydub."
                 )
 
+                if hasattr(audio_file, "seek"):
```
titu1994, line 222: Doesn't seem to be correct logic cause both offset and duration are ignored

jonghwanhyeon: They are not ignored. They are resolved after the `audio_file.seek(0)`.
```python
        if HAVE_PYDUB and samples is None:
            try:
                samples = Audio.from_file(audio_file)
                sample_rate = samples.frame_rate
                num_channels = samples.channels
                if offset > 0:
                    # pydub does things in milliseconds
                    seconds = offset * 1000
                    samples = samples[int(seconds) :]
                if duration > 0:
                    seconds = duration * 1000
                    samples = samples[: int(seconds)]
```

titu1994: Oh woops. Missed that. Looks good.

titu1994 (approved):

PR #5757, opened 2023-01-09
Add EMA Docs, fix common collection documentation

SeanNaren:
# What does this PR do ?

Adds simple documentation for EMA. Also fixes the docs on main due to an incorrect import in the EMA class.

**Collection**: Common

# Changelog 
- Add EMA documentation

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [x] Documentation


requirements/requirements_docs.txt
```diff
@@ -1,3 +1,4 @@
+Jinja2<3.1
```
SeanNaren, line 1: When running `bash update_docs_docker.sh` in the docs folder, I was getting this error: https://github.com/readthedocs/readthedocs.org/issues/9038

To fix this, I set the requirement in the docs build.

docs/source/common/callbacks.rst
```diff
@@ -0,0 +1,56 @@
...
+
+.. code-block:: bash
+
+    python examples/asr/asr_ctc/speech_to_text_ctc.py \
+        model.train_ds.manifest_filepath=/path/to/my/train/manifest.json \
+        model.validation_ds.manifest_filepath=/path/to/my/validation/manifest.json \
+        trainer.devices=2 \
+        trainer.accelerator='gpu' \
+        trainer.max_epochs=50 \
+        exp_manager.ema.enable=True \
```
titu1994, line 40: Let's use ... and add just the difference Rather than copy full command again

docs/source/common/callbacks.rst
```diff
@@ -0,0 +1,56 @@
...
+        exp_manager.ema.decay=0.999
+
+We also offer other helpful arguments.
+
+.. code-block:: bash
+
+    python examples/asr/asr_ctc/speech_to_text_ctc.py \
```
titu1994, line 47: List the arguments as a table maybe without the script part

requirements/requirements_docs.txt
titu1994: Wow it's still an issue after all this time ?? I could run it last year without this ..

titu1994 (commented):
Minor changes

requirements/requirements_docs.txt
SeanNaren: Tried again but still required this requirement, as it's building a docker image I'll assume that it would be the same for everyone else

titu1994 (approved):

PR #5758, opened 2023-01-09
Fix EMA topk checkpoint deletion

SeanNaren:
# What does this PR do ?

Fixes case where we need to also delete the EMA checkpoints from previous runs.

**Collection**: Common

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


nemo/utils/exp_manager.py
```diff
@@ -1001,6 +1007,12 @@ def _remove_checkpoint(self, trainer: "pytorch_lightning.Trainer", filepath: str
     def _ema_format_filepath(self, filepath: str) -> str:
         return filepath.replace(self.FILE_EXTENSION, f'-EMA{self.FILE_EXTENSION}')
 
+    def _has_ema_ckpts(self, best_k_models: List[str]) -> bool:
+        return any(self._is_ema_filepath(model_name) for model_name in best_k_models)
+
+    def _is_ema_filepath(self, filepath: Union[Path, str]) -> bool:
+        return filepath.endswith(f'-EMA{self.FILE_EXTENSION}')
```
suiyoubi, line 1014: Cannot directly call `.endswith` on `PosixPath` Object.
`return str(filepath).endswith(f'-EMA{self.FILE_EXTENSION}')`

nemo/utils/exp_manager.py
```diff
@@ -877,10 +877,16 @@ def nemo_topk_check_previous_run(self):
         else:
             models_to_delete = len(best_k_models) - self.save_top_k
         logging.debug(f'Number of models to delete: {models_to_delete}')
+
+        # If EMA enabled, delete the additional EMA weights
+        ema_enabled = self._has_ema_ckpts(best_k_models)
```
suiyoubi, line 882: `best_k_models` only has Non-EMA checkpoints, therefore, this would always be **FALSE.**
We need to provide the list of all the checkpoints in the folder here.
```
        all_checkpoints = list(path for path in Path(self.dirpath).rglob("*.ckpt"))
        ema_enabled = self._has_ema_ckpts(all_checkpoints)
```

nemo/utils/exp_manager.py
```diff
@@ -877,10 +877,16 @@ def nemo_topk_check_previous_run(self):
...
+
         for _ in range(models_to_delete):
```
suiyoubi, line 884: One note here: since all `models_to_delete` only have reg checkpoint path.
If in the folder, there are some EMA-checkpoints that do not have accompanied reg checkpoints (and they are supposed to be remove), these EMA-checkpoint will not be removed.

I think this edge case is fine to leave it as it is because it should never happen in an actual training setup: When we have EMA ckpts saved, the reg ckpts should have always have been created already before.

SeanNaren: agreed!

PR #5777, opened 2023-01-11
fix token id bug

yidong72:
fixed the tokenizer bug, add `bos_id` to the tabular tokenizer


MaximumEntropy (changes requested):
I don't think it should silently add an eos if bos is not available. It should fail and throw an error message saying bos is not part of the tokenizer.

yidong72:
> I don't think it should silently add an eos if bos is not available. It should fail and throw an error message saying bos is not part of the tokenizer.

It is for legacy support. the old tabular tokenizer doesn't have 'bos' token. but text generation only has "add bos" argument which under the hood uses 'eos'. If we throw an error the old trained model won't work anymore. But I can add a warning message so it is not silent.

MaximumEntropy (approved):

PR #5788, opened 2023-01-12
Update radtts' infer path

blisc:
# What does this PR do ?

Update radtts' infer path.

**Collection**: TTS

# Changelog 
- Update radtts' infer path

### DIFF
--- a/nemo/collections/tts/modules/radtts.py
+++ b/nemo/collections/tts/modules/radtts.py
@@ -18,6 +18,8 @@ import torch
 import torch.nn.functional as F
 from torch import nn
 
+import random
+
 from nemo.collections.tts.helpers.helpers import get_mask_from_lengths
 from nemo.collections.tts.helpers.helpers import mas_width1 as mas
 from nemo.collections.tts.helpers.helpers import regulate_len
--- a/nemo/collections/tts/modules/radtts.py
+++ b/nemo/collections/tts/modules/radtts.py
@@ -627,7 +627,14 @@ class RadTTSModule(NeuralModule, Exportable):
         pace = pace[:, :txt_len_pad_removed]
         pitch_shift = pitch_shift[:, :txt_len_pad_removed].unsqueeze(-1)
 
-        txt_enc_time_expanded, out_lens = regulate_len(dur, txt_enc.transpose(1, 2), pace)
+        txt_enc_time_expanded, out_lens = regulate_len(
+            dur,
+            txt_enc.transpose(1, 2),
+            pace,
+            replicate_to_nearest_multiple=True,
+            group_size=self.n_group_size,
+            in_lens=in_lens,
+        )
         n_groups = torch.div(out_lens, self.n_group_size, rounding_mode='floor')
         max_out_len = torch.max(out_lens)
 
@@ -638,6 +645,8 @@ class RadTTSModule(NeuralModule, Exportable):
                 voiced_mask = self.v_pred_module.infer(txt_enc_time_expanded, spk_vec_attributes, lens=out_lens)
                 voiced_mask_bool = torch.sigmoid(voiced_mask[:, 0]) > 0.5
                 voiced_mask = voiced_mask_bool.to(dur.dtype)
+            else:
+                voiced_mask_bool = None
         else:
             voiced_mask_bool = voiced_mask.bool()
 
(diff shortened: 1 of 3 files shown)


nemo/collections/tts/modules/radtts.py
```diff
@@ -777,33 +805,71 @@ def _prepare_for_export(self, **kwargs):
...
+        # index = 1
+        # while index < len(batch_lengths):
+        #     sum += batch_lengths[index] - batch_lengths[index - 1]
+        #     index += 1
+        # assert sum == sz[0], f"sum: {sum}, sz: {sz[0]}, lengths:{batch_lengths}"
+
+        # TODO: Shouldn't hardcode but self.tokenizer isn't initlized yet so unsure how
+        # to get the pad_id
+        pad_id = 94
+        inp[inp == pad_id] = pad_id - 1 if pad_id > 0 else pad_id + 1
```
blisc, line 835: @XuesongYang @redoctopus Any suggestions on this?

redoctopus: When is `input_example()` called? Is it possible to create the tokenizer first?

One way around it might be to have the `pad_id` be manually calculated beforehand using number of total valid tokens (with a note that explains that it's how the tokenizer assigns the `pad_id`). I'd be somewhat worried about it causing a subtle bug in the future, though, if something changes and they no longer match.

Another possibility--maybe the user can specify it in the config? They'll have to figure out their vocab length first though (annoying user experience).

The most ideal thing is just to have the tokenizer in existence first, if possible...

XuesongYang: I would suggest preserving `<pad>` ID (along with `<unk>` and `<blank>`) strictly in either the beginning of the file `https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py`, or `https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py`, as a shared resource to all tokenizers. 

for general tokenization purposes, token IDs for `<pad>`, `<unk>`, and `<blank>` should be preserved ahead to avoid any confusion and potential bugs. For example, predefine some mapping `{"<pad>": 0, `<unk>`: 1, `<blank>`: 2}`. Dynamically assigning token IDs for those may bring maintenance overheads...

From my current experience inside NeMo, any mel-generator model (not module) should call `self._setup_tokenizer()` when instantiating itself, i.e. https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/models/fastpitch.py#L93-L98. Instantiating g2p happens inside `self._setup_tokenizer()` as shown in https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/models/fastpitch.py#L213.

borisfom: I agree with @XuesongYang  - tokenizer has to be initialized at the time of input_example call. And we should use ```<pad>``` token in config.

nemo/collections/tts/modules/radtts.py
```diff
@@ -812,8 +878,18 @@ def forward_for_export(self, text, lens, speaker_id, speaker_id_text, speaker_id
             sigma_txt=0.0,
             sigma_f0=1.0,
             sigma_energy=1.0,
-            f0_mean=145.0,
-            f0_std=30.0,
+            f0_mean=0.0,
+            f0_std=0.0,
```
borisfom, line 882: Those mean values should be read from the checkpoint.

nemo/collections/tts/modules/radtts.py
```diff
@@ -812,8 +878,18 @@
...
             in_lens=lens,
+            pitch_shift=pitch,
+            pace=pace,
+        ).values()
+        volume_extended = volume
```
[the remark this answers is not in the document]
borisfom: Looks like volume_extended is being overwritten below with regulate_len(..., volume ..)

nemo/collections/tts/modules/radtts.py
```diff
@@ -691,7 +719,7 @@ def infer(
         if self.n_group_size > 1:
             mel = self.fold(mel)
 
-        return mel, out_lens, dur, f0, energy_avg
+        return {'mel': mel, 'out_lens': out_lens, 'dur': dur, 'f0': f0, 'energy_avg': energy_avg}
```
borisfom, line 723: This reverts my previous change - which is a good thing.

nemo/collections/tts/modules/radtts.py
blisc: A value of 0 means no shift with is what we want.

nemo/collections/tts/modules/radtts.py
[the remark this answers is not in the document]
blisc: Fixed

nemo/collections/tts/modules/radtts.py
blisc: You're right, it is instantiated in the model. Not sure why this functionality was defined in the module, but moving to from module to model solves this issue

borisfom (approved):

PR #5772, opened 2023-01-11
Add Esperanto ASR example

andrusenkoau:
Signed-off-by: andrusenkoau <andrusenkoau@gmail.com>

# What does this PR do ?

Adds ASR example for training Esperanto Conformer-CTC-large model.

**Collection**: ASR

# Changelog 
- Adds Esperanto example to docs/source/asr/examples/

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [x] Documentation

PR #5638, opened 2022-12-15
Add AWS SageMaker ASR Examples

SeanNaren:
# What does this PR do ?

Adds AWS SageMaker example notebooks for ASR!

**Collection**: ASR

# Changelog 
- Add AWS SageMaker examples to NeMo repo.

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

PR #5797, opened 2023-01-13
CommonVoice support for `convert_hf_dataset_to_nemo` script

SeanNaren:
# What does this PR do ?

Related #5638 #5772  

Renames `sentence` column to `text` before creating manifest. Exposes `ensure_ascii` for saving the transcripts + fix naming of config class.

**Collection**: ASR

# Changelog 
- CommonVoice support for `convert_hf_dataset_to_nemo` script.

# Usage

```bash
python scripts/speech_recognition/convert_hf_dataset_to_nemo.py \
        output_dir=/data/commonvoice/ \
        path="mozilla-foundation/common_voice_11_0" \
        name="eo" \ # esperanto
        split="validation" \
        ensure_ascii=False \
        use_auth_token=True
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/scripts/speech_recognition/convert_hf_dataset_to_nemo.py
+++ b/scripts/speech_recognition/convert_hf_dataset_to_nemo.py
@@ -255,8 +255,10 @@ def convert_offline_dataset_to_nemo(
             del sample['audio']
             if 'file' in sample:
                 del sample['file']
-
-            manifest_f.write(f"{json.dumps(sample, ensure_ascii=cfg.ensure_ascii)}\n")
+            if cfg.ensure_ascii:
+                manifest_f.write(f"{json.dumps(sample, ensure_ascii=cfg.ensure_ascii)}\n")
+            else:
+                manifest_f.write(f"{json.dumps(sample, ensure_ascii=cfg.ensure_ascii)}")
 
 
 def convert_streaming_dataset_to_nemo(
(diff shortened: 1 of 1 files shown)


scripts/speech_recognition/convert_hf_dataset_to_nemo.py
```diff
@@ -255,11 +256,11 @@ def convert_offline_dataset_to_nemo(
             if 'file' in sample:
                 del sample['file']
 
-            manifest_f.write(f"{json.dumps(sample)}\n")
+            manifest_f.write(f"{json.dumps(sample, ensure_ascii=cfg.ensure_ascii)}\n")
```
titu1994, line 259: If you are using ensure ASCII, note that you cannot have the trailing \n that this code will generate. Manifest will not be parsed correctly if there is trailing \n + ensure ascii

SeanNaren: Could you explain how so @titu1994? We've been disabling it when formatting the esperanto dataset as done here: https://github.com/andrusenkoau/NeMo/blob/esperanto_example/docs/source/asr/examples/esperanto_asr/scripts/get_commonvoice_data_v2.py#L96

Do you mean when `ensure_ascii` is True? This is the default value afaik (so we've been using it for all our manifests).

titu1994: You can simulate it with a manifest built with ensure ASCII false and a trailing new line, then run the char, and use ASR manifest processor to parse it. It will crash inside of the collection

SeanNaren: Done! thanks @titu1994 :)

SeanNaren:
I ran this with `ensure_ascii=True` and `ensure_ascii=False`. I then ran the manifest with the below code:

```python
from nemo.collections.asr.data.audio_to_text import ASRManifestProcessor
from nemo.collections.common.parts.preprocessing.parsers import CharParser


x = ASRManifestProcessor(
    manifest_filepath='/data/commonvoice/mozilla-foundation/common_voice_11_0/eo/validation/validation_mozilla-foundation_common_voice_11_0_manifest.json',
    parser=CharParser(['a','b','c'])
)
```

Both loaded successfully, seems no issue with the manifests @titu1994

titu1994 (approved):
LGTM, thanks !

PR #5805, opened 2023-01-16
[TTS][DE] Augment tokenization/G2P to preserve capitalization of words and mix phonemes with word-level graphemes for an input text.

XuesongYang:

# What does this PR do ?

```
1. support grapheme case options with either "upper", "lower", or "mixed".
2. automatically prepend a symbol "#" or any customized one to each grapheme of a word if both phonemes and graphemes are mixed in order to distinguish graphemes from phoneme symbols if there are overlaps in between.
3. normalize texts with "NFC" form to represent the letter with diacritics as a single letter. Similar reason as https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/g2p/data/data_utils.py#L140.
4. add de IPA dictionary and de heteronym set.
5. fix typos. 
```

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo_text_processing/g2p/modules.py
```diff
@@ -404,39 +474,44 @@ def _normalize_dict(self, phoneme_dict_obj: dict) -> Tuple[DefaultDict[str, List
...
-            else:
-                word_new = word.lower()
+            word_new = set_grapheme_case(word, case=self.grapheme_case)
 
             # add grapheme symbols if `use_chars=True`.
             if self.use_chars:
-                # remove punctuations after a word.
+                # remove punctuations within a word. Punctuations can exist at the start, middle, and end of a word.
+                # TODO @xueyang: is the removal of a hyphen within a word expected, e.g. up-to-date -> uptodate?
                 word_no_punct = self.PUNCT_REGEX.sub('', word_new)
```
XuesongYang, line 485:
```
import re
PUNCT_REGEX = re.compile(fr"[^{LATIN_CHARS_ALL}\d]")
PUNCT_REGEX.sub("", "up-to-date")
Out[11]: 'uptodate'
```

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -582,22 +585,30 @@ def __init__(
...
+        Tokenize the `g2p_text` that has been already run through G2P. Each item in the `g2p_text` would be encoded as
+        one of the integer IDs predefined in `self._token2id`. Note that this function should be called after
+        `self.text_preprocessing_func` and `self.g2p` functions
 
         Args:
-            g2p_text: G2P's output, could be a mixture of phonemes and graphemes,
-                e.g. "see OOV" -> ['ˈ', 's', 'i', ' ', 'O', 'O', 'V']
-            raw_text: original raw input
+            g2p_text (List[str]): a sequence of tokens from G2P's output. It could be a sequence of phonemes, a sequence
+                of graphemes, or a mixture of both. For example, `['ˈ', 's', 'i', ' ', '#O', '#O', '#V']`, which is the
```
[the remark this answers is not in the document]
XuesongYang: sure. will add that.

nemo_text_processing/g2p/data/data_utils.py
```diff
@@ -160,6 +160,18 @@ def german_text_preprocessing(text: str) -> str:
     return ''.join(res)
 
 
+def any_locale_text_preprocessing(text: str) -> str:
+    return german_text_preprocessing(text)
+
+
+def normalize_unicode_text(text: str) -> str:
+    # normalize word with NFC form
```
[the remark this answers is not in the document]
XuesongYang: sure will add that, same as here:  https://github.com/NVIDIA/NeMo/blob/main/nemo_text_processing/g2p/data/data_utils.py#L140

nemo_text_processing/g2p/modules.py
```diff
@@ -271,37 +277,52 @@ def __init__(
...
+            phoneme_probability (Optional[float]): The probability (0.0 <= ε <= 1.0) that is used to balance the action
+                that a word in a sentence is whether transliterated into a sequence of phonemes, or kept as a sequence
+                of graphemes. If a random number for a word is greater than ε, then the word is kept as graphemes;
+                otherwise, the word is transliterated as phonemes. Defaults to None which is equivalent to setting it
+                to 1.0, meaning always transliterating the word into phonemes. Note that this code path is only run if
+                the word can be transliterated into phonemes, otherwise, if a word does not have an entry in the g2p
+                dict, it will be kept as graphemes. If a word has multiple pronunciations as shown in the g2p dict and
+                `ignore_ambiguous_words` is True, it will be kept as graphemes as well.
+            use_stresses (Optional[bool]): Whether to include the stress symbols (ˈ and ˌ).
+            grapheme_case (Optional[str]): Trigger converting all graphemes to uppercase, lowercase, or mix-cases.
```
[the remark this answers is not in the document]
XuesongYang: the words in the dictionary and the words in the original text both are converted into the same case, either upper or lower. It indicates that the processed words with `grapheme_case` in the dict and original text should be case-matched. While for the mixed-case option, the word in the original text should not change, while we need a mixed-case word in dict in order to map its pronunciation accordingly. **It is required to have mixed-case words when preparing dictionaries**, but I did not enforce it for now. 

So to answer your question, if `grapheme_case="mixed"`, and if words in the dictionary are all upper-cased. Then g2p will not find corresponding pronunciations because "Query" will not match "QUERY q w eh r iy".

nemo_text_processing/g2p/modules.py
```diff
@@ -271,37 +277,52 @@ def __init__(
         use_chars: bool = False,
         phoneme_probability: Optional[float] = None,
         use_stresses: Optional[bool] = True,
-        set_graphemes_upper: Optional[bool] = True,
+        grapheme_case: Optional[str] = "upper",
+        grapheme_prefix: Optional[str] = "#",
```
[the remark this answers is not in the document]
XuesongYang: sure. I will revert to `grapheme_prefix=""` by default.

nemo_text_processing/g2p/modules.py
XuesongYang: @ekmb @redoctopus could you please confirm if it needs fixed?

nemo_text_processing/g2p/modules.py
```diff
@@ -348,36 +373,82 @@ def __init__(
...
+                # validate dict entry format
+                assert isinstance(
+                    prons, list
+                ), f"Pronunciation type <{type(prons)}> is not supported. Please convert to <list>."
+
+                # normalize word with NFC form
+                word = normalize_unicode_text(word)
+
+                # normalize phonemes with NFC form
+                prons = [[normalize_unicode_text(p) for p in pron] for pron in prons]
```
redoctopus, line 445: Do we expect to need to normalize the phonemes? Is it safe to assume that we can conflate two distinct IPA symbols that are the same after normalization?

nemo_text_processing/g2p/modules.py
redoctopus: I think it should be fine since this part is just to update the symbol set without including punctuation, and doesn't affect the word in the long-term. I would definitely count the hyphen as a punctuation character over a "grapheme" character, if that makes sense.

The only issue would be if the user doesn't include the hyphen in the punctuation list, in which case the entry would be invalid; I assume this would be intended if that's a deliberate decision...

nemo_text_processing/g2p/modules.py
```diff
@@ -404,39 +478,54 @@ def _normalize_dict(self, phoneme_dict_obj: dict) -> Tuple[DefaultDict[str, List
...
             # add grapheme symbols if `use_chars=True`.
             if self.use_chars:
-                # remove punctuations after a word.
+                # remove punctuations within a word. Punctuations can exist at the start, middle, and end of a word.
+                # TODO @xueyang: is the removal of a hyphen within a word expected, e.g. up-to-date -> uptodate?
                 word_no_punct = self.PUNCT_REGEX.sub('', word_new)
-                symbols.update(word_no_punct)
 
+                # add prefix to distinguish graphemes from phonemes.
+                symbols.update(self._prepend_prefix_for_one_word(word_no_punct))
```
redoctopus, line 492: Do we want to prepend the grapheme symbol here? This would just add it to the symbols dictionary, which I'm not sure we want to do if we want to count [# + grapheme] as one token post-processing, and not as user input with # as punctuation.

nemo_text_processing/g2p/modules.py
```diff
@@ -404,39 +478,54 @@ def _normalize_dict(self, phoneme_dict_obj: dict) -> Tuple[DefaultDict[str, List
...
 
+            # duplicate word entries if grapheme_case is mixed. Even though grapheme_case is set to mixed, the words in
+            # the original input text and the g2p_dict remain unchanged, so they could still be either lowercase,
+            # uppercase, or mixed-case as defined in `set_grapheme_case` func. When mapping an uppercase word, e.g.
+            # "HELLO", into phonemes using the g2p_dict with {"Hello": [["həˈɫoʊ"]]}, "HELLO" can't find its
+            # pronunciations in the g2p_dict due to the case-mismatch of the words. Augmenting the g2p_dict with its
+            # uppercase word entry, e.g. {"Hello": [["həˈɫoʊ"]], "HELLO": [["həˈɫoʊ"]]} would provide possibility to
+            # find "HELLO"'s pronunciations rather than directly considering it as an OOV.
+            if self.grapheme_case == GRAPHEME_CASE_MIXED and not word_new.isupper():
+                g2p_dict[word_new.upper()] = prons_new
```
redoctopus, line 518: Why not just set them all to `.upper()` to avoid potential duplicate entries?

nemo_text_processing/g2p/modules.py
```diff
@@ -445,82 +534,111 @@ def is_unique_in_phoneme_dict(self, word: str) -> bool:
...
-                    pron.extend(["-"])
-                pron.pop()
-
-            prons.extend(pron)
+                # for example: (["NVIDIA", "unchanged"], True). "NVIDIA" is considered as a single token.
+                prons.extend([f"{self.grapheme_prefix}{word}" for word in words])
+            else:
+                assert (
+                    len(words) == 1
+                ), f"{words} should only has a single item when `without_changes` is False, but found {len(words)}."
```
redoctopus, line 623: Nitpick: s/has/have

nemo_text_processing/g2p/modules.py
XuesongYang: this is not an expected behavior if setting all words to `upper()`. For example, if you look at German's dictionary, you will find different pronunciations between words `Weg` and `weg`.
```
Weg  vˈeːk
weg  vˈɛk
```

nemo_text_processing/g2p/modules.py
```diff
@@ -348,36 +369,82 @@ def __init__(
...
-            else heteronyms
-        )
-        if set_graphemes_upper and heteronyms:
-            self.heteronyms = [het.upper() for het in self.heteronyms]
+        if isinstance(heteronyms, str) or isinstance(heteronyms, pathlib.Path):
+            self.heteronyms = set(self._parse_file_by_lines(heteronyms))
+        elif isinstance(heteronyms, list) and all(isinstance(het, str) for het in heteronyms):
+            self.heteronyms = set(heteronyms)
+        else:
+            self.heteronyms = None
```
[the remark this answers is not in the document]
XuesongYang: > if they've passed something in that's not being used

This will not happen because we had type hints `heteronyms: Optional[Union[str, pathlib.Path, List[str]]] = None`, which only consider str, pathlib.Path, List, or None.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -610,7 +621,7 @@ def encode_from_g2p(self, g2p_text: List[str], raw_text: Optional[str] = None):
             elif (p in self.punct_list) and self.punct:
                 # Add punct
                 ps.append(p)
-            elif p != space:
+            else:
```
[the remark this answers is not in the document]
XuesongYang: good catch! this should be reverted because we don't want to mess up warnings to users for skipping a whitespace.

nemo_text_processing/g2p/modules.py
XuesongYang: the motivation here is similar to graphemes. In IPA symbol sets, it usually combines a base token with its diacritics, like a nasalized `/ẽ/`. Without this normalization, `/ẽ/` would be encoded like `/e/ + / ̃/`

nemo_text_processing/g2p/modules.py
XuesongYang: agreed. Thanks for clarification. I will keep as-is and removed the TODO comments.

nemo_text_processing/g2p/modules.py
redoctopus: Ah, I see! Didn't realize that there were cases like this. Very interesting.

nemo_text_processing/g2p/modules.py
XuesongYang: yes, IIUC, this is the right place to prepend a special symbol. `# + grapheme` is actually one token as the key in `self._token2id` dict. We don't use the predefined symbol set by the user, instead, we automatically summarize the grapheme and phoneme set from the dictionary. Without prepending the special symbol, it would not distinguish the token that overlaps grapheme and phoneme set.

redoctopus: Can you confirm that it adds each `# + grapheme` as one symbol? Or does it add `#` separately from the grapheme?

XuesongYang: yes. the unit test explains: 
```
tests/nemo_text_processing/g2p/test_modules.py::TestIPAG2P::test_normalize_dict_with_graphemes_and_phonemes PASSED
```

XuesongYang: and your comments remind me of adding a validation func that avoid `grapheme_prefix` overlaps with character and punctuation lists.

redoctopus: Got it, sounds good to me then! And thank you for adding the tests :)

redoctopus (approved):
LGTM!

ekmb (approved):

Issue #5637, opened 2022-12-15
Speaker Diarization Task: Clustering stage crashes(Process killed) for long audio files using CPU


gabitza-tech:
Hello,

I've been testing the NeMo diarization example, using clustering inference on many different datasets and inputs using CPU only. On files up to 90 minutes in length, the diarizer reached around 25Gb RAM consumed in the clusterization stage. When I tried to diarize a 4 hour file that has 2 speakers using CPU only, the process was killed in the clusterization stage, no matter the configuration file used.

The configuration I used:
- clustering_diarizer/offline_diar_infer
- vad_multilingual_marblenet
- embeddings extraction: titanet_l
- configurations: either meeting or general ( both crashed)

The way I solved the process crashing problem was by splitting the 4 hour audio in 4 parts, and then merge the predicted rttms at the end by taking in consideration the offset between parts and checking if the speaker labels match. The issue is that the cluster labels on each file may differ from the other files, even if it is the same original audio split into 4. With only 2 speakers, it may be easy to check and switch the speakers labels so they match between all corresponding parts, but when there are more than 2 speakers, it's impossible.

My questions are: is there a way to avoid the process being killed when using long audio files(above 2 hours for example) on CPU(I can't used GPU)? If there isn't, is there a way to match the speakers from each split from the original audio? (example: speaker 1 in part 1 may be corresponding to speaker 4 in part 2) I was thinking about extracting an mean affinity matrix for each speaker in each part and then calculating the cosine distance between all speakers. 

Any solution, idea or feature that is already implemented would be greatly appreciated!

tango4j:
Hi, sorry for the late reply. Currently, the clustering algorithm has limitation on 60~70 min audio (depending on the scale length). Currently, we are working on clustering algorithm that handles long-form audio inputs. This feature will be added along with online diarization capability.

gabitza-tech:
Hi, thank you for you response! Looking forward to the implementation!

gabitza-tech:
@tango4j I would like to ask you if there is an expected time for the feature to be released at the moment?

I would also like to make two proposals for solving this issue, from my research and what I have tested so far:

[...]

PR #5876, opened 2023-01-27
Reduce memory usage in getMultiScaleCosAffinityMatrix function

gabitza-tech:
Reduce memory usage in getMultiScaleCosAffinityMatrix function

Signed-off-by: gabitza-tech <gabriel.pirlogeanu@gmail.com>

# What does this PR do ?

Replace some tensor multiplication in the function for multi scale cos affinity matrix calculation with a single tensor in which we add the repeated embeddings multiplied with its scale weight.

Results show a decrease from 11GB to 2GB for 1 hour audio. It even managed to reduce from above 60GB memory usage for a [53000,53000] tensor to 28GB.
**Collection**: ASR (collections/asr/parts/utils/offline_clustering.py)

# Changelog 
`multiscale_weights = torch.squeeze(multiscale_weights,dim=0).to(device)`
The tensor was [1,6] shaped and had to squeeze the first dimension.
` fused_sim_d = torch.zeros(len(timestamps_in_scales[-1]),len(timestamps_in_scales[-1]))`
Initialized a tensor with the shape [s,s] with 0 values where s is the number of timestamps for the smallest scale. We will add the tensor for each scale to this tensor
` with torch.no_grad():  `
Gradients are not needed for inference.
`fused_sim_d = fused_sim_d + multiscale_weights[scale_idx] * repeated_tensor_1`
Instead of adding the tensor for each scale to a list, then stacking the tensor list and then matrix multiplying it with the weights, we simply add the repeated tensor to the previously initialized tensor before the for loop.

@tango4j is adding followings:
- Clear memory outside torch.jit.script functions since jit-script does not support torch.cuda.empty_cache()
- Fixed the bug where eigen value decomposition does not use GPU. Now GPU usage hits 100%
- Removed torch.jit.script decorators since it slows down if we run without jit.script export.
- Performance (without torch.jit.script) 

sess len in secs | Num of sessions | total time in secs | sec/Iter. | RTFx |  Peak Memory Usage
-- | -- | -- | -- | -- | --
1800 | 10 | 34 | 3.48 | 517.24138 | 3G
3600 | 10 | 70 | 7.02 | 512.82051 | 8G
7200 | 10 | 319 | 31.98 | 225.14071 | 20G

Now, 2hours of audio usually takes 30~40 seconds.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. I would like to ask @titu1994, @redoctopus, @jbalam-nv, or @okuchaiev to review this pull request.
# Additional Information
* Related to https://github.com/NVIDIA/NeMo/issues/5637


nemo/collections/asr/parts/utils/offline_clustering.py
```diff
@@ -486,7 +486,7 @@ def getMultiScaleCosAffinityMatrix(
     timestamps_in_scales: List[torch.Tensor],
     device: torch.device = torch.device('cpu'),
 ) -> torch.Tensor:
-    """
+      """
```
tango4j, line 489: Did you make this tab intentionally? seems like it is causing style error in the test.

nemo/collections/asr/parts/utils/offline_clustering.py
```diff
@@ -506,20 +506,22 @@ def getMultiScaleCosAffinityMatrix(
...
+
+    for scale_idx in scale_list:       
+        with torch.no_grad():            
+            mapping_argmat = session_scale_mapping_list[scale_idx]
+            emb_t = embeddings_in_scales[scale_idx].half().to(device)
+            score_mat_torch = getCosAffinityMatrix(emb_t)
+            repeat_list = getRepeatedList(mapping_argmat, torch.tensor(score_mat_torch.shape[0])).to(device)
+            repeated_tensor_0 = torch.repeat_interleave(score_mat_torch, repeats=repeat_list, dim=0)            
+            repeated_tensor_1 = torch.repeat_interleave(repeated_tensor_0, repeats=repeat_list, dim=1)
+            fused_sim_d = fused_sim_d + multiscale_weights[scale_idx] * repeated_tensor_1
```
tango4j, line 524: Thank you for accepting my suggestion and pushing this.
Wouldn't it be better to clear memory here by doing the following:

```python
del emb_t, repeated_tensor_0, repeated_tensor_1
torch.empty_cache()
```
I need to check whether this helps or not.

tango4j: This doesn't work in torch.jit.script decorator. I have added torch.empty.cache() outside the clustering code and now it is clearning memory.

nemo/collections/asr/parts/utils/offline_clustering.py
tango4j: Fixed this on my own.

tango4j (approved):

tango4j:
This is passing the test and seems like working well. I will merge this as soon as the newly rebased one passes.

tango4j (approved):
The suggested changes have been reflected.

gabitza-tech:
@tango4j I wanted to ask if removing the torch.jit.script leads to the same performances in terms of memory usage as before?

donishi:
Also @tango4j, could you please share the configuration you used in the table you added to the PR description? I'm running out of GPU memory (on an RTX3090) when trying this PR with an audio two hours long. This happens with the clustering diarizer + 'general' configuration as well as with the neural diarizer.

yoelcabo:
hey @donishi,

I managed to run a 2 hours and a half audio on a RTX4000 (8GB) with this config:

```
  diarizer.speaker_embeddings.parameters.window_length_in_sec="[3.0,1.5,1.0]" \
  diarizer.speaker_embeddings.parameters.shift_length_in_sec="[1.5,0.75,0.5]" \
  diarizer.speaker_embeddings.parameters.multiscale_weights="[0.33,0.33,0.33]" \
```

With smaller `shift_length_in_sec` and `window_length_in_sec` I was running out of memory. So on top of reducing the number of scales, reducing the amount of windows also helps.

donishi:
That makes sense @yoelcabo, thanks for sharing!

For some reason that setting uses much more than 8GB for me, but it certainly uses less memory than the default window lengths.

yoelcabo:
One other thing I changed and forgot to mention is the batch size:

```
  batch_size=32 \
```

Otherwise, I run out of memory when extracting the embeddings.

tango4j:
@gabitza-tech Correct, removing torch.jit.script does not affect memory allocations
@donishi Each user has differnt GPUs so it's meaningless to share my script. as @yoelcabo described, try with higher base scale parameter (1.0 window, 0.5 hoplength).

Currently, I am also testing low-memory low-speed batch mode for clustering since so many users are raising this issue. I hope we can have clustering system that is also based on batch_size to control the memory usage. 

The memory allocation happens when embedding vectors in  `embedding_in_scales` are loaded to GPU memory.

PR #5879, opened 2023-01-28
Dynamic freezing in Nemo

trias702:
Signed-off-by: Daniel Egert <degert@nvidia.com>

# What does this PR do ?

Allows you to selectively freeze modules in your model for the first N steps during training

**Collection**: Core

# Changelog 
- changed core/classes/modelPT.py to do dynamic freezing via PTL hooks
- updated core documentation at docs/source/core/core.rst

# Usage
You can add the following to your config.yaml:

```yaml
  model:
    ...
    freeze_updates:
      enabled: true  # set to false if you want to disable freezing
      
    modules:   # list all of the modules you want to have freezing logic for
      encoder: 200      # module will be frozen for the first 200 training steps
      decoder: [50, -1] # module will be frozen at step 50 and will remain frozen until training ends
      joint: [10, 100]  # module will be frozen between step 10 and step 100 (step >= 10 and step <= 100)
      transcoder: -1    # module will be frozen for the entire training run
```

Setting `enabled: false` will disable all of the freezing logic.

**Pre checks**:
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


titu1994 (commented):
@VahidooX  for final review.
I asked for the code to not be part of init() - since all subclasses call it and it would be problematic.

As to logic - it will freeze given module for some steps then unfreeze and train them. Apparantly some model will not fine-tune properly without this strategy.

As to docs, maybe we can add it to one of the tutorials? Or somewhere in the docs?

nemo/collections/asr/models/asr_model.py
```diff
@@ -136,9 +136,56 @@ def on_after_backward(self):
...
-                logging.warning(f'detected inf or nan values in gradients! Setting gradients to zero.')
+                logging.warning('detected inf or nan values in gradients! Setting gradients to zero.')
                 self.zero_grad()
 
+    def on_train_start(self):
+        super().on_train_start()
+
+        # dynamic freezing
+        # should fire only once, on the very first batch of training and never again
+        if not hasattr(self, '_freeze_cfg'):
```
VahidooX, line 147: One suggestion I have is to drop the enabled from the config and have something like this where -1 or None means not any freezing. It can make the config simpler. It is your call as the current one should also be OK:
```
  freeze_modules:
     preprocessor: -1
     encoder: -1
     joint: -1
```

nemo/collections/asr/models/asr_model.py
```diff
@@ -136,9 +136,56 @@ def on_after_backward(self):
...
+        # dynamic freezing
+        if hasattr(self, '_freeze_cfg') and self._freeze_cfg is not None:
+            if self.training and hasattr(self, "trainer") and self.trainer is not None:
+                num_updates = self.trainer.global_step + 1
+
+                for ml, m_steps in self._freeze_cfg['modules'].items():
+                    # we could do hasattr check here, but it's too expensive for each step
+                    # consequently you'll throw an error if the module name doesn't exist
+                    # or was spelled wrong in the config.yaml
+                    if num_updates <= m_steps and not self._freeze_cfg['is_frozen'][ml]:
```
VahidooX, line 174: Other suggestion which can be done later in future is to add a support of passing start and end steps for freezing like this. -1 means to the end:
  freeze_updates:
    enabled: true

```
    modules:
      preprocessor: 50
      encoder: [10, 500]
      joint: [600, -1]

```

nemo/collections/asr/models/asr_model.py
```diff
@@ -136,9 +136,56 @@ def on_after_backward(self):
                 torch.distributed.all_reduce(valid_gradients, op=torch.distributed.ReduceOp.MIN)
 
             if valid_gradients < 1:
-                logging.warning(f'detected inf or nan values in gradients! Setting gradients to zero.')
+                logging.warning('detected inf or nan values in gradients! Setting gradients to zero.')
                 self.zero_grad()
 
+    def on_train_start(self):
```
VahidooX, line 142: Would you please add some description of this feature in the nemo docs?

VahidooX (changes requested):
Mostly LGTM!
Just need some documentation. Also made some suggestions to consider.

nemo/collections/asr/models/asr_model.py
trias702: I originally wanted to do this, but Som explicitly requested that I add that global enabled flag. @titu1994 Can you please explain why we need that global enabled flag and can't do it like Vahid suggests?

nemo/collections/asr/models/asr_model.py
trias702: This is a PyTorch Lightning hook, it's not a part of Nemo. More info here: https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.core.hooks.ModelHooks.html

nemo/collections/asr/models/asr_model.py
trias702: I agree this would be something good to add later down the line. I'm curious to first release this initial version and see if people use it and find it helpful, and then maybe add this if there is demand for it? Or would you prefer I add this functionality from the getgo?

nemo/collections/asr/models/asr_model.py
titu1994: Yould have to update N different flags to disable it then, vs one arg for enabled=False

nemo/collections/asr/models/asr_model.py
titu1994: I'd prefer from the get go. It would be a breaking change later

trias702: Is this something we really need though? Have there been any papers or methodologies ever published wherein people would want to freeze layers in such a schedule form? Like the [10, 500] case, I can't really wrap my head around why that would ever be useful.

VahidooX: There may be cases where you may want to freeze them after some steps, then you may need to use something like [50,-1]. 
We can avoid later breaking change by supporting both cases. When it is a simple integer, it specifies the start and if it is a list it would specify a span?

nemo/collections/asr/models/asr_model.py
VahidooX: I meant adding some description about this freezing feature and how it works to the nemo docs.

nemo/collections/asr/models/asr_model.py
trias702: Okay, I will implement this, but I don't understand what the [50, -1] logic is supposed to do. Does that mean you freeze for the first 50 steps, or you freeze after you reach 50 and until the end (so only the first 50 steps are UNfrozen)?

nemo/collections/asr/models/asr_model.py
trias702: Okay, can you please send me a link to where these Nemo docs are?

nemo/collections/asr/models/asr_model.py
VahidooX: [50,-1] means start freezing at step 50 until the end. If there is a single integer, it means the start of freezing but when it is a list, it specifies a range where it is freezed.

nemo/collections/asr/models/asr_model.py
VahidooX: You may add s section in this file which explains how this freezing works:
https://github.com/NVIDIA/NeMo/blob/main/docs/source/asr/configs.rst

trias702: I have updated the docs as requested

nemo/collections/asr/models/asr_model.py
trias702: I haved committed new code which does this exactly as you have asked for

docs/source/asr/configs.rst
```diff
@@ -218,6 +218,31 @@ For example, a decoder config corresponding to a sub-word tokenization model sho
       vocabulary: []  # filled with vocabulary from tokenizer at runtime
 
 
+Dynamic Layer Freezing
+---------------------------------
+
+You can selectively freeze any modules inside a Nemo model by specifying a freezing schedule in the config. Freezing stops any gradient updates
```
VahidooX, line 224: It just works for ASR models, not all nemo models.

docs/source/asr/configs.rst
```diff
@@ -218,6 +218,31 @@ For example, a decoder config corresponding to a sub-word tokenization model sho
       vocabulary: []  # filled with vocabulary from tokenizer at runtime
 
 
+Dynamic Layer Freezing
+---------------------------------
```
VahidooX, line 222: The dashes should have the same length as the title.

nemo/collections/asr/models/asr_model.py
```diff
@@ -136,9 +136,61 @@ def on_after_backward(self):
...
+        # dynamic freezing
+        if hasattr(self, '_freeze_cfg') and self._freeze_cfg is not None:
+            if self.training and hasattr(self, "trainer") and self.trainer is not None:
+                num_updates = self.trainer.global_step + 1
+
+                for ml, m_steps in self._freeze_cfg['modules'].items():
+                    # we could do hasattr check here, but it's too expensive for each step
+                    # consequently you'll throw an error if the module name doesn't exist
+                    # or was spelled wrong in the config.yaml
+                    if isinstance(m_steps, list):
```
VahidooX, line 174: Have you tested this list feature?
I think the list which comes a conf has type of ListConfig, not list.

VahidooX (changes requested):

nemo/collections/asr/models/asr_model.py
trias702: Yes, I have tested it and it works. The type is actually a list, not ListConfig. If you look at line 153 in that same file, you will see we call `setattr(self, '_freeze_cfg', OmegaConf.to_container(self.cfg.freeze_updates))` which casts from OmegaConf types down to Pythonic types.

docs/source/asr/configs.rst
trias702: NA since we have now transferred to ModelPT

docs/source/asr/configs.rst
trias702: Fixed

titu1994 (approved):
It looks great.

docs/source/core/core.rst
```diff
@@ -756,3 +756,30 @@ It also means that ``.forward(...)`` and ``__call__(...)`` methods each produce
...
+To enable freezing, add the following to your config:
+
+.. code-block:: yaml
+
+  model:
+    ...
+    freeze_updates:
+      enabled: true  # set to false if you want to disable freezing
+      
+    modules:   # list all of the modules you want to have freezing logic for
```
VahidooX, line 780: This section needs more indentation.

VahidooX (dismissed):
LGTM! Just left one comment.

docs/source/core/core.rst
trias702: Fixed

VahidooX (approved):

PR #5880, opened 2023-01-28
ONNX export for RadTTS

borisfom:

# What does this PR do ?
ONNX export for RadTTS is now functional. 
Working around: https://github.com/pytorch/pytorch/pull/91526

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/tts/helpers/helpers.py
```diff
@@ -547,12 +547,11 @@ def regulate_len(
     reps = durations.float() / pace
     reps = (reps + 0.5).floor().long()
     dec_lens = reps.sum(dim=1)
-
     if replicate_to_nearest_multiple:
```
blisc, line 550: Can we test this function?

nemo/collections/tts/models/radtts.py
```diff
@@ -443,12 +442,13 @@ def input_example(self, max_batch=1, max_dim=400):
         par = next(self.parameters())
         sz = (max_batch, max_dim)
         # sz = (max_batch * max_dim,)
-        inp = torch.randint(0, 94, sz, device=par.device, dtype=torch.int64)
+        # Pick up only pronouncible tokens
+        inp = torch.randint(32, 64, sz, device=par.device, dtype=torch.int64)
```
blisc, line 446: Why is this needed?

nemo/collections/tts/modules/common.py
```diff
@@ -82,35 +82,36 @@ def __init__(self, input_size, hidden_size, num_layers=1, lstm_norm_fn="spectral
...
-        # this is only needed for Torchscript to run in Triton
+        # autocast guard is only needed for Torchscript to run in Triton
         # (https://github.com/pytorch/pytorch/issues/89241)
         with torch.cuda.amp.autocast(enabled=False):
-            ret = self.lstm_tensor(context.to(dtype=torch.float32), lens, enforce_sorted=True)
-        return ret[0].to(dtype=dtype)[unsort_ids]
+            return self.lstm(context.to(dtype=torch.float32), lens).to(dtype=dtype)
+
+    def forward(self, context: Tensor, lens: Tensor) -> Tensor:
+        self.bilstm.flatten_parameters()
```
blisc, line 111: Do we need this? It's called in init

nemo/collections/tts/modules/common.py
```diff
@@ -158,13 +159,13 @@ def __init__(
             self.dense = nn.Linear(n_channels, out_dim)
 
     def forward(self, context: Tensor, lens: Tensor) -> Tensor:
-        mask = get_mask_from_lengths(lens, context)
+        mask = get_mask_from_lengths(lens)
         mask = mask.to(dtype=context.dtype).unsqueeze(1)
+        context = context[:, :, : mask.shape[-1]]
```
blisc, line 164: Why are we dropping the masking here?

nemo/collections/tts/helpers/helpers.py
borisfom: Well I did quite a bit of ad hoc testing - what exactly do you have in mind ?

nemo/collections/tts/models/radtts.py
borisfom: This was my attempt to make sure there are no entries of zero length generated. LSTM cannot handle 0-length elements in input. That did not help 100% percent - also there was f0_pad that would add zeroes - so I had to modify clip() to ensure non-zero lengths are generated

nemo/collections/tts/modules/common.py
borisfom: Well that may be required for some distributed envirionments. It amounts to almost no-op when weights are already flat.

nemo/collections/tts/modules/common.py
borisfom: Why, it's not dropping it - it's making context the shape of the mask, not vice versa. Required to match ONNX operation. I actually am already moving that truncation higher up to radtts.py so it would be more intuitive and more efficient.

nemo/collections/tts/helpers/helpers.py
blisc: A small unit test in https://github.com/NVIDIA/NeMo/tree/main/tests/collections/tts that tests input of three size: [group_size-1, group_size, group_size+1]

nemo/collections/tts/models/radtts.py
blisc: I don't understand. The LSTMs will see either the text_length input, or the spectrogram_length input. For text, there will be no 0-length elements since each character has length 1. When expanded from text_len to spectrogram_length, there can be 0-length elements but those characters wouldn't exist in the expanded tensor. I don't see why the LSTM will ever see a 0-length element

nemo/collections/tts/modules/common.py
blisc: Can you point me to where the masking is done now?

blisc: Actually, I see that now you reverted this change. Was that intentional?

nemo/collections/tts/helpers/helpers.py
borisfom: Well, perhaps that test should have been added when replicate was first introduced (with a bug :) ). 
Anyway, you got it: added test_helpers.py

nemo/collections/tts/models/radtts.py
borisfom: Well, I definitely was encountering 0s - printed them out and observed by myself. Has to modify clamp() call for that. If that is not something expected to happen, we should investigate.

nemo/collections/tts/modules/common.py
borisfom: Yes - I have moved slicing up (and out), so mask length is determined by input context, as before.

nemo/collections/tts/models/radtts.py
blisc: Ok, will take a look after this PR

blisc (approved):

MikyasDesta (changes requested):
Can we change the sigma values to 0.7 ?

MikyasDesta (changes requested):

nemo/collections/tts/modules/radtts.py
```diff
@@ -581,13 +582,13 @@ def infer(
         self,
         speaker_id,
         text,
-        sigma,
-        sigma_txt=0.8,
-        sigma_f0=0.8,
-        sigma_energy=0.8,
+        sigma=0.0,
```
MikyasDesta, line 585: Can we change the sigma value to 0.7?

MikyasDesta (changes requested):

nemo/collections/tts/modules/radtts.py
borisfom: This is botched diff - it is at 0.7

blisc: Let's make `forward_for_export` and `infer` have the same default values. In case, let's put `sigma` for both at 0.7

borisfom: Done.

blisc (approved):

blisc (approved):

PR #5907, opened 2023-02-02
Ensure tokens-in-batch is always an upper bound

gabrielbretschner:
# What does this PR do ?

This PR fixes an edge case where a batch may contain more tokens than allowed by `tokens-in-batch`

**Collection**: NLP

# Changelog 
- In case a batch that contains too many tokens has fewer than 8 sentences the last sentence is evicted to a new batch to ensure we don't break the `tokens-in-batch` bound
- Before it would not evict any sentence leading to batches that are too big.

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


gabrielbretschner:
I'd really like to get a review. Maybe @MaximumEntropy?

AlexGrinch (approved):

MaximumEntropy:
@gabrielbretschner thanks for the contribution. Can you please add the appropriate license header to your unittest file?

PeganovAnton (changes requested):
Could you please move the test into `tests/collections/nlp`?

gabrielbretschner:
@MaximumEntropy Added the Licence header to the test file.

Numeri:
@AlexGrinch @MaximumEntropy Any way we could get this fix merged?

gabrielbretschner:
@MaximumEntropy @AlexGrinch Is there anything I can do to get this merged?

PR #5910, opened 2023-02-02
Refactor the retrieval services for microservice architecture 

yidong72:
# What does this PR do ?

Refactor the retrieval services so each component can be deployed alone inside a container. This allows it to be served easily in a cloud. Also fixed some issues with zero neighbor retrieval.


soumye (approved):
The refactor LGTM!

PR #5911, opened 2023-02-02
Moved eval notebook data to aws

redoctopus:

# What does this PR do ?

Moves the MCD DTW data tarball to AWS instead of storing it in NeMo.
Paths tested locally again.

**Collection**: TTS


Kipok (approved):
Looks good to me!

PR #5925, opened 2023-02-06
Support Alignment Extraction for all RNNT Beam decoding methods

titu1994:
# What does this PR do ?

Adds alignment extraction for all rnnt beam algorithms

**Collection**: [ASR]

# Changelog 
- Adds support for TSD, ALSD and mAES alignment extraction 
- Adds unittest that downloads a small model and runs on AN4 test set to compute alignments from greedy and beam search algorithms

# Usage
You can now submit `preserve_alignments=true` to in the RNNTDecodingConfig for all beam search algos and extract alignments.

```python

filepath = "<audio filepath>"

model = nemo_asr.models.ASRModel.from_pretrained('stt_en_conformer_transducer_large')  # type: nemo_asr.models.EncDecRNNTBPEModel
# Create default decoding config
model.change_decoding_strategy(None)
decoding_cfg = model.cfg.decoding
decoding_cfg.strategy = "beam"  # can be beam, tsd, alsd or maes
decoding_cfg.preserve_alignments = True  # the important argument !
decoding_cfg.beam.beam_size = 8
decoding_cfg.beam.return_best_hypothesis = False

# Prepare beam search decoding strategy
model.change_decoding_strategy(decoding_cfg)

best_hypothesis, all_hypothesis = model.transcribe([filepath], return_hypotheses=True)
print("Num hypothesis per sample :", len(best_hypothesis[0]))
print("Result type :", best_hypothesis[0][0].__class__.__name__)
for candidate in range(len(best_hypothesis[0])):
    print(f"Beam Search {candidate + 1} :", best_hypothesis[0][candidate].text, best_hypothesis[0][candidate].score)

# Use the following commented print statements to check
# the alignment of other algorithms compared to the default
all_hyps = all_hypothesis
for idx, hyp_ in enumerate(all_hyps):  # type: (int, rnnt_utils.Hypothesis)
    print("Hyp index", idx + 1, "text :", hyp_.text)

    for t in range(len(hyp_.alignments)):
        t_u = []
        for u in range(len(hyp_.alignments[t])):
            logp, label = hyp_.alignments[t][u]
            t_u.append(int(label))
        print(f"Tokens at timestep {t} = {t_u}")
    print()

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
@@ -266,10 +266,10 @@ class BeamRNNTInfer(Typing):
         self.maes_expansion_gamma = float(maes_expansion_gamma)
         self.maes_expansion_beta = int(maes_expansion_beta)
 
-        if self.maes_prefix_alpha < 0:
+        if self.search_type == 'maes' and self.maes_prefix_alpha < 0:
             raise ValueError("`maes_prefix_alpha` must be a positive integer.")
 
-        if self.vocab_size < beam_size + maes_expansion_beta:
+        if self.search_type == 'maes' and self.vocab_size < beam_size + maes_expansion_beta:
             raise ValueError(
                 f"beam_size ({beam_size}) + expansion_beta ({maes_expansion_beta}) "
                 f"should be smaller or equal to vocabulary size ({self.vocab_size})."
@@ -278,7 +278,7 @@ class BeamRNNTInfer(Typing):
         if search_type == 'maes':
             self.max_candidates += maes_expansion_beta
 
-        if self.maes_num_steps < 2:
+        if self.search_type == 'maes' and self.maes_num_steps < 2:
             raise ValueError("`maes_num_steps` must be greater than 1.")
 
         if softmax_temperature != 1.0 and language_model is not None:
@@ -582,6 +582,12 @@ class BeamRNNTInfer(Typing):
                 # keep those hypothesis that have scores greater than next search generation
                 hyps_max = float(max(hyps, key=lambda x: x.score).score)
                 kept_most_prob = sorted([hyp for hyp in kept_hyps if hyp.score > hyps_max], key=lambda x: x.score,)
+
+                temp_hyps = sorted([hyp for hyp in hyps], key=lambda x: x.score, reverse=True)
+                temp_kept_hyps =  sorted([hyp for hyp in kept_hyps], key=lambda x: x.score, reverse=True)
+                print("hyp max", hyps_max, "hyps", [(hyp.score, hyp.y_sequence) for hyp in temp_hyps])
+                print("kept hyps", [(hyp.score, hyp.y_sequence) for hyp in temp_kept_hyps])
+                print("final kept", [(hyp.score, hyp.y_sequence) for hyp in kept_most_prob])
 
                 # If enough hypothesis have scores greater than next search generation,
                 # stop beam search.
@@ -953,7 +959,7 @@ class BeamRNNTInfer(Typing):
         """
         if self.preserve_alignments:
             raise NotImplementedError(
-                "`preseve_alignments` is not implemented for Alignment-length Synchronous Decoding."
+                "`preseve_alignments` is not implemented for Modified Adaptive Expansion Search."
             )
 
         if partial_hypotheses is not None:
--- a/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
@@ -692,13 +692,13 @@ class BeamRNNTInfer(Typing):
                         # If the sequence is not in seq_A, add it as the blank token
                         # In this step, we dont add a token but simply update score
                         _temp_hyp = Hypothesis(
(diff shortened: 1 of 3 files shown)


tests/collections/asr/decoding/test_rnnt_decoding.py
```diff
@@ -0,0 +1,252 @@
...
+        with torch.no_grad():
+            hyps = beam(encoder_output=enc_out, encoded_lengths=enc_len)[0]  # type: rnnt_utils.Hypothesis
+            hyp = decode_text_from_greedy_hypotheses(hyps, model.decoding)
+            hyp = hyp[0]
+
+            assert hyp.alignments is not None
+
+            # Use the following commented print statements to check
+            # the alignment of other algorithms compared to the default
+            # print("Text", hyp.text)
```
Kipok, line 170: I usually prefer to leave the print statements "on" all the time - by default test output is not being printed, but if something fails, it's very helpful to see such messages right away without the need to change code and rerun

tests/collections/asr/decoding/test_rnnt_decoding.py
```diff
@@ -0,0 +1,252 @@
...
+                    for u in range(len(hyp_.alignments[t])):
+                        logp, label = hyp_.alignments[t][u]
+                        assert torch.is_tensor(logp)
+                        assert torch.is_tensor(label)
+
+                        t_u.append(int(label))
+
+                    # Blank token must be the last token in the current
+                    if len(t_u) > 1:
+                        assert t_u[-1] == blank_id
```
Kipok, line 245: Can we also add a check that all other tokens are non-blank? That was breaking in the bug-fix PR I did for batched alignment calculation, so might be a helpful check here as well

tests/collections/asr/decoding/test_rnnt_decoding.py
```diff
@@ -0,0 +1,252 @@
...
+
+    @pytest.mark.skipif(
+        not NUMBA_RNNT_LOSS_AVAILABLE, reason='RNNTLoss has not been compiled with appropriate numba version.',
+    )
+    @pytest.mark.with_downloads
+    @pytest.mark.unit
+    @pytest.mark.parametrize(
+        "beam_config",
+        [
+            {"search_type": "greedy"},
```
Kipok, line 191: I don't really know the details about all these algorithms, but if any of them are expected to produce exactly the same label sequence, that would be a nice thing to test for as well. Or maybe something like greedy = beam with beam size of 1. Very different code paths but should give equivalent results - always a great test to have, in my opinion.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -425,7 +429,8 @@ def greedy_search(
             not_blank = True
             symbols_added = 0
 
-            while not_blank and (symbols_added < self.max_candidates):
+            # TODO: Figure out how to remove this hard coding afterwords
+            while not_blank and (symbols_added < 5):
```
Kipok, line 433: Why do we have 5 here?

Kipok (commented):
Looks generally good to me with a few comments. I should say that I don't really know the details of all of those algorithms, so will be great if someone with more knowledge about them can also have a look before merging.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
titu1994: Heuristic. Depending on model stride (T / U ratio ~ 2) so the model could at worst predict 1 token every 2 timesteps - we give it some leeway with up to 5 tokens per timestep. Normally it will never reach such a high number, even for 8x stride.

tests/collections/asr/decoding/test_rnnt_decoding.py
titu1994: Agreed, made them print always

tests/collections/asr/decoding/test_rnnt_decoding.py
titu1994: It auto switches to greedy decoding at beam size 1 (so there's no actual beam search being done with beam size 1)

tests/collections/asr/decoding/test_rnnt_decoding.py
titu1994: Nice catch, will add.

tests/collections/asr/decoding/test_rnnt_decoding.py
Kipok: Maybe we can add some flag to allow to keep beam search logic with beam size of 1? Useful for the tests + might be quite useful for debugging issues as well.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
Kipok: Got it - maybe mention this in the comment, so that it's clear why we have this constant here

tests/collections/asr/decoding/test_rnnt_decoding.py
titu1994: The beam search logic doesn't work for size 1, but I can look info it in a separate pr

PR #5947, opened 2023-02-07
Move settings to `pyproject.toml`. Remove deprecated `pytest-runner`

artbataev:
# What does this PR do ?

- move settings from old-style `setup.cfg` to `pyproject.toml` (recommended by most of the tools)
- remove deprecated `pytest-runner` according to the [documentation](https://github.com/pytest-dev/pytest-runner#deprecation-notice)

# Changelog 
- create `pyproject.toml` with settings from `setup.cfg` for `isort` and `pytest`
- remove `setup.cfg`
- remove `pytest-runner` from `setup.py`

# Usage
It is not possible anymore to use `python setup.py test` (actually, not used in project). 

`pytest <options>` should be used for testing purpose.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[tool.isort]
+profile = "black"
+line_length = 119
+ignore_whitespace = true
+py_version = 38
```
titu1994, line 19: Can you inline some docstrings so that we know what to upgrade when you start using py3.9 as default min. Same for other args

pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+    "*.egg",
+    ".*",
+    "_darcs",
+    "build",
+    "CVS",
+    "dist",
+    "venv",
+    "{arch}"
+]
+markers = [
```
titu1994, line 50: Is this just docstring for the markers ? Where is the logic for markers still placed ? If it's in conf, can you add comment to update this docstring too if we create more markers

setup.py
```diff
@@ -134,8 +134,6 @@ def req_file(filename, folder="requirements"):
 
 extras_require['slu'] = list(chain([extras_require['slu'], extras_require['asr']]))
 
-tests_requirements = extras_require["test"]
```
titu1994, line 137: Why are these two being removed ?

artbataev: There is no need for `tests_requires` if we do not execute `pytest setup.py test`, see deprecation guide: https://github.com/pytest-dev/pytest-runner#deprecation-notice

These requirements still installed when `[all]` is used

pyproject.toml
artbataev: This is just a port for `markers` from `setup.cfg`, colon is the delimiter for `marker: docstring`
You can see all these markers by executing `pytest --markers`

pyproject.toml
artbataev: Added comments

pyproject.toml
artbataev: Also, I think it would be good to add permanently `--strict-markers` (to `addopts`) to mark tests with unregistered markers as failed (now it's just a warning that the marker is not registered, e.g there is a typo `@pytest.mark.uint` instead of `unit`). 

Would it be ok?

ericharper: Sounds okay to me

ericharper (dismissed):
LGTM. Thanks!

pyproject.toml
artbataev: Ok, I added `--strict-markers`

titu1994: Yeah makes sense to make it strict.

titu1994 (approved):

PR #6006, opened 2023-02-13
Use module-based k2 import guard

artbataev:
Signed-off-by: Vladimir Bataev <vbataev@nvidia.com>

# What does this PR do ?

Simplify `k2` library import guard: use module-based approach.

**Advantages:**
- less code is required for using guard
- no need to suppress linters (isort, etc.)

**Collection**: [ASR]

# Changelog 

# Usage

```python
from nemo.core.utils.k2_guard import k2  # import k2 with performing all required checks
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/core/utils/k2_guard.py
+++ b/nemo/core/utils/k2_guard.py
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Tuple
+"""
+Guard for importing optional NeMo dependency k2.
+Contains checks for k2 availability and version.
+Use `from nemo.core.utils.k2_guard import k2` to import k2 instead of direct import.
+If there is an error, the module will raise an exception with a helpful message.
+"""
+
+import textwrap
 
 __K2_MINIMUM_MAJOR_VERSION__ = 1
 __K2_MINIMUM_MINOR_VERSION__ = 11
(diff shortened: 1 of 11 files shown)


GNroy (approved):
LGTM

SeanNaren (approved):
LGTM!

nemo/core/utils/k2_guard.py
```diff
@@ -26,39 +38,20 @@
...
     "as different versions of k2 may not interact with the NeMo code as expected."
 )
-K2_IMPORT_SUCCESS = "Successfully imported k2."
-
 
-def k2_is_available() -> Tuple[bool, str]:
-    try:
-        import k2
+if not package_available("k2"):
+    raise ModuleNotFoundError(K2_INSTALLATION_MESSAGE)
```
titu1994, line 43: How is it an optional dependency if merely importing it will raise an error ? K2 is not installed by default, the check should be don't at init inside of where it will be used rather than during simply import of a file.

titu1994 (changes requested):

nemo/core/utils/k2_guard.py
titu1994: Currently, k2 is not included in the import path of init.py files for this reason, but at least the check is done at init of the class. This instead puts it in the import path so if used simply imports the file it will crash with helpful error. 

Numba checks specifically defers this crash to the moment user actually tries to use the module rather than simply import the module

artbataev: Importing `k2` raises an error only when importing `k2_guard` of `k2 from k2_guard`
`k2_guard` will not be imported and cached.

So, every module that tries to import `k2 from k2_guard` will raise an error (if no k2 / incorrect version) and will not be initialized (no classes in this module).
As a result, importing from **any** of the modules in nemo, that depend on `k2` (maybe not directly) will result in error with the necessary message, which is a correct behavior.

But importing not at the top level will allow to raise an error in runtime, we do not need to call `k2_import_guard()`:

```python
if backend == "k2":
    if criterion_type == "ml":
        from nemo.collections.asr.parts.k2.ml_loss import MLLoss as K2Loss  # maybe error here
    elif criterion_type == "map":
        from nemo.collections.asr.parts.k2.map_loss import MAPLoss as K2Loss  # or here
```

Also, **this doesn't change the previous behavior**: in modules `k2_import_guard()` call was used at the top level, so incorrect k2 version resulted in immediate crash **before** importing any defined class.

artbataev: You can set `__K2_MINIMUM_MINOR_VERSION = 111` and see that the behavior didn't change and everything works.

titu1994 (approved):
Approved for now, but K2 should switch to dynamic Numba style runtime check rather than static file import check.

GNroy (approved):

PR #6012, opened 2023-02-14
from_restore for NM models on HF plus popular properties

okuchaiev:
# What does this PR do ?

Add two popular read-only properties for tensor and pipeline parallelism values.
Change some of the self._cfg to self.cfg and cfg to self.cfg

Adds from_pretrained for NM Models on HF Hub

NLP

# Changelog 

# Usage

```python
print(model.tensor_model_parallel_size)
rint(model.pipeline_model_parallel_size)
```

**Pre checks**:


nemo/core/classes/modelPT.py
```diff
@@ -1704,6 +1704,33 @@ def on_train_end(self):
         if hasattr(self, '_freeze_cfg'):
             delattr(self, '_freeze_cfg')
 
+    @classmethod
+    def from_pretrained(
+        cls,
+        model_name: str,
```
titu1994, line 1710: This is not a requirement for ASR / TTS models. Should be moved up to NLPModel if such a constraint needs to be placed.

nemo/core/classes/common.py
```diff
@@ -938,14 +938,19 @@ def _get_hf_hub_pretrained_model_info(cls, model_name: str, refresh_cache: bool
             -   The path to the NeMo model (.nemo file) in some cached directory (managed by HF Hub).
         """
         # Resolve the model name without origin for filename
-        resolved_model_filename = model_name.split("/")[-1] + '.nemo'
+        if model_name.endswith('.nemo'):
```
titu1994, line 941: So the repo itself would hand ".Nemo" in the end of the name? Such as Nvidia/abc.nemo?

titu1994 (changes requested):
ModelPT change is incorrect for other domains

nemo/core/classes/common.py
okuchaiev: this is for the case when the user specifically specifies .nemo file path in repo (e.g. NLP models have several due to various parallelism configurations)

nemo/core/classes/modelPT.py
okuchaiev: Moved this to NLPModel. ModelPT is now unchanged

nemo/core/classes/common.py
titu1994: Right. I'm wondering what the signature is like 

ModelClass.from_pretrained("nvidia/???")

What is the value of ??? and how does the user get this value ? Normally it would be something like "Nvidia/stt_en_xyz" where Nvidia is user name / org name and stt_en_xyz is the model card name

PR #6013, opened 2023-02-14
InterCTC loss and stochastic depth implementation

Kipok:
# What does this PR do ?

Adding intermediate CTC loss and stochastic depth as described in https://arxiv.org/abs/2102.03216. 

The current implementation is only for conformer encoder, but I'm not really sure how to write a generic code for this case. Please let me know if you have some ideas here.

**Collection**: ASR

# Changelog 

# Usage

To use, specify parameters in the config. E.g., for stochastic depth:
```
model.encoder.stochastic_depth_mode=linear  
model.encoder.stochastic_depth_drop_prob=0.3 
model.encoder.stochastic_depth_start_layer=0
```

For intermediate CTC loss:
```
model.encoder.capture_output_at_layers=[9] 
model.intermediate_loss_weights=[0.3]
```

I've added the docs to ConformerEncoder, but not sure how to add the docs to the new `intermediate_loss_weights` parameter of the `EncDecCTCModel`. Please let me know what's the right place to put those docs in.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/docs/source/asr/configs.rst
+++ b/docs/source/asr/configs.rst
@@ -513,36 +513,6 @@ respectively. Some components of the configs of `Conformer-CTC <./models.html#Co
 These datasets are similar to other ASR models like `QuartzNet <./models.html#QuartzNet>`__. There should be a tokenizer section where you can
 specify the tokenizer if you want to use sub-word encoding instead of character-based encoding.
 
-Conformer-CTC also supports `InterCTC loss <https://arxiv.org/abs/2102.03216>`_. To use it, you need to specify
-2 parameters as in example below
-
-.. code-block:: yaml
-
-   model:
-      # ...
-      encoder:
-        # ...
-        capture_output_at_layers: [9]
-
-      intermediate_loss_weights: [0.3]
-
-which can be used to reproduce the default setup from the paper (assuming the total number of layers is 18).
-You can also specify multiple CTC losses from different layers, e.g., to get 2 losses from layers 4 and 9 with
-weights 0.1 and 0.3, specify:
-
-.. code-block:: yaml
-
-   model:
-      # ...
-      encoder:
-        # ...
-        capture_output_at_layers: [4, 9]
-
-      intermediate_loss_weights: [0.1, 0.3]
-
-Note that the final-layer CTC loss weight is automatically computed to normalize
-all weight to 1 (0.6 in the example about).
-
 
 The encoder section includes the details about the Conformer-CTC encoder architecture. You may find more information in the
 config files and also :ref:`nemo.collections.asr.modules.ConformerEncoder <conformer-encoder-api>`.
@@ -575,6 +545,38 @@ The config files for LSTM-Transducer and LSTM-CTC models can be found at ``<NeMo
 Most of the of the configs of are similar to other ctc or transducer models. The main difference is the encoder part.
 The encoder section includes the details about the RNN-based encoder architecture. You may find more information in the
 config files and also :ref:`nemo.collections.asr.modules.RNNEncoder <rnn-encoder-api>`.
+
+
+CTC Configurations
+------------------
+
+All CTC-based models also support `InterCTC loss <https://arxiv.org/abs/2102.03216>`_. To use it, you need to specify
+2 parameters as in example below
+
+.. code-block:: yaml
+
+   model:
+      # ...
+      interctc:
+        loss_weights: [0.3]
+        apply_at_layers: [8]
+
+which can be used to reproduce the default setup from the paper (assuming the total number of layers is 18).
+You can also specify multiple CTC losses from different layers, e.g., to get 2 losses from layers 3 and 8 with
+weights 0.1 and 0.3, specify:
+
+.. code-block:: yaml
+
+   model:
+      # ...
+      encoder:
+        # ...
+        capture_output_at_layers: [3, 8]
+
+      intermediate_loss_weights: [0.1, 0.3]
+
+Note that the final-layer CTC loss weight is automatically computed to normalize
+all weight to 1 (0.6 in the example about).
 
 
 Transducer Configurations
--- a/docs/source/asr/configs.rst
+++ b/docs/source/asr/configs.rst
@@ -547,8 +547,8 @@ The encoder section includes the details about the RNN-based encoder architectur
 config files and also :ref:`nemo.collections.asr.modules.RNNEncoder <rnn-encoder-api>`.
 
 
-CTC Configurations
-------------------
+InterCTC Config
+---------------
 
 All CTC-based models also support `InterCTC loss <https://arxiv.org/abs/2102.03216>`_. To use it, you need to specify
 2 parameters as in example below
@@ -574,7 +574,29 @@ weights 0.1 and 0.3, specify:
         apply_at_layers: [3, 8]
 
 Note that the final-layer CTC loss weight is automatically computed to normalize
-all weight to 1 (0.6 in the example about).
+all weight to 1 (0.6 in the example above).
+
+
+Stochastic Depth Config
+-----------------------
+
+`Stochastic Depth <https://arxiv.org/abs/2102.03216>`_ is a useful technique for regularizing ASR model training.
+Currently it's only supported for :ref:`nemo.collections.asr.modules.ConformerEncoder <conformer-encoder-api>`. To
+use it, specify the following parameters in the encoder config file to reproduce the default setup from the paper:
+
+.. code-block:: yaml
+
+   model:
+      # ...
+      encoder:
+        # ...
+        stochastic_depth_drop_prob: 0.3
+        stochastic_depth_mode: linear  # linear or uniform
+        stochastic_depth_start_layer: 0
+
+See :ref:`documentation of ConformerEncoder <conformer-encoder-api>` for more details. Note that stochastic depth
+is supported for both CTC and Transducer model variations (or any other kind of model/loss that's using
+conformer as encoder).
 
 
 Transducer Configurations
--- a/docs/source/asr/configs.rst
+++ b/docs/source/asr/configs.rst
@@ -1,12 +1,12 @@
 NeMo ASR Configuration Files
 ============================
 
-This section describes the NeMo configuration file setup that is specific to models in the ASR collection. For general information 
-about how to set up and run experiments that is common to all NeMo models (e.g. Experiment Manager and PyTorch Lightning trainer 
+This section describes the NeMo configuration file setup that is specific to models in the ASR collection. For general information
+about how to set up and run experiments that is common to all NeMo models (e.g. Experiment Manager and PyTorch Lightning trainer
 parameters), see the :doc:`../core/core` section.
 
-The model section of the NeMo ASR configuration files generally requires information about the dataset(s) being used, the preprocessor 
-for audio files, parameters for any augmentation being performed, as well as the model architecture specification. The sections on 
+The model section of the NeMo ASR configuration files generally requires information about the dataset(s) being used, the preprocessor
+for audio files, parameters for any augmentation being performed, as well as the model architecture specification. The sections on
 this page cover each of these in more detail.
 
 Example configuration files for all of the NeMo ASR scripts can be found in the
@@ -17,8 +17,8 @@ Dataset Configuration
 ---------------------
 
 Training, validation, and test parameters are specified using the ``train_ds``, ``validation_ds``, and
-``test_ds`` sections in the configuration file, respectively. Depending on the task, there may be arguments specifying the sample rate 
-of the audio files, the vocabulary of the dataset (for character prediction), whether or not to shuffle the dataset, and so on. You may 
+``test_ds`` sections in the configuration file, respectively. Depending on the task, there may be arguments specifying the sample rate
+of the audio files, the vocabulary of the dataset (for character prediction), whether or not to shuffle the dataset, and so on. You may
 also decide to leave fields such as the ``manifest_filepath`` blank, to be specified via the command-line at runtime.
 
 Any initialization parameter that is accepted for the Dataset class used in the experiment can be set in the config file.
@@ -80,7 +80,7 @@ Preprocessor Configuration
 --------------------------
 
 If you are loading audio files for your experiment, you will likely want to use a preprocessor to convert from the
-raw audio signal to features (e.g. mel-spectrogram or MFCC). The ``preprocessor`` section of the config specifies the audio 
+raw audio signal to features (e.g. mel-spectrogram or MFCC). The ``preprocessor`` section of the config specifies the audio
 preprocessor to be used via the ``_target_`` field, as well as any initialization parameters for that preprocessor.
 
 An example of specifying a preprocessor is as follows:
@@ -97,7 +97,7 @@ An example of specifying a preprocessor is as follows:
       ...
       # Other parameters for the preprocessor
 
-Refer to the `Audio Preprocessors <./api.html#Audio Preprocessors>`__ API section for the preprocessor options, expected arguments, 
+Refer to the `Audio Preprocessors <./api.html#Audio Preprocessors>`__ API section for the preprocessor options, expected arguments,
 and defaults.
 
 Augmentation Configurations
@@ -179,7 +179,7 @@ The following example sets up a ``SentencePiece Tokenizer`` at a path specified 
       dir: "<path to the directory that contains the custom tokenizer files>"
       type: "bpe"  # can be "bpe" or "wpe"
 
-The Aggregate (``agg``) tokenizer feature makes it possible to combine tokenizers in order to train multilingual 
+The Aggregate (``agg``) tokenizer feature makes it possible to combine tokenizers in order to train multilingual
 models. The config file would look like this:
 
 .. code-block:: yaml
@@ -188,21 +188,21 @@ models. The config file would look like this:
     ...
     tokenizer:
       type: "agg"  # aggregate tokenizer
-      langs: 
+      langs:
         en:
           dir: "<path to the directory that contains the tokenizer files>"
           type: "bpe"  # can be "bpe" or "wpe"
         es:
           dir: "<path to the directory that contains the tokenizer files>"
-          type: "bpe"  # can be "bpe" or "wpe"  
+          type: "bpe"  # can be "bpe" or "wpe"
 
-In the above config file, each language is associated with its own pre-trained tokenizer, which gets assigned 
-a token id range in the order the tokenizers are listed. To train a multilingual model, one needs to populate the 
+In the above config file, each language is associated with its own pre-trained tokenizer, which gets assigned
+a token id range in the order the tokenizers are listed. To train a multilingual model, one needs to populate the
 ``lang`` field in the manifest file, allowing the routing of each sample to the correct tokenizer. At inference time,
 the routing is done based on the inferred token id range.
 
-For models which utilize sub-word tokenization, we share the decoder module (``ConvASRDecoder``) with character tokenization models. 
-All parameters are shared, but for models which utilize sub-word encoding, there are minor differences when setting up the config. For 
+For models which utilize sub-word tokenization, we share the decoder module (``ConvASRDecoder``) with character tokenization models.
+All parameters are shared, but for models which utilize sub-word encoding, there are minor differences when setting up the config. For
 such models, the tokenizer is utilized to fill in the missing information when the model is constructed automatically.
 
 For example, a decoder config corresponding to a sub-word tokenization model should look similar to the following:
@@ -221,7 +221,7 @@ For example, a decoder config corresponding to a sub-word tokenization model sho
 Model Architecture Configurations
 ---------------------------------
 
-Each configuration file should describe the model architecture being used for the experiment. Models in the NeMo ASR collection need 
+Each configuration file should describe the model architecture being used for the experiment. Models in the NeMo ASR collection need
 an ``encoder`` section and a ``decoder`` section, with the ``_target_`` field specifying the module to use for each.
 
 Here is the list of the parameters in the model section which are shared among most of the ASR models:
@@ -478,7 +478,7 @@ A Citrinet-512 config should look similar to the following:
         se: ${model.model_defaults.se}
         se_context_size: ${model.model_defaults.se_context_size}
 
-As mentioned above, Citrinet uses the ``ConvASRDecoder`` as the decoder layer similar to QuartzNet. Only the configuration must be 
+As mentioned above, Citrinet uses the ``ConvASRDecoder`` as the decoder layer similar to QuartzNet. Only the configuration must be
 changed slightly as Citrinet utilizes sub-word tokenization.
 
 .. note::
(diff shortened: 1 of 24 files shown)


nemo/collections/asr/models/ctc_models.py
```diff
@@ -524,6 +536,23 @@ def forward(
         encoded_len = encoder_output[1]
         log_probs = self.decoder(encoder_output=encoded)
         greedy_predictions = log_probs.argmax(dim=-1, keepdim=False)
+        # generating decoding results for intermediate layers if necessary
+        if self.intermediate_loss_weights:
+            # we assume that encoder has to have property called "captured_layer_outputs"
+            # which is a list with the same length as loss weights
+            if len(self.encoder.captured_layer_outputs) != len(self.intermediate_loss_weights):
```
titu1994, line 543: Dont assume. First and foremost you should not be attaching tensors with grads to a module - ever. You can instead use the Tensor Registry framework in NeMo. Most models already support it due to other parts needing it. So do the following - In the training_step, enable the tensor registry, let modules register their forward activation (or a subset of it) and then access the registry here. Empty the registry at end of the train step to prevent memory leak.

So do a hasattr check, then raise a proper error (Value/Runtime) which details 
1) Provided CTC intermediate loss weights list is not empty
2) But the model does not add anything to the registry.

Also, this will waste memory during inference if the user does not need interctc outputs unless explicitly requested. So put a self.training check above for this part.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -478,6 +523,17 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
                 cache_last_channel_next=cache_last_channel_next,
                 cache_last_time_next=cache_last_time_next,
             )
+            if self.training:
+                should_drop = torch.rand(1) < drop_prob
+                # adjusting to match expectation
+                if should_drop:
+                    # that's not efficient, but it's hard to implement distributed
+                    # version of dropping layers without deadlock or random seed meddling
+                    # so multiplying the signal by 0 to ensure all weights get gradients
```
titu1994, line 532: This is fine.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -487,6 +543,9 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
                 _, pos_emb = self.pos_enc(x=audio_signal, cache_len=cache_len)
                 pad_mask, att_mask = self._create_masks(max_audio_length, length, audio_signal.device)
 
+            if lth in self.capture_output_at_layers:
```
titu1994, line 546: Use tensor registry here.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -496,6 +555,9 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
 
         audio_signal = torch.transpose(audio_signal, 1, 2)
 
+        for captured_output in self.captured_layer_outputs:
```
titu1994, line 558: do above when you use the tensor registry.

titu1994 (changes requested):
The idea is good, but the code is becoming complicated.
We have two options - 
1) Keep it inside ctc_model.py - but have mixin class that deals with interctc parts of the code including any and all functions needed by it. Then the ctc_model.py class simply calls these functions.
2) Write entirely separate class for ctc_models.py - subclass it and override the parts we need - which is primarily the forward, parts of training step and validation_step.

A bad option is to merge it as is right now - would significantly complicate the ctc_models.py training step and validation step plus cause issues with long audio inference where we don't want interctc tensor to waste memory.

@VahidooX What is your preference ?

nemo/collections/asr/models/ctc_models.py
```diff
@@ -106,6 +106,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         # Adapter modules setup (from ASRAdapterModelMixin)
         self.setup_adapters()
 
+        # setting up interCTC loss (from InterCTCMixin)
```
titu1994, line 109: Probably best to keep adapter setup for last, move it up by one level. Btw in a later PR we should update the execution flow diagram for CTC models - https://github.com/NVIDIA/NeMo/tree/main/examples/asr/asr_ctc#model-execution-overview

nemo/collections/asr/models/ctc_models.py
```diff
@@ -536,6 +540,9 @@ def training_step(self, batch, batch_nb):
         if AccessMixin.is_access_enabled():
             AccessMixin.reset_registry(self)
 
+        if self.interctc_enabled:
```
titu1994, line 543: Should be a function in InterCTCMixin, not a variable assigned to self.

nemo/collections/asr/models/ctc_models.py
```diff
@@ -544,27 +551,32 @@ def training_step(self, batch, batch_nb):
...
+        else:
+            log_every_n_steps = 1
+
         loss_value = self.loss(
             log_probs=log_probs, targets=transcript, input_lengths=encoded_len, target_lengths=transcript_len
         )
 
         # Add auxiliary losses, if registered
         loss_value = self.add_auxiliary_losses(loss_value)
+        loss_value, tensorboard_logs = self.add_interctc_losses(
```
titu1994, line 565: Very cool. Can you add a comment why we have the odd bool check in the `compute_wer` part.

nemo/collections/asr/models/ctc_models.py
```diff
@@ -607,29 +622,38 @@ def validation_step(self, batch, batch_idx, dataloader_idx=0):
...
-            'val_wer_denom': wer_denom,
-            'val_wer': wer,
-        }
+        # Reset access registry
+        if AccessMixin.is_access_enabled():
+            AccessMixin.reset_registry(self)
+
+        return metrics
+
+    def multi_validation_epoch_end(self, outputs, dataloader_idx: int = 0):
```
titu1994, line 644: Looks very clean, thanks !

nemo/collections/asr/models/ctc_models.py
```diff
@@ -607,29 +622,38 @@ def validation_step(self, batch, batch_idx, dataloader_idx=0):
...
 
     def test_step(self, batch, batch_idx, dataloader_idx=0):
         logs = self.validation_step(batch, batch_idx, dataloader_idx=dataloader_idx)
-        test_logs = {
-            'test_loss': logs['val_loss'],
-            'test_wer_num': logs['val_wer_num'],
-            'test_wer_denom': logs['val_wer_denom'],
-            'test_wer': logs['val_wer'],
-        }
+        test_logs = {name.replace("val_", "test_"): value for name, value in logs.items()}
```
titu1994, line 656: Wow, i forgot str.replace() exists. Good catch

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -192,6 +210,10 @@ def __init__(
         dropout_pre_encoder=0.1,
         dropout_emb=0.1,
         dropout_att=0.0,
+        stochastic_depth_drop_prob: float = 0.0,
+        stochastic_depth_mode: str = "linear",
+        stochastic_depth_start_layer: int = 0,
+        capture_output_at_layers: Optional[List[int]] = None,
```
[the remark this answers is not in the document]
titu1994: This arg should not be needed anymore right? You can specify in the interctc config what layers you need? Or some other way?
Im trying to see how to make this capture_output.. agnostic to model.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,6 +384,27 @@ def __init__(
...
+            # we are dividing by L - 1 to ensure we start from 0 probability
+            # (never drop the first layer) and end with desired drop probability.
+            layer_drop_probs += [l / (L - 1) * stochastic_depth_drop_prob for l in range(L)]
+        elif stochastic_depth_mode == "uniform":
+            layer_drop_probs += [stochastic_depth_drop_prob] * L
+        else:
+            raise ValueError('stochastic_depth_mode has to be one of ["linear", "uniform"].')
+        self.layer_drop_probs = layer_drop_probs
+        self.capture_output_at_layers = capture_output_at_layers
+        if self.capture_output_at_layers is None:
```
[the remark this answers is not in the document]
titu1994: ++

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -487,6 +541,10 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
                 _, pos_emb = self.pos_enc(x=audio_signal, cache_len=cache_len)
                 pad_mask, att_mask = self._create_masks(max_audio_length, length, audio_signal.device)
 
+            if lth in self.capture_output_at_layers and self.is_access_enabled():
```
titu1994, line 544: Can you add a comment with the registered tensor shapes as [B, T, D] or [B, D, T]?

titu1994: Plus, two things. One - the layer id part should be part of InterCTC Mixin's config management, not model specific. 
Two - after self.is_access_enabled() add a bool check to InterCTCMixin if the mixin is enabled or not. setup_interctc() should enable the flag depending on user's config setting.

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+    we want to get loss from. Then call
+
+        * ``setup_interctc`` in the init method
+        * ``self.add_interctc_losses`` after computing regular loss.
+        * ``self.finalize_interctc_metrics(metrics, outputs, prefix="val_")``
+          in the `multi_validation_epoch_end` method.
+        * ``self.finalize_interctc_metrics(metrics, outputs, prefix="test_")``
+          in the `multi_test_epoch_end` method.
+    """
+
```
titu1994, line 35: Note: Add method that returns whether interctc mixin is enabled or not rather than assign value to self.

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+                "loss will have weight of 0.6"
+            )
+        self.interctc_enabled = len(self.intermediate_loss_weights) > 0
+        if self.interctc_enabled and not hasattr(self.encoder, 'capture_output_at_layers'):
+            raise ValueError('To use intermediate CTC loss, encoder has to define "capture_output_at_layers" property')
+
+        if len(self.encoder.capture_output_at_layers) != len(self.intermediate_loss_weights):
+            raise ValueError('Length of encoder.capture_output_at_layers has to match intermediate_loss_weights')
+
+    def finalize_interctc_metrics(self, metrics, outputs, prefix):
```
titu1994, line 54: Methods need docstrings

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+                    {
+                        f"{prefix}inter_ctc_loss_l{layer_idx}": loss,
+                        f"{prefix}inter_wer_l{layer_idx}": wer_num / wer_denom,
+                    }
+                )
+            metrics["log"][f"{prefix}final_ctc_loss"] = torch.stack(
+                [x[f"{prefix}final_ctc_loss"] for x in outputs]
+            ).mean()
+
+    def get_captured_tensors(self):
```
titu1994, line 70: Lets have the keyword `interctc` somewhere in the name of all methods to avoid name collision. Or make them private if wont be used outside of this class.

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+        if not self.interctc_enabled:
+            return []
+        # if intermediate_loss_weights was set, the encoder has to register
+        # layer_output_X and layer_length_X tensors. We need to apply decoder
+        # to each of them and compute CTC loss.
+        module_registry = AccessMixin.get_module_registry(self.encoder)['']  # key for encoder
```
titu1994, line 76: [''] ?

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def add_interctc_losses(
+        self,
+        loss_value: torch.tensor,
+        transcript: torch.tensor,
+        transcript_len: torch.tensor,
+        compute_wer: bool,
+        log_wer_num_denom: bool = False,
+        log_prefix: str = "",
+    ) -> Tuple[torch.Tensor, Dict]:
```
titu1994, line 103: Docstring

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,140 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict, Tuple
+
+import torch
+
+from nemo.core.classes.mixins import AccessMixin
+
+
+class InterCTCMixin:
```
titu1994, line 21: Overall, this looks fantastic. Thanks !

titu1994 (commented):
Looks a lot better, minor comments and docstring additions to InterCTCMixin class, then can be merged.

nemo/collections/asr/parts/mixins/interctc_mixin.py
Kipok: That's the key that AccessMixin has when assigning tensors to the current module. I don't have control over it :)

docs/source/asr/configs.rst
```diff
@@ -546,6 +547,38 @@ The encoder section includes the details about the RNN-based encoder architectur
 config files and also :ref:`nemo.collections.asr.modules.RNNEncoder <rnn-encoder-api>`.
 
 
+CTC Configurations
```
Kipok, line 550: @titu1994 not sure if that's the best place to put these docs

examples/asr/conf/conformer/conformer_ctc_bpe.yaml
```diff
@@ -143,6 +143,15 @@ model:
     num_classes: -1
     vocabulary: []
 
+  # config for InterCTC loss: https://arxiv.org/abs/2102.03216
```
VahidooX, line 146: Are we going to support stochastic depth for Transducer models?

examples/asr/conf/conformer/conformer_ctc_bpe.yaml
```diff
@@ -143,6 +143,15 @@ model:
...
+  # specify loss weights and which layers to use for InterCTC
+  # e.g., to reproduce the paper results, set loss_weights: [0.3]
+  # and apply_at_layers: [9] (assuming 18 layers). Note that final
+  # layer loss coefficient is automatically adjusted (to 0.7 in above example)
```
VahidooX, line 150: Would you please add this section to the fastconformer and cache-aware configs too?

docs/source/asr/configs.rst
VahidooX: It is for the configuration of ASR models, doesn't look like that bad to me.

docs/source/asr/configs.rst
```diff
@@ -546,6 +547,38 @@ The encoder section includes the details about the RNN-based encoder architectur
...
+------------------
+
+All CTC-based models also support `InterCTC loss <https://arxiv.org/abs/2102.03216>`_. To use it, you need to specify
```
VahidooX, line 553: How about the stochastic depth docs?

VahidooX (changes requested):
LGTM, just minor comments!

examples/asr/conf/conformer/conformer_ctc_bpe.yaml
Kipok: It's supported, but only for ConformerEncoder. The other encoders are possible, but need to check the details, so let's do it in another PR

docs/source/asr/configs.rst
Kipok: What's the best place to put them in? They are currently in the code for ConformerEncoder and only supported there (but for any model that's using it)

examples/asr/conf/conformer/conformer_ctc_bpe.yaml
VahidooX: Why not adding stochastic_depth_start_layer and their params to the encoder section?
You have just added the InterCTC section, right?
Is it OK to just support Conformer. I meant for Conformer-Transducer configs.

docs/source/asr/configs.rst
VahidooX: The same place looks good to me. You may mention in the descriptions that it is just supported for Conformer-based models.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -362,6 +384,25 @@ def __init__(
         self.setup_streaming_params()
         self.export_cache_support = False
 
+        if not (0 <= stochastic_depth_drop_prob < 1.0):
```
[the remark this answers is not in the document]
Kipok: sure, let me move it to a utility (it does not really depend on any class parameters) so that it can also be re-used in other models later

examples/asr/conf/conformer/conformer_ctc_bpe.yaml
Kipok: added to the example configs (if that's what you meant)

docs/source/asr/configs.rst
Kipok: But this one is called "CTC Configurations", while stochastic depth is for both CTC and transducer (although only for conformer-based versions now)

VahidooX: I meant creating a new section in the same file titled "Stochastic Depth" or something like this.
You may also rename this section to "InterCTC Loss" instead of "CTC Configurations"?

nemo/collections/asr/modules/conformer_encoder.py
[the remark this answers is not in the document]
VahidooX: Even if put that method inside the Conformer encoder it is still OK. Other option is to put it in higher levels. Whatever you see fit.

docs/source/asr/configs.rst
titu1994: Same, looks fine

docs/source/asr/configs.rst
titu1994: Agreed. Rename to InterCTC Config and add new config Stochastic Depth. Note that conformer is only supported now. We can add more models in the future

docs/source/asr/configs.rst
```diff
@@ -546,6 +547,36 @@ The encoder section includes the details about the RNN-based encoder architectur
...
+2 parameters as in example below
+
+.. code-block:: yaml
+
+   model:
+      # ...
+      interctc:
+        loss_weights: [0.3]
```
titu1994, line 561: Note why list of weights is used

docs/source/asr/configs.rst
```diff
@@ -546,6 +547,36 @@ The encoder section includes the details about the RNN-based encoder architectur
...
+.. code-block:: yaml
+
+   model:
+      # ...
+      interctc:
+        loss_weights: [0.1, 0.3]
+        apply_at_layers: [3, 8]
+
+Note that the final-layer CTC loss weight is automatically computed to normalize
+all weight to 1 (0.6 in the example about).
```
titu1994, line 577:
```suggestion
all weight to 1 (0.6 in the example above).
```

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -487,6 +522,20 @@ def forward(self, audio_signal, length, cache_last_channel=None, cache_last_time
                 _, pos_emb = self.pos_enc(x=audio_signal, cache_len=cache_len)
                 pad_mask, att_mask = self._create_masks(max_audio_length, length, audio_signal.device)
 
+            # saving tensors if required for interctc loss
+            if self.is_access_enabled():
```
titu1994, line 526: If self.is_access_enabled() and self.is_interctc_enabled():

nemo/collections/asr/modules/squeezeformer_encoder.py
```diff
@@ -360,6 +363,20 @@ def forward_for_export(self, audio_signal, length):
 
             audio_signal = layer(x=audio_signal, att_mask=att_mask, pos_emb=pos_emb, pad_mask=pad_mask)
 
+            # saving tensors if required for interctc loss
+            if self.is_access_enabled():
+                if self.interctc_capture_at_layers is None:
```
titu1994, line 368: If self.is_access_enabled() and self.is_interctc_enabled():

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,206 @@
...
+    Then call
+
+        * ``setup_interctc`` in the init method
+        * ``self.add_interctc_losses`` after computing regular loss.
+        * ``self.finalize_interctc_metrics(metrics, outputs, prefix="val_")``
+          in the `multi_validation_epoch_end` method.
+        * ``self.finalize_interctc_metrics(metrics, outputs, prefix="test_")``
+          in the `multi_test_epoch_end` method.
+    """
+
```
titu1994, line 43: Add getter and setter for a bool variable - is_interctc_enabled() (already added) and set_interctc_enabled(enabled: bool). Instance methods, they check if hasattr() self._interctc_enabled and returns that or sets it explicitly. 

It will allow users to disable it even if present in config which gives more flexibility (say finetuning or want to experiment). Plus it gives better access to whether the class should allow interctc of not even if access is enabled for other use cases (adapter losses, SSL etc)

nemo/collections/asr/parts/mixins/interctc_mixin.py
```diff
@@ -0,0 +1,206 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict, List, Tuple
+
+import torch
+
+from nemo.core.classes.mixins import AccessMixin
+
+
+class InterCTCMixin:
```
titu1994, line 21: Add this mixin to the API docs of ASR mixin class docs.

nemo/collections/asr/parts/submodules/jasper.py
```diff
@@ -1041,8 +1046,20 @@ def forward(self, input_: Tuple[List[Tensor], Optional[Tensor]]) -> Tuple[List[T
 
                 out = out.transpose(1, 2)  # (B, C, T)
 
-        if self.is_access_enabled() and self.access_cfg.get('save_encoder_tensors', False):
-            self.register_accessible_tensor(name='encoder', tensor=out)
+        if self.is_access_enabled():
+            # for adapters
```
titu1994, line 1050: ... + self.is_interctc_enabled():

titu1994 (commented):
Minor comments. The PR is ready to merge after that

docs/source/asr/configs.rst
Kipok: There is an example right after this one, which explains why list is used by providing multiple values

nemo/collections/asr/modules/conformer_encoder.py
Kipok: So, this function is defined on the EncDec model level, not in the encoder. Here we can only check that by querying access mixin and checking if it needs to log something. Do you propose to add a new is_interctc_enabled function to all encoders and put that mixin access check there?

docs/source/asr/configs.rst
titu1994: Ah good enough. Thanks.

nemo/collections/asr/modules/conformer_encoder.py
titu1994: Hmm. The glob access mixin config can be used to see whether interctc is enabled of disabled (we can set a config value when you call set_interctc_available() in the mixin. 

So basically if is_access_enabled(): if self.access_cfg.get("is_interctc_enabled", False): ...

titu1994 (dismissed):
After review, seems I was mistaken. All cases are covered correctly. 
LGTM, ready to merge

titu1994 (approved):

VahidooX (approved):

PR #6023, opened 2023-02-14
Update TitaNet-L and MSDD models 

nithinraok:
# What does this PR do ?

* Update hydra config on published NGC titanet-model and release as v1 version
* Improve speaker label tests with latest hydra config
* Add a feature to write model labels `labels.txt` to exp directory 
* Update batch inference function to return gt labels and trained model labels (if available) 

**Collection**: ASR

# Changelog 
- Update label models to include latest v1 version of titanet-large
- Added titanet-small config file (model will be published soon)
- Now instead of returning index of gt labels from AudioSpeechLabelDataset, return the speaker label names directly as seen from manifest file

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [x] Documentation

### DIFF
--- a/nemo/collections/asr/models/label_models.py
+++ b/nemo/collections/asr/models/label_models.py
@@ -98,6 +98,7 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
         self.world_size = 1
         self.cal_labels_occurrence_train = False
         self.labels_occurrence = None
+        self.labels = None
 
         num_classes = cfg.decoder.num_classes
 
@@ -152,7 +153,6 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
 
         self._macro_accuracy = Accuracy(num_classes=num_classes, top_k=1, average='macro', task='multiclass')
 
-        self.labels = None
         if hasattr(self._cfg, 'spec_augment') and self._cfg.spec_augment is not None:
             self.spec_augmentation = EncDecSpeakerLabelModel.from_config_dict(self._cfg.spec_augment)
         else:
@@ -390,7 +390,7 @@ class EncDecSpeakerLabelModel(ModelPT, ExportableEncDecModel):
 
         self.log(f'{tag}_loss', loss_mean, sync_dist=True)
         for top_k, score in zip(self._accuracy.top_k, topk_scores):
-            self.log(f'{tag}_acc_micro_top@{top_k}', score, sync_dist=True)
+            self.log(f'{tag}_acc_micro_top_{top_k}', score, sync_dist=True)
         self.log(f'{tag}_acc_macro', macro_accuracy_score, sync_dist=True)
 
         return {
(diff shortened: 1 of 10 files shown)


nithinraok:
test `test_pretrained_titanet_embeddings` would be passed after v1 version of model gets published to NGC [ON GOING]

nemo/collections/asr/models/label_models.py
```diff
@@ -124,21 +122,6 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
             weight = [sum(self.labels_occurrence) / (len(self.labels_occurrence) * i) for i in self.labels_occurrence]
 
         if 'loss' in cfg:
-            # To support older version checkpoints
```
fayejf, line 127: how about older checkpoints such as speakernet?

nemo/collections/asr/models/label_models.py
```diff
@@ -546,21 +528,20 @@ def batch_inference(self, manifest_filepath, batch_size=32, sample_rate=16000, d
             embs: embeddings of files provided in manifest file
             logits: logits of final layer of EncDecSpeakerLabel Model
             gt_labels: labels from manifest file (needed for speaker enrollment and testing)
-            mapped_labels: Classification labels sorted in the order that they are mapped by the trained model
+            trained_labels: Classification labels sorted in the order that they are mapped by the trained model
```
fayejf, line 531: how about the groundtruth label does not belong to trained_label?  What whould be gt_labels in that case? 
If gt_label does not present and the model hasn't seen the class the audio, it will output most likely pred?

nemo/collections/asr/models/label_models.py
nithinraok: Latest versions are being published on NGC. Urls would be updated accordingly

nemo/collections/asr/models/label_models.py
nithinraok: gt_lables doesn;t depend on trained labels now, they just return what it gets from manifest.

nemo/collections/asr/models/label_models.py
fayejf: Then let's merge and update the link until the latest version is published? Or it will break things.

nemo/collections/asr/models/label_models.py
fayejf: then could you add a test to ensure these conditions, please?

nemo/collections/asr/models/label_models.py
nithinraok: Added links for speakernet and ecapa models as well. All models are updated now

nemo/collections/asr/models/label_models.py
fayejf: No, we didn't have test for groundtruth label does not belong to trained_label. Could you add this?

PR #6108, opened 2023-02-24
Add missing parameters to ModelCheckpoint CallbackParams

markelsanz14:
# What does this PR do ?
Adds missing parameters to ModelCheckpoint CallbackParams, to achieve feature parity with PTL. See ModelCheckpoint API here: https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.ModelCheckpoint.html

**Collection**: exp_manager

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -99,7 +99,7 @@ class CallbackParams:
     auto_insert_metric_name: bool = True
     every_n_epochs: Optional[int] = 1
     every_n_train_steps: Optional[int] = None
-    train_time_interval: Optional[timedelta] = None
+    train_time_interval: Optional[str] = None
     prefix: Optional[str] = None  # If None, exp_manager will attempt to handle the filepath
     postfix: str = ".nemo"
     save_best_model: bool = False
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -96,7 +96,10 @@ class CallbackParams:
     save_top_k: Optional[int] = 3
     save_weights_only: Optional[bool] = False
     mode: Optional[str] = "min"
+    auto_insert_metric_name: bool = True
     every_n_epochs: Optional[int] = 1
+    every_n_train_steps: Optional[int] = None
+    train_time_interval: Optional[timedelta] = None
     prefix: Optional[str] = None  # If None, exp_manager will attempt to handle the filepath
     postfix: str = ".nemo"
     save_best_model: bool = False


nemo/utils/exp_manager.py
```diff
@@ -96,7 +96,10 @@ class CallbackParams:
     save_top_k: Optional[int] = 3
     save_weights_only: Optional[bool] = False
     mode: Optional[str] = "min"
+    auto_insert_metric_name: bool = True
```
MaximumEntropy, line 99: What does this bit do? And how does it interact with specifying a filename that already has metrics specified like this in `megatron_gpt_config.yaml`. 

Ex: `filename: 'megatron_gpt--{val_loss:.2f}-{step}-{consumed_samples}`

markelsanz14: When this is true, it will not break anything and behaves the same as until now. It's already true by default on PTL. When it's true, PTL will replace {val_loss:.2f} with something like val_loss=0.12-.
The equal (=) sign and the dash (-) signs are what PTL auto-inserts to the metric name. So it won't change anything.

However, when set to False, you can specify the name of the checkpoint manually, to replace some of those characters with any character you want. This can help avoid logging issues with W&B, TB, or other systems that don't like the = and - characters, so it gives you full control over the checkpoint name.

nithinraok:
LGTM. `every_n_train_steps` is crucial as we move our training strategies to steps based rather than epoch based

nithinraok (dismissed):

nemo/utils/exp_manager.py
```diff
@@ -96,7 +96,10 @@ class CallbackParams:
...
     every_n_epochs: Optional[int] = 1
+    every_n_train_steps: Optional[int] = None
+    train_time_interval: Optional[timedelta] = None
```
titu1994, line 102: Looks good. Maybe we should look into some way of testing the callbacks data class (if it exists in PTL) against exp managers checkpoint subconfig similar to how we check PTL trainer args against our trainer config.

titu1994 (approved):

nemo/utils/exp_manager.py
markelsanz14: It doesn't exist as far as I know, I think it's just the parameters of the ModelCheckpoint init function. But I'm not 100% sure, I haven't looked at it in depth.

markelsanz14:
It seems like timedelta cannot be used with OmegaConf as a type annotation, but timedelta is what PTL uses (see link in PR description). Should I remove the type annotation? Or can I replace timedelta with something else?

markelsanz14:
I modified the train_time_interval type annotation from Optional[timedelta] to Optional[str] after a discussion with @MaximumEntropy . Otherwise OmegaConf complains that timedelta is not a supported type annotation class. Please review again @nithinraok @titu1994

nithinraok (approved):

PR #6155, opened 2023-03-09
[TTS] Make mel spectrogram norm configurable

rlangman:
# What does this PR do ?

Add optional configuration to AudioToMelSpectrogramPreprocessor to allow disabling area normalization in the librosa mel filter.

Applying area normalization to the mel spectrogram reduces its range of values from approximately [0, 100] to [0, 1], which is detrimental when you take the log of the [0, 1] values producing a poor distribution of negative features.

The intention is to experiment with disabling the area normalization and replacing the current mel spectrogram used in TTS which is:

> log(max(1E-5, area_norm(mel))

With a more accurate representation like:

> log(1.0 + mel)

This will also enable us to compute energy features using the mel spectrogram instead of the linear spectrogram.

**Collection**: [TTS], [ASR]

# Changelog 
- Add an optional parameter to AudioToMelSpectrogramPreprocessor to configure the mel filter normalization, defaulting to the librosa default of 'slaney'

# Usage
.yaml config:

> preprocessor:
>     _target_: nemo.collections.asr.modules.AudioToMelSpectrogramPreprocessor
>     log: true
>     log_zero_guard_type: add
>     log_zero_guard_value: 1.0
>     mag_power: 1.0
>     mel_norm: null

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/modules/audio_preprocessing.py
+++ b/nemo/collections/asr/modules/audio_preprocessing.py
@@ -926,9 +926,10 @@ class AudioToMelSpectrogramPreprocessorConfig:
     nb_augmentation_prob: float = 0.0
     nb_max_freq: int = 4000
     use_torchaudio: bool = False
+    mel_norm: str = "slaney"
     stft_exact_pad: bool = False  # Deprecated argument, kept for compatibility with older checkpoints.
     stft_conv: bool = False  # Deprecated argument, kept for compatibility with older checkpoints.
-    mel_norm: str = 'slaney'
+
 
 
 @dataclass
--- a/nemo/collections/asr/modules/audio_preprocessing.py
+++ b/nemo/collections/asr/modules/audio_preprocessing.py
@@ -162,10 +162,10 @@ class AudioToMelSpectrogramPreprocessor(AudioPreprocessor, Exportable):
             nb_max_freq (int) : Frequency above which all frequencies will be masked for narrowband augmentation.
                 Defaults to 4000
             use_torchaudio: Whether to use the `torchaudio` implementation.
+            mel_norm: Normalization used for mel filterbank weights.
+                Defaults to 'slaney' (area normalization)
             stft_exact_pad: Deprecated argument, kept for compatibility with older checkpoints.
             stft_conv: Deprecated argument, kept for compatibility with older checkpoints.
-            mel_norm: Normalization used for filterbanks in librosa.filters.mel.
-                Defaults to 'slaney' (area normalization)
         """
 
     def save_to(self, save_path: str):
@@ -229,9 +229,9 @@ class AudioToMelSpectrogramPreprocessor(AudioPreprocessor, Exportable):
         nb_augmentation_prob=0.0,
         nb_max_freq=4000,
         use_torchaudio: bool = False,
+        mel_norm="slaney",
         stft_exact_pad=False,  # Deprecated arguments; kept for config compatibility
         stft_conv=False,  # Deprecated arguments; kept for config compatibility
-        mel_norm='slaney',
     ):
         super().__init__(n_window_size, n_window_stride)
 
@@ -275,9 +275,9 @@ class AudioToMelSpectrogramPreprocessor(AudioPreprocessor, Exportable):
             rng=rng,
             nb_augmentation_prob=nb_augmentation_prob,
             nb_max_freq=nb_max_freq,
+            mel_norm=mel_norm,
             stft_exact_pad=stft_exact_pad,  # Deprecated arguments; kept for config compatibility
             stft_conv=stft_conv,  # Deprecated arguments; kept for config compatibility
-            mel_norm=mel_norm,
         )
 
     def input_example(self, max_batch: int = 8, max_dim: int = 32000, min_length: int = 200):
--- a/nemo/collections/asr/modules/audio_preprocessing.py
+++ b/nemo/collections/asr/modules/audio_preprocessing.py
@@ -931,7 +931,6 @@ class AudioToMelSpectrogramPreprocessorConfig:
     stft_conv: bool = False  # Deprecated argument, kept for compatibility with older checkpoints.
 
 
-
 @dataclass
 class AudioToMFCCPreprocessorConfig:
     _target_: str = 'nemo.collections.asr.modules.AudioToMFCCPreprocessor'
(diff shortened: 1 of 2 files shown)


racoiaws:
> Applying area normalization to the mel spectrogram reduces its range of values from approximately [0, 100] to [0, 1], which is detrimental when you take the log of the [0, 1] values producing a poor distribution of negative features.

I think there is a misunderstanding.
Let's say we have a white noise signal with uniform energy across the spectrum. With the standard "slaney" normalisation, the mel-spectrogram is supposed to also have uniform values in all bands. Reducing the range of values is not the goal.

Without normalisation we should get up big numbers in bands where the mel-scale 'compresses' (higher frequencies). Think sum instead of averaging. I think this is what you interpreted as 'extended' range.

This could be clarified by visualising distributions of values for different mel bands. Could you please do that? I would expect both higher mean and variance for upper bands.

racoiaws:
Note that the above is orthogonal to `log(1+mel)` suggestion. Wonder why people don't do that. `log(mel)` is almost decibels so makes sense in theory, but `log(1+mel)` would make much more sense in practice

rlangman:
> > Applying area normalization to the mel spectrogram reduces its range of values from approximately [0, 100] to [0, 1], which is detrimental when you take the log of the [0, 1] values producing a poor distribution of negative features.
> 
> I think there is a misunderstanding. Let's say we have a white noise signal with uniform energy across the spectrum (STFT amplitude). With the standard "slaney" normalisation, the mel-spectrogram is supposed to also have uniform values in all bands. Reducing the range of values is not the goal.
> 
> Without normalisation we would get bigger numbers in bands where the mel-scale 'compresses' (higher frequencies). Think sum instead of averaging (roughly). I think this is what you interpreted as 'extended' range.
> 
> This could be clarified by visualising distributions of values for different mel bands. Could you please do that? I would expect both higher mean and variance for upper bands.

True, the point of slaney normalization is definitely not to reduce the range of values. But it is a side effect, which does not work well with log compression. It could be that using the linear mels after normalization is better than using log mels without normalization (I suspect both would produce similar results).

If people are interested, I could create visuals showing the distribution of features using the different preprocessing configurations.

racoiaws:
> True, the point of slaney normalization is definitely not to reduce the range of values. But it is a side effect, which does not work well with log compression

My main point is, this side effect most likely materialises in the upper bands, with lesser effect on bands where the mel-scale is more linear. When used as input to a neural network, this effectively puts a bigger weight on higher frequencies from the start.

For your goal, I would rather suggest just scaling input before the logarithm (`log(alpha * mel)`),  with alpha somewhere from 1 to 100 (default 1). This way before-log range scales the same for all bands.

rlangman:
Below is 1 audio sample with different mel configs. There are limits to what we can infer by just looking at 1 piece of audio, but at a high level we see:

1. The spectrograms which add 1.0 to the log guard look a lot cleaner (most values in the spectrogram are very close to 0).
2. Without log compression, there are a large number of small values and small number of very large values/outliers. Taking the log results in a more even distribution.
3. Area normalization reduces the scale of higher frequency bands relative to smaller frequency bands, due to these mel bands having a larger range.

From this I guess that a function like log(1.0 + 50 * slaney(mel)) would be optimal, but is likely to produce the same behavior as 
log(1.0 + mel). Without running experiments, I would be OK with either but would slightly favor the latter by default due to its simplicity.

Would be happy to hear anyone else's thoughts on this.

mel

mel

log(1.0 + mel)

mel_log

slaney(mel)

mel_slaney

log(max(1E-5, slaney(mel))) - Our current TTS config

mel_slaney_log_guard

log(max(1E-5, 50 * slaney(mel)))

mel_slaney_50x_log_guard

log(1.0 + slaney(mel))

mel_slaney_log

log(1.0 + 50 * slaney(mel))

mel_slaney_50x_log

nemo/collections/asr/modules/audio_preprocessing.py
```diff
@@ -274,6 +277,7 @@ def __init__(
             nb_max_freq=nb_max_freq,
             stft_exact_pad=stft_exact_pad,  # Deprecated arguments; kept for config compatibility
             stft_conv=stft_conv,  # Deprecated arguments; kept for config compatibility
+            mel_norm=mel_norm,
```
XuesongYang, line 280: `featurizer_class` has two options. You may need to change same in `FilterbankFeaturesTA`.

XuesongYang (changes requested):

XuesongYang:
A little bit off-topic, this PR only makes functions configurable while it is the user's responsibility to decide which trick is more applicable to their needs. The discussion is very valuable, shall we move to other places for further discussion>?

nemo/collections/asr/modules/audio_preprocessing.py
titu1994: Torchaudio should not be the default for anything. We don't install it and it's not a requirement, but an optional one

titu1994: You'll need to add this argument and basically just ignore it in the TA variant (with logging that you are ignoring it)

titu1994 (commented):
This is fine with me, since it's backward compatible. Please ask for final review when PR is ready

titu1994:
Side note - while this is a TTS change, you are modifying ASR code, so please make sure to notify someone from ASR for reviews (usually me or @VahidooX) before merging.

nemo/collections/asr/modules/audio_preprocessing.py
rlangman: I added the parameter to the TA variant. Interestingly, `torchaudio.transforms.MelSpectrogram` supports the slaney normalization but (similar to Tensorflow and Kaldi) defaults to not using it.

XuesongYang (approved):

XuesongYang:
added a resource that showing diff between `slaney` and `htk` mel-scale: https://groups.google.com/g/librosa/c/JDzUZggitYM

titu1994 (approved):
Looks good ! Thanks

titu1994:
@XuesongYang Yes it's a good idea to make a discussion on the topic with a summary of the discussion + plots. Better still if experiments show notable improvement, then simply updating configs should be useful

PR #6159, opened 2023-03-09
convert Megatron LM ckpt to NeMo PP support.

yidong72:
1. Fixed the bug in the script to convert Megatron LM ckpt to NeMo.
2. Fixed a bug in the GPT PP that we should only add output layer for the last PP rank.
3. Overwrite the precision in the eval script and fix the web server.


MaximumEntropy (approved):
LGTM!

PR #6170, opened 2023-03-10
Update dataset preparation tool

Zhilin123:
# What does this PR do ?

Refactor hyperparameter recommendation and add defaults for epochs, num virtual tokens and lr

**Collection**: nlp

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


tools/customization_dataset_preparation/customization_dataset_preparation.py
```diff
@@ -79,31 +79,50 @@ def load_file_into_df(filename):
...
-
-    return message
 
-
-def estimating_customization_job_time(df, recommend_hyperparameters_message):
-    recommended_batch_size = int(recommend_hyperparameters_message.split("=")[1].split()[0])
+    df_char_length = df.apply(lambda x: len(x.prompt) + len(x.completion), axis=1)
+    length_by_chars = sorted(list(df_char_length))
+    n_samples_under_99p5_limit = int(len(df_char_length) * 0.995 + 1)
+    max_char_length = length_by_chars[:n_samples_under_99p5_limit][-1]
```
arendu, line 111: `length_by_chars[:n_samples_under_99p5_limit][-1]` -> `length_by_chars[n_samples_under_99p5_limit]`
won't this suffice?

tools/customization_dataset_preparation/customization_dataset_preparation.py
```diff
@@ -79,31 +79,50 @@ def load_file_into_df(filename):
...
-        message += additional_msg_for_small_df
-
-    return message
 
-
-def estimating_customization_job_time(df, recommend_hyperparameters_message):
-    recommended_batch_size = int(recommend_hyperparameters_message.split("=")[1].split()[0])
+    df_char_length = df.apply(lambda x: len(x.prompt) + len(x.completion), axis=1)
+    length_by_chars = sorted(list(df_char_length))
+    n_samples_under_99p5_limit = int(len(df_char_length) * 0.995 + 1)
```
arendu, line 110: `int(len(df_char_length) * 0.995 + 1)` -> `int(math.ceil(len(df_char_length) * 0.995))`

tools/customization_dataset_preparation/customization_dataset_preparation.py
```diff
@@ -367,7 +398,7 @@ def print_all_messages(messages):
     messages = []
     messages.append(str(args))
 
-    MAX_TOTAL_CHAR_LENGTH = 10000
+    MODEL_MAX_SEQ_LENGTH = 2048
```
arendu, line 401: The Rope models can support 4096 tokens (and even beyond that)

Zhilin123: Will update when they get released!

arendu (dismissed):
LGTM!

Zhilin123:
@arendu I've reverted the choice to use max seq/token length over max char length to filter out samples given that it introduces a new dependency to use huggingface tokenizers, which might fail sometimes due to it having to make an internet call to huggingface (which has might rate limit us), following a discussion with @okuchaiev

tools/customization_dataset_preparation/customization_dataset_preparation.py
```diff
@@ -79,31 +80,50 @@ def load_file_into_df(filename):
...
 
-
-def estimating_customization_job_time(df, recommend_hyperparameters_message):
-    recommended_batch_size = int(recommend_hyperparameters_message.split("=")[1].split()[0])
+    df_char_length = df.apply(lambda x: len(x.prompt) + len(x.completion), axis=1)
+    length_by_chars = sorted(list(df_char_length))
+    n_samples_under_99p5_limit = math.ceil(len(df_char_length) * 0.995)
+    max_char_length = length_by_chars[n_samples_under_99p5_limit - 1]
+    # every token is around 4 chars + 100 for extra capacity
+    max_seq_length = max_char_length // 4 + 100
```
okuchaiev, line 114: What if the users data is very uniform in terms of seq length? E.g. the 99p5_limit is off by very small number from average? 
I would only use 99p5_limit if it differs significantly from the mean + 2*std. E.g. - we should be cutting outliers and if there is no outliers we shouldn't be cutting.

okuchaiev (changes requested):

tools/customization_dataset_preparation/customization_dataset_preparation.py
Zhilin123: Hmm I think this current logic might work because if say we have 1000 examples with range([1. 1000]) chars each. Then the max limit will be 995 chars. When calculating the max_seq_length, we include an additional 100 tokens (partly for this and partly for the fact that calculation based on chars/tokens is inexact) so we can approximately consider ~400 additional characters. What this means to say is in the general case, we would likely not cut off any samples unless it has >400 characters more than the 99p5 limit. Do you think this will be enough? 

If it's not, I can also calculate mean + 2*std and use that to find max_char_length but that code will be a bit hard to read (since we need to first do 

`max_char_length = max( min(mean+2*std , length_by_chars[-1]), length_by_chars[n_samples_under_99p5_limit - 1]) `

and then add some spare tokens to account imprecise char/token).

okuchaiev (approved):

PR #6172, opened 2023-03-10
Tts fixed vocab

redoctopus:
# What does this PR do ?

Allows users to pass in a grapheme + phoneme list/set as an enforced symbol vocabulary when using the IPA classes. Other symbols, such as punctuation, OOV, etc. are still handled by other tokenizer arguments.

Once set, G2P dictionary entries will be filtered in the following way:
- Words with any illegal graphemes are removed entirely.
- Words with one unique pronunciation and any illegal phonemes are removed entirely.
- If a word has multiple pronunciations and all of them have at least one illegal phoneme, the word is removed entirely.
- If at least one pronunciation is valid, it is kept if `keep_alternate=True` (default), otherwise the word is removed entirely.

The `TTSDataset` will also filter out all manifest entries whose normalized forms have any illegal graphemes.

Note: Users should take care that the passed-in vocab is consistent with `grapheme_prefix` and `grapheme_case`, or else filtering may not work properly.

**Collection**: TTS

# Changelog 
- Added a `replace_symbols()` function to `IPAG2P` to enforce user-set vocab and filter existing G2P dict
- Modified `IPATokenizer` to support taking in a user-set vocab
- Modified `TTSDataset` to check for illegal graphemes if a fixed vocab is used
- Added unit tests for `replace_symbols()`
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -527,6 +527,7 @@ class IPATokenizer(BaseTokenizer):
                 Set only if overriding the default vocab generation process (reading from G2P dict).
                 If set, any dataset entries that have unincluded graphemes will be filtered out, and any words whose
                 pronunciations have unincluded phonemes will be treated as OOV.
+                Please make sure that the grapheme prefixes and cases are consistent with the G2P module's settings.
                 Defaults to None, which means default vocab generation is used.
             space: Space token as string.
             silence: Silence token as string (will be disabled if it is None).
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -505,6 +505,7 @@ class IPATokenizer(BaseTokenizer):
         locale="en-US",
         punct=True,
         non_default_punct_list=None,
+        fixed_vocab=None,
         *,
         space=' ',
         silence=None,
@@ -522,6 +523,11 @@ class IPATokenizer(BaseTokenizer):
                 Specify None if implementing custom logic for a new locale.
             punct: Whether to reserve grapheme for basic punctuation or not.
             non_default_punct_list: List of punctuation marks which will be used instead default, if any.
+            fixed_vocab: List of valid grapheme/phoneme tokens for the model.
+                Set only if overriding the default vocab generation process (reading from G2P dict).
+                If set, any dataset entries that have unincluded graphemes will be filtered out, and any words whose
+                pronunciations have unincluded phonemes will be treated as OOV.
+                Defaults to None, which means default vocab generation is used.
             space: Space token as string.
             silence: Silence token as string (will be disabled if it is None).
             apostrophe: Whether to use apostrophe or not.
@@ -546,8 +552,14 @@ class IPATokenizer(BaseTokenizer):
         if hasattr(g2p, "phoneme_probability"):
             self.phoneme_probability = g2p.phoneme_probability
 
-        # Build tokens list
-        tokens = set(g2p.symbols)
+        # Build tokens list if fixed_vocab isn't set
+        if fixed_vocab:
+            tokens = set(fixed_vocab)
+            self.set_fixed_vocab = True     # Used to check whether dataset entries need filtering
+            g2p.replace_symbols(tokens)
+        else:
+            tokens = set(g2p.symbols)
+            self.set_fixed_vocab = False
 
         if apostrophe:
             tokens.add("'")
@@ -572,6 +584,8 @@ class IPATokenizer(BaseTokenizer):
             self.silence, tokens = len(tokens), tokens + [silence]
 
         super().__init__(tokens, oov=oov, sep=sep, add_blank_at=add_blank_at)
+
+        self.tokens_set = set(self.tokens)  # To save some repeated work when filtering entries
 
         self.punct = punct
         self.pad_with_space = pad_with_space
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -555,7 +555,7 @@ class IPATokenizer(BaseTokenizer):
         # Build tokens list if fixed_vocab isn't set
         if fixed_vocab:
             tokens = set(fixed_vocab)
-            self.set_fixed_vocab = True     # Used to check whether dataset entries need filtering
+            self.set_fixed_vocab = True  # Used to check whether dataset entries need filtering
             g2p.replace_symbols(tokens)
         else:
             tokens = set(g2p.symbols)
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -598,7 +598,6 @@ class IPATokenizer(BaseTokenizer):
 
         self.g2p = g2p
 
-
     def encode(self, text: str) -> List[int]:
         """See base class for more information."""
         # normalize the input text with "NFC" form.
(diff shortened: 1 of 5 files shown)


XuesongYang (changes requested):
LGTM. Could you pls add a unit test for `IPATokenizer` with on/off param `fixed_vocab` at `tests.collections.common.tokenizers.text_to_speech.test_tts_tokenizers`? just make sure the behavior of overridden symbol set is expected. Thanks.

nemo/collections/tts/data/tts_dataset.py
```diff
@@ -248,6 +258,19 @@ def __init__(
...
+                    # If so, entries with illegal graphemes should be filtered out
+                    if getattr(self.text_tokenizer, "set_fixed_vocab", False):
+                        normalized_text = file_info["normalized_text"]
+                        text_set = set(set_grapheme_case(normalized_text, self.text_tokenizer.g2p.grapheme_case))
+                        # Skip if set difference between graphemes in text and valid tokens set is nonempty
+                        set_diff = text_set - tokens_set
+                        if set_diff:
+                            logging.warning(
+                                f"Skipping entry, found illegal grapheme(s) [{set_diff}] in text: [{normalized_text}]"
+                            )
```
rlangman, line 271: Right now it is possible for users to have characters in their training dataset that are not in their dictionary. At which point the tokenizer prints a warning that there is an unknown symbol. Why does the logic need to be different for the fixed vocab scenario?

redoctopus: Good point--I'll remove the filtering code here.

redoctopus:
> LGTM. Could you pls add a unit test for `IPATokenizer` with on/off param `fixed_vocab` at `tests.collections.common.tokenizers.text_to_speech.test_tts_tokenizers`? just make sure the behavior of overridden symbol set is expected. Thanks.

Added a test to make sure the phoneme dict in G2P and the tokenization output is as expected.

rlangman (dismissed):

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -546,8 +553,14 @@ def __init__(
         if hasattr(g2p, "phoneme_probability"):
             self.phoneme_probability = g2p.phoneme_probability
 
-        # Build tokens list
-        tokens = set(g2p.symbols)
+        # Build tokens list if fixed_vocab isn't set
+        if fixed_vocab:
+            tokens = set(fixed_vocab)
```
XuesongYang, line 558: I wonder if this function takes care of the case that, e.g. 'ö' can be encoded as
    b'\xc3\xb6' (one char) as well as b'o\xcc\x88' (two chars). We discussed similar long time ago. https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/common/tokenizers/text_to_speech/tokenizer_utils.py#L96-L101

redoctopus: I'm assuming here that it's reasonable to assume that the user has passed in a "correct"/"canonical" version of the symbols they want (mostly I'm assuming they're copy/pasting from a previous config or model).

Are you suggesting we run normalize over the user input fixed vocab?

XuesongYang: hmm...i was thinking to apply the same process (calling `normalize_unicode_text`) as what we did now. But this process is applied to `g2p/modules.py` in our current implementation rather than in `tts_tokenizers.py`. I guess `replace_symbols` func should be a better place?

XuesongYang: @rlangman for better comments.

rlangman: Right now it is done both in `tts_tokenizers.py` as part of `text_preprocessing_func`, as well as in `g2p/modules.py`. I would favor putting any text normalization in `tts_tokenizers.py` where possible.

redoctopus: Alright, added a bit of code to preprocess the fixed vocab symbols in the tokenizer init.

nemo/collections/tts/data/tts_dataset.py
```diff
@@ -185,7 +185,6 @@ def __init__(
         self.phoneme_probability = None
         if isinstance(self.text_tokenizer, BaseTokenizer):
             self.text_tokenizer_pad_id = text_tokenizer.pad
-            self.tokens = text_tokenizer.tokens
```
XuesongYang, line 188: this was not used anywhere, so that's why you remove it. Is my understanding correct?

nemo/collections/tts/data/tts_dataset.py
```diff
@@ -195,7 +194,6 @@ def __init__(
                 raise ValueError(f"tokens must be specified if text_tokenizer is not BaseTokenizer")
 
             self.text_tokenizer_pad_id = text_tokenizer_pad_id
-            self.tokens = tokens
```
XuesongYang, line 198: this was not used anywhere, so that's why you remove it. Is my understanding correct?

nemo/collections/tts/g2p/modules.py
```diff
@@ -542,6 +542,60 @@ def _normalize_dict(self, phoneme_dict_obj: Dict[str, List[List[str]]]) -> Tuple
...
         return g2p_dict, symbols
 
+    def replace_symbols(self, symbols, keep_alternate=True):
+        """Replaces the vocabulary of symbols with the one given.
+        Also filters out any entries with illegal graphemes or phonemes according to the new vocab.
+
+        Args:
+            symbols (List, Set): User-provided set of valid symbols, both graphemes and phonemes
+            keep_alternate (bool): Whether to keep the other pronunciation(s) of a word if not all contain
+                illegal phonemes (and the word doesn't containi illegal graphemes).
```
XuesongYang, line 552: typo: containi -> contain

nemo/collections/tts/g2p/modules.py
```diff
@@ -542,6 +542,60 @@ def _normalize_dict(self, phoneme_dict_obj: Dict[str, List[List[str]]]) -> Tuple
...
+                Warning: this may change a word from being ambiguous to having only one valid pronunciation.
+                Defaults to True.
+        """
+        new_symbols = set(symbols)
+        if self.symbols == new_symbols:
+            logging.info("Did not replace G2P valid symbol set since the given set is equivalent to the existing one.")
+            return
```
XuesongYang, line 559: `replace_symbols` function doesn't have return values because it directly update `self.symbols`. I was wondering if it is better to move this code block outside of this function. For example, https://github.com/NVIDIA/NeMo/pull/6172/files#diff-ccb3516827d242a91442d35ce0879d40a11d7021f1f22fd2a7e34d7cec4a220fR557-R559:~:text=g2p.replace_symbols(tokens)

XuesongYang (dismissed):

nemo/collections/tts/data/tts_dataset.py
redoctopus: Yes, I was not able to find any references to this and removed it. Test runs seem fine without it as well.

nemo/collections/tts/data/tts_dataset.py
redoctopus: Yep--same as above.

nemo/collections/tts/g2p/modules.py
redoctopus: Good catch, thanks. Updating now.

nemo/collections/tts/g2p/modules.py
redoctopus: Sure, that works for me.

I was thinking about removing it entirely, but someone may want to re-compute the phoneme dictionary with a different `keep_alternate` value. But since that probably won't happen at least in the tokenizer, I'm fine with moving this check there.

XuesongYang (dismissed):

XuesongYang (approved):

PR #6178, opened 2023-03-13
Add sequence parallel support to Rope positional embedding

yidong72:
Add sequence parallel support to Rope positional embedding


MaximumEntropy (approved):
LGTM. Thanks for the fix.

PR #6182, opened 2023-03-13
Filter p-tuning by example length

arendu:
# What does this PR do ?

Previously only training and validation examples longer than the max encoder length were filtered out. This PR makes it possible to filter out by a specified length. This will save training time, as larger batch sizes will be possible after removing a small % of training and validation examples.

**Collection**: [NLP]

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -403,10 +403,13 @@ def test_epoch_end(self, outputs):
...
     def setup_training_data(self, training_data_config=None):
         if self.cfg.data.get('train_ds', None):
+            max_seq_length = self.frozen_model.cfg.encoder_seq_length
+            if "max_seq_length" in self.cfg.data and self.cfg.data.max_seq_length:
+                max_seq_length = self.cfg.data.max_seq_length
             self._train_ds, self._train_dl = self.build_virtual_prompt_dataset(
                 data=self.cfg.data.train_ds,
                 batch_size=self.cfg.global_batch_size,
-                max_seq_length=self.frozen_model.cfg.encoder_seq_length,
+                max_seq_length=max_seq_length,
```
Zhilin123, line 412: Looking at https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/data/language_modeling/megatron/gpt_prompt_learning_dataset.py#L166

Why might the sample have longer than max_sequence_length even after truncation? IN practice, this might mean that all samples get truncated and none get filtered out.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -421,10 +424,13 @@ def setup_training_data(self, training_data_config=None):
 
     def setup_validation_data(self, validation_data_config=None):
         if self.cfg.data.get('validation_ds', None):
+            max_seq_length = self.frozen_model.cfg.encoder_seq_length
```
Zhilin123, line 427: What's the benefit of asking users to specify max_seq_length instead of using the default `self.frozen_model.cfg.encoder_seq_length`. Is there a situation in which they will be different? I can foresee how this might be a problem if `max_seq_length > self.frozen_model.cfg.encoder_seq_length`, so would prefer to leave it to default `self.frozen_model.cfg.encoder_seq_length` unless there's a clear advantage that I'm missing.

Zhilin123 (commented):
Left some questions

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
arendu: it makes training inefficient. consider if 99.99999% of training examples are less than 50 tokens. But 1 example is 4000. We have to make mbs=1 just to fit this one long example. If we remove it we could probably make mbs=64 or 128.

arendu: Ideallly the validation tool should suggest a good default here.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
arendu: The idea is to remove long examples relative to other examples in training set. So its not really related to max_encoder_seq_length.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
Zhilin123: I think this makes sense - how would you like the validation tool to suggest a good default? Maybe the length at which will preserve say 95/99% of samples?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
Zhilin123: What I am wondering about is this:

Why might the ' if self.min_seq_length <= len(input_ids) <= self.max_seq_length:' ever be true since we do this truncation step first?

`# Try to truncate input text to fit into the max sequence length
            if len(input_ids) > self.max_seq_length:
                input_ids = self._truncate_input(
                    truncation_field,
                    input_ids,
                    taskname,
                    doc,
                    prompt_template,
                    prompt_template_fields,
                    virtual_token_splits,
                )

            # Skip example if the final length doesn't fit length requirements even after truncation
            if self.min_seq_length <= len(input_ids) <= self.max_seq_length:
                if self.virtual_prompt_source == VirtualPromptSource.PROMPT_ENCODER:
                    taskname_id = self.tokenizer.text_to_ids(taskname)
                elif self.virtual_prompt_source == VirtualPromptSource.NO_PROMPT:
                    taskname_id = -1
                else:
                    raise ValueError("Invalid virtual prompt source specified")

                # Find answer field indices if training and answer_only_loss is True
                answer_start_idx = None
                if answer_only_loss and self.for_train:
                    answer_start_idx = self._find_answer_start(taskname, input_ids, answer_field, doc)

                self.examples.append((taskname_id, input_ids, answer_start_idx))
            else:
                skipped += 1`

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
Zhilin123: Current the validation tool drops samples > 10000 chars (which is around 2048 tokens) at line https://github.com/NVIDIA/NeMo/blob/main/tools/customization_dataset_preparation/customization_dataset_preparation.py#L268

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
arendu: Yeah it might become redundant. But not handling that in this PR.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
arendu: 95 is too aggressive, 99% seems ok to me. But this is a separate discussion.

Zhilin123 (approved):
LGTM - there remains some more things to handle but @arendu / @Zhilin123 will work on them in a separate PR.

okuchaiev (changes requested):
Can you please modify ``build_virtual_prompt_dataset`` to print a message that sequences longer than set max_seq_length will be dropped. Ideally, we'd also like to know how many were dropped. I guess for that you'll need to modify GPTPromptLearningDataset ?

okuchaiev (approved):
Saw that requested warning is already there. LGTM, thanks!

PR #6207, opened 2023-03-15
Make sure RETRO integrates well with the core

yidong72:
# What does this PR do ?
Make sure all the unit tests passes.
Make sure the training loss is the same as with apex.


ericharper (approved):
LGTM. Thanks!

PR #6218, opened 2023-03-16
Support tp pp conversion

titu1994:
# What does this PR do ?

Adds support for changing pipeline parallel version post construction for GPT

**Collection**: [Core, NLP]

# Changelog 
- Add new script for pp conversion to avoid breaking old script (should probably be deprecated eventually when new script fully supports all functionality)
- Add some modifications to allow partially loading model parallel models into memory to extract parameters
- Add PP TP conversion support for both Megatron GPT and Megatron T5 (only when shared embeddings are used between encoder and decoder)
- Add support for forcing CPU construction of T5 and GPT models
- Add support for loading model based on AppState parameters instead of Trainer parameters.

# Usage

```python
Usage:

# Megatron GPT
python megatron_change_num_partitions.py \
    --model_file=PATH_TO_SRC_FILE \
    --target_file=PATH_TO_TGT_FILE \
    --tensor_model_parallel_size=1 \
    --target_tensor_model_parallel_size=1 \
    --pipeline_model_parallel_size=1 \
    --target_pipeline_model_parallel_size=1 \
    --precision=bf16

# Megatron T5
python megatron_change_num_partitions.py \
    --model_file=PATH_TO_SRC_FILE \
    --target_file=PATH_TO_TGT_FILE \
    --model_class="nemo.collections.nlp.models.language_modeling.megatron_t5_model.MegatronT5Model" \
    --tensor_model_parallel_size=1 \
    --target_tensor_model_parallel_size=1 \
    --pipeline_model_parallel_size=1 \
    --target_pipeline_model_parallel_size=1 \
    --target_pipeline_model_parallel_split_rank=0 \
    --precision=bf16

# NOTE: When converting large models, always ensure that you pre-extract the nemo model and then only perform conversion

$ mkdir "unpacked_nemo_file"
$ tar -xvf "<path to nemo file>" -C "<absolute path to pwd>/unpacked_nemo_file/"

python megatron_change_num_partitions.py \
    ...
    --model_extracted_dir="<Absolute path to pwd>/unpacked_nemo_file/"

# NOTE: Conversion of other model types. 
# Default model type is MegatronGPTModel, if you want another model you need to pass classpath of the model
# For example - MegatronT5Model - 

python megatron_change_num_partitions.py \
    ...
    --model_class="nemo.collections.nlp.models.language_modeling.megatron_t5_model.MegatronT5Model"

# Additional arguments:

--num_gpu_per_node: Number of GPUs per node. Default is 8.
--megatron_legacy: Whether the model is a legacy Megatron model or not. Default is False. May be unsuported for 
    Pipeline Parallelism change.
--tokenizer_model_path: Path to tokenizer model. Default is None. When not None, overrides the tokenizer model path
    in the model config.
--tokenizer_vocab_file: Path to tokenizer vocab file. Default is None. When not None, overrides the tokenizer vocab
    file in the model config.

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
```diff
@@ -0,0 +1,753 @@
...
+
+    # If target model has TP > 1 or PP > 1
+    if tgt_pp_size > 1 or tgt_tp_size > 1:
+
+        # Preserve the TP 1 PP 1 model parameters and names
+        global_params = []
+        global_params.append([p for n, p in model.named_parameters()])  # params
+        global_params.append([n for n, p in model.named_parameters()])  # names
+
+        if DEBUG_PRINT:
```
yidong72, line 632: use `logging.debug` instead of print and this `DEBUG_PRINT` condition

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -86,10 +86,21 @@ def __init__(self, cfg: DictConfig, trainer: Trainer, no_lm_init=True):
         # buffer used during train_step for logging average loss over gradient accumulation steps
         self._reduced_loss_buffer = []
 
+        # Overrides used when converting checkpoints
+        if os.environ.get(NEMO_MEGATRON_MODEL_PARALLEL_APPSTATE_OVERRIDE, "false").lower() == "true":
```
yidong72, line 90: nice hack to fool the model initialization using AppState globals.

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
```diff
@@ -0,0 +1,753 @@
...
+    if shared_enc_dec_embeddings_intermediate:
+        offset_diff += 1
+
+    # Finalize the new offset
+    new_offset = offset_diff
+
+    # Save each of the TP ranks in reverse order
+    # This is done so that the last PP rank will save the last TP rank only after all other PP TP ranks are saved
+    # The final rank will then save a new NeMo file with all other ranks inside.
+    for tp_rank in range(tp_size - 1, -1, -1):
```
yidong72, line 425: nice work around for saving the nemo files with one worker.

yidong72 (commented):
pretty cool PR. just have one minor comment.

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
titu1994: Fixed all prints to debug / info / warning logging

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
titu1994: @aklife97 came up with this for TP change, turns out its needed for PP change too :D

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
ericharper, line 1: why `_pp` ? do we have multiple partition change scripts now?

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -658,3 +664,122 @@ def _fetch_next_batch(self, iterator: Iterator) -> None:
...
+    virtual_pipeline_model_parallel_size: Optional[int] = None,
+    model_type: 'ModelType' = None,  # Default of ModelType.encoder_or_decoder
+    *args: Any,
+    **kwargs: Any,
+) -> List[torch.nn.Module]:
+    """Build the model satisfying pipeline model parallel requirements.
+
+    This function sets `pre_process` and `post_process` to `**kwargs` and pass `*args` and `**kwargs` to
+    `model_provider_func`.
+
```
ericharper, line 681: Can you add a note explaining why we need this here?

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
titu1994: Yes, for now I am keeping the two scripts (earlier one with same name and this one with `_pp` separate, but in due time this script should completely supersede the earlier one (and at that time, we can replace the earlier script with this one). This is for two reasons -

1) This script is a strict superset of the earlier one, but has multiple modifications and overhead to support PP (but maintains the core of the tp change logic). The pp logic can make it difficult to test tp change logic in isolation, unless you stick to synthetic models (small random init model developed with tp 1 pp 1 in order to force exact equivalence test after conversion).

2) This script uses a lot of experimental features - mock init of the model using AppState instead of Trainer, modification to the Apex call to force init on CPU, quite odd logic to track and map PP parameters for each rank etc. If the earlier script is well tested, we should test this script more before we replace the older one.

titu1994: But if those are ok and theres no need to worry about preserving the old script, then I'm ok with overwriting it with this one.

nemo/collections/nlp/parts/nlp_overrides.py
titu1994: Done

examples/nlp/language_modeling/megatron_change_num_partitions_pp.py
titu1994: Overwrote the older version of the script according to discussion.

aklife97: Thanks! having 2 scripts is just unnecessary overhead, just having 1 is great

ericharper (dismissed):
LGTM. Thanks!

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -171,67 +499,290 @@ def main():
...
+            trainer = Trainer(devices=1, strategy=NLPDDPStrategy(), accelerator="cpu", precision=precision)
+            if args.tokenizer_model_path is not None:
+                model.cfg.tokenizer.model = args.tokenizer_model_path
+
+            model = cls(model.cfg, trainer).to('cpu')
+            model._save_restore_connector = NLPSaveRestoreConnector()
+            model.to(dtype=dtype)
+
+            # Update global batch size
+            if old_global_batch_size % new_global_batch_size != 0 or old_global_batch_size < new_global_batch_size:
```
aklife97, line 734: nit: can we just keep the first condition here? I don't think we need the second one

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -12,36 +12,204 @@
...
+        if idx >= pp_lens[pp_rank]:
+            pp_rank += 1
+            idx = 0
+
+            # For EncDec models, after the encoder-decoder PP split occurs,
+            # the vocab and positional embeddings are duplicated across the PP ranks at the
+            # beginning of the decoder rank. We can skip them during the merge step.
+            if pp_total_len > total_params_merged:
+                if og_pp_split_rank > 0 and og_pp_split_rank == pp_rank:
+                    logging.info(
```
aklife97, line 182: can we still check that the next 2 idx are vocab and positional embeddings? (some like if 'vocab' in name) Want to avoid cases where things can break silently in case new models classes get added that don't follow this regime

aklife97 (commented):
LGTM! just a couple of minor comments, everything else looks great!

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: The second one is needed because just being divisible is not enough, at minimum it must be the size of the new world size or assert breaks and model can't be built

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: There won't be silent errors, end of the config has exact param count check, every single param has to match or it will raise an error.

I can't rework the logic for this, because I cannot make assumptions of what the parameter types of names are. It would make pp change much more complicated

aklife97:
also, can we add a PP change CI as well? would be helpful to keep testing that since the PR brings in global overrides that may cause issues if someone changes it

examples/nlp/language_modeling/megatron_change_num_partitions.py
aklife97: sounds good

examples/nlp/language_modeling/megatron_change_num_partitions.py
aklife97: just to double-check, did you intend the second condition to be `old_global_batch_size < new_global_batch_size`?
if so, condition 1 is always True when condition 2 is since GBS is >0

but its just a nit, don't mind keeping it the same way if you prefer that

titu1994: I don't remember exactly what config of tp pp conversion I needed both flags for, but it is needed for at least one weirs T5 config conversion

aklife97: sounds good

titu1994:
Good point about jenkins test - updated old one from tp reduce and increase to jointly increase pp by even or odd number,
Though this tests only GPT. We need nightly tests that test the whole matrix of TP (inc x dec) x PP (inc x dec) x {GPT, T5} - but that is super expensive on this CI. Will need to look into how to set that up.

aklife97 (approved):
LGTM, thanks!

aklife97 (approved):

PR #6222, opened 2023-03-16
[NLP] Support T5 with Megatron Core

SeanNaren:
# What does this PR do ?

Adds megatron core support for our T5 model. This only works for pre-training, due to dynamic max lengths in the GLUE/XLNI datasets fine-tuning seems to be broken. 

The fix going forward I think will be to pad to the maximum length to ensure the size of sequences are always the same. This is a requirement of the iterator object now used in the `training_step`.

I also need to confirm that the weights of the model are the same between main and this branch.

- [ ] Fix support for GLUE/XLNI Fine-tuning
- [x] Ensure convergence is the same
- [x] Fix support for encode/decode functions

**Collection**: NLP

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
@@ -140,8 +140,8 @@ class BaseMegatronBatchSampler:
 
 class MegatronPretrainingBatchSampler(BaseMegatronBatchSampler):
     def get_start_end_idx(self) -> Tuple[int, int]:
-        start_idx = self.data_parallel_rank * self._global_batch_size_on_this_data_parallel_rank
-        end_idx = start_idx + self._global_batch_size_on_this_data_parallel_rank
+        start_idx = self.data_parallel_rank * self.micro_batch_size
+        end_idx = start_idx + self.micro_batch_size
         return start_idx, end_idx
 
     def __iter__(self):
@@ -149,12 +149,16 @@ class MegatronPretrainingBatchSampler(BaseMegatronBatchSampler):
         # Last batch will be dropped if drop_last is not set False
         for idx in range(self.consumed_samples, self.total_samples):
             batch.append(idx)
-            if len(batch) == self._global_batch_size:
+            if len(batch) == self.micro_batch_size:
                 # start_idx, end_idx = self.get_start_end_idx()
+
+                """
+                if data parallel size was 2, batch size was 8
+                """
                 indices = [
-                    batch[i] for i in range(self.data_parallel_rank, self._global_batch_size, self.data_parallel_size,)
+                    batch[i] for i in range(self.data_parallel_rank, self.micro_batch_size, self.data_parallel_size,)
                 ]
-                assert len(indices) == self._global_batch_size_on_this_data_parallel_rank
+                assert len(indices) == self.micro_batch_size
                 yield indices
                 # yield batch[start_idx:end_idx]
                 batch = []
--- a/nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
@@ -146,17 +146,18 @@ class MegatronPretrainingBatchSampler(BaseMegatronBatchSampler):
 
     def __iter__(self):
         batch = []
+
+        data_parallel_micro_batch_size = self.data_parallel_size * self.micro_batch_size
         # Last batch will be dropped if drop_last is not set False
         for idx in range(self.consumed_samples, self.total_samples):
             batch.append(idx)
-            if len(batch) == self.micro_batch_size:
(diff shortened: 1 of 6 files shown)


nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
```diff
@@ -140,21 +140,25 @@ def __iter__(self):
 
 class MegatronPretrainingBatchSampler(BaseMegatronBatchSampler):
```
SeanNaren, line 141: The changes here will impact retro and nmt models too. Before megatron core, the batch would consist of `global_batch_size` number of samples.

After megatron core and the introduction of `dataloader_iter` to the `step` functions, we should pass microbatches and allow megatron to handle the fetching of batches itself.

SeanNaren:
Seems formatting on the `GPT_integrate_core` was off and included in this PR. 

I've pushed formatting changes directly to the `GPT_integrate_core` branch @aklife97

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
```diff
@@ -200,25 +199,6 @@ def on_train_epoch_start(self) -> None:
         self.on_validation_epoch_end()
         return super().on_train_epoch_start()
 
-    def training_step(self, batch, batch_idx):
```
aklife97, line 203: why are we removing this?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
```diff
@@ -405,20 +360,36 @@ def training_step(self, batch, batch_idx):
...
         lr = self._optimizer.param_groups[0]['lr']
-        self.log('lr', lr, rank_zero_only=True)
-        self.log('global_step', self.trainer.global_step, prog_bar=True, rank_zero_only=True)
+        self.log('lr', lr, rank_zero_only=True, batch_size=self.cfg.global_batch_size)
+        self.log(
+            'global_step',
+            self.trainer.global_step,
+            prog_bar=True,
+            rank_zero_only=True,
+            batch_size=self.cfg.global_batch_size,
```
aklife97, line 373: should we set batch_size=1? I am not sure how it affects things on PTL side

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
SeanNaren: Ah thanks @aklife97 

Because we do not have access to the batch we're not able to run this check.

I think for now what we can do is replace this with an assertion to ensure that `drop_last=True` is always true. What do you think?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
SeanNaren: Followed up with @aklife97 I need to investigate what this really does.

@awaelchli @carmocca when using `dataloader_iter` and setting the `batch_size` in `self.log`, what does this batch size do? Is it ok to set to 1?

awaelchli: Hey @SeanNaren 

I think you are setting this here because for the special case of passing dataloader_iter to training_step, there is a warning in Lightning saying: 

> With `def training_step(self, dataloader_iter)`, `self.log(..., batch_size=...)` should be provided

You can set it to 1 to avoid the warning. Normally Lightning uses the batch size to weight each log when averaging it (on_epoch=True). If you know that all batch sizes are the same, or you don't care about the averaging, then the value can be arbitrary here (e.g. 1).

I'm not 100% sure why it is required for the dataloader_iter case, but I think it is just that Lightning won't be able to infer the batch size automatically otherwise.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
aklife97: I think we're good without an assert here, since we anyways don't support drop_last as False at train time. So looks good!

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
aklife97: Thanks @awaelchli, @SeanNaren 

@ericharper, FYI

carmocca: > I'm not 100% sure why it is required for the dataloader_iter case

This is because lightning needs access to the batch to get its size.
But with `dataloader_iter`, the dataloader is consumed in the `training_step` directly, so lightning doesn't have a reference to the batch.

PR #6233, opened 2023-03-17
[ASR] Re-organizing data simulator with faster audio loading speed for data-simulator

tango4j:
# What does this PR do ?

Added a feature to data-simulator that loads audio files faster by limiting the target audio duration.

**Collection**: ASR

# Changelog 
- Audio reading functions in `build_sentence` are updated
- Added fully reproducible seeding mechanism for all the randomness.
- Changed all the wav reading functions to `AudioSegment.from_file()`. 
- Moved util functions to `data_simulation_utils.py`
- Added unit tests in `test_data_simul_utils.py`

# Usage
For relatively longer audio file sources (e.g. Fisher, Swbd etc) set the following 
`data_simulator.session_params.max_audio_read_sec`
to 2.5~3.5 seconds   (min is 2.5 sec). 
The shorter the faster the simulation is, but it will concatenate lots of words to form a sentence (speaker turn)

For datasets that are already segmented (e.g VoxCeleb, LibriSpeech etc), `max_audio_read_sec` does not need to be short (no speed gains). For already-segmented datasets, I recommend 10~20 seconds.

When we choose 2.5 secs for `data_simulator.session_params.max_audio_read_sec` in case of unsegmented datasets (Fisher, SWBD, TEDtalks, AMI, etc), we can reach 80~90% speed of the segmented datasets (Libri, Vox, etc)

```python
  python <NEMO_ROOT>/tools/speech_data_simulator/multispeaker_simulator.py  \
    num_workers=10 \
    data_simulator.random_seed=42 \
    data_simulator.manifest_filepath=manifest_with_alignment_file.json \
    data_simulator.session_params.max_audio_read_sec=3.0 \
    data_simulator.outputs.output_dir=./simulated_data \
    data_simulator.outputs.output_filename=sim_spk2_sess20 \
    data_simulator.session_config.num_sessions=1000 \
    data_simulator.session_config.num_speakers=2 \
    data_simulator.session_config.session_length=20 \
    data_simulator.background_noise.add_bg=False \
    data_simulator.background_noise.background_manifest=background_noise.json \
    data_simulator.background_noise.snr=40 \

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

Anyone in ASR


nemo/collections/asr/data/data_simulation.py
```diff
@@ -979,6 +1000,164 @@ def _add_file(
...
+        # Find all silence indices
+        sil_inds = np.where((np.array(audio_manifest['words']) == '') == True)[0]
+        # If the audio file is shorter than the max_audio_read_sec, then we don't need to read a subset of the audio file.
+        if (
+            len(sil_inds) <= self.min_alignment_count
+            or (audio_manifest['alignments'][-1] - audio_manifest['alignments'][0]) < self.max_audio_read_sec
+        ):
+            return 0
+        else:
+            offset_index = np.random.randint(len(sil_inds) - 1)
```
stevehuang52, line 1097: I wonder it this random selection will make the resulting audio very short? For example, the audio is 21s and `max_audio_read_sec==20`, the if `offset_index` is very close to the end of audio, the resulting sub-audio could be just 1s. Shall we change the random range so that the resulting audio is about `max_audio_read_sec` long? One thought is using binary search to find the longest random range:
```python
left = 0
right = len(sil_inds) - 1
while left < right:
    mid = left + (right - left) // 2
    dur_left = audio_manifest['alignments'][-1] - audio_manifest['alignments'][sil_inds[mid]]
    if  dur_left < self.max_audio_read_sec:
        right = mid - 1
    elif dur_left > self.max_audio_read_sec:
        left = mid + 1
    else:
        break
offset_index = np.random.randint(left + (right - left) // 2)
```
However, if it's ok to have very short sub-audio, we can just keep your current design.

tango4j: Reflected this binary search for securing `min_audio_read_sec`.
However, this still does not guarantee that we will always have enough words cause there are cases where the last word (alignment) is far away from the second last word. 
In this rare case, we need to use 1 word. Other cases are all handled.

+ Added reproducibility (seeding all the necessary randomizations) and checked now it is reproducible with the input seed.

+ Added a warning where 1 word is longer than `min_audio_read_sec`. This happens with forced alignment (libriSpeech or Fisher) where laughter and some long words are longer than 2.5 seconds.
We pop out a warning message then truncate the audio to have `min_audio_read_sec`.

stevehuang52 (approved):
Really awesome work, thanks~!

PR #6239, opened 2023-03-17
Add the capability to stop generation based on customized end strings

yidong72:
# What does this PR do ?

Stop the text generation based on user provided end strings.


nemo/collections/nlp/modules/common/megatron_web_server.py
```diff
@@ -60,6 +62,7 @@ def get_demo(share, username, password, server_port=5555, web_port=9889, loop=No
                 repetition_penality = gr.Slider(
                     minimum=1.0, maximum=5.0, step=0.02, value=1.2, label='Repetition penalty'
                 )
+                end_strings = gr.Textbox(label="End strings (comma separated)", value="<|endoftext|>,", lines=1,)
```
MaximumEntropy, line 65: <|endoftext|> is hardcoded for the GPT2 tokenizer, we need to make this general no?

nemo/collections/nlp/modules/common/text_generation_server.py
```diff
@@ -156,6 +157,14 @@ def put(self):
             if neighbors < 0:
                 return "num of neighbors must be an integer no less than 0"
 
+        end_strings = ['<|endoftext|>']
```
MaximumEntropy, line 160: I think the default end_strings should be the empty list, to have nothing. You are already handling <|endoftext|> in your text generation utils by checking for tokenizer.eos_id no?

MaximumEntropy (changes requested):
Thanks! Just left a couple of minor comments about hardcoding for the GPT2 tokenizer and defaults for end strings.

nemo/collections/nlp/modules/common/megatron_web_server.py
yidong72: It is used as a text representation of eos_id. Each tokenizer will use its own eos_id. So it is general.

michalivne: I think it's default value for UI, right?

nemo/collections/nlp/modules/common/megatron_web_server.py
```diff
@@ -26,7 +26,7 @@
 
 
 def create_gen_function(port=5555):
-    def get_generation(prompt, greedy, add_BOS, token_to_gen, min_tokens, temp, top_p, top_k, repetition):
+    def get_generation(prompt, greedy, add_BOS, token_to_gen, min_tokens, temp, top_p, top_k, repetition, end_strings):
```
michalivne, line 29: Maybe `temperature` is useful here too?

nemo/collections/nlp/modules/common/text_generation_server.py
yidong72: if empty list, it will not stop generating. Now we use text generation strategy to decide when to stop. `['<|endoftext|>']` is the default behavior that uses eos_id to stop.

nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -132,6 +134,36 @@ def post_process(self, tokens: torch.Tensor, new_tokens: torch.Tensor, context_l
...
+        return whether the generation should stop based on the previous token
+        Args:
+            tokens (torch.Tensor): the generated tokens so far
+            prev  (torch.Tensor): the previous token
+            eod_id (int): the end of document token id
+            end_strings (List[str]): the list of end of generation strings
+        returns:
+            a boolean tensor indicating whether the generation should stop
+        """
+        if len(end_strings) == 1 and end_strings[0] == END_OF_SEQ:
```
michalivne, line 150: Can you add to comment near definition of `END_OF_SEQ`it is the default `<EOS>`?

michalivne (dismissed):
LGTM! See minor comments

nemo/collections/nlp/modules/common/megatron_web_server.py
yidong72: yes. it is default for text generation too

nemo/collections/nlp/modules/common/megatron_web_server.py
yidong72: `temp` is temperature here.

nemo/collections/nlp/modules/common/text_generation_strategy.py
yidong72: yes. will add the comment.

michalivne (approved):
LGTM!

MaximumEntropy (approved):

PR #6254, opened 2023-03-19
Fix buckeing seeding

VahidooX:
# What does this PR do ?

When the global seeding is set and controlled by the user, the fully_randomized bucketing uses the same order of buckets on all workers. This PR fixes this seeding bug of the bucketing dataset by adding the rank of data loader to the seed.

# Changelog 
- Add the rank of the data worker to the bucketing dataset to keep the randomness for fully_randomzied mode of buceking.

  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation


titu1994 (approved):
Looks good !

PR #6260, opened 2023-03-20
Hybrid Autoregressive Transducer (HAT)

andrusenkoau:
# What does this PR do ?

Add HAT model as a new joint network type (HATJoint) for RNNT model. The difference is only in decoding time -- HAT.joint.joint returns two outputs: hat_logprobs and internal_lm_logprobs (for internal lm subtraction in case of Shallow Fusion with external n-gram LM).

**Collection**: [ASR]

# Usage
* For HAT model training you need replace `_target_: nemo.collections.asr.modules.RNNTJoint` with `_target_: nemo.collections.asr.modules.HATJoint` in `joint` part of standard transducer config.

* For Shallow Fusion with external n-gram LM use RNNT `maes` decoding algorithm which is able to work with HATJoint model.  

```python
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/asr/modules/hat.py
```diff
@@ -0,0 +1,531 @@
...
+    LengthsType,
+    LogprobsType,
+    LossType,
+    NeuralType,
+    SpectrogramType,
+)
+from nemo.utils import logging
+
+
+class HATJoint(rnnt_abstract.AbstractRNNTJoint, Exportable, AdapterModuleMixin):
```
titu1994, line 55: This class is duplicating a lot of code from RNNTJoint. Would it make sense to subclass it ?

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -460,7 +466,12 @@ def greedy_search(
 
             # TODO: Figure out how to remove this hard coding afterwords
             while not_blank and (symbols_added < 5):
-                ytu = torch.log_softmax(self.joint.joint(hi, y) / self.softmax_temperature, dim=-1)  # [1, 1, 1, V + 1]
+                if isinstance(self.joint, HATJoint):
+                    ytu, _ = self.joint.joint(hi, y)
```
titu1994, line 470: This kinda logic is problematic in the long run. Why not take a bool in the HAT module that determine what self.joint returns - by default it's set and returns both items, otherwise return things in the form of RNNT so that this code doesn't need to change

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -34,6 +34,7 @@
 from omegaconf import DictConfig
 
 from nemo.collections.asr.modules import rnnt_abstract
+from nemo.collections.asr.modules.hat import HATJoint
```
titu1994, line 37: No modules should be imported inside of Greedy of Beam decoding libraries because it will eventually cause circular dependency

titu1994 (changes requested):
Needs a bit of refactoring

nemo/collections/asr/modules/hat.py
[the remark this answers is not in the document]
andrusenkoau: Done

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
andrusenkoau: I thought that jit compiler does not like variable outputs numbers. Now I made default mode -- return only logprobs (like the standard rnnt joint) and return both logprobs and internal_lm_logprobs (in case of special boolean flag). This is allowed to save more rnnt decoding code unchanged.

nemo/collections/asr/modules/hybrid_autoregressive_transducer.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+
+        self.pred, self.enc, self.joint_net, self.blank_pred = self._joint_hat_net_modules(
+            num_classes=self._vocab_size,  # non blank symbol
+            pred_n_hidden=self.pred_hidden,
+            enc_n_hidden=self.encoder_hidden,
+            joint_n_hidden=self.joint_hidden,
+            activation=self.activation,
+            dropout=jointnet.get('dropout', 0.0),
+        )
```
titu1994, line 146: Add a property (settable) to this class that sets whether to return_hat_ilm (default to false)

nemo/collections/asr/modules/hybrid_autoregressive_transducer.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def joint(
+        self, f: torch.Tensor, g: torch.Tensor, return_ilm: bool = False
+    ) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
```
titu1994, line 150: Remove return_ilm from here, use the properties

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -325,6 +328,9 @@ def __init__(
         else:
             self.ngram_lm = None
 
+        self.hat_subtract_ilm = hat_subtract_ilm
+        self.hat_ilm_weight = hat_ilm_weight
```
titu1994, line 332: If this is set, assert that hasattr(self.joint, "return_hat_ilm"). Only assert, don't set it yet

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -325,6 +328,9 @@ def __init__(
...
+
```
titu1994, line 333: Inside the  `__call__`, see if the ilm flag is set in the self.hat_subtract_ilm (self is the decoding beam search class), do a hasattr check again and now set the value. 

Keep a coph of the original value and then at the end of the call function, before returning, set the ilm value back to the original value. 

All of this is so that 1) hat model is by default running like a normal RNNT model. 2) only inside beam search it will become used as the ilm result.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -1152,9 +1158,13 @@ def modified_adaptive_expansion_search(
                 beam_dec_out = torch.stack([h.dec_out[-1] for h in hyps])  # [H, 1, D]
 
                 # Extract the log probabilities
-                beam_logp, beam_idx = torch.log_softmax(
-                    self.joint.joint(beam_enc_out, beam_dec_out) / self.softmax_temperature, dim=-1,
-                ).topk(self.max_candidates, dim=-1)
+                if isinstance(self.joint, hat.HATJoint) and self.hat_subtract_ilm:
```
titu1994, line 1161: Here, everywhere, simply call the self.joint.joint(with the ordinary arguments for RNNT). The output can now be either torch.Tensor - (RNNT joint, HAT without the ILM subtract) or it can be HATOutput dataclass. 

import RNNT utils and then check if torch.is_tensor(output) here - this is for og RNNT. Elif self.hat_subtract_ilm and isinstance(output, HATOutput): 

Then do the required code path. On else path, raise error saying could not resolve the output

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
```diff
@@ -1196,7 +1206,12 @@ def modified_adaptive_expansion_search(
                                     lm_score, new_hyp.ngram_lm_state = self.compute_ngram_score(
                                         hyp.ngram_lm_state, int(k)
                                     )
-                                    new_hyp.score += self.ngram_lm_alpha * lm_score
+                                    if isinstance(self.joint, hat.HATJoint) and self.hat_subtract_ilm:
```
titu1994, line 1209: Same for everywhere else below.

titu1994 (changes requested):
Currently the code is too circular for HAT import. Another thing is it requires too many modifications to an already very complicated function (mAES). 

The first thing we can make more generic with dataclass and property trucks. Those changes are relatively simple but require some refactor. 

The second one I dunno how to make more generic. Perhaps an abstract method inside of AbstractRNNTJoint that discussed how to do special forward of joint ? That's a heavy refactor so ignore it for now.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
andrusenkoau: Hi @titu1994! Thank you for detailed review. I tried to modify HAT related code according to your suggestions. For convenience I also added `resolve_joint_output` function. Check it pls.

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
titu1994, line 1: I see we support only basic beam and maes. Can you look into supporting HAT with other algos ? If it's simple, it can be done in this pr, if not in another pr.

titu1994 (commented):
It looks very good now, could you add some tests that assert that normally forward returns tensor and hat forward with and without flag set returns either tensor or HATJointOutput. 

Another thing is we support only mAES and normal beam, can you look into the complexity of the other beam algos to support hat ? If it's difficult we can leave it to another pr in the future

kobenaxie:
Hi @andrusenkoau , the HATJoint returns log softmaxed `log_prob`, but the rnnt_loss in torchaudio or  [rnnt_pytorch](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/losses/rnnt_pytorch.py#L48) receives `logits` without logsoftmax, should this be unified ?

titu1994:
@kobenaxie that's a template implementation of the loss using pure PyTorch, it is not used during actual training since it is super slow. Instead we use numba bases cuda compiled loss. 

Also, hat during training does not return the dataclass (which the loss anyway would not accept) so it is fine

nemo/collections/asr/parts/submodules/rnnt_beam_decoding.py
andrusenkoau: Do you mean n-gram LM fusion (with RNNT and HAT) for other decoding algorithms? Now only `maes` algorithm supports LM fusion. I did not do it for `default beam search` because it works too slow. I do not think anyone wants to use it because of speed.

BTW, all the decoding algorithms can work now with HAT model without LN fusion because HATJoint has the same default output type like RNNTJoint.

titu1994: Oh ok sounds good then

titu1994 (approved):

titu1994:
Looks great !

titu1994:
Final things to do are to add HAT decoder based conformer config to a conf dir called conf/hat_transducer/conformer/conformer_hat_bpe.yaml / char.yaml

titu1994:
That can be done when release bench is cut.

PR #6284, opened 2023-03-23
[TTS] Fix TTS audio preprocessing script

rlangman:
# What does this PR do ?

Fix some bugs and improve usability of TTS preprocessing script.

**Collection**: [TTS]

# Changelog
- Replace hydra runner based script with regular python script that uses OmegaConf to parse .yaml config. The hydra script used AudioPreprocessingConfig as input, but in some situations this might not work because classes in the "scripts" directory are typically not installed or importable.
- Update the audio trimmer class paths because they were moved to "tts.parts.preprocessing", and moved them from "scripts" to "examples".
- Support relative paths in manifest.
- Ignore min_duration and max_duration filters when they are not provided.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


examples/tts/conf/trim/energy.yaml
redoctopus, line 1: Should this be a `git mv` for better tracking of the file?

rlangman: I just modified it to be 2 commits; one which only renames with git mv and the other which updates the target path. Though the final diff on github with both commits still shows it as a deletion + creation for some reason.

redoctopus (dismissed):
Ah well, even if it shows up as a delete and an add, not the end of the world. LGTM

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -23,6 +23,17 @@
 from torch.special import gammaln
 
 
+def get_audio_paths(audio_path, base_path):
+    if os.path.isabs(audio_path):
+        abs_path = audio_path
+        rel_path = audio_path.relative_to(base_path)
```
XuesongYang, line 29: `relative_to` is a function for `pathlib.Path` object. Could you pls add type hints in the function?

XuesongYang, line 29: `relative_to` is a function for `pathlib.Path` object. Could you pls add type hints in the function?

scripts/dataset_processing/tts/audio_processing/preprocess_audio.py
```diff
@@ -99,53 +136,51 @@ def _process_entry(
...
 
+    output_path.parent.mkdir(exist_ok=True, parents=True)
     sf.write(file=output_path, data=audio, samplerate=sample_rate)
 
-    original_duration = librosa.get_duration(filename=str(audio_filepath))
+    original_duration = librosa.get_duration(filename=str(audio_path))
     output_duration = librosa.get_duration(filename=str(output_path))
 
-    entry["audio_filepath"] = output_path
+    entry["audio_filepath"] = str(output_audio_filepath)
```
XuesongYang, line 145: nit: this function actually applied many forced conversions from `pathlib.Path` to `str`. I feel it is redundant. Maybe we can only specify a `str` type for `input_audio_dir` and `output_audio_dir`?

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: Fixed

XuesongYang (approved):

scripts/dataset_processing/tts/audio_processing/preprocess_audio.py
XuesongYang: looks good to me. Thanks.

PR #6324, opened 2023-03-29
Move checkpoint consolidation to cpu

wdykas:
# What does this PR do ?

This aims to move the megatron checkpoint to nemo conversion to cpu rather than using pytorch distributed following additions introduced [here](https://github.com/NVIDIA/NeMo/pull/6218).  This is useful for large models where the user may have checkpoints to consolidate but don't want to use the massive amount of gpus just for conversion.

**Collection**: [NLP]

# Changelog 
- Change megatron_ckpt_to_nemo.py to use CPU only

# Usage
```python
python3 megatron_ckpt_to_nemo.py \
     --checkpoint_folder <path_to_PTL_checkpoints_folder> \
     --checkpoint_name <checkpoint_name> \
     --nemo_file_path <path_to_output_nemo_file> \
     --tensor_model_parallel_size <tensor_model_parallel_size>\
     --pipeline_model_parallel_size <pipeline_model_parallel_size> \
     --model_type <model type>
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
+++ b/examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
@@ -15,21 +15,22 @@
 r"""
 Conversion script to convert PTL checkpoints into nemo checkpoint.
   Example to run this conversion script:
-    python -m torch.distributed.launch --nproc_per_node=<tensor_model_parallel_size> * <pipeline_model_parallel_size> \
-     megatron_ckpt_to_nemo.py \
+     python3 megatron_ckpt_to_nemo.py \
      --checkpoint_folder <path_to_PTL_checkpoints_folder> \
      --checkpoint_name <checkpoint_name> \
      --nemo_file_path <path_to_output_nemo_file> \
-     --tensor_model_parallel_size <tensor_model_parallel_size> \
-     --pipeline_model_parallel_size <pipeline_model_parallel_size>
+     --tensor_model_parallel_size <tensor_model_parallel_size>\
+     --pipeline_model_parallel_size <pipeline_model_parallel_size> \
+     --model_type <model type>
 """
 
 import os
+import tempfile
+import itertools
+import shutil
 from argparse import ArgumentParser
 
 import torch
-from apex.transformer import parallel_state
-from pytorch_lightning.plugins.environments import TorchElasticEnvironment
 from pytorch_lightning.trainer.trainer import Trainer
 
 from nemo.collections.nlp.models.language_modeling.megatron_bart_model import MegatronBARTModel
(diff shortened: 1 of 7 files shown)


examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
```diff
@@ -82,91 +87,100 @@ def get_args():
...
+            checkpoint_path, hparams_file=args.hparams_file, trainer=trainer
+        ).to("cpu")
     elif args.model_type == 'retro':
         model = MegatronRetrievalModel.load_from_checkpoint(
             checkpoint_path, hparams_file=args.hparams_file, trainer=trainer
-        )
-    model._save_restore_connector = NLPSaveRestoreConnector()
+        ).to("cpu")
+
+    checkpoint_dir = args.checkpoint_folder
```
titu1994, line 138: Is the following section necessary or can be replaced with just Nemo model.save_to("basepth") and let it automatically calculate tp pp path + save the checkpoint and then finally repackage? 

If that doesn't work, below code is fine

examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
```diff
@@ -82,91 +87,100 @@ def get_args():
...
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+
+        if app_state.pipeline_model_parallel_size == 1:
+            # move weights to the tmpdir
+            for tp_rank in range(app_state.tensor_model_parallel_size):
+                os.makedirs(os.path.join(tmpdir, f'mp_rank_{tp_rank:02d}'))
+                mp_model_weights = os.path.join(checkpoint_dir, f'mp_rank_{tp_rank:02d}', checkpoint_name)
+                # TODO: Change to a move
+                shutil.copy(
```
titu1994, line 150: Copy then rm?

titu1994 (commented):
Overall this looks correct and much better approach than spinning up multi node just to modify checkpoint save

Fyi @ericharper for final review

examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
wdykas: Copy then RM makes more sense here.  If on the above comment I can just call save_to then I think this might be irrelevant anyway.

examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
wdykas: I think the only thing keep us from using save_to is the weird pathing https://github.com/NVIDIA/NeMo/blob/b9cf05cf76496b57867d39308028c60fef7cb1ba/nemo/collections/nlp/parts/nlp_overrides.py#L245.  I do the same thing as here but change the paths properly

examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
wdykas: I think this is ready for someone to take a look at. I just realized that because I have the copy under a tmp directory its actually like a copy+rm operation already.

PR #6346, opened 2023-04-01
[Core] return_config=True now extracts just config, not full tarfile

titu1994:
# What does this PR do ?

Make return_config=True now only extract out the yaml files from the tarfile, preventing opening and file extractions of all files (should significantly speedup return_config=True for Billion parameter models).

**Collection**: [Core, ASR, NLP, TTS]

# Changelog 
- Add flag to extract just config from core internals

# Usage

```python
model_config = ModelClass.restore_from(..., return_config=True)
OR
model_config = ModelClass.from_pretrained(..., return_config=True)
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/nlp/parts/nlp_overrides.py
+++ b/nemo/collections/nlp/parts/nlp_overrides.py
@@ -372,7 +372,7 @@ class NLPSaveRestoreConnector(SaveRestoreConnector):
         loaded_params = super().load_config_and_state_dict(
             calling_cls, restore_path, override_config_path, map_location, strict, return_config, trainer,
         )
-        if not isinstance(loaded_params, tuple):
+        if not isinstance(loaded_params, tuple) or return_config is True:
             return loaded_params
         conf, instance, state_dict = loaded_params
         state_dict = self.modify_state_dict(conf, state_dict)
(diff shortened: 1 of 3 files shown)


nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -372,7 +372,7 @@ def restore_from(
         loaded_params = super().load_config_and_state_dict(
             calling_cls, restore_path, override_config_path, map_location, strict, return_config, trainer,
         )
-        if not isinstance(loaded_params, tuple):
+        if not isinstance(loaded_params, tuple) or return_config is True:
```
SeanNaren, line 375:
```suggestion
        if not isinstance(loaded_params, tuple) or return_config:
```

Could we not just simplify to this? Seems it's always a bool

SeanNaren (approved):
Looks great, thanks for doing this!

nemo/collections/nlp/parts/nlp_overrides.py
titu1994: I prefer explicit checks generally, even for bool assertion

okuchaiev (approved):

PR #6347, opened 2023-04-02
specaug speedup

1-800-BAD-CODE:
# What does this PR do ?

Faster implementation of non-numba specaug

**Collection**: ASR

# Changelog 

Rather than repeatedly modify the features tensor in-place, build a mask and fill the features tensor once. By my measurements, about 20x faster on GPU.

It could be faster still, but it gets hard to compare output exactly to the original implementation to verify correctness.

Also, the original implementation seems to be biased away from the upper freq bins, but again fixing it would be make it impossible to compare the outputs to the original implementation.

# Usage
Running this snippet will run both the original and this implementation, verify similar output, and print latencies.

<details>
  <summary>Test snippet</summary>

```python
import random
import time
import torch

from nemo.collections.asr.parts.submodules.spectr_augment import SpecAugment
from nemo.core.classes import typecheck


# Copy of original `SpecAugment.forward` for comparison
@typecheck()
@torch.no_grad()
def forward_original(self, input_spec, length):
    sh = input_spec.shape

    for idx in range(sh[0]):
        for i in range(self.freq_masks):
            x_left = self._rng.randint(0, sh[1] - self.freq_width)

            w = self._rng.randint(0, self.freq_width)

            input_spec[idx, x_left : x_left + w, :] = self.mask_value

        for i in range(self.time_masks):
            if self.adaptive_temporal_width:
                time_width = max(1, int(length[idx] * self.time_width))
            else:
                time_width = self.time_width

            y_left = self._rng.randint(0, max(1, length[idx] - time_width))

            w = self._rng.randint(0, time_width)

            input_spec[idx, :, y_left : y_left + w] = self.mask_value

    return input_spec


seed = 12345
batch_size = 128
feat_dim = 80
max_length = 1000
device = "cuda"
num_iterations = 100

# Generate some inputs
spec = torch.randn(size=[batch_size, feat_dim, max_length], device=device)
lengths = torch.randint(low=max_length // 10, high=max_length, size=[batch_size], device=device)
# Usually, at least one element in a batch is the max length.
lengths[0] = max_length
print(f"Testing with input features shape {spec.shape}")

# This version has the new forward
augmentor: SpecAugment = SpecAugment(
    freq_masks=2, time_masks=10, freq_width=27, time_width=0.05, rng=random.Random(seed)
).to(device)
# cache the first output for comparison (need to be careful of RNG values)
first_output_new = augmentor(input_spec=spec, length=lengths)

# Warm up
augmented_spec = augmentor(input_spec=spec, length=lengths)
# Loop and record time
start = time.time()
for _ in range(num_iterations):
    augmented_spec = augmentor(input_spec=spec, length=lengths)
    if device == "cuda":
        torch.cuda.synchronize()
stop = time.time()
mean_duration = (stop - start) / num_iterations
print(f"Mean duration for new implementation: {mean_duration * 1000:0.1f} ms")

# Re-instantiate a spec aug with same RNG
augmentor: SpecAugment = SpecAugment(
    freq_masks=2, time_masks=10, freq_width=27, time_width=0.05, rng=random.Random(seed)
).to(device)
# Set class's forward method to original one
setattr(SpecAugment, "forward", forward_original)
# Get baseline output. Note that the original implementation modifies in-place and returns a reference to `spec`, so
# use a clone of `spec` since we'll call forward some more for getting times
first_output_old = augmentor(input_spec=spec.clone(), length=lengths)

# Warm up
augmented_spec = augmentor(input_spec=spec, length=lengths)
# Loop and record time
start = time.time()
for _ in range(num_iterations):
    augmented_spec = augmentor(input_spec=spec, length=lengths)
    if device == "cuda":
        torch.cuda.synchronize()
stop = time.time()
mean_duration = (stop - start) / num_iterations
print(f"Mean duration for original specaug: {mean_duration * 1000:0.1f} ms")

# Compare the output of each module. Can use hard tolerance since op is a constant fill.
print("Asserting old and new augmentations are all close...")
torch.testing.assert_close(actual=first_output_new, expected=first_output_old, rtol=0.0, atol=0.0)
print("Old and new outputs match")

```

</details>

Expected output:

```text
# With GPU
Testing with input features shape torch.Size([128, 80, 1000])
Mean duration for new implementation: 5.0 ms
Mean duration for original specaug: 100.1 ms
Asserting old and new augmentations are all close...
Old and new outputs match

# with CPU
Testing with input features shape torch.Size([128, 80, 1000])
Mean duration for new implementation: 8.5 ms
Mean duration for original specaug: 42.5 ms
Asserting old and new augmentations are all close...
Old and new outputs match
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/parts/submodules/spectr_augment.py
+++ b/nemo/collections/asr/parts/submodules/spectr_augment.py
@@ -44,15 +44,15 @@ class SpecAugment(nn.Module, Typing):
         """Returns definitions of module input types
         """
         return {
-            "input_spec": NeuralType(("B", "D", "T"), SpectrogramType()),
-            "length": NeuralType(tuple("B"), LengthsType()),
+            "input_spec": NeuralType(('B', 'D', 'T'), SpectrogramType()),
+            "length": NeuralType(tuple('B'), LengthsType()),
         }
 
     @property
     def output_types(self):
         """Returns definitions of module output types
         """
-        return {"augmented_spec": NeuralType(("B", "D", "T"), SpectrogramType())}
+        return {"augmented_spec": NeuralType(('B', 'D', 'T'), SpectrogramType())}
 
     def __init__(
         self, freq_masks=0, time_masks=0, freq_width=10, time_width=10, rng=None, mask_value=0.0,
@@ -82,7 +82,7 @@ class SpecAugment(nn.Module, Typing):
     def forward(self, input_spec, length):
         batch_size, num_freq_bins, _ = input_spec.shape
         # Move lengths to CPU before repeated indexing
-        length = length.cpu().numpy()
+        lengths_cpu = length.cpu().numpy()
         # Generate a numpy boolean mask. `True` elements represent where the input spec will be augmented.
         fill_mask: np.array = np.full(shape=input_spec.shape, fill_value=False)
         freq_start_upper_bound = num_freq_bins - self.freq_width
@@ -96,10 +96,10 @@ class SpecAugment(nn.Module, Typing):
 
             # Derive time width, sometimes based percentage of input length.
             if self.adaptive_temporal_width:
-                time_max_width = max(1, int(length[idx] * self.time_width))
+                time_max_width = max(1, int(lengths_cpu[idx] * self.time_width))
             else:
                 time_max_width = self.time_width
-            time_start_upper_bound = max(1, length[idx] - time_max_width)
+            time_start_upper_bound = max(1, lengths_cpu[idx] - time_max_width)
 
             # Set time masking
             for _ in range(self.time_masks):
@@ -107,7 +107,9 @@ class SpecAugment(nn.Module, Typing):
                 width = self._rng.randint(0, time_max_width)
                 fill_mask[idx, :, start : start + width] = True
         # Bring the mask to device and fill spec
-        return input_spec.masked_fill(mask=torch.tensor(fill_mask, device=input_spec.device), value=self.mask_value)
+        fill_mask = torch.from_numpy(fill_mask).to(input_spec.device)
+        masked_spec = input_spec.masked_fill(mask=fill_mask, value=self.mask_value)
+        return masked_spec
 
 
 class SpecCutout(nn.Module, Typing):
@@ -125,13 +127,13 @@ class SpecCutout(nn.Module, Typing):
     def input_types(self):
         """Returns definitions of module input types
         """
-        return {"input_spec": NeuralType(("B", "D", "T"), SpectrogramType())}
+        return {"input_spec": NeuralType(('B', 'D', 'T'), SpectrogramType())}
 
     @property
     def output_types(self):
         """Returns definitions of module output types
         """
-        return {"augmented_spec": NeuralType(("B", "D", "T"), SpectrogramType())}
+        return {"augmented_spec": NeuralType(('B', 'D', 'T'), SpectrogramType())}
 
     def __init__(self, rect_masks=0, rect_time=5, rect_freq=20, rng=None):
         super(SpecCutout, self).__init__()
--- a/nemo/collections/asr/parts/submodules/spectr_augment.py
+++ b/nemo/collections/asr/parts/submodules/spectr_augment.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import numpy as np
 import random
+
+import numpy as np
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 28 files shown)


nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -44,15 +45,15 @@ def input_types(self):
         """Returns definitions of module input types
         """
         return {
-            "input_spec": NeuralType(('B', 'D', 'T'), SpectrogramType()),
```
titu1994, line 47: Could you revert these changes ?

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -80,29 +81,34 @@ def __init__(
...
-                y_left = self._rng.randint(0, max(1, length[idx] - time_width))
-
-                w = self._rng.randint(0, time_width)
-
-                input_spec[idx, :, y_left : y_left + w] = self.mask_value
-
-        return input_spec
+        batch_size, num_freq_bins, _ = input_spec.shape
+        # Move lengths to CPU before repeated indexing
+        length = length.cpu().numpy()
```
titu1994, line 86: Name this var _cpu and don't override device of original length.

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -80,29 +81,34 @@ def __init__(
...
+                time_max_width = self.time_width
+            time_start_upper_bound = max(1, length[idx] - time_max_width)
+
+            # Set time masking
+            for _ in range(self.time_masks):
+                start = self._rng.randint(0, time_start_upper_bound)
+                width = self._rng.randint(0, time_max_width)
+                fill_mask[idx, :, start : start + width] = True
+        # Bring the mask to device and fill spec
+        return input_spec.masked_fill(mask=torch.tensor(fill_mask, device=input_spec.device), value=self.mask_value)
```
titu1994, line 111: Put return seperate lines.

titu1994: Plus use torch.from_numpy() and then do .to(device=device)

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -120,13 +126,13 @@ class SpecCutout(nn.Module, Typing):
     def input_types(self):
         """Returns definitions of module input types
         """
-        return {"input_spec": NeuralType(('B', 'D', 'T'), SpectrogramType())}
+        return {"input_spec": NeuralType(("B", "D", "T"), SpectrogramType())}
```
titu1994, line 129: Revert below.

titu1994 (commented):
Very cool pr ! The speedup pushes this on par with the Numba method (Numba is maybe a few milliseconds faster but it doesn't matter for CPU use and the speedup is substantial for GPU anymore) 

Pr looks good, but I'll ask @VahidooX for final review

titu1994 (commented):
Thanks !

nemo/collections/asr/parts/submodules/spectr_augment.py
1-800-BAD-CODE: Done... this was `black`s default (string normalization) that was done automatically

nemo/collections/asr/parts/submodules/spectr_augment.py
1-800-BAD-CODE: Renamed. The reason it's pulled to CPU is because it's accessed `2 * batch_size` times which is very slow on GPU.

nemo/collections/asr/parts/submodules/spectr_augment.py
1-800-BAD-CODE: done

nemo/collections/asr/parts/submodules/spectr_augment.py
1-800-BAD-CODE: done

nemo/collections/asr/parts/submodules/spectr_augment.py
titu1994: Yup makes sense

VahidooX (dismissed):
LGTM, thanks for the contribution!

titu1994:
Seems like you need to sign your commits - https://github.com/NVIDIA/NeMo/pull/6347/checks?check_run_id=12461951009

raman-r4334:
Just a suggestion, please have a look  at `torchaudio.functional.mask_along_axis_iid` which I hope doing the same thing

titu1994:
Torchaudio is an optional dependency in NeMo, and not installed automatically due to its dependency on fixed pytorch version and compilation. This is a more generic PR that has no requirements other than torch.

titu1994 (approved):

titu1994:
@1-800-BAD-CODE thanks for the PR !

PR #6370, opened 2023-04-05
Update script for ngram rnnt and hat beam search decoding

andrusenkoau:
# What does this PR do ?

Update previous script for ngram RNNT/HAT beam search decoding by analogy with CTC ngram decoding.

**Collection**: [ASR]

# Changelog 
- Add option to return `return_encoder_embiddings` for `rnnt_mode.transcribe()` method (analogy of logits for CTC)
- Update `eval_beamsearch_ngram_transducer.py` by analogy with `eval_beamsearch_ngram.py`
- Add RNNT/HAT ngram decoding description to docs

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


andrusenkoau:
Hi @titu1994!
I think that it is better to have two separate scripts for CTC and RNNT ngram decoding. There is some code intersection, but the attempts to merge the scripts lead to great code complication and overflow of `if` operators.

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -212,6 +212,7 @@ def transcribe(
         self,
         paths2audio_files: List[str],
         batch_size: int = 4,
+        return_encoder_embiddings: bool = False,
```
titu1994, line 215: Revert. Setup a parameter in the RNNTDecodingConfig and set this value there. 
We have to stop adding more arguments to model.transcribe(). Decoding framework exists for this purpose. Put the thing inside Hypothesis, and use return_hypotheses=True for model.transcribe().

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,420 @@
...
+    probs_cache_file: Optional[str] = None  # The cache file for storing the logprobs of the model
+
+    # Parameters for inference
+    acoustic_batch_size: int = 128  # The batch size to calculate log probabilities
+    beam_batch_size: int = 128  # The batch size to be used for beam search decoding
+    device: str = "cuda"  # The device to load the model onto to calculate log probabilities
+    use_amp: bool = False  # Whether to use AMP if available to calculate log probabilities
+
+    # The decoding scheme to be used for evaluation
+    decoding_strategy: str = "greedy_batch" # ["greedy_batch", "maes"]
```
titu1994, line 103: Revert these args below specific to decoding strategy (why is this greedy_batch for a beam search script btw) and use the values from BeamRNNTInferConfig here - similar to how it is done in the CTC beam search example. Use that config as you need. Dont redo maes, ILM etc arguments.

The only exception is alpha, beta and beam_width - to stay consistent with CTC beam search framework.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,420 @@
...
+def decoding_step(
+    model: nemo_asr.models.ASRModel,
+    cfg: EvalBeamSearchNGramConfig,
+    all_probs: List[torch.Tensor],
+    target_transcripts: List[str],
+    preds_output_file: str = None,
+    lm_path: str = None,
+    beam_alpha: float = 0.2,
+    beam_width: int = 128,
+    maes_prefix_alpha: int = 2,
```
titu1994, line 131: Revert all these additional arguments and pass full rnntbeamdecodingconfig here.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,420 @@
...
+        else:
+            description = f"{cfg.decoding_strategy} decoding with bw={beam_width}, ba={beam_alpha}, ma={maes_prefix_alpha}, mg={maes_expansion_gamma}, hat_ilmw={hat_ilm_weight}"
+        it = tqdm(range(int(np.ceil(len(all_probs) / beam_batch_size))), desc=description, ncols=120)
+    else:
+        it = range(int(np.ceil(len(all_probs) / beam_batch_size)))
+    for batch_idx in it:
+        # disabling type checking
+        probs_batch = all_probs[batch_idx * beam_batch_size : (batch_idx + 1) * beam_batch_size]
+        probs_lens = torch.tensor([prob.shape[-1] for prob in probs_batch])
+        with torch.no_grad():
```
titu1994, line 181: Why is manual decoding being dont here? Why not use model.transcribe() which will do proper decoding with high level API?

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,420 @@
...
+            packed_batch = torch.zeros(len(probs_batch), probs_batch[0].shape[0], max(probs_lens), device='cpu')
+
+            for prob_index in range(len(probs_batch)):
+                packed_batch[prob_index, :, : probs_lens[prob_index]] = torch.tensor(
+                    probs_batch[prob_index].unsqueeze(0), device=packed_batch.device, dtype=packed_batch.dtype
+                )
+            best_hyp_batch, beams_batch = model.decoding.rnnt_decoder_predictions_tensor(
+                packed_batch, probs_lens, return_hypotheses=True,
+            )
+        if cfg.decoding_strategy == "greedy_batch":
```
titu1994, line 191: Why do we need / want to support greedy decoding in a beam search script? Drop support for it, we have transcribe_speech script for that.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,420 @@
...
+            target_transcripts=target_transcripts,
+            beam_batch_size=cfg.beam_batch_size,
+            progress_bar=True,
+        )
+        logging.info(f"Greedy batch WER/CER = {candidate_wer:.2%}/{candidate_cer:.2%}")
+
+    asr_model = asr_model.to('cpu')
+
+    # 'greedy_batch' decoding_strategy would skip the beam search decoding
+    if cfg.decoding_strategy in ["maes"]:
```
titu1994, line 344: RNNT supports all beam search inference algos, do not special case like this to maes just because of ILM. 
This is meant to be a generic inference script of RNNT models with beam search, not special cased external LM inference with HAT alone using maes only.

titu1994 (changes requested):
Code needs to be significantly refactored to generalize. We are not writing a single script for just HAT with maes support.

titu1994:
> I think that it is better to have two separate scripts for CTC and RNNT ngram decoding. There is some code intersection, but the attempts to merge the scripts lead to great code complication and overflow of if operators.

Makes sense, lets have separate CTC and RNNT beam decoding scripts. However, the RNNT beam decoding script needs to have much more generality than what this PR currently has. Lets discuss some removal of hardcoding for maes, hat, ILM and other things and make this code more generic and robust.

titu1994:
This script also needs a test in Jenkinsfile. (So does CTC but that things dependencies are difficult, at least RNNT has pytorch level beam search).

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: Only maes decoding supports ngram lm fusion for RNNT/HAT now. Do we need to have all beamsearch decoding methods in the `eval_beamsearch_ngram_transducer.py` script even if they do not support ngram lm?

titu1994: Treat Ngram as special optional case. This script should support beam search in general (all RNNT capable beam search algos) and if you're passing external kenlm then enforce that only mAES is currently supported

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: Ok, I will try to do it in this way

andrusenkoau: "Manual decoding" with `model.decoding.rnnt_decoder_predictions_tensor()` is used here because we want to save time on rnnt encoder embeddings calculation during search of the best decoding parameters (alpha, ilm_weight, etc.). As far as I understand, with `model.transcribe()` we will calculate the whole rnnt pipeline from the beginning for all searching parameters. I took this code from the reference `eval_beamsearch_ngram.py` script for CTC. Thats why I added `return_encoder_embiddings` option to `rnnt.transcribe()`.

titu1994: Don't add more arguments to transcribe. Find another way of doing this. Hypothesis is fine for this, use that since you are anyway calling the rnnt_decoding class.

nemo/collections/asr/models/rnnt_models.py
andrusenkoau: I moved `return_encoder_embiddings` parameter to `RNNTDecodingConfig`.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: Done

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: `return_encoder_embiddings` parameter was moved to `RNNTDecodingConfig`.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: I added support for all the RNNT decoding algorithms.

andrusenkoau:
Hi @titu1994!
I tried to fix the code according to you suggestions. `return_encoder_embiddings` parameter was moved from `rnnt.trascribe()`  to `RNNTDecodingConfig`. `eval_beamsearch_ngram_transducer.py` also was generalized. It supports all the rnnt decoding methods now (but ngram fusion is enable only for maes). Could you check the changes? 

About test in Jenkinsfile -- I did not find any test examples for /NeMo/scripts/* files. Do we need to write a test for `eval_beamsearch_ngram_transducer.py` script?

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -242,6 +242,10 @@ def transcribe(
         """
         if paths2audio_files is None or len(paths2audio_files) == 0:
             return {}
+        if return_hypotheses and self.cfg.decoding.return_encoder_embeddings:
```
titu1994, line 245: Wait, why is this being done as such ? 
The hypothesis object can hold any tensor. Make a new value like "encoder_embedding" : Optional[torch Tensor] = None. 

In greedy and beam decoding scripts, you use this new flag - rename to "preserve_encoder_embedding" - to check if encoder embeddings should be preserved or not. 

If yes, inside the forward() pass of each of the beam, greedy and batch greedy decoding classes, you slice off the appropriate encoder embedding for each sample, call detach and CPU() and then set them into the hypothesis object. 

Then user just sets this flag and does return_hypotheses = True - and the hypothesis object will contain the actual encoder embedding for that sample.

andrusenkoau: Hi!
I am trying to implement the same functionality as in `ctc_models.trascribe()` -- it returns a list of decoded transcriptions or only raw logprobs from `ctc_encoder` (without decoded transcriptions -- https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/models/ctc_models.py#L202).

Raw logprobs are used for search of decoder hyper parameters by calling `model.decoding.ctc_decoder_predictions_tensor` method in the search loop. 

Do I understand correctly what your suggestion with `preserve_encoder_embedding` will do all the rnnt decoding pipeline (including getting decoded transcripts) and save encoder_embeddings additionally? I wanted to avoid transcripts decoding and return encoder_embeddings only (to save time) for later hyper parameters search by `model.decoding.rnnt_decoder_predictions_tensor`.

titu1994: The logprobs flag in CTC models transcribe is legacy from before we had Hypothesis object. These days you are meant to request return_hypotheses true in order to get the hypothesis object with the "alignment" field containing the logprobs. We won't be adding an equivalent field in RNNT because we want to follow the proper way of using decoding strategy and hypothesis object. 

In RNNT, alignment is decoder output - which we don't need. So for this script you can manually calculate the encoder forward pass and skip the decoder and joint for caching step, then use decoding . Rnnt_docoder_prediction_tensod from the cached inputs in two steps. No need to add flags then

andrusenkoau: Ok, I will try to avoid the call of `rnnt_model.transcribe()` for encoder_embeddings calculation. I will do it manually inside the `eval_beamsearch_ngram_transducer.py` script.

titu1994: Thanks, this is good enough for now. Let's discuss offline for a better design so that we can reuse transcribe if possible

titu1994 (approved):
This is fantastic work, thank you ! @VahidooX for last review. 

For the RNNT duplication of transcribe, let's discuss offline some better way to do this in the future

andrusenkoau:
@titu1994, thank you for the extensive review. Is the PR ready to merge or do we need @VahidooX approval?

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
```diff
@@ -13,257 +13,443 @@
...
+    beam_batch_size: int = 128  # The batch size to be used for beam search decoding
+    device: str = "cuda"  # The device to load the model onto to calculate log probabilities
+    use_amp: bool = False  # Whether to use AMP if available to calculate log probabilities
+    num_workers: int = 1  # Number of workers for DataLoader
+
+    # The decoding scheme to be used for evaluation
+    decoding_strategy: str = "greedy_batch" # ["greedy_batch", "beam", "tsd", "alsd", "maes"]
+
+    # Beam Search hyperparameters
+    beam_width: List[int] = field(default_factory=lambda: [128])  # The width or list of the widths for the beam search decoding
```
titu1994, line 108: RNNT beam size should be 8

titu1994 (approved):
Lgtm, @VahidooX for final review

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram_transducer.py
andrusenkoau: Done

docs/source/asr/asr_language_modeling.rst
```diff
@@ -7,7 +7,8 @@ Language models have shown to help the accuracy of ASR models. NeMo supports the
 *  :ref:`ngram_modeling`
 *  :ref:`neural_rescoring`
 
-It is possible to use both approaches on the same ASR model.
+It is possible to use both approaches on the same CTC ASR model.
+RNNT and Hybrid Autoregressive Transducer (HAT) models support only N-gram language modeling.
```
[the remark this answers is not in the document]
andrusenkoau: Ohh, you are absolutely right. I forgot about rescoring of n-best results with NNLM. I meant that RNNT/HAT models do not support NNLM in shallow fusion.

titu1994 (approved):

PR #6389, opened 2023-04-07
Add [Draft][TTS] FastPitch multi-speaker pre-train and adapter fine-tune

hsiehjackson:
# What does this PR do ?

Adds FastPitch speaker adaptation with adapters.

**Collection**: TTS

# Changelog 
- Adds multi-speaker FastPitch pre-training with 1) Looked-up speaker embedding 2) Speaker verification model speaker embedding 3) Global Style Tokens 4) Conditional Layer Normalization 
    - `nemo/collections/tts/data/tts_dataset.py`
    - `nemo/collections/tts/torch/tts_data_types.py`
    - `nemo/collections/tts/models/fastpitch.py`
    - `nemo/collections/tts/modules/fastpitch.py`
    - `nemo/collections/tts/modules/transformer.py`
    - `nemo/collections/tts/modules/submodules.py`
- Add adapter modules for FastPitch fine-tuning
    - `nemo/collections/tts/models/fastpitch.py`
    - `nemo/collections/tts/modules/fastpitch.py`
    - `nemo/collections/tts/modules/transformer.py`
    - `nemo/collections/tts/modules/aligner.py`
    - `nemo/collections/tts/parts/mixins/__init__.py`
    - `nemo/collections/tts/parts/mixins/fastpitch_adapter_mixins.py`
- Add config and fine-tuning python script
    - `examples/tts/conf/fastpitch_speaker_adaptation.yaml`
    - `examples/tts/fastpitch_finetune_adapters.py`
- Add tutorial
    - `tutorials/tts/FastPitch_SpeakerAdaptation.ipynb`
- Add fast text normalization script
    - `scripts/dataset_processing/tts/add_normalized_text.py`
- Fix aligner `nan` loss bug
    - `nemo/collections/tts/losses/aligner_loss.py`
 
# Usage
* Follow `tutorials/tts/FastPitch_SpeakerAdaptation.ipynb`
    * Pre-train multi-speaker FastPitch
    * Fine-tune HiFiGAN with mel-spectrograms from pre-trained FastPitch
    * Fine-tune FastPitch with adapters
    * Fine-tune HiFiFAN with mel-spectrograms from fine-tuned FastPitch

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/tts/torch/tts_data_types.py
```diff
@@ -67,6 +67,14 @@ class LMTokens(TTSDataType):
     name = "lm_tokens"
 
 
+class GSTRefAudio(TTSDataType, WithLens):
```
subhankar-ghosh, line 70: Let's rename this to `ReferenceAudio` and name = "reference_audio". So, we do not couple this with GST only.

nemo/collections/tts/data/dataset.py
```diff
@@ -453,6 +455,11 @@ def add_pitch(self, **kwargs):
             with open(Path(pitch_stats_path), 'r', encoding="utf-8") as pitch_f:
                 self.pitch_stats = json.load(pitch_f)
 
+    def add_speaker_embedding(self, **kwargs):
```
subhankar-ghosh, line 458: Add a comment/method description as to what kind of speaker embedding will be imported, any specific format that is expected.

nemo/collections/tts/data/dataset.py
```diff
@@ -483,6 +490,15 @@ def add_energy(self, **kwargs):
     def add_speaker_id(self, **kwargs):
         pass
 
+    def add_gst_ref_audio(self, **kwargs):
```
subhankar-ghosh, line 493: add_gst_ref_audio -> add_reference_audio

nemo/collections/tts/data/dataset.py
```diff
@@ -483,6 +490,15 @@ def add_energy(self, **kwargs):
...
+        if SpeakerID not in self.sup_data_types_set:
+            self.speaker_to_index = None
```
subhankar-ghosh, line 495: `speaker_to_index_map`

nemo/collections/tts/data/dataset.py
subhankar-ghosh: Add comment/description of what this method is doing, what does self.speaker_to_index represent.

nemo/collections/tts/data/dataset.py
```diff
@@ -683,6 +697,39 @@ def __getitem__(self, index):
         if SpeakerID in self.sup_data_types_set:
             speaker_id = torch.tensor(sample["speaker_id"]).long()
 
+        gst_ref_audio, gst_ref_audio_length = None, None
+        if GSTRefAudio in self.sup_data_types_set:
+            ref_sample = sample
```
subhankar-ghosh, line 702: reference_sample

nemo/collections/tts/data/dataset.py
```diff
@@ -683,6 +697,39 @@ def __getitem__(self, index):
...
+            if self.speaker_to_index is not None:
+                if len(self.speaker_to_index[sample["speaker_id"]]) > 1:
+                    ref_pool = self.speaker_to_index[sample["speaker_id"]] - set([index])
```
subhankar-ghosh, line 705: reference_pool

nemo/collections/tts/data/dataset.py
```diff
@@ -483,6 +490,15 @@ def add_energy(self, **kwargs):
...
         pass
 
+    def add_gst_ref_audio(self, **kwargs):
+        if SpeakerID not in self.sup_data_types_set:
+            self.speaker_to_index = None
+        else:
+            self.speaker_to_index = defaultdict(list)
+            for i, d in enumerate(self.data):
+                self.speaker_to_index[d.get('speaker_id', None)].append(i)
+            self.speaker_to_index = {k: set(v) for k, v in self.speaker_to_index.items()}
```
subhankar-ghosh, line 500: Make the defaultdict values as set and not list so we do not have to make another pass to convert list to set.

nemo/collections/tts/data/dataset.py
```diff
@@ -683,6 +697,39 @@ def __getitem__(self, index):
...
+            ref_features = self.featurizer.process(
+                ref_sample["audio_filepath"],
+                trim=self.trim,
+                trim_ref=self.trim_ref,
+                trim_top_db=self.trim_top_db,
+                trim_frame_length=self.trim_frame_length,
+                trim_hop_length=self.trim_hop_length,
+            )
+
+            gst_ref_audio, gst_ref_audio_length = ref_features, torch.tensor(ref_features.shape[0]).long()
```
subhankar-ghosh, line 719: rename gst_* to reference_*

nemo/collections/tts/data/dataset.py
```diff
@@ -683,6 +697,39 @@ def __getitem__(self, index):
...
+
+        speaker_emb = None
+        if SpeakerEmbedding in self.sup_data_types_set:
+            if self.speaker_embeddings is not None:
+                ref_sample_index = index
+                if self.speaker_to_index is not None:
+                    if len(self.speaker_to_index[sample["speaker_id"]]) > 1:
```
subhankar-ghosh, line 726: This logic needs to be converted into a method as it is being called again in GST feature extraction above.

nemo/collections/tts/data/dataset.py
```diff
@@ -765,18 +819,9 @@ def general_collate_fn(self, batch):
...
-            [],
-            [],
-            [],
-            [],
-            [],
-            [],
-        )
+            gst_ref_audios,
+            speaker_embs,
+        ) = ([], [], [], [], [], [], [], [], [], [], [], [])
```
subhankar-ghosh, line 824: keep the code format same.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -245,17 +313,27 @@ def forward(
 
         # Calculate speaker embedding
         if self.speaker_emb is None or speaker is None:
-            spk_emb = 0
+            spk_emb = None
         else:
             spk_emb = self.speaker_emb(speaker).unsqueeze(1)
 
+        if self.speaker_encoder is not None:
+            spk_emb = self.speaker_encoder(
```
subhankar-ghosh, line 321: @rlangman @redoctopus  We need your opinion on this. Currently speakers are represented as lookup table embeddings (Line 318), In Adapters project we represent speakers as GST Embeddings + Titanet embeddings + lookup table embeddings. Line 321 `self.speaker_encoder` is the class to represent speakers, this has GST, Titanet representation. We were thinking if we should include lookup table also in this speaker_encoder class https://github.com/NVIDIA/NeMo/blob/385d46acc15c7a48c25db9e8e1f0ccafb17fb2de/nemo/collections/tts/modules/submodules.py#L613. This will lead to just one speaker representation class, which will be much neater than having separate classes/methods and then somehow adding them.

This has only one down side -> the use of older models which do not use the SpeakerEncoder class to represent speakers but the lookup table was a part of the FastPitch class. To solve this problem, we either upload new versions of the model with SpeakerEncoder class to NGC or write a wrapper that will put the lookup table from older model checkpoints into the SpeakerEncoder class.

What do you think?

subhankar-ghosh: We can raise a separate PR for this, I would like to not make this PR more complicated.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -121,14 +137,57 @@ def output_types(self):
...
+        for layer in self.layers:
+            out = layer(out, conditioning=conditioning)
+
+        out = out.transpose(1, 2)
         out = self.fc(out) * enc_mask
         return out.squeeze(-1)
 
 
-class FastPitchModule(NeuralModule):
+class TemporalPredictorAdapter(TemporalPredictor, adapter_mixins.AdapterModuleMixin):
```
subhankar-ghosh, line 155: Add comments for each method, what they are doing and logic.

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -42,10 +42,15 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+from typing import List, Optional
+
 import torch
+from omegaconf import DictConfig
 
+from nemo.collections.asr.parts.utils import adapter_utils
+from nemo.collections.tts.modules.transformer import _LayerNorm
```
subhankar-ghosh, line 51: Change name _LayerNorm to ConditionalLayerNorm and move the function to submodules

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -204,6 +266,9 @@ def input_types(self):
             "attn_prior": NeuralType(('B', 'T_spec', 'T_text'), ProbsType(), optional=True),
             "mel_lens": NeuralType(('B'), LengthsType(), optional=True),
             "input_lens": NeuralType(('B'), LengthsType(), optional=True),
+            "gst_ref_spec": NeuralType(('B', 'D', 'T_spec'), MelSpectrogramType(), optional=True),
```
subhankar-ghosh, line 269: change names from gst_* to reference_*

nemo/collections/tts/models/fastpitch.py
```diff
@@ -19,13 +19,14 @@
 from hydra.utils import instantiate
 from omegaconf import DictConfig, OmegaConf, open_dict
 from pytorch_lightning import Trainer
-from pytorch_lightning.loggers import TensorBoardLogger
+from pytorch_lightning.loggers import TensorBoardLogger, WandbLogger
```
subhankar-ghosh, line 22: I think WandB logger can be a separate PR and needs more discussion.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -405,34 +470,54 @@ def training_step(self, batch, batch_idx):
...
+            else {
+                "loss": loss,
+                "mel_target": mels[0].data.cpu().float().numpy(),
+                "mel_predict": mels_pred[0].data.cpu().float().numpy(),
+                "attn_hard": attn_hard[0].data.cpu().float().numpy().squeeze().T if self.learn_alignment else None,
+                "attn_soft": attn_soft[0].data.cpu().float().numpy().squeeze().T if self.learn_alignment else None,
+            }
+        )
 
-            self.tb_logger.add_image(
```
subhankar-ghosh, line 412: Let's keep this logic as it is for now, we can have different PR for this.

examples/tts/fastpitch_finetune_adapters.py
```diff
@@ -0,0 +1,166 @@
...
+    model.freeze()
+
+    # Add weighted speaker embedding module in speaker representation
+    if add_weight_speaker and model.fastpitch.speaker_emb is not None:
+        old_emb = model.fastpitch.speaker_emb
+        new_emb = WeightedSpeakerEmbedding(pretrained_embedding=old_emb, speaker_list=add_weight_speaker_list)
+        model.fastpitch.speaker_emb = new_emb
+        model.fastpitch.speaker_emb.embedding_weight.requires_grad = True
+
+    ############ Current experiment - unfreeze gst and titanet embedding
```
subhankar-ghosh, line 120: Remove this comment.

scripts/dataset_processing/tts/add_normalized_text.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
rlangman, line 1: I will have a more streamlined script for this soon as part of the new preprocessing pipeline.

examples/tts/conf/fastpitch_speaker_adaptation.yaml
```diff
@@ -0,0 +1,307 @@
+# This config contains the default values for training FastPitch speaker adaptation
+# If you want to train model on other dataset, you can change config values according to your dataset.
+# Most dataset-specific arguments are in the head of the config file, see below.
+
+name: FastPitch
+
+train_dataset: ???
+validation_datasets: ???
+sup_data_path: ???
+sup_data_types: [ "align_prior_matrix", "pitch" ]
```
rlangman, line 10: Does this need SpeakerEmbedding and ReferenceAudio as a sup data type?

nemo/collections/tts/data/dataset.py
```diff
@@ -453,6 +456,13 @@ def add_pitch(self, **kwargs):
             with open(Path(pitch_stats_path), 'r', encoding="utf-8") as pitch_f:
                 self.pitch_stats = json.load(pitch_f)
 
+    # Add pretrained speaker embeddings (e.g. speaker-verification embeddings)
```
rlangman, line 459: Are we using titanet embeddings? Do we have a script to compute the speaker embeddings and store them like other data types in sup_data?

nemo/collections/tts/data/dataset.py
```diff
@@ -453,6 +456,13 @@ def add_pitch(self, **kwargs):
...
+    # The embedding order should be the same as manifest order
+    def add_speaker_embedding(self, **kwargs):
+        embedding_path = kwargs.pop("speaker_embedding_path", None)
+        assert embedding_path is not None, "Speaker embedding path is required but got None."
+        self.speaker_embeddings = torch.from_numpy(np.load(embedding_path))
```
rlangman, line 464: It sounds like you have a single file with all embeddings in it. Should we not follow the same convention with other sup_data and have each embedding in a separate .pt file?

nemo/collections/tts/data/dataset.py
```diff
@@ -683,6 +712,26 @@ def __getitem__(self, index):
         if SpeakerID in self.sup_data_types_set:
             speaker_id = torch.tensor(sample["speaker_id"]).long()
 
+        reference_audio, reference_audio_length = None, None
+        if ReferenceAudio in self.sup_data_types_set:
+            if self.speaker_to_index_map is not None:
+                reference_index = self.sample_reference_index(index, n=1)[0]
+                reference_audio = self.featurizer.process(
+                    self.data[reference_index]["audio_filepath"],
```
rlangman, line 720: I can understand the rationale for wanting to sample the speaker embeddings and reference audios from the non-training sample, but it is pretty unorthodox. Having training examples sample from other records in the manifest is going to be complicated to maintain unless the data processor is designed specifically for the use case.

In comparison, using the embeddings from the training utterance and stopping overfiting by using a bottleneck like VAE/VQ is a lot more standard and scalable. But might not work well on smaller datasets without lots of tuning.

I am curious what other's thoughts are.

examples/tts/fastpitch_finetune_adapters.py
```diff
@@ -0,0 +1,159 @@
...
+from nemo.core import adapter_mixins
+from nemo.core.config import hydra_runner
+from nemo.utils import logging
+from nemo.utils.exp_manager import exp_manager
+
+
+def update_model_config_to_support_adapter(config) -> DictConfig:
+    with open_dict(config):
+        enc_adapter_metadata = adapter_mixins.get_registered_adapter(config.input_fft._target_)
+        if enc_adapter_metadata is not None:
```
rlangman, line 34: We should think about whether there is a design that will let us use adapters without having to hardcode every layer in every model that needs to be adapted or frozen.

rlangman (commented):
I know this is a draft, but we had similar PR before and the first main comment was that the PR can be broken into independent commits/features:

1. Adapter finetuning
2. Reference audio handling and GST
3. External speaker embeddings/titanet support
4. Auxilliary speaker encoder system
5. Tutorial(s)

nemo/collections/tts/modules/fastpitch.py
rlangman: After thinking about it, I don't think we should merge the speaker embeddings with the speaker encoder. The speaker encoder is very specific to these experiments with custom voice, and unlikely to be used by most other TTS problems.

But now that we are supporting utterance level titanet embeddings, we should really support all 3 workflows:

1. Condition only on lookup table.
2. Condition only on utterance level embeddings
3. Condition on both together using speaker encoder.

All 3 could be represented as instances of SpeakerEncoder, but just having a branching if-statement will be clearer.

rlangman: So how about we add a configuration for whether to learn the embedding (defaulting to True for backwards compatibility):

        if self.learn_speaker_embedding and n_speakers > 1:
            self.speaker_emb = torch.nn.Embedding(n_speakers, symbols_embedding_dim)
        else:
            self.speaker_emb = None

Then you can manually set it to false when you provide the speaker encoder. Then we can cleanly support all 3 common training workflows.

```
def get_speaker_embedding(self, speaker, reference_speaker_embedding, reference_spec, reference_spec_lens):
  if self.speaker_encoder:
    spk_emb = self.speaker_encoder(speaker=speaker, ...)
  elif self.speaker_emb:
    if speaker is None:
      raise ValueError()
    spk_emb = self.speaker_emb(speaker)
  elif reference_speaker_embedding:
    spk_emb = reference_speaker_embedding
  else:
    spk_emb = None

  return spk_emb
```

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+    """
+    Global Style Token based Speaker Embedding
+    """
+
+    def __init__(
+        self,
+        cnn_filters=[32, 32, 64, 64, 128, 128],
+        dropout=0.2,
+        gru_hidden=128,
+        gst_size=128,
```
rlangman, line 452: Should we pass in the ReferenceEncoder directly to __init__?

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -237,6 +302,9 @@ def forward(
         attn_prior=None,
         mel_lens=None,
         input_lens=None,
+        reference_spec=None,
+        reference_spec_lens=None,
+        speaker_embedding=None,
```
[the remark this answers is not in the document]
hsiehjackson: Sure! Good idea!

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+    def forward(self, inputs, inputs_masks):
+        inputs = inputs.transpose(1, 2)
+
+        inputs = inputs * inputs_masks
+        out = inputs.unsqueeze(3)
+        out = self.conv(out)
+        out = out.view(out.shape[0], out.shape[1], -1).contiguous()
+        self.gru.flatten_parameters()
+        _, out = self.gru(out)
+
```
rlangman, line 544: I had this comment on the previous adapter CR last year:

Masking is required in between CNN layers (In theory, maybe not always in practice. Definitely required for final GRU layer). This is relatively complicated with reference encoder due to (2, 2) stride in CNN cutting sequence lengths in half at each step.

I suggest copying implementation from this or other github repo, use it to create mask after each CNN layer, and ideally adding a unit test to it: https://github.com/mozilla/TTS/blob/e9e07844b77a43fb0864354791fb4cf72ffded11/TTS/tts/layers/gst_layers.py#L84

rlangman: Main concern is not masking RNN input is a large bug likely to impact performance, and relatively easy to fix. Masking in between each CNN is less important and more difficult to implement.

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        self.token_size = self.output_size // self.n_head
+
+        self.tokens = torch.nn.Parameter(torch.FloatTensor(self.n_token, self.token_size))
+
+        self.q_linear = torch.nn.Linear(self.input_size, self.output_size)
+        self.k_linear = torch.nn.Linear(self.token_size, self.output_size)
+        self.v_linear = torch.nn.Linear(self.token_size, self.output_size)
+
+        self.tanh = torch.nn.Tanh()
+        self.softmax = torch.nn.Softmax(dim=2)
```
rlangman, line 564: Can we use https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html ?

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        style_emb = style_emb.permute(1, 0, 2).contiguous().view(bs, -1)
+
+        return style_emb
+
+
+class SVEmbeddingLinearProjectionNetwork(torch.nn.Module):
+    def __init__(self, speaker_embedding_dim, symbols_embedding_dim):
+        super(SVEmbeddingLinearProjectionNetwork, self).__init__()
+        self.speaker_proj = torch.nn.Linear(speaker_embedding_dim, symbols_embedding_dim)
+        self.act_fn = torch.nn.ReLU()
```
rlangman, line 609: Should we have a more generic LinearActivation class which takes (input_dim, output_dim, activation)?

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+        Constructor.
+        Parameters:
+            - gst_module: Neural module to get GST based speaker embedding
+            - sv_projection_module: Neural module to get Speaker Verification based speaker embedding
+            - lookup_emb_projection_module: Neural module to project lookup speaker embedding
+        """
+        super(SpeakerEncoder, self).__init__()
+        self.gst_module = gst_module
+        self.sv_projection_module = sv_projection_module
+        self.lookup_emb_projection_module = lookup_emb_projection_module
```
rlangman, line 634: `lookup_emb_projection_module` looks unused with example config. Do we need it?

nemo/collections/tts/modules/submodules.py
rlangman: Is it a good idea to use a ReLU activation on a residual? Other architectures like ResNet and transformer can do it because they apply batch norm or layer norm afterwards.

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+
+    def forward(self, spk_emb, reference_spec=None, reference_spec_lens=None, speaker_embedding=None):
+        x = spk_emb
+        # Project lookup speaker embedding
+        if self.lookup_emb_projection_module is not None:
+            x += self.lookup_emb_projection_module(spk_emb)
+
+        # Get GST based speaker embedding
+        if self.gst_module is not None and reference_spec is not None and reference_spec_lens is not None:
```
rlangman, line 643: Should we throw an error if gst_module is not None but reference spec is None?

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+            ref_spec_mask = (
+                torch.arange(reference_spec_lens.max())
+                .to(reference_spec.device)
+                .expand(reference_spec_lens.shape[0], reference_spec_lens.max())
+                < reference_spec_lens.unsqueeze(1)
+            ).unsqueeze(2)
+            x += self.gst_module(reference_spec, ref_spec_mask).unsqueeze(1)
+
+        # Project speaker embedding from Speaker Verification based
+        if self.sv_projection_module is not None and speaker_embedding is not None:
```
rlangman, line 653: Should we throw an error if sv_projection_module is not None but speaker_embedding is None?

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+            x += self.sv_projection_module(speaker_embedding)
+        return x
+
+
+class Transpose(torch.nn.Module):
```
rlangman, line 658: Why is this required over the normal transpose/permute operation?

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+                                stride=(2, 2),
+                                padding=(1, 1),
+                                dilation=(1, 1),
+                                bias=True,
+                            ),
+                        ),
+                        ("transpose_bhwc_{}".format(i + 1), Transpose([0, 2, 3, 1])),
+                        ("relu_{}".format(i + 1), torch.nn.ReLU()),
+                        ("layer_norm_{}".format(i + 1), torch.nn.LayerNorm(self.filter_size[i + 1]),),
+                        ("dropout_{}".format(i + 1), torch.nn.Dropout(self.dropout)),
```
rlangman, line 514: Would be convenient here and in the future if we have a generic Conv2dNorm layer that handles the awkward transposes and makes the activation/layer_norm optional.

rlangman: Binary dropout is basically designed to be used on relu activation output (effectively setting active neurons to inactive). Applying it after layer norm rather than after relu is strange.

nemo/collections/tts/modules/submodules.py
```diff
@@ -410,3 +419,284 @@ def forward(self, forward_input: Tuple[torch.Tensor, torch.Tensor]):
...
+
+class ReferenceEncoder(NeuralModule):
+    """
+    Encode mel-spectrograms to an utterance level feature
+    """
+
+    def __init__(self, cnn_filters=[32, 32, 64, 64, 128, 128], dropout=0.2, gru_hidden=128):
+        super(ReferenceEncoder, self).__init__()
+        self.filter_size = [1] + cnn_filters
+        self.dropout = dropout
```
rlangman, line 491: Nitpick: no need to save dropout as a class field here

nemo/collections/tts/modules/fastpitch.py
```diff
@@ -155,13 +215,15 @@ def __init__(
...
         self.aligner = aligner
+        self.speaker_encoder = speaker_encoder
         self.learn_alignment = aligner is not None
         self.use_duration_predictor = True
         self.binarize = False
         self.speaker_emb_condition_prosody = speaker_emb_condition_prosody
         self.speaker_emb_condition_decoder = speaker_emb_condition_decoder
         self.speaker_emb_condition_aligner = speaker_emb_condition_aligner
 
+        # TODO: combine self.speaker_emb with self.speaker_encoder
```
redoctopus, line 226: What's the plan for this TODO? Any timeline in mind?

nemo/collections/tts/modules/fastpitch.py
redoctopus: I'm in favor of including the lookup table in the speaker encoder class if it makes things simpler and improves the abstraction/organization. A little pain in re-training in the short term is okay!

redoctopus (commented):
Some initial comments

examples/tts/conf/fastpitch_speaker_adaptation.yaml
subhankar-ghosh: @hsiehjackson let's add reference audio and speaker embedding as ssup data type to the config.

nemo/collections/tts/modules/submodules.py
hsiehjackson: Masks between CNN layers and before RNN are added

nemo/collections/tts/modules/submodules.py
hsiehjackson: The reason is I used this module in `nn.Sequential`. Now, I change the implementation and remove this module.

nemo/collections/tts/modules/submodules.py
hsiehjackson: I think the parameters should be called `scale` instead of `temperature`. And It looks like `nn.MultiheadAttention` do the scaling inside in these two parts [line1](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L5294) and [line2](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L5348). Let me take a look at how to map my pre-trained weights to `nn.MultiheadAttention`. If it has some mismatch, I'll change this maybe for next PR.

nemo/collections/tts/data/dataset.py
subhankar-ghosh: What kind of complications do you think we will have in maintaining this random sampling code? I agree with your suggestion of using VAE/VQ as a projection layer on top of the reference/GT audio/embedding and that can be incorporated into the SpeakerEncoder class. But here this might be useful for the current use-case as well as other use-cases where we need contrastive losses and need to do negative sampling.

examples/tts/fastpitch_finetune_adapters.py
hsiehjackson: As discussed, this is by design. This is how adapters are implemented in ASR and we just used it as reference.
Do you mean we should give users the flexibility to use adapters to specific layers and not to entire modules like the whole of aligner/encoder etc but just few layers in these modules?

nemo/collections/tts/modules/fastpitch.py
rlangman: Either way technically works. It looks like we all agree the existing SpeakerEncoder should have an embedding array internally.

If we don't want to go with my above suggestion and do want to have a SpeakerEncoder as the only means later down, then we should create and finalize an interface in this PR to avoid breaking the adapter models later.

Current SpeakerEncoder in submodules.py is the "spaghetti code" approach with tons of if statements. The clean way to implement it would be like:

```
class SpeakerEncoder(ABC):

  @abstractmethod
  def get_speaker_encoder(self, speaker: Tensor, reference_speaker_embedding: Tensor, ...) -> Tensor:
    raise NotImplementedError
```
        
Then have one implementation which is the current GST logic with one logic flow and the if/else statements removed. And another simple one:

```
class LookupTableSpeakerEncoder(SpeakerEncoder):
    def __init__(self, num_speakers):
      self.speaker_emb = torch.nn.Embedding(n_speakers, symbols_embedding_dim)

    def get_speaker_encoder(self, speaker: Tensor, ...):
      if speaker is None:
        raise ValueError()
     return self.speaker_emb(speaker)
``'

File: nemo/collections/tts/modules/submodules.py (hunk 28 above)
Comment (hsiehjackson) [reply nemo/collections/tts/modules/submodules.py]: We don't use the activation function, so I remove it.

File: nemo/collections/tts/modules/submodules.py (hunk 27 above)
Comment (hsiehjackson) [reply nemo/collections/tts/modules/submodules.py]: I change the implementation with `torch.nn.MultiheadAttention`.

File: nemo/collections/tts/modules/fastpitch.py (hunk 11 above)
Comment (rlangman) [reply nemo/collections/tts/modules/fastpitch.py:321]: Then to deprecate the old system we can have a branch that uses the existing lookup table when speaker encoder is not provided, and does not create a lookup table in the FP model when it is.

Then we take our time to retrain all of the FP models on NGC with the new LookupTableSpeakerEncoder and delete the old branch when we are ready.

File: nemo/collections/tts/modules/submodules.py (hunk 29 above)
Comment (hsiehjackson) [reply nemo/collections/tts/modules/submodules.py:640]: You are right this is unused, we kept it as an option if we want to use it. All these three components (gst_module, sv_projection_module, lookup_emb_projection_module) can be instantiated and used or not based on what we define in the config to represent speaker.  
Once we decide on if we should include lookup table in this class or not we will make the appropriate changes.

File: nemo/collections/tts/data/dataset.py [hunk 36]
```diff
@@ -522,6 +541,16 @@ def _pad_wav_to_multiple(self, wav):
                 )
         return wav
 
+    # Random sample a reference sample from the same speaker
+    def sample_reference_index(self, curr_index, n=1):
+        speaker_id = self.data[curr_index]["speaker_id"]
+        if len(self.speaker_to_index_map[speaker_id]) > 1:
+            reference_pool = self.speaker_to_index_map[speaker_id] - set([curr_index])
+        else:
+            reference_pool = self.speaker_to_index_map[speaker_id]
```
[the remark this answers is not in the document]
hsiehjackson: Yes. In else part there is just a single sample for a particular speaker, hence we don't remove it.

nemo/collections/tts/modules/fastpitch.py
hsiehjackson: Once we merge this PR, we can raise another PR for this -> this might take some time around 2 weeks to do this because even thought the code change is easy, we have to change the state_dict of the existing NGC checkpoints, test them and upload them.

nemo/collections/tts/data/dataset.py
rlangman: It might not be too bad. Maybe we can discuss how to do it when the new data loader is in PR.

Having worked a lot with contrastive losses and negative sampling before, they tend to be manageable when you train on 1 host but can become problematic when the dataset is large enough that you have to shard it and each host only sees a relatively small subset of data. For example to optimize this code, you would want a shard data by speakers (eg. host one gets speakers 0-3, host two 4-9, etc). But that is far in the future.

nemo/collections/tts/modules/submodules.py
hsiehjackson: @rlangman  For this, do you know why we add dropout after norm?
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/modules/fastpitch.py#L92

nemo/collections/tts/modules/submodules.py
hsiehjackson: Yes. I add ReferenceEncoder as a module which can be init in config.

nemo/collections/tts/modules/submodules.py
hsiehjackson: Add the check logic in `models/fastpitch.py` https://github.com/NVIDIA/NeMo/blob/tts_adapters_cln/nemo/collections/tts/models/fastpitch.py#L151-L176

nemo/collections/tts/modules/submodules.py
hsiehjackson: Add the check logic in models/fastpitch.py https://github.com/NVIDIA/NeMo/blob/tts_adapters_cln/nemo/collections/tts/models/fastpitch.py#L151-L176

hsiehjackson:
This PR split into two parts.
https://github.com/NVIDIA/NeMo/pull/6416
https://github.com/NVIDIA/NeMo/pull/6417

PR #6417, opened 2023-04-12
[TTS] FastPitch speaker encoder

hsiehjackson:
# What does this PR do ?

Adds FastPitch speaker encoder with reference audio and reference speaker embedding

**Collection**: TTS

# Changelog 
- Adds multi-speaker FastPitch pre-training with 1) Speaker verification model speaker embedding (reference speaker embedding) and 2) Global Style Tokens (reference audio)
    - `nemo/collections/tts/data/tts_dataset.py`
    - `nemo/collections/tts/torch/tts_data_types.py`
    - `nemo/collections/tts/models/fastpitch.py`
    - `nemo/collections/tts/modules/fastpitch.py`
    - `nemo/collections/tts/modules/submodules.py`
- Modified config and fine-tuning python script
    - `examples/tts/conf/fastpitch_speaker_adaptation.yaml`
    - `examples/tts/fastpitch_finetune_adapters.py`
 
# Usage

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #6416, opened 2023-04-12
[TTS] FastPitch adapter fine-tune and conditional layer normalization

hsiehjackson:
# What does this PR do ?

Adds FastPitch pre-training with CLNs and fine-tuning with adapters.

**Collection**: TTS

# Changelog 
- Adds multi-speaker FastPitch pre-training with Conditional Layer Normalization 
    - `nemo/collections/tts/modules/fastpitch.py`
    - `nemo/collections/tts/modules/transformer.py`
    - `nemo/collections/tts/modules/submodules.py`
- Add adapter modules for FastPitch fine-tuning
    - `nemo/collections/tts/models/fastpitch.py`
    - `nemo/collections/tts/modules/fastpitch.py`
    - `nemo/collections/tts/modules/transformer.py`
    - `nemo/collections/tts/modules/aligner.py`
    - `nemo/collections/tts/parts/mixins/__init__.py`
    - `nemo/collections/tts/parts/mixins/fastpitch_adapter_mixins.py`
- Add config and fine-tuning python script
    - `examples/tts/conf/fastpitch_speaker_adaptation.yaml`
    - `examples/tts/fastpitch_finetune_adapters.py`
- Fix aligner `nan` loss bug
    - `nemo/collections/tts/losses/aligner_loss.py`
 
# Usage

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #6096, opened 2023-02-23
[TTS] Create skeleton for TTS inference API

rlangman:
# What does this PR do ?

Add prototype implementation of TTS pipeline that is intended to be used for inference before and after model training.

These are far from finalized. The intent is for this to be experimental and to slowly incorporate them into our inference workflows (eg. for preprocessing, evaluation, analysis, deployment) and expand and iterate on the interfaces and design as we figure out what needs to be changed.

**Collection**: [TTS]

# Changelog
- Add TTS pipeline and module definitions and documentation.
- Add module implementations using existing backends.
- Add unit and functional tests for all modules. 

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

PR #6431, opened 2023-04-14
[TTS] Add tutorials for FastPitch TTS speaker adaptation with adapters

hsiehjackson:
# What does this PR do ?

Add two tutorials for FastPitch TTS speaker adaptation with adapters. 
One is FastPitch multi-speaker pretraining, the other is FastPitch adapter finetuning.

**Collection**: [TTS]

# Changelog 
- `tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb`
- `tutorials/tts/FastPitch_Adapter_Finetuning.ipynb.`

# Usage

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #6416 #6417


tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "73bc625f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# e.g. NeMoTTS_logs/FastPitch/Y-M-D_H-M-S/checkpoints/FastPitch--val_loss=XXX-epoch=XXX-last.ckpt\n",
+    "last_checkpoint_dir = sorted(list([i for i in (Path(logsdir) / \"FastPitch\").iterdir() if i.is_dir()]))[-1] / \"checkpoints\"\n",
+    "YOUR_PRETRAINED_FASTPITCH_CHECKPOINT = list(last_checkpoint_dir.glob('*-last.ckpt'))[0]\n",
```
rlangman, line 449: Retrieving the latest checkpoint in a directory might be common enough for us to add it as a function in "tts/parts/utils/helpers" (but not necessary for this specific PR as it is effectively 1 line of code).

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8d17f232",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "str(YOUR_FINETUNED_HIFIGAN_ON_MULTISPEAKER_CHECKPOINT)"
```
rlangman, line 834: Can remove the stray prints at the end

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+    "heteronyms_path={normdir}/heteronyms-052722 \\\n",
+    "model.speaker_encoder.lookup_module._target_=\"nemo.collections.tts.modules.submodules.SpeakerLookupTable\" \\\n",
+    "model.speaker_encoder.lookup_module.n_speakers=5 \\\n",
+    "model.speaker_encoder.gst_module._target_=\"nemo.collections.tts.modules.submodules.GlobalStyleToken\" \\\n",
+    "model.input_fft.condition_types=\"['add', 'layernorm']\" \\\n",
+    "model.output_fft.condition_types=\"['add', 'layernorm']\" \\\n",
+    "model.duration_predictor.condition_types=\"['add', 'layernorm']\" \\\n",
+    "model.pitch_predictor.condition_types=\"['add', 'layernorm']\" \\\n",
+    "model.alignment_module.condition_types=\"['add']\" \\\n",
+    "model.train_ds.dataloader_params.batch_size=8 \\\n",
```
rlangman, line 419: Is pretraining supposed to use such a small batch size (usually we do 32)? Or is 8 only intended for fine-tuning?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3ac16425",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# additional files\n",
+    "!mkdir -p {normdir} && cd {normdir} \\\n",
+    "&& wget https://raw.githubusercontent.com/nvidia/NeMo/$BRANCH/scripts/tts_dataset_files/cmudict-0.7b_nv22.10 \\\n",
+    "&& wget https://raw.githubusercontent.com/nvidia/NeMo/$BRANCH/scripts/tts_dataset_files/heteronyms-052722 \\"
```
rlangman, line 196: Moving forward we would like tutorials to access files from a cloned version of NeMo repo, rather than wget individual files/scripts

https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/NeMo_TTS_Primer.ipynb?short_path=9930674#L46

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+    "for m in valid_datas: m['old_speaker'], m['speaker'] = m['speaker'], speaker2id[m['speaker']]\n",
+    "json_writer(valid_datas, valid_manifest)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "14c74cab",
+   "metadata": {},
+   "source": [
+    "### Calculate Pitch Stats"
```
rlangman, line 271: Can use script https://github.com/NVIDIA/NeMo/blob/main/scripts/dataset_processing/tts/compute_speaker_stats.py

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   "source": [
+    "# 2. Fine-tune HiFiGAN on multi-speaker data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "88364f18",
+   "metadata": {},
+   "source": [
+    "## a. Dataset Preparation"
```
rlangman, line 466: Can you use https://github.com/NVIDIA/NeMo/blob/main/scripts/dataset_processing/tts/generate_mels.py or https://github.com/NVIDIA/NeMo/blob/main/scripts/dataset_processing/tts/resynthesize_dataset.py ?

rlangman (commented):
I have only looked at the multispeaker pretraining so far. As a general comment, most code blocks should be have some text describing what it is doing and why it is important.

For example, what is the data prepartation doing? Why do we need to modify the manifest to include absolute paths? Why replace the speaker IDs with indices starting at 0? What pitch statistics are we computing and why? Why fine-tune the vocoder?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "03c33763",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "train_manifest = os.path.abspath(os.path.join(manidir, 'train.json'))\n",
+    "valid_manifest = os.path.abspath(os.path.join(manidir, 'dev.json'))"
```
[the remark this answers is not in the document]
hsiehjackson: Yes. I use the subset of VCTK as examples.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "73bc625f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# e.g. NeMoTTS_logs/FastPitch/Y-M-D_H-M-S/checkpoints/FastPitch--val_loss=XXX-epoch=XXX-last.ckpt\n",
+    "last_checkpoint_dir = sorted(list([i for i in (Path(logsdir) / \"FastPitch\").iterdir() if i.is_dir()]))[-1] / \"checkpoints\"\n",
+    "YOUR_PRETRAINED_FASTPITCH_CHECKPOINT = list(last_checkpoint_dir.glob('*-last.ckpt'))[0]\n",
+    "YOUR_PRETRAINED_FASTPITCH_CHECKPOINT"
```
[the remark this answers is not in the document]
hsiehjackson: It is a checkpoint path which will be used for multiple times. It is fine to change to `pretrained_fastpitch_checkpoint`

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+    "sample_rate={sample_rate} \\\n",
+    "train_dataset={train_manifest} \\\n",
+    "validation_datasets={valid_manifest} \\\n",
+    "sup_data_types=\"['align_prior_matrix', 'pitch', 'speaker_id','reference_audio']\" \\\n",
+    "sup_data_path={suppdir} \\\n",
+    "pitch_mean={pitch_mean} \\\n",
+    "pitch_std={pitch_std} \\\n",
+    "phoneme_dict_path={normdir}/cmudict-0.7b_nv22.10 \\\n",
+    "heteronyms_path={normdir}/heteronyms-052722 \\\n",
+    "model.speaker_encoder.lookup_module._target_=\"nemo.collections.tts.modules.submodules.SpeakerLookupTable\" \\\n",
```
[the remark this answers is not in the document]
hsiehjackson: `phoneme_dict_path` and `heteronyms_path ` are removed

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "73bc625f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# e.g. NeMoTTS_logs/FastPitch/Y-M-D_H-M-S/checkpoints/FastPitch--val_loss=XXX-epoch=XXX-last.ckpt\n",
```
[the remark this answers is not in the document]
hsiehjackson: I think default is using datetime as a version? https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L249

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: I use this to let user know which checkpoint they'll get after running this tutorial.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: I use 8 because it is just an example. User can change based on their preference. I'll add this in the comment.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: This script should be run after `extract_sup_data.py`. And `extract_sup_data.py` is overwhelmed to go through dataset and dataloader.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: Do I need to modify `generate_mels.py` and `resynthesize_dataset.py` to support reference audio? The inference function needs the reference audio, so I don't use this two scripts currently.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: What do you mean "overwhelmed"?

It would be better to run `extra_sup_data.py` before training to avoid effectively duplicating its code in the tutorial, and to avoid computing the dataset pitch twice (pitch calculation is extremely slow). Using `compute_speaker_stats.py` then also gives the user the option of using speaker-level pitch norm instead of global pitch norm (which tends to work better for multi-speaker).

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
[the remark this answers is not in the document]
rlangman: `phoneme_dict_path `and `heteronyms_path ` are needed because they are dependent on where the user stores the file.

I meant the model.speaker_encoder.*, model.input_fft.*, etc are providing the same values as in the .yaml config, and so do nothing.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,867 @@
...
+   "id": "6ffdce37",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Normally 200 epochs\n",
+    "!(python {codedir}/fastpitch.py \\\n",
+    "--config-path={os.path.abspath(confdir)} \\\n",
+    "--config-name=fastpitch_align_44100_adapter.yaml \\\n",
+    "+init_from_pretrained_model=\"tts_en_fastpitch\" \\\n",
+    "sample_rate={sample_rate} \\\n",
```
rlangman, line 402: Should avoid providing sample_rate as an override since the config `fastpitch_align_44100_adapter.yaml` is for 44khz specifically.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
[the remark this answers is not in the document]
rlangman: I see. I always override the version with the name of the experiment. I guess the idea is you can technically figure out which experiment is which by looking at the hparams in the checkpoint.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: Yes. If I remember correctly the two scripts are similar (`generate_mels.py` uses predicted pitch/energy while `resynthesize_dataset.py` uses ground truth pitch/energy). Supporting reference audio would make more sense in `resynthesize_dataset.py`, but depends on how exactly you are doing the fine-tuning.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: Ok, then how about we make it slightly more descriptive and put it in 1 cell.

```
print(f"FastPitch checkpoint: {YOUR_PRETRAINED_FASTPITCH_CHECKPOINT}")
print(f"HiFi-Gan checkpoint: {YOUR_FINETUNED_HIFIGAN_ON_MULTISPEAKER_CHECKPOINT}")
```

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: We may have this [PR](https://github.com/NVIDIA/NeMo/pull/6317) to compute pitch stats. Do I still use `extract_sup_data.py` currently?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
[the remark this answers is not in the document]
hsiehjackson: `model.speaker_encoder` and `model.input_fft.condition_types` are new arguments. I am wondering if I don't set here, do user know how these parameters work?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: If user already clone the repo to use tutorial, do they still clone the whole repo again?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: Current workflow is `extract_sup_data.py` --> `compute_speaker_stats.py`

After that PR there will be 2 new scripts 'extract_features.py' --> 'compute_feature_stats.py', which do more or less the same thing with slightly different CLI arguments.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: Tutorials can be run in many ways. We most commonly expect them to be uploaded and run on Google Colab. But for users are who are running locally or on docker, we can ask them to comment out the "git clone" command and set NEMO_DIR to their local nemo directory.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: I see. I'll use `exetract_sup_data.py` in this PR.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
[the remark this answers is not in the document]
rlangman: Well, the users will not understand what those parameters do anyways unless the tutorial explains it.  If we want to explain what they do and how a user might adjust them based on their dataset, then we can explain it in bullet points in a text section similar to https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/FastPitch_MixerTTS_Training.ipynb?short_path=558c0d9#L520-L525

Alternatively, you could explain it by pulling the values into python variables ahead of time and explaining in comments.

```
# Change this depending on your GPU memory
train_batch_size = 8
# Set this to["add"] if you want to disable layer norm adaptation.
adapter_norm_type = ["add", "layernorm"]

...
model.train_ds.dataloader_params.batch_size={train_batch_size}
model.input_fft.condition_types={adapter_norm_type}
model.output_fft.condition_types={adapter_norm_type}
...
```

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: For `FastPitch_Adapter_Finetuning.ipynb`, if we use `resynthesize_dataset.py`, we can't load adapter weights in `SpectrogramGenerator`; if we use `generate_mels.py`, we need to add `load_adapters`. Additionally, because of the new speaker encoder, we also need input `reference audio` and make input `speaker` optional. Do you think we can generate mel-spectrograms without using the script in the notebook?

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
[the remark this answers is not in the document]
hsiehjackson: I explain the parameters in a separate markdown block.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
hsiehjackson: Add cloned version of NeMo repo.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
rlangman: If there are significant problems with reusing the previous scripts, then we can leave it in the notebook.

Figuring out how to make inference logic consistent across models is a different project (eg. https://github.com/NVIDIA/NeMo/pull/6096) and I don't think extending base.SpectrogramGenerator or adding a 3rd duplicate inference script is a good idea at this point.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,877 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "78680a01",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Normally 100 epochs (15 mins)\n",
```
rlangman, line 442: I would avoid putting the minute estimations since it is highly dependent on GPU type.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,877 @@
...
+    "sup_data_types=\"['align_prior_matrix', 'pitch', 'speaker_id', 'reference_audio']\" \\\n",
+    "sup_data_path={suppdir} \\\n",
+    "pitch_mean={PITCH_MEAN} \\\n",
+    "pitch_std={PITCH_STD} \\\n",
+    "~model.speaker_encoder.lookup_module \\\n",
+    "model.train_ds.dataloader_params.batch_size=8 \\\n",
+    "model.validation_ds.dataloader_params.batch_size=8 \\\n",
+    "model.optim.name=adam \\\n",
+    "model.optim.lr=2e-4 \\\n",
+    "model.optim.weight_decay=0.0 \\\n",
```
rlangman, line 457: Why disable weight decay/regularization during fine tuning?

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,877 @@
...
+   "execution_count": null,
+   "id": "5a187093",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Reference Audio\n",
+    "with open(train_manifest, \"r\") as f:\n",
+    "    for i, line in enumerate(f):\n",
+    "        reference_record = json.loads(line)\n",
+    "        break\n",
```
rlangman, line 791: Is this randomly selecting the first line from the training data as the reference? Shouldn't the reference come from the validation set, and belong to the same speaker?

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -117,6 +118,15 @@ def resynthesize_batch(self, batch: Dict[str, Any]) -> Dict[str, Any]:
         batch = to_device_recursive(batch, self.device)
 
         mels, mel_lens = self.model.preprocessor(input_signal=batch["audio"], length=batch["audio_lens"])
+
+        reference_audio = batch.get("reference_audio", None)
+        reference_audio_len = batch.get("reference_audio_lens", None)
```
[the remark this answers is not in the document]
hsiehjackson: Yes this is not easy to modify for adapters. But, for reference audio (speaker encoder) parts, it is easy to change.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
hsiehjackson: I expect user only use one target speaker data to run this notebook, so the training set and validation set are from the same speaker. I think there is no much difference we use from training or validation set.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,877 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!cd {codedir} && python scripts/dataset_processing/tts/extract_sup_data.py \\\n",
+    "    manifest_filepath={valid_manifest} \\\n",
+    "    sup_data_path={suppdir} \\\n",
+    "    dataset.sample_rate={sample_rate} \\\n",
+    "    dataset.n_fft=2048 \\\n",
+    "    dataset.win_length=2048 \\\n",
+    "    dataset.hop_length=512"
```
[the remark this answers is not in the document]
hsiehjackson: This is for validation, and previous block is for training

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
hsiehjackson: This can be removed.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
rlangman: Got it, that makes sense for adapters.

scripts/dataset_processing/tts/resynthesize_dataset.py
```diff
@@ -80,6 +80,7 @@ def chunks(iterable: Iterable, size: int) -> Iterator[List]:
 
 
 def load_model(path: Path, device: torch.device) -> SpectrogramGenerator:
+    print(path)
```
rlangman, line 83: Nitpick: should make print statement clearer like "Loading checkpoint - {}"

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,868 @@
...
+    "model = FastPitchModel.restore_from(pretrained_fastpitch_checkpoint)\n",
+    "model.cfg = update_model_config_to_support_adapter(model.cfg)\n",
+    "model.save_to('Pretrained-FastPitch.nemo')\n",
+    "shutil.copyfile(finetuned_hifigan_on_multispeaker_checkpoint, \"Pretrained-HifiGan.nemo\")\n",
+    "\n",
+    "pretrained_fastpitch_checkpoint = os.path.abspath(\"Pretrained-FastPitch.nemo\")\n",
+    "finetuned_hifigan_on_multispeaker_checkpoint = os.path.abspath(\"Pretrained-HifiGan.nemo\")\n",
+    "#     state = torch.load(pretrained_fastpitch_checkpoint)\n",
+    "#     state['hyper_parameters']['cfg'] = update_model_config_to_support_adapter(state['hyper_parameters']['cfg'])\n",
+    "#     torch.save(state, pretrained_fastpitch_checkpoint)"
```
rlangman, line 218: Is the commented code needed?

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,868 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "93c0f45a",
+   "metadata": {},
+   "source": [
+    "### Extract Supplementary Data\n",
+    "\n",
+    "As mentioned in the [FastPitch and MixerTTS training tutorial](https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/FastPitch_MixerTTS_Training.ipynb) - To accelerate and stabilize our training, we also need to extract pitch for every audio, estimate pitch statistics (mean, std, min, and max). To do this, all we need to do is iterate over our data one time, via `extract_sup_data.py` script.\n",
+    "\n",
+    "Note: This is an optional step, if skipped, it will be automatically executed within the first epoch of training FastPitch."
```
rlangman, line 322: Some past tutorials said this, but it is not entirely correct. If the script is not run then the training script will compute the pitch during first epoch but it will either not normalize the pitch, or normalize using default pitch values (eg. hardcoded from LJSpeech).

For most of our example Fastpitch configs we now set `pitch_mean: ???` so that user cannot accidentally run the training without precomputing and providing the statistics.

tutorials/tts/FastPitch_MultiSpeaker_Pretraining.ipynb
```diff
@@ -0,0 +1,736 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c2a6f14a",
+   "metadata": {},
+   "source": [
+    "### Extract Supplementary Data\n",
+    "\n",
+    "As mentioned in the [FastPitch and MixerTTS training tutorial](https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/FastPitch_MixerTTS_Training.ipynb) - To accelerate and stabilize our training, we also need to extract pitch for every audio, estimate pitch statistics (mean, std, min, and max). To do this, all we need to do is iterate over our data one time, via `extract_sup_data.py` script.\n",
+    "\n",
+    "Note: This is an optional step, if skipped, it will be automatically executed within the first epoch of training FastPitch."
```
rlangman, line 263: Similar to comment on other tutorial file

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
```diff
@@ -0,0 +1,868 @@
...
+    "finetuned_adapter_checkpoint = list(last_checkpoint_dir.glob('adapters.pt'))[0]\n",
+    "print(finetuned_adapter_checkpoint)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "33dcbc8f",
+   "metadata": {},
+   "source": [
+    "# 4. Fine-tune HiFiGAN on adaptation data"
```
rlangman, line 460: Since this section has so little text/explanation, maybe we can reference/link to the main finetuning tutorial.

https://github.com/NVIDIA/NeMo/blob/main/tutorials/tts/FastPitch_Finetuning.ipynb

I do not see any way to link to the specific section though in github, other than to say it is in the "Finetuning HiFi-GAN" section.

rlangman (commented):
Looks good to me overall. A few minor comments.

tutorials/tts/FastPitch_Adapter_Finetuning.ipynb
hsiehjackson: I add explanation why we don't use existing script to generate mel-spectrograms.

nemo/collections/tts/modules/submodules.py
```diff
@@ -709,18 +709,29 @@ class SpeakerEncoder represents speakers representation.
...
         """
         super(SpeakerEncoder, self).__init__()
+
+        # Multi-speaker embedding
         self.lookup_module = lookup_module
+
+        # Reference speaker embedding
         self.gst_module = gst_module
 
+        if precompute:
```
rlangman, line 726: Can we have the check be if `precomputed_embedding_dim` is not None, instead of a separate `precompute` flag?

nemo/collections/tts/modules/submodules.py
```diff
@@ -709,18 +709,29 @@ class SpeakerEncoder represents speakers representation.
     This module can combine GST (global style token) based speaker embeddings and lookup table speaker embeddings.
     """
 
-    def __init__(self, lookup_module=None, gst_module=None):
+    def __init__(self, lookup_module=None, gst_module=None, precompute=False, precompute_embedding_dim=None):
         """
         lookup_module: Torch module to get lookup based speaker embedding
         gst_module: Neural module to get GST based speaker embedding
+        precompute: Use precompute speaker embedding
         """
```
rlangman, line 717: Missing `precompute_embedding_dim` param

nemo/collections/tts/modules/submodules.py
```diff
@@ -709,18 +709,29 @@ class SpeakerEncoder represents speakers representation.
...
         super(SpeakerEncoder, self).__init__()
+
+        # Multi-speaker embedding
         self.lookup_module = lookup_module
+
+        # Reference speaker embedding
         self.gst_module = gst_module
 
+        if precompute:
+            self.precomputed_emb = torch.nn.Parameter(torch.empty(precompute_embedding_dim))
```
rlangman, line 727: Would be nice to have a method in the class that shows how to overwrite this value.

nemo/collections/tts/modules/submodules.py
hsiehjackson: If user want to use precompute embedding, I think setting a bool value is better than setting a dimension value?

rlangman: Having 2 parameters that are coupled but optional is awkward. If you keep it as is there should be an assert which validates that the dimension is None when the flag is false, or dimension is not None when the flag is true (similar to the XOR in https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/data/dataset.py#L443-L445).

Normally one would provide dependent arguments as a single module or as a config object to ensure user provides all or none of them, but in this case there is only 1 argument to configure so checking if its None is also fine.

nemo/collections/tts/modules/submodules.py
hsiehjackson: Add function`overwrite_precomputed_emb()`

nemo/collections/tts/modules/submodules.py
hsiehjackson: Use only one parameter `precomputed_embedding_dim`

nemo/collections/tts/modules/submodules.py
```diff
@@ -732,9 +743,16 @@ def output_types(self):
             "embs": NeuralType(('B', 'D'), EncodedRepresentation()),
         }
 
-    def forward(self, speaker=None, reference_spec=None, reference_spec_lens=None):
+    def overwrite_precomputed_emb(self, emb):
+        self.precomputed_emb = torch.nn.Parameter(emb)
+
+    def forward(self, batch_size, speaker=None, reference_spec=None, reference_spec_lens=None):
```
rlangman, line 749: Could default batch_size to None since only the precomputed branch uses it.

rlangman (approved):
LGTM

nemo/collections/tts/modules/submodules.py
hsiehjackson: Done!

rlangman (approved):

zaidato:
I did follow the steps in the tutorial, but the output audio was just noise. The model could not pronounce any word.

PR #6433, opened 2023-04-16
Upgrade to pytorch lightning 2.0

athitten:
# What does this PR do ?

This PR upgrades NeMo to use lightning 2.0. Majority of the changes are done based on the migration guide provided by lightning: [https://lightning.ai/docs/pytorch/stable/upgrade/from_1_9.html](url)

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


athitten:
Note for the reviewers to give some background on the changes:
@ericharper @titu1994 @aklife97 @yaoyu-33 

1. Change in the function, argument names etc., can be found in the migration guide provided by lightning: [https://lightning.ai/docs/pytorch/stable/upgrade/from_1_9.html](url). 
2. A major change that can be seen in most of the files for all 3 domains (ASR, TTS, NLP) is the renaming of `validation/test_epoch_end(outputs)` hooks to `on_validation/test_epoch end()`. A detailed overview for this can be found here: [](https://github.com/Lightning-AI/lightning/pull/16520). Lightning's recommendation of making `outputs` an instance variable as its not a supported arg in the epoch end callbacks has been implemented. It's been initialized in the ModelPT class so that it applies to all the inherited classes. Also, the new PTL doesn't fetch outputs automatically from the `validation/test_step` and it needs to be appended to the instance variables explicitly in the `validation/test_step`.
3. PTL has added more options for precision argument in trainer.  16bit mixed precision can be passed with any of these values: 16, ‘16’, ‘16-mixed’ and bfloat16 mixed precision with: ‘bf16’ or ‘bf16-mixed’. Also PTL's `MixedPrecisionPlugin` class now expects the precision arg to be ‘16-mixed’ or ‘bf16-mixed’ and doesn't support passing it as just '16' or 'bf16'. Changes in the NLP/megatron files can be seen to accommodate this.
4. There are a bunch of other small changes to fix the errors that came with lightning 2.0, for which brief comments are added in the code itself explaining why it was done.

redoctopus (commented):
TTS side looks good to me, aside from a couple of minor typos.

nemo/collections/common/callbacks/ema.py
```diff
@@ -119,7 +119,7 @@ def on_load_checkpoint(
 
         # use the connector as NeMo calls the connector directly in the exp_manager when restoring.
         connector = trainer._checkpoint_connector
-        ckpt_path = connector.resume_checkpoint_path
+        ckpt_path = connector._ckpt_path
```
awaelchli, line 122: This would be equivalent to replace these lines:
```py
ckpt_path = trainer.ckpt_path
```

and avoids calling protected API

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -159,7 +163,7 @@ def _metrics_require_string2category_map(self):
     def setup(self, stage=None):
         # NOTE: super().__init__ will try and setup train/val/test datasets, but we sidestep this using a if self._train_ds is not None condition
         # We then set things up for real only once setup() of this class is called.
-        resume_checkpoint_path = self.trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+        resume_checkpoint_path = self.trainer._checkpoint_connector._ckpt_path
```
awaelchli, line 166:
```suggestion
        resume_checkpoint_path = self.trainer.ckpt_path
```

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -115,7 +115,7 @@ def configure_ddp(self):
             hasattr(self.model, 'with_distributed_adam') and self.model.with_distributed_adam
         ):
             # do not use DDP if using megatron amp O2 or distributed optimizer
-            self._model = LightningDistributedModule(self.model)
+            self._model = _LightningModuleWrapperBase(self.model)
```
awaelchli, line 118: LGTM. Note that in in the future 2.1 version, this minor annoyance of having to wrap the module with our special wrapper won't be necessary anymore 🎉

nemo/utils/exp_manager.py
```diff
@@ -599,7 +598,8 @@ def check_resume(
         logging.info(f"Resuming from {last_checkpoints[0]}")
         checkpoint = last_checkpoints[0]
 
-    trainer._checkpoint_connector.resume_from_checkpoint_fit_path = str(checkpoint)
+    # PTL 2.0 supports ckpt_path instead of resume_from_checkpoint as the trainer flag
+    trainer.ckpt_path = str(checkpoint)
```
awaelchli, line 602: Yep! Awesome. It also supports this as a setter, so I recommend replacing all calls and references to `_checkpoint_connector._ckpt_path` with this.

awaelchli (commented):
Awesome effort @athitten! Thanks so much for making the upgrade happen 🎉

nemo/core/classes/modelPT.py
```diff
@@ -177,12 +178,39 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
                     f"Test config : \n{OmegaConf.to_yaml(self._cfg.test_ds)}"
                 )
 
+        # Create list of lists for val and test outputs to support multiple dataloaders
```
titu1994, line 181: Oh now i see - they removed the `outputs` arg for validation_epoch_end and test_epoch_end so we need a cache storage inside the model..

Btw - this if check - it will fail to execute for all Megatron models + any model that adds `defer_setup=True` in their val/test_ds configs. We need to add these checks below as well somewhere.

nemo/core/classes/modelPT.py
```diff
@@ -177,12 +178,39 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+        if self._test_dl and type(self._test_dl) == list and len(self._test_dl) > 1:
+            for _ in range(len(self._test_dl)):
+                self.test_step_outputs.append([])
         # ModelPT wrappers over subclass implementations
         self.training_step = model_utils.wrap_training_step(self.training_step)
 
         # Setup nsys profiling if it has been enabled in the model config
         self._setup_nsys_profiling()
 
+    def _prefetch(self, iterator):
```
titu1994, line 201: Hmm does NeMo core really need to provide this? This seems to be more of a data thing than a NeMo core thing. I see its used in Megatron, maybe best to move it to NLPModel level if its not used anywhere else.

nemo/core/classes/modelPT.py
```diff
@@ -873,23 +899,24 @@ def validation_epoch_end(
...
         """
         # Case where we dont provide data loaders
-        if outputs is not None and len(outputs) == 0:
+        if self.validation_step_outputs is not None and len(self.validation_step_outputs) == 0:
             return {}
 
         # Case where we provide exactly 1 data loader
-        if type(outputs[0]) == dict:
-            output_dict = self.multi_validation_epoch_end(outputs, dataloader_idx=0)
+        if type(self.validation_step_outputs[0]) == dict:
```
titu1994, line 906: use isinstance(xxx, dict)

nemo/core/classes/modelPT.py
```diff
@@ -968,23 +995,24 @@ def test_epoch_end(
...
         """
         # Case where we dont provide data loaders
-        if outputs is not None and len(outputs) == 0:
+        if self.test_step_outputs is not None and len(self.test_step_outputs) == 0:
             return {}
 
         # Case where we provide exactly 1 data loader
-        if type(outputs[0]) == dict:
-            output_dict = self.multi_test_epoch_end(outputs, dataloader_idx=0)
+        if type(self.test_step_outputs[0]) == dict:
```
titu1994, line 1002: Same as above

tests/collections/asr/test_asr_interctc_models.py
```diff
@@ -234,7 +234,10 @@ def __getitem__(self, idx):
                 if model_class is EncDecCTCModel:
                     assert output[0].shape == logprobs.shape
 
-            trainer = pl.Trainer(max_epochs=1)
+            ## Explicitly pass acclerator as cpu, since deafult val in PTL >= 2.0 is auto and it picks cuda
```
titu1994, line 237: @Kipok Could you take a look at this test and the bugs (?) that may be there under these cases.

tests/collections/asr/test_asr_local_attn.py
```diff
@@ -171,8 +171,10 @@ def __getitem__(self, idx):
         asr_model = EncDecCTCModel(cfg=model_config)
         asr_model.train()
         _ = asr_model.forward(input_signal=input_signal, input_signal_length=input_length)
-
-        trainer = pl.Trainer(max_epochs=1)
+        ## Explicitly pass acclerator as cpu, since deafult val in PTL >= 2.0 is auto and it picks cuda
```
titu1994, line 174: TODO: @titu1994

tests/core/test_config_utils.py
```diff
@@ -117,12 +117,7 @@ class DummyDataClass:
 
     @pytest.mark.unit
     def test_ptl_config(self):
-        PTL_DEPRECATED = [
-            'gpus',
-            'num_processes',
-        ]
-
-        result = config_utils.assert_dataclass_signature_match(ptl.Trainer, TrainerConfig, ignore_args=PTL_DEPRECATED)
+        result = config_utils.assert_dataclass_signature_match(ptl.Trainer, TrainerConfig)
```
titu1994, line 120: Hmm lets keep empty list of PTL_DEPRECATED = [] here and pass the arg. We may have to use it in the futuer.

titu1994 (commented):
Incredible work! Minor comments here and there, but overall it looks great!

examples/nlp/language_modeling/megatron_bart_pretraining.py
```diff
@@ -67,10 +71,10 @@ def main(cfg) -> None:
     if cfg.model.resume_from_checkpoint is not None:
         resume_from_checkpoint = cfg.model.resume_from_checkpoint
     else:
-        resume_from_checkpoint = trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+        resume_from_checkpoint = trainer._checkpoint_connector._ckpt_path
     logging.info(f'Resuming training from checkpoint: {resume_from_checkpoint}')
 
-    trainer._checkpoint_connector = CheckpointConnector(trainer, resume_from_checkpoint=resume_from_checkpoint)
+    trainer._checkpoint_connector = _CheckpointConnector(trainer)
```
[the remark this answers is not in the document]
awaelchli: Hi @titu1994 
Adrian from the Lightning team here. For this specific code block here, I believe reaching into the checkpoint connector should no longer be necessary in Lightning 2.0, because it is now possible to set the path like this: `trianer.ckpt_path = cfg.model.resume_from_checkpoint`

But if you have other uses of the CheckpointConnector that call into protected API, please let me know so I can study your use case 🎉

examples/nlp/language_modeling/megatron_bart_pretraining.py
```diff
@@ -43,18 +43,22 @@ def main(cfg) -> None:
         gradient_as_bucket_view=cfg.model.gradient_as_bucket_view,
         find_unused_parameters=False,
     )
-    if cfg.trainer.precision in [16, 'bf16']:
+    if cfg.trainer.precision in [16, 'bf16', '16-mixed', 'bf16-mixed']:
```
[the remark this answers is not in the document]
awaelchli: In Lightning 2.0 we introduced "16-mixed" to make this less ambiguous. Yes, for backward compatibility `16` still does AMP and is equivalent to "16-mixed". We started promoting the string version in all our updated docs and examples. In Lightning 2.1, there will also be the options "16-true" and "bf16-true" for full fp16 training casting weights like you said with `model.half()` or `model.bfloat16()`

examples/nlp/language_modeling/megatron_gpt_pretraining.py
```diff
@@ -71,11 +76,11 @@ def main(cfg) -> None:
     if cfg.model.resume_from_checkpoint is not None:
```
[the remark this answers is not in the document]
athitten: Thanks @awaelchli this has been fixed in the latest commit.

tests/collections/asr/test_asr_interctc_models.py
Kipok: I'm a bit out of context - what exactly is the error that's happening here? Using CPU should be fine I guess, since we do not test anything GPU-specific here. Or does it not work in that way?

nemo/collections/common/callbacks/ema.py
athitten: @awaelchli this has been fixed in the latest commit, thanks!

nemo/utils/exp_manager.py
athitten: Thanks, Incorporated this @awaelchli

examples/nlp/entity_linking/self_alignment_pretraining.py
```diff
@@ -27,6 +27,10 @@
 
 @hydra_runner(config_path="conf", config_name="umls_medical_entity_linking_config.yaml")
 def main(cfg: DictConfig) -> None:
+    # PTL 2.0 has find_unused_parameters as False by default, so its required to set it to True
+    # when there are unused parameters here
+    if cfg.trainer.strategy == 'ddp':
```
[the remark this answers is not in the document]
athitten: @titu1994 quick FYI: in previous lightning versions like 1.9 find_unused_parameters was True by default. In lightning 2.0 it was made False by default to improve performance. Did we observe ACC regression with 1.9 versions ? If not, then setting "ddp_find_unused_parameters_true" should be okay right as it was the case by default in 1.9. More details in this PR: [https://github.com/Lightning-AI/lightning/pull/16611#:~:text=Case%201%3A%20User%20has%20unused%20parameters%20in%20their%20forward%2Dbackward](url)

nemo/collections/asr/models/classification_models.py
```diff
@@ -577,25 +577,35 @@ def validation_step(self, batch, batch_idx, dataloader_idx=0):
...
         acc = self._accuracy(logits=logits, labels=labels)
         correct_counts, total_counts = self._accuracy.correct_counts_k, self._accuracy.total_counts_k
-        return {
+        loss = {
             'val_loss': loss_value,
             'val_correct_counts': correct_counts,
             'val_total_counts': total_counts,
             'val_acc': acc,
         }
+        if type(self.trainer.val_dataloaders) == list and len(self.trainer.val_dataloaders) > 1:
```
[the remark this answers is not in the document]
athitten: Also adding more to this statement: _ModelPT does take care of initializing self.validation_step_outputs appropriately based on the number of dataloaders_
There are cases for ex when we are restoring the model, ModelPT doesn't do the setup of the dataloaders and it gets overridden by setup defined in the specific model class. For cases where this issue was caught through an error in CI,  the initialization of self.validation_step_outputs is added in the specific model class.

nemo/collections/nlp/metrics/classification_report.py
```diff
@@ -203,9 +205,10 @@ def validation_step(self, batch, batch_idx):
             ...
             tp, fn, fp, _ = self.classification_report(preds, labels)
 
-            return {'val_loss': val_loss, 'tp': tp, 'fn': fn, 'fp': fp}
+            self.val_outputs = {'val_loss': val_loss, 'tp': tp, 'fn': fn, 'fp': fp}
+            return self.val_outputs
```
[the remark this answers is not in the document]
athitten: Yup not needed here, removed it.

nemo/collections/nlp/models/language_modeling/megatron_t0_model.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
     def setup(self, stage=None):
         # NOTE: super().__init__ will try and setup train/val/test datasets, but we sidestep this using a if self._train_ds is not None condition
         # We then set things up for real only once setup() of this class is called.
-        resume_checkpoint_path = self.trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+        resume_checkpoint_path = self.trainer._checkpoint_connector._ckpt_path
```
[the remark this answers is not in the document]
athitten: Yes this is fixed!

examples/tts/g2p/g2p_heteronym_classification_train_and_evaluate.py
```diff
@@ -62,6 +62,10 @@
 
 @hydra_runner(config_path="conf", config_name="g2p_heteronym_classification.yaml")
 def main(cfg):
+    # PTL 2.0 has find_unused_parameters as False by default, so its required to set it to True
+    # when there are unused parameters like in this model
```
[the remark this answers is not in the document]
athitten: Thanks @redoctopus, this has been fixed.

nemo/collections/tts/models/fastpitch.py
```diff
@@ -588,6 +590,7 @@ def validation_epoch_end(self, outputs):
                 "val_mel_predicted", plot_spectrogram_to_numpy(spec_predict), self.global_step, dataformats="HWC",
             )
             self.log_train_images = True
+        self.validation_step_outputs.clear()  # free memory)
```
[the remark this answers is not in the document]
athitten: Fixed in the latest commit.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -550,6 +550,10 @@ def training_step(self, dataloader_iter, batch_idx):
             The input batch to each micro-batch is fetched using the dataloader function
             in the micro-batch fwd function.
         """
+        # Prefetch the dataloader_iter before fwd_bwd func to avoid PP rank 2 from waiting indefinitely when PP rank 1 reaches the end of dataloader_iter
+        dataloader_iter, done = self._prefetch(dataloader_iter)
+        if done:
```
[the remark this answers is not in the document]
athitten: Hmm I am unsure if there can be situations where data isn't ready. If the batch exists we should be able to retrieve it with `next(dataloader_iter)` right ? That's why we try getting it and if not it goes to `except StopIteration` indicating the end of iterator in which case we don't want to do `fwd_bwd_func` at all. @titu1994 let me know if I am missing something here. Thanks!

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -95,6 +96,23 @@ def __init__(
...
+        across all GPUs. Lightning's PR for adding _sync_module_states: https://github.com/Lightning-AI/lightning/pull/17370
+        """
+        # Currently overiding only in case of PP>1. Check if TP>1 also needs _sync_module_states(self.model) to be disabled
+        app_state = AppState()
+        if app_state.pipeline_model_parallel_size is not None and app_state.pipeline_model_parallel_size > 1:
+            trainer_fn = trainer.state.fn
+            if trainer_fn == TrainerFn.FITTING:
+                super().setup(trainer)
+            else:
+                assert self.model is not None
```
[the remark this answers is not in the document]
athitten: @titu1994 just realized an issue with this, fixing it.

nemo/core/classes/modelPT.py
athitten: Yes you are right, for cases where setup is not happening in ModelPT and multiple dataloaders were getting used, the tests gave error and I have added this check in those specific model files. Since they were cases where setup was overridden inside the model class. 
Any suggestions on if we want to add this condition check to all models that have setup overridden or add it when the need arises ?
I have added the check again in ModelPT inside def setup() to handle scenario when defere_setup is True.

nemo/collections/asr/models/classification_models.py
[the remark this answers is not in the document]
titu1994: That seems error prone. Why not initialize the two variables always to empty lists, whether or not the data loader is called. 

Then if multi data loader is detected in setup, change the default value of these two to nested list ? That way subclass doesn't have to worry about it. Setup is always called when train val begins

examples/nlp/language_modeling/megatron_gpt_pretraining.py
```diff
@@ -71,11 +76,11 @@ def main(cfg) -> None:
...
         resume_from_checkpoint = cfg.model.resume_from_checkpoint
     else:
-        resume_from_checkpoint = trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+        resume_from_checkpoint = trainer._checkpoint_connector._ckpt_path
 
     logging.info(f'Resuming training from checkpoint: {resume_from_checkpoint}')
 
-    trainer._checkpoint_connector = CheckpointConnector(trainer, resume_from_checkpoint=resume_from_checkpoint)
+    trainer._checkpoint_connector = _CheckpointConnector(trainer)
```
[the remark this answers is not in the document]
athitten: @awaelchli facing weird issue with removing  this line: `trainer._checkpoint_connector = _CheckpointConnector(trainer)`
Couple of the models error out with this error at this line: [https://github.com/Lightning-AI/lightning/blob/master/src/lightning/pytorch/loops/fit_loop.py#L259](url)

Couldn't really figure out why that's the case. Adding it back in these tests for now, to ensure the CI is passing until we have clarity on this.

examples/nlp/language_modeling/megatron_bert_pretraining.py
```diff
@@ -68,11 +72,12 @@ def main(cfg) -> None:
...
 
     # update resume from checkpoint found by exp_manager
-    resume_from_checkpoint = trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+    # Avoid calling protected API trainer._checkpoint_connector._ckpt_path as lightning 2.0 supports ckpt_path as trainer arg
+    resume_from_checkpoint = trainer.ckpt_path
     # resume_from_checkpoint = uninject_model_parallel_rank(resume_from_checkpoint)
     logging.info(f'Resuming training from checkpoint: {resume_from_checkpoint}')
 
-    trainer._checkpoint_connector = CheckpointConnector(trainer, resume_from_checkpoint=resume_from_checkpoint)
+    trainer._checkpoint_connector = _CheckpointConnector(trainer)
```
aklife97, line 80: nit: we can remove this

aklife97 (commented):
Thanks for the PR @athitten, this was a huge task!
I am leaving some comments for pieces I was able to review, still need to look into `collections/nlp`, will post the next part of my review soon

examples/nlp/language_modeling/megatron_bart_pretraining.py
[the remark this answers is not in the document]
awaelchli: We already show a warning that the use is discouraged, but we haven't discussed when we would start a proper deprecation.

examples/nlp/language_modeling/megatron_bart_pretraining.py
```diff
@@ -43,18 +43,22 @@ def main(cfg) -> None:
...
         scaler = None
-        if cfg.trainer.precision == 16:
+        if cfg.trainer.precision == 16 or cfg.trainer.precision == '16-mixed':
```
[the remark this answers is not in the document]
ericharper: We have this coming: https://github.com/maanug-nv/NeMo/blob/llm-trainer-builder/examples/nlp/language_modeling/megatron_gpt_pretraining.py#L33

examples/nlp/language_modeling/megatron_bart_pretraining.py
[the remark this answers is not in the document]
ericharper: @athitten should we follow this suggestion?

examples/nlp/language_modeling/megatron_lm_ckpt_to_nemo.py
```diff
@@ -42,11 +42,11 @@
 from typing import Any, Optional
 
 import torch
+from lightning_fabric.utilities.cloud_io import _load as pl_load
```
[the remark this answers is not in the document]
ericharper: Maybe this would work? https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.plugins.io.CheckpointIO.html#checkpointio

examples/nlp/language_modeling/megatron_retro_fine_tune.py
```diff
@@ -102,7 +102,7 @@ def main(cfg) -> None:
     exp_manager(trainer, cfg.exp_manager)
 
     # update resume from checkpoint found by exp_manager
-    resume_from_checkpoint = trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+    resume_from_checkpoint = trainer._checkpoint_connector._ckpt_path
     logging.info(f'Resuming training from checkpoint: {resume_from_checkpoint}')
 
     trainer._checkpoint_connector = CheckpointConnector(trainer, resume_from_checkpoint=resume_from_checkpoint)
```
ericharper, line 108: Didn't this become private?

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -83,9 +83,9 @@ def main(cfg) -> None:
         gradient_as_bucket_view=cfg.model.gradient_as_bucket_view,
         find_unused_parameters=False,
     )
-    if cfg.trainer.precision in [16, "bf16"]:
+    if cfg.trainer.precision in [16, "bf16", '16-mixed', 'bf16-mixed']:
         scaler = None
-        if cfg.trainer.precision == 16:
+        if cfg.trainer.precision == 16 or cfg.trainer.precision == '16-mixed':
```
ericharper, line 88: nit: why not `if cfg.trainer.precision in [16, '16', '16-mixed]` ? This is less characters and it's similar to what we do in line 86. Also 16 should probably be `[16, '16', 'bf16', ...`]

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
```diff
@@ -333,58 +306,73 @@ def fwd_bwd_step(self, dataloader_iter, batch_idx, forward_only):
...
-        # this will be ignored.
-        loss = self.fwd_bwd_step(itertools.chain([batch]), batch_idx, forward_only=True)
-
-        predicted_token_ids, _ = self.decode(
-            tokens_enc=batch['text_enc'],
-            enc_mask=batch['enc_mask'],
-            num_tokens_to_generate=30,
-            bos_id=self.tokenizer.pad_id if data_cfg.get('replace_bos_with_pad', False) else self.tokenizer.bos_id,
-        )
+        # Add try except since dataloader_iter in PTL 2.0 doesnt catch the end of the iterator
```
aklife97, line 309: ugh... this try catch looks ugly...
I dont understand the rationale why PTL removed the StopIteration check
Instead of doing this on every step function, can we handle this try-catch at a level above? maybe directly handle this in base class or create a function that calls our step functions that puts everything in try except?

nemo/collections/nlp/models/language_modeling/megatron_base_prompt_learning_model.py
```diff
@@ -133,14 +132,16 @@ def init_model(self, cfg: DictConfig, trainer: Trainer):
         self.pad_token_id = self.tokenizer.pad_id if self.tokenizer.pad_id is not None else self.tokenizer.unk_id
         self.decoder_seq_length = cfg.get('decoder_seq_length', 40)
 
-        if self.trainer.precision == 'bf16':
+        # Check for first few chars to be '16', '32' or 'bf16' as Precision can also be '16-mixed', '32-true' or 'bf16-mixed'
+        # along with 16, 32 or bf16 in case of PTL >= 2.0
+        if self.trainer.precision[0:4] == 'bf16':
```
aklife97, line 137: not sure if its only me, but this indexing feels slightly hacky, can we change it something like what you have in some code earlier: `cfg.trainer.precision in [16, '16-mixed', '16-true']`

examples/nlp/language_modeling/megatron_t5_lm_adaptation_finetune.py
```diff
@@ -67,7 +67,7 @@ def main(cfg) -> None:
     if cfg.model.resume_from_checkpoint is not None:
         resume_from_checkpoint = cfg.model.resume_from_checkpoint
     else:
-        resume_from_checkpoint = trainer._checkpoint_connector.resume_from_checkpoint_fit_path
+        resume_from_checkpoint = trainer._checkpoint_connector._ckpt_path
     logging.info(f'Resuming training from checkpoint: {resume_from_checkpoint}')
 
     trainer._checkpoint_connector = CheckpointConnector(trainer, resume_from_checkpoint=resume_from_checkpoint)
```
ericharper, line 73: `_CheckpointConnector`

nemo/collections/nlp/models/language_modeling/megatron_bert_model.py
```diff
@@ -90,14 +89,14 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
...
             self.autocast_dtype = torch.bfloat16
-        elif int(self.trainer.precision) == 32:
+        elif str(self.trainer.precision)[0:2] == '32':
             self.autocast_dtype = torch.float
-        elif int(self.trainer.precision) == 16:
+        elif str(self.trainer.precision)[0:2] == '16':
             self.autocast_dtype = torch.half
         else:
-            raise ValueError('precision must be in [32, 16, "bf16"]')
+            raise ValueError('precision must be in [32, "32-true", 16, "16-mixed", "bf16", "bf16-mixed"]')
```
ericharper, line 99: Let's update the error message so it suggest to use the values that won't be deprecated soon:
`["32-true", "16-mixed", "bf16-mixed"]

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
```diff
@@ -552,14 +577,20 @@ def write_predictions_to_file(self, outputs, output_file_path_prefix):
     def validation_step(self, dataloader_iter, batch_idx, dataloader_idx=0):
         return self.inference_step(dataloader_iter, batch_idx, 'validation', dataloader_idx)
 
-    def validation_epoch_end(self, outputs):
-        _ = self.inference_epoch_end(outputs, 'validation', self.cfg.data.validation_ds)
+    def on_validation_epoch_end(self):
+        _ = self.inference_epoch_end(self.validation_step_outputs, 'validation', self.cfg.data.validation_ds)
+        self.validation_step_outputs.clear()
```
aklife97, line 582: is it safe to run `.clear()` on `validation_step_outputs` ?
it looks like validation_step_outputs is created as a nested list in case of multiple dataloaders in init of ModelPT. Clearing it here will make it an empty list without any sublists, possibly causing errors when you try to append by specifying indexes in `validation_step`?

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
```diff
@@ -552,14 +577,20 @@ def write_predictions_to_file(self, outputs, output_file_path_prefix):
...
+        # return super().on_validation_epoch_end()
 
     def test_step(self, dataloader_iter, batch_idx, dataloader_idx=0):
         return self.inference_step(dataloader_iter, batch_idx, 'test', dataloader_idx)
 
-    def test_epoch_end(self, outputs):
-        _ = self.inference_epoch_end(outputs, 'test', self.cfg.data.test_ds)
+    def on_test_epoch_end(self):
+        _ = self.inference_epoch_end(self.test_step_outputs, 'test', self.cfg.data.test_ds)
+        self.test_step_outputs.clear()
```
aklife97, line 591: same concern as clearing `validation_step_outputs`

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -424,13 +440,13 @@ def on_validation_epoch_start(self) -> None:
         self._reconfigure_batch_sizes(gbs, mbs)
         return super().on_validation_epoch_start()
 
-    def validation_epoch_end(self, outputs):
-        if len(outputs) == 0:
+    def on_validation_epoch_end(self):
+        if len(self.validation_step_outputs) == 0:
```
ericharper, line 444: Nit: `self.validation_step_outputs` is initialized to `[]` ? So can we just do 
```
if not self.validation_step_outputs:
```

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -93,6 +94,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
         self.original_checkpointing_num_layers = base_module.language_model.encoder.activations_checkpoint_num_layers
         self.original_checkpointing_method = base_module.language_model.encoder.activations_checkpoint_method
         self.virtual_tokens = 0
+        # Create a different self.validation_step_outputs as we call val step of parent class(MegatronGPTModel) here,
+        # that appends to self.validation_step_outputs leading to wrong values in self.validation_step_outputs
+        self.validation_step_outputs_sft = []
```
ericharper, line 99: Do we need both `self.validation_step_outputs` (from the parent) and `self.validation_step_outputs_sft` (from the child)?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -180,6 +184,15 @@ def setup(self, stage=None):
...
             self._test_dl = self.setup_eval_dataloader(self._test_ds, self.cfg.data.test_ds)
 
+        # If setting val and test dataloaders via this method, create self.validation_step_outputs and self.test_step_outputs
+        if type(self._validation_dl) == list and len(self._validation_dl) > 1:
+            for _ in range(len(self._validation_dl)):
+                self.validation_step_outputs_sft.append([])
+
+        if type(self._test_dl) == list and len(self._test_dl) > 1:
+            for _ in range(len(self._test_dl)):
+                self.test_step_outputs.append([])
```
ericharper, line 194: I thought multiple dataloaders wasn't supported yet by PTL 2.0 when using `dataloader_iter` ?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -359,88 +372,100 @@ def fwd_bwd_step(self, dataloader_iter, batch_idx, forward_only):
...
-                metric_name=self.val_metric_name if mode == 'validation' else self.test_metric_name,
-                class_labels=self.cfg.data.validation_ds.metric.get('class_labels', None)
-                if mode == 'validation'
-                else self.cfg.data.test_ds.metric.get('class_labels', None),
-                labels_are_strings=self.cfg.data.validation_ds.metric.get('labels_are_strings', False)
-                if mode == 'validation'
-                else self.cfg.data.test_ds.metric.get('labels_are_strings', False),
+        # loss can be None as super().validation_step returns None when dataloader_iter is exhausted
+        # which can lead to error, adding check to prevent it
+        if not loss is None:  # Ensure its not None
```
ericharper, line 389: `if loss is not None` ?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -359,88 +372,100 @@ def fwd_bwd_step(self, dataloader_iter, batch_idx, forward_only):
...
+                if type(self.trainer.test_dataloaders) == list and len(self.trainer.test_dataloaders) > 1:
+                    self.test_step_outputs[dataloader_idx].append(outputs)
+                else:
+                    self.test_step_outputs.append(outputs)
+            return outputs
+            # TODO (sandeepsub): Figure out the subsequent decode bits.
+            length_params: LengthParam = {
+                "min_length": 0,
+                "max_length": batch['tokens'].size(1) - batch['context_lengths'].max(),
+            }
```
ericharper, line 411: Seems there's a problem with the formatting since adding the conditional

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
ericharper: I think we are missing an `else` for this conditional or there's some formatting error

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
```diff
@@ -662,39 +664,57 @@ def validation_step(self, dataloader_iter, batch_idx, dataloader_idx=0):
...
+            else:
+                self.test_step_outputs.append(loss)
 
-    def validation_epoch_end(self, outputs):
+        return loss
+
+    def on_validation_epoch_end(self):
         # NOTE: we need to make sure outputs is not empty (this is a workaround for a bug in pytorch lightning (?))
-        if len(outputs) == 0:
+        if len(self.validation_step_outputs) == 0:
```
ericharper, line 688: `if not self.validation_step_outputs:` ?

nemo/core/classes/modelPT.py
```diff
@@ -177,12 +178,39 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
                     f"Test config : \n{OmegaConf.to_yaml(self._cfg.test_ds)}"
                 )
 
+        # Create list of lists for val and test outputs to support multiple dataloaders
+        self.validation_step_outputs = (
+            []
+        )  # Initialize an empty list as sometimes self._validation_dl can be None at this stage
```
ericharper, line 184: can we move this comment above line 182 so we don't have the extra formatting for a long line?

Jenkinsfile
```diff
@@ -3729,8 +3731,8 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         model.data.validation_ds.micro_batch_size=1 \
         model.data.validation_ds.global_batch_size=4 \
         model.data.validation_ds.num_workers=0 \
-        model.data.validation_ds.file_names=[/home/TestData/nlp/megatron_sft/quarel.jsonl,/home/TestData/nlp/megatron_sft/trec.jsonl] \
```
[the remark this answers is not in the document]
athitten: @aklife97 thanks for the review! 
The issue here is Lightning 2.0 does not support multiple dataloaders. Lightning dropped the support in 2.0 since they thought it was too fragile in 1.9. I have asked lightning to provide a fix for this, since we consider it important. But until then it's a quick fix to disable the second validation dataset file here. Checked with MLPerf team and they are not using multiple dataloaders in their NeMo runs.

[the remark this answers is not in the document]
athitten: Also this only when we are using `dataloader_iter`. 
[https://github.com/Lightning-AI/lightning/blob/master/src/lightning/pytorch/loops/evaluation_loop.py#L237](url)

ericharper (changes requested):
I reviewed NLP and Core and overall I think the PR is very close. I left a few comments which I hope we can resolve and there are still pending comments from other reviewers. I'll try and summarize some action items.

- `_CheckpointConnector` and `_ckpt_path`. I believe we can get rid of all of these private methods. It's okay with me to do it in a subsequent PR.
- Precision configuration. I think for megatron models we should add a method to the base class which handles this logic as @aklife97  suggested. Can be done in a subsequent PR.
- `dataloader_iter` try catch. As @aklife97 pointed out this looks pretty ugly in the code. @awaelchli , is it possible for PTL to help with a more long-term solution?
- `dataloader_iter` multiple dataloaders, this seems to be the only feature we lost when upgrading to 2.0, @awaelchli  is it possible to get this feature back?

Jenkinsfile
```diff
@@ -3202,7 +3204,6 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
-        trainer.limit_val_batches=1 \
```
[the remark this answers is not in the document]
athitten: @aklife97 thanks, adding it back in. During the migration to 2.0 there was an issue using this and hence was removed temporarily. Missed to add it back.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
```diff
@@ -168,38 +169,10 @@ def on_test_epoch_start(self):
         )
         return super().on_test_epoch_start()
 
-    def on_test_epoch_end(self):
```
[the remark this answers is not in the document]
athitten: @aklife97 this is not removed. We had two functions `test_epoch_end` and `on_test_epoch_end`. With lightning 2.0 since `test_epoch_end` callback is deprecated and renamed to `on_test_epoch_end`, I have merged the functionality of previous `test_epoch_end` into `on_test_epoch_end`. You can find the configuration of batch size being handled here in `inference_epoch_end` [https://github.com/NVIDIA/NeMo/blob/athitten/ptl_2.0/nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py#L549](url)

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
athitten: @aklife97 very good catch, thanks! It needs to be cleared as `self.validation_step_outputs[dataloader_idx].clear()`. Let me fix it

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
athitten: Yes fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
athitten: Fixed.

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
athitten: Thanks @ericharper, this is added in the latest commit.

examples/nlp/language_modeling/megatron_retro_fine_tune.py
athitten: Fixed.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
athitten: Fixed in the latest commit.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
athitten: Fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
athitten: Yes that's right. Added it more for futuristic purpose. Let me know if we want to remove it.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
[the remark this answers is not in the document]
aklife97: sounds good, thanks for the clarification!

Jenkinsfile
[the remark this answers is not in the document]
aklife97: got it!

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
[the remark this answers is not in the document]
athitten: Resolved offline.

nemo/collections/nlp/parts/nlp_overrides.py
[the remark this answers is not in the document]
athitten: Resolved offline.

Jenkinsfile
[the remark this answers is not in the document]
athitten: @aklife97 sorry my bad.Actually this trainer.limit_val_batches=1 needs to be removed or made as 2 (which is the num of val batches). Otherwise, there's an error with the datalaoder_iter.

nemo/core/classes/modelPT.py
athitten: Fixed

nemo/core/classes/modelPT.py
athitten: Fixed.

tests/core/test_config_utils.py
athitten: Fixed.

nemo/collections/nlp/models/language_modeling/megatron_base_prompt_learning_model.py
```diff
@@ -133,14 +132,16 @@ def init_model(self, cfg: DictConfig, trainer: Trainer):
         self.pad_token_id = self.tokenizer.pad_id if self.tokenizer.pad_id is not None else self.tokenizer.unk_id
         self.decoder_seq_length = cfg.get('decoder_seq_length', 40)
 
-        if self.trainer.precision == 'bf16':
+        # Check for first few chars to be '16', '32' or 'bf16' as Precision can also be '16-mixed', '32-true' or 'bf16-mixed'
```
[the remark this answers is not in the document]
athitten: Deferring this to a subsequent PR.

nemo/collections/nlp/models/language_modeling/megatron_base_prompt_learning_model.py
athitten: Fixed.

examples/nlp/language_modeling/megatron_t5_lm_adaptation_finetune.py
athitten: Fixed

nemo/collections/nlp/models/language_modeling/megatron_bert_model.py
athitten: Fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
athitten: Discussed offline and fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
athitten: Discussed offline and fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
athitten: Discussed offline and fixed.

nemo/core/classes/modelPT.py
athitten: Fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
athitten: Fixed and discussed offline.

nemo/core/classes/modelPT.py
athitten: Discussed offline to add on_train_start, on_validation_start hooks to ModelPT and perform the check for multi dataloaders in the hooks. Deferring it to a subsequent PR.

nemo/collections/asr/models/classification_models.py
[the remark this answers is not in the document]
athitten: Discussed offline to add on_train_start, on_validation_start hooks to ModelPT and perform the check for multi dataloaders in the hooks. Deferring it to a subsequent PR.

examples/nlp/language_modeling/megatron_bert_pretraining.py
athitten: Deferring to a subsequent PR as removing this is causing errors hence needs further investigation.

examples/nlp/language_modeling/megatron_gpt_pretraining.py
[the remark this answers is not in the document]
athitten: Deferring to a subsequent PR.

nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py
athitten: Deferred to subsequent PR.

examples/nlp/language_modeling/megatron_bart_pretraining.py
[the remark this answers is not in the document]
athitten: Yes incorporated this.

Jenkinsfile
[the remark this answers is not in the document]
aklife97: setting it to 2 sounds good, can we also change other tests that remove it?

aklife97:
Adding to @ericharper's list of pending action items that need to be resolved:

1. set back `trainer.limit_val_batches` - will improve CI runtime
2. add on_train_start, on_validation_start hooks to ModelPT

We can keep addressing these 6 action items in subsequent PRs. I think the PR looks good to go, @ericharper

ericharper (approved):

ericharper:
LGTM. Thanks for the massive effort!

PR #6436, opened 2023-04-17
Update NeMo_TTS_Primer.ipynb

pythinker:
# What does this PR do ?

Fix a mistake in line 782. FFT calculates the magnitude and phase of each frequency bin not frequency band. Moreover, frequency bins in FFT are not related to pitch. So, we should change _**band (ie. pitch)**_ to _**bin**_. 

**Collection**: [TTS]

# Changelog 
tutorials/tts/NeMo_TTS_Primer.ipynb
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [x] Documentation

## Who can review?
@blisc & @okuchaiev


rlangman:
Thank you for the correction.

My understanding is that the FFT does produce a series of frequency **bins**. When we convert to the mel spectrogram, we combine FFT **bins** into a smaller number frequency **bands** and translate them on the **mel frequency** scale so that there is a roughly linear relationship between the **mel frequency** and perceived **pitch**.

Does that sound correct? In which case, the term "frequency band" in bullet point (3) is still used correctly even if we change (2) to "bin".

pythinker:
@rlangman 

Perceived pitch is related to the harmonic structure of human speech while mel frequency scale is inspired by how human ear divides the spectrogram into different frequency bands so I think there is no relationship between the mel frequency and perceived pitch. Moreover, note that pitch is only defined for voiced sounds while unvoiced sounds have no harmonic structure and therefore no pitch while we still use mel frequency scale to analyze them.

More importantly, going from spectrogram to mel-spectrogram is not done by translating frequency band from hertz (Hz) to units of mel frequency, it's more complicated. It's done by [matrix multiplication of spectrogram (magnitude squared) and mel filter-banks](https://pytorch.org/audio/main/_modules/torchaudio/transforms/_transforms.html#MelSpectrogram:~:text=mel_specgram%20%3D%20torch.matmul(specgram.transpose(%2D1%2C%20%2D2)%2C%20self.fb).transpose(%2D1%2C%20%2D2)). The point is that frequency bands are defined by mel filters not by translating Hz to mel scale.

We should not only change _**point (3)**_, but also change _**point (2)**_ and add another point:
_**2. For each audio frame, use the Fast Fourier transform (FFT) to calculate the magnitude (ie. energy, amplitude or "loudness") and phase (which we don't use) of each frequency bin (N bins in total). Get spectrogram by squaring the magnitude of frequency bins.**_

_**3. Define M mel filters (and therefore M frequency bands) by sliding a triangular filter in frequency and choosing its center and width based on mel scale.**_

_**4. In each frequency band defined by the center and width of each mel filter, compute a new number by multiplication of spectrogram and filter coefficients at each bin at the band and summation over all bins in band. The output is called a mel spectrogram.**_

rlangman:
> Perceived pitch is related to the harmonic structure of human speech while mel frequency scale is inspired by how human ear divides the spectrogram into different frequency bands so I think there is no relationship between the mel frequency and perceived pitch. Moreover, note that pitch is only defined for voiced sounds while unvoiced sounds have no harmonic structure and therefore no pitch while we still use mel frequency scale to analyze them.

According to https://en.wikipedia.org/wiki/Mel_scale: _mel scale (after the word melody) is a perceptual scale of pitches judged by listeners to be equal in distance from one another._

However the pitch it is referring to is the non-quantifiable version from music theory. From https://en.wikipedia.org/wiki/Pitch_(music): _Pitch may be quantified as a frequency, but pitch is not a purely objective physical property; it is a subjective psychoacoustical attribute of sound._

This may be confusing because a lot of people who do speech processing today use the term as a synonym for F0, which is a very different quantifiable physical property (and where the distinction between voiced and unvoiced comes in).

So originally this (and other NeMo tutorials like ASR) used the term "pitch" to describe frequency because it makes it more understandable to people without a background in speech processing. However it is also ambiguous/confusing in this context, so I do think it would be better to leave out any comparison between frequency and pitch.

> Get spectrogram by squaring the magnitude of frequency bins.

Technically most public TTS models use the magnitude spectrogram. Though with deep learning models pretty much any representation works the about the same including the square/power spectrogram or even the linear (non-mel) spectrogram.

> We should not only change point (3), but also change point (2) and add another point:

> 2. For each audio frame, use the Fast Fourier transform (FFT) to calculate the magnitude (ie. energy, amplitude or "loudness") and phase (which we don't use) of each frequency bin (N bins in total). Get spectrogram by squaring the magnitude of frequency bins.
> 
> 3. Define M mel filters (and therefore M frequency bands) by sliding a triangular filter in frequency and choosing its center and width based on mel scale.
> 
> 4. In each frequency band defined by the center and width of each mel filter, compute a new number by multiplication of spectrogram and filter coefficients at each bin at the band and summation over all bins in band. The output is called a mel spectrogram.

I think the wording is a little too low level for the intended audience of the primer. If someone wants to know the exact mathematical steps and dimensions and such, then we can link to a separate reference that goes in detail (if you know of one).

As another example, we could also specify that the FFT output is a complex number with the magnitude being the absolute value and the phase the arctangent, but I am not sure that information is useful here.

But the description should be accurate, albeit intuitive. Would something like this be reasonable?

> 2. For each audio frame, use the Fast Fourier transform (FFT) to calculate the magnitude (ie. energy, amplitude or "loudness") and phase (which we don't use) of each frequency bin. We refer to the magnitudes of the frequency bins as a **spectrogram**.
> 3. Map the original frequency bins onto the [mel scale](https://en.wikipedia.org/wiki/Mel_scale), using overlapped [triangular filters](https://en.wikipedia.org/wiki/Window_function#Triangular_window) to create _mel filterbanks_.
> 4. Multiply the original spectrogram by the mel filterbanks to produce a mel spectrogram.

pythinker:
@rlangman 

I just committed the additional description updates and MATLAB reference to this PR and I prefer to merge this pull request to the main branch of NeMo. So, please don't open a separate PR.

pythinker:
@rlangman 
@ericharper 

Who would give this commit an approval? Does it need another review while pending approval?

rlangman (approved):
Looks good to me.

But the CI check is failing. I suggest rebasing first to see if that fixes it.

pythinker:
@rlangman 
@ericharper 

I rebased and all checks are passed. Could you please merge this pull request?

PR #6438, opened 2023-04-18
add config and char tokenizer for russian tts

bene-ges:
# What does this PR do ?

Add config and char tokenizer for russian tts. It contains russian alphabet + ipa alphabet. Each unicode character is regarded as separate token, including stress.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -238,6 +238,72 @@ def __init__(
...
+        """Russian phoneme-based tokenizer.
+        Args:
+            punct: Whether to reserve grapheme for basic punctuation or not.
+            apostrophe: Whether to use apostrophe or not.
+            add_blank_at: Add blank to labels in the specified order ("last") or after tokens (any non None),
+             if None then no blank in labels.
+            pad_with_space: Whether to pad text with spaces at the beginning and at the end or not.
+            non_default_punct_list: List of punctuation marks which will be used instead default.
+            text_preprocessing_func: Text preprocessing function for correct execution of the tokenizer.
+             Currently, it only applies lower() function.
```
redoctopus, line 260: Maybe: s/Currently/By default

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -238,6 +238,72 @@ def __init__(
...
+        """
+
+        ru_ipa = "əɪɐɨaʊioueʉæɵɛ" + "vrmtpsnʂkflzɫbdjɕxɡʐ" + "аеёиоуыэюя" + "бвгджзйклмнпрстфхцчшщъь"
+        ru_ipa += "ʲ"  # soft
+        ru_ipa += "`"  # stress
+        ru_ipa += "ː̃"  # long
```
redoctopus, line 266: These are Russian characters, right? Should this be called `ru_chars` instead of `ru_ipa`?

bene-ges: These are symbols from ipa **and** russian alphabet. The idea is that most input words will be converted to ipa but some will stay as graphemes (e.g. heteronyms)

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
bene-ges: done

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
redoctopus: Ahh. Could you merge this class with the existing IPATokenizer class with the appropriate `RU-ru` locale character sets in `<base_dir>/collections/common/tokenizers/text_to_speechipa_lexicon.py`? Is there anything in the design of the IPATokenizer that prevents this?

bene-ges: I see 2 problems with using class IPATokenizer
1. It requires a g2p as parameter, but I use an external g2p module. 
2. In ipa_lexicon only standardised IPA are used, for example, as cited in Wikipedia.

I do not pretend that my phoneme set is exacty "gold standard". If I just want TTS to work with my particular character set - how can I achieve that? Is it possible to somehow pass my alphabet through a config, like we can do in ASR?

redoctopus: I see... the IPATokenizer does allow users to pass in a `fixed_vocab` argument with the character set you want to use, but that doesn't fix the external G2P module issue.

We do have the Chinese model using a different setup than IPATokenizer, so we can probably do the same for the Russian model. @XuesongYang do you have any guidelines/best practices to keep in mind if we move forward with a separate Russian TTS model?

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -238,6 +238,71 @@ def __init__(
...
+    def __init__(
+        self,
+        punct=True,
+        apostrophe=True,
+        add_blank_at=None,
+        pad_with_space=False,
+        non_default_punct_list=None,
+        text_preprocessing_func=any_locale_text_preprocessing,
+    ):
+        """Russian phoneme-based tokenizer.
```
racoiaws, line 251: Could you change this description to be more accurate?

Input text is assumed to be a mixture of IPA symbols (and cyrillic characters for words not handled by G2P).

Note that neither are phonemes. Russian has 5 to 6 vowel phonemes (depending on who you ask). Here I see a lot of tokens for vowels (əɪɐɨaʊioueʉæɵɛ).

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
racoiaws: Code in the `IPATokenizer` is concerned with
1) Creating a mapping `token -> integer id`; tokens are phones, graphemes and service tokens like spaces
2) Calling G2P (`string -> list of strings`)
3) Converting sequence of tokens to sequence of integer ids (`list of string -> list of integer ids`)

I haven't found any evidence it uses the IPA lexicons mentioned above in any way. So far I understand it as a wrapper for G2P. So, you can't just pass `fixed_vocab`. I'm perplexed at how that argument is supposed to be used.

So, to use `IPATokenizer` we would need to create something like a `PreprocessedRussianG2P`, that
1) Has a vocabulary (with phonetic symbols prefixed by "#")
2) Splits input text according to this vocabulary
3) Conforms to the G2P interface

Bonus points if it builds its vocabulary from IPA lexicons defined in `collections/common/tokenizers/text_to_speechipa_lexicon.py`

racoiaws: Tokenization means "splitting string into tokens".  `IPATokenizer` does not tokenize anything. A fitting `PreprocessedRussianG2P` would not do any actual G2P.

I can definitely see how this might be confusing for somebody adding a new language.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -238,6 +238,71 @@ def __init__(
...
+            punct: Whether to reserve grapheme for basic punctuation or not.
+            apostrophe: Whether to use apostrophe or not.
+            add_blank_at: Add blank to labels in the specified order ("last") or after tokens (any non None),
+             if None then no blank in labels.
+            pad_with_space: Whether to pad text with spaces at the beginning and at the end or not.
+            non_default_punct_list: List of punctuation marks which will be used instead default.
+            text_preprocessing_func: Text preprocessing function for correct execution of the tokenizer. By default, it only applies lower() function.
+        """
+
+        ru_ipa = "əɪɐɨaʊioueʉæɵɛ" + "vrmtpsnʂkflzɫbdjɕxɡʐ" + "аеёиоуыэюя" + "бвгджзйклмнпрстфхцчшщъь"
```
racoiaws, line 262: This assumes that every symbol in the strings is its own token. There are some caveats with how unicode handles characters with diacritics, so I would advise against this.

A better practice would be to have an honest list of tokens like in `collections/common/tokenizers/text_to_speechipa_lexicon.py`, e.g.
```
ru_tokens = ["ə", "ɪ", "ɐ", ...]
```

It is a bit unwieldy but will save debugging time down the road.

bene-ges: @racoiaws , what do you think about the idea to create a subclass of BaseTokenizer that takes the alphabet from the config (like in ASR)? In this case **no need** to add dummy g2p classes and alphabets to the python code. It also allows to have models with slightly different alphabets for the same language. Wouldn't it be an easier way to add custom alphabets for other languages?

racoiaws: Yes, that sounds great for experiments with pre-processed texts or external g2p.

It would need a list of symbols, a list of phonetic tokens, and some convention how to disambiguate between them in input strings (like phonetic symbols starting with # in IPATokenizer).

Note that the default NeMo way is still to do g2p on-the-fly, so when loading a pretrained model people could just run it on some text.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
redoctopus: `fixed_vocab` is a bit of a hacky workaround to allow a user-set vocab--it lets the G2P class build the list of tokens and then replaces it by calling [`replace_symbols()`](https://github.com/NVIDIA/NeMo/blob/cc0cbd3bf8d11856ecd29efcded7fd9cd1a354d0/nemo/collections/tts/g2p/models/i18n_ipa.py#L292), which scrubs through the dictionary. It's set at the tokenizer level so that any illegal tokens can be discarded before being assigned some int mapping.

bene-ges: @racoiaws @redoctopus I found a way to pass custom alphabet to BaseCharsTokenizer without any code change, just modifying yaml config.
```
  text_tokenizer:
    _target_: nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers.BaseCharsTokenizer
    chars: ["ʲ","ː","ə","ɪ","ɐ","ɨ","a","ʊ","i","o","u","e","ʉ","æ","ɵ","ɛ","v","r","m","t","p","s","n","ʂ","k","f","l","z","ɫ","b","d","j","ɕ","x","ɡ","ʐ","а","е","ё","и","о","у","ы","э","ю","я","б","в","г","д","ж","з","й","к","л","м","н","п","р","с","т","ф","х","ц","ч","ш","щ","ъ","ь"]
    non_default_punct_list: [',','.','!','?','-',':',';','/','"','(',')','[',']','{','}',"`","̃"]
    punct: true
    apostrophe: true
    pad_with_space: true
```
this works for my TTS recipe with external g2p.

The only observation is that BaseCharTokenizer's `.encode` method requires characters to be either `isalnum()` or be listed as punctuation symbols. So I have to override punctuation symbols too.

If you have no objections I will close this PR

redoctopus: > The only observation is that BaseCharTokenizer's .encode method requires characters to be either isalnum() or be listed as punctuation symbols. So I have to override punctuation symbols too.

Oof, that's unfortunate. I guess we'll probably run into this with some other non-English character sets too sometime, so we may want to revisit `encode()` to see if we can generalize it somehow.

I'm fine with this workaround if @racoiaws is okay with it, though! Could you add the working YAML config in a new PR if so?

bene-ges: @redoctopus, @racoiaws I put a [full recipe](https://github.com/bene-ges/nemo_compatible/blob/main/scripts/tts/ru_ipa_fastpitch_hifigan/train.sh) for Russian TTS data preparation and training to my own repository. It is compatible witn Nemo main branch. The recipe includes several python scripts, two config files and a bash-script example. 
I think it's no use to add a single config file to nemo - I can either add the whole thing or leave it in external repo.

redoctopus: Yes, please add the whole thing!

bene-ges:
Closing this PR because I found a way to pass custom alphabet to BaseCharsTokenizer without any code change, just modifying yaml config.
My external [full recipe](https://github.com/bene-ges/nemo_compatible/blob/main/scripts/tts/ru_ipa_fastpitch_hifigan/train.sh) for Russian TTS data preparation is compatible witn Nemo main branch.

PR #6466, opened 2023-04-20
Fix cache aware hybrid bugs

VahidooX:
# What does this PR do ?

Fixes some bugs to support Hybrid and FastConformer variants of cache-aware streaming models.

# Changelog 
- Fixed a bug which prevented switching between decoders for Hybrid models
- Added support of Hybrid models to the cache-aware streaming inference scripts
- Added the feature to switch between decoders to the cache-aware streaming simulation script
- Added support of causal downsamling for dw_striding approach.

# Usage
Users may use the --set_decoder "rnnt" or "ctc" to change the decoder with speech_to_text_cache_aware_streaming_infer.py.
They may use asr_model.change_decoding_strategy(decoder_type='ctc' or 'rnnt') to change the decoding for a Hybrid model in general.

**PR Type**:
- [X] New Feature
- [X] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
+++ b/examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
@@ -286,8 +286,10 @@ def main():
 
     logging.info(asr_model.encoder.streaming_cfg)
     if args.set_decoder is not None:
-        asr_model.change_decoding_strategy(decoder_type=args.set_decoder)
-        # asr_model.cur_decoder = args.set_decoder
+        if hasattr(asr_model, "cur_decoder"):
+            asr_model.change_decoding_strategy(decoder_type=args.set_decoder)
+        else:
+            raise ValueError("Decoder cannot get changed for non-Hybrid ASR models.")
 
     global autocast
     if (
--- a/examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
+++ b/examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
@@ -42,7 +42,9 @@ python speech_to_text_streaming_infer.py \
 
 You may drop the '--debug_mode' and '--compare_vs_offline' to speedup the streaming evaluation.
 If compare_vs_offline is not used, then significantly larger batch_size can be used.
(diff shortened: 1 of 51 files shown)


examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
```diff
@@ -259,6 +264,13 @@ def main():
         help="Enables padding the audio input and then dropping the extra steps after the pre-encoding for the first step. It makes the outputs of the downsampling exactly as the offline mode for some techniques like striding.",
     )
 
+    parser.add_argument(
+        "--set_decoder",
+        choices=["ctc", "rnnt"],
```
nithinraok, line 269: why not give 'rnnt+ctc' option to decoder as well?

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -126,42 +126,72 @@ def __init__(
...
-                    stride=self._stride,
-                    padding=self._left_padding,
+            if self.is_causal:
+                layers.append(
+                    CausalConv2D(
+                        in_channels=in_channels,
+                        out_channels=conv_channels,
+                        kernel_size=self._kernel_size,
+                        stride=self._stride,
+                        padding=None,
```
nithinraok, line 146: is padding not required for causal convs?

VahidooX: Causal needs padding but when padding is passed as None, it would use the default padding (kernel_size-1) to make it causal. User does not need to worry about the right number of paddings.

examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py
VahidooX: That option is not feasible as just one decoder can be used for inference time. To use both, the whole inference process needs to be done twice with different decoders.

nithinraok (approved):

nithinraok:
LGTM, one of the test is failing related to hybrid asr is failing

nithinraok (approved):

titu1994 (approved):
Very nice! But we need have a way to track changes in models - having to do these partial updates for multiple models individually is error prone.

PR #6498, opened 2023-04-26
Add interleaved pp support

titu1994:
# What does this PR do ?

Add support for virtual pipeline parallelism conversion of Megatron GPT models

**Collection**: [NLP]

# Changelog 
- 

# Usage

```python

python megatron_change_num_partitions.py \
    --model_extracted_dir="<Directory of Pytorch Lightning ckpt folders>" \
    --target_file="<Name of the target NeMo file>" \
    --ckpt_name="<Name of a single Pytorch Lightning ckpt file inside the extracted dir>" \
    --tensor_model_parallel_size=<TP SIZE> \
    --target_tensor_model_parallel_size=<TARGET TP SIZE> \
    --pipeline_model_parallel_size=<PP SIZE> \
    --target_pipeline_model_parallel_size=<TARGET PP SIZE> \
    --virtual_pipeline_model_parallel_size=<VP SIZE> \
    --hparams_file="<Path to HPARAMS.yaml file>" \
    --precision=bf16

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/nlp/language_modeling/megatron_change_num_partitions.py
+++ b/examples/nlp/language_modeling/megatron_change_num_partitions.py
@@ -13,10 +13,13 @@
 # limitations under the License.
 
 import os
+import tempfile
 from argparse import ArgumentParser
 from typing import Dict, List
+from omegaconf import OmegaConf
 
 import torch
+import torch.nn as nn
 from omegaconf import open_dict
 from pytorch_lightning import Trainer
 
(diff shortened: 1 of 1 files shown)


examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -830,98 +909,215 @@ def main():
...
+                    app_state.local_rank = global_rank % num_gpu_per_node
+                    app_state.pipeline_model_parallel_size = pp_size
+                    app_state.tensor_model_parallel_size = tp_size
+                    app_state.pipeline_model_parallel_split_rank = pipeline_model_parallel_split_rank
+                    app_state.model_parallel_size = (
+                        app_state.pipeline_model_parallel_size * app_state.tensor_model_parallel_size
+                    )
+
+                    if vp_size > 1:
+                        set_virtual_parallel_rank_safely(vp_rank)
```
aklife97, line 980: why are we doing this again? we seem to have already set it the outside loop

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -1046,10 +1281,16 @@ def main():
                 with open_dict(model.cfg):
                     model.cfg.tokenizer.model = args.tokenizer_model_path
 
-            model = cls(model.cfg, trainer).to('cpu')
+            model.cfg, restore_dict = force_cpu_model(model.cfg)
+
+            model = cls(model.cfg, trainer)
+            model = model.to('cpu')
```
aklife97, line 1287: why do we need this?

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: This gets reset to different values in parts of the script so for safety i reset this back to the value outside the loop. It can possibly be removed, but better keep it.

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: Due to NeMo bug in some layers, not all layers will be on CPU even after explicit setting of config to load on CPU - this is to force the remainder of the layers to move to cpu.

aklife97: ahh okay ://
can we go OOM because of this?

aklife97 (approved):
LGTM, thanks for the PR!

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: Thankfully for GPT, only one of the embedding layers has this issue, most other layers respect the cpu_init flag. So rarely possibly to get OOM due to this on GPT. T5 yes quite possible cause there's a bunch of places that's ignoring the flag.

PR #6507, opened 2023-04-27
[TTS] Create initial TTS dataset feature processors

rlangman:
# What does this PR do ?

Add an interface and initial implementations to process features at training time, as part of TTS data loading refactoring.

The intention is for the data loader to load and merge feature dictionaries (produced by tts.parts.preprocessing.feature.py) and to use these processors to apply on necessary on the fly calculations, such as feature normalization.

I can provide a diff for the data loader itself if it helps provide context. Otherwise I will post PR for the data loader after this one, as it depends on this change.

**Collection**: [TTS]

# Changelog 
- Define interface to processing dataset features
- Provide implementation for feature scaling, log compression, and mean-variance normalization.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/parts/preprocessing/feature_processors.py
+++ b/nemo/collections/tts/parts/preprocessing/feature_processors.py
@@ -178,7 +178,7 @@ class MeanVarianceSpeakerNormalization(FeatureProcessor):
             stats_path: JSON file with feature mean and variance.
             speaker_field: field containing speaker ID string.
             mask_field: Optional, field in example dictionary with boolean array indicating which values to
-                mask to 0. Defaults to 'voiced_mask', expected to be by pyin pitch estimator.
+                mask to 0. Defaults to 'voiced_mask', expected to be computed by pyin pitch estimator.
             fallback_to_default: Whether to use 'default' feature statistics when speaker is not found in
                 the statistics dictionary.
         """
--- a/nemo/collections/tts/parts/preprocessing/feature_processors.py
+++ b/nemo/collections/tts/parts/preprocessing/feature_processors.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import json
-import os
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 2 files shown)


nemo/collections/tts/parts/preprocessing/feature_processors.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+
+        self.field = field
+        self.mask_field = mask_field
+
+        if not os.path.exists(stats_path):
+            raise ValueError(f"Statistics file does not exist: {stats_path}")
+
+        with open(stats_path, 'r', encoding="utf-8") as pitch_f:
+            stats_dict = json.load(pitch_f)
```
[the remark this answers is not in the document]
rlangman: Fixed

nemo/collections/tts/parts/preprocessing/feature_processors.py
```diff
@@ -0,0 +1,223 @@
...
+        """
+
+        self.field = field
+        self.key_mean = f"{self.field}_mean"
+        self.key_std = f"{self.field}_std"
+        self.speaker_field = speaker_field
+        self.mask_field = mask_field
+        self.fallback_to_default = fallback_to_default
+
+        if not os.path.exists(stats_path):
```
racoiaws, line 197: This check is rather redundant, `open` will throw more or less the same thing if file does not exist
```
FileNotFoundError: [Errno 2] No such file or directory: <path>
```

Nitpick, not a blocker

racoiaws: https://docs.python.org/3/glossary.html#term-EAFP

rlangman: I was not aware until now of EAFP being a standard in Python. But since it is, then it makes sense to not have an explicit assert or try/catch block as the exception is more or less the same, as you pointed out.

racoiaws: I'm not talking about catching anything. Please see pseudocode below. I hope redundancy is more visible this way
```
def open(path):
    if not path.exists():
        raise FileNotFoundError
    ...

def do_things(path):
    if not path.exists():
        raise ValueError
    open(path)
```

racoiaws (approved):
LGTM

Experimental, so some leeway for later changes

PR #6509, opened 2023-04-27
[TTS] Add script for mapping speaker names to indices

rlangman:
# What does this PR do ?

The existing TTS workflow requires the user to provide speaker as an integer index in the training manifest. This is awkward because

1. User can provide raw ID from datasets like LibriTTS as an integer, but then we end up creating 10,000+ element speaker array for only hundreds of speakers.
2. User can manually create a map from speaker ID string to sequential integer index and then overwrite their training manifest.
3. Training with multiple manifests is awkward because you need to rewrite the speaker field for all manifests.

So the idea is for the new data loading logic, user will provide speaker as a string name/ID in the manifest and then provide a separate JSON file which maps the speaker IDs to integer IDs. The below script is an example of how to create that map.

**Collection**: [TTS]

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


scripts/dataset_processing/tts/create_speaker_map.py
```diff
@@ -0,0 +1,85 @@
...
+import argparse
+import json
+from pathlib import Path
+
+from nemo.collections.asr.parts.utils.manifest_utils import read_manifest
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter, description="Compute speaker level pitch statistics.",
```
redoctopus, line 46: Fix description

rlangman: Fixed

redoctopus (dismissed):

scripts/dataset_processing/tts/create_speaker_map.py
```diff
@@ -0,0 +1,96 @@
...
+from nemo.collections.asr.parts.utils.manifest_utils import read_manifest
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description="Create mapping from speaker names to numerical speaker indices.",
+    )
+    parser.add_argument(
+        "--manifest_paths", required=True, type=str, help="Path to training manifests, comma delimited",
```
racoiaws, line 50: Could you please change this to use an honest list, not split things by a comma?
Argparse allows to do this through [actions](https://docs.python.org/3/library/argparse.html#action)

Example:
```
import argparse

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--manifest_path", action="append") 

    args = parser.parse_args()
    print(args.manifest_path)
```
```
$ python test.py --manifest_path foo --manifest_path bar
['foo', 'bar']
```

racoiaws (changes requested):

racoiaws (approved):

PR #6526, opened 2023-05-01
Purely GPU-based optimized RNN-T greedy batched decoding

hainan-xv:
# What does this PR do ?

A highly optimized implementation for RNN-T greedy batched decoding with GPUs.

ASR

# Changelog 

# Usage

The method could be invoked by setting  rnnt_decoding.cuda_optimized as True, e.g.

```
python examples/asr/transcribe_speech.py \
        model_path=$model \
        dataset_manifest=$manifest  \
        rnnt_decoding.strategy=greedy_batch \
        +rnnt_decoding.cuda_optimized=True \
        batch_size=$batch \
        output_filename=$out_file
```

This gives significant inference speed-up for large batch sizes. E.g. Librispeech Transducer-large model with stateless decoders running inference on test-other, with batch=128,

| inference type  | WER | decoding time |
| --------------- | ----- | --------------- |
| baseline greedy batch | 4.70 | 119s |
| GPU optimized | 4.64 | 86s |

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/asr/modules/rnnt.py
```diff
@@ -368,6 +368,35 @@ def batch_concat_states(self, batch_states: List[List[torch.Tensor]]) -> List[to
 
         return state_list
 
+    def batch_subset_states(
```
titu1994, line 371: Add this function to the abstract base class - and preferably make it abstract to force all subclasses to implement it. Please provide concrete implementation for all subclasses if it is indeed added to the abstract base class.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -558,7 +560,13 @@ def __init__(
 
         # Depending on availability of `blank_as_pad` support
         # switch between more efficient batch decoding technique
-        if self.decoder.blank_as_pad:
+
+        if cuda_optimized:
+            if isinstance(decoder_model, StatelessTransducerDecoder):
```
titu1994, line 565: Dont import a class like this at the top. Import inside of this if clause instead - otherwise circular import issues may arise later. Also, i don't think `RNNTDecoder` is used.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+        x: torch.Tensor,
+        out_len: torch.Tensor,
+        device: torch.device,
+        max_symbols_per_hyp_factor: float = 0.5,
+        partial_hypotheses: Optional[List[rnnt_utils.Hypothesis]] = None,
+    ):
+        if partial_hypotheses is not None:
+            raise NotImplementedError("`partial_hypotheses` support is not supported")
+
+        with torch.inference_mode():
```
titu1994, line 629: Preferred to do just torch.no_grad() unless inferece mode is really needed - the modules are already frozen so they are effectively under inference mode. Tensors output of inference mode cannot be used for anything unless they are cloned.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+            # x: [B, T, D]
+            # out_len: [B]
+            # device: torch.device
+
+            # Initialize list of Hypothesis
+            batchsize = x.shape[0]
+
+            max_symbols_per_hyp = max(int(max_symbols_per_hyp_factor * x.shape[1]), 256)
```
titu1994, line 637: Revert max symbols per hyp to 10. Too large otherwise

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+            hyp2t = torch.zeros(
+                [batchsize], dtype=torch.long, device=x.device
+            )  # hyp2t[i] represent the time step of i'th hyp in kept_hyps
+
+            hyp2looptimes = torch.zeros(
+                [batchsize], dtype=torch.long, device=x.device
+            )  # to keep track of how many times the decoding stays at the same frame to compare with max_symbols_per_step
+
+            done = torch.zeros([batchsize], dtype=torch.bool, device=x.device)
+            hyp2b = torch.LongTensor([i for i in range(batchsize)]).to(x.device)
```
titu1994, line 664: Can use torch.tensor() rather than torch.LongTensor

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+
+            shift = hyp2b
+
+            x2 = torch.reshape(x, [-1, D])
+
+            bos = True
+            while True:
```
titu1994, line 671: Could you provide a detailed breakdown of decoding loop and every step inside. Basically please note the overall high level algorithm that is being done here and below. At the high level i do understand it but better to have it written down for future.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+                    hidden_prime = self.decoder.batch_copy_states(hidden_prime, None, blank_indices, value=0.0)
+
+                # Recover prior predicted label for all samples which predicted blank now/past
+                k[blank_indices] = last_label[blank_indices, 0]
+
+                # Update new label and hidden state for next iteration
+                last_label = k.clone().view(-1, 1)
+                hidden = hidden_prime
+
+                hyp2looptimes += 1 - k_is_blank
```
titu1994, line 723: Needs lots of detailed comments from here on out

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -607,6 +615,289 @@ def forward(
...
+
+                force_advance = hyp2looptimes == self.max_symbols
```
titu1994, line 866: Same as above, detailed comments needed

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -2202,3 +2494,4 @@ class GreedyBatchedRNNTInferConfig:
     preserve_alignments: bool = False
     preserve_frame_confidence: bool = False
     confidence_method_cfg: Optional[ConfidenceMethodConfig] = None
+    cuda_optimized: bool = True
```
titu1994, line 2497: Are the results exactly equivalent to greedy batched? If so this is fine, otherwise use default as false

nemo/collections/asr/parts/utils/rnnt_utils.py
```diff
@@ -140,6 +143,48 @@ def words(self) -> List[str]:
         return [] if self.text is None else self.text.split()
 
 
+# highly optimized Cuda structure representing hypotheses
+class CudaHypothesesStatelessTransducer:
```
titu1994, line 147: These two classes should come after NBestHypothesis

nemo/collections/asr/metrics/rnnt_wer.py
```diff
@@ -284,6 +284,7 @@ def __init__(self, decoding_cfg, decoder, joint, blank_id: int):
                     decoder_model=decoder,
                     joint_model=joint,
                     blank_index=self.blank_id,
+                    cuda_optimized=self.cfg.get('cuda_optimized', False),
```
titu1994, line 287: False as default here, but set to True in dataclass, so the dataclass value will always override this default here when you do change_decoding_strategy. Either set this to True to match dataclass, or update dataclass to match this.

titu1994 (changes requested):
Incredible perf improvement ! Some comments about doc, but the logic overall looks correct. We do need to check 1:1 exact check of results on some datasets compare to greedy batch. We should probably stick to this by default if the results are exact.

PR #6565, opened 2023-05-04
Update manifest.py for speedup

stevehuang52:
Re-order the checks for faster processing audio filepaths that are already absolute paths


titu1994 (approved):
Thanks !

PR #6574, opened 2023-05-05
Separate punctuation by whitespace

karpnv:
# What does this PR do ?

Fix separate_punctuation=True option 

**Collection**: ASR

# Changelog 
Do 
```
'some text.' -> 'some text .'
'some text .' -> some text .'
```

instead of 
```
'some text.' -> 'some text .'
'some text .' -> some text.'
```

**PR Type**:
- [V ] Bugfix

## Who can review?

@Kipok


Kipok:
Thanks @karpnv! Is it possible to add a couple of unit tests for this class to ensure that it works as expected on a few examples?

titu1994 (approved):
Thanks !

titu1994:
For unittest - best case would be to wait for the ASR post processor class to be developed and then we can add the regex + test it in isolation. This full script is hard to unittest

PR #6579, opened 2023-05-05
Online Code Switching Dataset for ASR

trias702:
# What does this PR do ?

Adds an IterableDataset class which generates on-the-fly code switched utterances for ASR

**Collection**: ASR

# Changelog 
- `nemo/collections/asr/data/audio_to_text.py`: added language support to the BPE classes to support code-switching
- `nemo/collections/asr/data/audio_to_text_dataset.py`: added the core create_cs_dataset function and various instantiation logic
- `nemo/collections/common/data/__init__.py`: exposed new CodeSwitchedDataset class
- `nemo/collections/common/data/dataset.py`: contains new CodeSwitchedDataset class
- `nemo/collections/common/parts/preprocessing/collections.py`: added discrete language support of the TokeniserWrapper via a hasattr check

- `nemo/collections/asr/models/*`: changed all of the models to check for IterableDataset instead of reading the is_tarred config item, as I think this makes it more robust

# Usage
There are many different permutations to using this new dataset, but a basic example shows how to add it to an existing training config yaml. To understand what each parameter does, please look at the class doc for the CodeSwitchedDataset class in nemo.collections.common.data.dataset.

```yaml
++model.train_ds.manifest_filepath=[/path/to/en/tarred_manifest.json,/path/to/de/tarred_manifest.json,/path/to/ja/tarred_manifest.json]
++model.train_ds.tarred_audio_filepaths=['/path/to/en/tars/audio__OP_0..511_CL_.tar','/path/to/de/tars/audio__OP_0..1023_CL_.tar', '/path/to/ja/tars/audio__OP_0..2047_CL_.tar']
++model.train_ds.is_code_switched=true
++model.train_ds.is_tarred=true
++model.train_ds.shuffle=true
++model.train_ds.code_switched.min_duration=12
++model.train_ds.code_switched.max_duration=20
++model.train_ds.code_switched.min_monolingual=0.3
++model.train_ds.code_switched.probs=[0.25, 0.5, 0.25]
++model.train_ds.code_switched.force_monochannel=true
++model.train_ds.code_switched.sampling_scales=[0.75, 1, 0.25]
++model.validation_ds.manifest_filepath=[[/path/to/en/non-tarred/manifest.json, /path/to/de/non-tarred/manifest.json, /path/to/ja/non-tarred/manifest.json]]
++model.validation_ds.is_code_switched=true
++model.validation_ds.is_tarred=false \
++model.validation_ds.code_switched.min_duration=6
++model.validation_ds.code_switched.max_duration=20
++model.validation_ds.code_switched.min_monolingual=0.3
++model.validation_ds.code_switched.probs=[0.33, 0.34, 0.33]
++model.validation_ds.code_switched.force_monochannel=true
++model.validation_ds.code_switched.sampling_scales=1
~model.tokenizer
++model.tokenizer.type=agg
++model.tokenizer.langs.en.dir=/path/to/en/tokeniser
++model.tokenizer.langs.en.type=bpe
++model.tokenizer.langs.de.dir=/path/to/de/tokeniser
++model.tokenizer.langs.de.type=bpe
++model.tokenizer.langs.ja.dir=/path/to/ja/tokeniser
++model.tokenizer.langs.ja.type=bpe
```

**Pre checks**:
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


trias702:
Can someone please review this PR for merging? We really need this functionality for Riva work.

nemo/collections/asr/data/audio_to_text.py
```diff
@@ -653,6 +653,7 @@ def __init__(
         use_start_end_token: bool = True,
         return_sample_id: bool = False,
         channel_selector: Optional[ChannelSelectorType] = None,
+        language: Optional[str] = None,
```
KunalDhawan, line 656: Could you please also add this parameter to the docstring of the class?

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,310 @@ def __len__(self):
...
+        self.infinity_mode = infinity_mode
+        self.global_rank = global_rank
+        self.world_size = world_size
+        self.augmentor = augmentor
+        self.sample_rate = sample_rate
+        self.length = 0
+        if lang_probs is None:
+            self.prob_dict = {l: 1.0 / len(self.langs) for l in self.langs}
+        else:
+            self.prob_dict = {l: lang_probs[i] for i, l in enumerate(self.langs)}
```
KunalDhawan, line 377: We should probably add a check here to ensure that len(lang_probs) == len(datasets.keys()).
Also, would it be better to use a dictionary with the format <lang_id_str>: prob_lang_float instead of simply a list for lang_probs, this would make it more user friendly and also help avoid errors where there might be a mismatch in the order of languages in lang_probs and datasets?

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,310 @@ def __len__(self):
...
+                dtype=created_sample_audios[0].dtype,
+            )
+        else:
+            comp_audio = np.zeros(
+                shape=(int(self.pause_start * self.sample_rate / 1000.0),), dtype=created_sample_audios[0].dtype
+            )
+
+        for idx, wav in enumerate(created_sample_audios):
+            if not multichannel:
+                wav = np.trim_zeros(wav)
```
KunalDhawan, line 491: The aim here is to remove leading and trailing silences from the sample right? In that case would it be better to achieve this using VAD and a low db threshold for the case where silence/noise is of small amplitude but non zero?

KunalDhawan:
Great work @trias702! Added a few minor comments but overall the implementation looks good. @bmwshop for final review.

nemo/collections/common/data/dataset.py
trias702: It already takes quite a bit of processing time to create the synthetic utterances on-the-fly, so I think VAD would only complicate that further. It's a good idea to revisit for a V2 approach though, perhaps if we can figure out how to do all the processing on GPU if one is detected.

nemo/collections/asr/data/audio_to_text.py
trias702: I have updated it

nemo/collections/common/data/dataset.py
trias702: I have changed lang_probs to be a dictionary, and also added an assert check for length. Can you please check the new code?

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+        and config['code_switch_probabilities'] is not None
+        and len(config['code_switch_probabilities']) == len(config['code_switch_languages'])
+    ):
+        cs_probs_dict = {
+            l: config['code_switch_probabilities'][i] for i, l in enumerate(config['code_switch_languages'])
+        }
+    else:
+        cs_probs_dict = None
+
+    dataset = CodeSwitchedDataset(
```
titu1994, line 460: is this a tarred dataset equivalent? Or loose file dataset?

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+        datasets,
+        shuffle=config.get('code_switch_shuffle', True),
```
titu1994, line 462: Too many base level arguments. How about
`model.train_ds.code_switch.*`

titu1994: Then you dont need a dozen code_switch_* prepending

nemo/collections/asr/data/audio_to_text_dataset.py
titu1994: If tarred it needs to be marked in its name as such

nemo/collections/asr/models/classification_models.py
```diff
@@ -174,7 +174,7 @@ def setup_training_data(self, train_data_config: Optional[Union[DictConfig, Dict
         # Need to set this because if using an IterableDataset, the length of the dataloader is the total number
         # of samples rather than the number of batches, and this messes up the tqdm progress bar.
         # So we set the number of steps manually (to the correct number) to fix this.
-        if 'is_tarred' in train_data_config and train_data_config['is_tarred']:
+        if isinstance(self._train_dl.dataset, torch.utils.data.IterableDataset):
```
titu1994, line 177: Train dl can be None. The earlier config check is safer. Or add more checks to your version.

nemo/collections/asr/models/ctc_bpe_models.py
```diff
@@ -106,7 +106,7 @@ def _setup_dataloader_from_config(self, config: Optional[Dict]):
             return dataset
 
         shuffle = config['shuffle']
-        if config.get('is_tarred', False):
+        if isinstance(dataset, torch.utils.data.IterableDataset):
```
titu1994, line 109: dataset can be None. do complete checks

nemo/collections/asr/models/ctc_models.py
```diff
@@ -365,7 +365,7 @@ def _setup_dataloader_from_config(self, config: Optional[Dict]):
             return dataset
 
         shuffle = config['shuffle']
-        if config.get('is_tarred', False):
+        if isinstance(dataset, torch.utils.data.IterableDataset):
```
titu1994, line 368: Same as above

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models.py
```diff
@@ -144,7 +144,7 @@ def _setup_dataloader_from_config(self, config: Optional[Dict]):
             return dataset
 
         shuffle = config['shuffle']
-        if config.get('is_tarred', False):
+        if isinstance(dataset, torch.utils.data.IterableDataset):
```
titu1994, line 147: ++

nemo/collections/asr/models/label_models.py
```diff
@@ -275,7 +275,7 @@ def setup_training_data(self, train_data_layer_config: Optional[Union[DictConfig
         # Need to set this because if using an IterableDataset, the length of the dataloader is the total number
         # of samples rather than the number of batches, and this messes up the tqdm progress bar.
         # So we set the number of steps manually (to the correct number) to fix this.
-        if 'is_tarred' in train_data_layer_config and train_data_layer_config['is_tarred']:
+        if isinstance(self._train_dl.dataset, torch.utils.data.IterableDataset):
```
titu1994, line 278: ++

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -523,7 +523,7 @@ def setup_training_data(self, train_data_config: Optional[Union[DictConfig, Dict
         # Need to set this because if using an IterableDataset, the length of the dataloader is the total number
         # of samples rather than the number of batches, and this messes up the tqdm progress bar.
         # So we set the number of steps manually (to the correct number) to fix this.
-        if 'is_tarred' in train_data_config and train_data_config['is_tarred']:
+        if isinstance(self._train_dl.dataset, torch.utils.data.IterableDataset):
```
titu1994, line 526: Do in all files above and below complete safe checks before doing your check

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,313 @@ def __len__(self):
     def __getitem__(self, idx):
         dataset_id, dataset_index = self.indices[idx]
         return self.datasets[dataset_id][dataset_index]
+
+
+class CodeSwitchedDataset(IterableDataset):
```
titu1994, line 294: Seems the dataset dosnt support tarred dataset. Most of our datasets are too vast to use loose file anymore.. Maybe refactor out the core components and wrap them to support at least tarred datasets (bucketing is also important but can be done later)

titu1994: Or is it a wrapper dataset like bucketing already? Im a bit confused so ill leave it to @bmwshop to review properly

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,313 @@ def __len__(self):
...
+                self.lang_kind[lang] = 'map'
+                self.length += int((len(dataset) // world_size) * self.sampling_scales[lang])
+
+        if seed is not None:
+            np.random.seed(seed)
+
+        # set this to ensure compatibility with models searching for the collate_fn
+        # since this class stores datasets as a dict, not list
+        # self.collate_fn = self.datasets[self.langs[0]].collate_fn
+        if hasattr(self.datasets[self.langs[0]], 'collate_fn'):
```
titu1994, line 409: These need a lot of checks - for example don't assume self.datasets is a list in the first place, or that it has 'datasets' again as its argument. Youll need to do multiple hasattr for each line here

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,313 @@ def __len__(self):
...
+            self.collate_fn = self.datasets[self.langs[0]].collate_fn
+        elif hasattr(self.datasets[self.langs[0]].datasets[0], 'collate_fn'):
+            # support datasets that are lists of entries
+            self.collate_fn = self.datasets[self.langs[0]].datasets[0].collate_fn
+        else:
+            # support datasets that are lists of lists
+            self.collate_fn = self.datasets[self.langs[0]].datasets[0].datasets[0].collate_fn
+
+    def get_iterable_by_lang(self, lang):
```
titu1994, line 418: Below methods must be documented

nemo/collections/common/data/dataset.py
titu1994: VAD is too expensive in a data loader + complicates data parallel with multiple threads (multiple CUDA instances cannot be run on threads, only sub-processes). 
User should do it outside of training.

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,313 @@ def __len__(self):
...
+                shape=(int(self.pause_start * self.sample_rate / 1000.0),), dtype=created_sample_audios[0].dtype
+            )
+
+        for idx, wav in enumerate(created_sample_audios):
+            if not multichannel:
+                wav = np.trim_zeros(wav)
+
+            wav_norm = wav * (10.0 ** (self.db_norm / 20.0) / np.maximum(0.01, (wav ** 2).mean(axis=0) ** 0.5))
+
+            if idx < len(created_sample_audios) - 1:
```
titu1994, line 498: There is a lot of logic in this function - every line must be documented for the future when modifications may be needed.

nemo/collections/common/data/dataset.py
```diff
@@ -286,3 +289,313 @@ def __len__(self):
...
+                comp_audio,
+                np.zeros(shape=(int(self.pause_end * self.sample_rate / 1000.0),), dtype=comp_audio.dtype),
+                axis=0,
+            )
+
+        if self.augmentor is not None:
+            # import here to avoid circular import error
+            from nemo.collections.asr.parts.preprocessing import AudioSegment
+
+            mb = io.BytesIO()
```
titu1994, line 540: This will be super slow right? Better to just not allow augmentation here then?

titu1994 (commented):
Overall, the pr needs some documentation (inline as well as in NeMo Docs) and checks.
Core logic of dataset, ill leave to others to review

nemo/collections/common/data/dataset.py
titu1994: Probably needs to be renamed a little if its a wrapped dataset

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+    ):
+        lang = config['code_switch_languages'][dataset_idx]
+        conf = copy.deepcopy(config)
+        conf['manifest_filepath'] = manifest_filepath
+        with open_dict(conf):
+            conf['tarred_audio_filepaths'] = tarred_audio_filepath
+            conf['language'] = lang
+        if tarred_audio_filepath is None or len(tarred_audio_filepath) == 0:
+            if tokenizer is None:
+                dataset = get_char_dataset(config=conf, augmentor=None)
```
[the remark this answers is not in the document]
trias702: Actually this is by design. We don't want to pass the augmentor to the individual dataset classes because we then sample from them and we want un-augmented audio returned to build the synthetic utterance with. The augmentor, if provided, is only used on the final, composite code-switched synthetic sample. This is why it's always None until you instantiate the final code_switched_dataset object and only then do you pass the augmentor to that object.

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+            else:
+                dataset = get_bpe_dataset(config=conf, tokenizer=tokenizer, augmentor=None)
+        else:
+            dataset = get_tarred_dataset(
+                config=conf,
+                tokenizer=tokenizer,
+                shuffle_n=shuffle_n,
+                global_rank=global_rank,
+                world_size=world_size,
+                augmentor=None,
```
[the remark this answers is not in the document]
trias702: Actually this is by design. We don't want to pass the augmentor to the individual dataset classes because we then sample from them and we want un-augmented audio returned to build the synthetic utterance with. The augmentor, if provided, is only used on the final, composite code-switched synthetic sample. This is why it's always None until you instantiate the final code_switched_dataset object and only then do you pass the augmentor to that object.

nemo/collections/asr/data/audio_to_text_dataset.py
trias702: Neither really. It's the same "style" as the ConcatDataset

nemo/collections/common/data/dataset.py
trias702: The constructor specifies that self.datasets is a dict, which is how it's used here. You want me to typecheck that it's actually a dict?

nemo/collections/common/data/dataset.py
trias702: So this is the part where we apply the augmentor to the final, synthetic utterance only, and why we always pass `Augmentor=None` to the individual language sub-datasets. We want all the individual samples from the individual mono-languages to be clean, and then we build the clean synthetic, and then we hit that final synthetic with the augmentation.

As for the logic, it's all in-memory operations with io.BytesIO, nothing is actually being written to disk. I have done several training runs with online CS and augmentation, and it doesn't slow it down by much at all, still very much usable.

There's sadly no way around it, we need to do augmentation on the final, full synthetic sample because many augmentations depend on features of the full waveform to work correctly (for example the new NormPerturbation).

nemo/collections/common/data/dataset.py
trias702: Correct, it's a wrapped dataset, just like ConcatDataset which I used as the base for this new class

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -546,8 +630,35 @@ def get_audio_to_text_char_dataset_from_config(
         )
         return dataset
 
+    # Instantiate a code-switched dataset if config is present
+    if config.get('is_code_switch', False):
```
bmwshop, line 634: maybe "is_code_switched" as the name here and some other params is a little more user friendly?

nemo/collections/common/data/dataset.py
bmwshop: Right, it is a wrapper class.

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+    # [[dataset1,dataset2]] otherwise ModelPT would interfere
+    if len(manifest_filepaths) == 1 and not isinstance(manifest_filepaths[0], str):
+        manifest_filepaths = config['manifest_filepath'][0]
+    if tarred_audio_filepaths is None:
+        tarred_audio_filepaths = [None] * len(manifest_filepaths)
+    datasets = {}
+    for dataset_idx, (tarred_audio_filepath, manifest_filepath) in enumerate(
+        zip(tarred_audio_filepaths, manifest_filepaths)
+    ):
+        lang = config['code_switch_languages'][dataset_idx]
```
bmwshop, line 425: What happens if the a manifest contains LID information and we pass LIDs here as well?

nemo/collections/asr/data/audio_to_text_dataset.py
bmwshop: We now have multiple wrapper classes; i think it's time to start thinking about a proper graph of such. Not sure that it's for this PR to cover, but we ought to consider, i think.

nemo/collections/asr/models/ctc_bpe_models.py
trias702: I don't think dataset can be None. If you look 5 lines above this one, there is:

nemo/collections/asr/models/ctc_models.py
trias702: Dataset cannot be None. There is the following code just above the line you have highlighted:

```python
if dataset is None:
    return None
```

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models.py
trias702: Dataset cannot be None. There is the following code just above the line you have highlighted:

```python
if dataset is None:
    return None
```

nemo/collections/asr/data/audio_to_text_dataset.py
trias702: It's not a problem, because LIDs in the manifest always take precedence, you can see that here: https://github.com/NVIDIA/NeMo/blob/40dd53d25b0d3a40775d32c6438580cd526af2d0/nemo/collections/common/parts/preprocessing/collections.py#L160

The new language parameter is only used as a fallback IFF using AggTokeniser and the manifest has no LID in the manifest.

nemo/collections/asr/data/audio_to_text_dataset.py
trias702: Versus code_switch_*?

nemo/collections/asr/data/audio_to_text_dataset.py
```diff
@@ -398,6 +400,88 @@ def get_tarred_dataset(
...
+    # needed to support validation Datasets that arrive here as
+    # [[dataset1,dataset2]] otherwise ModelPT would interfere
+    if len(manifest_filepaths) == 1 and not isinstance(manifest_filepaths[0], str):
+        manifest_filepaths = config['manifest_filepath'][0]
+    if tarred_audio_filepaths is None:
+        tarred_audio_filepaths = [None] * len(manifest_filepaths)
+    datasets = {}
+    for dataset_idx, (tarred_audio_filepath, manifest_filepath) in enumerate(
+        zip(tarred_audio_filepaths, manifest_filepaths)
+    ):
```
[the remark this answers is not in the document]
trias702: Fixed

nemo/collections/asr/models/classification_models.py
trias702: Fixed

nemo/collections/asr/models/label_models.py
trias702: Fixed

nemo/collections/common/data/dataset.py
trias702: Fixed, please check newly committed code

nemo/collections/common/data/dataset.py
trias702: Added comments on every line

nemo/collections/asr/data/audio_to_text_dataset.py
trias702: I have changed it to `is_code_switched` as requested

nemo/collections/common/data/dataset.py
trias702: I have documented them now

nemo/collections/asr/data/audio_to_text_dataset.py
trias702: I have added the sub dict as requested

KunalDhawan (approved):
Thanks @trias702, LGTM!

KunalDhawan (approved):

PR #6598, opened 2023-05-08
Update manifest.py to use os.path for get_full_path

stevehuang52:


nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -196,9 +191,7 @@ def get_full_path(
         ]
     elif isinstance(audio_file, str):
         # If input is a string, get the corresponding full path
-        audio_file = Path(audio_file)
-
-        if (len(str(audio_file)) < audio_file_len_limit) and not audio_file.is_absolute() and not audio_file.is_file():
+        if not os.path.isabs(audio_file) and not os.path.exists(audio_file):
```
titu1994, line 194: Let's keep the file length check please

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -210,23 +203,17 @@ def get_full_path(
...
 
             if is_datastore_path(audio_file_path):
                 # If audio was originally on an object store, use locally-cached path
                 audio_file_path = datastore_path_to_local_path(audio_file_path)
 
-            audio_file_path = Path(audio_file_path)
-
-            if audio_file_path.is_file():
-                audio_file = str(audio_file_path.absolute())
+            if os.path.exists(audio_file_path):
```
titu1994, line 215: Exists can be true even for dir - we need file only check here

titu1994 (approved):
Thanks !

VahidooX (approved):

PR #6608, opened 2023-05-09
[ASR] Fix for old models in change_attention_model

sam1373:
# What does this PR do ?

Fixes some issues:
-change_attention_model should work for old models which don't have some of the relevant parameters in config
-also it shouldn't run when nothing to change was submitted

**Collection**: ASR

  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/parts/mixins/mixins.py
+++ b/nemo/collections/asr/parts/mixins/mixins.py
@@ -412,6 +412,9 @@ class ASRModuleMixin(ASRAdapterModelMixin):
             update_config (bool): Whether to update the config or not with the new attention model.
                 Defaults to True.
         """
+        if self_attention_model is None and att_context_size is None:
+            return
+
         if not hasattr(self, 'encoder'):
             logging.info(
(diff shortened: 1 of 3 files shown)


nemo/collections/asr/parts/mixins/mixins.py
```diff
@@ -412,6 +412,9 @@ def change_attention_model(
             update_config (bool): Whether to update the config or not with the new attention model.
                 Defaults to True.
         """
+        if self_attention_model is None and att_context_size is None:
+            return
```
titu1994, line 416: Don't we want to reset the attention to full context ?

titu1994 (approved):
Lgtm thanks !

nemo/collections/asr/parts/mixins/mixins.py
sam1373: no, in this method setting either parameter to None means not changing it (this is default so you can just change one of them), so if both are None we can just return

PR #6644, opened 2023-05-12
Fix for interctc test random failure

Kipok:
# What does this PR do ?

I finally figured out the reason for random failure of this test. It turns out that sometimes our input signal is such that for some models it results in all-zeros in the output of some initial encoder layers. This is then propagated all the way through the model and so the assert that final layer output is different from middle layer fails. After changing the code to pass processed signal directly, this should not happen anymore.

**Collection**: ASR

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/tests/collections/asr/test_asr_interctc_models.py
+++ b/tests/collections/asr/test_asr_interctc_models.py
@@ -66,7 +66,6 @@ def squeezeformer_encoder_config() -> Dict:
 
 
 class TestInterCTCLoss:
-    @pytest.mark.pleasefixme
     @pytest.mark.unit
     @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
Seems reasonable but you sure the zeros generation won't affect model padding during actual training ?

Kipok:
> Seems reasonable but you sure the zeros generation won't affect model padding during actual training ?

There is no real issue here, it's just that I test that the output of different layers is different, but when they are all zeros, this breaks. The logic in the code needs no changes.

PR #6656, opened 2023-05-15
eval_beamsearch_ngram.py with hybrid ctc

karpnv:
# What does this PR do ?

- Fix eval_beamsearch_ngram.py with hybrid ctc

- Set default value text_processing.separate_punctuation=false for 
scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
examples/asr/speech_to_text_eval.py

**Collection**: ASR

# Usage

```python
python3 /path/to/NeMo/scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py \
nemo_model_file=/path/to/hybrid/model.nemo \
input_manifest=/path/to/manifest \
kenlm_model_file=/path/to/kenlm \
decoding_mode=beamsearch_ngram \
beam_width=[32] \
beam_alpha=[0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.1] \
beam_beta=[0.0,0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9] \

```

**PR Type**:
- [V ] Bugfix


nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
```diff
@@ -229,7 +235,10 @@ def transcribe(
                 self.joint.unfreeze()
                 if hasattr(self, 'ctc_decoder'):
                     self.ctc_decoder.unfreeze()
-        return hypotheses, all_hypotheses
+        if logprobs:
+            return logits_list
```
titu1994, line 239: Can you do logprobs, None so signature is always 2 elements out

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -151,7 +152,11 @@ def beam_search_eval(
     model.cfg.decoding.beam = cfg.decoding
 
     # Update model's decoding strategy
-    model.change_decoding_strategy(model.cfg.decoding)
+    if isinstance(model, EncDecHybridRNNTCTCModel):
+        model.change_decoding_strategy(model.cfg.decoding, decoder_type='ctc')
+        model.decoding = model.ctc_decoding
```
titu1994, line 157: Don't in place replace like this, use the change decoding strategy above

titu1994 (commented):
Looks ok to me @VahidooX for final reviews

examples/asr/speech_to_text_eval.py
```diff
@@ -82,7 +82,7 @@ class EvaluationConfig(transcribe_speech.TranscriptionConfig):
     only_score_manifest: bool = False
 
     text_processing: Optional[TextProcessingConfig] = TextProcessingConfig(
-        punctuation_marks=".,?", separate_punctuation=True, do_lowercase=False, rm_punctuation=False,
+        punctuation_marks=".,?", separate_punctuation=False, do_lowercase=False, rm_punctuation=False,
```
VahidooX, line 85: Do our current PC models on NGC follow the separate_punctuation=False?
The default should work with our current PC models.

nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
karpnv: yes I could, but I think it is better to return one, because it replicates EncDecCTCModel.transcribe(files, logprobs=True) function which present in many tutorials, for instance   https://github.com/NVIDIA/NeMo/blob/88816f0b87364ee539731a0f59f298e958ab68f9/tutorials/asr/Offline_ASR.ipynb#L248

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -151,7 +152,11 @@ def beam_search_eval(
     model.cfg.decoding.beam = cfg.decoding
 
     # Update model's decoding strategy
-    model.change_decoding_strategy(model.cfg.decoding)
+    if isinstance(model, EncDecHybridRNNTCTCModel):
```
VahidooX, line 155: Why we assume the Hybrid models are supposed to be used with ctc decoder? Doesn't this script work with RNNT models?

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -151,7 +152,11 @@ def beam_search_eval(
...
+        model.change_decoding_strategy(model.cfg.decoding, decoder_type='ctc')
+        model.decoding = model.ctc_decoding
+    else:
+        model.change_decoding_strategy(model.cfg.decoding)
```
VahidooX, line 159: You may save the decoding or ctc_decoding based on the type of the model in a variable named decoding and just use that in the rest of the model instead of using model.decoding.

nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
karpnv: Oh I see, EncDecCTCModel.transcribe() always return one

VahidooX (changes requested):

examples/asr/speech_to_text_eval.py
karpnv: Yes, Som made a temporary regexp fix to follow this. So, assume that bought target and predicted punctuation is not separated from words.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
karpnv: moved it into inside change_decoding_strategy()

nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
karpnv: done

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
karpnv: moved it into change_decoding_strategy()

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -316,7 +325,15 @@ def default_autocast():
 
         with autocast():
             with torch.no_grad():
-                all_logits = asr_model.transcribe(audio_file_paths, batch_size=cfg.acoustic_batch_size, logprobs=True)
+                if isinstance(asr_model, EncDecHybridRNNTCTCModel):
```
VahidooX, line 328: The EncDecHybridRNNTCTCModel and regular ctc models should return the same number of outputs to be consistent.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
```diff
@@ -331,11 +348,17 @@ def default_autocast():
     for batch_idx, probs in enumerate(all_probs):
         preds = np.argmax(probs, axis=1)
         preds_tensor = torch.tensor(preds, device='cpu').unsqueeze(0)
-        pred_text = asr_model._wer.decoding.ctc_decoder_predictions_tensor(preds_tensor)[0][0]
+        if isinstance(asr_model, EncDecHybridRNNTCTCModel):
```
VahidooX, line 351: If you save the right decoding into a variable in the top, then there is no need to check again here.

nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
```diff
@@ -339,6 +348,7 @@ def change_decoding_strategy(self, decoding_cfg: DictConfig = None, decoder_type
         decoding_cfg = OmegaConf.merge(decoding_cls, decoding_cfg)
 
         self.ctc_decoding = CTCDecoding(decoding_cfg=decoding_cfg, vocabulary=self.ctc_decoder.vocabulary)
+        self.decoding = self.ctc_decoding
```
VahidooX, line 351: I don't think you need to do it here. It would overwrite the RNNT decoding and we cannot switch back. There are other places in the code which are not based on this assumption.

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models.py
```diff
@@ -436,6 +436,7 @@ def change_decoding_strategy(self, decoding_cfg: DictConfig = None, decoder_type
             decoding_cfg = OmegaConf.merge(decoding_cls, decoding_cfg)
 
             self.ctc_decoding = CTCBPEDecoding(decoding_cfg=decoding_cfg, tokenizer=self.tokenizer)
+            self.decoding = self.ctc_decoding
```
VahidooX, line 439: self.decoding is for RNNT.

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
VahidooX: It would overwrite the RNNT decoding. How about doing something like the following?
```
if isinstance(model, EncDecHybridRNNTCTCModel):
        model.change_decoding_strategy(model.cfg.decoding, decoder_type='ctc')
	decoding = model.ctc_decoding
    else:
        model.change_decoding_strategy(model.cfg.decoding)
	decoding = model.decoding

```

VahidooX (changes requested):

scripts/asr_language_modeling/ngram_lm/eval_beamsearch_ngram.py
karpnv: done

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models.py
karpnv: removed

nemo/collections/asr/models/hybrid_rnnt_ctc_models.py
karpnv: removed

VahidooX (approved):
LGTM!

PR #6666, opened 2023-05-16
GPT extrapolatable position embedding (xpos/sandwich/alibi/kerple) and Flash Attention

hsiehjackson:
# What does this PR do ?
* Add extrapolatable position embedding, e.g. xpos, sandwich.
* Fix extrapolatable position embedding, e.g. alibi, kerple. 
* Add flash attention (FA)

**Collection**: [NLP]

# Changelog 
* GPT
  - `examples/nlp/language_modeling/conf/megatron_gpt_config.yaml`
  - `nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py`
  - `nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py`
  - `nemo/collections/nlp/modules/common/megatron/language_model.py`
* T5
  - `nemo/collections/nlp/modules/common/megatron/token_level_encoder_decoder.py`
  - `nemo/collections/nlp/modules/common/megatron/megatron_encoders.py`
  - `nemo/collections/nlp/modules/common/megatron/megatron_decoders.py`
  - `nemo/collections/nlp/modules/common/megatron/megatron_transformer_encoder.py`
  - `nemo/collections/nlp/modules/common/megatron/megatron_transformer_decoder.py`
* PE
  - `nemo/collections/nlp/modules/common/megatron/transformer.py`
  - `nemo/collections/nlp/modules/common/megatron/attention.py`
  - `nemo/collections/nlp/modules/common/megatron/kerple_relative_position.py`
  - `nemo/collections/nlp/modules/common/megatron/sandwich_relative_position.py`
  - `nemo/collections/nlp/modules/common/megatron/xpos_relative_position.py`
* FA
  - `nemo/collections/nlp/modules/common/megatron/attention.py`

# Usage
* PE

```python
# GPT
model.position_embedding_type='' # Options ['learned_absolute', 'rope', 'alibi', 'xpos', 'sandwich', 'kerple']
# T5
model.encoder.position_embedding_type='' # Options ['learned_absolute', 'rope', 'alibi', 'xpos', 'sandwich', 'kerple']
model.decoder.position_embedding_type='' # Options ['learned_absolute', 'rope', 'alibi', 'xpos', 'sandwich', 'kerple']
```
* FA

```python
# GPT
model.use_flash_attention=true
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Todo
- [x] Add tests for PE and FA
- [x] Integrate multiple PE classes into one file 
- [x] Move if/else of different attention bias outside CoreAttention
- [ ] Update requirements files
- [x] Update dockerfile
- [ ] Add a warning for FA


examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -127,6 +127,12 @@ def main(cfg) -> None:
         peft_model_cfg.data.test_ds = cfg.model.data.test_ds
         peft_model_cfg.activations_checkpoint_granularity = None
         peft_model_cfg.activations_checkpoint_method = None
+        peft_model_cfg.encoder_seq_length = cfg.model.data.test_ds.max_seq_length
+        if 'use_flash_attention' not in peft_model_cfg:
+            peft_model_cfg.use_flash_attention = False
+
+        if 'use_flash_attention' in cfg.model:
+            peft_model_cfg.use_flash_attention = cfg.model.use_flash_attention
```
hsiehjackson, line 135: @ekmb Do we need to include evaluation in this PR?

ekmb: no, you can revert

trias702:
> @trias702 found a bug in NeMo GPT Alibi, could you check if this PR fixes it if not notify @hsiehjackson of your fix?

The initial bugs I found have been fixed, however, there is currently another problem with the Alibi class in that it sets the relative positions tensor to cuda manually with a .cuda() call. This breaks functionality of all models which instantiate Alibi with a PL Trainer set to CPU. We should not be hardcoding the device for tensors inside Megatron classes, unless this has been agreed upon as a design pattern? There should be a device argument passed in, which is read from the Trainer object at runtime. MegatronGPTModel does explicitly support CPU-only instantiation, see here: https://github.com/NVIDIA/NeMo/blob/5428a97e3c9578b79fa7b30b6c53f2ae9759f418/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L215

hsiehjackson:
> 



> > @trias702 found a bug in NeMo GPT Alibi, could you check if this PR fixes it if not notify @hsiehjackson of your fix?
> 
> The initial bugs I found have been fixed, however, there is currently another problem with the Alibi class in that it sets the relative positions tensor to cuda manually with a .cuda() call. This breaks functionality of all models which instantiate Alibi with a PL Trainer set to CPU. We should not be hardcoding the device for tensors inside Megatron classes, unless this has been agreed upon as a design pattern? There should be a device argument passed in, which is read from the Trainer object at runtime. MegatronGPTModel does explicitly support CPU-only instantiation, see here:
> 
> https://github.com/NVIDIA/NeMo/blob/5428a97e3c9578b79fa7b30b6c53f2ae9759f418/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L215

Change `.cuda()` with `torch.cuda.is_available()`.

Dockerfile
```diff
@@ -72,6 +72,10 @@ WORKDIR /tmp/nemo
 COPY requirements .
 RUN for f in $(ls requirements*.txt); do pip3 install --disable-pip-version-check --no-cache-dir -r $f; done
 
+# install flash attention dependencies
+RUN pip install flash-attn
+RUN pip install triton==2.0.0.dev20221202
```
ericharper, line 77: Does this need to be pinned?

ekmb:
> Amazing! Did you happen to launch a small experiment to check the training (or inference) of both models leads to (almost) identical results?

@michalivne training loss curves match with and w/o FA

khcs:
Nice PR! 👏

1. Good that we have `experimental` decorator from `from nemo.utils.decorators import experimental` for `sandwich` and `xpos` - do we want to add comment about that in the .yaml file as well?
2. Flash-Attention does not work with `kerple` at this point - do we want to add a check, warning, or comment about that?

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -77,7 +77,7 @@ model:
   transformer_block_type: 'pre_ln' # Options ['pre_ln', 'post_ln', 'normformer']
   openai_gelu: False # Use OpenAI's GELU instead of the default GeLU
   normalize_attention_scores: True # Whether to scale the output Q * K^T by 1 / sqrt(hidden_size_per_head). This arg is provided as a configuration option mostly for compatibility with models that have been weight-converted from HF. You almost always want to se this to True.
-  position_embedding_type: 'learned_absolute' # Position embedding type. Options ['learned_absolute', 'rope']
+  position_embedding_type: 'learned_absolute' # Position embedding type. Options ['learned_absolute', 'rope', 'alibi', 'xpos', 'sandwich', 'kerple']
```
khcs, line 80: Do we want to add a comment mentioning at this point `xpos` and `sandwich` being experimental, and `kerple` does not work with flash-attention?

Dockerfile
hsiehjackson: FA has fixed triton version currently. I have tried other version which will raise errors.
https://github.com/HazyResearch/flash-attention/blob/main/flash_attn/flash_attn_triton.py#L3

nemo/collections/nlp/modules/common/megatron/language_model.py
```diff
@@ -551,6 +558,39 @@ def __init__(
...
+            self.encoder_relative_position_embedding = ALiBiRelativePositionEmbedding(
+                bidirectional=encoder_attn_mask_type != AttnMaskType.causal,
+                num_attention_heads=num_attention_heads,
+                layer_type=LayerType.encoder,
+                num_attention_heads_alibi=None,
+                max_seq_len=max_position_embeddings,
+            )
+
+        elif position_embedding_type == 'kerple':
+            # TODO: If this is used for encoder-decodemax_position_embeddingsr model, implement proper logic and following
```
khcs, line 574: Do we want to add a test to make sure we're not using flash-attention with `kerple`, as it's not supported at the moment, like:
```
assert use_flash_attention == False # flash-attention not supported with kerple at this point
```

khcs (commented):
Nice PR! 👏

1. Good that we have `experimental` decorator from `from nemo.utils.decorators import experimental` for `sandwich` and `xpos` - do we want to add comment about that in the .yaml file as well?
2. Flash-Attention does not work with `kerple` at this point - do we want to add a check, warning, or comment about that?

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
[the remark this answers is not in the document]
hsiehjackson: Added

nemo/collections/nlp/modules/common/megatron/language_model.py
hsiehjackson: Added

Jenkinsfile
```diff
@@ -3144,6 +3151,88 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
       }
     }
     stage('L2: Megatron GPT Pretraining and Resume Training TP=2') {
+      when {
```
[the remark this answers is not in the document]
hsiehjackson: Originally, we test using rope, and we don't test the default GPT Pretraining. So, I add the default test back.

Jenkinsfile
```diff
@@ -3229,6 +3318,174 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
...
+    stage('L2: Megatron GPT with ALiBi Pretraining and Resume Training TP=2') {
+      when {
+        anyOf {
+          branch 'main'
+          changeRequest target: 'main'
+        }
+      }
+      failFast true
+      steps {
+        sh "python examples/nlp/language_modeling/megatron_gpt_pretraining.py \
```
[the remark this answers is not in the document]
hsiehjackson: I can enable rope (wo/ bias) and alibi (w/ bias)

nemo/collections/nlp/modules/common/megatron/attention.py
```diff
@@ -25,12 +25,17 @@
...
+from nemo.collections.nlp.modules.common.megatron.position_embedding.rotary_position_embedding import (
+    apply_rotary_pos_emb,
+)
 from nemo.collections.nlp.modules.common.megatron.utils import ApexGuardDefaults, attention_mask_func
 from nemo.collections.nlp.parts import utils_funcs
 from nemo.core import adapter_mixins
+from nemo.utils import logging
 
 try:
+    from apex._autocast_utils import _cast_if_autocast_enabled
```
ericharper, line 38: @trias702 just added an implementation of this, should we use that one instead since it is more robust?

nemo/collections/nlp/modules/common/megatron/attention.py
```diff
@@ -53,6 +58,19 @@
...
+    from flash_attn.flash_attn_interface import flash_attn_unpadded_func
+    from flash_attn.flash_attn_triton import flash_attn_func
+
+except (ImportError, ModuleNotFoundError):
+    logging.warning(
+        "flash_attn was not found. Please see the installation instructions: https://github.com/HazyResearch/flash-attention."
+        "If you use flash_attn with triton. Please install triton==2.0.0.dev20221202."
+    )
+    flash_attn_unpadded_func, flash_attn_func = None, None
+    unpad_input, pad_input = None, None
```
ericharper, line 72: You might want to add a `HAVE_FLASH_ATTENTION` variable here, then raise an error if a class that needs it is instantiated and you don't have it

ericharper: We can also do a config check very on in megatron_base_model.py if user configure `use_flash_attention=True` but `HAVE_FLASH_ATTENTION=False` then raise an error

nemo/collections/nlp/modules/common/megatron/attention.py
```diff
@@ -911,3 +930,84 @@ def forward(
...
+            assert len(attention_mask.shape) == 2
+            attention_mask_q = attention_mask
+            attention_mask_kv = attention_mask
+
+        q, indices_q, cu_seqlens_q, max_seqlen_q = unpad_input(query_layer, attention_mask_q)
+        k, _, cu_seqlens_k, max_seqlen_k = unpad_input(key_layer, attention_mask_kv)
+        v, _, _, _ = unpad_input(value_layer, attention_mask_kv)
+        causal = self.reset_fa_causal(
+            query_layer.shape[1], key_layer.shape[1], self.attn_mask_type == AttnMaskType.causal
+        )
```
ericharper, line 964: I really would prefer a wrapper class with all of this logic, then if `use_flash_attention` we use the wrapper class. 

Can we do a refactor for this?

nemo/collections/nlp/modules/common/megatron/attention.py
```diff
@@ -911,3 +930,84 @@ def forward(
         context_layer = context_layer.view(*new_context_layer_shape)
 
         return context_layer
+
+    def reset_fa_causal(self, query_length, key_length, causal):
```
MaximumEntropy, line 934: Unless I'm misunderstanding this function is doing more of an "is_causal" check rather than a reset of the property right? Can we rename it to reflect this if so?

nemo/collections/nlp/modules/common/megatron/megatron_decoders.py
```diff
@@ -88,6 +88,8 @@ def get_decoder_model(
     moe_dropout=0.0,
     turn_off_rop=False,  # turn off the RoP positional embedding
     version=1,
+    position_embedding_type='learned_absolute',
+    use_flash_attention=False,
```
ericharper, line 92: if you're going to add this, shouldn't you add it to the t5 configs as well?

nemo/collections/nlp/modules/common/megatron/position_embedding/alibi_relative_position_embedding.py
```diff
@@ -42,20 +42,31 @@ def build_slopes(num_attention_heads, num_attention_heads_alibi):
...
+    relative_position = torch.arange(1 - max_seq_len, 1)[None, :].mul(-1)  # (1, max_seq_len)
+
+    if full:
+        memory_position = torch.arange(1 - max_seq_len, 1)[:, None].mul(-1)
+        relative_position = torch.abs(memory_position - relative_position)  # (max_seq_len, max_seq_len)
 
-    # shape (num_attention_heads, max_seq_len, max_seq_len)
-    relative_position = torch.abs(relative_position).unsqueeze(0).expand(num_attention_heads, -1, -1)
+    if torch.cuda.is_available():
+        relative_position = relative_position.to(torch.cuda.current_device())
```
ericharper, line 69: This seems strange why are you checking for cuda here and moving the tensor manually?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -205,9 +205,10 @@ def _build_tokenizer(self):
         self.tokenizer = get_nmt_tokenizer(
             library=self._cfg.tokenizer.library,
             model_name=self._cfg.tokenizer.type,
-            tokenizer_model=self.register_artifact("tokenizer.model", self._cfg.tokenizer.model),
-            vocab_file=self.register_artifact("tokenizer.vocab_file", self._cfg.tokenizer.vocab_file),
-            merges_file=self.register_artifact("tokenizer.merge_file", self._cfg.tokenizer.merge_file),
+            tokenizer_model=self.register_artifact("tokenizer.model", self._cfg.tokenizer.get('model', None)),
```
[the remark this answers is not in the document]
ekmb: @trias702 could you please comment on that

ericharper (changes requested):
Overall really great PR. Have some minor comments.
Major comments are: 

1. Should update the model configs so this can be easily enabled/disabled.
2. Having something like `FlashCoreAttention` would be better than just `CoreAttention(use_flash_attention)` Is it possible to do? Would it be better?
3. This PR is very large and would like to see multiple reviews from @khcs , @MaximumEntropy and @yidong72  if possible

README.rst
```diff
@@ -280,6 +280,15 @@ It is highly recommended to use the NVIDIA PyTorch or NeMo container if having i
...
 
+
+Flash Attention
+~~~~~~~~~~~~~~~~~~~~
+Install flash-attn <https://github.com/HazyResearch/flash-attention> if you want to use Flash Attention to reduce memory overhead. 
+.. code-block:: bash
+
+  pip install flash-attn
+  pip install triton==2.0.0.dev20221202
+
```
ericharper, line 291: Can you also clarify that if the TransformerEngine path is used, it will use it's own version of Flash Attention?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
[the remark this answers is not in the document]
trias702: @MaximumEntropy It happens if you use a HF tokenizer. To use a pretrained HF tokenizer like `EleutherAI/gpt-neox-20b` or `bert-base-uncased` you don't need to pass any `model`, `vocab_file`, or `merge_file` in the config, you just need to pass `tokenizer.library=huggingface` and `tokenizer.type=EleutherAI/gpt-neox-20b`. But because the old code would call register_artifact for `cfg.tokenizer.model` which would be None, it would throw an error. That's why I changed this block, to ensure it works correctly with pure HF tokenizers at model instantiation.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -294,7 +294,7 @@ def get_inference_config(self):
     def model_provider_func(self, pre_process, post_process):
         """Model depends on pipeline paralellism."""
         model = GPTModel(
-            vocab_size=self.padded_vocab_size,
+            vocab_size=self.cfg.get('override_vocab_size', self.padded_vocab_size),
```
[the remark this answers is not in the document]
trias702: This is needed for converting models from one platform (Mosaic, HF, etc) into Megatron, because you need to instantiate the model with your Embedding layer for the exact dimensionality you require in order to import the weights, otherwise you get a mismatch or you need to do complex weight surgery to cut out the weights you need and pad with zeros. It was easier to add a special override flag to just force the model to use the vocab_size necessary to make downstream conversion work. If the `override_vocab_size` is None, which it will be 99.8% of the time, it uses padded_vocab_size same as before, so it felt like a really harmless bypass to make model conversion much easier for all future conversion work.

nemo/collections/nlp/modules/common/megatron/layer_norm_1p.py
```diff
@@ -47,3 +48,36 @@ def forward(self, x):
...
+def _cast_if_autocast_enabled(tensor):
+    if torch.is_autocast_enabled():
+        if tensor.device.type == 'cuda':
+            dtype = torch.get_autocast_gpu_dtype()
+        elif tensor.device.type == 'cpu':
+            dtype = torch.get_autocast_cpu_dtype()
+        else:
+            raise NotImplementedError()
+        return tensor.to(dtype=dtype)
+    return tensor
```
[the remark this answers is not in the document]
trias702: I agree, as this method is more robust to the one offered in Apex. The Apex one will only work for GPU autocasting but not CPU.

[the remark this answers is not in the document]
hsiehjackson: Let me move in `nemo.collections.nlp.modules.common.megatron.utils`

nemo/collections/nlp/modules/common/megatron/position_embedding/alibi_relative_position_embedding.py
hsiehjackson: The original code use `.cuda()` . However @trias702 suggest we need support CPU-only one.

nemo/collections/nlp/modules/common/megatron/attention.py
```diff
@@ -767,138 +796,128 @@ def forward(
...
+        # Rearrange query_layer, key_layer, value_layer
+        # ==================================================
+        if self.use_flash_attention:
+            query_layer = rearrange(query_layer, 'sq b np hn -> b sq np hn')
+            key_layer = rearrange(key_layer, 'sk b np hn -> b sk np hn')
+            value_layer = rearrange(value_layer, 'sv b np hn -> b sv np hn')
+        elif self.multi_query_attention:
+            query_layer = rearrange(query_layer, 'sq b np hn -> b (np sq) hn')
+            key_layer = rearrange(a, 'sk b 1 hn -> b hn sk')
+            value_layer = rearrange(value_layer, 'sv b np hn -> (b np) sv hn')
```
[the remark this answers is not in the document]
hsiehjackson: There is a typo here. Let me fix this.

README.rst
hsiehjackson: Currently, TransformerEngine also use the same open source Flash Attention. They don't have their own version.  The difference is whether we implement FA in NeMo or TransformerEngine.

nemo/collections/nlp/modules/common/megatron/attention.py
hsiehjackson: I think `CoreAttention` is already the lowest module to compute attention between q, k, v. Flash attention is used to replace the operation of `softmax`, `add_attention_bias`, and `multiply value`. Therefore, I refactor with different attention function `torch_attention()` and `flash_attention()`.

nemo/collections/nlp/modules/common/megatron/attention.py
hsiehjackson: Added

nemo/collections/nlp/modules/common/megatron/attention.py
hsiehjackson: Removed and use `_cast_if_autocast_enabled` from `nemo.collections.nlp.modules.common.megatron.utils `

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -77,7 +77,7 @@ model:
   transformer_block_type: 'pre_ln' # Options ['pre_ln', 'post_ln', 'normformer']
   openai_gelu: False # Use OpenAI's GELU instead of the default GeLU
   normalize_attention_scores: True # Whether to scale the output Q * K^T by 1 / sqrt(hidden_size_per_head). This arg is provided as a configuration option mostly for compatibility with models that have been weight-converted from HF. You almost always want to se this to True.
-  position_embedding_type: 'learned_absolute' # Position embedding type. Options ['learned_absolute', 'rope']
+  position_embedding_type: 'learned_absolute' # Position embedding type. Options ['learned_absolute', 'rope', 'alibi', 'kerple' , 'xpos', 'sandwich'] xpos and sandwich are experimental.
```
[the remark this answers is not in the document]
hsiehjackson: Added

nemo/collections/nlp/modules/common/megatron/attention.py
hsiehjackson: changed to `reset_is_causal()`

nemo/collections/nlp/modules/common/megatron/megatron_decoders.py
hsiehjackson: Added

README.rst
ericharper: Right but if user has `TransformerEngine=True` and `use_flash_attention=False` they will still be using flash attention, for example. Better to help clarify this

hsiehjackson: Added

nemo/collections/nlp/modules/common/megatron/position_embedding/alibi_relative_position_embedding.py
trias702: Yes, I saw that we have explicit support for CPUAccelerator inside MegatronGPTModel here: https://github.com/NVIDIA/NeMo/blob/ebfcef7902c9232da04ecf90f505c3c996e7e059/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L215

That would cause a device incompatibility to be thrown if you're doing a manual .cuda() call somewhere in the downstream Megatron classes, because your model at instantiation expects everything to be on CPU.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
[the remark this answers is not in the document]
MaximumEntropy: So, for MPT-7B, the override_vocab_size is not equal to the padded vocab size computed based on the tokenizer vocab size plus divisibility by 128 upscaling?

I have concerns about this for two reasons 1) This is a "dangling" property with no reference in the yaml file so the only way for people to use this is to override from CLI via something like `+model.override_vocab_size=32000` 2) There are divisibility constraints on the vocab size based on the tensor parallel size and mod 128 divisibility for tensor cores, this sidesteps both of them.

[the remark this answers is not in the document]
ekmb: @MaximumEntropy for the final call, but @trias702, if it's for external only - how about renaming it to make it clear or adding a comment?

nemo/collections/nlp/modules/common/megatron/position_embedding/kerple_relative_position_embedding.py
```diff
@@ -65,21 +65,26 @@ def __init__(
 
         # initialize the slopes
         self.kerple_b = torch.nn.Parameter(build_slopes(num_attention_heads, num_attention_heads_kerple))
-        self.kerple_a = torch.zeros_like(self.kerple_b)
-        self.kerple_p = torch.ones_like(self.kerple_b)
+        self.kerple_a = torch.nn.Parameter(torch.rand_like(self.kerple_b))
```
michalivne, line 68: This initialization is unstable. Can you please change it to:
```
        self.kerple_a = torch.nn.Parameter(torch.ones_like(self.kerple_b))
        self.kerple_p = torch.nn.Parameter(torch.ones_like(self.kerple_b))
```

michalivne (changes requested):
Added an important request. Could you please address before merging?

nemo/collections/nlp/modules/common/megatron/position_embedding/kerple_relative_position_embedding.py
hsiehjackson: Added

examples/nlp/language_modeling/conf/megatron_model_base_config.yaml
```diff
@@ -36,4 +36,5 @@ megatron_legacy: False # Whether to use the legacy Megatron model. This affects
 normalize_attention_scores: True # Whether to scale the output Q * K^T by 1 / sqrt(hidden_size_per_head). This arg is provided as a configuration option mostly for compatibility with models that have been weight-converted from HF. You almost always want to se this to True.
 num_moe_experts: 1 # When >1, FFNs are changed to MoE layers
 moe_frequency: 1 # every Nth ffn layer will be made MoE 
-moe_dropout: 0.0 # Dropout value for MoE layers
\ No newline at end of file
+moe_dropout: 0.0 # Dropout value for MoE layers
+use_flash_attention: false # Use flash attention in self-attention module
```
ericharper, line 40:
```suggestion
use_flash_attention: false # Use flash attention in self-attention module, this config does nothing when transformer_engine=True
```

hsiehjackson: I add this comment in `examples/nlp/language_modeling/conf/megatron_gpt_config.yaml` since `transformer_engine` only used for GPT models

ekmb (approved):

ericharper (approved):
LGTM. Thanks!

hsiehjackson:
@michalivne Can you help me to take a look at the changes you request? Thanks!

michalivne (approved):
LGTM!

nemo/collections/nlp/modules/common/megatron/position_embedding/xpos_position_embedding.py
```diff
@@ -0,0 +1,78 @@
...
+        self.scale_base = scale_base
+        self.register_buffer("scale", (torch.arange(0, head_dim, 2) + 0.4 * head_dim) / (1.4 * head_dim))
+
+    def forward(self, x, offset=0, downscale=False):
+        length, b = x.shape[0], x.shape[1]
+        x = rearrange(x, 's b np hn -> (b np) s hn')
+        min_pos = -(length + offset) // 2
+        max_pos = length + offset + min_pos
+        scale = self.scale ** torch.arange(min_pos, max_pos, 1).to(self.scale).div(self.scale_base)[:, None]
+        sin, cos = fixed_pos_embedding(scale)
```
yidong72, line 66: It is a bit costly to dynamically calculate fixed_pos_embedding in the forward method.
Why don't we pre-calculate `inv_freq` tensor and `torch.arange(0, seq_len, dtype=torch.float)` in buffer and use them by indexing them instead of allocating GPU memory dynamically in the forward?

nemo/collections/nlp/modules/common/megatron/position_embedding/xpos_position_embedding.py
```diff
@@ -0,0 +1,78 @@
...
+    A simple version of `torch.repeat_interleave` for duplicating a matrix while interleaving the copy.
+    """
+    dim0 = m.shape[0]
+    m = m.view(-1, 1)  # flatten the matrix
+    m = m.repeat(1, 2)  # repeat all elements into the 2nd dimension
+    m = m.view(dim0, -1)  # reshape into a matrix, interleaving the copy
+    return m
+
+
+def apply_rotary_pos_emb(x, sin, cos, scale=1):
```
yidong72, line 46: It is strange to have `apply_rotary_pos_emb` inside the xpos position embedding. shall we rename it as `apply_pos_emb` ?

arendu:
FYI looks like this introduces a typo, made a fix here #6882

PR #6575, opened 2023-05-05
[TTS] Implement new TextToSpeech dataset

rlangman:
# What does this PR do ?

Add initial implementation of new TTS data loader which supports multi-manifest training and uses mostly preprocessed features.

The data loader is intended to be for TTS acoustic model and as such includes some specific functionality for text processing. Other models like vocoder will need another data loader, with most of the common code being shared in the features & tts_dataset_utils.

**Collection**: [TTS]

# Changelog 
- Implement TextToSpeechDataset class (following ASR naming convention of having separate X to X datasets for each task). 
- Add example FastPitch config for using new dataset.
- Modify feature.yaml files to use relative paths so they can be reused directly in the training script without needing to be nested in the fastpitch conf directory itself.
- Add code branches to FastPitch model to allow configuring new data loader for weighted sampling.
- Add `collate_fn` to featurizer class.
- Make log energy optional in FastPitch module so user can experiment with different normalization approaches.

# Usage
* This is an example command to train FastPitch on LJSpeech & VCTK with 0.2 and 0.8 sampling weight, respectively.

```bash
/home/rlangman/miniconda3/envs/nemo_dev/bin/python /home/rlangman/Code/NeMo/examples/tts/fastpitch.py
--config-path=/home/rlangman/Code/NeMo/examples/tts/conf/fastpitch
--config-dir=/home/rlangman/Code/NeMo/examples/tts/conf
--config-name=fastpitch_22050.yaml
n_speakers=110
speaker_path=/home/rlangman/Data/multi/ljspeech_vctk_speakers.json
max_epochs=1
weighted_sample_steps=10
phoneme_dict_path=/home/rlangman/Code/NeMo/scripts/tts_dataset_files/ipa_cmudict-0.7b_nv23.01.txt
heteronyms_path=/home/rlangman/Code/NeMo/scripts/tts_dataset_files/heteronyms-052722
feature_stats_path=/home/rlangman/Data/multi/ljspeech_vctk_stats.json
exp_manager.exp_dir=/home/rlangman/exps/ljspeech_vctk
+train_ds_meta.ljspeech._target_=nemo.collections.tts.data.text_to_speech_dataset.DatasetMeta
+train_ds_meta.ljspeech.manifest_path=/home/rlangman/Data/ljspeech/train_manifest.json
+train_ds_meta.ljspeech.audio_dir=/home/rlangman/Data/ljspeech/audio
+train_ds_meta.ljspeech.feature_dir=/home/rlangman/Data/ljspeech/features
[...]

PR #6670, opened 2023-05-17
[TTS] Implement new vocoder dataset

rlangman:
# What does this PR do ?

Implement new VocoderDataset to support multi-manifest training, which follows the same design pattern as the new TextToSpeechDataset (https://github.com/NVIDIA/NeMo/pull/6575).

**Collection**: [TTS]

# Changelog
- Add VocoderDataset implementation
- Add example config for using data loader
- Add logic in HiFi-GAN recipe for switching to new data loader (matching changes to support UnivNet recipe will be in a future PR).
- Add vocoder logging callback
- Use modified configure_optimizers() implementation from UnivNet recipe (https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/models/univnet.py#L106) to support schedulers other than CosineAnnealing. New config uses scheduler from original HiFi-GAN paper. 

# Usage

Example of training on 2 datasets, LJSpeech with 10% training examples and VCTK with 90%:

```
/home/rlangman/miniconda3/envs/nemo_dev/bin/python /home/rlangman/Code/NeMo/examples/tts/hifigan.py
--config-path=/home/rlangman/Code/NeMo/examples/tts/conf/hifigan
--config-dir=/home/rlangman/Code/NeMo/examples/tts/conf
--config-name=hifigan_22050_data.yaml
sample=22050
max_epochs=1000
weighted_sample_steps=1000
batch_size=16
trainer.devices=1
log_dir=/home/rlangman/exps/ljspeech_vctk/HifiGan/test/samples
exp_manager.exp_dir=/home/rlangman/exps/ljspeech_vctk
+exp_manager.version=test
+train_ds_meta.ljspeech.manifest_path=/home/rlangman/fs_mount/data/ljspeech/train_manifest.json
+train_ds_meta.ljspeech.audio_dir=/home/rlangman/fs_mount/data/ljspeech/audio
+train_ds_meta.ljspeech.sample_weight=0.1
+val_ds_meta.ljspeech.manifest_path=/home/rlangman/fs_mount/data/ljspeech/val_manifest.json
+val_ds_meta.ljspeech.audio_dir=/home/rlangman/fs_mount/data/ljspeech/audio
+log_ds_meta.ljspeech._target_=nemo.collections.tts.data.vocoder_dataset.DatasetMeta
+log_ds_meta.ljspeech.manifest_path=/home/rlangman/fs_mount/data/ljspeech/log_manifest.json
+log_ds_meta.ljspeech.audio_dir=/home/rlangman/fs_mount/data/ljspeech/audio
+train_ds_meta.vctk.manifest_path=/home/rlangman/fs_mount/data/vctk/train_manifest.json
+train_ds_meta.vctk.audio_dir=/home/rlangman/fs_mount/data/vctk/audio
+train_ds_meta.vctk.sample_weight=0.9
+val_ds_meta.vctk.manifest_path=/home/rlangman/fs_mount/data/vctk/val_manifest.json
+val_ds_meta.vctk.audio_dir=/home/rlangman/fs_mount/data/vctk/audio
+log_ds_meta.vctk._target_=nemo.collections.tts.data.vocoder_dataset.DatasetMeta
+log_ds_meta.vctk.manifest_path=/home/rlangman/fs_mount/data/vctk/log_manifest.json
+log_ds_meta.vctk.audio_dir=/home/rlangman/fs_mount/data/vctk/audio
exp_manager.create_wandb_logger=true
+exp_manager.wandb_logger_kwargs.project=hifigan_ljspeech_vctk
+exp_manager.wandb_logger_kwargs.name=test
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/tts/conf/hifigan/hifigan_22050_data.yaml
+++ b/examples/tts/conf/hifigan/hifigan_22050_data.yaml
@@ -1,0 +1,132 @@
+# This config contains the default values for training a 22khz HiFi-GAN model.
+# If you want to train model on other dataset, you can change config values according to your dataset.
+# Most dataset-specific arguments are in the head of the config file, see below.
+
+name: "HifiGan"
+
+max_epochs: ???
+batch_size: 16
+weighted_sample_steps: null
+
+train_ds_meta: ???
+val_ds_meta: ???
+log_ds_meta: ???
+
+log_dir: ???
+
+defaults:
+  - model/generator: v1
+  - feature: feature_22050
+
+model:
+
+  max_epochs: ${max_epochs}
+  steps_per_epoch: ${weighted_sample_steps}
+  l1_loss_factor: 60
+
+  preprocessor:
+    _target_: nemo.collections.asr.parts.preprocessing.features.FilterbankFeatures
+    nfilt: ${feature.mel_feature.mel_dim}
+    lowfreq: ${feature.mel_feature.lowfreq}
+    highfreq: ${feature.mel_feature.highfreq}
+    n_fft: ${feature.win_length}
+    n_window_size: ${feature.win_length}
+    n_window_stride: ${feature.hop_length}
+    pad_to: 0
+    pad_value: 0
+    exact_pad: true
+    sample_rate: ${feature.sample_rate}
+    window: hann
+    normalize: null
+    preemph: null
+    dither: 0.0
+    frame_splicing: 1
+    log: true
+    log_zero_guard_type: add
+    log_zero_guard_value: 1.0
+    mag_power: 1.0
+    mel_norm: null
+    use_grads: false
+
+  train_ds:
+    dataset:
+      _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
+      weighted_sample_steps: ${weighted_sample_steps}
+      sample_rate: ${feature.sample_rate}
+      n_segments: 8192
+      min_duration: 0.5
+      max_duration: null
+      dataset_meta: ${train_ds_meta}
+
+    dataloader_params:
+      batch_size: ${batch_size}
+      num_workers: 4
+
+  validation_ds:
+    dataset:
+      _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
+      sample_rate: ${feature.sample_rate}
+      n_segments: 66048
+      min_duration: 3.0
+      max_duration: null
+      dataset_meta: ${val_ds_meta}
+
+    dataloader_params:
+      batch_size: ${batch_size}
+      num_workers: 2
+
+  log_config:
+    log_dir: ${log_dir}
+    log_epochs: [10, 50]
+    epoch_frequency: 100
+    log_tensorboard: false
+    log_wandb: false
+
+    generators:
+      - _target_: nemo.collections.tts.parts.utils.callbacks.VocoderArtifactGenerator
+
+    dataset:
+      _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
+      sample_rate: ${feature.sample_rate}
+      n_segments: null
+      min_duration: null
+      max_duration: null
+      dataset_meta: ${log_ds_meta}
+
+    dataloader_params:
+      batch_size: 4
+      num_workers: 2
+
+  optim:
+    _target_: torch.optim.AdamW
+    lr: 2e-4
+    betas: [0.8, 0.99]
+    weight_decay: 1e-6
+    sched:
+      name: ExponentialLR
+      gamma: 0.999
+
+trainer:
+  num_nodes: 1
+  devices: 1
+  accelerator: gpu
+  strategy: ddp
+  precision: 16
+  max_epochs: ${max_epochs}
+  accumulate_grad_batches: 1
+  enable_checkpointing: False  # Provided by exp_manager
+  logger: false # Provided by exp_manager
+  log_every_n_steps: 100
+  check_val_every_n_epoch: 10
+  benchmark: false
+
+exp_manager:
+  exp_dir: null
+  name: ${name}
+  create_tensorboard_logger: true
+  create_checkpoint_callback: true
+  create_wandb_logger: false
+  checkpoint_callback_params:
+    monitor: val_loss
+  resume_if_exists: false
+  resume_ignore_no_checkpoint: false
--- a/examples/tts/conf/hifigan/hifigan_data.yaml
+++ b/examples/tts/conf/hifigan/hifigan_data.yaml
@@ -1,8 +1,13 @@
-# This config contains the default values for training a 22khz HiFi-GAN model.
+# This config contains the default values for training a HiFi-GAN model.
 # If you want to train model on other dataset, you can change config values according to your dataset.
 # Most dataset-specific arguments are in the head of the config file, see below.
 
 name: "HifiGan"
+
+defaults:
+  - sample: ???
+  - feature: feature_${sample}
+  - model/generator: v1_${sample}
 
 max_epochs: ???
 batch_size: 16
@@ -13,10 +18,6 @@ val_ds_meta: ???
 log_ds_meta: ???
 
 log_dir: ???
-
-defaults:
-  - model/generator: v1
-  - feature: feature_22050
 
 model:
 
@@ -35,7 +36,7 @@ model:
     pad_to: 0
     pad_value: 0
     exact_pad: true
-    sample_rate: ${feature.sample_rate}
+    sample_rate: ${sample.sample_rate}
     window: hann
     normalize: null
     preemph: null
@@ -52,9 +53,9 @@ model:
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
       weighted_sample_steps: ${weighted_sample_steps}
-      sample_rate: ${feature.sample_rate}
-      n_segments: 8192
-      min_duration: 0.5
+      sample_rate: ${sample.sample_rate}
+      n_segments: ${sample.train_n_segments}
+      min_duration: 0.4
       max_duration: null
       dataset_meta: ${train_ds_meta}
 
@@ -65,8 +66,8 @@ model:
   validation_ds:
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
-      sample_rate: ${feature.sample_rate}
-      n_segments: 66048
+      sample_rate: ${sample.sample_rate}
+      n_segments: ${sample.val_n_segments}
       min_duration: 3.0
       max_duration: null
       dataset_meta: ${val_ds_meta}
@@ -87,7 +88,7 @@ model:
 
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
-      sample_rate: ${feature.sample_rate}
+      sample_rate: ${sample.sample_rate}
       n_segments: null
       min_duration: null
       max_duration: null
--- a/examples/tts/conf/hifigan/hifigan_data.yaml
+++ b/examples/tts/conf/hifigan/hifigan_data.yaml
@@ -5,9 +5,9 @@
 name: "HifiGan"
 
 defaults:
+  - feature: ???
   - sample: ???
-  - feature: feature_${sample}
-  - model/generator: v1_${sample}
+  - model/generator: ???
 
 max_epochs: ???
 batch_size: 16
@@ -36,7 +36,7 @@ model:
     pad_to: 0
     pad_value: 0
     exact_pad: true
-    sample_rate: ${sample.sample_rate}
+    sample_rate: ${feature.sample_rate}
     window: hann
     normalize: null
     preemph: null
@@ -53,8 +53,8 @@ model:
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
       weighted_sample_steps: ${weighted_sample_steps}
-      sample_rate: ${sample.sample_rate}
-      n_segments: ${sample.train_n_segments}
+      sample_rate: ${feature.sample_rate}
+      n_samples: ${sample.train_n_samples}
       min_duration: 0.4
       max_duration: null
       dataset_meta: ${train_ds_meta}
@@ -66,8 +66,8 @@ model:
   validation_ds:
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
-      sample_rate: ${sample.sample_rate}
-      n_segments: ${sample.val_n_segments}
+      sample_rate: ${feature.sample_rate}
+      n_samples: ${sample.val_n_samples}
       min_duration: 3.0
       max_duration: null
       dataset_meta: ${val_ds_meta}
@@ -88,8 +88,8 @@ model:
 
     dataset:
       _target_: nemo.collections.tts.data.vocoder_dataset.VocoderDataset
-      sample_rate: ${sample.sample_rate}
-      n_segments: null
+      sample_rate: ${feature.sample_rate}
+      n_samples: null
       min_duration: null
       max_duration: null
       dataset_meta: ${log_ds_meta}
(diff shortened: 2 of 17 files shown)


nemo/collections/tts/models/hifigan.py
```diff
@@ -69,9 +72,20 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
         if self._train_dl:
             self.input_as_mel = self._train_dl.dataset.load_precomputed_mel
 
+        self.log_audio = cfg.get("log_audio", False)
+        self.log_config = cfg.get("log_config", None)
         self.automatic_optimization = False
 
     def _get_max_steps(self):
```
racoiaws, line 79: Is it possible to change this into a property? Would be more pythonic
```
@property
def max_steps(self):
   ... same code ...
```

examples/tts/conf/hifigan/hifigan_22050_data.yaml
```diff
@@ -0,0 +1,132 @@
...
+# If you want to train model on other dataset, you can change config values according to your dataset.
+# Most dataset-specific arguments are in the head of the config file, see below.
+
+name: "HifiGan"
+
+max_epochs: ???
+batch_size: 16
+weighted_sample_steps: null
+
+train_ds_meta: ???
```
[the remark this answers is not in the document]
rlangman: Yes. I have added a usage example to the PR writeup.

nemo/collections/tts/parts/utils/callbacks.py
```diff
@@ -229,6 +229,39 @@ def on_train_epoch_end(self, trainer: Trainer, model: LightningModule):
...
+        audio_ids = [create_id(p) for p in audio_filepaths]
+
+        audio = batch_dict.get("audio")
+        audio_len = batch_dict.get("audio_lens")
+
+        spec, spec_len = model.audio_to_melspec_precessor(audio, audio_len)
+
+        with torch.no_grad():
+            # [B, T]
+            audio_pred = model.forward(spec=spec).squeeze(1)
```
racoiaws, line 253: (A nitpick)
Would be a nice place to use `einops.rearrange`, it is in NeMo dependencies. Makes tensor dimension dances **much** more explicit and readable.
```
audio_pred = rearrange(audio_pred, "batch 1 time -> batch time")
```

examples/tts/conf/hifigan/hifigan_22050_data.yaml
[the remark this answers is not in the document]
racoiaws: Do you think this these dataclasses could be turned into `TypedDict`?

The flow of "YAML dicts <-> Python dicts" is broken here, with additional verbosity.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -0,0 +1,181 @@
...
+        self.data_samples = []
+        self.sample_weights = []
+        for dataset_name, dataset in dataset_meta.items():
+            samples, weights = self._process_dataset(
+                dataset_name=dataset_name, dataset=dataset, min_duration=min_duration, max_duration=max_duration,
+            )
+            self.data_samples += samples
+            self.sample_weights += weights
+
+    def get_sampler(self, batch_size: int) -> Optional[torch.utils.data.Sampler]:
```
racoiaws, line 86: Also looks like a `@property`

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -0,0 +1,181 @@
...
+            audio_array, _ = librosa.load(audio_filepath, sr=self.sample_rate)
+        else:
+            audio_segment = self._segment_audio(audio_filepath)
+            audio_array = audio_segment.samples
+        audio = torch.tensor(audio_array)
+        audio_len = torch.tensor(audio.shape[0])
+        return audio, audio_len
+
+    @staticmethod
+    def _process_dataset(
```
racoiaws, line 119: The name is misleading, this method doesn't process a `Dataset`. Maybe `_preprocess_manifest`?

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self.sample_rate = sample_rate
+        self.n_segments = n_segments
+        self.weighted_sample_steps = weighted_sample_steps
+        self.num_audio_retries = num_audio_retries
+        self.load_precomputed_mel = False
+
+        if feature_processors:
+            logging.info(f"Found feature processors {feature_processors.keys()}")
+            self.feature_processors = feature_processors.values()
```
racoiaws, line 73: (A nitpick)

`.values()` [doesn't return a list](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects)

If anyone ever does something like this:
```
feature_processors = ...
dataset1 = VocoderDataset(..., feature_processors, ...)
feature_processors["another_processor"] = ...
dataset2 = VocoderDataset(..., feature_processors, ...)
```
then, both datasets would keep reference to the original collection, and would have the same list of feature processors. Which is likely not intended.

Turning it into a list will make it a bit more foolproof:
```
self.feature_processors = list(feature_processors.values())
```

examples/tts/conf/hifigan/hifigan_22050_data.yaml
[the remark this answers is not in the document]
rlangman: I have not used TypedDict much. My understanding is that the advantages/disadvantages of using it would be:

**Advantages**
1. Dict type is naturally supported by YAML config, avoiding repeated _target_ specification.
2. TypedDict is backwards compatible with code that previously used dictionary (not applicable here).

**Disadvantages**
1. Does not validate keys; user could be missing a required field or have a typo and the error would not be caught immediately.
2. Code using it is required to access fields like dictionary instead of object (ie. metadata["field"] vs metadata.field)

Of these I think the most important is dataclass validating that the correct keys are provided.

In the rest of NeMo, there is one class which uses TypedDict for input parameter groups (https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/modules/transformer/text_generation.py#L26). Everything else uses dataclass.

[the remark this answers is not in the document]
racoiaws: Here's my concern. The hydra "_target_" pattern is typically a signal to the user that something could be customised using their own class. Here, this dataclass is intricately tied to a specific Dataset, making the `_target_` just a boilerplate in the log. Which is also easy to have a typo in.

Maybe the data class could be used only internally? Keeps the validation advantage.
```
class VocoderDataset(Dataset):
    def __init__(șelf, ..., dataset_meta: Dict[str, Dict], ...):
        ...
        for dataset_name, dataset in dataset_meta.items():
            dataset = DatasetMeta(**dataset)
            ...
```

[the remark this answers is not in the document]
rlangman: Makes sense to me. I will try to refactor it using the data class internally.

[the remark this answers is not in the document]
rlangman: Changed

nemo/collections/tts/parts/utils/callbacks.py
rlangman: Changed

nemo/collections/tts/data/vocoder_dataset.py
rlangman: For it to be a property it cannot take `batch_size` as input. Where the batch size (and sampler) are inputs to the pytorch DataLoader rather than the Dataset.

The main alternative would be to move this function out out of the Dataset class and into `dataset_utils`.

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -115,26 +115,31 @@ def __init__(
...
+            self.feature_processors = list(feature_processors.values())
         else:
             self.feature_processors = []
 
-        self.align_prior_config = align_prior_config
-        if self.align_prior_config.use_beta_binomial_interpolator:
-            self.beta_binomial_interpolator = BetaBinomialInterpolator()
+        if align_prior_config:
+            self.align_prior_config = AlignPriorConfig(**align_prior_config)
+            if self.align_prior_config.use_beta_binomial_interpolator:
```
racoiaws, line 130: Since this a new dataset class, could we just drop the binomial interpolator?

I think we just forgot to deprecate it, it's been obsolete for some time now

examples/tts/conf/hifigan/hifigan_data.yaml
```diff
@@ -0,0 +1,133 @@
+# This config contains the default values for training a HiFi-GAN model.
+# If you want to train model on other dataset, you can change config values according to your dataset.
+# Most dataset-specific arguments are in the head of the config file, see below.
+
+name: "HifiGan"
+
+defaults:
+  - sample: ???
```
racoiaws, line 8: (Let's say I'm new user reading this config)

I don't quite get what this `sample` means neither from name nor from context. What should i put here if want to train a model on my dataset?

nemo/collections/tts/data/vocoder_dataset.py
rlangman: It is called in the model/ classes, so It can be reused in multiple training different training recipes. Right now in this PR FastPitch is the only one using the TTS sampler and HiFi-GAN the vocoder, but I will be adding it to UnivNet and EnCodec recipes soon and if we deprecate old data loader eventually then all recipes would use it.

examples/tts/conf/hifigan/hifigan_data.yaml
rlangman: In the defaults lists, `sample: ???` can only mean it is the name of config file in the `sample` directory. It cannot be anything else, like a string value.

Otherwise, I would like to do something like:

```
sample_rate: ???

defaults:
  - sample: sample_${sample_rate}
  - feature: feature_${sample_rate}
  - model/generator: v1_${sample_rate}
```

But my current understanding is that this is impossible to do.

Or is your point more than I should rename the parameter and config directory name from `sample` to `sample_rate`?

I could also add more comments to try to tell a user what some of the other `???` parameters mean, but I doubt anyone will be able to figure out how to use it regardless without seeing a full example/tutorial.

nemo/collections/tts/data/text_to_speech_dataset.py
rlangman: I think we can drop it. I did not want to remove it because I was having issues with alignment during testing. But I think I found the root cause being that me redefining an "epoch" as 1,000 steps was causing the aligner binarization loss to warm up too quickly on large datasets. Which can be fixed by reducing the scale on the binarization loss, or making 'bin_loss_warmup_epochs' larger.

I think it is similar to the alignment issues we were seeing with FastPitch recently where aligner would not learn properly during fine-tuning.

examples/tts/conf/hifigan/hifigan_data.yaml
racoiaws: I don't think grouping of these parameters is good in the first place.

Sampling rate is a property of dataset we're building a model for. Other two parameters have more to do with available GPU memory than with signals themselves. I think this is why it is hard to come up with a good name for the group - they don't really belong together. I hope you agree with me that `cfg.sample.sample_rate` looks a bit awkward

I'd ask for renaming `train_n_segments` to `train_n_samples` though, if it doesn't harm backward compatibility.

examples/tts/conf/hifigan/hifigan_data.yaml
```diff
@@ -0,0 +1,133 @@
...
+train_ds_meta: ???
+val_ds_meta: ???
+log_ds_meta: ???
+
+log_dir: ???
+
+model:
+
+  max_epochs: ${max_epochs}
+  steps_per_epoch: ${weighted_sample_steps}
```
racoiaws, line 25: What does `weighted_sample_steps` mean?

examples/tts/conf/hifigan/hifigan_data.yaml
rlangman: I think of having an easy one line way to group all parameters than depend on the sample rate is a good idea. However the pre-existing config hierarchy and vocoder code does not provide an optimal way to do it.

So if we think this looks too awkward, I will require the user to provide all 3 configs separately in the CLI.

nemo/collections/tts/data/vocoder_dataset.py
racoiaws: OK, if you think having this method is preferable to directly calling `get_weighted_sampler`, let's leave it as is

examples/tts/conf/hifigan/hifigan_data.yaml
rlangman: Dataloader docstring: `Optional int, If provided, then data will be sampled (with replacement) based on the sample weights provided in the dataset metadata. If None, then sample weights will be ignored.`

In other words, with weighted sampling (and with large datasets in general) the standard definition of an epoch no longer works so you refer to an epoch as a fixed number of training steps.

The name is not as clear as I would like, but alternatives I thought of were either too long (like `weighted_sampling_steps_per_epoch`) or required breaking it into 2 arguments like:

```
WeightedSamplingConfig:
  use_weighted_sampling: true
  steps_per_epoch: 1000
```

Where `steps_per_epoch` does nothing if `use_weighted_sampling: false`.

examples/tts/conf/hifigan/hifigan_data.yaml
rlangman: Changed

examples/tts/conf/hifigan/hifigan_data.yaml
racoiaws: > In other words, with weighted sampling (and with large datasets in general) the standard definition of an epoch no longer works so you refer to an epoch as a fixed number of training steps.

Can't really agree, I always thought of epoch as of something dependent on size of a specific dataset. Number of training examples in the dataset still works, with the caveat that some samples might not have been utilized in a given epoch run.

(Not a blocker)

>  but alternatives I thought of were either too long (like `weighted_sampling_steps_per_epoch`)
I see. True, this is not something that can be expressed consisely. But I see no harm in a long-named argument. I would  even suggest `steps_per_epoch_if_using_weighted_sampling`

Do you thing your could you rename this into a longer and more descriptive version?

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -78,8 +71,8 @@ class TextToSpeechDataset(Dataset):
         featurizers: Optional, list of featurizers to load feature data from. Should be the same config provided
             when running scripts.dataset_processing.tts.compute_features.py before training.
         feature_processors: Optional, list of feature processors to run on training examples.
-        align_prior_config: Optional, if provided alignment prior will be calculated and included in
-            batch output.
+        align_prior_hop_length: Optional int, hop length of audio features.
+            If provided alignment prior will be calculated and included in batch output.
```
racoiaws, line 75: Could you please also add a comment on how it usually is equal to `hop_lengh` of spectrogram STFT?

Wouldn't hurt for this to be explicitly stated somewhere outside the default config.

racoiaws:
Apologies if this is taking too long, after resolving two threads above this PR should be good to go.

nemo/collections/tts/data/text_to_speech_dataset.py
rlangman: Added. Mentioned it as feature hop_length as we may not always be using STFT based features.

examples/tts/conf/hifigan/hifigan_data.yaml
rlangman: I tried adding in alternate config which could optionally set `epoch_size = dataset_size` (even if not all samples are seen each epoch), but the way that an epoch is defined for distributed training is relatively complicated (https://github.com/NVIDIA/NeMo/blob/main/nemo/core/optim/lr_scheduler.py#L934-L936). Getting it wrong will create a mismatch with the LR scheduler. So I would rather not support it for now unless there is an urgent need to.

nemo/collections/tts/data/text_to_speech_dataset.py
racoiaws: Yes, thank you, that makes more sense

racoiaws (approved):

Issue #6584, opened 2023-05-07
TTS Aligner Inference Example outputs attn_soft_tensor with nan values 


kn:
**Describe the bug**

I simply followed the [TTS Aligner Inference Example](https://github.com/NVIDIA/NeMo/blob/stable/tutorials/tts/Aligner_Inference_Examples.ipynb) but the visualized `attn_soft` is empty and indeed `attn_soft_tensor` has all `nan` values.

<img width="1082" alt="Screen Shot 2023-05-06 at 10 43 29 PM" src="https://user-images.githubusercontent.com/621009/236655018-9c65c525-684c-42f6-ba1a-45e67024113f.png">

**Steps/Code to reproduce bug**

You can simply proceed through [this notebook]((https://github.com/NVIDIA/NeMo/blob/stable/tutorials/tts/Aligner_Inference_Examples.ipynb)

**Expected behavior**

Display a graph with values described in the instruction.

**Environment overview (please complete the following information)**

Google Collab

PR #5653, opened 2022-12-16
No-script TS export, prepared for ONNX export 

borisfom:
Signed-off-by: Boris Fomitchev <bfomitchev@nvidia.com>

# What does this PR do ?
Changed unfold to reshape for inference to be able to export in ONNX.
Also using extra hx parameter to LSTM call to work around Pytorch issues with tracing the hx=None.
Now Torchscript export works without any scripting - so I was also able to get rid of unnecessary autocast=False block.
ONNX export still fails due to internal bugs, even though each part of the network is now ONNX-exportable.

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #6677, opened 2023-05-18
Fix masking bug for TTS Aligner

redoctopus:
# What does this PR do ?

Fixes a masking bug where boolean values were inverted from what was expected. Reverted change from PR #5653.

**Collection**: TTS

# Changelog 
- Added check for `== 0` (applied to the input mask) back to the Aligner forward call. Otherwise the mask is applied incorrectly, as the values are flipped from what is expected in the Aligner module.

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information
* See Issue #6584


rlangman (approved):
It is weird that the aligner _forward()_ method reverses the standard convention for mask inputs (it should flip the entries internally before calling masked_fill_). But for backwards compatibility I suppose we should not modify the aligner interface.

PR #6702, opened 2023-05-23
[TTS] corrected misleading deprecation warnings.

XuesongYang:
# What does this PR do ?

1. correct the misleading deprecation warnings;
2. organize the duplicate warning and replacing code snippet into a single function.
3. passed `pytest --nightly tests/collections/tts/models`.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/tts/models/aligner.py
+++ b/nemo/collections/tts/models/aligner.py
@@ -104,7 +104,11 @@ class AlignerModel(ModelPT):
         text_tokenizer_kwargs = {}
         if "g2p" in cfg.text_tokenizer:
             # for backward compatibility
-            if self._is_model_being_restored():
+            if (
+                self._is_model_being_restored()
+                and (cfg.text_tokenizer.g2p.get('_target_', None) is not None)
+                and cfg.text_tokenizer.g2p["_target_"].startswith("nemo_text_processing.g2p")
+            ):
                 cfg.text_tokenizer.g2p["_target_"] = g2p_backward_compatible_support(
                     cfg.text_tokenizer.g2p["_target_"]
                 )
--- a/nemo/collections/tts/models/aligner.py
+++ b/nemo/collections/tts/models/aligner.py
@@ -24,7 +24,12 @@ from pytorch_lightning.loggers import WandbLogger
 from torch import nn
 
 from nemo.collections.tts.losses.aligner_loss import BinLoss, ForwardSumLoss
-from nemo.collections.tts.parts.utils.helpers import binarize_attention, get_mask_from_lengths, plot_alignment_to_numpy
+from nemo.collections.tts.parts.utils.helpers import (
+    binarize_attention,
+    g2p_backward_compatible_support,
+    get_mask_from_lengths,
+    plot_alignment_to_numpy,
+)
 from nemo.core.classes import ModelPT
 from nemo.core.classes.common import PretrainedModelInfo
 from nemo.utils import logging, model_utils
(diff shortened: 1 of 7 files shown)


nemo/collections/tts/models/aligner.py
```diff
@@ -99,11 +104,10 @@ def _setup_tokenizer(self, cfg):
         text_tokenizer_kwargs = {}
         if "g2p" in cfg.text_tokenizer:
             # for backward compatibility
-            if self._is_model_being_restored() and cfg.text_tokenizer.g2p.get('_target_', None):
```
XuesongYang, line 102: as an FYI, I remvoed `cfg.text_tokenizer.g2p.get('_target_', None)` because "AttributeError: 'NoneType' object has no attribute 'replace'"

redoctopus: Sure, as long as we're certain a config won't have this missing.

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -825,3 +826,14 @@ def sample_tts_input(
             0, export_config["num_speakers"], (max_batch,), device=device, dtype=torch.int64
         )
     return inputs
+
+
+@deprecated(
+    explanation="But it will not be removed until a further notice. G2P object root directory "
+    "`nemo_text_processing.g2p` has been replaced with `nemo.collections.tts.g2p`. "
```
redoctopus, line 833: Should we commit to a particular release to remove this? Maybe within 1-2 releases?

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -825,3 +826,14 @@ def sample_tts_input(
...
+    "Please use the latter instead since NeMo 1.18.0."
```
redoctopus, line 834: Nitpick: s/since/as of

nemo/collections/tts/models/aligner.py
XuesongYang: hmm...`_target_` should be a required argument for `hydra.instantiate()`.

nemo/collections/tts/parts/utils/helpers.py
XuesongYang: I think 1.18 was the release that we are trying to remove this replacement for `nemo_text_processing.g2p`, which means all old models before 1.18 will raise errors for the import path. We have no plan now to re-train all old models. But we can modify the state_dict for all old models if we plan to remove such deprecation warnings and back-compatibility replacement.

XuesongYang: So for now, it is good to keep the warning and replacement op until re-train or update state_dict is done.

rlangman: I don't see why retraining all of our models would be a priority (unless we have several other deprecations we also want to resolve).

But if we do not commit to a timeline, then it would be clearer to only print the deprecation warning when the checkpoint is using the old class. Will this annotation result in all model restorations printing the message?

XuesongYang: good point. This warning should be only triggered when model class uses `nemo_text_processing.g2p`.

redoctopus (dismissed):
LGTM

redoctopus (approved):

PR #6708, opened 2023-05-23
check for first or last stage

ericharper:
# What does this PR do ?

Fix bug when `inference.compute_logprob=True`. Only first and last pipeline stages have a response.

**Collection**: NLP

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/modules/common/text_generation_utils.py
+++ b/nemo/collections/nlp/modules/common/text_generation_utils.py
@@ -136,7 +136,7 @@ def megatron_gpt_generate(model, inputs, tokenizer, length_params, sampling_para
 
 def get_computeprob_response(tokenizer, response, inputs):
     if parallel_state.is_pipeline_first_stage() or parallel_state.is_pipeline_last_stage():
-        # we only have a reseponse on the first and last pipeline stages
+        # we only have a response on the first and last pipeline stages
         compute_prob_response = {}
         new_token_ids = []
         new_tokens = []
--- a/nemo/collections/nlp/modules/common/text_generation_utils.py
+++ b/nemo/collections/nlp/modules/common/text_generation_utils.py
@@ -135,36 +135,45 @@ def megatron_gpt_generate(model, inputs, tokenizer, length_params, sampling_para
 
 
 def get_computeprob_response(tokenizer, response, inputs):
-    compute_prob_response = {}
-    new_token_ids = []
-    new_tokens = []
-    new_texts = []
-    log_probs = []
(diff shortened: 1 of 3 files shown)


titu1994 (dismissed):
Thanks!

nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -135,36 +135,45 @@ def megatron_gpt_generate(model, inputs, tokenizer, length_params, sampling_para
...
-        offsets.append(response['offsets'][batch_id][:-1])
-    compute_prob_response['sentences'] = new_texts
-    compute_prob_response['tokens'] = new_tokens
-    compute_prob_response['token_ids'] = new_token_ids
-    compute_prob_response['logprob'] = log_probs
-    compute_prob_response['full_logprob'] = full_logprobs
-    compute_prob_response['offsets'] = offsets
-    return compute_prob_response
+    if (
+        not parallel_state.model_parallel_is_initialized()
```
yidong72, line 139: why we need this condition `not parallel_state.model_parallel_is_initialized()`? 
if the model_parallel is not initialized, it will throw an exception at the very beginning right?

yidong72 (commented):
Looks good. just one comment.

nemo/collections/nlp/modules/common/text_generation_utils.py
ericharper: I was just trying to make it robust. This method will only be called after parallel_state is initialized?

yidong72: this `parallel_state.model_parallel_is_initialized()` method will always be true.
```python
def model_parallel_is_initialized():
    """Check if model and data parallel groups are initialized."""
    if _TENSOR_MODEL_PARALLEL_GROUP is None or \
        _PIPELINE_MODEL_PARALLEL_GROUP is None or \
        _DATA_PARALLEL_GROUP is None:
        return False
    return True
```

ericharper: I mean will this method ever be used without model parallel init?

yidong72: no. it is only used with model parallel.

yidong72 (approved):

PR #6716, opened 2023-05-24
[TTS] Filter out silent audio files during preprocessing

rlangman:
# What does this PR do ?

First version of preprocessing ignored audio files where start/end of speech could not be detected. During testing I find that all such examples are silent audio files with no speech. And that most public datasets have several such files. This updates the script to remove such files from the final training manifest.

**Collection**: [TTS]

# Changelog 
- Update preprocessing script to remove silent/empty audio files from final training manifest.
- Add `overwrite_audio` so that is script fails partway, you can resume where you left off instead of having to reprocess the entire dataset.
- Fix parsing of `overwrite` arguments, as trying to parse them as a bool results in `True` regardless of what value you provide in the CLI.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


redoctopus (approved):
LGTM, thanks for the fixes

PR #6806, opened 2023-06-02
[TTS] Add cosine distance option to TTS aligner

rlangman:
# What does this PR do ?

Add configuration to TTS aligner which uses scaled cosine distance instead of euclidean distance.

**Collection**: [TTS]

# Changelog 
- Add code for cosine distance based alignment

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

# Additional Information

The current aligner used for FastPitch training has a few problem:

1. It aligns slowly for datasets with lots of speakers.
2. If binarization loss is enabled/warmup too soon then alignment learning stops before accurate alignments are learned.
3. When fine-tuning, the model does not learn alignment of target speaker well.

Using cosine distance instead of L2 distance can fix these problems. Cosine distance in general a lot more accurate and stable when it comes to similarity problems. It also avoids possible issues with the binarization loss because the scale of the cosine distance is fixed regardless of the scale or content of the alignment embeddings.

To test this I trained FastPitch on the VCTK dataset for 300k steps and provide the alignments and loss graphs below.

As seen, using cosine distance aligns faster, produces more accurate and sharper alignments, has a lot lower alignment loss (especially binarization loss), and as a result converges to a model with significantly better accuracy all around (mel, duration, pitch, energy). Though the gap between the two would likely shrink if we train it for several million steps.

**Alignments**

_Example 1_

`L2`
align_l2_1
`Cosine`
align_cos_1

_Example 2_

`L2`
align_l2_2
`Cosine`
align_cos_2

**Alignment Losses**

ctc_loss
bin_loss

**Reconstruction Losses**

mel_loss
dur_loss
pitch_loss
energy_loss


examples/tts/conf/fastpitch/fastpitch.yaml
```diff
@@ -193,6 +193,8 @@ model:
   alignment_module:
     _target_: nemo.collections.tts.modules.aligner.AlignmentEncoder
     n_text_channels: ${model.symbols_embedding_dim}
+    dist_type: cosine
```
XuesongYang, line 196: nit: it seems we now only support L2 and Cosine that are hidden in the aligner.py. It would be great to mention supported ones as a comment.

examples/tts/conf/fastpitch/fastpitch.yaml
```diff
@@ -193,6 +193,8 @@ model:
...
+    temperature: 15.0
```
XuesongYang, line 197: any guidance on the empirical number?

nemo/collections/tts/modules/aligner.py
```diff
@@ -57,15 +78,35 @@ def get_dist(self, keys, queries, mask=None):
         Output:
             dist (torch.tensor): B x T1 x T2 tensor.
         """
-        keys_enc = self.key_proj(keys)  # B x n_attn_dims x T2
-        queries_enc = self.query_proj(queries)  # B x n_attn_dims x T1
-        attn = (queries_enc[:, :, :, None] - keys_enc[:, :, None]) ** 2  # B x n_attn_dims x T1 x T2
-        dist = attn.sum(1, keepdim=True)  # B x 1 x T1 x T2
+        # B x n_attn_dims x T2
```
XuesongYang, line 81: shall we use consistent dimension alias in the context? I believe `n_attn_dims` and `C` are meaning the same. It's up to you to decide one of them.

nemo/collections/tts/modules/aligner.py
```diff
@@ -57,15 +78,35 @@ def get_dist(self, keys, queries, mask=None):
...
+        keys_enc = self.key_proj(keys)
+        # B x n_attn_dims x T1
```
XuesongYang, line 83: same comment as the above.

nemo/collections/tts/modules/aligner.py
```diff
@@ -57,15 +78,35 @@ def get_dist(self, keys, queries, mask=None):
...
+        self._apply_mask(dist, mask, float("inf"))
 
-        if mask is not None:
-            dist.data.masked_fill_(mask.permute(0, 2, 1).unsqueeze(2), float("inf"))
+        return dist
 
-        return dist.squeeze(1)
+    @staticmethod
+    def get_euclidean_dist(queries, keys):
+        queries = rearrange(queries, "B C T1 -> B C T1 1")
```
XuesongYang, line 95: same comment as the above. Please choose either `C` or `n_attn_dims`, but not mix usage of them.

XuesongYang (approved):
LGTM. Added some neat-picks.

One discussion: I observed many `@staticmethod` member functions defined in `class AlignmentEncoder`. I wonder if `nemo.collections.tts.parts.utils.helpers` is a better place to hold them all?

examples/tts/conf/fastpitch/fastpitch.yaml
rlangman: Added documentation to the aligner constructor.

examples/tts/conf/fastpitch/fastpitch.yaml
rlangman: I just found ~15 to work well by testing multiples of 5 (I assume the L2 distance temperature of 0.0005 was similarly found by trial and error). Its been about a year since I experimented with it, but I vaguely recall 30 not working and 10 converging too slow.

XuesongYang (approved):

SaltedSlark:
@XuesongYang Hello friend, I got a problem on aligner. When I try to train fastpitch on a chinese dialect dataset, it seems that the ctc_loss cannot converge, and the generated audio is on a low quality and tends to missing some words,like this:

image

but when I train the same code on another manderin dataset, it converges normally, the generate audio is pretty good.
image


I have no idea what wrong. 
Looking for your relply

PR #6865, opened 2023-06-13
Update transcribe_utils.py

stevehuang52:
fix ctc decoding for hybrid model in partial transcribe


erastorgueva-nv (approved):
Change looks good. I can confirm partial transcription with Hybrid model in CTC decoding mode runs for me now 👍

PR #6866, opened 2023-06-14
fix ptuning residuals bug

arendu:
# What does this PR do ?

- fixes peft ptuning bug which was causing ptuning to use a residual connection
- simplifies mixin use for all NLP PEFT models
- ptuning includes a inference only table which is needed for FT 
- 
**Collection**: [NLP]

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -268,7 +274,15 @@ def load_state_dict(self, state_dict, strict: bool = True):
 
     def setup_optimizer_param_groups(self):
         if self.first_stage_of_pipeline():
-            super().setup_optimizer_param_groups()
+            # super().setup_optimizer_param_groups()
+            self.freeze()  # Freeze the entire model
+            opt_params = []
+            for n, p in self.named_parameters():
+                if n in self.trainable_keys:
```
titu1994, line 281: Are trainable keys the full list of all the params of the adapters? Or just prefix of the adapter name?

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
```diff
@@ -112,7 +110,6 @@ def __init__(
         row_init_method: str = 'zero',  # TODO: (@adithyare) should rename this to output_init_method to be more precise.
         gather_output: bool = True,
         dropout: float = 0.0,
-        adapter_strategy: adapter_mixin_strategies.ResidualAddAdapterStrategyConfig = None,
```
titu1994, line 115: I think in some place, the adapter engine does try to call the module with this strategy - hopefully removing it does not break anything. 
An alternative is to simply replace the default value.

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
```diff
@@ -301,15 +292,47 @@ def __init__(
...
+        self.inference_table.set_prompt_table(prompt_representation)
+
+    def clear_inference_table(self,):
+        self.inference_table.clear_prompt_table()
+
+    def get_inference_table(self,):
+        return self.inference_table.get_prompt_table()
 
-    def forward(self, batch_size):
+    def _forward(self,):
```
titu1994, line 310: Better to not make this private for subclasses, rename to forward_inner()

nemo/collections/nlp/modules/common/megatron/language_model.py
```diff
@@ -746,10 +746,7 @@ def forward(
                 ptuning_adapter = self.get_adapter_module(AdapterName.PTUNING_ADAPTER)
                 v = ptuning_adapter.virtual_tokens
                 if ptuning_adapter and _sq >= v:  # The sequence should be longer the v to insert virtual embeddings.
-                    strategy = ptuning_adapter.adapter_strategy
-                    virtual_embeddings = self.forward_single_enabled_adapter_(
-                        _bs, ptuning_adapter, adapter_name=AdapterName.PTUNING_ADAPTER, adapter_strategy=strategy,
-                    )
+                    virtual_embeddings = ptuning_adapter(_bs)
```
titu1994, line 749: Each adapter type has its own strategy - by removing this you're hard coding the logic and side stepping the strategy. Its not necessary to do that is it?

nemo/collections/nlp/modules/common/prompt_encoder.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         self.prompt_embeddings.weight.requires_grad = False
 
         # Set fixed indicies for forward pass
-        self.register_buffer('indices', torch.LongTensor(list(range(self.total_virtual_tokens))))
+        self.register_buffer("indices", torch.LongTensor(list(range(self.total_virtual_tokens))), persistent=False)
```
titu1994, line 73: Probably a breaking change since older peft modules will have this but newer ones wont. Need to check

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
arendu: They are the list of full names of all adapter parameters which will be trained (i.e. weights will be updates). For backward compatibility reasons there are some adapter params which are not-trained but should still remain in the state_dict.

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
arendu: Calls to ParallelLinearAdapter don't explicitly pass a strategy arg so nothing breaks. I want the strategy to be explicitly set rather than defaults because its easier to catch bugs this way.

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
arendu: good point!

nemo/collections/nlp/modules/common/megatron/language_model.py
arendu: Agreed, its not necessary, but just makes the readability and maintainability easier. Its just clearer what is happening to read the code and see a residual connection rather than follow up with where a default strategy is coming from.

nemo/collections/nlp/modules/common/prompt_encoder.py
arendu: good catch! will check!

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
titu1994: Makes sense

nemo/collections/nlp/modules/common/prompt_encoder.py
arendu: ok this will indeed break older p-tuning checkpoints. However, it is a nice "feature" because older checkpoints will need to be converted anyway to a new param naming format. In that conversion step (which need to be written) I will remove the indices.

nemo/collections/nlp/modules/common/megatron/language_model.py
titu1994: I think this is mostly bypassing the entire system. I would suggest reverting this so that the strategy kicks in and to only update the adapter strategy defaults in all adapters.

arendu: oh one other limitation i think is that i can not pass additional args? for example for ptuning_adapter the forward now accepts an additional arg like `used_cached_reps`

titu1994: After discussion, seems this is never going to be needed by nlp domain, so its fine to directly call adaper

titu1994 (approved):

PR #6889, opened 2023-06-20
[TTS] Add output audio format to preprocessing

rlangman:
# What does this PR do ?

Add option in TTS audio preprocessing to convert audio to desired output format. Default to flac since it is lossless and uses much less space than wav.

Adding this to save on disk space for large datasets, and because I was hitting cryptic dataloader errors when training with mp3 files.

**Collection**: [TTS]

# Changelog 
- Add `output_format` CLI parameter.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


scripts/dataset_processing/tts/preprocess_audio.py
```diff
@@ -97,6 +97,12 @@ def get_args():
     parser.add_argument(
         "--output_sample_rate", default=0, type=int, help="If provided, rate to resample the audio to."
     )
+    parser.add_argument(
+        "--output_format",
+        default=".flac",
```
XuesongYang, line 102: Double check if `.` brings in ambiguation in CLI. Is it better to default as `flac` instead?

scripts/dataset_processing/tts/preprocess_audio.py
```diff
@@ -123,12 +129,18 @@ def _process_entry(
...
+    output_format: str,
     volume_level: float,
 ) -> Tuple[dict, float, float]:
     audio_filepath = Path(entry["audio_filepath"])
 
     audio_path, audio_path_rel = get_abs_rel_paths(input_path=audio_filepath, base_path=input_audio_dir)
+
+    if not output_format:
+        output_format = audio_path.suffix
+
```
XuesongYang, line 141: we need a safeguard to cover only supported audio formats, otherwise, raises a `ValueError`. Are the formats supported the same to Librosa.load() or torchaudio.load()?

XuesongYang (changes requested):
I could be wrong, but I did not find any system calls using either `sox` or `ffmpeg` for the conversion. Could you please elaborate it?

scripts/dataset_processing/tts/preprocess_audio.py
rlangman: I changed it so it expects it without the ".", which is also consistent with soundfile.

scripts/dataset_processing/tts/preprocess_audio.py
rlangman: I added a check. I am not sure if torchaudio and librosa support the same audio formats.

rlangman:
> I could be wrong, but I did not find any system calls using either `sox` or `ffmpeg` for the conversion. Could you please elaborate it?

The conversion is done by changing the extension after reading with librosa and writing with soundfile. I think those libraries used `libsndfile` for handling different formats.

XuesongYang (approved):

PR #6899, opened 2023-06-21
Fix transcribe_utils.py for hybrid models in partial transcribe mode

stevehuang52:
The bug is that the hybrid-ctc model has a slightly different transcribe function implementation from the original CTC model.

### DIFF
--- a/nemo/collections/asr/parts/utils/transcribe_utils.py
+++ b/nemo/collections/asr/parts/utils/transcribe_utils.py
@@ -23,7 +23,7 @@ from omegaconf import DictConfig
 from tqdm.auto import tqdm
 
 import nemo.collections.asr as nemo_asr
-from nemo.collections.asr.models import ASRModel, EncDecCTCModel, EncDecHybridRNNTCTCModel
+from nemo.collections.asr.models import ASRModel, EncDecHybridRNNTCTCModel
 from nemo.collections.asr.parts.utils import rnnt_utils
 from nemo.collections.asr.parts.utils.streaming_utils import FrameBatchASR
 from nemo.collections.common.parts.preprocessing.manifest import get_full_path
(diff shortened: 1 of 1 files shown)


erastorgueva-nv:
Thanks for the quick fix Steve, I can confirm that partial transcription using `stt_en_fastconformer_hybrid_large_pc` & `decoder_type="ctc"` now works for me - no errors thrown and `pred_text` looks good

fayejf (dismissed):
LGTM thanks!

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -421,6 +420,8 @@ def transcribe_partial_audio(
                 input_signal=test_batch[0].to(device), input_signal_length=test_batch[1].to(device)
             )
             logits, logits_len = outputs[0], outputs[1]
+            if isinstance(asr_model, EncDecHybridRNNTCTCModel) and decoder_type == "ctc":
+                logits = asr_model.ctc_decoder(encoder_output=logits)
```
erastorgueva-nv, line 424: Am I right in thinking that we will do a forward pass twice if it's a Hybrid model and decoder_type == "ctc"? That will essentially double the time to get a transcription, right? Could we avoid this?

stevehuang52: Not really. Hybrid-ctc separates encoder and decoder (the forward() function only runs encoder), so we need to run the decoder separately, while in the original ctc model the forward() has decoder included as well

erastorgueva-nv: I see, thank you for explaining, LGTM

titu1994 (approved):

fayejf (approved):

VahidooX (approved):

PR #6906, opened 2023-06-22
Add hybrid model support to transcribe_speech_parallel.py

stevehuang52:
# What does this PR do ?

Add hybrid model support to transcribe_speech_parallel.py

**Collection**: [ASR]


erastorgueva-nv (approved):
Thanks Steve! The changes look good, and I can confirm the script works: `decoder_type="ctc"` causes the model to switch to CTC decoding, and `offset` fields, if present, are also saved in the output manifests.

PR #6907, opened 2023-06-22
Reduce number of nodes produced in ConformerEncoder export

ilyasher:
This commit breaks up the output cache tensor into a list of tensors. Each layer which writes to the output cache now writes to its own tensor. The output cache tensors are then stacked into one tensor before being returned, so ConformerEncoder still has the same inputs and outputs as before.

The reason for this change is from an ONNX perspective, updating a tensor creates a new copy with the changes. So minimizing the sizes of the tensors that get updated makes the exported ONNX model more efficient.

The number of nodes produced in the ONNX export is also reduced. In the example I tested, the exported ONNX model had 11.484 nodes before this change, and only 4,892 nodes after this change.

Signed off by: Ilya Sherstyuk <isherstyuk@nvidia.com>


ilyasher:
@VahidooX could you please review? CC @borisfom

ilyasher:
Closing in favor of #6918

PR #6915, opened 2023-06-24
Fix gpt trainer test

hsiehjackson:
# What does this PR do ?

- Fix gpt trainer inference_step to compute generation results during validation and test
- Add dataset with metadata, to output in prediction file
- Add token_f1 metrics for Scrolls

**Collection**: [NLP]

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -353,74 +354,24 @@ def test_epoch_end(self, outputs):
...
-            inputs=(
-                batch['tokens'].cuda(),
-                (batch['context_lengths'] - 1).cuda(),
-            ),  # NOTE: We do -1 here to remove the space between context and response.
-            tokenizer=self.tokenizer,
-            sampling_params=sampling_params,
-            length_params=length_params,
-            check_sequence_parallel_and_checkpointing=False,  # We need to skip these checks since we'll manually enbale and disable checkpointing between training and validation.
-        )
+        batch = next(dataloader_iter)
```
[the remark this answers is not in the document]
hsiehjackson: I think this part is only used for inference (without gradient accumulation) not training. The code is referenced from [t5 inference](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron_finetune_model.py#L335-L346).

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -163,40 +163,11 @@ def main(cfg) -> None:
...
-                                f.write(json.dumps(d, sort_keys=True, indent=2) + '\n')
-                    else:
-                        for s in batch_sentences:
-                            d = {'sentence': s}
-                            f.write(json.dumps(d) + '\n')
-            print("predictions saved to {}".format(cfg.inference.outfile_path))
-        else:
-            print(response)
-    print("***************************")
+
```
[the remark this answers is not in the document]
ericharper: This is interesting to me. They are different APis from PTL: https://lightning.ai/docs/pytorch/stable/common/trainer.html#test vs https://lightning.ai/docs/pytorch/stable/common/trainer.html#predict

We should come up with a uniform design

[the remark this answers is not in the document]
arendu: @hsiehjackson have you checked if predict and test give same results?

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -163,40 +163,11 @@ def main(cfg) -> None:
...
+    if hasattr(cfg.model.data, 'test_ds'):
+        trainer.test(model)
```
arendu, line 170: previously the response object was used to produce the final output jsonl file. where does the output get saved to with this refac?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         seed: int = 1234,
         context_key: str = "text",
         label_key: str = "answer",
+        reference_key: str = None,
```
arendu, line 41: why do you need this extra field? why not use label_key are the reference for inferencing?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -292,6 +293,7 @@ def _determine_log_key(self, data_config, dataloader_idx, metric_name, mode):
 
     def fwd_bwd_step(self, dataloader_iter, batch_idx, forward_only):
         batch = next(dataloader_iter)
+        batch = {k: v for k, v in batch.items() if isinstance(v, torch.Tensor)}
```
arendu, line 296: can you comment why this is needed? cc @aklife97

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
hsiehjackson: We will call [write_predictions_to_file](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L560-L564) to write prediction file.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: Currently, our dataset json file only has column for `context_key` and `label_key` both are in string format. However, users may want to pass other columns to the prediction files, e,g. IDs, other labels, etc. In this scenario, I use `reference_key` to pass the column with multiple labels (some tasks has multiple reference answers than just one). I am not sure do we have another good way to pass columns to prediction files since we may face order issue due to using distributed sampler.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
hsiehjackson: I pass `reference_texts` in the batch which is not a torch.Tensor and will get errors when doing `data_iter = get_iterator_k_split(batch, get_num_microbatches())`

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
arendu: do you mean this is just for including some metadata for each test input? i.e. the model does not use this information in any way?

Jenkinsfile
arendu, line 1: @hsiehjackson could you add merge test https://github.com/NVIDIA/NeMo/pull/6931 into this PR? This way we can actually verify if the peft inference changes are working.

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
arendu: ok i recall having trouble with this for peft_eval script. did you verify that it was generating correctly?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
arendu: for multiple possible correct labels for evaluation a better way is to allow a list of labels instead of a single string.

hsiehjackson: I currently use `reference_key` to pass multiple answers and calculate metrics between predictions and answers. 
There are two issues related to these changes.
1. We want to pass some metadata for each test input.
2. We want to calculate metrics with multiple answers.
For 1., I am not sure what is the best way for the implementation. 
For 2., I can change to directly use `label_key` as the answer and calculate metrics, which may make more sense. Then user can calculate their metrics if we have implementation 1.

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
hsiehjackson: I use `predict_step()` to generate outputs, so it should be the same logic as previous implementation.

Jenkinsfile
hsiehjackson: Sure!

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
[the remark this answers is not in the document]
hsiehjackson: Yes. They give the same results. The difference is test can used data parallel.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: I remove this field and simply copy all columns except context_key and label_key to the prediction file.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
hsiehjackson: In the batch, we may have other metadata column from dataset. So I simply select those are tensors to pass to `get_iterator_k_split()`. If I don't do so, we may see errors for element not be tensor.

examples/nlp/language_modeling/tuning/megatron_gpt_peft_tuning.py
```diff
@@ -85,7 +85,21 @@ def _modify_config(gpt_cfg, cfg, add_cfg_to_tree=False):
         gpt_cfg.activations_checkpoint_granularity = cfg.model.get("activations_checkpoint_granularity", None)
         gpt_cfg.activations_checkpoint_num_layers = cfg.model.get("activations_checkpoint_num_layers", None)
         gpt_cfg.activations_checkpoint_method = cfg.model.get("activations_checkpoint_method", None)
+        gpt_cfg.activations_checkpoint_layers_per_pipeline = cfg.model.get(
```
yzhang123, line 88: what does this do and why is this needed?

hsiehjackson: We call `_reset_activation_checkpointing_args()` and `_restore_activation_checkpointing_args()` in validation and test step start ([code](https://github.com/NVIDIA/NeMo/blob/fix-gpt-trainer-test/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L669)) which we need the config of `activations_checkpoint_layers_per_pipeline`.

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -125,8 +125,20 @@ def main(cfg) -> None:
         # update the model config of the trained model with params we want to set at inference time.
         peft_model_cfg.precision = cfg.trainer.precision
         peft_model_cfg.data.test_ds = cfg.model.data.test_ds
```
yzhang123, line 127: could you pls add , since it otherwise uses the global batch size of pretrained model instead of configuration

        peft_model_cfg.global_batch_size = cfg.model.global_batch_size
        peft_model_cfg.micro_batch_size = cfg.model.micro_batch_size

hsiehjackson: don't we use `model.data.test_ds.global_batch_size` to set this parameter?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -38,6 +38,7 @@ def __init__(
         seed: int = 1234,
         context_key: str = "text",
         label_key: str = "answer",
+        query_key: str = None,
```
ekmb, line 41: missing docstring for this key
could you make it Optional?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -134,91 +137,152 @@ def __getitem__(self, idx):
                 idx = idx.item()
 
         assert idx < len(self.indexed_dataset)
+        # idx may < 0 because we pad_samples_to_global_batch_size, e.g. id = -1
+        if idx < 0:
+            idx = len(self) + idx
         example = self.indexed_dataset[idx]
         return self._process_example(example)
 
+    def _calculate_total_ids(self, example):
```
ekmb, line 146: doctring

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -134,91 +137,152 @@ def __getitem__(self, idx):
...
+
+        # Only training need to consider eos token
+        if self.add_eos and self.tokens_to_generate == 0:
             input_ids = input_ids + [self.tokenizer.eos_id]
 
-        if len(input_ids) < self.min_seq_length or len(input_ids) > self.max_seq_length:
-            input_ids = input_ids[: self.max_seq_length]
+        assert len(input_ids) <= self.max_seq_length
 
+        # store metadata in dataset
```
ekmb, line 277: could add in the comment why this is needed

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -134,91 +137,152 @@ def __getitem__(self, idx):
...
         input_ids = input_ids + answer_ids
 
         if self.add_bos:
+            context_ids = [self.tokenizer.bos_id] + context_ids
             input_ids = [self.tokenizer.bos_id] + input_ids
             answer_start_idx += 1
-        if self.add_eos:
+
+        # Only training need to consider eos token
+        if self.add_eos and self.tokens_to_generate == 0:
```
ekmb, line 272: don't we want to stop generation during inference if eos is omitted?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -134,91 +137,152 @@ def __getitem__(self, idx):
...
-
-        if self.separate_prompt_and_response_with_newline and self.prompt_template is None:
-            text = context + '\n' + output
-        elif not self.separate_prompt_and_response_with_newline and self.prompt_template is None:
-            text = context + ' ' + output
+            context = self.prompt_template.replace(context_string, context).replace(label_string, '').strip(' ')
+            text = self.prompt_template.replace(context_string, original_context).replace(label_string, label)
+            if self.query_key is not None:
+                query_string = f'{{{self.query_key}}}'
+                assert query_string in self.prompt_template
```
ekmb, line 232: could you add error message here?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
ekmb: this function duplicates multiple lines from process_example(), why is it needed?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -265,13 +329,15 @@ def collate_fn(self, batch):
         contexts = [item['context_ids'] for item in batch]
         context_lengths = torch.LongTensor([item['context_length'] for item in batch])
         loss_mask = [self._build_loss_mask(item)[1:] for item in batch]
+        metadata = [item['metadata'] for item in batch]
 
-        max_length = max([len(x) for x in input_ids]) + self.tokens_to_generate
+        max_length = max(max([len(x) for x in input_ids]), max([len(x) for x in contexts]) + self.tokens_to_generate)
```
ekmb, line 334: when does context could be longer input_ids?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
ekmb: could you comment in the code as well

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
```diff
@@ -122,6 +125,7 @@ model:
       truncation_field: "context" # Options: ['context', 'answer']
       index_mapping_dir: null # Path to a directory to write index mapping files.
       prompt_template: ${model.data.train_ds.prompt_template} # fstring to use for assistant prompt. Example: "Q: {input}\nA: {output}"
+      tokens_to_generate: ???
 
       metric:
         name: "loss" # Name of the evaluation metric to use. Options: ['exact_string_match', 'loss']
```
ekmb, line 131: could add comment about supported metrics and how to use them?

ekmb (changes requested):

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: This is used to calculate how long we'll get for total tokens, so we can know how long we can truncate contexts.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: But during generation, we don't need to add EOS right? EOS is generated by model

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: It is `contexts` + `tokens_to_generate` compared to `input_ids` (`contexts` + `answers`).

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: I try to refactor and only remain two functions `_process_prompt()` and `_process_truncation()`

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         seed: int = 1234,
         context_key: str = "text",
         label_key: str = "answer",
+        query_key: Optional[str] = None,
```
arendu, line 43: could you add query_key to the doc string below

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
arendu: Whats the difference between the old reference_key and the new query_key that you've added?

hsiehjackson: `reference_key` is used to pass some key-value metadata from dataset jsonl file to prediction jsonl file. I remove this by just passing all key-value pairs except context_key and label_key.

`query_key` is used to have the flexibility to construct prompts. Previously, we only have `context_key` and the query part (questions) is at the front or the end of the context. It is hard to truncate the contexts. Therefore, we add one key for query, so user can construct the prompt based on their preference. 

Here is an example from Qasper in Scrolls:
context_key: `Introduction\\nCategories such as animal or furniture are fundamental cognitive building blocks allowing humans to efficiently represent ...`
query_key: `do language share categories?`
label_key: `Yes`
prompt: `{context} <extra_id_1>{query}<extra_id_2>\n{label}`

nemo/collections/common/metrics/classification_accuracy.py
```diff
@@ -205,3 +208,51 @@ def update(self, pred: str, target: str):
...
+        num_same = sum(common.values())
+        if num_same == 0:
+            return 0.0
+        precision = 1.0 * num_same / len(prediction_tokens)
+        recall = 1.0 * num_same / len(ground_truth_tokens)
+        f1 = (2 * precision * recall) / (precision + recall)
+        return f1
+
+    def normalize(self, s):
+        """Lower text and remove punctuation, articles and extra whitespace."""
```
yzhang123, line 243: could you add this is oriented on squad ?

nemo/collections/common/metrics/classification_accuracy.py
```diff
@@ -205,3 +208,51 @@ def update(self, pred: str, target: str):
...
+
+
+class TokenF1Score(Metric):
+    def __init__(self, dist_sync_on_step=False, *args, **kwargs):
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+
+        self.add_state("correct", default=torch.tensor(0.0), dist_reduce_fx="sum")
+        self.add_state("total", default=torch.tensor(0), dist_reduce_fx="sum")
+
+    def update(self, pred: str, target):
```
yzhang123, line 220: add type for target?

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: Sure! `query_key` is needed if user want to separate question from the context and construct their customized prompt.

yzhang123 (changes requested):
overall LGTM, great work!

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
arendu: I think its best to leave the context field as is. We can merge the "query" into the context in a preprocessing step.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
arendu: please comment in the code as @ekmb mentioned.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: When we have a extreme long context and we need to truncate them, we may truncate the query part. How can we put the query in the end and we still truncate the context part? I think most dataset can have the format with context, query, label.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
hsiehjackson: Sure!

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: Dataset part is moved to this PR https://github.com/NVIDIA/NeMo/pull/7147

examples/nlp/language_modeling/tuning/megatron_gpt_peft_eval.py
```diff
@@ -127,11 +127,25 @@ def main(cfg) -> None:
...
         peft_model_cfg.activations_checkpoint_granularity = None
         peft_model_cfg.activations_checkpoint_method = None
+        peft_model_cfg.activations_checkpoint_layers_per_pipeline = None
         if peft_model_cfg.get("use_flash_attention", False):
             peft_model_cfg.use_flash_attention = cfg.model.use_flash_attention
+        if peft_model_cfg.get("apply_query_key_layer_scaling", False):
```
arendu, line 133: What does "apply_query_key_layer_scaling" do?

examples/nlp/language_modeling/tuning/megatron_gpt_sft.py
```diff
@@ -70,6 +75,18 @@ def _modify_config(gpt_cfg, cfg, add_cfg_to_tree=False):
...
+        if (
+            gpt_cfg.get('position_embedding_type', None) in [None, 'learned_absolute']
+            and max_seq_length > gpt_cfg.max_position_embeddings
+        ):
+            logging.warning(
+                f"Set dataset max_seq_length to max_position_embeddings {gpt_cfg.max_position_embeddings} if using learned_absolute position embedding"
+            )
+            gpt_cfg.data.train_ds.max_seq_length = gpt_cfg.max_position_embeddings
+            gpt_cfg.data.validation_ds.max_seq_length = gpt_cfg.max_position_embeddings
+
```
arendu, line 89: can this logic be moved inside `MegatronGPTSFTModel._build_dataset`?

examples/nlp/language_modeling/tuning/megatron_gpt_peft_tuning.py
```diff
@@ -100,6 +103,18 @@ def _modify_config(gpt_cfg, cfg, add_cfg_to_tree=False):
...
+        if (
+            gpt_cfg.get('position_embedding_type', None) in [None, 'learned_absolute']
+            and max_seq_length > gpt_cfg.max_position_embeddings
+        ):
+            logging.warning(
+                f"Set dataset max_seq_length to max_position_embeddings {gpt_cfg.max_position_embeddings} if using learned_absolute position embedding"
+            )
+            gpt_cfg.data.train_ds.max_seq_length = gpt_cfg.max_position_embeddings
+            gpt_cfg.data.validation_ds.max_seq_length = gpt_cfg.max_position_embeddings
+
```
arendu, line 117: can this logic be moved inside MegatronGPTSFTModel._build_dataset?

arendu (changes requested):

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -141,99 +142,120 @@ def __getitem__(self, idx):
...
 
         assert idx < len(self.indexed_dataset)
+        # idx may < 0 because we pad_samples_to_global_batch_size, e.g. id = -1
+        if idx < 0:
+            idx = len(self) + idx
         example = self.indexed_dataset[idx]
         return self._process_example(example)
 
-    def _process_example(self, example):
+    def _process_prompt(self, context: str, label: str, query: str):
```
arendu, line 151: were you going to revert this for this PR?

arendu (changes requested):

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: I remove the query part. However, I still leave some refactoring codes to split out `_process_truncation()` and `_process_prompt()`.

arendu (approved):
LGTM with minor comments. I would also seek input from folks using SFT code path heavily.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -193,7 +196,8 @@ def _process_example(self, example):
             total_ids += 1
         if self.add_sep:
             total_ids += 1
-        if self.add_eos:
+        # Only training need to consider eos token
+        if self.add_eos and self.tokens_to_generate == 0:
```
arendu, line 200: are you using `self.tokens_to_generate == 0` to decide if we are in training phase or eval phase?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -452,82 +432,94 @@ def inference_epoch_end(self, outputs, mode, data_cfg):
...
             averaged_loss.append(loss)
-
-            # Skip the rest of this loop if the user wants to monitor the loss only.
-            if self.val_metric is None:
-                continue
-            # Determine the key used to log the eval metric based on the user provided name of the dataset or the dataloader index.
-            metric_log_key = self._determine_log_key(data_cfg, dataloader_idx, metric_name, mode)
-            metric_object = (
-                self.val_metric[dataloader_idx] if mode == 'validation' else self.test_metric[dataloader_idx]
+            # Gather the outputs object from all data parallel ranks since we are using the DistributedSampler which splits data across DDP ranks.
```
arendu, line 441: is it still possible to skip all this is one just wants to use val loss as the monitor metric?

hsiehjackson: We have this condition to bypass if we use metric.name='loss'
https://github.com/NVIDIA/NeMo/blob/fix-gpt-trainer-test/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L479

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: Yes. Do you have a safer idea?

ekmb (approved):

arendu (approved):

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
```diff
@@ -217,34 +221,41 @@ def _process_example(self, example):
...
             input_ids = input_ids + [self.tokenizer.eos_id]
 
-        if len(input_ids) < self.min_seq_length or len(input_ids) > self.max_seq_length:
+        if len(input_ids) > self.max_seq_length:
+            logging.warning(f'Input ids length {len(input_ids)} exceed max sequence length {self.max_seq_length}')
             input_ids = input_ids[: self.max_seq_length]
 
+        # store metadata in dataset, in case user may have keys required in the prediction json files
+        metadata = {k: v for k, v in example.items() if k not in [self.context_key, self.label_key]}
```
Kipok, line 250: Not sure if I'm doing something wrong, but I just tried this branch today and getting errors. I think the reason is that you only changed the `gpt_sft_dataset.py`, but I am using the `gpt_sft_chat_dataset.py`. Still the validation/inference part is shared for them and I'm getting errors, e.g. that metadata key is not available.

Am I doing something incorrectly or are the "chat"-based datasets broken in this branch @hsiehjackson?

hsiehjackson: Did you see the error because of this? Let me add the check for metadata. https://github.com/NVIDIA/NeMo/blob/fix-gpt-trainer-test/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L407

yzhang123 (dismissed):

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
Kipok: that's just the first error - I fixed it in my code, but then everything breaks downstream as well. So I think more changes are needed here

Kipok: just confirmed that if I switch to the non-chat format, everything works well. I think you just need to make sure the dataset is prepared in the same way there including metadata, context, length, etc.

And as a side note - seems like we don't have any tests for chat-datasets? Something we should look into in the future, I guess

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -779,27 +809,9 @@ def setup_eval_dataloader(self, datasets, data_cfg):
...
-            base_module = self.model.module
-        else:
-            base_module = self.model
-        base_module.language_model.encoder.activations_checkpoint_granularity = self.original_checkpointing_granularity
-        base_module.language_model.encoder.activations_checkpoint_method = self.original_checkpointing_method
-        base_module.language_model.encoder.activations_checkpoint_num_layers = self.original_checkpointing_num_layers
-
     def on_validation_epoch_start(self):
         self._reset_activation_checkpointing_args()
+        self._reset_sequence_parallelism_args()
```
yidong72, line 814: where is the restore activation checkpointing and sequence parallel ?

yidong72 (commented):
Looks good overall, please don't change the default configuration behavior as the old code depends on it.

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
```diff
@@ -109,15 +110,15 @@ model:
       names: null # Names of the corresponding datasets used to log metrics.
       global_batch_size: ${model.global_batch_size}
       micro_batch_size: ${model.micro_batch_size}
-      shuffle: True
+      shuffle: False
```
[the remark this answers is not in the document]
hsiehjackson: Why we do shuffling for validation dataset?

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
```diff
@@ -109,15 +110,15 @@ model:
...
       num_workers: 4
       memmap_workers: ${model.data.train_ds.memmap_workers}
       pin_memory: True
       max_seq_length: 2048
       min_seq_length: 1
-      drop_last: True
-      context_key: 'input'
-      label_key: 'output'
+      drop_last: False
```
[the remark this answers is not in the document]
hsiehjackson: Why we drop last if we evaluate with the validation dataset?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
hsiehjackson: These two functions are inherited from `MegatronGPTModel`

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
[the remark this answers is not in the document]
yidong72: okay. didn't notice it is validation dataset.  This is fine.

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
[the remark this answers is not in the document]
yidong72: I have one use case that I have to drop the last because I am preparing a dataset that computes contrastive loss.  I need to make sure all the batch has the same batch size.

nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py
hsiehjackson: `GPTSFTChatDataset` is fixed. Wait for @Kipok to try.

examples/nlp/language_modeling/tuning/conf/megatron_gpt_sft.yaml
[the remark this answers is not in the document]
hsiehjackson: If we set `drop_last=False`, we can use `pad_samples_to_global_batch_size=True` right? Or this doesn't fulfill your settings? https://github.com/NVIDIA/NeMo/blob/fix-gpt-trainer-test/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L788 
If we set default `drop_last=True`, then we may drop some samples for evaluation which may show incorrect results for comparison.

ekmb (approved):

PR #6911, opened 2023-06-23
Fix FP8 scales conversion issue from non-nemo checkpoint to nemo checkpoint

sudhakarsingh27:
# What does this PR do ?

Converting a non-nemo checkpoint to NeMo with Transformer Engine modules results in truncating of scales in `_extra_state`. This WAR temporarily fixes the issue by manually toggling the `fp8` attribute in TE modules. 

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #6929, opened 2023-06-27
Fix FP8 scales conversion issue from non-nemo checkpoint to nemo checkpoint (try 2) (failed try 1: #6911)

sudhakarsingh27:
# What does this PR do ?

TLDR; 
Converting a non-nemo checkpoint to NeMo with Transformer Engine modules results in truncating of scales in _extra_state. This WAR temporarily fixes the issue by manually toggling the fp8 attribute in Transformer Engine modules.

More context:
1. When exporting TE module data for checkpointing (fp8 scale info, etc.), `fp8` attribute needs to be set (see here: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/module/base.py#L344) but this attribute is only set during the forward (and when using `fp8_autocast` context manager). If the `fp8` attribute isn't set, the relevant scale information isn't included in the pickled `_extra_state` tensor (and this is the "truncation of scales" issue referred in the TLDR, basically an empty tensor is pickled which makes it seem like the scale information or `_extra_state` tensor is truncated). (Note that I tried adding the `fp8_autocast` context manager around the `model.save_to()` call and it didn't work as expected)
2. Until, there's a cleaner way to enable the `fp8` atribute for the Transformer Engine modules, this WAR allows externally enabling it before the `model.save_to()` call in the checkpoint converter script.

**Collection**: NLP

# Changelog 
- Toggle the `fp8` attribute in a module if it exists (if it exists, its probably a TE module)

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
+++ b/examples/nlp/language_modeling/megatron_ckpt_to_nemo.py
@@ -30,6 +30,7 @@ from argparse import ArgumentParser
 import torch
 from megatron.core import parallel_state
 from omegaconf import open_dict
+from pytorch_lightning.core.saving import load_hparams_from_yaml
 from pytorch_lightning.plugins.environments import TorchElasticEnvironment
 from pytorch_lightning.trainer.trainer import Trainer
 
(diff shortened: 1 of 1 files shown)


ksivaman:
@sudhakarsingh27 Could you explain what toggling that attribute does and what is the root cause of the truncation of the scales?

sudhakarsingh27:
@ksivaman, I've added more information about the issue above. Thanks for pointing it out!

ksivaman:
The explanation in the post description makes sense, however I don't think this is the best solution since this would break this script for TE/non-fp8 checkpoint conversion.

sudhakarsingh27:
I see. I think to incorporate that, we might need to unpack the accompanying `*.yaml` file to check if the `fp8` attribute is set. But I do agree that this WAR is not fool-proof solution and we definitely need a cleaner solution. Do you have any ideas?

ksivaman:
For future generate checkpoints, I think the change belongs in TE, since the root cause here is that TE uses `fp8` and `calibration` attributes that are only set during forward, which doesn't work for the case of loading and saving checkpoints without actually training (a case used in converters that we didn't account for in TE). Meanwhile, a better solution here would be to inspect the checkpoint and see if it's an fp8 one, and only then set the attribute to `True` as a WAR.

sudhakarsingh27:
I agree. I've created an issue on TE repo (https://github.com/NVIDIA/TransformerEngine/issues/304) to track this long term. Meanwhile, let me add the code to check whether the checkpoint supports `fp8` or not to the above WAR.

ksivaman:
I've fixed the underlying issue in [this](https://github.com/NVIDIA/TransformerEngine/pull/323) TE PR. However, for older checkpoints we'll still have to implement the suggested solution above.

sudhakarsingh27:
@ksivaman, thanks for fixing the underlying issue! I've updated the WAR with your suggestion. Pls let me know if this looks okay.

PR #6937, opened 2023-06-28
Fix confidence ensembles RNNT logprobs selection logic for exclude_blank scenario

KunalDhawan:
# What does this PR do ?

- Fixes the algorithm for filtering blank frames while building confidence ensembles
- Adds checks for confidence value calculation to handle the edge case of all blank frames

ASR 

# Changelog 
- changes to get_filtered_logprobs() and compute_confidence() functions in nemo/collections/asr/models/confidence_ensembles.py

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

### DIFF
--- a/nemo/collections/asr/models/confidence_ensemble.py
+++ b/nemo/collections/asr/models/confidence_ensemble.py
@@ -102,6 +102,8 @@ def get_filtered_logprobs(hypothesis: Hypothesis, exclude_blank: bool) -> torch.
         if exclude_blank:  # filtering blanks
             labels = logprobs.argmax(dim=-1)
             filtered_logprobs = logprobs[labels != logprobs.shape[1] - 1]
+            if filtered_logprobs.shape[0] == 0:  # for the edge-case of all blanks
+                filtered_logprobs = logprobs[:1]
         else:
             filtered_logprobs = logprobs
     return filtered_logprobs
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/models/confidence_ensemble.py
```diff
@@ -135,7 +136,10 @@ def compute_confidence(hypothesis: Hypothesis, confidence_cfg: ConfidenceConfig)
         alpha = confidence_cfg.method_cfg.temperature
     conf_func = get_confidence_measure_bank()[conf_type]
 
-    conf_value = aggr_func(conf_func(filtered_logprobs, v=vocab_size, t=alpha)).cpu().item()
+    if filtered_logprobs.shape[0] == 0:  # for the edge-case of all blanks, no frames
```
Kipok, line 139: I'm not sure that returning 0 is the best choice here. The model might be very confident in predicting only blanks when there is a silence and we probably shouldn't ignore that. So maybe safer to use the same approach as for rnnt and return a confidence of a single blank token. Something like this in the `get_filtered_logprobs` function
```
            if filtered_logprobs.shape[0] == 0:  # for the edge-case of all blanks
                filtered_logprobs = logprobs[:1]
```

Kipok (changes requested):
Just a small issue to fix for consistency with rnnt logic.

By the way, were you able to figure out why rnnt probabilities are not correctly normalized?

nemo/collections/asr/models/confidence_ensemble.py
KunalDhawan: Good point! Fixed this issue in the latest commit by following a similar approach for CTC as done for RNNT in get_filtered_logprobs(). 
Still investigating the rnnt normalization issue of why log_normalize isn't being propagated as true

Kipok (approved):
Thanks!

PR #6957, opened 2023-07-01
Stop at the stop token

yidong72:
Detects the stop token and stop the generation once the stop token is detected.


aklife97 (approved):
LGTM!

PR #6989, opened 2023-07-07
rnnt_greedy_decoding.py: typos? auto-repressively -> auto-regressively

vadimkantorov:
Fixes typos (typos, right?) in the code comments "auto-repressively" -> "auto-regressively"


titu1994 (approved):
Indeed very old typo. Thanks !

PR #7086, opened 2023-07-20
[Fix] load_state_dict in nlp_model.py

stevehuang52:
# What does this PR do ?

Fix error in loading checkpoints. The pytorch [load_state_dict()](https://pytorch.org/docs/stable/generated/torch.nn.Module.html?highlight=load_state_dict#torch.nn.Module.load_state_dict) function has return variable that was not returned by this function in this file. 

**Collection**: [NLP]


fayejf (approved):
LGTM. better get approval from NLP people? Maybe @ekmb could help?

PR #7091, opened 2023-07-21
Fix tensor_parallel function calls to megatron-lm

stevehuang52:
# What does this PR do ?

Megatron-LM has changed its function APIs in tensor_parallel, where a lot of params (e.g., `use_cpu_initialization`) are moved into a config object (e.g., `megatron.core.model_parallel_config.ModelParallelConfig`), for example [here](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/tensor_parallel/layers.py#L148)

This PR fixes such errors when trying to use those functions with the latest Megatron-LM codebase, but there may be some other missed functions. 

**Collection**: [NLP]

### DIFF
--- a/nemo/collections/nlp/modules/common/megatron/attention.py
+++ b/nemo/collections/nlp/modules/common/megatron/attention.py
@@ -55,7 +55,7 @@ except (ImportError, ModuleNotFoundError):
 
 try:
     from megatron.core import parallel_state, tensor_parallel
-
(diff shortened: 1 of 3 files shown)


arendu:
I believe @ericharper is working on this so that nemo will be compatible with new megatron-core. This change will make things break for all our current containers.

stevehuang52:
I think some other people are working on adapting nemo to the latest megatron-lm, so I'll close this PR

PR #7112, opened 2023-07-26
Include the scripts for preprocessing OAST and unit tests for chat sft datasets

yidong72:
In this PR.

1. include scripts for preprocess OAST datasets.
2. annotate the datasets using reward model.
3. clean up the data for training.
4. include the unit tests to make sure the Chat SFT datasets produced expected masks. 
5. add support in the UI to run attribute condition generations.
6. add HF dataset as JSON dataset backend.


tests/collections/nlp/test_chat_sft_dataset.py
```diff
@@ -0,0 +1,317 @@
...
+    def setup_class(cls):
+        pass
+
+    @pytest.mark.unit
+    def test_43B_tokenizer_mask_user(self):
+        random.seed(5)
+        temp_file = '/tmp/test_file.jsonl'
+        turn_num = 5
+        records = 5
+        try:
```
aklife97, line 85: why are we doing this inside a try block?

examples/nlp/language_modeling/conf/megatron_gpt_inference.yaml
```diff
@@ -38,6 +38,55 @@ web_port: 9889 # the port number of the web server
 chat: False # use the chat interface
 chatbot_config:
   value: False   # whether to inject the value attributes
+  attributes:
```
aklife97, line 41: is it possible to move these attributes to some other file?
don't want to end up bloating the main gpt_inference config file by adding multiple value attributes (esp if we end up adding more)

aklife97 (approved):
LGTM! everything looks good to me, just have a couple of questions

examples/nlp/language_modeling/conf/megatron_gpt_inference.yaml
yidong72: It is meant to be an example and not a full list. I can remove a few items. The user can always use a customized yaml file.

tests/collections/nlp/test_chat_sft_dataset.py
yidong72: so we can make sure to remove the file in the `finally`

examples/nlp/language_modeling/conf/megatron_gpt_inference.yaml
aklife97: okay, outside this PR, IMO we should handle the chatbot configs in a new file.... bloating this one up for chatbot does not feel good...

PR #7123, opened 2023-07-28
Refactoring to make MegatronGPTModel dataset building more flexible

odelalleau:
# What does this PR do ?

Refactoring to make MegatronGPTModel dataset building more flexible.

In particular this allows subclasses to use a different `build_datasets_strategy` without having to copy / paste the full `build_train_valid_test_datasets()` method.

**Collection**: `nlp`

# Changelog 
- Add a new attribute `build_datasets_strategy` to the constructor of `MegatronGPTModel` to make it easier to override the dataset building logic

# Usage

```python
from my_module import my_build_datasets_strategy

class MyClass(MegatronGPTModel):

    def __init__(self):
        super().__init__(build_datasets_strategy=my_build_datasets_strategy)
```

# Discussion

There are at least two potential controversial design choices, happy to discuss them here (marking PR as draft in the meantime):
1. Naming: the most "natural" name would be `build_train_valid_test_datasets_strategy`, but I felt it was starting to get annoyingly long, which is why I picked a shorter variant
2. Implementation: alternatively, we could (a) avoid changing the signature of `__init__()`, letting subclasses override `self._build_datasets_strategy` as they see fit (I didn't go this route because I tend to prefer customization options to be clearly exposed through the API), or (b) make it part of the config, using Hydra's instantiate feature to obtain the strategy (although I actually like this option better, I wanted to discuss it here before making changes to the config). [edit: see [comment below](https://github.com/NVIDIA/NeMo/pull/7123#issuecomment-1665501843) for a follow-up discussion]

I'll look into adding a test once the implementation is approved.

**Pre checks**:
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -189,7 +189,7 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
     Megatron GPT pretraining
     """
 
-    def __init__(self, cfg: DictConfig, trainer: Trainer):
+    def __init__(self, cfg: DictConfig, trainer: Trainer, build_datasets_strategy=build_train_valid_test_datasets):
         if not HAVE_APEX:
             raise ImportError(
                 "Apex was not found. Please see the NeMo README for installation instructions: https://github.com/NVIDIA/NeMo#megatron-gpt."
@@ -202,6 +202,8 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
         # this prevents base constructor from initializing tokenizer
         self.tokenizer = None
         super().__init__(cfg, trainer=trainer, no_lm_init=True)
+
+        self._build_datasets_strategy = build_train_valid_test_datasets
 
         self._validate_trainer()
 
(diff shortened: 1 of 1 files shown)


odelalleau:
@ericharper I could use your feedback on the implementation!

odelalleau:
Un-drafting so as to hopefully get the discussion going.

There's also another option I didn't mention above, which is to (c) have a `get_build_train_valid_test_datasets_func()` method that could be overridden in subclasses. That would be favoring inheritance over composition -- I personally tend to prefer the other way around but I'm happy to go with whatever approach is preferred by others.

odelalleau:
Gave it some more thoughts, here is a recap of the various options I have in mind:

* This current commit a4ae5e482bf17569424f11edd53ba027778dbfd1
* (a) Same as above but do not change the signature of `__init__()`
* (b) Pass this option through the config instead of adding an extra argument to `__init__()`
* (c) Add a new method to fetch the dataset building function instead of providing it in `__init__()`

I'm actually not a fan of my current commit because someone might create a model with a custom dataset building function, then forget to set it again when re-loading it. In a way option (a) is safer because it really encourages you to create your own subclass to override the default behavior.

As much as I like (b), I feel like it's not well aligned with NeMo's current usage of Hydra. I think it would work best if there were structured configs associated to each model class, but that's not currently the case.

Option (c) might actually be the best fit to current NeMo's design, which is heavily based on inheritance. It might be a bit confusing though to have two methods `build_train_valid_test_datasets()` and `get_build_train_valid_test_datasets_func()`.

Happy to hear more thoughts from NeMo devs!

PR #7141, opened 2023-08-01
Fix default context size

nithinraok:
# What does this PR do ?

There is a bug when performing RNNT/CTC decoding with hybrid models. 

By default `att_context_size` should be [-1,-1] which is what set as default, but if None is passed through cfg, it replaces the current value, which shouldn't be the case. 

**Collection**: ASR

# Changelog 
- change value only if  `cfg.att_context_size` is not None

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


titu1994 (approved):
Thanks !

PR #7150, opened 2023-08-02
fix evaluator.py for various exceptions by ast

stevehuang52:
the python `ast` package throws both `SyntaxError` and `ValueError` during evaluating an invalid dict object, so exception handling needs to catch all possible kinds of exceptions.


titu1994 (approved):
Thanks !

PR #7154, opened 2023-08-02
[TTS] Remove nested TTS configs

rlangman:
# What does this PR do ?

In a separate PR (https://github.com/NVIDIA/NeMo/pull/6852) it was brought up that TTS is using nested hydra config, which is against NeMo 1 config file per model policy.

This PR removes nested config from our experimental FastPitch and HiFiGan in favor of a flat structure with duplicated code between 22kHz and 44kHz versions.

Our older non-experimental configs are left unchanged for backwards compatibility.

**Collection**: [TTS]

# Changelog 
- Remove nested configuration from FastPitch and HiFiGan configs using new data loader
- Add separate 22kHz and 44kHz examples for each model.
- Moved hifigan_data.yaml config to separate "hifigan_dataset" folder (not sure if that is a good name) now that it does not need to reuse the generator config files. 

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


redoctopus (commented):
Looks fine to me. @subhankar-ghosh or @XuesongYang -- is there anything you guys want to check before this is approved and merged?

examples/tts/conf/fastpitch/fastpitch_22050.yaml
```diff
@@ -0,0 +1,286 @@
...
+hop_length: 256
+
+mel_feature:
+  _target_: nemo.collections.tts.parts.preprocessing.features.MelSpectrogramFeaturizer
+  sample_rate: ${sample_rate}
+  win_length: ${win_length}
+  hop_length: ${hop_length}
+  mel_dim: 80
+  lowfreq: 0
+  highfreq: 8000
```
XuesongYang, line 39: It was discussed before, and I remembered we decided to go with `null` rather than 8000. Please correct me if I am wrong.

XuesongYang, line 39: It was discussed before, and I remembered we decided to go with `null` for 44100kHz config, while keep `8000` unchanged for 22050kHz. Is it a good time to switch to `null` for 22050kHz config?

examples/tts/conf/feature/feature_22050.yaml
```diff
@@ -4,25 +4,25 @@ hop_length: 256
...
   _target_: nemo.collections.tts.parts.preprocessing.features.MelSpectrogramFeaturizer
-  sample_rate: ${..sample_rate}
-  win_length: ${..win_length}
-  hop_length: ${..hop_length}
+  sample_rate: ${sample_rate}
+  win_length: ${win_length}
+  hop_length: ${hop_length}
   mel_dim: 80
   lowfreq: 0
   highfreq: 8000
```
XuesongYang, line 12: same comment as the above.

examples/tts/conf/hifigan_dataset/hifigan_22050.yaml
```diff
@@ -19,6 +14,19 @@ log_ds_meta: ???
 
 log_dir: ???
 
+mel_dim: 80
+lowfreq: 0
+highfreq: 8000
```
XuesongYang, line 19: same comment as the above.

examples/tts/conf/hifigan_dataset/hifigan_22050.yaml
```diff
@@ -19,6 +14,19 @@ log_ds_meta: ???
...
+
+# Change these values depending on your sampling rate.
+sample_rate: 22050
+win_length: 1024
+hop_length: 256
+upsample_rates: [8, 8, 2, 2]
+train_n_samples: 8192
+val_min_duration: 3.0
```
XuesongYang, line 27: nit: could you pls add a comment showing the unit is "seconds"?

XuesongYang (commented):
LGTM. Minor comments.

examples/tts/conf/fastpitch/fastpitch_22050.yaml
rlangman: If we did want to switch it to null, now might be the last good opportunity. Though it may not matter since most experimentation now focuses on 44.1kHz and we will likely add support for training on acoustic tokens/codes soon.

The interesting thing about having highfreq=8000 is it means FastPitch is effectively predicting 16kHz audio while HiFi-GAN upsamples it to 22.05kHz. This allows training the acoustic model on 16kHz data, which greatly increases the amount of training data available.

examples/tts/conf/hifigan_dataset/hifigan_22050.yaml
rlangman: Sure, will change it.

examples/tts/conf/fastpitch/fastpitch_22050.yaml
XuesongYang: that's true for 16kHz data, but TTS typically recommend a sampling rate at least 24kHz, so I guess directly extracting features from 16kHz data is uncommon in this context.

In any case, when playing with 22050kHz data, nyquist freq (i.e. `null`) may be a standard trick for signal reconstruction purpose? 

I am good with either actions though. @subhankar-ghosh @redoctopus any comments?

redoctopus: Fine with changing it to `null`. Out of curiosity, has anyone trained a 22.05kHz model with this set to `null` yet?

rlangman: Changed to `null`.

XuesongYang: > Fine with changing it to `null`. Out of curiosity, has anyone trained a 22.05kHz model with this set to `null` yet?

Yes, I did for German: https://github.com/NVIDIA/NeMo/blob/main/examples/tts/conf/de/fastpitch_align_22050_grapheme.yaml#L28

XuesongYang (approved):
LGTM.

examples/tts/conf/fastpitch/fastpitch_22050.yaml
racoiaws: Great news, I'm glad it was finally changed to `null`, just for consistency purposes.

> The interesting thing about having highfreq=8000 is it means FastPitch is effectively predicting 16kHz audio while HiFi-GAN upsamples it to 22.05kHz. This allows training the acoustic model on 16kHz data, which greatly increases the amount of training data available

The issue with this approach is you need to match STFT window and hop. E.g. 1024 window for 16 kHz becomes 1411.2 for 22.05 kHz. Or, 1024 for 22.05 kHz would turns into 743 for 16 kHz. Tricky to think about (and to set up).

rlangman: > The issue with this approach is you need to match STFT window and hop. E.g. 1024 window for 16 kHz becomes 1411.2 for 22.05 kHz. Or, 1024 for 22.05 kHz would turns into 743 for 16 kHz. Tricky to think about (and to set up).

It is not space efficient, but couldn't you upsample the 16kHz to 22.05kHz and then it would line up?

racoiaws: Yes, sure, that is an viable option if you need this setup

PR #7156, opened 2023-08-02
[TTS] Support audio offsets in TTS data loaders

rlangman:
# What does this PR do ?

This PR contains all of the changes needed to support training TTS models with large, multi-utterance files.

Related features are all experimental, and the change is not backwards compatible as I had to change feature files from _.pt_ to _.npy_ so I could memory map them to avoid reading the entire feature file into memory.

**Collection**: [TTS]

# Changelog 
- Add support for "offset" and "duration" in manifest (similar to ASR) to allow reading a subset of audio in the data loader.
- Modify featurizer classes so that they save features as .npy files, and reads subsets of the files using numpy memory mapping when "offset" is present in the manifest.
- Add logic for computing pitch in small chunks for large audio files (to avoid out of memory issues).
- Add on-the-fly volume normalization to featurizers and data loader, as we can not do it as a preprocessing step for multi-utterance files. Volume effects feature computation, as energy is proportional to volume, and volume can change the voiced/unvoiced classification during pitch computation.
- Modify feature_stats.py so it can compute feature statistics across multiple manifests at once (technically unrelated to the rest of the PR).
- Unit tests to validate that the saving/loading logic works as expected.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/parts/preprocessing/features.py
+++ b/nemo/collections/tts/parts/preprocessing/features.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import math
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -23,7 +24,7 @@ import torch
 from torch import Tensor
 
 from nemo.collections.asr.modules import AudioToMelSpectrogramPreprocessor
-from nemo.collections.tts.parts.utils.tts_dataset_utils import get_audio_filepaths, stack_tensors
+from nemo.collections.tts.parts.utils.tts_dataset_utils import get_audio_filepaths, normalize_volume, stack_tensors
 from nemo.utils.decorators import experimental
 
 
@@ -75,13 +76,13 @@ def _get_feature_filepath(
         feature_filepath "<feature_dir>/<feature_name>/speaker1/audio1.pt"
     """
     _, audio_filepath_rel = get_audio_filepaths(manifest_entry=manifest_entry, audio_dir=audio_dir)
-    feature_filepath = feature_dir / feature_name / audio_filepath_rel.with_suffix(".pt")
+    feature_filepath = feature_dir / feature_name / audio_filepath_rel.with_suffix(".npy")
     return feature_filepath
 
 
-def _save_pt_feature(
+def _save_feature(
     feature_name: Optional[str],
-    feature_tensor: Tensor,
+    features: np.ndarray,
     manifest_entry: Dict[str, Any],
     audio_dir: Path,
     feature_dir: Path,
@@ -96,15 +97,16 @@ def _save_pt_feature(
         manifest_entry=manifest_entry, audio_dir=audio_dir, feature_dir=feature_dir, feature_name=feature_name
     )
     feature_filepath.parent.mkdir(exist_ok=True, parents=True)
-    torch.save(feature_tensor, feature_filepath)
+    np.save(file=str(feature_filepath), arr=features)
 
 
-def _load_pt_feature(
+def _load_feature(
     feature_dict: Dict[str, Tensor],
     feature_name: Optional[str],
     manifest_entry: Dict[str, Any],
     audio_dir: Path,
     feature_dir: Path,
+    indices: Optional[Tuple[int, int]] = None,
 ) -> None:
     """
     If feature_name is provided, load feature into feature_dict from .pt file.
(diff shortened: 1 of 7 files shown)


racoiaws:
Is "peak normalization" really that useful to have it as argument everywhere? Yes it is a very simple thing to do, but it is not that well connected to volume we care about.

A much better measure would be LUFS, but that's a candidate for preprocessing step, not on-the-fly loading

rlangman:
> Is "peak normalization" really that useful to have it as argument everywhere? Yes it is a very simple thing to do, but it is not that well connected to volume we care about.
> 
> A much better measure would be LUFS, but that's a candidate for preprocessing step, not on-the-fly loading

It is useful for some things. The scale of spectrogram features, and its corresponding energy, are dependent on the scale of the audio samples. For things like VAD and pitch it is less important, as those would only be impacted if the peak is so low that it is almost silent (but you will find several such audio files in most public TTS datasets). For single speaker datasets where the recording is relatively consistent, it is likely irrelevant.

I am sure there are better metrics we could use to estimate the volume or control the scale, if someone wants to implement them.

There is probably a better way to ensure that all audio that gets loaded from disk receives the same preprocessing, without needing to pass the same parameters to every featurizer. But it might be overkill if peak norm is the only step that is being done now.

Also In general an issue I see with storing and training large audio files is that it will complicate or limit the types of preprocessing we can do. Things like speech restoration, audio compression, LUFS, are going to need some generic version of what I did here with the pitch in order to preprocess the audio in chunks and stick the results afterwards.

racoiaws:
Sorry, I actually did not made the point I originally wanted to make. You are right that there are cases when normalizing to 0 dBFS is better than not doing it.

But do we really need to treat peak normalization as a [0, 1] parameter instead of True/False? This API suggests that normalizing to different peaks might be useful, but I don't see how it could be.

rlangman:
> Sorry, I actually did not made the point I originally wanted to make. You are right that there are cases when normalizing to 0 dBFS is better than not doing it.
> 
> But do we really need to treat peak normalization as a [0, 1] parameter instead of True/False? This API suggests that normalizing to different peaks might be useful, but I don't see how it could be.

I cannot think of a situation where you care what the peak is. Do you think it would make more sense to have it as a true/false parameter?

nithinraok:
@XuesongYang @redoctopus can you please review this PR

nemo/collections/tts/parts/preprocessing/features.py
```diff
@@ -75,13 +73,27 @@ def _get_feature_filepath(
         feature_filepath "<feature_dir>/<feature_name>/speaker1/audio1.pt"
     """
     _, audio_filepath_rel = get_audio_filepaths(manifest_entry=manifest_entry, audio_dir=audio_dir)
-    feature_filepath = feature_dir / feature_name / audio_filepath_rel.with_suffix(".pt")
+    feature_filepath = feature_dir / feature_name / audio_filepath_rel.with_suffix(".npy")
```
XuesongYang, line 76: could you double-check .npy or .pt? The docstring seems use .pt.

scripts/dataset_processing/tts/compute_features.py
```diff
@@ -51,9 +52,18 @@ def get_args():
     parser.add_argument(
         "--feature_dir", required=True, type=Path, help="Path to directory where feature data will be stored.",
     )
+    parser.add_argument(
+        "--dedupe_files",
+        action=argparse.BooleanOptionalAction,
+        help="If given, will only process the first manifest entry found for each audio file.",
+    )
+    parser.add_argument(
+        "--overwrite", action=argparse.BooleanOptionalAction, help="Whether to overwrite existing feature files.",
```
XuesongYang, line 61: nit: this is new, not sure if it would be an issue. I usually used `action='store_true'`.

XuesongYang:
LGTM. @racoiaws, what do you think about the follow-ups by Ryan?

XuesongYang (dismissed):

nemo/collections/tts/parts/preprocessing/features.py
rlangman: Updated docstrings to `.npy`

scripts/dataset_processing/tts/compute_features.py
rlangman: Reading some threads online, it sounds like the two are equivalent. `BooleanOptionalAction` was is recommended for Python 3.9+.

racoiaws:
Making peak normalization ceiling a floating-point parameter is effectively telling users "this might be a useful thing to tune". I don't think it is.

Could this please be changed to a boolean True/False? LGTM otherwise

(It would be interesting to see if vrms (or lufs) normalization would help with model convergence/quality though, but it's definitely out of scope for this PR)

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -61,7 +66,7 @@ def get_audio_filepaths(manifest_entry: Dict[str, Any], audio_dir: Path) -> Tupl
     return audio_filepath_abs, audio_filepath_rel
 
 
-def normalize_volume(audio: np.array, volume_level: float) -> np.array:
+def normalize_volume(audio: np.array, volume_level: float = 0.95) -> np.array:
```
racoiaws, line 69: Just one little thing, feel free to skip

Digital volume is commonly measured is dBFS. 0.95 max absolute sample value corresponds to 20*log10(0.95)=-0.44 dBFS.

Would be nice to see this acknowledged in the docstring

racoiaws (approved):

PR #7174, opened 2023-08-06
Allow applications to control the encoder mask type

nzmora-nvidia:
# What does this PR do ?

When TensorRT optimizes Nemo GPT models it uses a different mask shape in the Context phase (triangular mask, i.e. causal) and in the Generation phase (1D mask, i.e. padding).
To support this TensorRT needs to control GPT's encoder mask type (which defaults to causal) via the model configuration.

**Collection**: NLP

# Changelog
- `MegatronGPTModel::model_provider_func` reads an optional `encoder_attn_mask_type` argument from the config file and passes it to the GPTModel constructor. The default value is a causal mask.
- Add an optional `encoder_attn_mask_type` argument to GPTModel constructor. The argument defaults to `AttnMaskType.causal`.
- Pass this argument to `get_language_model`  (previously hard-coded: `encoder_attn_mask_type=AttnMaskType.causal`)

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -60,8 +60,8 @@ from nemo.utils import logging
 
 try:
     import apex.transformer.pipeline_parallel.utils
+    from apex.transformer.enums import AttnMaskType
     from apex.transformer.pipeline_parallel.utils import get_num_microbatches
-    from apex.transformer.enums import AttnMaskType 
 
     HAVE_APEX = True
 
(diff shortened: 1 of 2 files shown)


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -65,6 +65,8 @@
     HAVE_APEX = True
```
yen-shi, line 65: We need to import `AttnMaskType` here if I read correctly:
```suggestion
    from apex.transformer.enums import AttnMaskType

    HAVE_APEX = True
```

nzmora-nvidia: Thanks! Good catch

nzmora-nvidia:
I will discuss this PR with @oyilmaz once I'm back at work.

nzmora-nvidia:
@ericharper @oyilmaz before you close this PR, please look again and consider merging. I think this extra flexibility in the API is important for NeMo regardless of whether TRT requires (or does not) this change.

PR #7211, opened 2023-08-10
Mcore LLAMA sync up with Mcore gpt model

blahBlahhhJ:
mcore_llama sync up with mcore_gpt_model


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -426,9 +444,10 @@ def configure_optimizers(self):
             if self.cfg.get('virtual_pipeline_model_parallel_size', None) is not None:
                 # Initialize a bucket for each virtual pipeline stage
                 for module in self.model:
-                    if isinstance(module, Float16Module):
+                    if isinstance(module, (Float16Module, MCoreFloat16Module)):
                         module = module.module
                     stage_bucket = []
+                    layers = module.decoder.layers if self.mcore_gpt else module.language_model.encoder.layers
                     for layer in module.language_model.encoder.layers:
```
blahBlahhhJ, line 451: @ericharper should this be `for layer in layers`?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -119,14 +120,16 @@ def init_model(self, cfg: DictConfig, trainer: Trainer):
             )
             frozen_model_cfg.activations_checkpoint_method = self.cfg.get("activations_checkpoint_method", None)
 
-        if self.trainer.precision == 'bf16':
+        if self.trainer.precision in ['bf16', 'bf16-mixed']:
+            # set hidden size in the model parallel config for pipeline parallel schedules
+            setattr(self.config, 'hidden_size', frozen_model_cfg.hidden_size)
```
blahBlahhhJ, line 125: @ericharper why is this only in 'bf16' or 'bf16-mixed'? shouldn't it be a common attribute?

ericharper: good catch, I think this may have by accidentally from the rebase

PR #7262, opened 2023-08-17
Resume checkpoint priority

maanug-nv:
# What does this PR do ?

Fix resume from checkpoint behavior to have lower priority than experiment dir.

# Changelog 
- resume from checkpoint behavior
- reorganize `check_resume` for prints

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/utils/exp_manager.py
```diff
@@ -333,24 +333,15 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
         resume_if_exists=cfg.resume_if_exists,
     )
 
-    if cfg.resume_if_exists:
```
titu1994, line 336: Id suggest keeping `if resume_if_exists` outside as well as passing it as an arg. Its the first bool flag that checks if you even want to resume or not.

nemo/utils/exp_manager.py
```diff
@@ -565,69 +558,70 @@ def check_resume(
...
-            raise ValueError(
-                f"Found {end_checkpoints[0]} indicating that the last training run has already completed."
-            )
-    elif not len(last_checkpoints) > 0:
-        if resume_ignore_no_checkpoint:
-            logging.warning(f"There were no checkpoints found in {checkpoint_dir}. Training from scratch.")
-            return
+    if resume_from_checkpoint:
+        checkpoint = resume_from_checkpoint
+    if resume_if_exists:
```
titu1994, line 564: Resume if exists will always be true for all runs - even the first run when there's no other checkpoint to resume for - that's why the second flag `resume_ignore_no_checkpoint` comes into play.

nemo/utils/exp_manager.py
```diff
@@ -565,69 +558,70 @@ def check_resume(
...
+        # Use <log_dir>/checkpoints/ unless `dirpath` is set
+        checkpoint_dir = Path(dirpath) if dirpath else Path(Path(log_dir) / "checkpoints")
+
+        end_checkpoints = list(checkpoint_dir.rglob("*end.ckpt"))
+        last_checkpoints = list(checkpoint_dir.rglob("*last.ckpt"))
+        if not checkpoint_dir.exists() or (not len(end_checkpoints) > 0 and not len(last_checkpoints) > 0):
```
titu1994, line 570: Why the new checks about number of checkpoints?

nemo/utils/exp_manager.py
```diff
@@ -565,69 +558,70 @@ def check_resume(
...
-                    run_count += 1
-            new_run_dir = Path(Path(log_dir) / f"run_{run_count}")
-            new_run_dir.mkdir()
-            for _file in files_to_move:
-                move(str(_file), str(new_run_dir))
+    if checkpoint is not None:
+        trainer.ckpt_path = str(checkpoint)
+        logging.info(f'Resuming training from checkpoint: {trainer.ckpt_path}')
+
+        if is_global_rank_zero():
```
titu1994, line 607: This should be moved left indent one time - earlier code has it outside.

titu1994 (commented):
Minor comments but looks good

nemo/utils/exp_manager.py
maanug-nv: My understanding was that `resume_if_exists` only applies to the checkpoints in the log dir, and `resume_from_checkpoint` is outside that. So this will enforce that `resume_if_exists` must be set to use any kind of checkpoint, either from the log dir or manually specified with resume from, is that correct?

nemo/utils/exp_manager.py
maanug-nv: There was an `elif not len(last_checkpoints) > 0:` block before that was doing the exact same things as this first if block. I felt this is simpler to just check both cases at the top, if the dir doesn't exist or if it has no checkpoints.

nemo/utils/exp_manager.py
maanug-nv: Earlier code also had an early return when checkpoint dir doesn't exist but I had to remove that for `resume_from_checkpoint`. So I'm not sure if this section should run when `trainer.ckpt_path` is `None`.

nemo/utils/exp_manager.py
titu1994: Makes sense

maanug-nv:
Closing and moving to #7335

maanug-nv:
Closing and moving to #7335

Issue #7187, opened 2023-08-08
Concerns about the implementation of `end_of_generation_condition()`


odelalleau:
The code of `end_of_generation_condition()` https://github.com/NVIDIA/NeMo/blob/13791d27ec8b322da11e8324ff07124aa4e6c55f/nemo/collections/nlp/modules/common/text_generation_strategy.py#L138 seems weird to me => opening an issue to clarify whether it is working as intended (cc @gshennvm @yidong72)

I have two concerns:

1. I find it weird to see `<extra_id_1>` hardcoded in this function -- I thought that the goal of providing `end_strings` as argument would be to avoid having to hardcode special strings
2. I am confused by something that is best explained through an example. Let's say I want to use `end_strings = ["abc"]`, which is tokenized into [1, 2, 3] (tokens for "a", "b", "c"), while `<extra_id_1>` is tokenized into [0]. It seems to me that this code will lead to `ids_1 = tokenizer.text_to_ids(f'<extra_id_1>abc') = [0, 1, 2, 3]`, `ids_2 = [0]`, and `token_id = ids_1[len(ids_2) :][0] = 1`. As a result, as soon as `p` (the last token generated) is equal to 1 (i.e., as soon as an `"a"` is generated), we consider that the generation has ended. Am I missing something?

yidong72:
It is because sentence piece tokenize tends to add an extra token in front of the string. the "<extra_id_1>" is added  in the front so it won't merge with the later string after tokenization. eventually we only take the difference between id2 and id1. This is a hack to deal with the sentencepiece behavior.

yidong72:
> I am confused by something that is best explained through an example. Let's say I want to use end_strings = ["abc"], which is tokenized into [1, 2, 3] (tokens for "a", "b", "c"), while <extra_id_1> is tokenized into [0]. It seems to me that this code will lead to ids_1 = tokenizer.text_to_ids(f'<extra_id_1>abc') = [0, 1, 2, 3], ids_2 = [0], and token_id = ids_1[len(ids_2) :][0] = 1. As a result, as soon as p (the last token generated) is equal to 1 (i.e., as soon as an "a" is generated), we consider that the generation has ended. Am I missing something?

This is a bug. The stop criteria are too aggressive as it only matches the first token of the string.  If we set stop sting as ['abc'] it will stop at 'a' even the model wants to generate other string starts with 'a'. 
This is the fix to the problem.
```python
            for end_string in end_strings:
[...]

PR #7267, opened 2023-08-18
Bugfix and optimization in `end_of_generation_condition()`

odelalleau:
# What does this PR do ?

It fixes a bug in `end_of_generation_condition()` (#7187) and makes it significantly faster in some cases.

**Collection**: nlp

# Changelog 
- Fixed edge case that could make text generation stop earlier than intended
- Speed up text generation when using custom end strings

# Detailed explanation

1. Bugfix

The previous implementation did not verify that `end_string` was encoded
into a single token, which could trigger the end of generation earlier
than intended (see discussion in #7187)

2. Optimization

The previous implementation was scaling linearly with the batch size and
quadratically with the length of the generated sequence, which could
lead to a significant overhead in some situations.

The new implementation is much more efficient in "normal" situations
(where the end of generation is identified by a set of unique tokens),
and raises a warning when it needs to fallback to the inefficient string
matching case.

Note that it does not behave exactly the same as before, because we skip
the string comparison when the end strings all have unique tokens
associated to them. For instance, in the previous implementation, if the
model had generated the string
        "Some string.<|endoftext|>"
(where "<|endoftext|>" would really be generated as a string, and not as
a single token), then the previous implementation would have considered
it to be the end of generation (assuming `end_strings` has length > 1),
while the new one would not. The previous behavior was likely a bug
though, since we expect models to generate the special tokens associated
to end strings when they exist (for instance, the standard case
`end_strings=["<|endoftext|>"]` has always been handled by just
comparing the last token to `eod_id`).

3. Minor change

See commit message of ab56968fd for the explanation regarding the warning that was added in `__init__()`

# Tests

Hopefully there are existing tests on CI -- I have tested this myself on my own jobs.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

Fixes ##7187


nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -176,6 +180,65 @@ def post_generation_process(self, output):
...
+                end_tokens.add(end_string_to_token[end_string])
+                continue
+            except KeyError:
+                if end_string in self._end_of_generation_cache["end_strings_to_check"]:
+                    end_strings_to_check.append(end_string)
+                    continue
+
+            # `end_string` does not exist in the cache yet: check if `end_string` is a special token for
+            # the tokenizer. Ideally, we would simply use `tokenizer.text_to_ids(end_string)`, but some
+            # tokenizers (e.g., SentencePiece) may prefix the special token with another token associated
```
aklife97, line 222: an easier route than doing all this would be to just use `token_to_id`. That function will return unk_id if the token does not exist / the ID for token if token exists

Example:

>>> nvllm.unk_id
1

>>> nvllm.token_to_id("abcd")
237433

>>> nvllm.token_to_id("abcd1233")
1

aklife97: Actually, this might be tricky for autotokenizer, so the current way looks fine

nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -147,26 +155,22 @@ def end_of_generation_condition(
...
-            tokenizer = self.model.tokenizer
-            conditions = []
-            end_tokens = set()
-            end_tokens.add(eod_id)
-            for end_string in end_strings:
-                ids_1 = tokenizer.text_to_ids(f'<extra_id_1>{end_string}')
-                ids_2 = tokenizer.text_to_ids('<extra_id_1>')
-                if len(ids_1) <= len(ids_2):
-                    continue
-                token_id = ids_1[len(ids_2) :][0]
```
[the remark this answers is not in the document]
odelalleau: Just to be sure we're on the same page, I think you mean "why don't we check that `len(ids_1) == len(ids_2) + 1` (because `token_id` in that code is an integer so it has no length)

This indeed fixes the bug, but there is still a need to optimize this function because calling `text = tokenizer.ids_to_text(token_item.tolist())` below is quite inefficient (we are decoding the whole model output at each step, over the whole batch).

The optimization part of this PR has two parts:

1. (Most important) Try to skip the text decoding if possible, by only comparing `prev` to the list of end tokens
2. (Less important) Cache the list of end tokens / end strings so that we don't re-compute it at each step. This matters less and I implemented it mostly to be safe in all cases, especially when using a tokenizer that would somehow be pretty slow (compared to the model generation speed). I can get rid of it if you feel like this isn't worth it (NB: I tried to keep the complexity within `_get_end_of_generation_tokens_and_strings(()` and simplify `end_of_generation_condition()`)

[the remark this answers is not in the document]
odelalleau: NB: also we can't cache `self.end_tokens` / `self.end_strings` because the API of the method supports arbitrary values for `end_strings`, so it would be dangerous to cache these values (unless we use `end_strings` for the cache key -- I thought about it but did not see an obvious advantage so I didn't go for this option).

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: I wasn't aware of `token_to_id()`, thanks for pointing it out. Out of curiosity, why wouldn't it work with autotokenizer?

nemo/collections/nlp/modules/common/text_generation_strategy.py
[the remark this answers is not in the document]
aklife97: I'm fine with whichever way we cache, so all good there

okay, I see so (1) brings the major speed improvement here, makes sense. Added a new comment below that can help further improve speed

nemo/collections/nlp/modules/common/text_generation_strategy.py
aklife97: its a mess... autotokenizer does not always maintain unk_id in nemo

nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -147,26 +155,22 @@ def end_of_generation_condition(
...
+
+        end_tokens, end_strings_to_check = self._get_end_of_generation_tokens_and_strings(eod_id, end_strings)
+        assert end_tokens
+
+        is_end = torch.isin(prev, torch.tensor(list(end_tokens), dtype=prev.dtype, device=prev.device))
+
+        if end_strings_to_check:
+            # The loop below is inefficient (see warning in `_get_end_of_generation_tokens_and_strings()`)
+            for idx, token_seq in enumerate(tokens):
+                text = self.model.tokenizer.ids_to_text(token_seq.tolist())
```
aklife97, line 170: instead of de-tokenizing the whole sentence, why not just de-tokenize the last N tokens where N is the len(end_string) since at worst 1 char can be 1 token?

aklife97: I'm assuming this can be significant savings over detokenizing the entire text which can be very long

nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -176,6 +180,65 @@ def post_generation_process(self, output):
...
+                if end_string in self._end_of_generation_cache["end_strings_to_check"]:
+                    end_strings_to_check.append(end_string)
+                    continue
+
+            # `end_string` does not exist in the cache yet: check if `end_string` is a special token for
+            # the tokenizer. Ideally, we would simply use `tokenizer.text_to_ids(end_string)`, but some
+            # tokenizers (e.g., SentencePiece) may prefix the special token with another token associated
+            # to an empty string. The code below is thus meant to extract the special token associated to
+            # `end_string` (if it exists). Note that using "This is a sequence." as reference is arbitrary.
+            ids_ref = tokenizer.text_to_ids("This is a sequence.")
```
yidong72, line 225: `"This is a sequence."` is not a good prefix string since the `.` character might be merged with other characters in the `end_string`.  Ideally we should use some special token so it will always stand alone.

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: Ok, let's forget about it then :)

nemo/collections/nlp/modules/common/text_generation_strategy.py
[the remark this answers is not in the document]
odelalleau: Alright, resolving this comment then

nemo/collections/nlp/modules/common/text_generation_strategy.py
aklife97: was exactly commenting this, but the code checks `ids_with_end_string[:-1] == ids_ref` below and also `.` is usually an independent token in most tokenizers

ideally a seaprate special token might be better, but do we have any token that we can guarantee will always be present?

aklife97: odds wise, I feel having `.` as the special token is more likely than any other thing. worst case we do a string check but it still doesnt give incorrect answer

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: Yeah I thought about trying something like that, but didn't do it for two reasons:
1. It's not obvious to me how many tokens we should decode, since some tokens can actually be decoded into an empty string. I was thinking I'd need to start from some "best guess" value (N in your example is a good start), then increase it if somehow it's not enough. It's definitely doable, but then point 2 below came to mind...
2. I'm not convinced there's a real use case for string-based matching. AFAIK for our models we make sure that the end of response is identified by special tokens. So making the code more complex to speed-up a use case we may not care about didn't seem worth it.

That being said, I may very well be wrong w.r.t. (2), in which case I'm happy to implement something like that. Just let me know if you're aware of situations where this is actually useful (beyond debugging).

aklife97: > It's not obvious to me how many tokens we should decode
N is a hard bound, you dont need to de-tokenize more than N

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: I started to write an explanation for my reasoning here, but this brought up a doubt in my mind that is very important to clear up, as it may invalidate a strong assumption made in this PR:

Is it possible that we may want to use an end string associated to a unique token (e.g., `<extra_id_1>`) and yet expect the model to end a response with the *string* `"<extra_id_1>"` but *without* generating this token?
I assumed no (similar to how we don't stop generation if the model generates the string "<|endoftext|>" instead of `eos_id`), but can you confirm this is correct?

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: > A whole token is never really an empty string

I learned that this is not true -- there are special tokens that get decoded into empty strings, and this is tokenizer-dependent.

Anyway, I'm happy to implement it, I just want someone to tell me that there's actually a real use case for this :)

aklife97: yes, thats specifically true for control_symbols but we shouldn't have them in the text

but its fine, we can let it be as is. can add it later in a different PR if we feel we're stopping on strings for any usecase since I currently dont see us doing that

aklife97 (commented):
LGTM! Everything looks good to me, there are some pending comments by Yi but once we feel those are resolved I think we should be good to merge

nemo/collections/nlp/modules/common/text_generation_strategy.py
odelalleau: > `"This is a sequence."` is not a good prefix string since the `.` character might be merged with other characters in the `end_string`. Ideally we should use some special token so it will always stand alone.

So, following offline discussion, there are three cases:

1. `end_string` is actually a special token for the tokenizer (e.g. `<extra_id_1>` typically is, when we use it). Then we are guaranteed that it will be tokenized as a single token, and we will identify it properly.

2. `end_string` is *not* a special token for the tokenizer, and is tokenized into more than one token: in that case we don't care whether or not it is merged with the preceding `.`, because we will need to use string comparisons anyway.

3. `end_string` is *not* a special token for the tokenizer, and is tokenized into a single token: in that case either we identify the single token and rely on token comparison, or we don't (because of the tokenizer merging the `.`) and we fall back to string matching. I would argue that the latter is safer, because (a) it will ensure we always end generation correctly, and (b) it shows a warning that may alert the user that something may be off (there's a good chance they didn't expect the tokenizer to merge `end_string` with other characters). And if the tokenizer merges the `.` with `end_string` it seems particularly important to be aware of it, since this sequence of characters is likely to be quite common to finish responses. So it seems to me that it's actually a better option than `<extra_id_1>` (though I admit I hadn't thought it through before ;))

yidong72: the assmption is either end_string is a special token or not, the generation will end with it inclusively. It is up to tokenizer to decide whether the `ids_to_text` method wants to show it. I think `<extra_id_1>` is a better prefix because if it is a special token, it will work as expected. If it is not, the `>` is less likely to be merged with the end_string, which is a cleaner case as you mentioned in the 3.  If '.' is merged with the other partial end-string during the generation, the string match might not capture it correctly. E.g. end_string is "hello", '.' is merged to 'hel', the generated tokens are '.hel', 'lo,'  it won't trigger the string match check as we use 'endswith` method.

odelalleau: Alright, let's go with `<extra_id_1>`, I don't think it's a big deal anyway, it shouldn't matter much in practice. This is done in c9a6d71f2778b7b85407e8aeca3d3f3e5b6df55e (I also rebased on top of main, but there are no changes in previous commits).

I still want to address the point below since a similar situation may still happen with `<extra_id_1>`:

> If '.' is merged with the other partial end-string during the generation, the string match might not capture it correctly. E.g. end_string is "hello", '.' is merged to 'hel', the generated tokens are '.hel', 'lo,' it won't trigger the string match check as we use 'endswith` method.

If '.' is merged with 'hel', then this will trigger the string match because the second condition of this check won't be satisfied (we  check both that there are N+1 tokens *and* also that the first N tokens are the same):
```python
if len(ids_with_end_string) == len(ids_ref) + 1 and ids_with_end_string[:-1] == ids_ref:
```
As a result, there will be a warning displayed, and the comparison will be made with `text.endswith("hello")` which will match any generation ending with "hello", regardless of what tokens this corresponds to.
Note that there could still be situations where the model generates "hello" without generation ending, e.g. if it generates tokens ".hel" followed by "lo world". But this case was not handled previously either, and it is unclear that we should stop there (since we can't truncate the model output mid-token, so the response would actually end with "hello world" rather than "hello".

yidong72 (approved):
LGTM. thanks for refining the end of generation logics.

yidong72 (approved):

aklife97 (approved):
LGTM, thank you!

PR #7276, opened 2023-08-19
add PEFT adapter support for mcore gpt path

blahBlahhhJ:
Add PEFT adapter support for mcore gpt path.

contains a nemo.core change to accept kwargs to pass into hydra instantiate.

currently only support Ptuning and Lora now.

tested on llama2 7b TP SP PP DP

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
@@ -70,7 +70,6 @@ class MegatronGPTPEFTModel(MegatronGPTSFTModel):
         assert len(self.name_key_to_cfg) > 0, "name_key_to_cfg has not been set no PEFT modules will be added"
         logging.info(f"Before adding PEFT params:\n{self.summarize()}")
         for name, module in self.named_modules():
-            # if there's an infinite loop here, check that the added adapter modules doesn't use the same key as mcore targets.
             if self.mcore_gpt:
                 for peft_key in self.peft_name_keys:
                     for mcore_target, mcore_mixin in self.name_key_to_mcore_mixins[peft_key]:
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
@@ -518,10 +518,14 @@ class MegatronGPTLoRAModel(MegatronGPTLayerwisePEFTModel):
         else:
             kv_channels = cfg.kv_channels
         projection_size = kv_channels * cfg.num_attention_heads
+        num_query_groups = cfg.get("num_query_groups", None)
+        if num_query_groups is None:
+            num_query_groups = cfg.num_attention_heads
+        qkv_projection_size = projection_size + 2 * kv_channels * num_query_groups
 
         adapter_cfg = LoraKQVAdapterConfig(
             in_features=cfg.hidden_size,
-            out_features=3 * projection_size,
+            out_features=qkv_projection_size,
             dim=lora_cfg.adapter_dim,
             norm_position=None,
             norm_type=None,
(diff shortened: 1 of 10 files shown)


nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
```diff
@@ -0,0 +1,79 @@
...
+    from megatron.core.transformer.custom_layers.transformer_engine import TELinear
+
+    HAVE_MEGATRON_CORE = True
+
+except (ImportError, ModuleNotFoundError):
+
+    HAVE_MEGATRON_CORE = False
+
+    logging.info("Megatron-core is required to use McoreAdapterModuleMixin.")
+    raise RuntimeError("McoreAdapterModuleMixin can not run without Megatron-core.")
```
blahBlahhhJ, line 24: didn't find a good place to raise this error. Should I do it here? @ericharper

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -55,16 +60,34 @@ def init_peft_modules(self):
...
+            # if there's an infinite loop here, check that the added adapter modules doesn't use the same key as mcore targets.
+            if self.mcore_gpt:
                 for peft_key in self.peft_name_keys:
-                    peft_cfg = self.name_key_to_cfg[peft_key]
-                    if model_utils.import_class_by_path(peft_cfg._target_) in module.get_accepted_adapter_types():
-                        module.add_adapter(
-                            name=peft_key, cfg=peft_cfg,
-                        )
+                    for mcore_target, mcore_mixin in self.name_key_to_mcore_mixins[peft_key]:
+                        if name.split(".")[-1] == mcore_target:
```
blahBlahhhJ, line 73: should I do a suffix match or a full regex match?

e.g. the template for lora mcore target will be 
```
"model\.decoder\.layers\.[0-9]+\.self_attention\.linear_qkv"
```

blahBlahhhJ:
update: will pull some general PEFT changes out to merge to main first, and then rebase this branch after those were merged.
#7287 
#7285

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
```diff
@@ -0,0 +1,66 @@
...
+    module.__class__ = mcore_mixin
+    module.mcore_register_adapters()
+
+
+class McoreAdapterModuleMixin(adapter_mixins.AdapterModuleMixin):
+    def mcore_register_adapters(self):
+        raise NotImplementedError("Mcore mixins should implement setup_adapters on a subclass of MyBase")
+
+
+class LoraTELinear(TELinear, McoreAdapterModuleMixin):
```
arendu, line 27: Could we use the [SelfAttention class from Megatron-LM ](https://github.com/NVIDIA/Megatron-LM/blob/f24fac4ed0dcf0522056521a93445d9a82f501a9/megatron/core/transformer/attention.py#L245) and implement the [get_query_key_value_tensors](https://github.com/NVIDIA/Megatron-LM/blob/f24fac4ed0dcf0522056521a93445d9a82f501a9/megatron/core/transformer/attention.py#L266C20-L266C20) here instead of going to the lower level TELinear? It is a few more lines of code but is gives me a more general recipe for implementing future PEFT methods. LoRA is a bit of an outlier because it can be fully encapsulated into a low-level Linear layer.

arendu:
Thanks, @blahBlahhhJ this looks like a great start with some useful key ingredients -- eg. class swapping with NeMo's subclass of Mcore is great!. I left a comment [here](https://github.com/NVIDIA/NeMo/pull/7276/files#r1301979030) which would be useful to address.

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
blahBlahhhJ: I've seen people injecting lora on not only `linear_qkv` but also other linears in the attention, although nemo only applies lora on `linear_qkv` for now...

nemo/collections/nlp/modules/common/megatron/adapters/parallel_adapters.py
```diff
@@ -322,7 +322,7 @@ def __init__(
         gradient_accumulation_fusion = False
         # (@adithyare) the persistent=False will not pollute the indices into the state_dict of this module.
         self.register_buffer("indices", torch.LongTensor(list(range(self.virtual_tokens))), persistent=False)
-        self.embedding = torch.nn.Embedding(self.virtual_tokens, self.embedding_dim)
+        self.prompt_embedding = torch.nn.Embedding(self.virtual_tokens, self.embedding_dim)
```
arendu, line 325: is this change required? if not, could we leave it as is.

PR #7232, opened 2023-08-14
Add precision and megatron_amp_O2 configs to adapters

guyueh1:
- Add precision and megatron_amp_O2 parameters to adapters
- Make adapter modules follow the config of base model for precision and megatron_amp_O2

# What does this PR do ?

Make adapter modules follow the config of base model for precision and megatron_amp_O2

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #7285, opened 2023-08-21
Fix adapters and ptuning for amp O2

guyueh1:
# What does this PR do ?

Fix the issues when using megatron_amp_O2 on PEFT models

# Changelog 
- Transform adapters to fp16/bf16 when the base model uses megatron_amp_O2. When using amp_O2 we should make model parameters in fp16/bf16. Previous PR #7232 handles the precision issue in ColumnParallelLinear and RowParallelLinear in PEFT models, and this PR handles precision casting of the rest of the modules such as layernorm.
- Explicitly set the ptuning inference table to untrainable for FP16/BF16 modules. Previous whitelist only includes the name of the inference table as in a FP32 model. We add its name in FP16/BF16 models to whitelist.
- Explicitly set ptuning inference table to be untrainable in AdapterPTuningModel like in PTuningModel

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #7232

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
@@ -69,7 +69,7 @@ class MegatronGPTPEFTModel(MegatronGPTSFTModel):
                     peft_cfg = self.name_key_to_cfg[peft_key]
                     if model_utils.import_class_by_path(peft_cfg._target_) in module.get_accepted_adapter_types():
                         module.add_adapter(
-                            name=peft_key, cfg=peft_cfg, base_model_cfg=self.cfg
+                            name=peft_key, cfg=peft_cfg, model_parallel_config=self.model_parallel_config
                         )
         logging.info(f"After adding PEFT params:\n{self.summarize()}")
         return True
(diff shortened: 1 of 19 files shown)


nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -47,6 +48,15 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
     def first_stage_of_pipeline(self):
         if hasattr(self, "model") and hasattr(self.model, "pre_process"):
             return self.model.pre_process
+        elif (
+            hasattr(self, "model")
+            and isinstance(self.model, Float16Module)
```
blahBlahhhJ, line 53: is this type check necessary? mcore uses a different `Float16Module` class so if you want to do this, I'll need to make it `... and (isinstance(self.model,Float16Module) or isinstance(self.mode, MCoreFloat16Module)) and ...`

will look pretty messy.

guyueh1: This `isinstance` check is essentially the same as `hasattr(self.model, 'module')`. It is not necessary, eliminating it still works in my test, and I was just adding it to make the code more readable (the long sequence of `hasattr` is very confusing). 

It really comes to the question that whether checking `hasattr(self.model, 'module'` will consistently work. If provided that Float16Module and MCoreFloat16Module both use .module to save the original network, I think we can remove the `isinstance` check.

arendu (approved):
LGTM!! left one minor comment

PR #7299, opened 2023-08-23
NeMo MCore llama2 support + MCore PEFT adapters

blahBlahhhJ:
# What does this PR do ?

Support Llama2 models in mcore.
Support PEFT adapters in mcore.

# Changelog 

# Usage
* `mcore_gpt` needs to be enabled for LLAMA2.

```python
model.mcore_gpt = True
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -173,16 +177,18 @@ def init_model(self, cfg: DictConfig, trainer: Trainer):
...
+                self.word_embeddings = self.frozen_model.model.language_model.embedding.word_embeddings
 
         self.padded_vocab_size = self.frozen_model.padded_vocab_size
 
         # Prepare pseudo token ids for virtual/virtual prompt tokens
         self.pseudo_tokens = get_pseudo_tokens(self.max_virtual_tokens)
         if isinstance(self.tokenizer, SentencePieceTokenizer):
             if not self.tokenizer.legacy:
-                self.tokenizer.pad_token = self.tokenizer.ids_to_tokens([self.tokenizer.pad_id])[0]
+                self.tokenizer.pad_token = self.tokenizer.ids_to_tokens([self.tokenizer.eos_id])[0]
```
blahBlahhhJ, line 191: llama2 tokenizer does not have a pad token, so we need to turn it into an eos token for [launcher script -> prompt learning] to work.
@ericharper should we rollback this change?

lhb8125: + @aklife97 to review.

aklife97: but this change is not model-specific. this would likely be an issue.
let's have a check that uses pad_id if it is present in the tokenizer. If not, we can use eos_id

blahBlahhhJ: fixed.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -133,7 +133,7 @@ def load_config_and_state_dict(
                 # Change current working directory to
                 os.chdir(tmpdir)
                 if override_config_path is None:
-                    config_yaml = os.path.join(tmpdir, self.model_config_yaml)
+                    config_yaml = self.model_config_yaml
```
ericharper, line 136: Why are we changing this?

blahBlahhhJ: @lhb8125 might be able to answer this

examples/nlp/language_modeling/conf/megatron_llama_inference.yaml
```diff
@@ -0,0 +1,37 @@
+inference:
```
aklife97, line 1: why do we need this file? shouldn't the .nemo inference be identical to any other .nemo file?

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -900,7 +903,31 @@ def main():
...
+        restore_path=args.model_file,
+        trainer=Trainer(devices=1, strategy=NLPDDPStrategy(), accelerator="cpu", precision=precision),
+        map_location=torch.device("cpu"),
+        return_config=True,
+    )
+    plugins = []
+    if precision in [16, '16', 'bf16', '16-mixed', 'bf16-mixed']:
+        scaler = None
+        if precision in [16, '16', '16-mixed']:
+            scaler = GradScaler(
```
aklife97, line 916: why do we need to setup the GradScaler in this script?

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
```diff
@@ -0,0 +1,106 @@
...
+
+class MCoreAdapterModuleMixin(adapter_mixins.AdapterModuleMixin):
+    def mcore_register_adapters(self):
+        raise NotImplementedError("Mcore mixins should implement setup_adapters on a subclass of MyBase")
+
+
+class MCoreSelfAttentionMixin(SelfAttention, MCoreAdapterModuleMixin):
+    def mcore_register_adapters(self):
+        self.set_accepted_adapter_types([LoraKQVAdapterConfig._target_])  # only self attn (packed qkv) for now
+        self.linear_qkv.return_layernorm_output = True  # need layernorm output for lora mlp
```
ericharper, line 43: Should this be private? Could we add a docstring?

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        # [sq, b, ng, np/ng * hn] -> [sq, b, np, hn]
+        query = query.reshape(query.size(0), query.size(1), -1, self.hidden_size_per_attention_head)
+
+        return query, key, value
+
+
+class MCoreGPTEmbeddingMixin(GPTEmbedding, MCoreAdapterModuleMixin):
+    def mcore_register_adapters(self):
+        self.set_accepted_adapter_types([PromptEncoderAdapterConfig._target_])
```
ericharper, line 91: Docstring

examples/nlp/language_modeling/conf/megatron_llama_inference.yaml
blahBlahhhJ: Because llama requires `add_BOS: True` and a special EOS token `end_strings: ["</s>"]`

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1294,7 +1311,11 @@ def on_load_checkpoint(self, checkpoint) -> None:
                     key.replace('model.', ''): checkpoint_state_dict.pop(key)
                     for key in list(checkpoint_state_dict.keys())
                 }
-                module.load_state_dict(checkpoint_state_dict, strict=True)
+                missing_keys, unexpected_keys = module.load_state_dict(checkpoint_state_dict, strict=False)
+
+                assert len(unexpected_keys) == 0, 'Unexpected key(s) in state_dict: {}. '.format(
+                    ', '.join('"{}"'.format(k) for k in unexpected_keys)
+                )
```
ericharper, line 1318: Why do we need this? Doesn't strict=True do the same thing?

examples/nlp/language_modeling/megatron_change_num_partitions.py
blahBlahhhJ: https://github.com/NVIDIA/NeMo/blob/fb87053f62780256029632494228389396ce1fd9/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L774

this line requires scalar if precision is fp16.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
blahBlahhhJ: it's different because this doesn't check `missing_keys`. During checkpoint conversion, all the TE `_extra_state` will be in missing keys and we don't want to error. Unexpected keys, on the other hand, is abnormal behavior so we want them to error.

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -151,21 +190,53 @@ def init_peft_modules(self):
         Randomly initialize the peft params and add them to the appropriate modules.
         """
         assert len(self.peft_name_keys) > 0, "peft_name_keys have not been set no PEFT modules will be added"
+        assert not self.mcore_gpt or hasattr(
+            self, 'name_key_to_mcore_mixins'
+        ), "some of the PEFT modules are not supported in megatron core mode yet."
```
ericharper, line 195: could we print self here so that the error message says which module it is?

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -526,14 +622,20 @@ def __init__(
 
     def tie_weights(self,):
         pos_idx = 0
-        layer0 = self.model.language_model.encoder.layers[0]
+
+        if self.mcore_gpt:
+            layers = self.model.decoder.layers
+        else:
+            layers = self.model.language_model.encoder.layers
```
ericharper, line 629: Do we need to check if self.model is being wrapped here?

examples/nlp/language_modeling/conf/megatron_llama_inference.yaml
aklife97: I am not fully sure if it makes sense to have a new inference file for this here...

@ericharper what do you feel? we have similar issues for T5/HF T5 as well - the way we handle that currently is pretty bad, we just manually know what model file is supposed to use what inference param. But if we keep on creating a new config file for every new community model I am afraid we'll make this folder extremely messy

maybe a middle ground could be to have a new folder, maybe called `community_models` and have all these config files there? I'm still afraid we might end up having a lot of files there, but dont see any better alternative

aklife97: Ideally this info should live inside the .nemo

we have the add_bos_to_input/add_eos_to_input tags in the data config, maybe longer term we need to move that to the model config so that we can directly use it during inference

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -275,7 +290,16 @@ def forward(
...
+        if self.frozen_model.mcore_gpt:
+            output = self.frozen_model.model(
+                input_ids=None,
+                position_ids=None,
+                decoder_input=encoder_input,
+                attention_mask=attention_mask,
+                labels=labels,
+                inference_params=inference_params,
+            )
+        elif self.autocast_dtype == torch.float32:
```
ericharper, line 302: What does autocast_dtype have to do with this? with and without it looks like identical calls to me?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -506,6 +530,12 @@ def on_test_epoch_end(self):
         logging.info(f'test_loss: {averaged_loss[0]}')
         self.test_step_outputs.clear()  # free memory
 
+    def setup(self, stage=None):
+        super().setup(stage)
+
+        if self.frozen_model.cfg.get('transformer_engine', False):
+            self.frozen_model.setup_transformer_engine_tp_groups()
```
ericharper, line 537: I think we need to update this for mcore_gpt. We need to setup those groups when we are using it.

examples/nlp/language_modeling/megatron_change_num_partitions.py
aklife97: this is just a TP/PP conversion script, I am not sure what needs a GradScaler here

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -186,7 +186,7 @@ def setup(self, stage=None):
             self.setup_training_dataloader()
         if hasattr(self, '_validation_ds'):
             self._validation_dl = self.setup_eval_dataloader(self._validation_ds, self.cfg.data.validation_ds)
-        if hasattr(self.cfg.data, 'test_ds'):
+        if hasattr(self.cfg.data, 'test_ds') and self.cfg.data.test_ds.get('file_names', None) is not None:
```
ericharper, line 189: How come we don't have the same check for validation?

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
blahBlahhhJ: done.

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
blahBlahhhJ: done.

nemo/collections/nlp/modules/common/megatron/adapters/mcore_mixins.py
blahBlahhhJ: done.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -197,14 +197,15 @@ def setup(self, stage=None):
...
-                parallel_state.set_virtual_pipeline_model_parallel_rank(0)
-            else:
-                self.model.sync_initial_word_embeddings()
+        if not self.cfg.get('mcore_gpt', False):
+            if parallel_state.get_pipeline_model_parallel_world_size() > 1:
+                if isinstance(self.model, list):
+                    for i, module in enumerate(self.model):
+                        parallel_state.set_virtual_pipeline_model_parallel_rank(i)
+                        module.sync_initial_word_embeddings()
+                    parallel_state.set_virtual_pipeline_model_parallel_rank(0)
```
ericharper, line 206: I'm not sure why this code is here. Do we ever do sft with interleaved?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
```diff
@@ -197,14 +197,15 @@ def setup(self, stage=None):
...
         # when using pipeline model parallel the final stage need to initialize word embeddings
-        if parallel_state.get_pipeline_model_parallel_world_size() > 1:
-            if isinstance(self.model, list):
-                for i, module in enumerate(self.model):
-                    parallel_state.set_virtual_pipeline_model_parallel_rank(i)
-                    module.sync_initial_word_embeddings()
-                parallel_state.set_virtual_pipeline_model_parallel_rank(0)
-            else:
-                self.model.sync_initial_word_embeddings()
+        if not self.cfg.get('mcore_gpt', False):
```
ericharper, line 200: Why wouldn't this code apply to mcore_gpt?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
blahBlahhhJ: yes. I'm wondering the same thing here. this is not my code. I only added the mcore if statement. Should we combine the other 2 cases?

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
blahBlahhhJ: Adi says test set should be optional (hence this check) but val set is not optional.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
blahBlahhhJ: @lhb8125 might be able to answer this.

Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN git clone https://github.com/NVIDIA/apex.git && \
 # install megatron core, this can be removed once 0.3 pip package is released
 RUN git clone https://github.com/NVIDIA/Megatron-LM.git && \
   cd Megatron-LM && \
-  git checkout 84f64880b3651c4f7cf90da337ee4e7d9968acab && \
+  git checkout 1b1a798bca149ecf98d43b6082cd1ee80079eadf && \
```
ericharper, line 58: Which mcore commit is this? Does it contain the dist ckpt fixes?

Jenkinsfile
```diff
@@ -59,10 +59,9 @@ pipeline {
 
     stage('Megatron Core installation') {
       steps {
-        // commit has api fix for TE
         sh 'git clone https://github.com/NVIDIA/Megatron-LM.git && \
             cd Megatron-LM && \
-            git checkout 84f64880b3651c4f7cf90da337ee4e7d9968acab && \
+            git checkout 1b1a798bca149ecf98d43b6082cd1ee80079eadf && \
```
ericharper, line 64: Which mcore commit is this? does it contain the dist ckpt fixes?

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
blahBlahhhJ: you mean
```
if self.cfg.get('transformer_engine', False) or self.cfg.get('mcore_gpt', False):
            self.setup_transformer_engine_tp_groups()
```

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
blahBlahhhJ: done.

Dockerfile
blahBlahhhJ: this is the most recent one https://github.com/NVIDIA/Megatron-LM/commit/1b1a798bca149ecf98d43b6082cd1ee80079eadf

will need to update this once my dist ckpt PR is merged and synced on github.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
blahBlahhhJ: @lhb8125 might be able to answer this.

examples/nlp/language_modeling/megatron_change_num_partitions.py
blahBlahhhJ: the code I point to runs when initializing the model. Thus it will error if user set precision to 16 when doing the conversion without a scalar.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -893,10 +898,22 @@ def fwd_output_only_func(dataloader_iter, model):
                 if attention_mask is not None:
                     attention_mask = attention_mask.cuda()
                     attention_mask = attention_mask[0:1]
+            if self.mcore_gpt:
+                # if first step, then clear KV cache, otherwise reuse inference_paarms
+                if set_inference_key_value_memory[0].item():
+                    self.inference_params = InferenceParams(
```
aklife97, line 904: I don't personally like the name InferenceParam being passed around during model training. But since this is in core already, we can go ahead with it and probably change later if we fee like

aklife97: okay, since this is forward_only, should be good

examples/nlp/language_modeling/megatron_change_num_partitions.py
aklife97: ugh... I see, looks like this moved from fwd_bwd to model init with config update
cool, can fix it outside this PR

nemo/core/connectors/save_restore_connector.py
lhb8125: Since the working dir has changed to `tmpdir `by `os.chdir(tmpdir)`, we don't need to join the `tmpdir `with `model_config_yaml`.
And if a relative dir path of nemo checkpoint is specified in the config, the `config_yaml `will be `tmpdir/model_config_yaml`. However, the path is wrong since we already enter the `tmpdir`.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
lhb8125: The logic is handled by mcore here: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/models/gpt/gpt_model.py#L119

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
lhb8125: The code was already there, so I don't know the reason.

examples/nlp/language_modeling/conf/megatron_llama_inference.yaml
ericharper: There will be a larger effort for managing configs. Let's just make the this config make the most sense for this model for now

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
ericharper: Then do we need this in the model, can it just be in the conversion script?

Dockerfile
blahBlahhhJ: updated mcore with my dist ckpt fix.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
ericharper: Yes let's delete the redundant conditional and just check if mcore or not

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
blahBlahhhJ: done.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
blahBlahhhJ: done.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1476,13 +1517,23 @@ def build_transformer_config(self) -> TransformerConfig:
         recompute_method = self.cfg.get('activations_checkpoint_method', None)
         recompute_num_layers = self.cfg.get('activations_checkpoint_num_layers', None)
 
+        if not self.cfg.get('fp8', False):
```
aklife97, line 1520: why do we want to have all these bools individually listed out?
lets make them like core where fp8 cfg variable is a string that is e4m3 or hybrid

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
blahBlahhhJ: I think checking unexpected keys is a must? Are there any cases where we want to allow unexpected keys? (keys exist in the checkpoint file but not in the actual model)

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
aklife97: cool, can do that outside this PR

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -17,6 +17,12 @@
 from pytorch_lightning.trainer.trainer import Trainer
 
 from nemo.collections.nlp.models.language_modeling.megatron_gpt_sft_model import MegatronGPTSFTModel
+from nemo.collections.nlp.modules.common.megatron.adapters.mcore_mixins import (
```
aklife97, line 20: @arendu you might wanna review these changes

blahBlahhhJ: those are from #7276 which are already reviewed by @arendu

aklife97: awesome

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
```diff
@@ -368,6 +459,16 @@ def load_state_dict(self, state_dict, strict: bool = True):
         else:
             super().load_state_dict(state_dict, strict=True)
 
+    def on_load_checkpoint(self, checkpoint) -> None:
```
aklife97, line 462: why do we need to do this?

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
ericharper: I mean keeping strict = True, and then only in ckpt conversion allowing for missing keys

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -275,15 +290,14 @@ def forward(
...
             output = self.frozen_model.model(
                 input_ids=None,
                 position_ids=None,
-                encoder_input=encoder_input,
+                decoder_input=encoder_input,
                 attention_mask=attention_mask,
                 labels=labels,
-                set_inference_key_value_memory=set_inference_key_value_memory,
-                inference_max_sequence_len=inference_max_sequence_len,
+                inference_params=inference_params,
```
aklife97, line 300: outside this PR, since inference_params were added, we'd want to remove the other inference related keys and just use the config

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
blahBlahhhJ: there's some inconsistency in save/load logic after the distributed checkpoint PR. `on_load_checkpoint()` became the actual method to load checkpoints :( I talked to Eric about this. not sure what's the plan.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
blahBlahhhJ: done.

nemo/collections/nlp/models/language_modeling/megatron_gpt_peft_models.py
ericharper: We put the dist ckpt logic there for now but we'll probably refactor it a bit.

nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py
ericharper: We should probably remove, but not in this PR

ericharper (approved):
LGTM. Thanks!

examples/nlp/language_modeling/conf/megatron_llama_config.yaml
```diff
@@ -0,0 +1,219 @@
+name: megatron_llama
+restore_from_path: null # used when starting from a .nemo file
+
+trainer:
+  devices: 1
+  num_nodes: 1
+  accelerator: gpu
+  precision: 32
```
aklife97, line 8: @blahBlahhhJ @lhb8125 
is it possible to write weights in bf16 O2 at coversion? I'm not familiar with how HF handles weights for llama but the original model was trained with bf16 so would prefer if we can write the model directly in bf16 so we consume less space and can also easily use it directly

blahBlahhhJ: no it converts it into O1 format and doesn't support O2 format. 

That being said, I convert the entire model to the target precision before saving it. so if precision is 16 or bf16, the storage it takes is one half of the full fp32 size. Most of the examples in Nemo can configure amp O2 at initialization so that's also not a problem.

aklife97: > no it converts it into O1 format and doesn't support O2 format.

whats the reason for this? since the original model was trained with bf16

I'm confused with what is happening with precision, since we're converting model from HF, what precision does HF keep the model weights in? are the model weights ever going to fp16/having lossy conversion?

blahBlahhhJ: hf stores the model in fp16. When loading the model, it becomes fp32. and then I re-map the state dict keys into Nemo (still fp32), finally I cast the model to whatever precision set in the command and save it. (So I think technically fp16->fp32->fp16) doesn't lose info.

aklife97: Meta's original checkpoints are in bf16, HF's conversion to fp16 is dangerous...

blahBlahhhJ: idk maybe it's for compatibility. some GPUs don't support bf16.

PR #7318, opened 2023-08-25
Replace prefetch with val iterator check in megatron models

athitten:
# What does this PR do ?

1) Catch the end of `dataloader_iter` to exit gracefully from `validation_step` without having to prefetch  all `microbatches` in a step which was being done previously. This is a temporary piece of code that is needed until we have lightning's fix that takes care of catching the end of `dataloader_iter`
2) Set `limit_val_batches` and `num_sanity_val_steps` for `pretraining` to be a multiple of num of `microbatches`. `limit_val_batches` is reconfigured so that we run as many global batches or validation_steps as the user entered `limit_val_batches`. 

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3157,24 +3157,25 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         sh "rm -rf examples/nlp/language_modeling/retro_results"
       }
     }
-    stage('L2: BioMegatron Bert NER Task') {
-      when {
-        anyOf {
-          branch 'main'
-          changeRequest target: 'main'
-        }
-      }
-      failFast true
-      steps {
-        sh "python examples/nlp/token_classification/token_classification_train.py \
-        exp_manager.exp_dir=examples/nlp/language_modeling/token_classification_results \
-        trainer.max_epochs=1 \
-        model.dataset.data_dir=/home/TestData/nlp/ner \
-        model.language_model.pretrained_model_name=biomegatron345m_biovocab_30k_cased \
-        model.tokenizer.tokenizer_name=null"
-        sh "rm -rf examples/nlp/language_modeling/token_classification_results"
-      }
-    }
+    // Comment temporarily
+    // stage('L2: BioMegatron Bert NER Task') {
+    //   when {
+    //     anyOf {
+    //       branch 'main'
+    //       changeRequest target: 'main'
+    //     }
+    //   }
+    //   failFast true
+    //   steps {
+    //     sh "python examples/nlp/token_classification/token_classification_train.py \
+    //     exp_manager.exp_dir=examples/nlp/language_modeling/token_classification_results \
+    //     trainer.max_epochs=1 \
+    //     model.dataset.data_dir=/home/TestData/nlp/ner \
+    //     model.language_model.pretrained_model_name=biomegatron345m_biovocab_30k_cased \
+    //     model.tokenizer.tokenizer_name=null"
+    //     sh "rm -rf examples/nlp/language_modeling/token_classification_results"
+    //   }
+    // }
     stage('L2: Megatron GPT Pretraining and Resume Training TP=2') {
       when {
         anyOf {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3189,7 +3189,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
-        trainer.limit_val_batches=1 \
+        trainer.limit_val_batches=2 \
         trainer.accumulate_grad_batches=1 \
         trainer.max_steps=3 \
         trainer.precision=16 \
@@ -4652,7 +4652,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
-        trainer.limit_val_batches=1 \
+        trainer.limit_val_batches=5 \
         trainer.accumulate_grad_batches=1 \
         trainer.max_steps=6 \
         trainer.precision=16 \
@@ -4838,7 +4838,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
           steps {
             sh "python examples/nlp/language_modeling/megatron_gpt_pretraining.py \
             trainer.max_steps=10 \
-            trainer.limit_val_batches=1 \
+            trainer.limit_val_batches=7 \
             trainer.val_check_interval=10 \
             exp_manager.exp_dir=examples/nlp/language_modeling/gpt_pretrain_results \
             model.data.data_impl=mock \
@@ -4851,7 +4851,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
           steps {
             sh "python examples/nlp/language_modeling/megatron_t5_pretraining.py \
             trainer.max_steps=10 \
-            trainer.limit_val_batches=1 \
+            trainer.limit_val_batches=3 \
             trainer.val_check_interval=10 \
             exp_manager.exp_dir=examples/nlp/language_modeling/t5_pretrain_results \
             model.data.data_impl=mock \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3189,7 +3189,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
-        trainer.limit_val_batches=2 \
+        trainer.limit_val_batches=1 \
         trainer.accumulate_grad_batches=1 \
         trainer.max_steps=3 \
         trainer.precision=16 \
@@ -4652,6 +4652,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
+        trainer.limit_val_batches=1 \
         trainer.accumulate_grad_batches=1 \
         trainer.max_steps=6 \
         trainer.precision=16 \
@@ -4837,6 +4838,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
           steps {
             sh "python examples/nlp/language_modeling/megatron_gpt_pretraining.py \
             trainer.max_steps=10 \
+            trainer.limit_val_batches=1 \
             trainer.val_check_interval=10 \
             exp_manager.exp_dir=examples/nlp/language_modeling/gpt_pretrain_results \
             model.data.data_impl=mock \
@@ -4849,6 +4851,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
           steps {
             sh "python examples/nlp/language_modeling/megatron_t5_pretraining.py \
             trainer.max_steps=10 \
+            trainer.limit_val_batches=1 \
             trainer.val_check_interval=10 \
             exp_manager.exp_dir=examples/nlp/language_modeling/t5_pretrain_results \
             model.data.data_impl=mock \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3157,25 +3157,24 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         sh "rm -rf examples/nlp/language_modeling/retro_results"
       }
     }
-    // Comment temporarily
-    // stage('L2: BioMegatron Bert NER Task') {
-    //   when {
-    //     anyOf {
-    //       branch 'main'
-    //       changeRequest target: 'main'
-    //     }
-    //   }
-    //   failFast true
-    //   steps {
-    //     sh "python examples/nlp/token_classification/token_classification_train.py \
-    //     exp_manager.exp_dir=examples/nlp/language_modeling/token_classification_results \
-    //     trainer.max_epochs=1 \
-    //     model.dataset.data_dir=/home/TestData/nlp/ner \
-    //     model.language_model.pretrained_model_name=biomegatron345m_biovocab_30k_cased \
-    //     model.tokenizer.tokenizer_name=null"
-    //     sh "rm -rf examples/nlp/language_modeling/token_classification_results"
-    //   }
-    // }
+    stage('L2: BioMegatron Bert NER Task') {
+      when {
+        anyOf {
+          branch 'main'
+          changeRequest target: 'main'
+        }
+      }
+      failFast true
+      steps {
+        sh "python examples/nlp/token_classification/token_classification_train.py \
+        exp_manager.exp_dir=examples/nlp/language_modeling/token_classification_results \
+        trainer.max_epochs=1 \
+        model.dataset.data_dir=/home/TestData/nlp/ner \
+        model.language_model.pretrained_model_name=biomegatron345m_biovocab_30k_cased \
+        model.tokenizer.tokenizer_name=null"
+        sh "rm -rf examples/nlp/language_modeling/token_classification_results"
+      }
+    }
     stage('L2: Megatron GPT Pretraining and Resume Training TP=2') {
       when {
         anyOf {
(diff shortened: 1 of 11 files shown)


nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -190,6 +190,17 @@ def __init__(self, cfg: DictConfig, trainer: Trainer, no_lm_init=True):
...
+
+    def _reconfigure_val_batches(self):
+        """
+        Reconfigure trainer.limit_val_batches for pretraining
+        """
+        # Override limit_val_batches to be a multiple of num microbatches and so there are limit_val_batches//num_micro_batches num of global batches
+        self.trainer.limit_val_batches *= get_num_microbatches()
+        # Override num sanity steps equal to num of microbatches and perform one val_step
+        self.trainer.num_sanity_val_steps = get_num_microbatches()
+
```
athitten, line 203: This method is called only in pretraining models. Not setting `limit_val_batches` to be a multiple of `microbatches` for downstream models as we are passing just 1 batch to `val_step` each time and limit is not required. Val step is run as many times as the user entered `limit_val_batches` in case of downstream tasks. Also leaving the `num_sanity_val_step` for downstream at the default val of 2.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -370,12 +370,12 @@ def optimizer_zero_grad(self, *args, **kwargs):
         return
 
     def validation_step(self, dataloader_iter, batch_idx):
-        # Prefetch the dataloader_iter before fwd_bwd func to avoid PP rank 2 from waiting indefinitely when PP rank 1 reaches the end of dataloader_iter
-        dataloader_iter, done = self._prefetch(dataloader_iter)
-        if done:
-            return
         mode = 'test' if self.trainer.testing else 'val'
-        batch = next(dataloader_iter)
```
athitten, line 378: This was not needed for prompt learning as we are just passing 1 batch.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -805,19 +816,18 @@ def build_model_parallel_config(self) -> ModelParallelConfig:
...
-        for _ in range(num_microbatches):
+        if (
+            self._val_micro_batches_consumed == self.trainer.num_sanity_val_steps
+            or self._val_micro_batches_consumed == self.trainer.limit_val_batches
+        ):
+            self._val_micro_batches_consumed = 0
             try:
-                element = next(iterator)
-                elements.append(element)
+                _ = next(iterator)  # exhausting the iterator so that PTL knows to go to validation_epoch_end
```
athitten, line 829: Lightning requires to hit a `StopIteration` to break out of the evaluation loop, otherwise it keeps running indefinitely. It hits a StopIteration at the (n+1)th step so we check for `StopIteration` after `limit_val_batches` is exhausted.

ericharper (dismissed):
LGTM. Thanks!

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
```diff
@@ -144,6 +144,9 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
 
         self.enc_dec_model.model_type = ModelType.encoder_and_decoder
 
+        # Override limit_val_batches to be a multiple of num microbatches to prevent val_step from exiting in between a step
+        self._reconfigure_val_batches()
```
aklife97, line 148: are we sure no other child class ends up calling this init?

aklife97: there are too many child classes for LM enc dec, so need to ensure we keep an explicit track...

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
```diff
@@ -370,12 +370,12 @@ def optimizer_zero_grad(self, *args, **kwargs):
...
+        # Add try except to catch the end of the iterator and exit
+        try:
```
aklife97, line 375: what about all other models that inherit from enc_dec? T5 etc, are those handled properly? they dont seem to be changed in this PR

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -805,19 +816,18 @@ def build_model_parallel_config(self) -> ModelParallelConfig:
...
-        is called to avoid PP rank 2 from wait indefinitely to get outpits from PP 1
+    def _val_iterator_done(self, iterator):
+        """
+        Check if we reached trainer.limit_val_batches, if so exhaust the iterator to raise a StopIteration and exit validation_step
         """
-        elements = []
-        num_microbatches = get_num_microbatches()
-        for _ in range(num_microbatches):
+        if (
+            self._val_micro_batches_consumed == self.trainer.num_sanity_val_steps
```
aklife97, line 824: is `num_sanity_val_steps` deleted/changed after sanity check is over? 
if it is not, what if `num_sanity_val_steps = 2 `and `self.trainer.limit_val_batches = 4`. inside validation will we end after 2 batches? wouldnt that be incorrect?

athitten: @aklife97 `_val_iterator_done` is called only after we run through all microbatches in a `validation_step`. So in case of regular validation `self._val_micro_batches_consumed` will be same as `self.trainer.num_sanity_val_steps` only if `num_sanity_val_steps == limit_val_batches`. The method never gets called in between for ex when we are done with 2 microbatches and `limit_val_batches=4`. It calls this method after finishing the remaining 2 microbatches.

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
athitten: @aklife97 wrt `MegatronGPTPromptLearningModel`, it's used only in `DialogueGPTClassificationModel` and `DialogueGPTGenerationModel`. Both of them have their own `validation_step` defined that are using batch (not `dataloader_iter`). So we should be good.

aklife97: I meant generally, not specific to `MegatronGPTPromptLearningModel`
are there other models that might be affected?

aklife97: also, all validation_steps should have this try except for StopIteration right?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
athitten: Yes I have checked for that. Wrt `MegatronLMEncoderDecoderModel` it's inherited by `MegatronT5Model` and `MegatronNMTModel`. `MegatronT5Model` calls `super().__init__()` so I haven't added reconfiguring in the T5 model since it's taken care in the parent class.  `MegatronNMTModel` uses batch in its val_step and also has it's own __init__ where we are not doing any reconfiguring.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
aklife97: > _val_iterator_done is called only after we run through all microbatches in a validation_step

do you mean that validation_step is only run once? because we seem to be calling `_val_iterator_done` in the beginning of validation_step which would mean it runs everytime validation_step is entered?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
aklife97: perfect! just wanted to be sure

nemo/collections/nlp/models/language_modeling/megatron_gpt_prompt_learning_model.py
athitten: Yes it's been taken care of either with just `try, except` or with `_val_iterator_done` depending on pre-training or downstream in all megatron models that are using `dataloader_iter`.

aklife97: cool!

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
aklife97: oh btw, what about `MegatronT5FinetuneModel` and its derivatives?
and `MegatronT5PromptLearningModel`?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
athitten: No `validation_step` can be run multiple times. But `self._val_micro_batches_consumed` is incremented by num of microbatches at the end of val_step and is always a multiple of num of microbatches. Let's say `limit_val_batches = 4 ` and `num_microbatches = 4` and `num_sanity_val_steps=2`. So the value of `self._val_micro_batches_consumed` will be 0 initially when `_val_iterator_done` is called, followed by `fwd_bwd_func` of val_step and then `self._val_micro_batches_consumed` is incremented to 4. So it never becomes 2 and will remain a multiple of 4  whenever it enters `_val_iterator_done` next time unless num_microbatches = 2 (and in this case it's okay if it passes the condition `self._val_micro_batches_consumed == self.trainer.num_sanity_val_steps`)

aklife97: but what if 
`num_microbatches = 2 `in the same example?

so `limit_val_batches = 4`, `num_sanity_val_steps=2 `and `num_microbatches = 2,` what will happen now?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
athitten: Yup they are taken care as well.

aklife97: wouldn't the `super.__init__` call in `MegatronT5FinetuneModel` mess things? `MegatronT5FinetuneModel` samples global batches so we _dont_ want it to call `_reconfigure_val_batches`, but it looks like it will

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
athitten: Yes I just realized this case, it'll break. Let me fix.

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
athitten: Yes you are right, this won't break but will end up running num_microbatches times more validation_step than what it is.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
athitten: Fixed this is the latest commit, let me know if it looks okay.

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
athitten: @aklife97 bcoz of this issue so that child classes that don't need reconfiguring don't end up calling it via `__init__` (I had a `BioMegatron BERT` CI test fail bcoz of this. It din't have global_batch_size and couldn't calculate `get_num_microbatches`), I am now calling reconfiguring in `build_train_valid_test_datasets` method of the models that need reconfiguring (basically pretraining models: `MegatronGPTModel`, `MegatronBERTModel`, `MegatronT5Model`). All other models that use `dataloader_iter` are either passing one batch or they have  build_train_valid_test_datasets implemented in the child class where reconfiguring is handled ex: `MegatronLMEncoderDecoderModel` that doesn't have `build_train_valid_test_datasets`. Or another case is, the model needs reconfiguring (ex: `MegatronBARTModel`) but it uses build_train_valid_test_datasets of the parent class where reconfiguring is done. The CI is now passing fully. Let me know if this is okay.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
athitten: I was observing a hang in the CI test bcoz of this issue, after the fix it runs fine.

Jenkinsfile
```diff
@@ -3189,7 +3189,7 @@ assert_frame_equal(training_curve, gt_curve, rtol=1e-3, atol=1e-3)"'''
         trainer.accelerator=gpu \
         trainer.log_every_n_steps=1 \
         trainer.val_check_interval=2 \
-        trainer.limit_val_batches=2 \
+        trainer.limit_val_batches=1 \
```
aklife97, line 3192: can we increase this back to a higher number?
would be good to ensure we have diversity of values that check most cases?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
aklife97: why do we need the if condition to check microbatches consumed if we're going to rely on iterator length?
why not just check for StopIteration and return accordingly? will also save from keeping track of _val_micro_batches_consumed

Jenkinsfile
athitten: Yes ofc, I decreased it back to what it was originally (before PTL 2.0 upgrade) to reduce the time spent on CI tests since we now have limit_val_batches = 1 working okay (with it being made a multiple of num_microbatches inside our code). One caveat with increasing it back to higher number would be the CI run might take longer especially in cases where the default limit_val_batches is 50 in configs since we now do 50 full validation_steps instead of say 50//num_microbatches. Maybe in cases where limit_val_batches is high like 50, do limit_val_batches = 10 or something to save time on CI?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
athitten: > why do we need the if condition to check microbatches consumed if we're going to rely on iterator length? why not just check for StopIteration and return accordingly? will also save from keeping track of _val_micro_batches_consumed

We have to check if next(iter) is available num_microbatches times and since that's not a good idea (this is the prefetching logic we had previously) we want to check it just once when `_val_micro_batches_consumed` is equal to trainer.limit_val_batches or num_sanity_val_steps. Hence we are keeping track of _val_micro_batches_consumed.

athitten: > also are we sure that PTL handles the lengths of the iterator completely correctly? if my validation set has a total of 100 batches, but limit_val_batches is set as 5 - do we have a guarantee that the iterator length will be exactly 5?
> 
> even if there is a single case where this assumption fails/PTL does not handle this properly, we can end up with huge non-trivial bugs, so would be great to double check on this assumption

Yes this has been confirmed with PTL, and they have reviewed this code too. Also, this is being added as temp snippet of code until we have PTL's patch working that'll take care of handling the end of iterator. I plan to work with PTL this week to get that patch in as soon as we can.

aklife97: the current code is equivalent to just the iterator is not empty checking code without any if condition and _val_micro_batches_consumed tracking.

we can skip the if part and the code will still work the same

aklife97: to clarify, here's what I mean: the current code is effectively just doing a stop iteration check. Removing the if-else will not change the functioning. The benefit of removing will be that all models (irrespective of downstream/pre-train) can use this function and check if the iterator is over and not worry about anything else

athitten: I see your point, let me check on this.

athitten: @aklife97 addressed this in the latest commit.

Jenkinsfile
athitten: This has been addressed as well.

aklife97 (approved):
LGTM, thank you!!

@athitten before finally merging this in, it would be great to re-ensure once again that we:
1. call `_val_iterator_done` in all models
2. do `_reconfigure_val_batches` only in the pre-training models

any missing/misconfigured models above can make it super difficult for anyone using the model to figure out what might be wrong

Overall, looks amazing!

Jenkinsfile
aklife97: thanks!

aklife97 (approved):

ericharper (approved):
LGTM. Thanks!

PR #7321, opened 2023-08-25
[TTS] Add additional config to preprocess_text and compute_feature_stats

rlangman:
# What does this PR do ?

Add some configuration to TTS preprocessing to make them more usable.

Most important, adding `batch_size` to text normalization multi-processing which I have found to make it about 10x faster than using the default 'auto' batch size in joblib. I do not fully understand why, but the documentation indicates that large batch sizes work for small, fast tasks:

> batch_size: int or 'auto', default: 'auto'
> The number of atomic tasks to dispatch at once to each
> worker. When individual evaluations are very fast, dispatching
> calls to workers can be slower than sequential computation because
> of the overhead. Batching fast computations together can mitigate
> this.

**Collection**: [TTS]

# Changelog 
- Add `batch_size` parameter to `preprocess_text.py`
- Add input and output field parameters to `preprocess_text.py`
- Fix boolean flag parsing for `lower_case` for `preprocess_text.py`
- Modify `compute_feature_stats.py` so that it can take a list of manifests, making it easy to compute pitch/energy stats across multiple datasets.

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/scripts/dataset_processing/tts/preprocess_text.py
+++ b/scripts/dataset_processing/tts/preprocess_text.py
@@ -72,7 +72,9 @@ def get_args():
     parser.add_argument(
         "--num_workers", default=1, type=int, help="Number of parallel threads to use. If -1 all CPUs are used."
     )
-    parser.add_argument("--batch_size", type=int, help="Batch size for jobs. Defaults to 'auto' if not provided.")
+    parser.add_argument(
+        "--joblib_batch_size", type=int, help="Batch size for joblib workers. Defaults to 'auto' if not provided."
+    )
     parser.add_argument(
         "--max_entries", default=0, type=int, help="If provided, maximum number of entries in the manifest to process."
     )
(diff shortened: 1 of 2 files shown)


scripts/dataset_processing/tts/preprocess_text.py
```diff
@@ -94,8 +109,11 @@ def main():
 
     input_manifest_path = args.input_manifest
     output_manifest_path = args.output_manifest
+    text_key = args.text_key
+    normalized_text_key = args.normalized_text_key
     lower_case = args.lower_case
     num_workers = args.num_workers
+    batch_size = args.batch_size
```
racoiaws, line 116: Could this be renamed to something like `joblib_batch_size`?

`batch_size` is too ubiquitous in deep learning and it instantly makes me worry about GPU memory etc etc when seen in scripts like this

racoiaws:
Otherwise LGTM

scripts/dataset_processing/tts/preprocess_text.py
rlangman: Sure, makes sense to me.

racoiaws (approved):

PR #7326, opened 2023-08-25
enable selective unfreeze

arendu:
# What does this PR do ?

allow for selective unfreeze of PEFT params. Useful for continued training of PEFT.

**Collection**: [NLP]

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/common/parts/adapter_modules.py
```diff
@@ -61,6 +61,10 @@ def get_default_strategy_config(self) -> 'dataclass':
         """
         return adapter_mixin_strategies.ResidualAddAdapterStrategyConfig()
 
+    def unfreeze(self,):
```
titu1994, line 64: Should be reverted. This is a mixin class - and if you inherit a NeuralModule - it will have a multiple inheritance diamond type inheritance. 

If you need unfreeze then you should inherit NeuralModule instead of torch.nn.Module (its almost exactly the same thing with utility methods)

titu1994 (changes requested):

titu1994:
This approach works well. Just make sure to inherit NeuralModule and revert the unfreeze from the mixin class

nemo/collections/common/parts/adapter_modules.py
arendu: fair point, but what about adapters like `MultiHeadAttentionAdapter` which do not inherit NeuralModel this call from `adapter_mixins` will fail `module.adapter_layer[name].unfreeze()`

nemo/collections/common/parts/adapter_modules.py
```diff
@@ -61,6 +61,13 @@ def get_default_strategy_config(self) -> 'dataclass':
         """
         return adapter_mixin_strategies.ResidualAddAdapterStrategyConfig()
 
+    def adapter_unfreeze(self,):
+        """
+        Sets the requires grad for all parameters in the adapter to True
```
titu1994, line 66: Hmm if this is intended to be for adapters, let's do two things - 
1) add the equivalent freeze_adapter()
2) add docstring that this should be overriden for custom adapter freezing and unfreezing instead of neural_module.freeze/unfreeze.

arendu: sg!

nemo/collections/common/parts/adapter_modules.py
```diff
@@ -61,6 +61,24 @@ def get_default_strategy_config(self) -> 'dataclass':
...
+        """
+        Sets the requires grad for all parameters in the adapter to True.
+        This method should be overridden for any custom unfreeze behavior that is required.
+        For example, if not all params of the adapter should be unfrozen.
+        """
+        for param in self.parameters():
+            param.requires_grad_(True)
+        self.train()
+    
+    def adapter_freeze(self,):
```
arendu, line 74: @titu1994 added adapter_freeze and adapter_unfreeze hooks in the adapter mixin per your suggestion. `adapter_unfreeze` is called internally when `unfreeze_enabled_adapters` is called. but there is no equivalent call for `freeze_enabeld_adapters` in the mixin -- So currently `adapter_freeze` will not be triggered at all.

titu1994: Its ok, its for mirrored api of freeze and unfreeze of adapter. For freeze_enabled_adapters, you can do model.unfreeze() so that's ok.

nemo/collections/common/parts/adapter_modules.py
```diff
@@ -61,6 +61,24 @@ def get_default_strategy_config(self) -> 'dataclass':
...
+        self.train()
+
+    def adapter_freeze(self,):
+        """
+        Sets the requires grad for all parameters in the adapter to False.
+        This method should be overridden for any custom freeze behavior that is required.
+        """
+        for param in self.parameters():
+            param.requires_grad_(False)
+
```
titu1994, line 81: Should have self.eval()

titu1994 (approved):

PR #7384, opened 2023-09-07
[TTS] Added a callback for logging initial data

anteju:
# What does this PR do ?

This PR adds an option to log artifacts before fit starts, for log input data only once.

**Collection**: TTS

# Changelog 
- Adds `initial_log` to `ArtifactGenerator::generate_artifacts`
- Adds `_log_artifacts` to `LoggingCallback`
- Adds `on_fit_start` to `LoggingCallback`

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/tts/parts/utils/callbacks.py
+++ b/nemo/collections/tts/parts/utils/callbacks.py
@@ -110,7 +110,7 @@ def create_id(filepath: Path) -> str:
 class ArtifactGenerator(ABC):
     @abstractmethod
     def generate_artifacts(
-        self, model: LightningModule, batch_dict: Dict
+        self, model: LightningModule, batch_dict: Dict, initial_log: bool = False
     ) -> Tuple[List[AudioArtifact], List[ImageArtifact]]:
         """
         Create artifacts for the input model and test batch.
@@ -118,6 +118,8 @@ class ArtifactGenerator(ABC):
         Args:
             model: Model instance being trained to use for inference.
             batch_dict: Test batch to generate artifacts for.
+            initial_log: Flag to denote if this is the initial log, can
+                         be used to save ground-truth data only once.
 
         Returns:
             List of audio and image artifacts to log.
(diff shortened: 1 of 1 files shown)


rlangman (approved):
LGTM

nemo/collections/tts/parts/utils/callbacks.py
```diff
@@ -214,17 +216,48 @@ def _log_image(self, image: ImageArtifact, log_dir: Path, step: int):
...
+                audio_list += audio
+                image_list += images
+
+        if len(audio_list) == len(image_list) == 0:
+            logging.debug('List are empty, no initial artifacts to log.')
+            return
+
+        log_dir = self.output_dir / f"initial" if self.output_dir else None
+
+        self._log_artifacts(audio_list=audio_list, image_list=image_list, log_dir=log_dir)
```
bzp83, line 252: wasn't this supposed to forward `global_step=model.global_step`?

nemo/collections/tts/parts/utils/callbacks.py
```diff
@@ -237,11 +270,13 @@ def on_train_epoch_end(self, trainer: Trainer, model: LightningModule):
...
-            self._log_audio(audio=audio, log_dir=log_dir, step=model.global_step)
+        if len(audio_list) == len(image_list) == 0:
+            logging.debug('List are empty, no artifacts to log at epoch %d.', epoch)
+            return
 
-        for image in image_list:
-            self._log_image(image=image, log_dir=log_dir, step=model.global_step)
+        log_dir = self.output_dir / f"epoch_{epoch}" if self.output_dir else None
+
+        self._log_artifacts(audio_list=audio_list, image_list=image_list, log_dir=log_dir)
```
bzp83, line 279: same here...

nemo/collections/tts/parts/utils/callbacks.py
rlangman: This is a pretty old PR, but I think the intention was for `on_fit_start` to only log at the start of training on the first epoch, in which case the global step would always be 0.

Though, `on_fit_start` is actually called at the beginning of each training run, even if you are resuming from a checkpoint. In that case it could make sense to pass through the global step. It might also be more correct to skip the logic in `on_fit_start` in that situation.

Issue #7383, opened 2023-09-06
megatron_change_num_partitions.py does not seem to handle `dense_h_to_4h.bias`


odelalleau:
**Describe the bug**

In megatron_change_num_partitions.py  there is dedicated code to handle splitting some layers with fast GLU activations:
* https://github.com/NVIDIA/NeMo/blob/c5ef26cafa935f9966ab3f074248913aef7fa509/examples/nlp/language_modeling/megatron_change_num_partitions.py#L236
* https://github.com/NVIDIA/NeMo/blob/c5ef26cafa935f9966ab3f074248913aef7fa509/examples/nlp/language_modeling/megatron_change_num_partitions.py#L278

However, this code seems to only handle the `.weight` part, not the `.bias` part. And it seems like the bias would also require some special handling: https://github.com/NVIDIA/NeMo/blob/8508e07a20edbfac6e37fdf252c036f4e2215654/nemo/collections/nlp/modules/common/megatron/mlp.py#L220

**Steps/Code to reproduce bug**

I haven't actually tried to trigger a bug because of this -- I am opening this issue because I suspect there is a bug.

**Expected behavior**

I would have expected the `.bias` to be handled as well (or a `NotImplementedError` exception if it is known that this is not supported yet).

**Additional context**

Tagging people who dealt with this code: @titu1994 / @hsiehjackson

titu1994:
Indeed this is a bug. Ive sent PR #7386 for a fix. 
While NeMo models don't use bias, Llama models do.

PR #7386, opened 2023-09-07
Add support for bias conversion in Swiglu models

titu1994:
# What does this PR do ?

Adds support for bias conversion in TP PP Script for swiglu supported modls

**Collection**: [NLP]

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/nlp/language_modeling/megatron_change_num_partitions.py
+++ b/examples/nlp/language_modeling/megatron_change_num_partitions.py
@@ -1410,12 +1410,13 @@ def main():
                                 f.write(extracted_file.read())
 
                 if tokenizer_model_path is None:
-                    raise ValueError("Could not extract tokenizer model file from checkpoint.")
+                    logging.warning("Could not extract tokenizer model file from checkpoint.")
 
-                # Extract tokenizer info
-                with open_dict(model.cfg):
-                    model.cfg.tokenizer.model = tokenizer_model_path
-                    logging.info(f"Tokenizer model path: {model.cfg.tokenizer.model}")
+                else:
+                    # Extract tokenizer info
+                    with open_dict(model.cfg):
+                        model.cfg.tokenizer.model = tokenizer_model_path
+                        logging.info(f"Tokenizer model path: {model.cfg.tokenizer.model}")
 
             model.cfg, restore_dict = force_cpu_model(model.cfg)
 
--- a/examples/nlp/language_modeling/megatron_change_num_partitions.py
+++ b/examples/nlp/language_modeling/megatron_change_num_partitions.py
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 import os
+import shutil
 import tarfile
 import tempfile
-import shutil
 from argparse import ArgumentParser
 from typing import Dict, List
 
--- a/examples/nlp/language_modeling/megatron_change_num_partitions.py
+++ b/examples/nlp/language_modeling/megatron_change_num_partitions.py
@@ -233,7 +233,7 @@ def compute_tp_splits(
             for i in range(tp_size):
                 tp_qkv = torch.cat([tp_qkv_splits[item] for item in range(i, tp_size * 2, tp_size)])
                 split.append(tp_qkv)
-        elif ('dense_h_to_4h.weight' in param_name or 'linear_fc1.weight' in param_name) and fast_glu_activation:
+        elif ('dense_h_to_4h' in param_name or 'linear_fc1' in param_name) and fast_glu_activation:
             # For Megatron GPT model with Fast Glu activation
             # Handle gated linear units
             # concat all the first halves ('W's) and all the second halves ('V's)
@@ -275,7 +275,7 @@ def compute_tp_merge(idx, name, param, partitions_pp, model_cfg):
         concated = torch.cat([partitions_pp[i][idx].data for i in range(len(partitions_pp))], dim=0)
 
     # Logic for Fast Glu activation
-    if 'dense_h_to_4h.weight' in name and fast_glu_activation:
+    if 'dense_h_to_4h' in name and fast_glu_activation:
         # concat all the first halves ('W's) and all the second halves ('V's)
         wk_splits = []
         for tpr in range(len(partitions_pp)):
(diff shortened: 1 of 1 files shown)


odelalleau (commented):
Looks good to me, but do you know why `linear_fc1` only appears in the code that does the merge and not in the code that does the split?

(I realize it was the case before this PR, but it seems a bit weird at first glance)

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -1384,6 +1385,38 @@ def main():
                 with open_dict(model.cfg):
                     model.cfg.tokenizer.model = args.tokenizer_model_path
 
+            else:
+                # Extract all artifacts that are not binary files
+                # and save them to the model directory
```
odelalleau, line 1390: What's the motivation for this change? This seems unrelated to #7383

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -1384,6 +1385,38 @@ def main():
...
+                                continue
+
+                            extracted_file = tar.extractfile(member)
+                            extracted_file_path = os.path.join(temp_dir, member.name)
+
+                            if '.model' in member.name:
+                                tokenizer_model_path = extracted_file_path
+
+                            with open(extracted_file_path, "wb") as f:
+                                f.write(extracted_file.read())
```
odelalleau, line 1410: Besides the tokenizer ".model" file, are there other files required by the tokenizer that need to be extracted? (at first glance it seems like we're not using any of the other extracted files)

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -1384,6 +1385,38 @@ def main():
                 with open_dict(model.cfg):
                     model.cfg.tokenizer.model = args.tokenizer_model_path
 
+            else:
+                # Extract all artifacts that are not binary files
+                # and save them to the model directory
+                logging.info("Extracting artifact from NeMo file...")
```
odelalleau, line 1391: Could this logic be moved outside of the `for pp_rank` loop so that it's run only once?

(also, ideally what comes after the extraction code would be in a `try...finally` block so that `temp_dir` is always deleted on exit)

titu1994: Not really. This logic is shared for all TP PP and must execute in a loop.
It should be fast. The extraction only works on the base level of the nemo file and does not recurse the actual checkpoint dirs.

examples/nlp/language_modeling/megatron_change_num_partitions.py
odelalleau: In that case, if only the ".model" file is needed, then why not only extract this file?

titu1994: There are models which dont have it, or have multiple. This is a fallback to not require the tokenizer all the time.

odelalleau: Sorry, I'm not sure I understand:
* If there's no ".model" file, you may end up extracting a bunch of files that will never be used
* If there are multiple ".model" files, you end up using only the last one (which I guess is an arbitrary choice) and extracting a bunch of other files that will never be used

Why not just extract at most one ".model" file and stop there? (skipping over anything that's not ".model")

titu1994: This only extracts at the root directory, not at checkpoint level.  So the cost of extraction is trivial and order of kb or mb for huge Tokenizer. 

For either of those cases, if it has no Tokenizer it will fail when calling the constructor of the model - it always needs one tokenizer. For multiple tokenizer - this code path isn't supported at the moment. It's arbitrary yes, and will still fail when loading the constructor because such models require two Tokenizers and youre providing just one.

examples/nlp/language_modeling/megatron_change_num_partitions.py
titu1994: No. Simply calling tarfile open does not untar the file - it only opens a object that can traverse the tar object, not it's contents inside it. You need to explicitly call member.extractfile() to get the file pointer to the actual object

examples/nlp/language_modeling/megatron_change_num_partitions.py
odelalleau: Thanks for the clarification, but even if it's cheap, why extract multiple files when it's trivial to extract only the one you need? Note that (with my admittedly limited knowledge of NeMo) I see no reason why future models may not include very large files other than .ckpt or .pt...

examples/nlp/language_modeling/megatron_change_num_partitions.py
odelalleau: I quoted that line to show that it was the same file, the actual extraction occurs below in the code. My point is that the same file(s) are extracted for each PP rank which seems pointless, it would be cleaner to do this only once before the `for` loop.

examples/nlp/language_modeling/megatron_change_num_partitions.py
```diff
@@ -978,6 +978,30 @@ def main():
     if vp_size > 1:
         set_virtual_parallel_rank_safely(0)
 
+    # Extract tokenizer artifact from the model to temp directory
+    logging.info("Extracting tokenizer artifact from NeMo file...")
```
odelalleau, line 982: Should probably be inside an
```python
if args.tokenizer_model_path is None:
```

odelalleau (commented):
Thanks for the refactor! I think it's easier to follow the logic now :)

NB: you could delete `temp_dir` afterwards (not sure it's worth wrapping everything in a `try... finally` since it would cause an extra indentation, so just deleting it at the end would probably be enough)

odelalleau (approved):
Looks good to merge!

ericharper (approved):
LGTM. Thanks!

PR #7399, opened 2023-09-08
Implement bucket_cap_mb automatic search for GPT model

shjwudp:
# What does this PR do ?

The optimizer `apex.distributed_fused_adam` is a high-performance adam optimizer implementation. It provides a bucket function to update the gradient aggregation to improve performance. However, bad bucket configuration can lead to GB-level memory loss. This PR realizes the automatic search for the optimal bucket size for GPT model, maximizing the bucket size while minimizing memory loss, thereby maintaining excellent optimizer performance.

**Collection**: NLP

# Changelog 
- add bucket_cap_mb automatic search function
- add unit test

# Usage

When using distributed_fused_adam, you can configure bucket_cap_mb to auto to enable this function.

```yaml
...
  optim:
    name: distributed_fused_adam
    lr: 0.0001
    weight_decay: 0.1
    betas:
    - 0.9
    - 0.95
    bucket_cap_mb: auto
    overlap_grad_sync: true
    overlap_param_sync: true
    contiguous_grad_buffer: true
    sched:
      name: CosineAnnealing
      warmup_steps: 107
      constant_steps: 11873
      min_lr: 1.0e-05
...
```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

Help me review this PR, thank you! @MaximumEntropy, @ericharper, @ekmb, @yzhang123, @VahidooX, @vladgets, @okuchaiev


timmoon10 (commented):
I think auto-detecting the bucket size would be a useful feature, but I think this heuristic can be improved. As far as I am aware, the considerations for bucket size are:
- Larger bucket sizes improve performance (less kernel overhead and NCCL latency overhead)
- There is wasted memory and communication if the bucket sizes don't match the param sizes
- Param sizes and bucket sizes are padded to 128B to improve alignment
- The bucket size should be smaller than existing buffers in the caching memory allocator, e.g. from temporary activation tensors. Otherwise the allocator will need to allocate new buffers for temporary distopt buffers (e.g. for all-gathers).

This last point is the hardest to handle automatically, and I remember it causing some trouble with T5. If we wanted to make this more robust, we could estimate the activation size and use that as an upper bound instead of 500 MB.

From an API perspective, I think we should also keep in mind that not all models explicitly specify buckets and that non-language models call the distributed optimizer. I think we would want the [wrapper class](https://github.com/NVIDIA/NeMo/blob/5cb76a53f21d7470773a80d62fd513defa24ddbf/nemo/core/optim/distributed_adam.py#L67) to complain if it gets a `str` for `bucket_cap_mb`, and maybe to then fall back to the 100 MB default.

shjwudp:
> * The bucket size should be smaller than existing buffers in the caching memory allocator, e.g. from temporary activation tensors. Otherwise the allocator will need to allocate new buffers for temporary distopt buffers (e.g. for all-gathers).

Hi, Tim. Thank you for taking the time to review my work and for providing your expert advice. I appreciate your insights, and I'm preparing to make the necessary modifications to the PR as per your suggestions.

I do have a query regarding one of your recommendations. You suggested aligning according to 128B, and I was wondering if you could provide some clarification on this. Is this alignment a specific feature of distopt, or is it a common practice for most CUDA kernels?

shjwudp:
@timmoon10 I have refactor the strategy for automatically detecting bucket size, and have now incorporated downgradeability for models that are not yet supported. Would you be interested in reviewing this updated approach?

timmoon10:
Sorry for the delay. Clarifying my fourth point, suppose our training loop looks like:
```
x = get_data()
y = layer1(x)
loss = layer2(y)
loss.backward()
optim.step()
```
If `y` takes up 100 MB, then PyTorch's memory allocator will have a free 100 MB buffer sitting around during the optimizer step. Ideally we want the bucket size to be <=100 MB so we don't need to allocate any extra memory for the distributed optimizer's temporary buffers.

I've made some modifications to this PR at https://github.com/timmoon10/nemo/tree/auto_find_optimizer_bucket_size. This takes the activation sizes into account. Also, the user has to explicitly specify `bucket_cap_mb: auto` to enable this and there are warnings if the model doesn't have the required configuration. When I run with some small models, the wasted bucket space is much smaller, although I haven't looked into the actual memory usage yet.

shjwudp:
Hi @timmoon10 , glad to see you come back to this question. I really want to discuss the fourth point with you. This condition can lead to poor performance in some cases. For a reasonable activation estimate, 2sbh, consider the commonly used in LLM training. bf16 activation, for llama 7b, s=2k, b=1, h=4k, this value is only 16MB. This upper bound is smaller than our commonly used 125MB and will cause poor optimizer performance. Would releasing restrictions be a good option in this case?

In addition, can we develop directly based on your code?

shjwudp:
> That's a good point. I'm a little uneasy with my memory heuristic as well since it's so uncertain what tensors will create the largest memory blocks (attention logits? feed-forward network activations?). I think a better API might be something like:
> 
> ```
> bucket_cap_mb: 125  # user-specified, should match model
> tune_bucket_cap_mb: True  # reduce bucket_cap_mb to minimize waste
> ```

LGTM!

PR #7413, opened 2023-09-11
Allow disabling sanity checking when num_sanity_val_steps=0

athitten:
# What does this PR do ?

Modifies the `_reconfigure_val_batches` method in `MegatronBaseModel` to perform sanity steps equal to num of microbatches only when `num_sanity_val_steps` is not 0 so that user has the option to disable sanity checking by passing `num_sanity_val_steps=0`

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -211,8 +211,9 @@ def _reconfigure_val_batches(self):
         """
         # Override limit_val_batches to be a multiple of num microbatches and so there are limit_val_batches//num_micro_batches num of global batches
         self.trainer.limit_val_batches *= get_num_microbatches()
-        # Override num sanity steps equal to num of microbatches and perform one val_step
-        self.trainer.num_sanity_val_steps = get_num_microbatches()
+        # Override num sanity steps equal to num of microbatches and perform one val_step if num_sanity_val_steps is not 0
+        if self.trainer.num_sanity_val_steps:
+            self.trainer.num_sanity_val_steps = get_num_microbatches()
```
aklife97, line 216: this should also be *= like limit val batches since everything is measured in global batches

aklife97: we should not ignore num_sanity_val_steps, we use it nemo side in a lot of use-cases like doing validation before training/tuning on a finetuned model

we should rather set it to:
`self.trainer.num_sanity_val_steps *= get_num_microbatches()`

athitten: Yeah we can have that, and avoid the if condition check. Let me modify it. Initially with `self.trainer.num_sanity_val_steps = get_num_microbatches()` my intention was to do just 1 sanity `validation_step`. But yeah we can keep it as `self.trainer.num_sanity_val_steps *= get_num_microbatches()` to be in sync with limit_val_batches.

athitten: Fixed.

aklife97 (approved):
LGTM, thank you!

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
aklife97: qq: if num_sanity_val_steps is not set, what is the default value? 
just want to ensure we dont get a multiplying by None error

PR #7455, opened 2023-09-18
Fix None dataloader issue in PTL2.0

KunalDhawan:
# What does this PR do ?

With PTL2.0, returning None from the *_dataloader methods is no longer supported (reference - https://github.com/Lightning-AI/lightning/issues/17154). This change causes an issue with ASR decoding when we don't provide a valid trainer.validation_ds.manifest_filepath or trainer.test_ds.manifest_filepath. For example running the following code:
`python NeMo/examples/asr/asr_ctc/speech_to_text_ctc_bpe.py --config-path=../conf/citrinet/ --config-name=citrinet_1024.yaml model.tokenizer.dir=data/tokenizer/tokenizer_bpe_v1024/ model.tokenizer.type="bpe" model.train_ds.manifest_filepath=data/LibriSpeech/train_clean_100.json model.validation_ds.manifest_filepath=data/LibriSpeech/dev_clean.json model.train_ds.batch_size=16 +accelerator="gpu" +strategy="ddp" +devices=1 trainer.max_epochs=1 exp_manager.checkpoint_callback_params.monitor="train_loss"`

throws the error - 
`TypeError: An invalid dataloader was returned from EncDecCTCModelBPE.val_dataloader(). Found None.`

The following PR fixes this issue by making the default return value of val_dataloader and test_dataloader as `[]`

**Collection**: All

# Changelog 

```
  def val_dataloader(self):
      if self._validation_dl is not None:
          return self._validation_dl
      else: # None dataloader no longer supported in PTL2.0
          return []

  def test_dataloader(self):
      if self._test_dl is not None:
          return self._test_dl
      else: # None dataloader no longer supported in PTL2.0
          return []
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -866,9 +866,9 @@ class ModelPT(LightningModule, Model):
         if self._test_dl is None:
             # None dataloader no longer supported in PTL2.0
             self._test_dl = []
-            
+
         return self._test_dl
-    
+
     def on_validation_epoch_end(self) -> Optional[Dict[str, Dict[str, torch.Tensor]]]:
         """
         Default DataLoader for Validation set which automatically supports multiple data loaders
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -858,10 +858,14 @@ class ModelPT(LightningModule, Model):
     def val_dataloader(self):
         if self._validation_dl is not None:
             return self._validation_dl
+        else: # None dataloader no longer supported in PTL2.0
+            return []
 
     def test_dataloader(self):
         if self._test_dl is not None:
             return self._test_dl
+        else: # None dataloader no longer supported in PTL2.0
+            return []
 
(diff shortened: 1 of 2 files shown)


nemo/core/classes/modelPT.py
```diff
@@ -858,10 +858,14 @@ def train_dataloader(self):
     def val_dataloader(self):
         if self._validation_dl is not None:
             return self._validation_dl
+        else: # None dataloader no longer supported in PTL2.0
```
titu1994: Set self._validation_dl to [] as well

titu1994 (dismissed):
Thanks for the fix !

nemo/core/classes/modelPT.py
KunalDhawan: Fixed

titu1994 (approved):
Thanks !

PR #7485, opened 2023-09-22
Safeguard nemo_text_processing installation on ARM

blisc:
# What does this PR do ?

Removes nemo_text_processing as a requirement on ARM.

Some scripts will fail on ARM due to missing nemo_text_processing

**Collection**: TTS

# Changelog 
- removes nemo_text_processing as a requirement on arm machines

# Usage
N/A

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
+++ b/examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
@@ -61,7 +61,7 @@ from nemo.utils import logging
 
 try:
     from nemo_text_processing.text_normalization.data_loader_utils import post_process_punct
-except ModuleNotFoundError:
+except (ImportError, ModuleNotFoundError):
     raise ModuleNotFoundError(
         "The package `nemo_text_processing` was not installed in this environment. Please refer to"
         " https://github.com/NVIDIA/NeMo-text-processing and install this package before using "
--- a/examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
+++ b/examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
@@ -50,7 +50,6 @@ import os
 from typing import List
 
 from helpers import DECODER_MODEL, TAGGER_MODEL, instantiate_model_and_trainer
-from nemo_text_processing.text_normalization.data_loader_utils import post_process_punct
 from nn_wfst.en.electronic.normalize import ElectronicNormalizer
 from nn_wfst.en.whitelist.normalize import WhitelistNormalizer
 from omegaconf import DictConfig, OmegaConf
@@ -59,6 +58,15 @@ from nemo.collections.nlp.data.text_normalization import constants
 from nemo.collections.nlp.models import DuplexTextNormalizationModel
 from nemo.core.config import hydra_runner
 from nemo.utils import logging
+
+try:
+    from nemo_text_processing.text_normalization.data_loader_utils import post_process_punct
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
+        "The package `nemo_text_processing` was not installed in this environment. Please refer to"
+        " https://github.com/NVIDIA/NeMo-text-processing and install this package before using "
+        "this script"
+    )
 
 
 @hydra_runner(config_path="conf", config_name="duplex_tn_config")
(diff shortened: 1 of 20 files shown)


requirements/requirements_tts.txt
```diff
@@ -4,7 +4,8 @@ jieba
 kornia
 librosa
 matplotlib
-nemo_text_processing
+# pynini does not currently support aarch, disable nemo_text_processing for now
+nemo_text_processing; 'arm' not in platform_machine and 'aarch' not in platform_machine
```
blisc, line 8: should we move this up to requirements or leave it in tts? I see it's used in

- examples/nlp
- NeMo/nemo/collections/asr/data/text_to_text.py
- NeMo/nemo/collections/common/tokenizers/text_to_speech/ipa_lexicon.py
- NeMo/nemo/collections/nlp/models/machine_translation/mt_enc_dec_model.py

and others. This is not a complete list

examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
```diff
@@ -60,6 +59,15 @@
 from nemo.core.config import hydra_runner
 from nemo.utils import logging
 
+try:
+    from nemo_text_processing.text_normalization.data_loader_utils import post_process_punct
+except ModuleNotFoundError:
+    raise ModuleNotFoundError(
```
titu1994, line 65: Dont raise at runtime. It will not fix anything cause we have multiple `__init__` that imports from these files and so this path will be called. i.e. - even if its unused, import will still raise this error.


Do the delayed check please - 
First create a bool - https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/losses/rnnt.py#L53-L58 
Next check bool at runtime - https://github.com/NVIDIA/NeMo/blob/2cc09425aba3e9b3cfdba43a3188eaef58227055/nemo/collections/asr/losses/rnnt.py#L143-L150

Or at init - https://github.com/NVIDIA/NeMo/blob/2cc09425aba3e9b3cfdba43a3188eaef58227055/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L85-L88

examples/nlp/duplex_text_normalization/nn_wfst/en/electronic/normalize.py
```diff
@@ -12,16 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nemo_text_processing.text_normalization.normalize import Normalizer
-from nemo_text_processing.text_normalization.token_parser import TokenParser
+try:
+    from nemo_text_processing.text_normalization.normalize import Normalizer
```
titu1994, line 16: For class imports that are used in-file like this, use the dummy class here - https://github.com/NVIDIA/NeMo/blob/2cc09425aba3e9b3cfdba43a3188eaef58227055/nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py#L36

Class definition - https://github.com/NVIDIA/NeMo/blob/2cc09425aba3e9b3cfdba43a3188eaef58227055/nemo/collections/nlp/modules/common/megatron/utils.py#L36-L45

requirements/requirements_tts.txt
titu1994: Lets guard all of this at their init and remove nemo text processing install entirely. Unless you are sure this `'arm' not in platform_machine and 'aarch' not in platform_machine` thing actually works?

titu1994: I can push the changes to the rest of the files if you want.

examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
titu1994: You can check this by removing the nemo text processing from the requirements, all these imports will raise error

requirements/requirements_tts.txt
blisc: I would prefer to keep the nemo text processing install.

examples/nlp/duplex_text_normalization/duplex_text_normalization_infer.py
blisc: I don't think this is necessary. All the files that are touched in this PR are under either examples, scripts, or tutorials. None of these files should be imported in the NeMo library.

titu1994: But this doesn't actually fix the issue - you still have NeMo toolkit in requirements, so pip install all still can't be done when pynini is not available

requirements/requirements_tts.txt
titu1994: We should either remove it, or test that the arm and aarch stuff works. If it works, ok to move it to commons, but its fine if its in tts too since all the text processing accesses are guarded inside nemo/collections

titu1994 (approved):
Thanks !

PR #7490, opened 2023-09-22
Fix errors caused by different dimensionality in get_dist function

redoctopus:
# What does this PR do ?

The TTS Aligner's `get_dist()` now has an extra dimension (from PR #6806); this PR fixes errors stemming from this in the Aligner notebook.

@rlangman -- was the change from returning a (B, C, T)-dimensional tensor to a  (B, 1, C, T)-dimensional one intentional? Do we want to keep that or change it back by adding a squeeze(1)?

**Collection**: TTS
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


XuesongYang:
this is a bugfix for r1.21.0. Please change the base.

rlangman:
> @rlangman -- was the change from returning a (B, C, T)-dimensional tensor to a (B, 1, C, T)-dimensional one intentional? Do we want to keep that or change it back by adding a squeeze(1)?

That was unintentional. I would change it back to have the squeeze(1) in get_dist.

redoctopus:
> That was unintentional. I would change it back to have the squeeze(1) in get_dist. Thanks for finding that.

Gotcha, will open a new PR with this fix (and on top of `r1.12.0`, whoops).

Issue #7460, opened 2023-09-19
The moving of files in `exp_manager` may cause crashes in other processes


odelalleau:
**Describe the bug**

`exp_manager()` calls `check_resume()` which moves current files to a `run_x` folder ([link to code](https://github.com/NVIDIA/NeMo/blob/cfb1f449831a9a0309aab71187e3192dbd26d1a7/nemo/utils/exp_manager.py#L630))

However, this may cause issues due to these files being accessed by other processes, here's an example of a stack trace:
```
  File "my_script.py", line 94, in main
    trainer.fit(model)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 532, in fit
    call._call_and_handle_interrupt(
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/call.py", line 42, in _call_and_handle_interrupt
    return trainer.strategy.launcher.launch(trainer_fn, *args, trainer=trainer, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/strategies/launchers/subprocess_script.py", line 93, in launch
    return function(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 565, in _fit_impl
    ckpt_path = self._checkpoint_connector._select_ckpt_path(
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 108, in _select_ckpt_path
    ckpt_path = self._parse_ckpt_path(
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 121, in _parse_ckpt_path
    if ckpt_path is None and SLURMEnvironment.detect() and self._hpc_resume_path is not None:
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 60, in _hpc_resume_path
    max_version = self.__max_ckpt_version_in_folder(dir_path_hpc, "hpc_ckpt_")
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 537, in __max_ckpt_version_in_folder
    files = [os.path.basename(f["name"]) for f in fs.listdir(uri)]
[...]

PR #7497, opened 2023-09-23
add init group during exp manager

gshennvm:
# What does this PR do ?

Fixes https://github.com/NVIDIA/NeMo/issues/7460. Exp manager now initializes distributed groups to force a barrier.

# Changelog 

  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information
* Related to # (issue)
#7460

### DIFF
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -309,6 +309,7 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
     # Add rank information to logger
     local_rank = int(os.environ.get("LOCAL_RANK", 0))
     global_rank = trainer.node_rank * trainer.num_devices + local_rank
+    logging.rank = global_rank
 
     # the exp manager calls some operations that require explicit
     # synchronization, therefore we need to initialize the process
@@ -321,10 +322,6 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
         if trainer.strategy.launcher is not None:
             trainer.strategy.launcher.launch(dummy, trainer=trainer)
         trainer.strategy.setup_environment()
-
-        global_rank = torch.distributed.get_rank()
-
-    logging.rank = global_rank
 
     if cfg is None:
         logging.error("exp_manager did not receive a cfg argument. It will be disabled.")
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -28,6 +28,7 @@ import pytorch_lightning
 import torch
 from hydra.core.hydra_config import HydraConfig
 from hydra.utils import get_original_cwd
+from megatron.core import parallel_state
 from omegaconf import DictConfig, OmegaConf, open_dict
 from pytorch_lightning.callbacks import Callback, ModelCheckpoint
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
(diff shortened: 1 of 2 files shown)


nemo/utils/exp_manager.py
```diff
@@ -307,9 +308,21 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
...
+    if parallel_state.is_unitialized():
+
+        def dummy():
+            return
+
+        if trainer.strategy.launcher is not None:
+            trainer.strategy.launcher.launch(dummy, trainer=trainer)
+        trainer.strategy.setup_environment()
+
+    global_rank = torch.distributed.get_rank()
```
ericharper, line 325: This forces even single gpu runs to launch torch.distributed, let's only do this if there's more than one gpu

nemo/utils/exp_manager.py
```diff
@@ -307,9 +308,21 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
...
     """
     # Add rank information to logger
-    # Note: trainer.global_rank and trainer.is_global_zero are not set until trainer.fit, so have to hack around it
     local_rank = int(os.environ.get("LOCAL_RANK", 0))
-    global_rank = trainer.node_rank * trainer.num_devices + local_rank
+
+    # the exp manager calls some operations that require explicit
+    # synchronization, therefore we need to initialize the process
+    # group to initiate a barrier
+    if parallel_state.is_unitialized():
```
ericharper, line 316: just check if torch.distrib is init here, don't need megatron.core

ericharper: also make sure there's more than one gpu here

gshennvm: fixed https://github.com/NVIDIA/NeMo/pull/7497/commits/07ff8ee4e1c6bf3ac75ba68ece1716aeec2ba921#diff-cacb4d3cf39593d980bcfcd9c7ed4b9a342b9565c5447511758daf1f02e0f60fR316

nemo/utils/exp_manager.py
gshennvm: 07ff8ee4e1c6bf3ac75ba68ece1716aeec2ba921 only updated from before if we initialized the process group

ericharper (approved):
LGTM. Thanks!

nemo/utils/exp_manager.py
```diff
@@ -307,11 +307,22 @@ def exp_manager(trainer: 'pytorch_lightning.Trainer', cfg: Optional[Union[DictCo
...
     local_rank = int(os.environ.get("LOCAL_RANK", 0))
     global_rank = trainer.node_rank * trainer.num_devices + local_rank
     logging.rank = global_rank
 
+    # the exp manager calls some operations that require explicit
+    # synchronization, therefore we need to initialize the process
+    # group to initiate a barrier
+    if not torch.distributed.is_initialized() and trainer.num_nodes * trainer.num_devices > 1:
+
+        def dummy():
```
titu1994, line 319: Hmm that seems reasonable to avoid race conditions.

It also should probably not have an effect on 1 GPU things. 

However I'm worried about what hacking ptl like this will do to ddp training and stuff. What happens when you call trainer.fit() after already setting up ddp with that dummy ? Will it reinitialize ? Not likely. So what does the dummy do then to normal PTL behaviour ?

gshennvm: when testing on my local setup, doing trainer.fit after this initialization seems totally fine. But I don't really know how it behaves on larger runs. To be save will you rather me modify NLPDDPStrategy as well and set a flag that checks if setup_environment has already been called?

gshennvm: actually looking closely, i think this is safe because PTL guards against already initialized case  https://github.com/Lightning-AI/lightning/blob/984f49f7195ddc67e961c7c498ee6e19fc0cecb5/src/lightning/fabric/utilities/distributed.py#L237

titu1994:
To me, personally I really do not want to make Exp Manager deal with distributed computing under any circumstances. 

While this may be foolproof solution with barrier and stuff, it messes around ptl internals. 

Instead, please see if you can do two things - 
If trauner.num_nodes * num_gpus > 1 and trainer.global_rank == 0 (assume you can  properly access trainer.global_rank), then simply sleep on rank 0 for 5-10 seconds. It's a minor waste, but way simpler without making all these hacks with distributed.

gshennvm:
> To me, personally I really do not want to make Exp Manager deal with distributed computing under any circumstances.
> 
> While this may be foolproof solution with barrier and stuff, it messes around ptl internals.
> 
> Instead, please see if you can do two things - If trauner.num_nodes * num_gpus > 1 and trainer.global_rank == 0 (assume you can properly access trainer.global_rank), then simply sleep on rank 0 for 5-10 seconds. It's a minor waste, but way simpler without making all these hacks with distributed.

I know of runs that have slept 30seconds + but still have this issue. I agree this may mess with PTL internals but sleep is going to have to depend on the size of the run and size of the cluster. We'll have this issue forever unless we have a barrier

ericharper:
Closing in favor of #7498

Issue #7495, opened 2023-09-23
How do I make nemo log only on rank 0?


stas00:
I can't figure out how to make the logging not repeat on 512 ranks. Is there a setting to tell nemo to log only on rank 0? at the very least for the iteration logs?

I currently have:
```
  exp_manager.create_dllogger_logger: True
```
but I'd happy with any other way.

What I ideally would like is the original Megatron-LM-like style logging where it logs iteration metrics from rank 0.

Thank you!

titu1994:
Yup you can use this flag in Exp Manager - https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L166-L167

Either 1 per local rank 0, or just 1 for global rank 0 in total.

stas00:
oh, I already had:

```
exp_manager:
[...]
  log_local_rank_0_only: True
```

but it doesn't help, I get logging on each rank

titu1994:
Odd, this should be fine - https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/exp_manager.py#L393-L402

stas00:
this looks like a different logger to a specific exp_manager file, I'm talking about what goes to `stdout`

I'd imagine it's because of `exp_manager.create_dllogger_logger: True` but I have no idea how to control its behavior as there are no such config options that I can see.

titu1994:
Dllogger doesn't have that many fine grained options. If they do enable it, please do add a PR to add those arguments !

stas00:
Hmm, are you sure this is even in the Dllogger land? Its API appears to be totally unaware of ranks.

Should it be something like this on nemo side instead?
```
if (log_local_rank_0_only and rank==0) or not log_local_rank_0_only:
  DLLogger.log(...)
```
using pseudo-code as I don't yet know the actual variables you use.

stas00:
As @jan-golda confirmed here: https://github.com/NVIDIA/dllogger/issues/9#issuecomment-1734541555
this is indeed a job for NeMo.

titu1994:
NeMo does not special case loggers, we get the implementation from here - https://github.com/NVIDIA/NeMo/blob/main/nemo/utils/loggers/dllogger.py

If you want, you can submit a pr to add a check here to only log on rank 0. But the proposed solution isn't a general one so it can't be used.

stas00:
Could you please why?

I see that you have this call:

[...]

PR #7513, opened 2023-09-26
dllogger - log on rank 0 only

stas00:
# What does this PR do ?

As discussed in https://github.com/NVIDIA/NeMo/issues/7495 @titu1994 helped me to make Dllogger to log on rank0 as otherwise the metrics log happens 512 times per step on 512 gpus.

Fixes: https://github.com/NVIDIA/NeMo/issues/7495

# Changelog 
- make Dllogger to log on rank0 as otherwise the metrics log happens 512 times per step on 512 gpus.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to https://github.com/NVIDIA/NeMo/issues/7495

### DIFF
--- a/nemo/utils/loggers/dllogger.py
+++ b/nemo/utils/loggers/dllogger.py
@@ -20,6 +20,7 @@ from typing import Optional
 from lightning_utilities.core.apply_func import apply_to_collection
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from pytorch_lightning.loggers import Logger
+from pytorch_lightning.utilities import rank_zero_only
 from pytorch_lightning.utilities.parsing import AttributeDict
 
 from nemo.utils import logging
@@ -81,6 +82,7 @@ class DLLogger(Logger):
             )
         dllogger.init(backends=backends)
 
+    @rank_zero_only
     def log_hyperparams(self, params, *args, **kwargs):
         if isinstance(params, Namespace):
             params = vars(params)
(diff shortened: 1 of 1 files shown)


titu1994:
This looks good ! Could you add a comment to the init docstring ?

titu1994 (approved):
Thanks !

titu1994:
Let's merge as is for now, thanks for the PR !

stas00:
After sleeping on this, are we sure this is a good solution? What if someone wants all ranks logged on Dllogger?

After all the users already define the value of `log_local_rank_0_only` so why not call `.log` conditionally on this value instead of hardcoding things to a way that would be hard to reverse down the road.

titu1994:
Well the best case would be dllogger supports distributed logging. 
Barring that, your suggestion makes sense but it's not just one flag, its also the global rank 0 only. Also, how would you make it work with local rank 0 only - we have a function only to check if your global rank is 0, not to check local rank is 0. 

I would say, let the dllogger class constructor take a bool, rank_zero_only - and when creating the object in Exp Manager, we do rank_zero_only=local rank zero only OR global rank zero only. 

Next part is doing this bool check and only then checking if ptl rank zero check is true or not, then calling dllogger.log(). 

It's complicated but doable I guess

titu1994:
Your new proposal with both flags is fine but put all that logic inside of the constructor of DLLogger class, not outside it inside exp Manager. Exp Manager will not specialize to any one logger, that's the individual loggers responsibility to manage itself.

titu1994:
Well we can generalize later if needed. At the moment just dl logger needs this support

PR #7518, opened 2023-09-26
Fix learning rate schedule in Megatron models when `max_steps` is not set

odelalleau:
# What does this PR do ?

This fixes the learning rate schedule in Megatron models when relying on `max_epochs` instead of `max_steps`.

The issue is that [compute_max_steps](https://github.com/NVIDIA/NeMo/blob/bdea326811ef81d3aff3530b26b67f7a7ce5119c/nemo/core/optim/lr_scheduler.py#L934) is incorrect for Megatron models since `accumulate_grad_batches` is equal to 1 from PTL's perspective, even if we are actually using multiple micro-batches.

**Collection**: nlp

# Changelog 
- Fix learning rate schedule in Megatron models when `max_steps` is not set

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

I considered two potential implementations:

1. (this one) setting `max_steps` directly in `MegatronBaseModel`
2. Setting `t_accumulate_grad_batches` correctly in the scheduler config so that the formula is correct

I didn't go with (2) because it seemed trickier to ensure that all parameters are correct so that the formula works, including with drop_last which [still raises an old warning](https://github.com/NVIDIA/NeMo/blob/a420f9083ea9993152a5221b8875c34470561f13/nemo/core/optim/lr_scheduler.py#L943). It's easier to just compute `max_steps` within `MegatronBaseModel`.

cc @aklife97


nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -449,6 +449,12 @@ def on_validation_batch_end(self, outputs, batch: Any, batch_idx: int, dataloade
     def setup_optimization(
         self, optim_config: Optional[Union[DictConfig, Dict]] = None, optim_kwargs: Optional[Dict[str, Any]] = None,
     ):
+        # Ensure `max_steps` is set correctly
+        optim_config = self._get_optim_config_copy(optim_config)
```
odelalleau, line 453: This creates a copy of the config, and another one is also created in `ModelPT` ==> we end up copying it twice. It's a bit unsatisfying but since it's a very small config the performance hit is negligible, and trying to avoid the extra copy would probably make the code more compelx.

nemo/core/classes/modelPT.py
```diff
@@ -592,32 +592,20 @@ def setup_optimization(
...
-            if not isinstance(optim_config, DictConfig):
-                optim_config = OmegaConf.create(optim_config)
-
-            # See if internal config has `optim` namespace before preservation
-            if self._cfg is not None and hasattr(self._cfg, 'optim'):
-                if self._cfg.optim is None:
-                    self._cfg.optim = copy.deepcopy(optim_config)
-                else:
-                    with open_dict(self._cfg.optim):
-                        self._cfg.optim = copy.deepcopy(optim_config)
```
odelalleau, line 617: I simplified a bit this code -- for instance the `with open_dict(self._cfg.optim)` was not needed since we are setting `self._cfg.optim` directly (we're not modifying its attributes).

nemo/core/classes/modelPT.py
```diff
@@ -1754,3 +1742,20 @@ def cuda(self, device=None):
         elif isinstance(device, int):
             device = torch.device("cuda", index=device)
         return super().cuda(device=device)
+
+    def _get_optim_config_copy(self, optim_config: Optional[Union[DictConfig, Dict]]) -> Optional[DictConfig]:
```
titu1994, line 1746: Why not just "_optim_copy_config"

titu1994 (commented):
Overall logic seems reasonable, let's let tests pass

nemo/core/classes/modelPT.py
odelalleau: Happy to rename it but did you really mean `_optim_copy_config()`, or would it be `_optim_config_copy()` / `_copy_optim_config()`?

titu1994: _optim_config_copy is better

odelalleau: Sounds good, done in 8e556a9dc

odelalleau:
_Commenting to un-stale PR_

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -527,7 +533,7 @@ def configure_optimizers(self):
             assert self._trainer.max_steps is not None, "'max_steps' is missing in trainer config."
             if hasattr(self._cfg.optim, 'sched'):
                 sched_config = self._cfg.optim.sched
-                sched_config['max_steps'] = self._trainer.max_steps
+                assert sched_config.get('max_steps', -1) >= 0
```
titu1994, line 536: Why should this be set manually on the config above ? Trainer has this value

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -836,3 +842,37 @@ def _val_iterator_done(self, iterator):
...
+                )
+            return self._trainer.max_steps
+
+        if self._trainer.max_epochs is None or self._trainer.max_epochs < 0:
+            raise ValueError(
+                "Cannot compute `max_steps` if neither `trainer.max_steps` nor `trainer.max_epochs` is set"
+            )
+
+        # The number of training step per epoch is typically the number of global batches in the training set...
+        assert self._train_dl is not None
```
titu1994, line 867: Don't use asserts everywhere, check if trainer is none or not and fall back. It's not guaranteed to have trainer when building a non LLM model, but setup optimization and gets called on model restore when there is no trainer. So all these asserts raise the possibility of breaking model restore in untested cases.

titu1994 (commented):
The logic seems fine, but I dunno about the asserts. Core changes need 2+ reviews, so @ericharper if you are ok with this change in nlp then let's merge

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
odelalleau: The whole point of this PR is to fix the situation where `trainer.max_steps` is `None`. We need `cfg.optim.sched.max_steps` to be set for the scheduler to work, so we use `trainer.max_epochs` to compute `max_steps` in this situation (in the new `_get_max_steps()` function added in this PR).

This assert is a sanity check, it shouldn't possible for it to fail in the current code since we do `optim_config.sched.max_steps = self._get_max_steps()`. I added it to (i) make sure my code is working as intended, and (ii) make sure no-one breaks it in the future.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
odelalleau: Ah thanks for pointing this out (though just a note that this is in `MegatronBaseModel` in nlp collection, so this shouldn't affect non-LLM models, right?)

I pushed a tentative fix in 4c991f553. Will mark this PR as draft while I'm testing it.

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
odelalleau: > oh, I just noticed the assert self._trainer.max_steps is not None above... now I'm a bit confused as to how I was testing this, it's been a long time and I don't remember :/

After testing it I now remember: it's because `max_steps` is set to -1 when undefined (which is confusing because `max_epochs` is using `None`). I think this assert was actually a mistake, assuming that a `None` value could appear when undefined, but I tested setting `max_steps: null` in the config and it actually crashes earlier during PTL trainer initialization because `None` is not valid.

=> I removed this assert in ff09438c5 (and added another one that validates that nobody uses a negative value for `max_steps` in the scheduler config, since we use `None` here for undefined).

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
odelalleau: > I pushed a tentative fix https://github.com/NVIDIA/NeMo/commit/3399ce842046c18e5852a075f1b457e426a8ff52. Will mark this PR as draft while I'm testing it.

Testing it made me realize I didn't remember correctly the code flow when writing this fix, and now I remember why I was triggering exceptions in `_get_max_steps()` if it couldn't be computed: this is because if we don't set `max_steps` properly, we fall back to the original issue where it's computed incorrectly.

One reason for my confusion (which I'm mentioning to be sure it doesn't confuse you as well) is that the `prepare_lr_scheduler()` call in `MegatronBaseModel.configure_optimizers()` ([here](https://github.com/NVIDIA/NeMo/blob/fd8f234070b21c64191824bf6b0f3d0cbdc2b61b/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L531)) is not actually the "usual" place where the scheduler is configured (this code is guarded by an `if` block) -- usually this happens in `ModelPT.setup_optimization()` ([here](https://github.com/NVIDIA/NeMo/blob/fd8f234070b21c64191824bf6b0f3d0cbdc2b61b/nemo/core/classes/modelPT.py#L733)).

So if we don't crash in `_get_max_steps()` (when we can't compute it), since we don't usually go through the check I added in `MegatronBaseModel.configure_optimizers()`, then we end up using a wrong schedule (same issue that this PR is trying to fix).

I pushed a new version in a26ac7535. This one actually fully relaxes the constraint that `max_steps` must be computed, and just skips the scheduler if it can't. This seems a bit dangerous to me though and I wanted to ask you if you think it'd be fine to trigger an exception in `MegatronBaseModel.configure_optimizers()` (not `setup_optimization()`), or if we also need that one to work even when the model is not fully setup for training?

odelalleau:
Just realized I had forgotten to un-draft this PR... should be ready for a second look now!

odelalleau:
Re-opening as this bit us again.

@titu1994 would you be able to have another look? (let me know if anything's still unclear, I realize there was some back and forth on some points that may have been confusing)

odelalleau:
jenkins

odelalleau:
@titu1994 would you be able to help review this one? Thanks!

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
titu1994: Silent about scheduler - it was a (bad) default in NeMo but it's easily detected by looking at logs.
For the megatron case, fine to assert for nlp only, other domains let them continue on without scheduler if max steps cannot be calculated (old default)

titu1994:
jenkins

titu1994:
Jenkins seems to be failing but overall this is fine. Core changes need two reviews, so @ericharper for second review

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
odelalleau: > For the megatron case, fine to assert for nlp only, other domains let them continue on without scheduler if max steps cannot be calculated (old default)

Sounds good, I added a check in b6354f120 to raise an exception when the scheduler can't be instantiated in Megatron.

odelalleau:
jenkins

odelalleau:
> Jenkins seems to be failing

I'll have a look after it re-runs on the latest version.

ericharper (approved):
LGTM. Thanks!

Ran nightly nemo-ci tests as well and all tests passed.

nemo/core/optim/lr_scheduler.py
```diff
@@ -832,8 +832,13 @@ def prepare_lr_scheduler(
         monitor = 'loss'
 
     # Store exact max_steps if it is provided
-    if 'max_steps' in scheduler_config and scheduler_config['max_steps'] is not None:
-        max_steps = scheduler_config['max_steps']
+    max_steps_from_cfg = scheduler_config.get('max_steps')
+    if max_steps_from_cfg is not None:
+        if max_steps_from_cfg == -1:
```
stevehuang52, line 837: what should we do if we want epoch based training? by default `max_steps` is set to -1, which means it'll return here before calculating the actual max_steps later in this function. `scheduler_config.get('max_steps')` will never be None, since otherwise PTL will throw an error saying max_teps must be an integer. @odelalleau @titu1994 @ericharper

PR #7523, opened 2023-09-26
layernorm1p fix

dimapihtar:
# What does this PR do ?

Passes layernorm_zero_centered_gamma param properly which fixes layernorm1p issue.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


gshennvm (commented):
can you add an option to pass normalization == "layernorm1p"? if we used layernorm1p this code still crashes

examples/nlp/language_modeling/conf/megatron_gpt_config.yaml
```diff
@@ -72,6 +72,7 @@ model:
   kv_channels: null # Projection weights dimension in multi-head attention. Set to hidden_size // num_attention_heads if null
   apply_query_key_layer_scaling: False # scale Q * K^T by 1 / layer-number.
   normalization: 'layernorm' # Normalization layer to use. Options are 'layernorm', 'rmsnorm'
+  layernorm_zero_centered_gamma: True
```
gshennvm, line 75: i don't think we need this flag. When normalization is 'layernorm1p' we will overwrite this flag anyway to be True. So we can hide it from the user to keep it consistent with the previous NeMo activation selection

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1483,10 +1483,14 @@ def build_transformer_config(self) -> TransformerConfig:
         activation_func = activation_to_func(activation)
 
         normalization = self.cfg.get('normalization', 'layernorm')
+        layernorm_zero_centered_gamma = self.cfg.get('layernorm_zero_centered_gamma', 'False')
```
gshennvm, line 1486: see above comment. Can set this to be False instead

gshennvm: can you change this line to just `layernorm_zero_centered_gamma = False` ?

dimapihtar: @gshennvm I think it's better lo leave it as it is since we have ` layernorm_zero_centered_gamma` param in Launcher configs for improved models. What do you think?

gshennvm: I see, ok with me. Users can customize it if needed

gshennvm: @MaximumEntropy given that we're moving to mcore and the user is likely not going to be exposed to the layernorm1p argument if they are coming from mcore. I was thinking maybe we can have both options. (i.e. if the user is more used to mcore they can say LayerNorm + layernorm_zero_centered_gamma). wdyt?

MaximumEntropy (approved):

Issue #7475, opened 2023-09-21
Request to change setuptools version


pplantinga:
**Describe the bug**

In my environment, the pinned version of setuptools is unavailable (`setuptools==65.5.1`). This is a request to unpin or update to a more recent version (latest is `setuptools==68.2.2`).

RobinDong:
The `nemo-toolkit` need the package `nemo-text-processing`, which is also set a fixed version for `setuptools` https://github.com/NVIDIA/NeMo-text-processing/blob/main/requirements/requirements.txt#L9
To unpin the version, we may need to change `requirement.txt` for at least these two repos.

PR #7525, opened 2023-09-26
Unpin setuptools

pplantinga:
This PR seeks to support alternate versions of `setuptools` for those (like me) who cannot use the pinned version.

See: https://github.com/NVIDIA/NeMo/issues/7475


fayejf:
Thanks for the PR. The unpin has been merged to [r1.21.0](https://github.com/NVIDIA/NeMo/pull/7534) and [main](https://github.com/NVIDIA/NeMo/pull/7535).

PR #7538, opened 2023-09-27
Create per.py

ssh-meister:
Script for calculation Punctuation Error Rate and related rates (correct rate, deletions rate, etc.)
@karpnv @ekmb @vsl9 @KunalDhawan


karpnv:
move it from `nemo/collections/common/metrics/per.py` to `nemo/collections/asr/metrics/`

ssh-meister:
@jubick1337 @karpnv because this metric is applicable for ASR and NLP PC models both (there is noticed in the paper that PER allows to compare prediction accuracy between PC models in general)

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
+import json
```
ekmb, line 1: @ssh-meister please add the following:
- header
- docstrings
- pytests
- comments on how to run/use the script (one the model is trained, during training etc)
 I haven't looked at the code yet

ekmb (changes requested):

nemo/collections/common/metrics/per.py
ssh-meister: sure!

jubick1337 (changes requested):

itzsimpl:
> move it from `nemo/collections/common/metrics/per.py` to `nemo/collections/asr/metrics/`

Note that coneptually this implementation of PER will give "useful" results only for NLP, where the input is the text that is to be punctuated and is immutable (i.e. the reference is this same text punctuated by hand; a clear ground truth to compare against). 

For the ASR case, eg. with hybrid_pc models, the input is audio, and the reference is the punctuated transcription (typically obtained by hand). If recognition is flawless then PER will give the correct information (the quality of the punctuation subtask), if recognition is not flawless punctuation might be affected by WER, and PER might be affected as well (eg. a word that is typically followed by a comma gets deleted). In certain occasions the punctuation of a imperfect transcription might still be flawless from the standpoint of punctuation, but since the metric is computed agains the original reference punctuated transcription, the PER score will be affected as well. There is no simple solution to this except to manually re-punctuate the transcription obtained by the ASR task and use that as reference, which might not be viable, but is something to keep in mind when reporting or. using the PER metric.

jubick1337:
Could you please add tests for PER [here](https://github.com/NVIDIA/NeMo/blob/main/tests/collections/common/test_metrics.py)?

examples/asr/punctuation_rates.py
```diff
@@ -0,0 +1,168 @@
...
+
+    with open(input_manifest_path, "r") as manifest:
+        lines = manifest.readlines()
+        samples = [json.loads(line) for line in lines]
+        return samples
+
+
+def write_manifest(output_manifest_path: str, samples: list[dict]) -> None:
+
+    '''Writes samples to .json file.
```
jubick1337, line 76: format

jubick1337 (changes requested):
minor format changes, otherwise looks good to me

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+
+class PER:
+    punctuation_marks = None
+    punctuation_mask = "[PUNCT]"
+
+    def __init__(self, punctuation_marks: list[str], punctuation_mask: str = "[PUNCT]"):
+        PER.punctuation_marks = punctuation_marks
+        PER.punctuation_mask = punctuation_mask
+
+    class Statistics:
```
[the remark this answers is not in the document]
ssh-meister: Thank you, `Statistics` class was removed and necessary functions were added to `PER` class

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+        def __init__(self) -> None:
+            self.punctuation_marks_statistics = pd.DataFrame(
```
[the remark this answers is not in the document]
ssh-meister: Thank you, I've changed implementation from pandas dataframes to dicts

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+            if amounts_sum == 0:
+                return
+
+            self.correct_rate = self.punctuation_marks_statistics.loc["Correct"].sum() / amounts_sum
+            self.deletions_rate = self.punctuation_marks_statistics.loc["Deletions"].sum() / amounts_sum
+            self.insertions_rate = self.punctuation_marks_statistics.loc["Insertions"].sum() / amounts_sum
+            self.substitutions_rate = self.punctuation_marks_statistics.loc["Substitutions"].sum() / amounts_sum
+            self.per = self.deletions_rate + self.insertions_rate + self.substitutions_rate
+
+            return
```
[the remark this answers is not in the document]
ssh-meister: Thank you, removed

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+
+    class Sample:
```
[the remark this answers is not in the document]
ssh-meister: sure! done

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+if __name__ == "__main__":
+    parser = ArgumentParser("Calculate Punctuaton Prediction Accuracy Rates")
+
+    parser.add_argument(
+        "--manifest",
+        required=True,
+        type=str,
+        help="Path .json manifest file with ASR predictions saved at `pred_text` field.",
+    )
+    parser.add_argument(
```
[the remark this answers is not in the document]
ssh-meister: `nargs` is added for direct parsing of punctuation marks in the form of a list, thank you

nemo/collections/common/metrics/per.py
ssh-meister: @ekmb, I've added the information that you have mentioned and pytests. Could I ask for a review, please?

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,343 @@
...
+            print(f"Substitutions rate: {substitutions_rate} ({substitutions_rate * 100:2f}%)")
+
+            per = self.statistics.per
+            print(f"Punctuation Error Rate: {per} ({per * 100:2f}%)")
+
+            print("\nRATES BY PUNCTUATION MARK:")
+            print(tabulate(self.statistics.punctuation_marks_rates, headers='keys', tablefmt='psql'))
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
ssh-meister: @titu1994, I've made a refactoring. Could I ask for a review, please?

ssh-meister:
> Could you please add tests for PER [here](https://github.com/NVIDIA/NeMo/blob/main/tests/collections/common/test_metrics.py)?

@jubick1337, I've added tests, could you please review them again?

karpnv (changes requested):
Lets move rate_punctuation.py to
https://github.com/NVIDIA/NeMo/blob/main/examples/asr/speech_to_text_eval.py by adding use_per=False option.

ssh-meister:
> Lets move rate_punctuation.py to https://github.com/NVIDIA/NeMo/blob/main/examples/asr/speech_to_text_eval.py by adding use_per=False option.

This type of modification will result in a reduction of certain features. For instance, in the `speech_to_text_eval.py` script, there is currently no provision for computing the supported metrics (WER/CER) separately for each sample and then saving them to the manifest. This capability could be essential when considering the use of punctuation rates (correct rate / deletions rate / etc.) as thresholds for further sample filtering (e.g. SDE).

It may be advisable to consider adding a feature to `speech_to_text_eval.py` to calculate only the total PER value on a dataset, but in addition to the existing `rate_punctuation.py` functionality, rather than replacing it.

jubick1337 (dismissed):

examples/asr/rate_punctuation.py
```diff
@@ -0,0 +1,207 @@
...
+# limitations under the License.
+
+import json
+import os
+from argparse import ArgumentParser
+
+import pandas as pd
+from tabulate import tabulate
+
+from nemo.collections.common.metrics.per import PERData
```
titu1994, line 22: Why is this at root of the example directory? It is not a universal script, applicable only to PC models no ?

Better to put under some directory such as "scoring" (better name please)

titu1994: If it's universal, merge the logic with the current eval speech script. And guard all the extra dependencies like tabulate etc in that case

titu1994 (changes requested):

examples/asr/rate_punctuation.py
ssh-meister: > Better to put under some directory such as "scoring" (better name please)

@titu1994, does it make sense to move it to `https://github.com/NVIDIA/NeMo/blob/main/scripts/metric_calculation/`?

titu1994:
I just saw the above comment. If that's the case that per sample metrics can't be calculated, I would say why not add support for that to the eval script ? 

However, usually WER and CER are global metrics. I assume PER is also a global metric. Its somewhat incorrect to have row by row value for wer and cer but we could at least add row by row per data. 

Please discuss with @fayejf and @stevehuang52

examples/asr/rate_punctuation.py
karpnv: need to remove file

karpnv (approved):
Thanks to add PER metric with use_per=True

examples/asr/rate_punctuation.py
ssh-meister: @karpnv @titu1994 thank you! the script rate_punctuation has been removed, and per calculation has been added to speech_to_text_eval.py. Could I ask for a review?

karpnv (approved):
well done!

examples/asr/speech_to_text_eval.py
```diff
@@ -132,6 +152,9 @@ def main(cfg: EvaluationConfig):
                 invalid_manifest = True
                 break
 
+            if cfg.scores_per_sample:
+                samples.append(data)
+
```
titu1994, line 157: Add else samples = None

examples/asr/speech_to_text_eval.py
```diff
@@ -154,6 +177,52 @@ def main(cfg: EvaluationConfig):
             f"contain value for `pred_text`."
         )
 
+    if cfg.use_per:
+        per_data_obj = PERData(
+            hypotheses=predicted_text,
+            references=ground_truth_text,
+            punctuation_marks=list(cfg.text_processing.punctuation_marks),
```
titu1994, line 184: Which config is this ? This file doesn't have text_processing.xyz

examples/asr/speech_to_text_eval.py
```diff
@@ -154,6 +177,52 @@ def main(cfg: EvaluationConfig):
...
+        )
+        per_data_obj.compute()
```
titu1994, line 186: What about reset after compute?

examples/asr/speech_to_text_eval.py
```diff
@@ -154,6 +177,52 @@ def main(cfg: EvaluationConfig):
...
+        per = per_data_obj.per
+
+    if cfg.scores_per_sample:
```
titu1994, line 189: Could you move this to a function inside of NeMo so that all this is hidden ? Maybe in transcribe utils ?

examples/asr/speech_to_text_eval.py
```diff
@@ -173,6 +242,30 @@ def main(cfg: EvaluationConfig):
 
     logging.info(f'Dataset WER/CER ' + str(round(100 * wer, 2)) + "%/" + str(round(100 * cer, 2)) + "%")
 
+    if cfg.use_per:
+        logging.info(f'Dataset PER ' + str(round(100 * per, 2)) + '%')
+
+        if HAVE_TABLUATE_AND_PANDAS:
```
titu1994, line 248: Put into transcribe utils

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,428 @@
...
+# limitations under the License.
+
+import re
+from collections import namedtuple
+from tqdm import tqdm
+
+from nemo.utils import logging
+
+
+def punctuation_error_rate(
```
titu1994, line 22: This .. is punctuation and not phoneme ? Then the metric shorthand needs to be changed everywhere. 

Per is phoneme error rate in asr and tts.

nemo/collections/common/metrics/per.py
```diff
@@ -0,0 +1,428 @@
...
+        punctuation_marks=punctuation_marks,
+        punctuation_mask=punctuation_mask,
+    )
+
+    per_data_obj.compute()
+
+    return per_data_obj.per
+
+
+class PER:
```
titu1994, line 52: Maybe name this to PuncER or something like that to denote it's different from phoneme error rate

titu1994 (commented):
Looks better now, some cleanup necessary. 

Main issue is name. PER is phoneme error rate in asr and tts. We need another name for punctuation error rate 

@ekmb any suggestions ?

examples/asr/speech_to_text_eval.py
ssh-meister: It referred to this - https://github.com/ssh-meister/NeMo/blob/per/examples/asr/speech_to_text_eval.py#L106

nemo/collections/common/metrics/per.py
ssh-meister: Thank you, I have fixed it with commit [c01baad](https://github.com/NVIDIA/NeMo/pull/7538/commits/c01baad459747978e1847cf4a2cdd6c21ee637d5)

nemo/collections/common/metrics/per.py
ssh-meister: Thank you, I have fixed it with commit [c01baad](https://github.com/NVIDIA/NeMo/pull/7538/commits/c01baad459747978e1847cf4a2cdd6c21ee637d5)

examples/asr/speech_to_text_eval.py
ssh-meister: Sure, fixed with the commit [dfc809a](https://github.com/NVIDIA/NeMo/pull/7538/commits/dfc809a2a891e96070a69c9a93d34d2f29cd90f2)

examples/asr/speech_to_text_eval.py
ssh-meister: fixed

examples/asr/speech_to_text_eval.py
ssh-meister: Probably, it would be clearer to add it as a method to `DatasetPunctuationErrorRate` class. Done it with the commit [3c9bfc5](https://github.com/NVIDIA/NeMo/pull/7538/commits/3c9bfc5191e2a689b2d4ff671bbe5449051058fd)

examples/asr/speech_to_text_eval.py
ssh-meister: Added [fda77ce](https://github.com/NVIDIA/NeMo/pull/7538/commits/fda77ce9594d1e757303688e6d0c34cb8c7e9d71)

titu1994 (approved):
This is excellent ! @fayejf @stevehuang52 please review the addition to transcribe utils

karpnv (approved):

jubick1337 (dismissed):

nemo/collections/common/metrics/punct_er.py
```diff
@@ -0,0 +1,473 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
```
ekmb, line 1: The first line should be:
`# Copyright (c) 2023, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.`

ssh-meister: Thank you, fixed

jubick1337 (dismissed):

karpnv (approved):

titu1994 (approved):
Lgtm

PR #7590, opened 2023-10-01
generate tokens_to_generate for all items in a batch

arendu:
# What does this PR do ?

ensure that all examples in a batch generate only `tokens_to_generate` number of tokens.

# Changelog 

# Usage

```python
```

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -748,6 +750,9 @@ def sample_sequence_batch(
 
                 # Clamp the predicted out of vocabulary tokens
                 prev = torch.clamp(prev, max=tokenizer.vocab_size - 1)
+                # force EOS token when tokens_to_generate has been reached for an item in the batch.
+                prev[context_length >= maxlens] = tokenizer.eos_id
```
odelalleau, line 754: This looks ok to me, but there might be an edge case if we happen to use a text generation strategy that doesn't consider `EOS` as the end of generation (the computed length would be incorrect). Is this something that needs to be supported? If yes, I think two things could be done:

1. Do something like `done_token = done_token | (context_length == maxlens).byte()` after the definition of `done_token` below (to be sure we consider generation to be done even if the generation strategy ignores the `EOS` token)
2. Use `prev[context_length == maxlens] = tokenizer.eos_id` here (no need for `>=` since `is_done` will be set for future steps, which will force `EOS` on l.759)

nemo/collections/nlp/modules/common/text_generation_utils.py
```diff
@@ -697,9 +697,11 @@ def sample_sequence_batch(
         all_generated_indices = None  # used to track all generated indices
         # Generate enough tokens for the longest sequence
         maxlen = tokens_to_generate + context_lengths.max().item()
-
         maxlen = inference_strategy.clip_max_len(maxlen)
 
+        maxlens = tokens_to_generate + context_lengths  # used to force EOS for shorter contexts in a batch.
```
odelalleau, line 702: Do we need to also implement a similar logic in `tab_sample_sequence_batch()`? (by the way this function seems to already have drifted away a bit from `sample_sequence_batch()`, and I'm not sure this is intended)

odelalleau (commented):
Looks good to me but a couple of things that might require a closer look

nemo/collections/nlp/modules/common/text_generation_utils.py
yidong72: what about using the padding token?

nemo/collections/nlp/modules/common/text_generation_utils.py
yidong72: yes. let 's fix it in the `tab_sample_sequence_batch` which is used to tabular data generation.

arendu: updated for tab_sample

nemo/collections/nlp/modules/common/text_generation_utils.py
odelalleau: > Is this something that needs to be supported?

As discussed, I'm ok not supporting it for now.

> what about using the padding token?

As discussed, there are some additional subtleties in properly handling padded inputs => personally I'm ok with not doing it in this PR.

arendu:
jenkins

ericharper:
jenkins

arendu:
jenkins

PR #7734, opened 2023-10-16
Set `RotaryEmbedding` base frequency from config

shan18:
### What does this PR do?

Add the ability to change the base value of RoPE embeddings from config.

**Collection**: nlp

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py
@@ -167,7 +167,7 @@ class GPTModel(MegatronModule):
         ub_tp_comm_overlap=False,
         use_flash_attention=False,
         seq_len_interpolation_factor=None,
-        base=10000,
+        rotary_base=10000,
     ):
         super(GPTModel, self).__init__(config=config, share_token_embeddings=share_embeddings_and_output_weights)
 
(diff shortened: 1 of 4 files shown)


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -384,6 +384,7 @@ def model_provider_func(self, pre_process, post_process):
                 use_flash_attention=self.cfg.get('use_flash_attention', False),
                 megatron_legacy=self.cfg.get('megatron_legacy', False),
                 seq_len_interpolation_factor=self.cfg.get('seq_len_interpolation_factor', None),
+                base=self.cfg.get('base', 10000),
```
hsiehjackson, line 387: Should we change the name to `rotary_base` (similar to `rotary_percentage`), so people can know what is this param used for?

hsiehjackson (commented):
Overall, looks good to me! Leave a minor comment.

nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py
```diff
@@ -167,6 +167,7 @@ def __init__(
         ub_tp_comm_overlap=False,
         use_flash_attention=False,
         seq_len_interpolation_factor=None,
+        base=10000,
```
titu1994, line 170: Give it better name - rotary_base or base_positional_scale or something

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
shan18: Done

nemo/collections/nlp/models/language_modeling/megatron/gpt_model.py
shan18: Done. Changed to `rotary_base`

titu1994 (approved):
Thanks !

Issue #7420, opened 2023-09-12
[TTS] Global Style Tokens implementation in FastPitch doesn't follow the original paper


anferico:
**Describe the bug**

The implementation of Global Style Tokens (GSTs) in FastPitch introduced in #6417 does not follow the prescription of the original [paper](https://arxiv.org/abs/1803.09017). In particular, the difference lies in the choice of the reference audio for a given training/validation sample `<text, ground_truth_audio>`:
- In the original paper, it is recommended to use the ground-truth audio as the reference audio
- In the NeMo implementation, the reference audio is chosen at random among audio samples from the same speaker

I discovered this after training a FastPitch model with GSTs and observing that the choice of the reference audio at inference time would have virtually no impact at all. More precisely, given a text T and **n** _different_ reference audios R_1, ..., R_n, passing <T, R_i> for any i in [1, n] would result in almost the exact same audio A. I say "almost" because the files appeared different when compared using `diff`, but they had the _exact same length_ and they sounded _exactly the same_ to the ear.

So what I did was to modify the [code](https://github.com/NVIDIA/NeMo/blob/v1.20.0/nemo/collections/tts/data/dataset.py#L704) that's responsible for selecting the reference audio in `nemo.collections.tts.data.dataset.TTSDataset.__getitem__`, effectively changing this:
`reference_audio = self.featurizer.process(self.data[reference_index]["audio_filepath"], ...)`
to this:
`reference_audio = self.featurizer.process(sample["audio_filepath"], ...)`

which is exactly what the GST paper prescribes. In this case, I observed that the choice of the reference audio at inference time _did_ make a difference, however small (although this is probably due to the fact that the training set I used was not very varied in terms of speaking styles).

My intuition is that if an audio A1 is used to compute the style embedding that is then used to condition the generation of the Mel spectrogram for another audio A2, the model will learn to ignore that style embedding because the information that can be extracted from A1 is useless to generate the spectrogram for A2. You could argue that this isn't quite true because A1 and A2 come from the same speaker, hence you could extract speaker information from A1 to generate the spectrogram for A2, but FastPitch already contains a SpeakerLookup module that takes care of encoding speaker information.

**Steps/Code to reproduce bug**

1. Train a FastPitch model using GSTs, for example using the following configuration file: `examples/tts/conf/fastpitch_align_44100_adapter.yaml`
2. Perform multiple inference steps where the input text is always the same, but the reference audio is different
[...]

PR #7777, opened 2023-10-22
[TTS] Fix Global Style Tokens (GSTs) implementation in FastPitch

anferico:
# What does this PR do ?

1. Fix bug in `nemo.collections.tts.modules.submodules.ConditionalInput.forward()`
2. Fix implementation of Global Style Tokens in FastPitch

**Collection**: TTS

# Changelog 
- When using `ConditionalInput` in `"concat"` mode, concatenate `inputs` and `conditioning` along the _feature_ dimension, not the _batch_ dimension as before
  - Despite the misleading docstring in `ConditionalInput.forward()` (corrected in this PR), `inputs` and `conditioning` have shapes `B x T x H` and `B x T x C`, respectively (to be even more precise, `conditioning` has shape `B x 1 x C` _before_ the call to `conditioning.repeat()` and `B x T x C` _after_). Consequently, concatenating the two along the batch dimension has 2 problems:
    - If `H != C`, the concatenation will fail
    - If `H == C`, the concatenation results in a `2B x T x C` tensor
       - This causes an error downstream when calling `self.concat_proj()`, which is defined as:
`self.concat_proj = torch.nn.Linear(hidden_dim + condition_dim, hidden_dim)`
       - `self.concat_proj` is meant to map `B x T x (H+C)` tensors to `B x T x H` tensors, so passing a `2B x T x (H=C)` tensor to it results in the following error:
`RuntimeError: mat1 and mat2 shapes cannot be multiplied ((2B*T)x(H=C) and (H+C)xH)` (the actual error message contains constant values, but here I have extracted variables to make my point clearer)
  - **Solution**: concatenate along the feature dimension to obtain a `B x T x (H+C)` tensor
- Fix the implementation of GSTs in FastPitch, specifically regarding the selection of the _reference audio_ for a given training example
  - The current NeMo implementation selects the reference audio for a given training example as a random example from the same speaker. Note that this also forces the user to add `speaker_id` to `sup_data_types` even when they don't want to use speaker embeddings
  - However, the [GSTs paper](https://arxiv.org/abs/1803.09017) clearly states that:
    > During training, the reference signal is ground-truth audio.
  - While this is enough reason to justify amending the current NeMo implementation, I have expressed my thoughts on what selecting the reference audio at random may lead to in [this issue](https://github.com/NVIDIA/NeMo/issues/7420) I opened over a month ago (which went totally unnoticed)

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?
@blisc, @okuchaiev 

# Additional Information
* Related to #7420
* I have run tests, but some failed or errored out. None of the failed tests makes use of my changes though. See [test_short_summary.txt](https://github.com/NVIDIA/NeMo/files/13063436/test_short_summary.txt) for further details

### DIFF
--- a/nemo/collections/tts/data/dataset.py
+++ b/nemo/collections/tts/data/dataset.py
@@ -485,11 +485,7 @@ class TTSDataset(Dataset):
         pass
 
     def add_reference_audio(self, **kwargs):
-        assert SpeakerID in self.sup_data_types, "Please add speaker_id in sup_data_types."
-        """Add a mapping for each speaker to their manifest indexes"""
-        self.speaker_to_index_map = defaultdict(set)
-        for i, d in enumerate(self.data):
-            self.speaker_to_index_map[d['speaker_id']].add(i)
+        pass
 
     def get_spec(self, audio):
         with torch.cuda.amp.autocast(enabled=False):
@@ -529,12 +525,6 @@ class TTSDataset(Dataset):
                     [wav, torch.zeros(self.pad_multiple - wav.shape[0] % self.pad_multiple, dtype=torch.float)]
                 )
         return wav
-
-    # Random sample a reference index from the same speaker
-    def sample_reference_index(self, speaker_id):
-        reference_pool = self.speaker_to_index_map[speaker_id]
-        reference_index = random.sample(reference_pool, 1)[0]
-        return reference_index
 
     def __getitem__(self, index):
         sample = self.data[index]
@@ -699,9 +689,8 @@ class TTSDataset(Dataset):
 
         reference_audio, reference_audio_length = None, None
         if ReferenceAudio in self.sup_data_types_set:
-            reference_index = self.sample_reference_index(sample["speaker_id"])
             reference_audio = self.featurizer.process(
-                self.data[reference_index]["audio_filepath"],
+                sample["audio_filepath"],
                 trim=self.trim,
                 trim_ref=self.trim_ref,
                 trim_top_db=self.trim_top_db,
--- a/nemo/collections/tts/modules/submodules.py
+++ b/nemo/collections/tts/modules/submodules.py
@@ -493,7 +493,7 @@ class ConditionalInput(torch.nn.Module):
     def forward(self, inputs, conditioning=None):
         """
         Args:
-            inputs (torch.tensor): B x T x C tensor.
+            inputs (torch.tensor): B x T x H tensor.
             conditioning (torch.tensor): B x 1 x C conditioning embedding.
         """
         if len(self.condition_types) > 0:
--- a/nemo/collections/tts/modules/submodules.py
+++ b/nemo/collections/tts/modules/submodules.py
@@ -510,7 +510,7 @@ class ConditionalInput(torch.nn.Module):
 
             if "concat" in self.condition_types:
                 conditioning = conditioning.repeat(1, inputs.shape[1], 1)
-                inputs = torch.cat([inputs, conditioning])
+                inputs = torch.cat([inputs, conditioning], dim=-1)
                 inputs = self.concat_proj(inputs)
 
         return inputs


hsiehjackson:
Thanks for the contributions! Great catch for the "concat" mode bug. I have some questions regarding the selection of reference audio. If possible, I would prefer this PR only including the fix of concat bug unless you provide more reasons or results we should use ground-truth audios for reference audios. 

> Fix the implementation of GSTs in FastPitch, specifically regarding the selection of the reference audio for a given training example

If people plan to use GSTs in FastPitch, then we think they suppose to use a multi-speaker dataset with speaker id for each sample. With speaker id, we can choose the utterances from the same speaker as reference audio to make sure that model can only obtain speaker information and prevent copying other speech information from ground-truth audio.

> While this is enough reason to justify amending the current NeMo implementation, I have expressed my thoughts on what selecting the reference audio at random may lead to in https://github.com/NVIDIA/NeMo/issues/7420 I opened over a month ago (which went totally unnoticed)

In our experiments, if we use both SpeakerLookup module and GSTs, then model will almost ignore SpeakerLookup module. These two have different scenarios. If people want to use learned speakers for inference, then SpeakerLookup module is a good choice. If people want to use new speakers for inference, then GSTs is a better choice since we learn how to extract speaker information from reference audios.

rlangman:
Thanks for the PR!

The GST implementation here is not the one everyone is familiar with which is intended to capture speaking style information. Rather it is based off a few lesser known papers later that try to use the same system to extract speaker embeddings from the reference spectrogram (to get speaker information, not speaking style information). The purpose of these alternate methods is to train a speaker embedding extractor that makes it easier to fine-tune on new speakers with very little data.

It would be nice to support the original GST though. Instead of requiring one or the other, I suggest adding a configuration which lets user decide whether they want the reference to be from the ground truth, or a random other utterance from the same speaker.

anferico:
@hsiehjackson 
> Thanks for the contributions! Great catch for the "concat" mode bug. I have some questions regarding the selection of reference audio. If possible, I would prefer this PR only including the fix of concat bug unless you provide more reasons or results we should use ground-truth audios for reference audios.

Sure, that sounds fair 👍🏼 I guess I'll close this PR and open two separate ones, one for the concat bug and one for GSTs.
 
> In our experiments, if we use both SpeakerLookup module and GSTs, then model will almost ignore SpeakerLookup module. These two have different scenarios. If people want to use learned speakers for inference, then SpeakerLookup module is a good choice. If people want to use new speakers for inference, then GSTs is a better choice since we learn how to extract speaker information from reference audios.

I found something very similar in my experiments, except it was the style embeddings (and not the speaker embeddings) that were ignored (cf. #7420). 


---
@rlangman 
> The GST implementation here is not the one everyone is familiar with which is intended to capture speaking style information. Rather it is based off a few lesser known papers later that try to use the same system to extract speaker embeddings from the reference spectrogram (to get speaker information, not speaking style information). The purpose of these alternate methods is to train a speaker embedding extractor that makes it easier to fine-tune on new speakers with very little data.

That's quite interesting, would you mind linking some of those papers?
By the way, I feel like this implementation choice wasn't documented properly in NeMo. Given that the speaker IDs are required anyway by the GST module for the selection of reference audios, users would think that it makes sense to use both the speaker lookup _and_ the GST modules, however as I pointed out in #7420 that can lead the model to learn to ignore style embeddings altogether, which completely defeats the purpose of having a GST module.
If your suggestion, assuming I understood correctly, is to use the GST module alone _without_ the speaker lookup module, I think it should be made very clear to the end user.

> It would be nice to support the original GST though. Instead of requiring one or the other, I suggest adding a configuration which lets user decide whether they want the reference to be from the ground truth, or a random other utterance from the same speaker.

That's a very good point, let me do exactly that in the new PR I'll open.

anferico:
@hsiehjackson @rlangman I've opened 2 new PRs: #7785 and #7788. Let's continue the discussion there, and please feel free to close this PR.

rlangman:
> That's quite interesting, would you mind linking some of those papers?

The FastPitch setup is: https://arxiv.org/abs/2211.00585

The idea to use GST system for speaker embeddings is originally from: https://www1.se.cuhk.edu.hk/~hccl/publications/pub/201909_INTERSPEECH_HuiLU.pdf

PR #7785, opened 2023-10-23
ConditionalInput: cat along the feature dim, not the batch dim

anferico:
# What does this PR do ?

Fix bug in `nemo.collections.tts.modules.submodules.ConditionalInput.forward()`: concatenate `inputs` and `conditioning` along the _feature_ dimension, not the _batch_ dimension.

**Collection**: TTS

# Changelog 
- When using `ConditionalInput` in `"concat"` mode, concatenate `inputs` and `conditioning` along the _feature_ dimension, not the _batch_ dimension as before
  - Despite the misleading docstring in `ConditionalInput.forward()` (corrected in this PR), `inputs` and `conditioning` have shapes `B x T x H` and `B x T x C`, respectively (to be even more precise, `conditioning` has shape `B x 1 x C` _before_ the call to `conditioning.repeat()` and `B x T x C` _after_). Consequently, concatenating the two along the batch dimension has 2 problems:
    - If `H != C`, the concatenation will fail
    - If `H == C`, the concatenation results in a `2B x T x C` tensor
       - This causes an error downstream when calling `self.concat_proj()`, which is defined as:
`self.concat_proj = torch.nn.Linear(hidden_dim + condition_dim, hidden_dim)`
       - `self.concat_proj` is meant to map `B x T x (H+C)` tensors to `B x T x H` tensors, so passing a `2B x T x (H=C)` tensor to it results in the following error:
`RuntimeError: mat1 and mat2 shapes cannot be multiplied ((2B*T)x(H=C) and (H+C)xH)` (the actual error message contains constant values, but here I have extracted variables to make my point clearer)
  - **Solution**: concatenate along the feature dimension to obtain a `B x T x (H+C)` tensor

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?
@subhankar-ghosh, @hsiehjackson

# Additional Information
* Related to #7777

### DIFF
--- a/nemo/collections/tts/modules/submodules.py
+++ b/nemo/collections/tts/modules/submodules.py
@@ -493,7 +493,7 @@ class ConditionalInput(torch.nn.Module):
     def forward(self, inputs, conditioning=None):
         """
         Args:
-            inputs (torch.tensor): B x T x C tensor.
+            inputs (torch.tensor): B x T x H tensor.
             conditioning (torch.tensor): B x 1 x C conditioning embedding.
         """
         if len(self.condition_types) > 0:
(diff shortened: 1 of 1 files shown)


hsiehjackson (approved):
LGTM! Thanks for your PR!

PR #7867, opened 2023-11-06
Tar codec

nithinraok:
Add webdataset support for VocoderDataset

**Collection**: TTS

# Changelog
- Add TarredVocderDataset class for dataset fetching
- sampler is not supported with IterableDataset
- added a conditional code to calculate correct number of max steps when webdataset is used

Known issue:
- Haven;t tested for passing multiple tarred datasets. Will be pushing a PR later to support it. 

# Usage
* add `is_sharded=True`  in cfg for train_ds to load webdataset

# Before your PR is Ready for review
**Pre checks**:

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/data/audio_to_label.py
+++ b/nemo/collections/asr/data/audio_to_label.py
@@ -566,7 +566,6 @@ class _TarredAudioLabelDataset(IterableDataset):
             world_size=world_size,
             global_rank=global_rank,
         )
-        print("world_size", world_size)
         # Put together WebDataset
         self._dataset = wd.WebDataset(urls=audio_tar_filepaths, nodesplitter=None)
 
--- a/nemo/collections/asr/data/audio_to_label.py
+++ b/nemo/collections/asr/data/audio_to_label.py
@@ -566,6 +566,7 @@ class _TarredAudioLabelDataset(IterableDataset):
             world_size=world_size,
             global_rank=global_rank,
         )
+        print("world_size", world_size)
         # Put together WebDataset
         self._dataset = wd.WebDataset(urls=audio_tar_filepaths, nodesplitter=None)
 
--- a/nemo/collections/tts/data/vocoder_dataset.py
+++ b/nemo/collections/tts/data/vocoder_dataset.py
@@ -292,8 +292,10 @@ class TarredVocoderDataset(IterableDataset):
         shard_strategy: str = "scatter",
         global_rank: int = 0,
         world_size: int = 2,
+        **kwargs,
     ):
         super().__init__()
+
         self.sample_rate = sample_rate
         self.n_samples = n_samples
 
@@ -312,18 +314,18 @@ class TarredVocoderDataset(IterableDataset):
         self.audio_tar_filepaths = []
         for dataset_name, dataset_info in dataset_meta.items():
             audio_tar_filepaths = dataset_info.audio_tar_filepaths
-            self.audio_tar_filepaths += audio_tar_filepaths
+            self.audio_tar_filepaths += [audio_tar_filepaths]
             dataset = DatasetMeta(**dataset_info)
             samples, _ = preprocess_manifest(
                 dataset_name=dataset_name, dataset=dataset, min_duration=min_duration, max_duration=max_duration,
             )
             self.data_samples += samples
 
-        self.mapping = {}
-        for sample in enumerate(self.data_samples):
+        self.file_id_to_sample_map = {}
+        for sample in self.data_samples:
             file_id = os.path.splitext(os.path.basename(sample.manifest_entry["audio_filepath"]))[0]
-            if file_id not in self.mapping:
-                self.mapping[file_id] = sample
+            if file_id not in self.file_id_to_sample_map:
+                self.file_id_to_sample_map[file_id] = sample
             else:
                 raise ValueError(
                     f"Duplicate file_id {file_id} found in manifest {sample.manifest_entry['audio_filepath']}"
@@ -331,7 +333,7 @@ class TarredVocoderDataset(IterableDataset):
 
         logging.info(f"world size: {world_size}")
         audio_tar_filepaths = expand_sharded_filepaths(
-            sharded_filepaths=self.audio_tar_filepaths,
+            sharded_filepaths=audio_tar_filepaths,
             global_rank=global_rank,
             world_size=world_size,
             shard_strategy=shard_strategy,
@@ -353,9 +355,9 @@ class TarredVocoderDataset(IterableDataset):
 
     def _filter(self, iterator):
         class FilteredIterator:
-            def __init__(self, mapping):
+            def __init__(self, file_id_to_sample_map):
                 self.iterator = iterator
-                self.mapping = mapping
+                self.file_id_to_sample_map = file_id_to_sample_map
 
             def __iter__(self):
                 return self
@@ -364,15 +366,15 @@ class TarredVocoderDataset(IterableDataset):
                 while True:
                     audio_bytes, audio_filename = next(self.iterator)
                     file_id = os.path.splitext(os.path.basename(audio_filename))[0]
-                    if file_id in self.mapping:
+                    if file_id in self.file_id_to_sample_map:
                         return audio_bytes, audio_filename
 
-        return FilteredIterator(self.mapping)
+        return FilteredIterator(self.file_id_to_sample_map)
 
     def _build_sample(self, tup):
         audio_bytes, audio_filename = tup
         file_id = os.path.splitext(os.path.basename(audio_filename))[0]
-        data = self.mapping[file_id]
+        data = self.file_id_to_sample_map[file_id]
 
         audio_array, sr = sf.read(file=io.BytesIO(audio_bytes), dtype='float32')
         if sr != self.sample_rate:
@@ -420,10 +422,4 @@ class TarredVocoderDataset(IterableDataset):
         return self._dataset.__iter__()
 
     def __len__(self):
-        return len(self.mapping)
-
-    def __iter__(self):
-        return self._dataset.__iter__()
-
-    def __len__(self):
-        return len(self.mapping)
+        return len(self.file_id_to_sample_map)
--- a/nemo/collections/tts/data/vocoder_dataset.py
+++ b/nemo/collections/tts/data/vocoder_dataset.py
@@ -218,7 +218,7 @@ class VocoderDataset(Dataset):
 
 class TarredVocoderDataset(IterableDataset):
     """
-    A similar Dataset to the VocoderDataset, but which loads tarred audio files.
+    A similar Dataset to the VocoderDataset, but loads tarred audio files.
 
     Accepts a single comma-separated JSON manifest file (in the same style as for the VocoderDataset),
     as well as the path(s) to the tarball(s) containing the wav files. Each line of the manifest should
@@ -280,7 +280,6 @@ class TarredVocoderDataset(IterableDataset):
 
     def __init__(
         self,
-        *,
         dataset_meta: Dict,
         sample_rate: int = 24000,
         n_samples: Optional[int] = None,
@@ -295,6 +294,11 @@ class TarredVocoderDataset(IterableDataset):
         **kwargs,
     ):
         super().__init__()
+
+        if len(kwargs) > 0:
+            logging.warning(
+                f"Arguments {kwargs.keys()} does not support for TarredVocoderDataset, they will be ignored."
+            )
 
         self.sample_rate = sample_rate
         self.n_samples = n_samples
--- a/nemo/collections/tts/data/vocoder_dataset.py
+++ b/nemo/collections/tts/data/vocoder_dataset.py
@@ -16,13 +16,13 @@ import io
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional
 
 import soundfile as sf
 import torch.utils.data
 import webdataset as wd
 
-from nemo.collections.asr.data.audio_to_text import cache_datastore_manifests, expand_sharded_filepaths
+from nemo.collections.asr.data.audio_to_text import expand_sharded_filepaths
 from nemo.collections.asr.parts.preprocessing.segment import available_formats as valid_sf_formats
 from nemo.collections.asr.parts.utils.manifest_utils import read_manifest
 from nemo.collections.tts.parts.preprocessing.feature_processors import FeatureProcessor
@@ -53,33 +53,6 @@ class DatasetSample:
     dataset_name: str
     manifest_entry: dict
     audio_dir: Path
-
-
-def audio_collate_fn(batch: List[dict]):
-    dataset_name_list = []
-    audio_filepath_list = []
-    audio_list = []
-    audio_len_list = []
-
-    for example in batch:
-        dataset_name_list.append(example["dataset_name"])
-        audio_filepath_list.append(example["audio_filepath"])
-        audio_list.append(example["audio"])
-        audio_len_list.append(example["audio_len"])
-
-    batch_audio_len = torch.IntTensor(audio_len_list)
-    audio_max_len = int(batch_audio_len.max().item())
-
-    batch_audio = stack_tensors(audio_list, max_lens=[audio_max_len])
-
-    batch_dict = {
-        "dataset_names": dataset_name_list,
-        "audio_filepaths": audio_filepath_list,
-        "audio": batch_audio,
-        "audio_lens": batch_audio_len,
-    }
-
-    return batch_dict
 
 
 def audio_collate_fn(batch: List[dict]):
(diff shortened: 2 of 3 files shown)


nemo/collections/tts/models/audio_codec.py
```diff
@@ -490,6 +510,28 @@ def _setup_test_dataloader(cfg):
 
     def setup_training_data(self, cfg):
         self._train_dl = self._setup_train_dataloader(cfg)
+        batch_size = cfg['dataloader_params']['batch_size']
+        # Need to set this because if using an IterableDataset, the length of the dataloader is the total number
+        # of samples rather than the number of batches, and this messes up the tqdm progress bar.
+        # So we set the number of steps manually (to the correct number) to fix this.
```
rlangman, line 516: Would it work if we defined the `__len__` method of the dataset as `len(self.mapping) / self.batch_size`?

Alternatively would the wrapper `webdataset.ResizedDataset` help? It looks like it allows you to define dataset length and number of steps to limit to in an epoch as separate values.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+                        return audio_bytes, audio_filename
+
+        return FilteredIterator(self.mapping)
+
+    def _build_sample(self, tup):
+        audio_bytes, audio_filename = tup
+        file_id = os.path.splitext(os.path.basename(audio_filename))[0]
+        data = self.mapping[file_id]
+
+        audio_array, _ = sf.read(file=io.BytesIO(audio_bytes), dtype='float32')
```
rlangman, line 486: It looks like webdataset requires reading entire audio file off disk even if we only want a very small sample from it. Is there a risk of an I/O bottleneck if we have to read data on the order of 20 seconds long just to sample <= 1 second of it?

nithinraok: Very good point. As far as I know, we cannot change the core indexing of WebDataset, which reads the tar file and returns the bytes.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+    def __iter__(self):
+        return self._dataset.__iter__()
+
+    def __len__(self):
+        return len(self.mapping)
+
+
+class TarredVocoderDataset(_TarredVocoderDataset):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
nithinraok: TarredVocoderDataset is a base class and if other users want to create another version for this vocoder they can use it and create.

nemo/collections/tts/models/audio_codec.py
```diff
@@ -473,10 +477,26 @@ def validation_step(self, batch, batch_idx):
...
+            is_sharded = False
+
+        if is_sharded:
+            cfg.dataset._target_ = 'nemo.collections.tts.data.vocoder_dataset.TarredVocoderDataset'
+            dataset = instantiate(cfg.dataset)
+            sampler = None
+        else:
+            dataset = instantiate(cfg.dataset)
+            sampler = dataset.get_sampler(cfg.dataloader_params.batch_size)
+        return dataset, sampler
```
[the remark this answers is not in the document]
nithinraok: Good point i will create a simpler case here, but I created this function to set the target so as to use existing config files and provide toggle option for sharding.

nemo/collections/asr/data/audio_to_label.py
```diff
@@ -566,6 +566,7 @@ def __init__(
             world_size=world_size,
             global_rank=global_rank,
         )
+        print("world_size", world_size)
```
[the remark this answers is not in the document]
nithinraok: Will move to logging

nemo/collections/tts/data/vocoder_dataset.py
rlangman: I don't think we can access the original file when using Webdataset. You can only access the content as bytes, at which point I/O to read the full file has already happened.

If this ends up problematic, the best solution may be to segment large files into smaller files during in the tar script.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
-        batch_dict = {
-            "dataset_names": dataset_name_list,
-            "audio_filepaths": audio_filepath_list,
-            "audio": batch_audio,
-            "audio_lens": batch_audio_len,
+    def __init__(
+        self,
+        *,
+        dataset_meta: Dict,
+        sample_rate: int,
```
rlangman, line 390: `sample_rate` in this implementation only used with `trunc_duration`, which is awkward but if we keep them both then they should both be optional.

Though how should we validate that the sample rates of all of the input tarred datasets are the same? Should we read it off the tarred dataset version info?

Or should we write the sample rate of each audio into the manifest and then resample it in this class if needed?

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+        num_audio_retries: int = 5,
+        shard_strategy: str = "scatter",
+        global_rank: int = 0,
+        world_size: int = 2,
+    ):
+        super().__init__()
+        self.sample_rate = sample_rate
+        self.n_samples = n_samples
+        self.num_audio_retries = num_audio_retries
+        self.load_precomputed_mel = False
```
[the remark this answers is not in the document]
rlangman: `load_precomputed_mel` is a flag used in the HiFi-GAN and UnivNet recipes.

`num_audio_retries` can be removed here.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+            self.trunc_samples = None
+
+        if feature_processors:
+            logging.info(f"Found feature processors {feature_processors.keys()}")
+            self.feature_processors = list(feature_processors.values())
+        else:
+            self.feature_processors = []
+
+        self.data_samples = []
+        self.sample_weights = []
```
rlangman, line 420: Sample weights are not used here.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+        self.audio_tar_filepaths = []
+        for dataset_name, dataset_info in dataset_meta.items():
+            audio_tar_filepaths = dataset_info['audio_tar_filepaths']
+            self.audio_tar_filepaths += audio_tar_filepaths
+            dataset = DatasetMeta(**dataset_info)
+            samples, weights = VocoderDataset._preprocess_manifest(
+                dataset_name=dataset_name, dataset=dataset, min_duration=min_duration, max_duration=max_duration,
+            )
```
rlangman, line 428: We should not use a protected method of a different class. `preprocess_manifest` could be made a generic method, or we could define a new one for tarred datasets since we do not need the sample weights here.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+        else:
+            self.feature_processors = []
+
+        self.data_samples = []
+        self.sample_weights = []
+        self.audio_tar_filepaths = []
+        for dataset_name, dataset_info in dataset_meta.items():
+            audio_tar_filepaths = dataset_info['audio_tar_filepaths']
+            self.audio_tar_filepaths += audio_tar_filepaths
+            dataset = DatasetMeta(**dataset_info)
```
rlangman, line 425: We should either add `audio_tar_filepaths` as an optional field to `DetasetMeta`, or define a new dataclass like `TarredDatasetMeta`.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+            samples, weights = VocoderDataset._preprocess_manifest(
+                dataset_name=dataset_name, dataset=dataset, min_duration=min_duration, max_duration=max_duration,
+            )
+            self.data_samples += samples
+            self.sample_weights += weights
+        self.mapping = {}
```
rlangman, line 431: Name is not very clear. Something like `self.file_id_to_sample_map` will be easier to read.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+
+        self._dataset = (
+            self._dataset.rename(audio=VALID_FILE_FORMATS, key='__key__')
+            .to_tuple('audio', 'key')
+            .pipe(self._filter)
+            .map(f=self._build_sample)
+        )
+
+    def _filter(self, iterator):
+        class FilteredIterator:
```
rlangman, line 464: Consider using a lambda instead of function currying. Like `.pipe(lambda iterator: FilteredIterator(iterator))`

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
+        self._dataset = wd.WebDataset(audio_tar_filepaths, nodesplitter=None)
+
+        if shuffle_n > 0:
+            self._dataset = self._dataset.shuffle(shuffle_n)
+        else:
+            logging.info("WebDataset will not shuffle data. Consider setting shuffle_n > 0.")
+
+        self._dataset = (
+            self._dataset.rename(audio=VALID_FILE_FORMATS, key='__key__')
+            .to_tuple('audio', 'key')
```
rlangman, line 458: Nitpick: it is not very important for audio only dataset, but code would be more readable and extensible using original dictionary format instead of converting to tuple.

nemo/collections/tts/data/vocoder_dataset.py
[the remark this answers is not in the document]
nithinraok: We don;t want to use private class with `_` hence created  `TarredVocoderDataset` its just a convention

nemo/collections/tts/models/audio_codec.py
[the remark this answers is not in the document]
nithinraok: I am against creating one new config for each feature.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +238,294 @@ def __getitem__(self, index):
...
 
-        for example in batch:
-            dataset_name_list.append(example["dataset_name"])
-            audio_filepath_list.append(example["audio_filepath"])
-            audio_list.append(example["audio"])
-            audio_len_list.append(example["audio_len"])
 
-        batch_audio_len = torch.IntTensor(audio_len_list)
-        audio_max_len = int(batch_audio_len.max().item())
+class _TarredVocoderDataset(IterableDataset):
```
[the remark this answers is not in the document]
nithinraok: Mistake, probably this occured when I was trying to pull from earlier merge request and merging was not done probably. Thanks for catching this, fixed.

nemo/collections/tts/data/vocoder_dataset.py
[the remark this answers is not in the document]
nithinraok: Done

nemo/collections/tts/models/audio_codec.py
```diff
@@ -473,10 +477,26 @@ def validation_step(self, batch, batch_idx):
         }
         self.log_dict(metrics, on_epoch=True, sync_dist=True)
 
+    @staticmethod
+    def get_dataset(cfg):
+        if 'is_sharded' in cfg.dataset:
+            is_sharded = cfg.dataset.is_sharded
+            del cfg.dataset.is_sharded
+        else:
+            is_sharded = False
```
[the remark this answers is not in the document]
nithinraok: Unfortunately DictConfig in struct mode does not support pop. I will use open_dict to achieve this.

nemo/collections/tts/data/vocoder_dataset.py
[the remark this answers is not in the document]
nithinraok: Removed the wrapper for now

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: very good point, currently it reads bytes so we don;t have much option with tarred reading. So I added a condition to check the target sample rate and audio sample rate, and throw warning otherwise.

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: I wonder why you made it as a protected method, when you defined it as static?

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: it is already added as an optional argument (?)

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: I think current way is easier to read? Any reason for preference?

nemo/collections/tts/data/vocoder_dataset.py
rlangman: Oh, so it is. Then it can be accessed as `dataset.audio_tar_filepaths` instead of `dataset_info['audio_tar_filepaths']`.

nithinraok:
jenkins

nemo/collections/tts/data/vocoder_dataset.py
rlangman: I guess I was not expecting another class to use it, or if it did then I would move it to a non-class method or tts_dataset_utils.py.

Given that method is static, it can be moved outside the VocoderDataset class.

nemo/collections/tts/models/audio_codec.py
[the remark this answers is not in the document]
rlangman: Do you mean you are against creating separate config files for new features in general (eg. sampling rate, new architectures)? Or specifically for needing separate config for tarred vs non-tarred datasets?

The former of creating lots of different overlapping configs is expected across NeMo. The latter can be addressed in several different ways, if needed.

nemo/collections/tts/data/vocoder_dataset.py
rlangman: Main reason would be I will need to move FilteredIterator to a static standalone class when I reuse it for other tarred dataset implementation, but I can modify that in a future PR if needed

nemo/collections/tts/models/audio_codec.py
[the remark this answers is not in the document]
nithinraok: We don;t want to create multiple configs with same model architecture but different setting (like tarred/sampling_rate etc).

nemo/collections/tts/models/audio_codec.py
nithinraok: `len(self.mapping) / self.batch_size` I dont think this would work, as we are facing issue with missing world_size but not batch_size. Or we should have a way to pass global batch size than batch_size per GPU

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: It doesn;t make sense to move it out of `VocoderDataset` unless other tts samples also use similar DatasetSample like dataclasses

nithinraok: I moved to a separate function

nithinraok:
jenkins

nithinraok:
jenkins

nemo/collections/tts/models/audio_codec.py
```diff
@@ -474,9 +478,20 @@ def validation_step(self, batch, batch_idx):
...
     @staticmethod
-    def _setup_train_dataloader(cfg):
+    def get_dataset(cfg):
+        with open_dict(cfg):
+            is_sharded = cfg.dataset.pop('is_sharded', False)
+
+        if is_sharded:
+            cfg.dataset._target_ = 'nemo.collections.tts.data.vocoder_dataset.TarredVocoderDataset'
+
         dataset = instantiate(cfg.dataset)
```
rlangman, line 488: For `TarredVocoderDataset` don't we need to pass the trainer `world size` and `global rank` into `instantiate`?

nemo/collections/tts/models/audio_codec.py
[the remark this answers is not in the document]
rlangman: For the same model architecture with different configs are main options I have heard before are:

1. Create different example configs.
2. Have one example config and provide web documentation and tutorials clearly indicating to users how to manually change fields for different valid configs (eg. sample rate, language).

From past discussions around this, generally (1) was preferable, especially for changes like sample rate which are relatively complicated and error prone if you attempt to update all needed parameters manually.  A combination of (1) and (2) is needed if you have too many configs to support, such as when having to support multiple sample rates and languages at the same time.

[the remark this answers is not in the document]
rlangman: In this particular case, there are 2 problems with the existing approach.

- The model class takes the data loader class/config as input, however the constructor requires passing information `global_rank` and `world_size` which are known only to the model class and trainer.
- Different data loaders require significantly different inputs to instantiate (and this will diverge more as we add additional sampling strategies).

Perhaps the cleanest way to address these would be to change it so the model is expected to receive a subset of dataset parameters, and then add a module which maps the parameters to the logic to create each dataset.

For example:

**In .yaml**
```
train_ds:
  dataset_type: "vocoder_tarred"
  dataset_args:
    batch_size: ${batch_size}
    ... 
```

**In tts_dataset_utils.py**

```
def create_dataset(dataset_type, dataset_args, global_rank, world_size, ...):
    if dataset_type == "vocoder":
        return VocoderDataset(*dataset_args)
    elif dataset_type == "vocoder_tarred":
        return TarrredVocoderDataset(*dataset_args, global_rank=global_rank, world_size=world_size)
     ...
```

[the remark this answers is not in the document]
nithinraok: Good catch, missed it. Updated to use those as well.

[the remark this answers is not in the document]
nithinraok: for logging, testing, validation we rely only on non tarred datasets.

nithinraok:
jenkins

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +212,218 @@ def __getitem__(self, index):
...
+
+    See the WebDataset documentation for more information about accepted data and input formats.
+
+    If using multiple processes the number of shards should be divisible by the number of workers to ensure an
+    even split among workers. If it is not divisible, logging will give a warning but training will proceed.
+    In addition, if using mutiprocessing, each shard MUST HAVE THE SAME NUMBER OF ENTRIES after filtering
+    is applied. We currently do not check for this, but your program may hang if the shards are uneven!
+
+    Notice that a few arguments are different from the AudioToBPEDataset; for example, shuffle (bool) has been
+    replaced by shuffle_n (int).
```
rlangman, line 239: Could remove this part as it is related to the ASR set.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +212,218 @@ def __getitem__(self, index):
...
+        shuffle_n (int): How many samples to look ahead and load to be shuffled.
+            See WebDataset documentation for more details.
+            Defaults to 0.
+        shard_strategy (str): Tarred dataset shard distribution strategy chosen as a str value during ddp.
+            -   `scatter`: The default shard strategy applied by WebDataset, where each node gets
+                a unique set of shards, which are permanently pre-allocated and never changed at runtime.
+            -   `replicate`: Optional shard strategy, where each node gets all of the set of shards
+                available in the tarred dataset, which are permanently pre-allocated and never changed at runtime.
+                The benefit of replication is that it allows each node to sample data points from the entire
+                dataset independently of other nodes, and reduces dependence on value of `shuffle_n`.
```
rlangman, line 267: If I am not mistaken, `shuffle_n` is the size of the shuffle buffer as the gpu reads in data from its shards. Allocating more shards to the gpu would not really reduce its dependence on `shuffle_n`. Intuitively, I would expect it to become more important (larger dataset requires more shuffling).

Somewhat related, `wd. WebDataset()` has a `shardshuffle=True` parameter that ensures that the shard list itself is fully shuffled.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -175,28 +212,218 @@ def __getitem__(self, index):
...
 
-        batch_dict = {
-            "dataset_names": dataset_name_list,
-            "audio_filepaths": audio_filepath_list,
-            "audio": batch_audio,
-            "audio_lens": batch_audio_len,
+        self._dataset = wd.WebDataset(audio_tar_filepaths, nodesplitter=None)
+
+        if shuffle_n > 0:
+            self._dataset = self._dataset.shuffle(shuffle_n)
```
rlangman, line 349: Should we do something like` dataset.shuffle(size=shuffle_n, initial=shuffle_n)`? The initial buffer size of 100 is really small, which I saw slow convergence a lot on my small test datasets (but might be irrelevant on very large datasets).

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -71,7 +132,7 @@ class VocoderDataset(Dataset):
     def __init__(
         self,
         dataset_meta: Dict,
-        sample_rate: int,
+        sample_rate: int = 24000,
```
rlangman, line 135: Nitpick: other classes in the TTS collection either provide no default sample_rate or defaults to 22050.

nithinraok: Default for our audio codec models is 24000. Did we have a 22050 Hz model?

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: From webdataset Docs: 
https://webdataset.github.io/webdataset/gettingstarted/ 
`shuffle(n): shuffle the dataset with a buffer of size n; also shuffles shards (see below)`

nemo/collections/tts/data/vocoder_dataset.py
rlangman: No, but this is a data loader for all TTS vocoders. So I suppose I would favor not having a default sample rate.

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: If the set is small then yes number of samples per worker would become small and `shuffle` might not have an effect and wouln;t be a problem for larger sets as number of samples per worker would be large. As per [this](https://github.com/webdataset/webdataset/issues/71) discussion, it looks like if we want to get a true shuffle of over all sets, its better to unbatch() on loader and shuffle with a larger number.

nithinraok:
jenkins

nemo/collections/tts/data/vocoder_dataset.py
rlangman: Based on the thread linked below, it sounds like the comment is effectively saying "larger datasets are less sensitive to how they are shuffled".

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: updated

nithinraok:
jenkins

nemo/collections/tts/data/vocoder_dataset.py
nithinraok: updated to have both buffer and inital arguments to have more control over shuffling

rlangman (approved):
LGTM. Thanks a lot for the feature and changes.

PR #7881, opened 2023-11-13
add comment on script and fix target check

gshennvm:
# What does this PR do ?

fix issue with target checking in nemo -> mcore conversion
  
**PR Type**:
- [x] Bugfix


cuichenx (approved):
LGTM, thanks for fixing!

PR #7888, opened 2023-11-14
Distributed checkpoint averaging supports bf16 type

yidong72:
Add support of bf16 type.


Kipok (approved):
Thanks, looks good!

yidong72:
jenkins

PR #7904, opened 2023-11-17
Pad sequences to multiples of 16 for GPTSFTDataset

vysarge:
One-line change to pad sequences to the next highest multiple of 16 instead of 8 in GPTSFTDataset to allow this dataset's outputs to work with TransformerEngine which expects multiples of 16 when using fp8.


ericharper:
jenkins

ericharper (approved):
LGTM. Thanks!

PR #7916, opened 2023-11-20
[NLP] Raise informative error on FP16 model and trainer precision mismatch

janekl:
# What does this PR do ?

When trying to load FP16 model with `trainer.precision != 16` (or `16-mixed`) the user gets an mysterious error message like:
```
  File "/opt/NeMo/nemo/collections/nlp/models/language_modeling/megatron_base_model.py", line 122, in __init__
    self.model_parallel_config: ModelParallelConfig = self.build_model_parallel_config()
  File "/opt/NeMo/nemo/collections/nlp/models/language_modeling/megatron_base_model.py", line 782, in build_model_parallel_config
    "grad_scale_func": self.trainer.precision_plugin.scaler.scale
AttributeError: 'NoneType' object has no attribute 'scale'
```
This PR identifies the problem before and suggests a user what to fix to properly create the model.

**Collection**: NLP

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [x] Documentation

## Who can review?


nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -773,8 +773,12 @@ def build_model_parallel_config(self) -> ModelParallelConfig:
         cfg = OmegaConf.to_container(self.cfg, resolve=True)
 
         # map precision related configs
-        precision = cfg.get('precision', 32)  # PTL trainer precision
         megatron_amp_O2 = cfg.get('megatron_amp_O2', False)
+        if self.torch_dtype == torch.float16 and not self.trainer.precision.startswith("16"):
+            # Make sure that trainer.precision_plugin.scaler exists for config_mapping below
+            raise ValueError(
+                "Creating a half-precision model requires setting trainer.precision to '16' or '16-mixed'"
+            )
```
ericharper, line 781: Should we do the same check for bf16?

janekl: I think it is needed only for `model.precision=16` which is `self.torch_dtype` here, see [this](https://github.com/NVIDIA/NeMo/blob/08937c8e7dd2e782fac99c6c230ae0170c277700/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L112). More specifically, the check is to make sure [this will work](https://github.com/NVIDIA/NeMo/blob/08937c8e7dd2e782fac99c6c230ae0170c277700/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L793-L795) -- loading FP16 model assumes that a scaler is available in `trainer` so it needs to be configured properly.

In my understanding in other cases it is the trainer precision to handle training precision if `model.precision` is different, and this would result in necessary type casting. But I might have not considered all the possibilities.

Which case exactly you mean for **model.precision** and **trainer.precision**, respectively?

nemo/collections/nlp/models/language_modeling/megatron_base_model.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, cfg: DictConfig, trainer: Trainer, no_lm_init=True):
         self.tokenizer = None
 
         with open_dict(cfg):
-            if cfg.get('precision', None) is None and trainer is not None:
```
janekl, line 106: The latter condition is guaranteed by [the check in L93](https://github.com/NVIDIA/NeMo/blob/52d50e9e09a3e636d60535fd9882f3b3f32f92ad/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L93)

ericharper (approved):
LGTM. Thanks!

PR #7917, opened 2023-11-20
Remove deprecated arguments from TE's TransformerLayer

jbaczek:
# What does this PR do ?

Transformer engine doesn't use this arguments since v0.7 and now in v1.0 they removed them completely.

NLP

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x ] Bugfix
- [ ] Documentation

## Who can review?

@ericharper

### DIFF
--- a/nemo/collections/nlp/modules/common/megatron/transformer.py
+++ b/nemo/collections/nlp/modules/common/megatron/transformer.py
@@ -797,8 +797,6 @@ class AutocastTransformerLayer(TransformerLayer):
         params_dtype: torch.dtype = torch.float32,
         get_rng_state_tracker: Optional[Callable] = None,
         fuse_wgrad_accumulation: bool = False,
-        apply_query_key_layer_scaling: bool = False,
-        attention_softmax_in_fp32: bool = False,
         seq_length: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


aklife97:
the PR changes look fine but we might want to note that this only changes things for NeMo local GPT models, not m-core based ones which are the ones we majorly use now

ericharper:
I think the main issue is that user are used to configuring this from the config. Now TE expects an environment variable to be set. Mcore makes sure that the env var and the config are consistent. Can NeMo set the env var automatically so that the user doesn't have to set the env var?

aklife97:
yes, thats right. It would probably be better if we automatically set the TE env variable if someone passes it in the config

aklife97:
Ideally, I would prefer if this env var setting happens in m-core since m-core also relies on arg passing format and is the place where we actually interface with TE and setup TE/local layers. But if we feel that it might need broader discussion it might make sense to add this in nemo for now so that it unblocks the issue?

ShriyaPalsamudram:
@ericharper @aklife97 can you approve this for now in nemo so it unblocks?

ericharper (approved):
LGTM. Thanks!

PR #7940, opened 2023-11-27
SDE2: GPU acceleration

ssh-meister:
GPU acceleration added, 
Data loading optimization with adding multiprocessing,
Code refactoring


vsl9 (changes requested):
Sasha, can you please address the CodeQL issues?

PR #7952, opened 2023-11-29
Unfinished checkpoints handling

jbieniusiewi:
# What does this PR do ?

Handle incomplete checkpoints.

With this fix, if last checkpointing operation was interrupted for some reason and left incomplete data, training should be resumed from the last available full checkpoint, instead of trying to read the incomplete checkpoint and crashing. This should simplify training auto-resume logic.

Implementation overview:
Create "unfinished" marker file for a checkpoint that is being written or removed. 
The marker file is deleted after the checkpoint IO operation completes.
When resuming, checkpoints marked as unfinished are ignored and subsequently removed.

**Collection**:
Core

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

@ericharper, @titu1994, @blisc, @okuchaiev


ericharper:
jenkins

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -257,7 +297,52 @@ def _ema_callback(self, trainer: 'pytorch_lightning.Trainer') -> Optional[EMA]:
                 ema_callback = callback
         return ema_callback
 
+    @staticmethod
+    def format_checkpoint_unfinished_marker_path(checkpoint_filepath: Union[Path, str]) -> Path:
```
titu1994, line 301: Needs documentation of when the barrier after before should be used.

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -257,7 +297,52 @@ def _ema_callback(self, trainer: 'pytorch_lightning.Trainer') -> Optional[EMA]:
...
+            torch.distributed.barrier()
+        if is_global_rank_zero():
+            marker_path = NeMoModelCheckpoint.format_checkpoint_unfinished_marker_path(checkpoint_filepath)
+            marker_path.parent.mkdir(parents=True, exist_ok=True)
+            marker_path.touch()
+        if barrier_after and torch.distributed.is_initialized():
+            torch.distributed.barrier()
+
+    @staticmethod
+    def remove_checkpoint_unfinished_marker(
```
titu1994, line 331: Every function should be documented. Especially for core functions that work on save restore

titu1994 (approved):
Pretty neat way of detecting and handling incomplete checkpoints with a marker. 

Minor comments about needing to add docstring to the functions. 

One minor question - when you filter out the checkpoints in Exp Manager - what if you don't find any usable last checkpoint ? Do you just crash? Or restart training from random checkpoint ? I'd rather crash

nemo/utils/callbacks/nemo_model_checkpoint.py
jbieniusiewi: I extended comments to better explain the purpose of the barriers.

nemo/utils/callbacks/nemo_model_checkpoint.py
jbieniusiewi: Good point. Fixed.

athitten:
@jbieniusiewi PTL 2.1 creates a symlink for the last checkpoint so we wouldn't have an issue where saving last checkpoint is incomplete. Would this PR be still relevant with 2.1 ?
CC @titu1994 @ericharper

jbieniusiewi:
Hi, 
Can you please reopen this PR?
I would like to update it for current main.

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -271,14 +393,23 @@ def _save_checkpoint(self, trainer: 'pytorch_lightning.Trainer', filepath: str)
...
+        self.set_checkpoint_unfinished_marker(filepath, barrier_after=True)
         super()._remove_checkpoint(trainer, filepath)
         ema_callback = self._ema_callback(trainer)
         if ema_callback is not None:
             # remove EMA copy of the state dict as well.
             filepath = self._ema_format_filepath(filepath)
             super()._remove_checkpoint(trainer, filepath)
+        # barrier_before=True, so all ranks synchronize before removing the unfinished checkpoint marker
+        # we don't want to remove the marker until the checkpoint is actually removed.
+        self.remove_checkpoint_unfinished_marker(filepath, barrier_before=True)
```
athitten, line 412: @jbieniusiewi do we want to have `barrier_after` in `remove_checkpoint_unfinished_marker` ? Similarly `barrier_before` in case of `set_checkpoint_unfinished_marker`. I am not able to understand the use case of it, since we always want to have `barrier_after=true` for `set_checkpoint_unfinished_marker` and `barrier_before=true` for `remove_checkpoint_unfinished_marker`.

athitten (commented):
Overall the PR is good and very helpful. Thanks @jbieniusiewi. Left one comment.

athitten:
jenkins

nemo/utils/callbacks/nemo_model_checkpoint.py
jbieniusiewi: Hi @athitten 
Thank you for the review!

Please let me clarify the purpose of the barriers, with `_save_checkpoint`
```
 def _save_checkpoint(self, trainer: 'pytorch_lightning.Trainer', filepath: str) -> None:
        self.set_checkpoint_unfinished_marker(filepath, barrier_after=True)

        # (...) checkpoint writing code (...)
        
        self.remove_checkpoint_unfinished_marker(filepath, barrier_before=True)
```
The goal of `self.set_checkpoint_unfinished_marker(filepath, **barrier_after=True**)`
is to ensure that when ranks start to write the checkpoint, the [checkpoint] unfinished marker file is already put in the filesystem. In other words, any checkpointing writing starts **after** the marker is placed.
Without the barrier, some ranks might write some data, before the marker file is placed [by rank0]; in such case, if something crashed before the marker was actually placed, we would get incomplete data and no [unfinished] marker.

On the other hand `self.remove_checkpoint_unfinished_marker(filepath, **barrier_before=True**)`
ensures that all ranks finishes the checkpoint writing **before** we remove the unfinished marker file. Without the barrier, it would be possible that if checkpointing crashed after the marker was removed,  we would get incomplete data and no marker.

I hope this helps!

athitten:
jenkins

athitten (approved):
LGTM, can be merged once the CI passes.Thanks for addressing the comments @jbieniusiewi

athitten:
@titu1994 is this good to be merged or do you have any comments to be resolved ?

titu1994:
This will need to be rechecked for PTL 2.1 release when we update

PR #7962, opened 2023-12-02
Modify GPTSFTPackedDataset to respond to pad_to_max_length setting

vysarge:
Several-line change to make GPTSFTPackedDataset pad sequences to max length if pad_to_max_length=true.


vysarge:
Closed, will be added directly to chcui/packed_sft branch.

PR #7976, opened 2023-12-06
Draft: Cuda graphs greedy rnnt inference

galv:
# What does this PR do ?

Greedy RNNT inference currently sees the GPU idle 90% of the time because of CPU overheads. This experiments with using cuda graphs to do greedy inference, so that we can decoder at much higher throughputs. Very WIP. I began to work on this because I discovered that using TensorRT was not helpful at all, because it speeds up such a small portion of the code right now.

**Collection**: ASR

# Changelog 
- Will have to do after cleaning up

Not ready for serious review.


galv:
This is now at parity with the original blank_as_pad greedy RNN-T decoding implementation. Currently, I can decode librispeech test clean in 1 min, 39 seconds, while the original implementation runs in 2 min, 21 seconds, at batch size 16. This may seem disappointing, but it is clear that there is still CPU overhead being counted because loading data off disk is not fast enough. In addition, this result is using float32, not float16.

That is to say that the WER output here is 0:

https://github.com/NVIDIA/NeMo/pull/7976/files#diff-e87be798c4c213fbdf438595c4c06898e1dabea0757aaa60334331cbf9f44cb1R83

However, this still needs to be cleaned up. It has a lot of cruft, including an attempt to use TensorRT from before my realization that the problem was CPU overhead, not slowness in the GPU implementation.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,431 @@
...
+        hypotheses = [
+            rnnt_utils.Hypothesis(score=0.0, y_sequence=[], timestep=[], dec_state=None) for _ in range(batch_size)
+        ]
+
+        # At batch size 16, this for loop takes 25-31 milliseconds... How can we speed it up?
+        # It is likely that most of the overhead is in memory loads...
+        # Let's profile just this block
+        for i in range(batch_size):
+            j = 0
+            for t in range(out_len_cpu[i]):
```
artbataev, line 416: This can be potentially done on GPU
If I correctly understand what's going on here, the algorithm can be the following:
- for each sequence find first non-blank label index, possibly with torch.arange + mask, see https://stackoverflow.com/a/60202801
- make a mask for resulting symbols
- reshape (flatten) and take non-masked elements
- convert to list (`Tensor.tolist`)

PR #8027, opened 2023-12-14
Support torch jit script

artbataev:
# What does this PR do ?

Fix compatibility of the core classes (`NeuralModule`, `Exportable`, etc.) with `torch.jit.script`.
Currently inheriting from `NeuralModule` instead of `nn.Module` breaks `torch.jit.script` compiler.

**Collection**: [Core]

# Changelog 

# Usage

```python
class SimpleLinear(NeuralModule):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(10, 20, bias=False)

    def forward(self, x):
        return self.linear(x)

module = torch.jit.script(SimpleLinear())
x = torch.zeros(2, 10)
result = module(x)
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


artbataev:
jenkins

artbataev:
jenkins

artbataev:
jenkins

nemo/core/classes/exportable.py
```diff
@@ -243,19 +244,19 @@ def _export(
         return (output, output_descr, output_example)
 
     @property
-    def disabled_deployment_input_names(self):
+    def disabled_deployment_input_names(self) -> List[str]:
```
titu1994, line 247: Set to list needs to be confirmed by @borisfom

nemo/core/neural_types/elements.py
```diff
@@ -69,28 +70,35 @@ class ElementType(ABC):
...
+    def type_parameters(self) -> Dict[str, Any]:
         """Override this property to parametrize your type. For example, you can specify 'storage' type such as
         float, int, bool with 'dtype' keyword. Another example, is if you want to represent a signal with a
         particular property (say, sample frequency), then you can put sample_freq->value in there.
         When two types are compared their type_parameters must match."""
         return {}
 
     @property
-    def fields(self) -> Optional[Tuple]:
+    def fields(self):
```
titu1994, line 91: This should be optional tuple no ?

nemo/core/neural_types/elements.py
```diff
@@ -69,28 +70,35 @@ class ElementType(ABC):
...
         """This should be used to logically represent tuples/structures. For example, if you want to represent a
         bounding box (x, y, width, height) you can put a tuple with names ('x', y', 'w', 'h') in here.
         Under the hood this should be converted to the last tesnor dimension of fixed size = len(fields).
         When two types are compared their fields must match."""
         return None
 
     def compare(self, second) -> NeuralTypeComparisonResult:
+        if torch.jit.is_scripting():
+            # Suppress in torch.jit.script
+            return NeuralTypeComparisonResult.SAME
```
titu1994, line 101: Why should this be hard coded ? It will give incorrect result for any scripted model ?

nemo/core/neural_types/elements.py
```diff
@@ -69,28 +70,35 @@ class ElementType(ABC):
     We are relying on Python for inheritance checking"""
 
     def __str__(self):
+        if torch.jit.is_scripting():
```
titu1994, line 73: How about just adding one init in this class and no need to put init in all subclasses ?

nemo/core/neural_types/elements.py
```diff
@@ -140,49 +152,93 @@ class ChannelType(ElementType):
     """Element to represent convolutional input/output channel.
     """
 
+    def __init__(self):
+        """Dummy init for TorchScript compatibility"""
```
titu1994, line 156: These classes should call super not pass

titu1994 (commented):
Most of this PR seems quite reasonable. I would like another review from a core maintainer, but only minor comments for now

nemo/core/neural_types/elements.py
artbataev: This is the correct type annotation, but it is not supported by TorchScript:
- TorchScript supports only full annotation. `Optional[Tuple]` does not have an underlying type (error in TorchScript)
- TorchScript does not support tuples of an arbitrary size (only lists). Adding underlying type will result in an error in TorchScript, thus I prefer to delete annotation here

nemo/core/neural_types/elements.py
artbataev: Thanks, removed (now suppression here is redundant, this was an artifact from the first attempt)

nemo/core/neural_types/elements.py
artbataev: Good suggestion, but unfortunately this is not supported in TorchScript(
Inability to call `super().method()` is a known limitation, still not fixed https://github.com/pytorch/pytorch/issues/42885

nemo/core/neural_types/elements.py
artbataev: See comment about `super` above. I can change everywhere, if you think it's a better solution: 

```python
def __init__(self):
    """Dummy init for TorchScript compatibility"""
    if not torch.jit.is_scripting():
        # torch.jit does not support super() call
        super().__init__()
```

artbataev:
jenkins

nemo/core/neural_types/elements.py
titu1994: Ok, no need to change then

nemo/core/neural_types/elements.py
titu1994: Yes this is better

titu1994:
Tests seem to have failed, could you check ?

artbataev:
jenkins

artbataev:
jenkins

artbataev:
jenkins

nemo/core/neural_types/elements.py
artbataev: Done

artbataev:
> Tests seem to have failed, could you check ?

All tests are passed. There were some weird issues with connection unrelated to this PR.

titu1994 (commented):
Looks good to me @okuchaiev can you review and approve ?

ericharper (approved):
LGTM. Thanks!

PR #8084, opened 2023-12-26
fix device setting to allow using accelerator cpu

orena1:
# What does this PR do ?

It is currently impossible to run wav2vec with `trainer.accelerator=cpu` if the machine has a gpu. 
The padding_mask is already on the device set in `trainer.accelerator` but if the machine has a GPU the DEVICE variable will contain cuda and this error message will be received:

```
File /mnt/NVM/oren/anaconda3_10/envs/nemo/lib/python3.10/site-packages/nemo/collections/asr/modules/wav2vec_modules.py:348, in Wav2VecTransformerEncoder.create_padding_mask(self, length)
    345 padding_mask = torch.arange(max_len, device=DEVICE)
    347 # Switch to binary for transformer, 1 for valid tokens, 0 for padding
--> 348 padding_mask = (padding_mask.expand(len(length), max_len) < length.unsqueeze(1)).type(torch.uint8)
    350 return padding_mask

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

```
  
**PR Type**:
- [x] Bugfix

# Additional Information
* Related to # (issue)
* 
@titu1994, @redoctopus, @jbalam-nv, or @okuchaiev

### DIFF
--- a/nemo/collections/asr/modules/wav2vec_modules.py
+++ b/nemo/collections/asr/modules/wav2vec_modules.py
@@ -32,7 +32,6 @@ from nemo.collections.nlp.modules.common.transformer import TransformerEncoder
 from nemo.core.classes.module import NeuralModule
 from nemo.core.neural_types import AcousticEncodedRepresentation, AudioSignal, LengthsType, NeuralType, SpectrogramType
 
-DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
 
 
 class TransposeLast(torch.nn.Module):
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
Makes sense, thanks for the fix ! @tbartley94 for final review

titu1994:
jenkins

orena1:
Thanks @titu1994 , see also here:
https://github.com/NVIDIA/NeMo/issues/8085

titu1994:
jenkins

tbartley94:
@titu1994 i know you pushed but just for record I'm not seeing any issues so review is approved

titu1994:
Oh thanks for the update, I think I misremembered as this having your approval on slack somehow, turns out it was a different topic different pr. Sorry about that.

orena1:
do you @titu1994  or @tbartley94 have any idea about this one:
https://github.com/NVIDIA/NeMo/issues/8085

tbartley94:
Hmm, lemme take a look. I'd wager it's some tech debt between PTL versions. (fervor for wav2vec dipped before PTL 2.0 was released, so the code is likely working off a dated namespace.)

PR #8087, opened 2023-12-27
Add text metrics to asr eval

stevehuang52:
# What does this PR do ?

Add some text metrics to ASR evaluator to support AST models


nemo/collections/asr/parts/utils/eval_utils.py
```diff
@@ -15,10 +15,18 @@
 import json
 from typing import Optional, Tuple, Union
 
+from torchmetrics.text import SacreBLEUScore
```
titu1994, line 18: These need to be guarded because we don't always install sacrebleu and rouge

stevehuang52: won't the dependencies be installed when installing torchmetrics? I didn't know where torchmetrics requires sacrebleu and rouge, thought it uses its own implementaion

titu1994: I don't want to take a chance with those two. They are annoying libraries to install on other machines than Linux

stevehuang52:
jenkins

nemo/collections/asr/parts/utils/eval_utils.py
stevehuang52: I searched through torchmetrics dependencies and didn't find sacrebleu or rouge. I also looked into its implementation, the SacreBLEUScore implementation doesn't need other dependencies, while the ROUGEScore implementation only needs NLTK, so I think it's safe to add those two metrics?

titu1994 (commented):
Nltk has issues on clusters, I'll send you a thread on slack later. Make the class with imports nltk run only inside of the place that's needed, not at the top of the file.

nemo/collections/asr/parts/utils/eval_utils.py
titu1994: Nltk has issues on clusters, I'll send you a thread on slack later. Make the class with imports nltk run only inside of the place that's needed, not at the top of the file.

titu1994 (approved):
Nevermind above comment, nltk itself is fine, it was another library that was acting poorly by installing nltk stuff at import time. 
Lgtm

nemo/collections/asr/parts/utils/eval_utils.py
titu1994: Nvm above. Nltk itself is fine.

titu1994:
jenkins

Issue #8116, opened 2024-01-03
Installation error on M1 Macbook - No matching distribution found for triton


stephenmcconnachie:
I am installing NeMo on M1 Macbook, following the installation guide for M1:
https://github.com/NVIDIA/NeMo?tab=readme-ov-file#mac-computers-with-apple-silicon

All works fine in Conda environment, up to running ./reinstall.sh. After a few successful packages installed I get this error:

```
INFO: pip is looking at multiple versions of nemo-toolkit to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement triton (from nemo-toolkit) (from versions: none)
ERROR: No matching distribution found for triton
```

The triton compatibility statement on the GitHub readme says Linux only:
https://github.com/openai/triton#compatibility

Is there a way to work around this in the M1 Mac installation?

- OS version: 14.1.2 (23B92)
- PyTorch version: 2.1.2 
- Python version: 3.11.0

titu1994:
Triton dependency comes from pytorch 2.0 - you can try to install pytorch CPU first and then install NeMo instead of installing everything at the same time

titu1994:
Btw only the asr toolkit can be guaranteed installed on MacBook. Don't use reinstall.sh, use pip install nemo_toolkit[asr]

titu1994:
Those instructions were fine for pytorch 1.x but triton complicates things

stephenmcconnachie:
Thank you so much for your help - that worked with `pip 'install nemo_toolkit[asr]'`

PR #8122, opened 2024-01-04
README edit to change Apple Silicon install instructions (to fix a break introduced by pytorch 2)

stephenmcconnachie:
# What does this PR do ?

This PR aims to correct a small error in the Apple Silcon installation instructions in README, where a triton dependency is problematic due to pytorch 2. The use of `/reinstall.sh` doesn't work, so using `pip install` instead. Also added a note about MacBook, to clarify that only asr toolkit is supported.

# Additional Information
* Related to #8116

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -274,7 +274,7 @@ To install NeMo on Mac with Apple M-Series GPU:
     cd NeMo
     pip install 'nemo_toolkit[all]'
 
-    Note that only the ASR toolkit is guaranteed to work on MacBook - so for MacBook use pip install 'nemo_toolkit[asr]'
+    # Note that only the ASR toolkit is guaranteed to work on MacBook - so for MacBook use pip install 'nemo_toolkit[asr]'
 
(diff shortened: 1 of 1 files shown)


titu1994 (approved):
Thanks !

TomTom101:
git clone and cd'ing into NeMo became obsolete when installing from pypi. Imo those commands can be removed?
```shell
git clone https://github.com/NVIDIA/NeMo
cd NeMo
```
In my case, I only get v1.21 installed from pypi. With `pip install 'nemo-toolkit[asr]==1.22.0'` I again run into an triton dependency issue:
`ERROR: Could not find a version that satisfies the requirement triton (from nemo-toolkit) (from versions: none)`

TomTom101:
According to [this comment](https://github.com/openai/triton/issues/1057#issuecomment-1500691831), macOS is not supported by triton at all.
What eventually _did_ work for me was removing `triton` from the requirements altogether

PR #8129, opened 2024-01-05
Use latest apex internal API

jbaczek:
# What does this PR do ?

Apex _fast_layer_norm added mandatory argument `memory_efficient`. This PR fixes call to this function to accommodate to the change.

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x ] Bugfix
- [ ] Documentation

## Who can review?

@mikolajblaz can you review this one? I can see that you introduced this line 17 months ago. Can we move to the APEX implementation entirely as suggested in the TODO? 

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/modules/common/megatron/layer_norm_1p.py
+++ b/nemo/collections/nlp/modules/common/megatron/layer_norm_1p.py
@@ -40,7 +40,7 @@ if HAVE_APEX:
             torch.nn.init.zeros_(self.bias)
 
         def forward(self, x):
-            return _fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon)
+            return _fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon, memory_efficient=False)
 
 
 else:


mikolajblaz (approved):
Looks good. Apex version still has no LayerNorm**1p** version so can't switch completely

nemo/collections/nlp/modules/common/megatron/layer_norm_1p.py
```diff
@@ -40,7 +40,7 @@ def reset_parameters(self):
             torch.nn.init.zeros_(self.bias)
 
         def forward(self, x):
-            return _fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon)
+            return _fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon, memory_efficient=False)
```
ShriyaPalsamudram, line 43: why is this hardcoded to False? Instead it can be run by just passing self.memory_efficient as a parameter right?

jbaczek: AFAIU memory_efficient=False is the previous implementation of _fast_layer_norm. As non memory efficient implementation has been tested I left it like this.
We can make it switchable but I think it would be beneficial to test it first. What do you think?

ShriyaPalsamudram: Sounds good, we can keep it as it is

ShriyaPalsamudram (approved):

layalir (approved):

PR #8130, opened 2024-01-05
add war fix for sync issues

gshennvm:
# What does this PR do ?

this is a war around a deeper issue where checkpointing is not guarded by pytorch distributed calls. So sometimes it removes checkpoints that don't exist and rmtree errors


gshennvm:
jenkins

cuichenx (approved):
LGTM thanks. Let's plan to fix the underlying issue soon

PR #8139, opened 2024-01-09
G2P CTC - allow setting vocab_file from config

ppisljar:
# What does this PR do ?

ctc G2P model always generates the vocab_file which is only suitable for english. This allows you to set custom vocab_file thru config.

**Collection**: TTS/G2p

# Changelog 
- allow passing custom vocab file to grapheme_tokenizer


ppisljar:
jenkins

ppisljar:
@titu1994 what should i do to move this forward ? thank you.

titu1994:
G2P would be @rlangman 
Plus, we cannot write to file inside of training step or val step. File writing must be done offline somewhere else cause ddp will introduce too many cases to it.

ppisljar:
file writing already happened here, i didnt change anything about it, except i skip file writing if dict is provided, so i guess for DDP this should work if file is provided and not generated, so i think this should be an improvement for tht case as well

ekmb (approved):

ekmb:
jenkins

nemo/collections/tts/g2p/models/ctc.py
```diff
@@ -125,10 +125,14 @@ def setup_grapheme_tokenizer(self, cfg):
...
-            vocab_file = "/tmp/char_vocab.txt"
-            with open(vocab_file, "w") as f:
-                [f.write(f'"{ch}"\n') for ch in chars]
-                f.write('"\\""\n')  # add " to the vocab
+            if cfg.tokenizer_grapheme.vocab_file:
+                vocab_file = cfg.tokenizer_grapheme.vocab_file
+            else:
+                vocab_file = "/tmp/char_vocab.txt"
+
+                with open(vocab_file, "w") as f:
```
ekmb, line 133: we actually do need to have this for a proper model restoration [tests are failing], could you move lines 133-135 outside of `else`?

ekmb (changes requested):

nemo/collections/tts/g2p/models/ctc.py
ppisljar: shouldn't restoration work as the tokenizer_grapheme.vocab_file is not set ? i will look into it,
but just moving the lines out of if file would make the if do nothing.

ppisljar:
i hope you can reopen this, i was still waiting for the review, as far as i am aware this is ready to merge

ekmb:
jenkins

ekmb:
jenkins

ekmb:
jenkins

ekmb:
jenkins

ekmb:
@ppisljar [this test](https://github.com/NVIDIA/NeMo/blob/main/Jenkinsfile#L1216) is still failing

PR #8191, opened 2024-01-17
Dgalvez/cuda graphs greedy rnnt inference squash

galv:
# What does this PR do ?

Speeds up RNN-T greedy decoding greatly by eliminating the 90% of the time that the GPU is idle, waiting on the CPU, via cuda graphs with conditional nodes. 

Here are some results for transcribing librispeech test other, a 5.4 hour dataset, on an A100, with bfloat16, at batch size 16:

image

You can see that we get a 3.125x speed up with a 600 million parameter model, and a 2.65x speedup with a 1.1 billion parameter model.

This benchmark comes from running the following. Note that I exclude the time required to create the cuda graph from the timing measurement. This fits an inference use case where that is a one time task.

```
for model_name in stt_en_fastconformer_transducer_xlarge nvidia/parakeet-rnnt-1.1b; do
for use_cuda_graph_decoder in false true; do
for amp in true; do
for batch_size in 16; do
echo "GALVEZ: ${model_name} fast:${use_cuda_graph_decoder}"
$(which python) examples/asr/speech_to_text_eval.py  pretrained_name=$model_name dataset_manifest=/home/dgalvez/scratch/data/test_other.json  batch_size=$batch_size  output_filename=test_other_decoded.js\
onl  amp=$amp  amp_dtype=bfloat16  rnnt_decoding.greedy.use_cuda_graph_decoder=$use_cuda_graph_decoder  use_cer=false num_workers=1
done
done
done
done
```

This is a squashing of #7976 . I wanted to squash that, but unfortunately I reference a few commits in various bugs I filed, and didn't want the links to break.

**Collection**: ASR. Adds some utils for cuda-python to common.

# Changelog 
- Adds a new class `RNNTGreedyDecodeCudaGraph`, which uses cuda graphs with conditional nodes to remove the CPU overhead.

# Usage

```bash
python examples/asr/speech_to_text_eval.py \
  pretrained_name=stt_en_fastconformer_transducer_xlarge \
  dataset_manifest=/home/dgalvez/scratch/data/test_other.json \
  batch_size=16 \
  output_filename=test_other_decoded.jsonl \
  amp=true  amp_dtype=bfloat16 \
  rnnt_decoding.greedy.loop_labels=false \
  rnnt_decoding.greedy.use_cuda_graph_decoder=true \
  use_cer=false num_workers=1
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #7976

### DIFF
--- a/nemo/collections/asr/modules/rnnt_abstract.py
+++ b/nemo/collections/asr/modules/rnnt_abstract.py
@@ -282,7 +282,7 @@ class AbstractRNNTDecoder(NeuralModule, ABC):
     def batch_replace_states_mask(
         cls, src_states: list[torch.Tensor], dst_states: list[torch.Tensor], mask: torch.Tensor,
     ):
-        """Replace states in dst_states with states from src_states using the mask"""
+        """Replace states in dst_states with states from src_states using the mask, in a way that does not synchronize with the CPU"""
         raise NotImplementedError()
 
     def batch_split_states(self, batch_states: list[torch.Tensor]) -> list[list[torch.Tensor]]:
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -115,11 +115,23 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
     to decide both whether to enter the loop, and also whether to
     execute the next iteration of the loop).
     """
-    capture_status, _, graph, _, _ = cu_call(cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream))
+    capture_status, _, graph, _, _ = cu_call(
+        cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream)
+    )
     assert capture_status == cudart.cudaStreamCaptureStatus.cudaStreamCaptureStatusActive
 
     cuda.cuLaunchKernel(
-        while_loop_kernel, 1, 1, 1, 1, 1, 1, 0, torch.cuda.current_stream(device=device).cuda_stream, while_loop_args.ctypes.data, 0
+        while_loop_kernel,
+        1,
+        1,
+        1,
+        1,
+        1,
+        1,
+        0,
+        torch.cuda.current_stream(device=device).cuda_stream,
+        while_loop_args.ctypes.data,
+        0,
     )
 
     capture_status, _, graph, dependencies, _ = cu_call(
@@ -277,7 +289,9 @@ class RNNTGreedyDecodeCudaGraph:
 
         # Always create a new stream, because the per-thread default stream disallows stream capture to a graph.
         stream_for_graph = torch.cuda.Stream(self.device)
-        with torch.cuda.stream(stream_for_graph), torch.inference_mode(), torch.cuda.graph(self.graph, stream=stream_for_graph):
+        with torch.cuda.stream(stream_for_graph), torch.inference_mode(), torch.cuda.graph(
+            self.graph, stream=stream_for_graph
+        ):
             # This is failing...
             self.f = torch.zeros(
                 (self.batch_size, 1, self.encoder_output.shape[-1]),
@@ -343,7 +357,9 @@ class RNNTGreedyDecodeCudaGraph:
                     ],
                     dtype=np.uint64,
                 )
-                with with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle, self.device):
+                with with_conditional_node(
+                    while_loop_kernel, while_loop_args, while_loop_conditional_handle, self.device
+                ):
                     g, hidden_prime = self.caller._pred_step(
                         self.last_label.unsqueeze(1), hidden, batch_size=self.batch_size
                     )
@@ -415,9 +431,7 @@ class RNNTGreedyDecodeCudaGraph:
         if torch.is_autocast_enabled():
             x = x.to(torch.get_autocast_gpu_dtype())
 
-        if (max_time > self.max_time or
-            batch_size > self.batch_size or
-            self.device != x.device):
+        if max_time > self.max_time or batch_size > self.batch_size or self.device != x.device:
             # In the first two cases, we need to recreate the cuda
             # graph to handle larger tensor sizes. In the third case,
             # we need to recreate the graph, as well as all tensors,
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -26,14 +26,11 @@ except ImportError:
 from typing import List, Optional
 
 from nemo.collections.asr.parts.utils import rnnt_utils
-from nemo.core.classes import Typing, typecheck
-from nemo.core.neural_types import AcousticEncodedRepresentation, ElementType, HypothesisType, LengthsType, NeuralType
 from nemo.core.utils.cuda_python_utils import (
     assert_drv,
     check_cuda_python_cuda_graphs_conditional_nodes_supported,
     cu_call,
 )
-from nemo.utils import logging
 
 
 def run_nvrtc(kernel_string, kernel_name):
@@ -45,7 +42,9 @@ def run_nvrtc(kernel_string, kernel_name):
     # https://stackoverflow.com/questions/48283009/nvcc-get-device-compute-capability-in-runtime
     opts = []
     (err,) = nvrtc.nvrtcCompileProgram(prog, len(opts), opts)
+    assert_drv(err)
     err, size = nvrtc.nvrtcGetProgramLogSize(prog)
+    assert_drv(err)
     buf = b" " * size
     (err,) = nvrtc.nvrtcGetProgramLog(prog, buf)
     assert_drv(err)
@@ -178,7 +177,7 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
 
 
 class RNNTGreedyDecodeCudaGraph:
-    def __init__(self, max_symbols: int, cuda_device, caller):
+    def __init__(self, max_symbols: int, caller):
         if HAVE_CUDA_PYTHON:
             check_cuda_python_cuda_graphs_conditional_nodes_supported()
         else:
@@ -186,15 +185,15 @@ class RNNTGreedyDecodeCudaGraph:
 
         assert max_symbols is not None
 
-        self.symbols_added_t = torch.tensor(0, dtype=torch.int64, device=cuda_device)
-        self.max_symbols_t = torch.tensor(max_symbols, dtype=torch.int64, device=cuda_device)
         self.max_symbols = max_symbols
-        self.not_all_blank_t = torch.tensor(True, dtype=torch.bool, device=cuda_device)
 
-        self.cuda_device = cuda_device
-
-        self.time_idx_t = torch.tensor(0, dtype=torch.int64, device=self.cuda_device)
-        self.max_out_len_t = torch.tensor(0, dtype=torch.int64, device=self.cuda_device)
+        # These are cuda torch.Tensors which will be lazily allocated the first time _reinitialize() is called.
+        # We don't do it here because we don't know which cuda device we are using yet.
+        self.symbols_added_t = None
+        self.max_symbols_t = None
+        self.not_all_blank_t = None
+        self.time_idx_t = None
+        self.max_out_len_t = None
 
         self.encoder_output = None
         self.encoder_output_length = None
@@ -209,7 +208,6 @@ class RNNTGreedyDecodeCudaGraph:
         self.scores_cpu = None
         self.labels_cpu = None
         self.graph = None
-        self.graph_exec = None
 
         self.first_call = True
 
@@ -228,6 +226,14 @@ class RNNTGreedyDecodeCudaGraph:
             self.caller._greedy_decode_blank_as_pad_loop_frames(
                 encoder_output, encoder_output_length, encoder_output.device
             )
+
+            self.symbols_added_t = torch.tensor(0, dtype=torch.int64, device=encoder_output.device)
+            self.max_symbols_t = torch.tensor(self.max_symbols, dtype=torch.int64, device=encoder_output.device)
+            self.not_all_blank_t = torch.tensor(True, dtype=torch.bool, device=encoder_output.device)
+
+            self.time_idx_t = torch.tensor(0, dtype=torch.int64, device=encoder_output.device)
+            self.max_out_len_t = torch.tensor(0, dtype=torch.int64, device=encoder_output.device)
+
             self.first_call = False
 
         self.max_time = max(self.max_time, max_time)
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -13,14 +13,14 @@
 # limitations under the License.
 
 import contextlib
-from packaging.version import Version
 
 import numpy as np
 import torch
+from packaging.version import Version
 
 try:
-    from cuda import cuda, cudart, nvrtc
     from cuda import __version__ as cuda_python_version
+    from cuda import cuda, cudart, nvrtc
 
     HAVE_CUDA_PYTHON = True
 except ImportError:
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 import contextlib
+from packaging.version import Version
 
 import numpy as np
 import torch
 
 try:
     from cuda import cuda, cudart, nvrtc
+    from cuda import __version__ as cuda_python_version
 
     HAVE_CUDA_PYTHON = True
 except ImportError:
@@ -144,13 +146,19 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
     driver_params.conditional.handle = while_loop_conditional_handle
     driver_params.conditional.type = cuda.CUgraphConditionalNodeType.CU_GRAPH_COND_TYPE_WHILE
     driver_params.conditional.size = 1
-    # Work around until https://github.com/NVIDIA/cuda-python/issues/55 is fixed
-    driver_params.conditional.phGraph_out = [cuda.CUgraph()]
+    if Version(cuda_python_version) == Version("12.3.0"):
+        # Work around for https://github.com/NVIDIA/cuda-python/issues/55
+        # Originally, cuda-python version 12.3.0 failed to allocate phGraph_out
+        # on its own.
+        # This bug is fixed in cuda-python version 12.4.0. In fact, we can
+        # no longer write to phGraph_out in cuda-python 12.4.0, so we must
+        # condition on the version number.
+        driver_params.conditional.phGraph_out = [cuda.CUgraph()]
     (ctx,) = cu_call(cuda.cuCtxGetCurrent())
     driver_params.conditional.ctx = ctx
 
     # Use driver API here because of bug in cuda-python runtime API: https://github.com/NVIDIA/cuda-python/issues/55
-    # TODO: Change call to this after fix goes in:
+    # TODO: Change call to this after fix goes in (and we bump minimum cuda-python version to 12.4.0):
     # node, = cu_call(cudart.cudaGraphAddNode(graph, dependencies, len(dependencies), driver_params))
     (node,) = cu_call(cuda.cuGraphAddNode(graph, dependencies, len(dependencies), driver_params))
     body_graph = driver_params.conditional.phGraph_out[0]
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -228,7 +228,6 @@ class RNNTGreedyDecodeCudaGraph:
         self.caller = caller
 
     def _reinitialize(self, max_time, batch_size, encoder_output, encoder_output_length):
-        torch.cuda.nvtx.range_push("Init")
         if self.first_call:
             # We need to call the original _greedy_decode_blank_as_pad
             # implementation at least once beforehand in order to make
@@ -282,10 +281,6 @@ class RNNTGreedyDecodeCudaGraph:
         self.graph = None
 
         self.graph = torch.cuda.CUDAGraph()
-
-        torch.cuda.nvtx.range_pop()
-
-        torch.cuda.nvtx.range_push("build graph")
 
         # Always create a new stream, because the per-thread default stream disallows stream capture to a graph.
         stream_for_graph = torch.cuda.Stream(self.device)
@@ -402,8 +397,6 @@ class RNNTGreedyDecodeCudaGraph:
             self.last_label.fill_(self.caller._SOS)
             self.time_idx_t.fill_(0)
 
-        torch.cuda.nvtx.range_pop()
-
     def __call__(
         self,
         x: torch.Tensor,
@@ -440,17 +433,14 @@ class RNNTGreedyDecodeCudaGraph:
             # set self.first_call to True to make sure that all
             # possibly blocking initializers are initialized properly
             # again on the new device.
-            print("GALVEZ: reinit!")
             if self.device != x.device:
                 self.first_call = True
             self._reinitialize(max_time, batch_size, x, out_len)
 
-        torch.cuda.nvtx.range_push("Graph")
         self.encoder_output[: x.shape[0], : x.shape[1], ...].copy_(x)
         self.encoder_output_length[: out_len.shape[0]].copy_(out_len)
         self.graph.replay()
         torch.cuda.current_stream(device=self.device).synchronize()
-        torch.cuda.nvtx.range_pop()
 
         self.scores_cpu[self.labels_cpu == self.caller._blank_index] = 0.0
         total_scores = self.scores_cpu.sum(dtype=torch.float32, axis=(1, 2))
--- a/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
+++ b/nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
@@ -106,7 +106,7 @@ def create_inner_while_loop_kernel():
 
 
 @contextlib.contextmanager
-def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle):
+def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle, device):
     """
     Even though we add a conditional node only once, we need to
     capture the kernel that calls cudaGraphSetConditional() both
@@ -115,15 +115,15 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
     to decide both whether to enter the loop, and also whether to
     execute the next iteration of the loop).
     """
-    capture_status, _, graph, _, _ = cu_call(cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream().cuda_stream))
+    capture_status, _, graph, _, _ = cu_call(cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream))
     assert capture_status == cudart.cudaStreamCaptureStatus.cudaStreamCaptureStatusActive
 
     cuda.cuLaunchKernel(
-        while_loop_kernel, 1, 1, 1, 1, 1, 1, 0, torch.cuda.current_stream().cuda_stream, while_loop_args.ctypes.data, 0
+        while_loop_kernel, 1, 1, 1, 1, 1, 1, 0, torch.cuda.current_stream(device=device).cuda_stream, while_loop_args.ctypes.data, 0
     )
 
     capture_status, _, graph, dependencies, _ = cu_call(
-        cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream().cuda_stream)
+        cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream)
     )
     assert capture_status == cudart.cudaStreamCaptureStatus.cudaStreamCaptureStatusActive
 
@@ -145,14 +145,14 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
 
     cu_call(
         cudart.cudaStreamUpdateCaptureDependencies(
-            torch.cuda.current_stream().cuda_stream,
+            torch.cuda.current_stream(device=device).cuda_stream,
             [node],
             1,
             cudart.cudaStreamUpdateCaptureDependenciesFlags.cudaStreamSetCaptureDependencies,
         )
     )
-    body_stream = torch.cuda.Stream()
-    previous_stream = torch.cuda.current_stream()
+    body_stream = torch.cuda.Stream(device)
+    previous_stream = torch.cuda.current_stream(device=device)
     cu_call(
         cudart.cudaStreamBeginCaptureToGraph(
             body_stream.cuda_stream,
@@ -198,6 +198,8 @@ class RNNTGreedyDecodeCudaGraph:
         self.encoder_output = None
         self.encoder_output_length = None
         self.f = None
+        # We also lazily initialize a variable holding the current device
+        self.device = None
 
         # Reasonable default maximum time. 375 frames * (80ms / frame) = 30 seconds
         # 80ms is the frame size of recent fastconformer models
@@ -214,6 +216,7 @@ class RNNTGreedyDecodeCudaGraph:
         self.caller = caller
 
     def _reinitialize(self, max_time, batch_size, encoder_output, encoder_output_length):
+        torch.cuda.nvtx.range_push("Init")
         if self.first_call:
             # We need to call the original _greedy_decode_blank_as_pad
             # implementation at least once beforehand in order to make
@@ -223,9 +226,12 @@ class RNNTGreedyDecodeCudaGraph:
             # initializes things like a cudnnHandle_t via
             # cudnnCreate(), which can involve synchronizing with the
             # host. Such actions are not stream capturable to a graph.
-            self.caller._greedy_decode_blank_as_pad_loop_frames(
-                encoder_output, encoder_output_length, encoder_output.device
-            )
+            with torch.cuda.stream(torch.cuda.Stream(self.device)):
+                self.caller._greedy_decode_blank_as_pad_loop_frames(
+                    encoder_output, encoder_output_length, encoder_output.device
+                )
+
+            self.device = encoder_output.device
 
             self.symbols_added_t = torch.tensor(0, dtype=torch.int64, device=encoder_output.device)
             self.max_symbols_t = torch.tensor(self.max_symbols, dtype=torch.int64, device=encoder_output.device)
@@ -261,11 +267,18 @@ class RNNTGreedyDecodeCudaGraph:
             (self.batch_size, self.max_time, self.max_symbols), dtype=torch.int64, device="cpu", pin_memory=True
         )
 
+        self.graph = None
+
         self.graph = torch.cuda.CUDAGraph()
 
-        # Always create a new stream, because the per-thread default stream disallows stream capture to a graph.
-        with torch.cuda.stream(torch.cuda.Stream()), torch.inference_mode(), torch.cuda.graph(self.graph):
+        torch.cuda.nvtx.range_pop()
 
+        torch.cuda.nvtx.range_push("build graph")
+
+        # Always create a new stream, because the per-thread default stream disallows stream capture to a graph.
+        stream_for_graph = torch.cuda.Stream(self.device)
+        with torch.cuda.stream(stream_for_graph), torch.inference_mode(), torch.cuda.graph(self.graph, stream=stream_for_graph):
+            # This is failing...
             self.f = torch.zeros(
                 (self.batch_size, 1, self.encoder_output.shape[-1]),
                 dtype=encoder_output.dtype,
@@ -295,7 +308,7 @@ class RNNTGreedyDecodeCudaGraph:
             self.max_out_len_t = self.encoder_output_length.max()
 
             capture_status, _, graph, _, _ = cu_call(
-                cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream().cuda_stream)
+                cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=self.device).cuda_stream)
             )
             assert capture_status == cudart.cudaStreamCaptureStatus.cudaStreamCaptureStatusActive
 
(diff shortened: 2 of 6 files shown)


nithinraok:
jenkins

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,187 @@
...
+    nemo_model.preprocessor.featurizer.pad_to = 0
+
+    # Switch model to evaluation mode
+    nemo_model.eval()
+    # Freeze the encoder and decoder modules
+    nemo_model.encoder.freeze()
+    nemo_model.decoder.freeze()
+    nemo_model.joint.freeze()
+
+    audio_filepaths = glob.glob("/home/dgalvez/scratch/data/LibriSpeech/test-clean-processed/*.wav")
```
galv, line 54: @titu1994 I am looping over a dataset and making sure that the results are the same between the old implementation and the new one. Does that work for you as a test?

Secondly, how do you actually run tests on datasets? I'm depending on a local path on my machine right now, which is obviously not okay.

galv:
This PR is ready for review. @artbataev @titu1994 would you be willing?

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
titu1994: We don't run tests on datasets usually. We do have a few samples of AN4 inside the NeMo test dir, could you use that instead of hard coded dataset path? We only need to check if the text is somewhat similar over 4-5 files. 

See examples of how to access the test_data_dir pytest fixture and relative path to the AN4 files inside the ASR tests (I think the dataset ones specifically).

titu1994: As to how to run with datasets, cd to root of NeMo dir and run pytest tests/asr/ with any additional flags you want usually --relax_numba_compat is necessary

titu1994:
I can review this tomorrow @artbataev could you review it too ?

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
galv: Sounds good.

artbataev:
Cool, I will review the PR today or tomorrow.

@galv Please fix DCO (anyway, you will need to fix it for merging).

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+    {
+     cudaGraphSetConditional(handle, *time_idx < *trip_count);
+    }
+    """
+    return run_nvrtc(kernel_string, b"for_loop_conditional")
+
+
+# Observations: If cudaGraphSetConditional is true once, the kernel never completes...
+# The GPU is doing *something*. I'm just not sure what...
+# No way to query that at runtime...
```
titu1994, line 113: Uhh is this still an issue or was it resolved during development?

examples/asr/transcribe_speech.py
```diff
@@ -340,14 +340,19 @@ def autocast(dtype=None):
...
             else:
-                transcriptions = asr_model.transcribe(
-                    paths2audio_files=filepaths,
-                    batch_size=cfg.batch_size,
-                    num_workers=cfg.num_workers,
-                    return_hypotheses=cfg.return_hypotheses,
-                    channel_selector=cfg.channel_selector,
-                    augmentor=augmentor,
-                )
+                for _ in range(5):
```
artbataev, line 343: I agree that we should measure speed after warmup, but 5x transcriptions is too much by default.

I suggest to do something like:
- add `warmup_batches=0` flag to config
- warmup with `warmup_batches` batches (from `filepaths`)
- transcribe as usual

By default user will just get a transcription, but for measuring speed we can use e.g., `warmup_batches=10`.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+    ptx = np.char.array(ptx)
+    err, module = cuda.cuModuleLoadData(ptx.ctypes.data)
+    ASSERT_DRV(err)
+    err, kernel = cuda.cuModuleGetFunction(module, kernel_name)
+    ASSERT_DRV(err)
+
+    return kernel
+
+
+def create_outer_for_loop_kernel():
```
titu1994, line 96: Could you provide a short comment of those kernels as to what they are doing / what are they necessary for ?

titu1994: It will help in case they need to be modified in the future for some reason

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+    driver_params.type = cuda.CUgraphNodeType.CU_GRAPH_NODE_TYPE_CONDITIONAL
+    driver_params.conditional.handle = while_loop_conditional_handle
+    driver_params.conditional.type = cuda.CUgraphConditionalNodeType.CU_GRAPH_COND_TYPE_WHILE
+    driver_params.conditional.size = 1
+    # Work around until https://github.com/NVIDIA/cuda-python/issues/55 is fixed
+    driver_params.conditional.phGraph_out = [cuda.CUgraph()]
+    (ctx,) = cu_call(cuda.cuCtxGetCurrent())
+    driver_params.conditional.ctx = ctx
+
+    # Use driver API here because of bug in cuda-python runtime API: https://github.com/NVIDIA/cuda-python/issues/55
```
titu1994, line 155: Let's add #TODO comments here to remind to fix in future

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+        self.cuda_device = cuda_device
+
+        self.time_idx_t = torch.tensor(0, dtype=torch.int64, device=self.cuda_device)
+        self.max_out_len_t = torch.tensor(0, dtype=torch.int64, device=self.cuda_device)
+
+        self.encoder_output = None
+        self.encoder_output_length = None
+        self.f = None
+
+        # Reasonable default maximum time. 375 frames * 40ms / frame = 15 seconds
```
titu1994, line 211: Fast Conformer is 80 Ms per frame

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+    cudart.cudaStreamEndCapture(body_stream.cuda_stream)
+
+    torch.cuda.set_stream(previous_stream)
+
+
+class RNNTGreedyDecodeFast:
+    def __init__(self, max_symbols: int, cuda_device, caller):
+        assert max_symbols is not None
+
+        self.symbols_added_t = torch.tensor(0, dtype=torch.int64, device=cuda_device)
```
titu1994, line 197: These values, are they somehow reset after end of execution ?

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+    cuda.cuLaunchKernel(
+        while_loop_kernel, 1, 1, 1, 1, 1, 1, 0, body_stream.cuda_stream, while_loop_args.ctypes.data, 0
+    )
+
+    cudart.cudaStreamEndCapture(body_stream.cuda_stream)
+
+    torch.cuda.set_stream(previous_stream)
+
+
+class RNNTGreedyDecodeFast:
```
titu1994, line 193: @artbataev please document class docstring

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+        self.scores_cpu = torch.zeros(
+            (self.batch_size, self.max_time, self.max_symbols),
+            dtype=encoder_output.dtype,
+            device="cpu",
+            pin_memory=True,
+        )
+        self.labels_cpu = torch.zeros(
+            (self.batch_size, self.max_time, self.max_symbols), dtype=torch.int64, device="cpu", pin_memory=True
+        )
+
```
titu1994, line 255: Add comment this is start of decoding loop capture 

In general, document explicitly each step so that future read can keep track of what is going on  in case there is need to modify

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+        if torch.is_autocast_enabled():
+            x = x.to(torch.get_autocast_gpu_dtype())
+
+        # What do we do if batch_size is actually smaller for the
+        # input? Is this a problem? the clone() call will fail...
+        if max_time > self.max_time or batch_size > self.batch_size:
+            self._reinitialize(max_time, batch_size, x, out_len)
+
+        torch.cuda.nvtx.range_push("Graph exec")
```
titu1994, line 393: Let's remove (or comment out) nvtx calls in final release code

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+        torch.cuda.nvtx.range_pop()
+
+        torch.cuda.nvtx.range_push("labels")
+        valid_labels_mask = self.labels_cpu != self.caller._blank_index
+        labels_segments = valid_labels_mask.sum(axis=(1, 2))
+        labels_packed = self.labels_cpu[valid_labels_mask]
+        torch.cuda.nvtx.range_pop()
+
+        torch.cuda.nvtx.range_push("Convert to Hypotheses")
+        hypotheses = [
```
titu1994, line 424: Was this the section you mentioned you wanted to speedup with numba CPU ?

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -598,12 +599,20 @@ def __init__(
...
-            if loop_labels:
+            if loop_labels and use_cuda_graph_decoder:
+                raise ValueError("loop_labels and use_cuda_graph_decoder is unsupported configuration")
+            elif loop_labels:
                 # default (faster) algo: loop over labels
                 self._greedy_decode = self._greedy_decode_blank_as_pad_loop_labels
+            elif use_cuda_graph_decoder:
+                from nemo.collections.asr.parts.submodules.fast_rnnt_greedy_decoding import RNNTGreedyDecodeFast
+
+                self._greedy_decode = RNNTGreedyDecodeFast(max_symbols_per_step, torch.device("cuda"), self)
```
titu1994, line 615: So the init variables auto reset at the end of each `__call__` ?

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nemo.core.utils.cuda_python_utils import check_cuda_python_cuda_graphs_conditional_nodes_supported
+
+check_cuda_python_cuda_graphs_conditional_nodes_supported()
```
titu1994, line 17: Since this check raises an error of cuda python is not correct version, please move this check to the init of the RNNTGreedyDecodeFast

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+import contextlib
+import ctypes
+import time
+from dataclasses import dataclass, field
+from itertools import product
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+import torch
+from cuda import cuda, cudart, nvrtc
```
titu1994, line 28: It is not guaranteed that pytorch-cuda will be installed by user, please add a static try catch check 

try:
  from cuda import xyz
  HAVE_CUDA_PYTHON=True
except (Import exception, module not found write):
  HAVE_CUDA_PYTHON=False

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+from nemo.collections.asr.parts.utils import rnnt_utils
+from nemo.collections.asr.parts.utils.asr_confidence_utils import ConfidenceMethodConfig, ConfidenceMethodMixin
+from nemo.collections.common.parts.rnn import label_collate
+from nemo.core.classes import Typing, typecheck
+from nemo.core.neural_types import AcousticEncodedRepresentation, ElementType, HypothesisType, LengthsType, NeuralType
+from nemo.utils import logging
+
+
+def ASSERT_DRV(err):
+    if isinstance(err, cuda.CUresult):
```
titu1994, line 41: Inside any function that uses cuda python, check if HAVE_CUDA_PYTHON, and if not call check_cuda_python_cuda_graphs_conditional_nodes_supported() to give users a meaningful error.

titu1994 (commented):
Overall it looks decent but needs some minor fixes. Needs removal of nvtx, docstring/comments for complex functions, and dynamic import guard instead of directly importing cuda python without check. 

Apart from that, just wanted to note that is an incredibly complex work, serious kudos to developing this @galv

examples/asr/transcribe_speech.py
titu1994: I like this idea but actually let's not put benchmarking code inside a user facing inference script. We can keep this version internal for our analysis

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,187 @@
...
+from nemo.collections.asr.models.rnnt_bpe_models import EncDecRNNTBPEModel
+from nemo.collections.asr.parts.submodules.fast_rnnt_greedy_decoding import RNNTGreedyDecodeFast
+from nemo.collections.asr.parts.submodules.rnnt_greedy_decoding import GreedyBatchedRNNTInfer
+
+
+@pytest.mark.parametrize(
+    ("model_name", "batch_size", "use_subset", "enable_bfloat16"),
+    [
+        # The xfail is not catching for some reason...
+        # pytest.param("stt_en_fastconformer_transducer_large", 8, True, True, marks=pytest.mark.xfail(reason="Cannot instantiate graph with persistent RNN")),
```
artbataev, line 24: Is `xlarge` model different from `large`? Maybe add `TODO` for this?

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,187 @@
...
+    ],
+)
+def test_for_loop(model_name, batch_size, use_subset, enable_bfloat16):
+    nemo_model = ASRModel.from_pretrained(model_name, map_location="cuda")
+    conf = nemo_model.to_config_dict()
+    with open_dict(conf):
+        conf["decoding"]["greedy"]["max_symbols"] = 5
+        conf["decoding"]["greedy"]["loop_labels"] = False
+
+    with tempfile.NamedTemporaryFile() as fp:
```
artbataev, line 39: `change_decoding_strategy` should be enough. Saving/restoring models takes a lot of time, it's better to avoid serialization if possible in unit tests.

artbataev (commented):
Overall looks good to me. Agree with comments from @titu1994
@galv Please, clean up prints, unused imports, and make tests using data on our CI system (not local).

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+from nemo.collections.asr.modules import rnnt_abstract
+from nemo.collections.asr.parts.utils import rnnt_utils
+from nemo.collections.asr.parts.utils.asr_confidence_utils import ConfidenceMethodConfig, ConfidenceMethodMixin
+from nemo.collections.common.parts.rnn import label_collate
+from nemo.core.classes import Typing, typecheck
+from nemo.core.neural_types import AcousticEncodedRepresentation, ElementType, HypothesisType, LengthsType, NeuralType
+from nemo.utils import logging
+
+
+def ASSERT_DRV(err):
```
artbataev, line 40: Is this a universal approach, that can be reused in other code? If yes, is it worth moving this function and several functions below to a separate file?

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+        device: torch.device,
+        partial_hypotheses: Optional[List[rnnt_utils.Hypothesis]] = None,
+    ):
+
+        # Need to copy x and out_len into "staging buffers", or do a graph update.
+
+        if partial_hypotheses is not None:
+            raise NotImplementedError("`partial_hypotheses` support is not supported")
+
+        assert not self.caller.preserve_alignments
```
artbataev, line 378: Maybe `raise NotImplementedError` is better?

examples/asr/transcribe_speech.py
galv: Yeah, no benchmarking code in the final code, I agree.

nemo/collections/asr/modules/rnnt.py
```diff
@@ -1078,6 +1078,32 @@ def mask_select_states(
         # LSTM in PyTorch returns a tuple of 2 tensors as a state
         return states[0][:, mask], states[1][:, mask]
 
+    def batch_copy_states_mask(
```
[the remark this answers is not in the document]
galv: It's a bit strange, but basically mask_select_states does a host<->device synchronization. `states[0][:, mask]` creates a smaller tensor whose shape is determined by the number of True values in mask. This ends up doing a device to host copy of the number of 1's, so that a memory allocation can be done on host.

batch_copy_states does something similar, since indexing with the integer tensor `ids` also creates a new allocation. 

This is different from what I end up doing, which just rewrites some values based on the mask using torch.where()

Anyway I need to make an abstract method and implement this for the convolution-based prediction networks as well.

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
galv: Yes, the LSTM is different in terms of its parameter size. "large" triggers a persistent RNN kernel in cudnn, which is a cooperative kernel launch. And cooperative kernel launches are disallowed inside of the body graphs of conditional nodes in cuda graphs right now.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
titu1994: So the static try catch at the top is the global flag at python instantiation time whether the library could be imported or not. It is safe to import this file when if the bool is set to false and the library is not available. 

However a user might try to run cuda graph op without installing it, then at init of the class, which is instantiated dynamically at runtime, we can check the function to assert that the library is actually installed correctly, otherwise raise an appropriate error

titu1994: So yes, we'd need both. Static check outside unit inside file, dynamic check with function inside init.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
galv: It's not that simple, though. It's not just about whether we can `import cuda`. We also need to check that the appropriate version of cuda-python itself and the cuda driver are installed. https://github.com/NVIDIA/NeMo/pull/8191/files#diff-acab1a9f3d702862ddbe5720bfa6c7fd0a57f7c3dc0b59eb9878ed5cd1e3513aR28-R45

Maybe what you want in this case is something like https://github.com/NVIDIA/NeMo/blob/main/nemo/core/utils/k2_guard.py. We will simply expect those developing with cuda-python to do `from cuda_python_guard.cuda import ...` instead `from cuda import ...`

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,440 @@
...
+def run_nvrtc(kernel_string, kernel_name):
+    err, prog = nvrtc.nvrtcCreateProgram(str.encode(kernel_string), b"while_loop_conditional.cu", 0, [], [])
+
+    ASSERT_DRV(err)
+
+    # Compile program
+    # Not specifying --gpu-architecture will default us to a fairly low compute capability, which is a safe bet.
+    # Otherwise, there are ways to query the current device's compute capability.
+    # https://stackoverflow.com/questions/48283009/nvcc-get-device-compute-capability-in-runtime
+    opts = []
```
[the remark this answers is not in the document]
galv: No. This is a kernel that launches one thread inside one block just for the sake of calling a function: `cudaGraphSetConditional`

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
galv: resolved a long time ago. Sorry, I write comments to myself during development.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
galv: I agree. Thank you for pointing it out.

tests/collections/asr/decoding/test_fast_rnnt_greedy_decoding.py
galv: Thanks for the tip. It sped up test execution almost 3 times.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
```diff
@@ -319,6 +319,7 @@ def __init__(self, decoding_cfg, decoder, joint, blank_id: int):
                         preserve_frame_confidence=self.preserve_frame_confidence,
                         confidence_method_cfg=self.confidence_method_cfg,
                         loop_labels=self.cfg.greedy.get('loop_labels', False),
+                        use_cuda_graph_decoder=self.cfg.greedy.get('use_cuda_graph_decoder', False),
```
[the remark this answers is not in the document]
galv: I actually want to stick with my original name, "use_cuda_graph_decoder". My concern with "use_cuda_graphs" is that is rather vague. There is nothing stopping a user from wrapping existing nemo code (including this code, since cuda graphs can be children of other cuda graphs!) in a cuda graph, for example.

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
galv: I agree with you. I moved cu_call and assert_drv to cuda_python_utils.py. I didn't move the run_nvrtc() function at this time. It is not fully general (for example, it allows extracting only one compiled cuda kernel from a compiled module, but there is no reason why we shouldn't be able to extract multiple).

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,416 @@
...
+    )
+
+    cudart.cudaStreamEndCapture(body_stream.cuda_stream)
+
+    torch.cuda.set_stream(previous_stream)
+
+
+class RNNTGreedyDecodeCudaGraph:
+    def __init__(self, max_symbols: int, cuda_device, caller):
+        if not HAVE_CUDA_PYTHON:
```
galv, line 175: Fix this. error out `if not HAVE_CUDA_PYTHON`. And always check check_cuda_python_cuda_graphs_conditional_nodes_supported() afterwards.

nemo/collections/asr/modules/rnnt_abstract.py
```diff
@@ -342,3 +342,25 @@ def mask_select_states(self, states: Any, mask: torch.Tensor) -> Any:
             states filtered by mask (same type as `states`)
         """
         raise NotImplementedError()
+
+    def batch_copy_states_mask(
```
artbataev, line 346: You can use `batch_replace_states_mask` for this (for both stateless and lstm decoders) - it is in `main` branch

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,427 @@
...
+from nemo.core.utils.cuda_python_utils import (
+    assert_drv,
+    check_cuda_python_cuda_graphs_conditional_nodes_supported,
+    cu_call,
+)
+from nemo.utils import logging
+
+
+def run_nvrtc(kernel_string, kernel_name):
+    err, prog = nvrtc.nvrtcCreateProgram(str.encode(kernel_string), b"while_loop_conditional.cu", 0, [], [])
```
pzelasko, line 40: looks like this could be made generic

```suggestion
    err, prog = nvrtc.nvrtcCreateProgram(str.encode(kernel_string), str.encode(f"{kernel_name}.cu"), 0, [], [])
```

galv: The trouble with that suggestion is that it assumes a file contains only a single kernel inside. I actually already broke this assumption when I made two "timer kernels". One to record the time, and one to print out the time different from the recorded time. This also required a global `__device__` int64_t array of times as well. That code is not committed, since I used it just for debugging. But the point is that it is not necessarily correct that we can always assume one kernel per file. So I don't want to modify this code right now. The actual "file name" doesn't matter anyway.

nemo/collections/asr/modules/rnnt_abstract.py
galv: 👍 Thank you

nemo/collections/asr/parts/submodules/fast_rnnt_greedy_decoding.py
galv: Ended up doing it with pytorch, which was definitely fast enough. One tricky bit about numba was the lack of bfloat16 support. And we can end up with scores in bfloat16 if the model is running in bfloat16.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
galv: Not sure I understand your question, but I will try to respond. 

No they don't. I made a class with a `__call__` method so I could save some values that I need to initialize (like pinned CPU memory and the cuda graph itself) as instance variables, which do not need to be recreated across decoding calls. If they did have to, this would be very bad for throughput.

titu1994 (dismissed):
The PR looks great! Thanks! @artbataev please go through it once more and merge when ready

artbataev:
jenkins

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -2271,6 +2283,7 @@ class GreedyRNNTInferConfig:
     preserve_alignments: bool = False
     preserve_frame_confidence: bool = False
     confidence_method_cfg: Optional[ConfidenceMethodConfig] = field(default_factory=lambda: ConfidenceMethodConfig())
+    use_cuda_graph_decoder: bool = False
```
artbataev, line 2286: Please, add this parameter to `GreedyBatchedRNNTInferConfg`

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -615,6 +620,12 @@ def __init__(
                     preserve_frame_confidence=preserve_frame_confidence,
                     confidence_method_cfg=confidence_method_cfg,
                 )
+            elif use_cuda_graph_decoder:
+                from nemo.collections.asr.parts.submodules.cuda_graph_rnnt_greedy_decoding import (
+                    RNNTGreedyDecodeCudaGraph,
+                )
+
+                self._greedy_decode = RNNTGreedyDecodeCudaGraph(max_symbols_per_step, torch.device("cuda"), self)
```
artbataev, line 628: Is it possible to avoid hardcoding first cuda device here?
The problem is that I get an error when trying to transcribe on the other device (I have 2 GPUs on my local machine):

```shell
python examples/asr/speech_to_text_eval.py \
 pretrained_name=stt_en_fastconformer_transducer_large \
 dataset_manifest=~/code/asr_particles/manifests/librispeech/test_other.json \
 batch_size=16 \
 output_filename=test_other_decoded.jsonl \
 amp=false \
 amp_dtype=bfloat16 \
 ++rnnt_decoding.greedy.use_cuda_graph_decoder=true \
 rnnt_decoding.greedy.loop_labels=false \
 cuda=1
```
fails with "RuntimeError: Capture must end on the same stream it began on."

artbataev (commented):
@galv Minor comments. Please, also fix the CodeQL suggestions to clean up unused imports.
Major issue for now: is it possible to avoid hardcoding the device for decoding with the cuda graph decoder?

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,431 @@
...
+        # Reasonable default maximum time. 375 frames * (80ms / frame) = 30 seconds
+        # 80ms is the frame size of recent fastconformer models
+        # This does not affect correctness.
+        self.max_time = 375
+        self.batch_size = 0
+
+        self.scores_cpu = None
+        self.labels_cpu = None
+        self.graph = None
+        self.graph_exec = None
```
artbataev, line 212: is `self.graph_exec` used anywhere?

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,431 @@
...
+        if self.first_call:
+            # We need to call the original _greedy_decode_blank_as_pad
+            # implementation at least once beforehand in order to make
+            # sure that pytorch is "initialized". Pytorch may be
+            # uninitialized if this code runs before any other pytorch
+            # operation in this process. Pytorch often lazily
+            # initializes things like a cudnnHandle_t via
+            # cudnnCreate(), which can involve synchronizing with the
+            # host. Such actions are not stream capturable to a graph.
+            self.caller._greedy_decode_blank_as_pad_loop_frames(
```
artbataev, line 228: Can we just return the results from this greedy decoding call (just check `self.first_call` in `__call__`), and run `_reinitialize` only on the second call?

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
galv: Ooops. Thank you.

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
galv: I suppose so... but does it really matter?

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
galv: I'm looking into it. https://pytorch.org/docs/stable/tensor_attributes.html#torch.device says:

> If the device ordinal is not present, this object will always represent the current device for the device type, even after [torch.cuda.set_device()](https://pytorch.org/docs/stable/generated/torch.cuda.set_device.html#torch.cuda.set_device) is called; e.g., a [torch.Tensor](https://pytorch.org/docs/stable/tensors.html#torch.Tensor) constructed with device 'cuda' is equivalent to 'cuda:X' where X is the result of [torch.cuda.current_device()](https://pytorch.org/docs/stable/generated/torch.cuda.current_device.html#torch.cuda.current_device).

Maybe the current device is being changed at some point inside nemo...

galv: Okay, it's a little bit complicated...

https://github.com/NVIDIA/NeMo/issues/8436#issue-2137592510

I believe I have a workaround to get this through, though.

galv: Should be fixed after https://github.com/NVIDIA/NeMo/pull/8191/commits/2c2da1a77f0cbb1f6b8a79c31e1fe20d417fe953

artbataev:
jenkins

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
galv: You're right. I was super confused between GreedyRNNTInferConfig and GreedyBatchedRNNTInferConfig. I can't even find any uses of GreedyRNNTInferConfig in the code base. So it should probably be removed.

tbartley94:
jenkins

galv:
@artbataev Let me know if you think this is good to go.

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -0,0 +1,437 @@
...
+    void while_loop_conditional(cudaGraphConditionalHandle handle, const bool *not_blank, const long *symbols_added, const long *max_symbols)
+    {
+     cudaGraphSetConditional(handle, *not_blank && *symbols_added < *max_symbols);
+    }
+    """
+    return run_nvrtc(kernel_string, b"while_loop_conditional")
+
+
+@contextlib.contextmanager
+def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle):
```
artbataev, line 109: Is this context manager reusable? Can we move it to a separate file?
This is not a blocker, we can later move it to a separate file

artbataev (commented):
@galv, the issue still exists. See below.
Also, please change the PR name (e.g., "Cuda graphs rnnt inference") and description to match the current cude (e.g., `go_very_fast` is no more valid).

**Issue with `cuda != 0`**

I still see an error, but the error is different now. Did you test it? Do you need a multi-gpu machine?

I do not want to block the PR, but I think that it is a bug and should be fixed if possible.

Cuda 1 + cuda graphs: Fail
```shell
python examples/asr/speech_to_text_eval.py  pretrained_name=stt_en_fastconformer_transducer_large  dataset_manifest=test_other.json  batch_size=16  output_filename=test_other_decoded.jsonl  amp=false amp_dtype=bfloat16  rnnt_decoding.greedy.use_cuda_graph_decoder=true rnnt_decoding.greedy.loop_labels=false cuda=1
```

> ...
> in _reinitialize
> logp = self.caller._joint_step(self.f, g, log_normalize=None)[:, 0, 0, :]
> ...
> RuntimeError: CUDA error: operation not permitted when stream is capturing
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
> During handling of the above exception, another exception occurred:
> ...
> RuntimeError: Capture must end on the same stream it began on.

Cuda 0 + cuda graphs: OK
```shell
python examples/asr/speech_to_text_eval.py  pretrained_name=stt_en_fastconformer_transducer_large  dataset_manifest=test_other.json  batch_size=16  output_filename=test_other_decoded.jsonl  amp=false amp_dtype=bfloat16  rnnt_decoding.greedy.use_cuda_graph_decoder=true rnnt_decoding.greedy.loop_labels=false cuda=0
```

Cuda 1 + no cuda graphs (both loop frames/loop labels): OK

artbataev:
@galv as I see, the issue can be fixed when passing appropriate device with cuda streams:
```
def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle, device):
    ...
    body_stream = torch.cuda.Stream(device=device)
    previous_stream = torch.cuda.current_stream(device=device)

def _reinitialize(...):
    ...
    with torch.cuda.stream(torch.cuda.Stream(device=device)), torch.inference_mode(), torch.cuda.graph(self.graph):
    ... 
    capture_status, _, graph, _, _ = cu_call(
                cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream)
    )
```

galv:
@artbataev thank you for the initial suggestion. It works when the decoder has not been run yet. However, it doesn't work if the decoder has already been run. You can see my failing test here: https://github.com/NVIDIA/NeMo/commit/36b32738edccfda31b480f96e5b86ce28a6850d8

Clearly something obscure is happening here. The commit message provides more details. I've spent a few hours trying to debug this so I need to stop for the day.

artbataev:
@galv I tried some changes, and it seems I'm able to get it work.
But I'm not sure why these changes are required, and why everything works when creating graph first time on any device. Maybe there are some bugs in PyTorch.

1. Pass stream explicitly to `torch.cuda.graph`. After this change I'm able to run the test, but the final comparison of the results fails (the results seems to be incorrect for second run)
2. Pass device explicitly to all calls `torch.cuda.current_stream`. After this change the test is passed.

```python
# Always create a new stream, because the per-thread default stream disallows stream capture to a graph.
stream_for_graph = torch.cuda.Stream(self.device)
with torch.cuda.stream(stream_for_graph), torch.inference_mode(), torch.cuda.graph(self.graph, stream=stream_for_graph):
    ... # capture graph
   # pass device explicitly
   capture_status, _, graph, _, _ = cu_call(
       cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=self.device).cuda_stream)
   )
   ...

@contextlib.contextmanager
def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditional_handle, device):
    ...
    # pass device explicitly here and in other calls
    capture_status, _, graph, _, _ = cu_call(cudart.cudaStreamGetCaptureInfo(torch.cuda.current_stream(device=device).cuda_stream))
    ... 
```     

You can see the full commit here: https://github.com/artbataev/NeMo/commit/77fc36e199d5c804b447cef0c817340ef4e0334c

galv:
jenkins

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
galv: We can move it to a separate file, but I will leave that up to you with your own loop labels + cuda graphs PR.

galv:
jenkins

Previous failure seems to be a spurious failure caused by git clone failing.

@artbataev I incorporated your change. Thank you again. I made one more commit https://github.com/NVIDIA/NeMo/pull/8191/commits/fb2bd7aeb6a99540b7ca1d2334e1502ad989b8a7 as well which makes this work with cuda-python version 12.4.0 and greater. It turns out that the bug fix in that version makes the phGraph_out variable not writable. So I must not use my workaround when cuda-python > 12.3.0.

artbataev (dismissed):
Thanks a lot for the really speed-of-light decoding with Cuda graphs!

artbataev:
@galv I manually restarted Jenkins, but it is still waiting for an executor

galv:
jenkins

Sorry for missing the guard in that test. Hopefully things go through now.

artbataev (approved):

artbataev:
jenkins

PR #8212, opened 2024-01-22
Add support in Neural Typecheck to disable semantic checks

titu1994:
# What does this PR do ?

Adds support to temporarily / permanently (and globally) disable semantic type checks.

**Collection**: [Core]

# Changelog 
- Adds utility methods to disable global semantic type checking, or local context manager.
- Adds test to see if semantic types are properly ignored (but still input kwargs, shape and container checks)

# Usage

```python
import torch
from nemo.core import Typing, typecheck
from nemo.core.neural_types import *

class InputOutputTypes(Typing):
    @property
    def input_types(self):
        return {"x": NeuralType(('B',), LogprobsType())}

    @property
    def output_types(self):
        return {"y": NeuralType(('B',), LabelsType())}

    @typecheck()
    def __call__(self, x):
        x += 1
        return x

 # Provide input with wrong type after disabling semantic type checks
 with typecheck.disable_semantic_checks():
    input_data = torch.zeros(10)
    input_data.neural_type = NeuralType(('B',), LabelsType())  # Should be LogprobsType()
    result = obj(x=input_data)

    # assert that even if semantic types are disabled, output is attached with appropriate types
    assert result.sum() == torch.tensor(10.0)
    assert result.neural_type.compare(NeuralType(('B',), LabelsType())) == NeuralTypeComparisonResult.SAME

```

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -43,6 +43,7 @@ from nemo.utils.model_utils import import_class_by_path, maybe_update_config_ver
 __all__ = ['Typing', 'FileIO', 'Model', 'Serialization', 'typecheck', 'PretrainedModelInfo']
 
 _TYPECHECK_ENABLED = True
+_TYPECHECK_SEMANTIC_CHECK_ENABLED = True
 # TODO @blisc: Remove _HAS_HYDRA
 _HAS_HYDRA = True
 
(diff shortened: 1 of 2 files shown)


titu1994:
jenkins

ericharper (approved):
LGTM. Thanks!

nemo/core/classes/common.py
```diff
@@ -1114,3 +1125,26 @@ def disable_checks():
             yield
         finally:
             typecheck.set_typecheck_enabled(enabled=True)
+
+    @staticmethod
+    def set_semantic_check_enabled(enabled: bool = True):
```
nithinraok, line 1130: better to rename to `update_semantic_check` ?

nithinraok (approved):
Minor comment, LGTM.

nemo/core/classes/common.py
titu1994: No. It mirrors the set_enabled() API

PR #8239, opened 2024-01-25
Allow MegatronPretrainingRandomSampler to do multi-epoch training

trias702:
# What does this PR do ?

Allows the MegatronPretrainingRandomSampler class to correctly handle multi-epoch training by calculating length correctly when consumed_samples > total_samples

**Collection**: NLP

# Changelog 
- nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py: modified MegatronPretrainingRandomSampler to correctly return length for multi-epoch training when consumed_samples > total_samples
- nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py: fixed len calculation to be correct when drop_last=False, and also cleaned up constructor logic caused by these fixes: https://github.com/NVIDIA/NeMo/pull/7728/files#diff-26575f60355f81991aef1f635e947a368b086afe220c5a81817e35b1c00b2e38

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
@@ -147,16 +147,17 @@ class MegatronPretrainingRandomSampler(BaseMegatronSampler):
             pad_samples_to_global_batch_size=pad_samples_to_global_batch_size,
         )
         assert (
-            pad_samples_to_global_batch_size == False
+            not pad_samples_to_global_batch_size
         ), "`MegatronPretrainingRandomSampler` does not support sample padding"
+        if (not drop_last) and self.micro_batch_times_data_parallel_size > 1:
+            raise RuntimeError("`MegatronPretrainingRandomSampler` does not support drop_last=False when micro_batch_size * data_parallel_size > 1. \
+                                please reduce your MBS and data parallelism to 1 if you want to use drop_last=False, or switch to drop_last=True to avoid this error")
         self.last_batch_size = self.total_samples % self.micro_batch_times_data_parallel_size
         self.seed = seed
 
     def __len__(self):
         active_total_samples = self.total_samples - (self.last_batch_size if self.drop_last else 0)
-        num_available_samples = (
-            active_total_samples * (1 + (self.consumed_samples // active_total_samples))
-        ) - self.consumed_samples
+        num_available_samples = active_total_samples - self.consumed_samples % active_total_samples
         if self.global_batch_size is not None:
             if self.drop_last:
                 return num_available_samples // self.global_batch_size
--- a/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
@@ -152,7 +152,7 @@ class MegatronPretrainingRandomSampler(BaseMegatronSampler):
         if (not drop_last) and self.micro_batch_times_data_parallel_size > 1:
             raise RuntimeError(
                 "`MegatronPretrainingRandomSampler` does not support drop_last=False when micro_batch_size * data_parallel_size > 1. \
-                                please reduce your MBS and data parallelism to 1 if you want to use drop_last=False, or switch to drop_last=True to avoid this error"
+                  please reduce your MBS and data parallelism to 1 if you want to use drop_last=False, or switch to drop_last=True to avoid this error"
             )
         self.last_batch_size = self.total_samples % self.micro_batch_times_data_parallel_size
         self.seed = seed
--- a/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
+++ b/nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
@@ -134,6 +134,7 @@ class MegatronPretrainingRandomSampler(BaseMegatronSampler):
         drop_last: bool = True,
         global_batch_size: Optional[int] = None,
         pad_samples_to_global_batch_size: Optional[bool] = False,
+        seed: int = 0,
     ) -> None:
         super().__init__(
             total_samples=total_samples,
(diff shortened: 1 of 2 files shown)


trias702:
jenkins

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
```diff
@@ -149,9 +150,13 @@ def __init__(
             pad_samples_to_global_batch_size == False
         ), "`MegatronPretrainingRandomSampler` does not support sample padding"
         self.last_batch_size = self.total_samples % self.micro_batch_times_data_parallel_size
+        self.seed = seed
 
     def __len__(self):
-        num_available_samples: int = self.total_samples
+        active_total_samples = self.total_samples - (self.last_batch_size if self.drop_last else 0)
```
odelalleau, line 156: Probably that the same formula should be used in `__iter__()`?

Edit: although I'm afraid this would trigger the `assert current_epoch_samples % self.micro_batch_times_data_parallel_size == 0`... what about just adding a comment in `__iter__()` that `self.epoch` may be incorrect when `drop_last` is False (and not try to fix it in this PR unless you're confident in the fix, since this bug already exists anyway).

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
```diff
@@ -149,9 +150,13 @@ def __init__(
...
+        num_available_samples = (
+            active_total_samples * (1 + (self.consumed_samples // active_total_samples))
+        ) - self.consumed_samples
```
odelalleau, line 159: I find this formula hard to understand, while the following is more readable (IMO):
```
num_available_samples = active_total_samples - self.consumed_samples % active_total_samples
```

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
```diff
@@ -149,9 +150,13 @@ def __init__(
             pad_samples_to_global_batch_size == False
         ), "`MegatronPretrainingRandomSampler` does not support sample padding"
         self.last_batch_size = self.total_samples % self.micro_batch_times_data_parallel_size
+        self.seed = seed
 
     def __len__(self):
-        num_available_samples: int = self.total_samples
```
odelalleau, line 154: It seems like previously `__len__()` was ignoring `consumed_samples`. This seems like a bug to me (that this PR would fix) but it'd be good to get the confirmation from whoever wrote that class to make sure it's not intended.

nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
```diff
@@ -192,32 +193,21 @@ def __init__(
...
+            consumed_samples=consumed_samples,
+            micro_batch_size=micro_batch_size,
+            data_parallel_rank=data_parallel_rank,
+            data_parallel_size=data_parallel_size,
+            drop_last=drop_last,
+            global_batch_size=global_batch_size,
+            pad_samples_to_global_batch_size=pad_samples_to_global_batch_size,
+        )
+        assert (
+            pad_samples_to_global_batch_size == False
```
odelalleau, line 207:
```suggestion
            not pad_samples_to_global_batch_size
```

nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
```diff
@@ -226,7 +216,7 @@ def __len__(self) -> int:
             When `rampup_batch_size` is enabled, the return value can be not exactly precise.
 
         """
-        active_total_samples = self.total_samples - self.last_batch_size
+        active_total_samples = self.total_samples - (self.last_batch_size if self.drop_last else 0)
```
odelalleau, line 219: Same comment as above on the fact that the epoch in `__iter__()` won't be correct when `drop_last` is False and a comment may be needed.

odelalleau (commented):
Sorry, I had to take a look

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
trias702: I think that the way these samplers are built to work (sharding out to each DP), they can never work correctly with `drop_last=False` because then you do not have an equivalent shard length on each DP which will cause a hang. `drop_last=False` would only ever work correctly if DP=1.

Perhaps we should add an assert in the constructor which would only ever allow `drop_last=False` if and only if `data_parallel_size == 1`?

trias702: I have added a new raise exception which ensures nobody can use the class with `drop_last=False` and MBS * DP > 1. I think this solves all issues and circumvents the necessity of adding a comment under `__iter__`

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
trias702: Done, thanks

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
trias702: Will find out and check

nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
trias702: Done

nemo/collections/nlp/data/language_modeling/megatron/megatron_batch_samplers.py
trias702: I've added the new raise exception so you never end up running with `drop_last=False` and MBS * DP > 1

trias702:
jenkins

nemo/collections/nlp/data/language_modeling/megatron/data_samplers.py
odelalleau: From where I stand I agree with you, but I'm not sure exactly how these samplers are used (by other code / other users) so I'm afraid I can't really confirm whether it's ok to raise an exception in such a situation.

ericharper (approved):
LGTM. Thanks!

PR #8285, opened 2024-01-31
[canary] Test for CanaryTokenizer + refactoring

pzelasko:


nemo/collections/common/tokenizers/canary_tokenizer.py
```diff
@@ -96,3 +100,22 @@ def to_language_id(self, language):
             return token_id
 
         raise KeyError(f"Language {language} not found in tokenizer.")
+
+    @staticmethod
+    def build_special_tokenizer(output_dir: str | Path) -> SentencePieceTokenizer:
```
titu1994, line 105: Why this hardcoding of vocab size and other things? Convenience?

titu1994 (approved):
Overall looks good, agreed that we need to make this much better for 2.0

PR #8345, opened 2024-02-06
Fix Megatron losses that were incorrect for micro_batch_size > 1

### LINKED ISSUE #8343: The loss of `MegatronGPTModel` is incorrect for `micro_batch_size` > 1
**Describe the bug**

The loss implemented [here](https://github.com/NVIDIA/NeMo/blob/13c1db4ffdaa3b2125ce6e66c93509a39d55b2cf/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L1169) is incorrect for `micro_batch_size` > 1 because it computes the average over all tokens in the micro-batch, instead of computing a per-sample loss then averaging it. This leads to inconsistent results when varying `micro_batch_size`.

Compare results with different values of `micro_batch_size` and notice they are very different.

**Expected behavior**

Varying `micro_batch_size` shouldn't affect results.

**Additional context**

I am creating this issue for visibility (in case someone else runs into the same problem), and will be submitting a PR to fix it.

odelalleau:
# What does this PR do ?

Fixes the losses of several Megatron models, that were incorrect for `micro_batch_size` > 1

**Collection**: nlp, multimodal

# Changelog 
- Fix losses of several Megatron models (`MegatronGPTModel`, `MegatronNevaModel`, `MegatronBertModel`, `MegatronLMEncoderDecoderModel`) that were incorrect for `micro_batch_size` > 1

# What still needs to be done

- [ ] Probably that some new tests are needed. I first want to check which tests fail to identify potential gaps in testing.
- [ ] There is a `# TODO does this also need fixing?` in `MegatronBertModel` as I'm not familiar with the Sequence Order Prediction loss. Would be helpful if someone could double check it, or at least tell me what are the shapes of `sop_logits` and `sentence_order` so I can figure it out.
- [ ] Check whether the handling of `context_parallel_size` > 1 is correct in `MegatronGPTModel`

# Other breaking changes

While fixing this problem I ran into some related issues that forced me to make some design choices leading to "collateral" breaking changes:

1. The validation loss in `MegatronGPTModel` with `validation_drop_last=False` was inconsistent with `drop_last=True` since it was computing a per-token average across the global batch (at least I believe that's what's happening [here](https://github.com/NVIDIA/NeMo/blob/9940ec600/nemo/collections/nlp/models/language_modeling/megatron_gpt_sft_model.py#L483)), while with `drop_last=True` it was using the (incorrect) per-token average across the micro-batch. I changed it so that the behavior should now be a consistent per-sample average across valid samples.

2. The loss could be NaN when all samples in a micro-batch would be fully masked (which typically would only happen with `micro_batch_size=1`). This also led to inconsistency between different micro-batch sizes. I changed it so that there is no NaN anymore, instead fully masked samples have a zero loss. The alternative would be to trigger a NaN as soon as one sample is fully masked, but I decided against it because (i) in order to handle validation with `validation_drop_last=False` we would need to make the loss function depend on this flag, which would make the code more complex (and potentially inconsistent in the presence of fully masked samples in the validation set), and (ii) NaN losses due to fully masked samples is a recurring pain point currently (it forces users to either go through a cumbersome model-dependent data filtering step, or hack the NeMo codebase to work around it).

3. (Follow-up to 2, splitting it for readability) That being said, I think that silently ignoring some samples is also potentially dangerous, so I added a warning when the loss mask is all zeros for some samples. I decided to do it at the dataset level rather than in the model so as to avoid forcing a CPU-GPU synch point on such a check. We could add a flag in the datasets to control whether this is an error or a warning, but I thought I'd get some feedback first.

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

Fixes #8343

Minor remark: I removed some `# TODO: add nemo version here` that seem to date back several years ago and don't make sense to me.

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -253,6 +253,8 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
         # TODO: add type hint once pip package is out
         self.transformer_config = self.build_transformer_config()
 
+        self._validation_drop_last = cfg.data.get('validation_drop_last', True)
+
         self.megatron_amp_O2 = cfg.get('megatron_amp_O2', False)
 
         self.mcore_gpt = cfg.get('mcore_gpt', False)
@@ -563,7 +565,7 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
 
         # only the last stages of the pipeline return losses
         if losses_reduced_per_micro_batch:
-            if (not forward_only) or self.cfg.data.get('validation_drop_last', True):
+            if (not forward_only) or self._validation_drop_last:
                 # average loss across micro batches
                 loss_tensors_list = [loss_reduced['avg'] for loss_reduced in losses_reduced_per_micro_batch]
                 loss_tensor = torch.concat(loss_tensors_list)
@@ -571,9 +573,9 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
             else:
                 # Get the total loss since micro batches sizes are not uniform
                 loss_sum_tensors_list = [
-                    loss_sum['loss_sum_and_ub_size']
+                    loss_sum['loss_sum_and_ub_num_valid']
                     for loss_sum in losses_reduced_per_micro_batch
-                    if loss_sum['loss_sum_and_ub_size'][1] > 0
+                    if loss_sum['loss_sum_and_ub_num_valid'][1] > 0
                 ]
                 loss_sum = (
                     torch.vstack(loss_sum_tensors_list).sum(axis=0)
@@ -914,9 +916,9 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
         return batch
 
     def get_batch_on_this_context_parallel_rank(self, batch):
-        num_valid_tokens_in_ub = None
-        if 'loss_mask' in batch and batch['loss_mask'] is not None:
-            num_valid_tokens_in_ub = batch['loss_mask'].sum()
+        num_valid_tokens_per_sample = None
+        if batch.get('loss_mask') is not None:
+            num_valid_tokens_per_sample = batch['loss_mask'].sum(dim=1)
 
         cp_size = parallel_state.get_context_parallel_world_size()
         if cp_size > 1:
@@ -935,7 +937,7 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
                     val = val.view(*val.shape[0:seq_dim], -1, *val.shape[(seq_dim + 2) :])
                     batch[key] = val
 
-        batch['num_valid_tokens_in_ub'] = num_valid_tokens_in_ub
+        batch['num_valid_tokens_per_sample'] = num_valid_tokens_per_sample
 
         return batch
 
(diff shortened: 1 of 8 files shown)


odelalleau:
jenkins

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -931,7 +933,7 @@ def get_batch_on_this_context_parallel_rank(self, batch):
                     val = val.view(*val.shape[0:seq_dim], -1, *val.shape[(seq_dim + 2) :])
                     batch[key] = val
 
-        batch['num_valid_tokens_in_ub'] = num_valid_tokens_in_ub
+        batch['num_valid_tokens_per_sample'] = num_valid_tokens_per_sample
```
yidong72, line 936: so `num_valid_tokens_per_sample` can be none if batch.get('loss_mask') is None? And I didn't see None handling in the loss calculation.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1162,11 +1159,12 @@ def on_test_epoch_end(self):
...
 
-    def loss_func(self, loss_mask, num_valid_tokens_in_ub, output_tensor):
+    def loss_func(self, loss_mask, num_valid_tokens_per_sample, output_tensor):
         losses = output_tensor.float()
-        loss_mask = loss_mask.view(-1).float()
-        # TODO: add nemo version here
-        loss = torch.sum(losses.view(-1) * loss_mask) / num_valid_tokens_in_ub  # sequence level nll
+        loss_mask = loss_mask.float()
+        # Compute (per-sample) sequence-level NLL. Fully masked samples have zero loss.
+        loss = torch.sum(losses * loss_mask, dim=1) / num_valid_tokens_per_sample.clamp(min=1)
```
yidong72, line 1166: For pretraining, we don't need per-sample average but per valid token average. Are we going to support these two types of losses?

yidong72 (commented):
need to support loss for both sft and pre-training.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
odelalleau: It shouldn't be None in practice -- I think `batch["loss_mask"]` can only be None on PP ranks that don't compute the loss. I didn't look at this very closely because I am not changing this specific aspect in this PR (it was the same before).

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
odelalleau: I'm not planning to implement per-valid-token average in this PR, though I agree it could be worth adding in the future. I'm not sure how pre-training uses the loss mask though, it seems to me it should be possible to concatenate data such that there is (almost) no data being masked out (but I may be wrong).

To be sure it's clear: note that the previous code was doing per-sample average with `micro_batch_size=1` (this PR isn't changing this).

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
```diff
@@ -783,10 +783,11 @@ def loss_func(self, loss_mask, tokens_loss):
         """
         This function takes as input per-token loss and masks non-required values.
         """
-        losses = tokens_loss.view(-1).float()
-        loss_mask = loss_mask.view(-1).float()
-        # TODO: add nemo version here
-        loss = torch.sum(losses * loss_mask) / loss_mask.sum()  # sequence level nll
+        losses = tokens_loss.float()
```
jbaczek, line 786: Can you add more verbose doc string? It would be nice to know what shapes are expected here and which dimension correspond to what.

odelalleau: Indeed, this is always a good idea => done in 97e9d05486042b586e121b3fd21751d0fe452041

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
yidong72: okay. it makes sense.  The none can happen for intermediate PP ranks.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
yidong72: hmm, before it calculated the per token average per micro-batch. I think it is never per-token loss for the whole batch. It is a bug for pretraining that we need to fix. is there a separate PR to address it?

nemo/collections/nlp/models/language_modeling/megatron_lm_encoder_decoder_model.py
jbaczek: I hoped for something more verbose :) Like
"""
This function takes as input per-token loss and masks non-required values.
Input:
    loss_mask: binary torch.Tensor of shape batch_size x seq_len
    loss_tokens: torch.Tensor of shape batch_size x seq_len
Returns:
    loss: 0-dimension torch.Tensor (single value)
"""

odelalleau: There isn't a single loss function in NeMo (that I can see) that has such a detailed docstring. I'm not saying it wouldn't be a good idea to add them, but this is out-of-scope for this PR which is focused on fixing a bug. It seems to me that adding shapes in comments should be enough to follow the loss calculation.

odelalleau:
jenkins

PR #8395, opened 2024-02-10
Added memory logger

sanandaraj5597:
This PR adds an env variable to be set when you want to log peak memory usage at the end of each training step. 
This env variable "NEMO_LOG_MEMORY_USAGE" should be set to 1 for the logging to be enabled, this prints the memory usage of rank 0 only.
This env variable will be added to perf CI for NeMo.


ericharper (approved):
LGTM. Thanks!

ericharper:
jenkins

sanandaraj5597:
jenkins

PR #8431, opened 2024-02-15
SDE: Added support of GPU-acceleration and running inside Docker

ssh-meister:
#7940 continue
GPU acceleration added,
Data loading optimization with adding multiprocessing,
Code refactoring started

+ Added support for running inside the docker container 

@vsl9


tools/speech_data_explorer/Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Use the base image
+FROM nvcr.io/nvidia/rapidsai/rapidsai:23.06-cuda11.8-runtime-ubuntu22.04-py3.10
```
Jorjeous, line 2: Prb. will not work for external users

ssh-meister: This image is open source

PR #8476, opened 2024-02-21
Update black to the latest version incrementally

artbataev:
# What does this PR do?
Update black using incremental changes only for files changed in PRs.

- the `black` package is updated to the latest version
- all collections are excluded, so `black .` will reformat nothing
  - it is possible also to exclude collection from the list, once it is reformatted though PRs
- there will be 2 approaches to reformat the files manually: 
  - pass the file(s) explicitly (e.g., `black my_file.py`) 
  - use a pre-commit hook
- the automatic reformatting is done by GitHub Actions only for PRs and only for changed files

Tests (target branch is `update_black`):
- #9017 - Test from fork (code reformatted, pushed back to the branch from fork)
- #8996 - Test from fork, with updates
- #9019 - Test from branch in the main repo

Approach:
- create a custom GitHub Action responsible for reformatting the code. For black use only changed files
- custom token is used to push changes, since only custom token can trigger workflows after reformatting

# Q&A About the Approach

> Why do we need a custom GitHub Action if we have pre-commit.ci?

The pre-commit hook is still active and works, but pre-commit.ci reformats all the code (the pre-commit.ci team explicitly denied changing this behavior, see https://github.com/pre-commit-ci/issues/issues/90), thus it is replaced with a custom GitHub Action, which does the same, but only for the changed files (in the PR)

> Why do we need `isort` GitHub action?

Since both `isort` and `black` can change files, this should be done in one action to avoid concurrent changes. So, the GitHub Action will execute both formatters and commit changes if needed (same behavior as for pre-commit.ci).

> Why do we need a custom GitHub token?

When GitHub action pushes the commit with the default secret token, it cannot trigger other workflows. 
See the discussion: https://github.com/orgs/community/discussions/25702
This can only be done with custom tokens.

**Collection**: [CI]

# Changelog 

# Usage

```python
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


artbataev:
@titu1994, @ericharper To make this finally work we need a separate token with minimal permissions:
- Read access to metadata
- Read and Write access to actions and code

The token will have fewer permissions than the default `GITHUB_TOKEN`, so I do not think that using custom token introduces any new issues.

The reason why we need a custom token is that `GITHUB_TOKEN` cannot trigger workflows, see https://github.com/orgs/community/discussions/25702 for details.

I tested the solution on the separate repository, and everything works (code reformatted, pushed if needed, after push the workflows are triggered).

nemo/collections/asr/__init__.py
```diff
@@ -23,3 +23,10 @@
 
 # Set collection name.
 __description__ = "Automatic Speech Recognition collection"
+
```
titu1994, line 26: Revert ?

pyproject.toml
```diff
@@ -27,8 +27,34 @@ extend_skip = ["setup.py", "docs/source/conf.py"]
 [tool.black]
 line_length = 119
 skip_string_normalization = true
-required_version = "19.10b0"  # recongized by future versions, disallows to reformat code with incompatible versions
-
+# major year version is stable, see details in
+# https://black.readthedocs.io/en/stable/the_black_code_style/index.html
+required_version = "24"
+target-version = ['py310', 'py311', 'py312']
```
titu1994, line 33: There are four places we are referencing black version now, in the action, in the toml and the pre commit, requirements txt. Can we somehow have a unified place that sets version for all three. 

If not easy, we need to write somewhere that all 4 places need to be updated simultaneously.

requirements/requirements_test.txt
```diff
@@ -1,4 +1,4 @@
-black==19.10b0
+black~=24.3
```
titu1994, line 1: Shouldn't this be 24.3?

titu1994 (commented):
Really interesting solution, I'm looking into if custom token is necessary since it's permissions are less than GitHub token

requirements/requirements_test.txt
artbataev: `black~=24.3` means (`black>=24.3` and `black<25`). Major version is stable, so `black~=24.3` is safe.

nemo/collections/asr/__init__.py
artbataev: This should be reverted automatically by black :)
The action is not working now since it has no custom github token

pyproject.toml
artbataev: For actions (pre-commit, GitHub) we need to specify action version, but not black. It is not possible to do this in one place. But `required_version="24"` in this file **guarantees the consistency**: 
any action with black will fail if it `black` major version is different from `required_version`

artbataev:
Test changes from fork: #8996

pyproject.toml
artbataev: It is also beneficial in cases when the user locally has a different black version (not from the requirements). Instead of incorrectly reformatting the files, it will fail.

pablo-garay (approved):

titu1994 (approved):

PR #8480, opened 2024-02-21
Fix PTL2.2 saving multiple `*-last.ckpt` checkpoints in resumed training

stevehuang52:
# What does this PR do ?

Fix PTL2.2 saving multiple `*-last.ckpt` checkpoints in resumed training. 

In current PTL2.2 [model_checkpoint.py](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/callbacks/model_checkpoint.py#L783), it doesn't delete the previous `*-last.ckpt` checkpoint that the model was resumed on when saving a new `*-last.ckpt` checkpoint, which results in multiple checkpoints. This will cause errors of "multiple *-last.ckpt checkpoints" when trying to resume from a previous job.

The fix is to check whether the checkpoint to be saved is also a `-last.ckpt` checkpoint. If it is, delete the previous `-last.ckpt` if they are under the same directory.

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


athitten:
@stevehuang52 this problem basically happens when we try to resume 2nd time right ?

stevehuang52:
> @stevehuang52 this problem basically happens when we try to resume 2nd time right ?

Yes, it happens when we resume for the 2nd time, the first resume is fine

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -454,3 +454,29 @@ def _remove_unfinished_checkpoints(checkpoint_dir: Union[Path, str]) -> None:
...
+        previous = Path(previous).absolute()
+        resume_path = Path(trainer.ckpt_path).absolute() if trainer.ckpt_path is not None else None
+
+        if resume_path is not None and previous == resume_path:
+            if str(current).endswith("-last.ckpt") and resume_path.name.endswith("-last.ckpt"):
+                # delete the previous `-last.ckpt` checkpoint when current saved checkpoint is also `-last.ckpt`, if they're in the same directory
+                pass
+            else:
+                return False
+        if self.dirpath is None:
```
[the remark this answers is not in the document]
stevehuang52: @athitten I remembered @titu1994 told me to avoid using `assert` in nemo, and to raise exceptions instead so that they can be caught by error handling code. However, it's up to you if you would like to keep it identical to PTL

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -454,3 +454,29 @@ def _remove_unfinished_checkpoints(checkpoint_dir: Union[Path, str]) -> None:
         # delete markers
         for marker_path in existing_marker_filepaths:
             os.remove(marker_path)
+
+    def _should_remove_checkpoint(self, trainer: "pl.Trainer", previous: str, current: str) -> bool:
```
athitten, line 458: @stevehuang52 we want to avoid overriding PTL's protected methods in NeMo as much as possible so that NeMo doesn't break when PTL changes something in the protected methods.

Maybe we can create a PR for this in PTL itself ? Since its not a lot of modification and just addition of couple of lines in this function.

stevehuang52: @athitten thanks for the advice, I've created a PR to PTL https://github.com/Lightning-AI/pytorch-lightning/pull/19613, let's see if they'll approve

athitten: Great, thank you very much @stevehuang52.

stevehuang52:
jenkins

athitten:
jenkins

athitten:
jenkins

stevehuang52:
jenkins

nemo/utils/callbacks/nemo_model_checkpoint.py
titu1994: Seems they closed the PR above. Is the resolution implemented here?

nemo/utils/callbacks/nemo_model_checkpoint.py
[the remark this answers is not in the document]
titu1994: Agreed, assert hides actual error, we should raise it properly with useful error message (like here)

nemo/utils/callbacks/nemo_model_checkpoint.py
stevehuang52: @titu1994 Yeah they don't want to do the fix in PTL, we will do the fix in NeMo (which is this PR).

stevehuang52:
@athitten seems that Jenkins passed without issues. Could you please review the PR again and see if there's other changes or tests needed? Thanks~

athitten (approved):
LGTM, thanks very much @stevehuang52

PR #8521, opened 2024-02-27
Accelerate CTC greedy decoding by around 10%

galv:
# What does this PR do ?

Accelerate CTC greedy decoding by around 10x, speeding up overall CTC runtime for parakeet-ctc-1.1b by close to 10%.

**Collection**: ASR

# Changelog 

There were a few problems with CTC greedy decoding before:

- It would copy GPU memory to pageable memory originally, which is very slow compared to copying to pinned memory. Unfortunately, cudaMallocHost() is synchronous and "slow", but fortunately pytorch has a free list of recent pinned memory allocations in its pinned memory allocator, allowing us to reduce the number of these calls. The cost of the memcpy before my change is excessive, taking ~20-40ms. After changing to use pinned memory, it takes only 2-4ms. This is the primary source of the 10x speedup. Note that doing the initial pinned memory allocations is slow (discussed later in this message).

<img width="789" alt="image" src="https://github.com/NVIDIA/NeMo/assets/4767568/11a1addd-b9a7-4060-84f7-712ea36062e3">

- A single scalar of logits_len/decoder_lengths was copied from GPU to CPU at a time, rather than all at once in a single call. This caused neddless overhead. We could also use a pinned memory allocation for copying logits_len as well, but using pinned memory is less important for small allocations. The original code's trace looks like this:
<img width="1061" alt="image" src="https://github.com/NVIDIA/NeMo/assets/4767568/9467224f-a1a0-492d-a4a5-9f3804e975ae"> Each red box followed by a green box is copying a single scalar value from GPU to CPU

- detatch() was called in function marked with @nograd

- For some reason, someone was using torch_tensor.numpy().tolist() instead of torch_tensor.tolist(). I don't believe numpy() makes a copy of the data, but it is unnecessary.

The main important improvements are the first two bullet points. Everything else is unimportant.

There are more opportunities for improvement. In particular, logits gets copied to cpu twice if trcfg.return_hypotheses is True. However, this overhead is now small since it will always just use the previous pinned host allocation from the free list.

Performance improvements:

I ran this code on an A100 GPU, on a machine with 16 CPU cores:

```
python examples/asr/speech_to_text_eval.py \
pretrained_name=nvidia/parakeet-ctc-1.1b \
dataset_manifest=/home/dgalvez/scratch/data/test_other.json \
batch_size=16  output_filename=test_other_decoded.jsonl  amp=true \
amp_dtype=bfloat16  use_cer=false num_workers=1
```

I applied this diff to run multiple times:

```
modified   examples/asr/transcribe_speech.py                                                                                                                                                              
@@ -367,14 +367,19 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis                                                                                                    
                     decoder_type=cfg.decoder_type,                                                                                                                                                       
                 )                                                                                                                                                                                        
             else:                                                                                                                                                                                        
-                transcriptions = asr_model.transcribe(                                                                                                                                                   
-                    audio=filepaths,                                                                                                                                                                     
-                    batch_size=cfg.batch_size,                                                                                                                                                           
-                    num_workers=cfg.num_workers,                                                                                                                                                         
-                    return_hypotheses=cfg.return_hypotheses,                                                                                                                                             
-                    channel_selector=cfg.channel_selector,                                                                                                                                               
-                    augmentor=augmentor,                                                                                                                                                                 
-                )                                                                                                                                                                                        
+                for i in range(5):                                                                                                                                                                       
+                        torch.cuda.cudart().cudaProfilerStart()                                                                                                                                          
+                    transcriptions = asr_model.transcribe(                                                                                                                                               
+                        audio=filepaths,                                                                                                                                                                 
+                        batch_size=cfg.batch_size,                                                                                                                                                       
+                        num_workers=cfg.num_workers,                                                                                                                                                     
+                        return_hypotheses=cfg.return_hypotheses,                                                                                                                                         
+                        channel_selector=cfg.channel_selector,                                                                                                                                           
+                        augmentor=augmentor,                                                                                                                                                             
+                    )                                                                                                                                                                                    
+                        torch.cuda.cudart().cudaProfilerStop()                                                                                                                                           
                                                                                                                                                                                                          
     if cfg.dataset_manifest is not None:                                                                                                                                                                 
         logging.info(f"Finished transcribing from manifest file: {cfg.dataset_manifest}")                                                                                                                
```

Time to do each of 5 evaluations of librispeech test other before my changes: 

32 seconds
28
29
29
29

Average: 29.4 seconds
Average excluding first (warmup): 28.75 seconds

Time to do each of 5 evaluations of librispeech test other after my changes:

30 seconds
26
26
25
25
Average: 26.4 seconds
Average excluding first (warmup): 25.5 seconds

This corresponds to an almost 10% speedup. This meets expectations, since 10% of the time was originally spent on CTC greedy decoding before.

P99 latency can increase for the first few calls to the model with this change. This is because we are now calling cudaMallocHost(), and then doing a pinned memory copy from GPU to CPU, which is slower than doing a paged memory copy from GPU to CPU. However, the cudaMallocHost() calls will be cached over time, allowing us to avoid the overhead of them in later evaluations of the dataset. The expense of a cudaMallocHost() (or cudaHostAlloc()) call can be seen here:

<img width="921" alt="image" src="https://github.com/NVIDIA/NeMo/assets/4767568/fd7fd4bc-833d-43ba-9226-959ebb7c9ee6">



# Usage

See above

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -666,9 +666,7 @@ class EncDecCTCModel(ASRModel, ExportableEncDecModel, ASRModuleMixin, InterCTCMi
             # See comment in
             # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
             # understand this idiom.
-            logits_cpu = torch.empty(logits.shape,
-                                     dtype=logits.dtype,
-                                     pin_memory=True)
+            logits_cpu = torch.empty(logits.shape, dtype=logits.dtype, pin_memory=True)
             logits_cpu.copy_(logits)
             logits_len = logits_len.cpu()
             # dump log probs per file
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -667,10 +667,7 @@ class EncDecCTCModel(ASRModel, ExportableEncDecModel, ASRModuleMixin, InterCTCMi
                 # See comment in
                 # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
                 # understand this idiom.
-                logits_cpu = torch.empty(logits.shape,
-                                         dtype=logits.dtype,
-                                         device=torch.device("cpu"),
-                                         pin_memory=True)
+                logits_cpu = torch.empty(logits.shape, dtype=logits.dtype, device=torch.device("cpu"), pin_memory=True)
                 logits_cpu.copy_(logits, non_blocking=True)
             else:
                 logits_cpu = logits
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -662,14 +662,21 @@ class EncDecCTCModel(ASRModel, ExportableEncDecModel, ASRModuleMixin, InterCTCMi
         current_hypotheses, all_hyp = self.decoding.ctc_decoder_predictions_tensor(
             logits, decoder_lengths=logits_len, return_hypotheses=trcfg.return_hypotheses,
         )
-        logits = logits.cpu()
-
         if trcfg.return_hypotheses:
+            # See comment in
+            # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
+            # understand this idiom.
+            logits_cpu = torch.empty(logits.shape,
+                                     dtype=logits.dtype,
+                                     pin_memory=True)
+            logits_cpu.copy_(logits)
+            logits_len = logits_len.cpu()
             # dump log probs per file
-            for idx in range(logits.shape[0]):
-                current_hypotheses[idx].y_sequence = logits[idx][: logits_len[idx]]
+            for idx in range(logits_cpu.shape[0]):
+                current_hypotheses[idx].y_sequence = logits_cpu[idx][: logits_len[idx]]
                 if current_hypotheses[idx].alignments is None:
                     current_hypotheses[idx].alignments = current_hypotheses[idx].y_sequence
+            del logits_cpu
 
         # cleanup memory
         del logits, logits_len
--- a/nemo/collections/asr/models/ctc_models.py
+++ b/nemo/collections/asr/models/ctc_models.py
@@ -663,11 +663,17 @@ class EncDecCTCModel(ASRModel, ExportableEncDecModel, ASRModuleMixin, InterCTCMi
             logits, decoder_lengths=logits_len, return_hypotheses=trcfg.return_hypotheses,
         )
         if trcfg.return_hypotheses:
-            # See comment in
-            # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
-            # understand this idiom.
-            logits_cpu = torch.empty(logits.shape, dtype=logits.dtype, pin_memory=True)
-            logits_cpu.copy_(logits)
+            if logits.is_cuda:
+                # See comment in
+                # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
+                # understand this idiom.
+                logits_cpu = torch.empty(logits.shape,
+                                         dtype=logits.dtype,
+                                         device=torch.device("cpu"),
+                                         pin_memory=True)
+                logits_cpu.copy_(logits, non_blocking=True)
+            else:
+                logits_cpu = logits
             logits_len = logits_len.cpu()
             # dump log probs per file
             for idx in range(logits_cpu.shape[0]):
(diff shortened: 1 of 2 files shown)


nemo/collections/asr/models/ctc_models.py
```diff
@@ -662,14 +662,19 @@ def _transcribe_output_processing(self, outputs, trcfg: TranscribeConfig) -> Gen
         current_hypotheses, all_hyp = self.decoding.ctc_decoder_predictions_tensor(
             logits, decoder_lengths=logits_len, return_hypotheses=trcfg.return_hypotheses,
         )
-        logits = logits.cpu()
-
         if trcfg.return_hypotheses:
+            # See comment in
+            # ctc_greedy_decoding.py::GreedyCTCInfer::forward() to
+            # understand this idiom.
+            logits_cpu = torch.empty(logits.shape, dtype=logits.dtype, pin_memory=True)
```
titu1994, line 669: This is a very neat trick. Just a question - what happens if i execute this on an environment with no GPU at all (not even cuda is setup). Will it work fine?

titu1994: Reason being - CPU only inference is quite common with our ASR models, especially on Gradio / HF space.

titu1994: Finally, no need to set device=something? The copy_ automatically does device to host copy?

titu1994 (commented):
Very nice speedup ! Minor questions, code looks good so we can merge easily once tests pass.

nemo/collections/asr/models/ctc_models.py
galv: > what happens if i execute this on an environment with no GPU at all (not even cuda is setup). Will it work fine?

You're right, it doesn't just work in a CPU-only environment. I get:

```
Python 3.9.18 (main, Sep 11 2023, 08:25:10) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import torch

In [2]: a = torch.empty((2,2,2), pin_memory=True)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In [2], line 1
----> 1 a = torch.empty((2,2,2), pin_memory=True)

RuntimeError: Pinned memory requires CUDA. PyTorch splits its backend into two shared libraries: a CPU library and a CUDA library; this error has occurred because you are trying to use some CUDA functionality, but the CUDA library has not been loaded by the dynamic linker for some reason.  The CUDA library MUST be loaded, EVEN IF you don't directly use any symbols from the CUDA library! One common culprit is a lack of -Wl,--no-as-needed in your link arguments; many dynamic linkers will delete dynamic library dependencies if you don't depend on any of their symbols.  You can check if this has occurred by using ldd on your binary to see if there is a dependency on *_cuda.so library.
```

Getting this to work in a sensible way without too many conditional statements (after all, in the existing code, .cpu() is a no-op in the case that the user was running on CPU; meanwhile, here, I would impose an extra memory copy on those users) is probably the main challenge with this PR.

> Finally, no need to set device=something?

I also made a mistake here, it turns out. I assumed that device=None would default to the "cpu", but it actually defaults to whatever you set with https://pytorch.org/docs/stable/generated/torch.set_default_device.html#torch.set_default_device So I need to explicitly use torch.device("cpu") here.

galv:
@titu1994 FYI, I made a fix. I tested by simply running running

```
python examples/asr/speech_to_text_eval.py  model_path=$HOME/.cache/huggingface/hub/models--nvidia--parakeet-ctc-1.1b/snapshots/085a3de63c7598065b072cd8f2182e6a5fa593eb/parakeet-ctc-1.1b.nemo  dataset_manifest=/home/dgalvez/scratch/data/test_other.json  batch_size=16  output_filename=test_other_decoded.jsonl  amp=true  amp_dtype=bfloat16  use_cer=false num_workers=1
```

with and without "CUDA_VISIBLE_DEVICES="

But the existing CTC test cases would have caught the error anyway if jenkins had run successfully, if jenkins had worked instead of getting stuck on fetching the git repo.

I also double checked my throughput results. They are a little bit better than originally reported because of a dumb error on my part. We are going from 28.75 seconds (RTFx 676) to 25.5 seconds (RTFx 762) after warmup on librispeech test other. Which is a nice speedup.

titu1994 (approved):
Very nice ! Thanks for the speedup @galv !

Issue #8592, opened 2024-03-05
ValueError when higher minimum_timestamp_duration given : "start_time must be a float or str containing float, but got int"


Stephane-Lpt:
**Describe the bug**
I wanted to put a `minimum_timestamp_duration` to visualize the result given by nemo forced aligner with [Gecko](https://github.com/gong-io/gecko) as the documentation recommends. 
When I change the value of `minimum_timestamp_duration` to 0.4, it's all good.
If I try to change the `minimum_timestamp_duration` for a higher value like 0.5, I get a `ValueError` : 
![image](https://github.com/NVIDIA/NeMo/assets/66870443/ca90e8df-a4f2-41d4-a3a2-1867e0b37cae)
**Steps/Code to reproduce bug**

- Notice the option **ctm_file_config.minimum_timestamp_duration=0.5** to change the value of `minimum_timestamp_duration`. To be sure to have the error, you can put an extremely high value like 50.
```
python <path to NeMo>/tools/nemo_forced_aligner/align.py pretrained_name="stt_en_fastconformer_hybrid_large_pc" manifest_filepath=nemo/manifest.json output_dir=nemo/output ctm_file_config.minimum_timestamp_duration=0.5
```
The value that is too high and causes the error **depends on** the **respective duration of your tokens/words/segments** that are calculated in this [loop](https://github.com/NVIDIA/NeMo/blob/1882d0d0025832a530063e5b699347253b2ab768/tools/nemo_forced_ligner/utils/make_ctm_files.py#L83). The loop iterates over **each token/word/segment** of the transcription and finally produces the `ctm` file).

**Expected behavior**

- No error should be raised.

**Environment overview**

 - Method of NeMo install: source

**Additional context**
**_This is a minor error that I've already fixed locally, so I'll make a pull request in a few minutes_**

erastorgueva-nv:
PR with fix has been merged, so closing this. Thank you for the fix.

PR #8593, opened 2024-03-05
fix: start_time type of int instead of float

Stephane-Lpt:
# What does this PR do ?

Fix : You can now completely use the `minimum_timestamp_duration` option, which bugged if the value was too large.

**Collection**: ASR

# Changelog 
As we could see in the traceback, the problem appeared during the use of the `get_ctm_line` function because `start_time` was a type of **int** at [this line](https://github.com/NVIDIA/NeMo/blob/main/tools/nemo_forced_aligner/utils/make_ctm_files.py#L98) :
At the beginning, `boundary_info_.t_start` is definitely a **float** with the value 0.0 and `start_time` is also a float after [copying](https://github.com/NVIDIA/NeMo/blob/main/tools/nemo_forced_aligner/utils/make_ctm_files.py#L88). Everything is fine for now.
The source of the problem is at [this line](https://github.com/NVIDIA/NeMo/blob/main/tools/nemo_forced_aligner/utils/make_ctm_files.py#L98) : When using the `max` function, if we have set a `minimum_timestamp_duration` value that is too high. `start_time` will become 0 BUT type of **int** and **no more** type of **float**.
So when we get to the use of the `get_ctm_line` function, this one checks the type of `start_time` and since it's not the right type, it raises a type exception [here](https://github.com/NVIDIA/NeMo/blob/935b9c78f4e75e873f8511121bd846e7dabca4f3/nemo/collections/asr/parts/utils/manifest_utils.py#L97).

Consequently, the **max function argument** should be changed to **0.0** to be always a type of **float**.

# Usage
- You can now really use the option **ctm_file_config.minimum_timestamp_duration=0.5** to change the value of `minimum_timestamp_duration` no matter how long the tokens/words/segments of your transcription is.
Example : 
```
python <path to NeMo>/tools/nemo_forced_aligner/align.py pretrained_name="stt_en_fastconformer_hybrid_large_pc" manifest_filepath=nemo/manifest.json output_dir=nemo/output ctm_file_config.minimum_timestamp_duration=0.5
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #8592


Stephane-Lpt:
jenkins

titu1994 (approved):
Thanks for the fix !

PR #8604, opened 2024-03-07
fix race condition in exp manager

gshennvm:
# What does this PR do ?

in exp manager we move files on rank0, but this move is not always completed by the time the other ranks go and do other things such as try to run `ls` on the file system. So this adds a sync to make sure

### DIFF
--- a/tests/core_ptl/check_for_ranks.py
+++ b/tests/core_ptl/check_for_ranks.py
@@ -112,13 +112,16 @@ def get_rank_info(texts: list, rank_key: str) -> int:
 
 @rank_zero_only
 def check_model_ranks(model: ExampleModel):
-    basedir = os.path.join('./ddp_check/', 'default', 'version_0')
+    basedir = os.path.join('./ddp_check/', 'default')
+    to_add = ""
     file_template = "nemo_log_globalrank-{rank}_localrank-{rank}.txt"
 
     world_size = torch.cuda.device_count()
     for rank in range(world_size):
+        to_add = "version_0" if rank == 0 else ""
+
         filename = file_template.format(rank=rank)
-        filepath = os.path.join(basedir, filename)
+        filepath = os.path.join(basedir, to_add, filename)
 
         with open(filepath, 'r', encoding='utf-8') as f:
             texts = f.readlines()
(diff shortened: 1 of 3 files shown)


gshennvm:
jenkins

ericharper (approved):
LGTM. Thanks!

gshennvm:
jenkins

tests/core_ptl/check_for_ranks.py
```diff
@@ -83,7 +83,9 @@ def instantiate_multinode_ddp_if_possible():
     num_gpus = torch.cuda.device_count()
     ## Change logger=None to logger=False to support PTL 2.0
     trainer = Trainer(devices=num_gpus, accelerator='gpu', strategy='ddp', logger=False, enable_checkpointing=False)
-    exp_manager_cfg = ExpManagerConfig(exp_dir='./ddp_check/', use_datetime_version=False, version="")
+    exp_manager_cfg = ExpManagerConfig(
+        exp_dir='./ddp_check/', use_datetime_version=False, version="", explicit_log_dir="./ddp_check/log_dir"
```
titu1994, line 87: I don't think there is a race condition here, or else it would have popped up at least a few times in the last 3 years of the test. Why is explicit logdir being used ?

gshennvm: rank 0 has a logger https://github.com/NVIDIA/NeMo/blob/7de8f87e7637a51461159c3f00295e72c3609904/nemo/utils/exp_manager.py#L782 which makes log_dir `./ddp_check/default/version_0` but rank 1 doesn't have a logdir, which makes log_dir `./ddp_check/default` without the version 0 then this test fails.

if we set explicit log dir here: https://github.com/NVIDIA/NeMo/blob/7de8f87e7637a51461159c3f00295e72c3609904/nemo/utils/exp_manager.py#L740 it short circuits this logic

gshennvm:
jenkins

ericharper:
jenkins

tests/core_ptl/check_for_ranks.py
titu1994: We avoid explicit log dir because it circumvents normal operation of exp  Manager and leads to bugs going undetected. 

You mean to say, over 2.5 years of testing and without fail the race condition always executed and therefore this code ran without fail for that long ? Doesn't seem likely. In either case, id suggest removing explicit log dir and updating the test to work without explicit log dir.

gshennvm:
jenkins

gshennvm:
jenkins

titu1994 (approved):
Perfect, thanks !

gshennvm:
jenkins

gshennvm:
jenkins

PR #8662, opened 2024-03-14
Support dumping nbest into manifest in ASR

zhehuaichen:
Support dumping nbest into manifest if exists from transcribe_speech tool
Previously it only generates but does not store.

- Test on Parakeet
```
python \
        ../transcribe_speech.py \
        pretrained_name="nvidia/parakeet-rnnt-1.1b"  \
        dataset_manifest=$VAL extract_nbest=True presort_manifest=False  rnnt_decoding.beam.beam_size=5 rnnt_decoding.strategy='beam' gt_text_attr_name="answer" output_filename=$output_filename
```

- Test on Canary

```
python \
        ../transcribe_speech.py \
        pretrained_name="nvidia/canary-1b"  \
        dataset_manifest=$VAL extract_nbest=True presort_manifest=False  multitask_decoding.beam.beam_size=5 gt_text_attr_name="answer" output_filename=$output_filename
```

Collection: [asr]

### DIFF
--- a/examples/asr/transcribe_speech.py
+++ b/examples/asr/transcribe_speech.py
@@ -280,7 +280,7 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
             cfg.multitask_decoding.compute_langs = cfg.compute_langs
             cfg.multitask_decoding.preserve_alignments = cfg.preserve_alignment
             if cfg.extract_nbest:
-                cfg.multitask_decoding.beam.return_best_hypothesis=False
+                cfg.multitask_decoding.beam.return_best_hypothesis = False
                 cfg.return_hypotheses = True
             asr_model.change_decoding_strategy(cfg.multitask_decoding)
         elif cfg.decoder_type is not None:
@@ -290,7 +290,7 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
 
             decoding_cfg = cfg.rnnt_decoding if cfg.decoder_type == 'rnnt' else cfg.ctc_decoding
             if cfg.extract_nbest:
-                decoding_cfg.beam.return_best_hypothesis=False
+                decoding_cfg.beam.return_best_hypothesis = False
                 cfg.return_hypotheses = True
             decoding_cfg.compute_timestamps = cfg.compute_timestamps  # both ctc and rnnt support it
             if 'preserve_alignments' in decoding_cfg:
@@ -301,11 +301,11 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
                 asr_model.change_decoding_strategy(decoding_cfg, decoder_type=cfg.decoder_type)
             else:
                 asr_model.change_decoding_strategy(decoding_cfg)
-        
+
         # Check if ctc or rnnt model
         elif hasattr(asr_model, 'joint'):  # RNNT model
             if cfg.extract_nbest:
-                cfg.rnnt_decoding.beam.return_best_hypothesis=False
+                cfg.rnnt_decoding.beam.return_best_hypothesis = False
                 cfg.return_hypotheses = True
             cfg.rnnt_decoding.fused_batch_size = -1
             cfg.rnnt_decoding.compute_timestamps = cfg.compute_timestamps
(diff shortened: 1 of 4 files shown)


titu1994:
There's a lot of additional flags to set - what about reading the value of extract nbest and setting the corresponding flags in the data classes ?

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -391,7 +391,8 @@ def write_transcription(
             if not cfg.decoding.beam.return_best_hypothesis:
                 beam = []
                 for hyp in hyps:
-                    beam.append((hyp.text, hyp.score))
+                    score = hyp.score.numpy().tolist() if isinstance(hyp.score, torch.Tensor) else hyp.score
```
titu1994, line 394: Why would score be a tensor or a list ?

examples/asr/transcribe_speech.py
```diff
@@ -400,9 +400,16 @@ def autocast(dtype=None):
         logging.info(f"Finished transcribing {len(filepaths)} files !")
     logging.info(f"Writing transcriptions into file: {cfg.output_filename}")
 
-    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses), extract just best hypothesis
+    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses)
     if type(transcriptions) == tuple and len(transcriptions) == 2:
-        transcriptions = transcriptions[0]
+        if isinstance(transcriptions[1][0], list) and isinstance(transcriptions[1][0][0], Hypothesis):
```
titu1994, line 405: Triple list shouldn't be needed. Let's stick with two nest list max. It will break multiple other parts external to Nemo

titu1994 (changes requested):

nemo/collections/asr/parts/utils/transcribe_utils.py
zhehuaichen: edited

examples/asr/transcribe_speech.py
zhehuaichen: removed

zhehuaichen:
jenkins

examples/asr/transcribe_speech.py
```diff
@@ -278,13 +279,20 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
         if isinstance(asr_model.decoding, MultiTaskDecoding):
             cfg.multitask_decoding.compute_langs = cfg.compute_langs
             cfg.multitask_decoding.preserve_alignments = cfg.preserve_alignment
+            if cfg.extract_nbest:
+                cfg.multitask_decoding.beam.return_best_hypothesis = False
+                cfg.return_hypotheses = True
             asr_model.change_decoding_strategy(cfg.multitask_decoding)
         elif cfg.decoder_type is not None:
+            breakpoint()
```
titu1994, line 287: Remove breakpoint

examples/asr/transcribe_speech.py
```diff
@@ -400,9 +416,14 @@ def autocast(dtype=None):
         logging.info(f"Finished transcribing {len(filepaths)} files !")
     logging.info(f"Writing transcriptions into file: {cfg.output_filename}")
 
-    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses), extract just best hypothesis
+    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses)
     if type(transcriptions) == tuple and len(transcriptions) == 2:
-        transcriptions = transcriptions[0]
+        if isinstance(transcriptions[1][0], list):
```
titu1994, line 421: Why would this be a list? Make sure to return an NBestHypothesis object. Inside it, the list of all hypotheses is packed.

titu1994: NBestHypothesis object should be returned by both the MultiTaskDecoding so the list is preserved inside of it.

titu1994 (changes requested):

examples/asr/transcribe_speech.py
titu1994: Don't check like this. See if the flag is set or not, then extract out the second item in the top level. Almost always, the second item is None rather than the beam values.

zhehuaichen: done

examples/asr/transcribe_speech.py
zhehuaichen: done

examples/asr/transcribe_speech.py
```diff
@@ -400,9 +415,14 @@ def autocast(dtype=None):
         logging.info(f"Finished transcribing {len(filepaths)} files !")
     logging.info(f"Writing transcriptions into file: {cfg.output_filename}")
 
-    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses), extract just best hypothesis
+    # if transcriptions form a tuple of (best_hypotheses, all_hypotheses)
     if type(transcriptions) == tuple and len(transcriptions) == 2:
-        transcriptions = transcriptions[0]
+        if cfg.extract_nbest:
+            # extract all hypotheses if exists
+            transcriptions = transcriptions[1]
```
titu1994, line 422: Write transcriptions function expects a single list, not a nester list of lists

PR #8679, opened 2024-03-16
injecting text_field when not present in manifest in AED models transcribe fn[ASR]

krishnacpuvvada:
# What does this PR do ?
injecting text_field when not present in manifest in AED models transcribe fn

**Collection**: [ASR]

**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation


titu1994 (approved):
Very nice ! Thanks, this should remove most of the confusion for inference. We still need to remove this answer thing but probably can do it a later

PR #8705, opened 2024-03-20
[Nemo CICD] Github Actions CICD init - Part 3: Unique container name per workflow run

pablo-garay:
Unique container name per workflow run.

This is so that each container name per workflow is unique and different workflows do not collide / end up using the same name & use the container produced by another workflow.


pablo-garay:
CI pipeline passed

PR #8727, opened 2024-03-22
Embedding Wgrad defering

sanandaraj5597:
# What does this PR do ?

The PR adds functionality to defer the embedding table wgrad GEMM, so that it happens during pipeline flush. This is only done on the last pipeline stage.
The embedding wgrad compute is drained by a function in megatron.core.utils which NeMo will import and use.


ericharper:
jenkins

sanandaraj5597:
jenkins

PR #8729, opened 2024-03-22
Pagaray/nemo cicd part5

pablo-garay:
Adds cleaning step for CI runs


pablo-garay:
CI had passed: 
image

PR #8739, opened 2024-03-25
Fix "index 0 is out of bounds for dimension 1 with size 0"

ssh-meister:
In the code, the variable `audio` containing the original audio recording gets overwritten during the operation `audio[start: start + duration]`. Consequently, on subsequent iterations of the loop, a segment of a new length cannot be processed correctly because the timestamps of the segment exceed the boundaries of the overwritten `audio`.

This PR addresses this issue by proposing a fix to ensure that the overwritten `audio` variable does not cause incorrect processing of subsequent segments.


karpnv (dismissed):
Thanks, LGTM!

karpnv (approved):
LGTM

vsl9 (dismissed):
Thanks, Sasha.

PR #8807, opened 2024-04-03
Revert "add fsdp fix for tp > 1"

mwawrzos:
Reverts NVIDIA/NeMo#8689

The NVIDIA/NeMo#8689 initializes the `PipelineMixedPrecisionPlugin` incorrectly. The `FSDPPrecision.__init__` is not called, and in result the `self._desired_input_dtype` field is not set.

Minimalistic repro:
```bash
 python -c "from nemo.collections.nlp.parts.nlp_overrides import PipelineMixedPrecisionPlugin as PM; PM('16-mixed', 'cuda').tensor_init_context()"
```

The `FSDPPrecision.__init__` is not called, because it doesn't have the `device` argument.


mwawrzos:
jenkins

dimapihtar (approved):
LGTM. Thank you!

PR #8819, opened 2024-04-04
Readme upgrade (experimental installation on windows)

Jorjeous:
# What does this PR do ?

All
# Changelog 
Pipeline for installation on pure windows
# Usage

```python
```

# Jenkins CI

**Pre checks**:
- [+] Did you add or update any necessary documentation?
  
**PR Type**:
- [+] New Feature
- [ ] Bugfix
- [+] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


README.rst
```diff
@@ -294,6 +294,53 @@ Next, follow the instructions for Linux systems, as provided above. For example:
     cd NeMo
     ./reinstall.sh
 
+Pure Windows Installation (Experimental)
```
titu1994, line 297: I wonder if this is really necessary to hand the pinned requirements file. Furthermore, there's no guarantee this won't break in the future. Do we need to support native windows install for ASR ? If so it can go into ASR docs rather than main readme cause no other domains will work on windows anyway. (Maybe TTS?)

requirements/win/win_reqs.txt
```diff
@@ -0,0 +1,149 @@
+absl-py==2.1.0
```
titu1994, line 1: Avoid exact pinned dependency list like this. We're never going to update it.

titu1994 (changes requested):

requirements/win/win_reqs.txt
Jorjeous: And we should not. Main problem is with one of the packages (triton) not supported on windows.
I found the replacement that can be installed on windows and there is only one version. I do not think it will update in the future so hardcoded all dependencies. 
But yes, except this triton package we can have unspecified libs

README.rst
Jorjeous: Yes TTS as well, but if we put it in ASR doc no one will ever find it (Imho)

PR #8841, opened 2024-04-05
Akoumparouli/fix get params for weight decay optimization

akoumpa:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/modules/common/megatron/utils.py
+++ b/nemo/collections/nlp/modules/common/megatron/utils.py
@@ -372,7 +372,7 @@ def get_params_for_weight_decay_optimization(
                     list(filter(lambda p: p is not None, module_._parameters.values()))
                 )
             else:
-                for name, param in module_.named_parameters():
+                for name, param in module_._parameters.items():
                     if param is None:
                         continue
                     if name.endswith('bias'):
--- a/nemo/collections/nlp/modules/common/megatron/utils.py
+++ b/nemo/collections/nlp/modules/common/megatron/utils.py
@@ -366,17 +366,27 @@ def get_params_for_weight_decay_optimization(
     is_expert = lambda param: not getattr(param, 'allreduce', True)
     # Do the actual param classification
     for module in modules:
-        for name, param in module.named_parameters():
-            if param is None:
-                continue
-            if name.endswith('.bias'):
-                no_weight_decay_params['params'].extend([param])
+        for module_ in module.modules():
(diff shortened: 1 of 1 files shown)


akoumpa:
jenkins

mikolajblaz:
Thanks @akoumpa. This change **does** fix the mlperf backward-compatibility, although I'm worried it might introduce another one for different scenarios, i.e. for models that previously had no params in "no_weight_decay_params" bucket? Did you test this case maybe?

akoumpa:
jenkins

akoumpa:
@ericharper will remove draft once jenkins passes.
@mikolajblaz I assumed that if a bucket does not have any parameters it should not be part of optimizer's param groups, I'm not sure what the use-case would be here since it's empty.

mikolajblaz:
> @ericharper will remove draft once jenkins passes. @mikolajblaz I assumed that if a bucket does not have any parameters it should not be part of optimizer's param groups, I'm not sure what the use-case would be here since it's empty.

I totally agree, just not sure what are the consequences of omitting that group altogether in terms of backwards compatibility

akoumpa:
@mikolajblaz I think it's fine, the code will yield a group if it has parameters, consumer code does not appear to make explicit assumptions on the number of returned groups and from the PyTorch documentation that also seems to be the right way to handle it. The problem would be to yield groups with empty parameters, rather than the current case. If you have an explicit example in mind happy to look into it. Apex also [seems ok](https://github.com/NVIDIA/apex/blob/141bbf1cf362d4ca4d94f4284393e91dda5105a5/apex/contrib/optimizers/distributed_fused_adam.py#L523) with Iterable[dict]

mikolajblaz:
> @mikolajblaz I think it's fine, the code will yield a group if it has parameters, consumer code does not appear to make explicit assumptions on the number of returned groups and from the PyTorch documentation that also seems to be the right way to handle it. The problem would be to yield groups with empty parameters, rather than the current case. If you have an explicit example in mind happy to look into it. Apex also [seems ok](https://github.com/NVIDIA/apex/blob/141bbf1cf362d4ca4d94f4284393e91dda5105a5/apex/contrib/optimizers/distributed_fused_adam.py#L523) with Iterable[dict]

ok, thanks for checking
I would approve the PR but I don't think I can

ShriyaPalsamudram (approved):
Approving based on the latest comment

PR #8865, opened 2024-04-09
dehardcode test string

JimmyZhang12:
# What does this PR do ?

By default `to_word_list_format` uses '<extra_id_1>' as its test string to map a string to its tokens, but using '<extra_id_1>' may not map correctly, so add option to specifiy the test string.

# Changelog 

# Usage

```python
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/export/trt_llm/nemo_utils.py
```diff
@@ -213,7 +215,7 @@ def to_word_list_format(word_dict: List[List[str]], tokenizer=None):
     # We use a similar trick as in NeMo to deal with the fact that the encoding of a single word
     # can't always be trusted. See
     #   https://github.com/NVIDIA/NeMo/blob/bb575b72fd0be51ae10cc77d9f89ddb9e9d3b96d/nemo/collections/nlp/modules/common/text_generation_strategy.py#L229
```
odelalleau, line 217: Can you please edit this comment? It looks like this was copy/pasted from NeMo-FW and it seems weird now that it's in NeMo. Something like:
```
# We use a similar trick as when generating text to deal with the fact that the encoding of a single word
# can't always be trusted. See
#   https://github.com/NVIDIA/NeMo/blob/bb575b72fd0be51ae10cc77d9f89ddb9e9d3b96d/nemo/collections/nlp/modules/common/text_generation_strategy.py#L229
```

Ideally this logic would be factored in a single place but I'm not sure if this is easy to do (may also not be worth the trouble since I believe the generation code will be deprecated at some point in favor of its upcoming mcore version).

odelalleau (commented):
Just a minor request to update the comment.

Besides this lgtm, but I just want to point out that using `<extra_id_1>` as stop word with the Llama2 tokenizer is a bad idea since as you mentioned it may be tokenized in various ways when merged with what comes before. When using the text generation from NeMo at least we fall back to the (slower) string matching ([here](https://github.com/NVIDIA/NeMo/blob/3f3df1c34aa73cfc890b93af678e3c3a63760fe5/nemo/collections/nlp/modules/common/text_generation_strategy.py#L251)), but with TRT-LLM I don't know if such a string matching exists (?) => one solution could be to postprocess the output to be sure we stop on the first stop word (this has nothing to do with this PR though -- just mentioning it in case this would be relevant to your work)

nemo/export/trt_llm/nemo_utils.py
jiemingz: Sure will do. We did ask TRTLLM for string level matching but they felt like it was infeasible to implement.

odelalleau: Yeah there's no obvious way to achieve this (that I know of). The best I can think of would be to have an option to regularly check the last chunk of text that has been generated to see if/when we need to stop (doing it on each step and over the whole input as in NeMo is overkill and very inefficient).

oyilmaz-nvidia (approved):
LGTM.

JimmyZhang12:
Can this be merged @oyilmaz-nvidia ?

odelalleau (approved):

oyilmaz-nvidia (approved):
LGTM, thanks!

oyilmaz-nvidia:
jenkins

oyilmaz-nvidia (approved):

PR #8972, opened 2024-04-18
RNN-T and TDT inference: use CUDA graphs by default

artbataev:
# What does this PR do ?

- enable CUDA graphs by default for inference (`transcribe_speech.py`, `speech_to_text_eval.py`)
- add fallback behavior with partial CUDA graphs for Label-Looping algorithm implementation (TDT, RNN-T)
  - Use Cuda graphs _with while loops_ when all requirements are met (driver 545+, Cuda 12.3+, cuda-python package)
  - Use Cuda graphs _without while loops_ (pure PyTorch functionality) in most other cases. The idea: use several separate Cuda graphs for parts of the decoding algorithm (`<graph before outer loop>` -> `while loop (python)` -> `<graph before inner loop>` -> `inner while loop (python)` -> `<graph for inner loop code>` etc.)

On my local machine, FastConformer L, LibriSpeech test-other decoding time, bfloat16, bs=16

|                | no graphs | partial graphs | full graph |
| -------------- | --------- | -------------- | ---------- |
| unsorted batch | 34s       | 21s            | 19s        |
| sorted batch   | 23s       | 15s            | 14s        |

**Collection**: [ASR]

# Changelog 

# Usage

```shell
python examples/asr/speech_to_text_eval.py <with rnnt/tdt model>
```

# Jenkins CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


artbataev:
jenkins

artbataev:
jenkins

artbataev:
@galv, @titu1994 Please, review the PR

The only issue I see for now is that the Frame-Looping algorithm does not have fallback behavior, and setting  `loop_labels=false` without `use_cuda_graph_decoder=false` can cause failures (if not all requirements are met for Frame-Looping+CUDA graphs).

On the other hand, this case is not important. Most users will use the Label-Looping algorithm since it is the default and produces the same result as the Frame-Looping algorithm.

artbataev:
jenkins

galv:
There is something I'm not quite understanding. Is there a reason why you did not set the default value of these values to true? 

https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L592

and

https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L2305

and

https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L2655

It looks like you are turning cuda graphs on by default only when someone runs transcribe_speech.py or transcribe_speech_parallel.py right now.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -430,16 +470,27 @@ def loop_labels_cuda_graphs(
...
-        # while self.state.active_mask_any.item():
-        #     self._before_inner_loop_get_decoder_output()
-        #     self._before_inner_loop_get_joint_output()
-        #     while self.state.advance_mask_any.item():
-        #         self._inner_loop_code()
-        #     self._after_inner_loop()
+        if self.cuda_graphs_mode is self.CudaGraphsMode.FULL_GRAPH:
+            self.full_graph.replay()
+        elif self.cuda_graphs_mode is self.CudaGraphsMode.NO_WHILE_LOOPS:
+            self.separate_graphs.before_outer_loop.replay()
```
galv, line 476: This is cool. I would like to speak to you about a way we could possibly do this more easily using torch.cond and torch.while_loop. They're not in the newest versions of pytorch yet.

As we improve our implementations of beam search, I don't think it is realistic to keep doing special case code like this. I'm thinking we can get the right level of abstraction with torch.cond and torch.while_loop, such that people can write things more naturally.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -207,20 +230,35 @@ def __init__(
...
-        if self.use_cuda_graphs and self.max_symbols is None:
-            logging.warning("Max symbols is None, which is not allowed with Cuda graphs.")
-            self.use_cuda_graphs = False
+        if not allow_cuda_graphs:
+            self.cuda_graphs_mode = None
+        else:
+            # cuda graphs are allowed
+            # check basic requirements for cuda graphs
+            if self.max_symbols is None:
+                logging.warning("Max symbols is None, which is not allowed with Cuda graphs.")
```
galv, line 239: People don't check warnings often enough. I recommend you throw an exception here.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -207,20 +230,35 @@ def __init__(
...
-                check_cuda_python_cuda_graphs_conditional_nodes_supported()
-            except ImportError as e:
-                logging.warning(f"No conditional node support. Cuda graphs will be disabled,\n{e.msg}")
-                self.use_cuda_graphs = False
+        self.state = None
+        self.full_graph = None
+        self.separate_graphs = None
 
-        self.state: Optional[LoopLabelsState] = None
+    def force_cuda_graphs_mode(self, mode: Optional[Union[str, CudaGraphsMode]]):
```
galv, line 258: Do we need this? I don't see any usages of it.

galv:
> On the other hand, this case is not important. Most users will use the Label-Looping algorithm since it is default and produces the same result as the Frame-Looping algorithm.

I agree. That is was my first attempt. It was very educational, but I don't believe that we need to do the work to add a feature to it that won't be used very much.

galv:
I didn't hit approve yet. The changes seem good to me. But I want to give some time to defer to @titu1994

examples/asr/transcribe_speech.py
```diff
@@ -161,7 +162,9 @@ class TranscriptionConfig:
     ctc_decoding: CTCDecodingConfig = CTCDecodingConfig()
 
     # Decoding strategy for RNNT models
-    rnnt_decoding: RNNTDecodingConfig = RNNTDecodingConfig(fused_batch_size=-1)
+    rnnt_decoding: RNNTDecodingConfig = RNNTDecodingConfig(
+        fused_batch_size=-1, greedy=GreedyBatchedRNNTInferConfig(use_cuda_graph_decoder=True)
```
titu1994, line 166: You can set the default config inside `GreedyBatchedRNNTInferConfig` to have this set to True by default rather than do it with this explicit override

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
titu1994: Max symbols is none on some older models if i recall, that means out of the box they will crash to infer due to error for unrelated reason of cuda graphs not supporting None. It should remain a warning, but i agree that instead of crashing, set it to large default value for cuda graphs..

Ie warn that its not supported so instead a default of 10 timesteps or higher is being used for cuda graphs optimization

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -207,20 +230,35 @@ def __init__(
...
+            if self.max_symbols is None:
+                logging.warning("Max symbols is None, which is not allowed with Cuda graphs.")
+                self.cuda_graphs_mode = None
+            else:
+                # basic requirements met, need to check while loops
+                try:
+                    check_cuda_python_cuda_graphs_conditional_nodes_supported()
+                    self.cuda_graphs_mode = self.CudaGraphsMode.FULL_GRAPH
+                except (ImportError, ModuleNotFoundError) as e:
+                    logging.warning(
```
titu1994, line 247: Im wondering if this should be visible to users. The problem is the vast majority of users will NOT be on latest driver and cudapython install - ie the vast majority wont be using Cuda graphs and will log this warning - repeatedly polluting the inference in a loop call to transcribe().

We need to make this log just once - see loggermode and how to pass it inside of logging messages.

tests/collections/asr/decoding/test_cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -11,19 +11,38 @@
...
 import pytest
 import torch
-from omegaconf import OmegaConf, open_dict
+from omegaconf import open_dict
 
 from nemo.collections.asr.models import ASRModel
 from nemo.core.utils.cuda_python_utils import skip_cuda_python_test_if_cuda_graphs_conditional_nodes_not_supported
 
 
+@pytest.fixture(scope="module")
```
titu1994, line 26: Shouldnt this be with a marker for `with_download`

titu1994 (commented):
Some important comments for logging, otherwise looks good

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
galv: I agree with Som's suggestion, actually. Using a large value like 10 when someone passes in None seems like the right move.

artbataev:
> There is something I'm not quite understanding. Is there a reason why you did not set the default value of these values to true?
> 
> https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L592
> 
> and
> 
> https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L2305
> 
> and
> 
> https://github.com/NVIDIA/NeMo/blob/7e805bc8d0683cc76fa44b6a847611aaaedb9794/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L2655
> 
> It looks like you are turning cuda graphs on by default only when someone runs transcribe_speech.py or transcribe_speech_parallel.py right now.

This is done intentionally. The problem is that using `True` as a default value here will result in using CUDA graphs when decoding in training mode, which is can result in issues with bucketing (dynamic batch sizes).
I think it is cleaner for now to use CUDA graphs only for transcribing speech.

See `training_step` for RNN-T models 
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/models/rnnt_models.py#L663
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/modules/rnnt.py#L1490
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/metrics/wer.py#L327

tests/collections/asr/decoding/test_cuda_graph_rnnt_greedy_decoding.py
artbataev: Pytest marks (`with_downloads`) are useful only for tests, not fixtures. All the tests that use these fixtures are marked with `with_downloads` tag.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
artbataev: As I see it, the `transcribe` does not change the decoding strategy. Since this is logged only when instantiating the class (when instantiating the model or changing the decoding strategy), this should be fine (no repeated logs).

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
artbataev: Fully agree, thanks! I fixed the behavior.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
artbataev: Thanks! I forgot to add an explicit test. Fixed.
This is useful for debugging to set `no_graphs` mode (since it's impossible to debug CUDA graphs directly).

examples/asr/transcribe_speech.py
artbataev: Same issue as with setting `use_cuda_graph_decoder=True` as default in the classes.
I added comments above this line.
`RNNTDecodingConfig` is used, e.g., in `change_vocabulary`. After this operation (useful for finetuning), the model will have the field in the config `use_cuda_graph_decoder=True`, and further training will use the CUDA decoder. Since it is not compatible with bucketing (pre-allocated memory for maximum batch_size*sequence_length can be too large in this case), I prefer to conservatively enable it only for transcription.

The alternative is to enable it everywhere by default, but in the training loop, explicitly use the decoder without CUDA graphs. However, this can make the code too complicated.

If you see that there is a more straightforward solution for this issue, please let's discuss it!

artbataev: > The alternative is to enable it everywhere by default, but in the training loop, explicitly use the decoder without CUDA graphs. However, this can make the code too complicated.
Moved to this solution. I think it is much cleaner. So, set everywhere `use_cuda_graph_decoder=True` by default

artbataev:
@galv, @titu1994 Please, review once again

I reworked the PR.
- CUDA graphs enabled by default now in all classes but turned off in the training pipeline
- I also added more tests and fixed some issues with confidence
- Now max_symbols is set to 10 if it is initially None (with an appropriate warning)

nemo/collections/asr/models/rnnt_models.py
```diff
@@ -660,6 +661,16 @@ def forward(
         return encoded, encoded_len
 
     # PTL-specific methods
+    def on_train_epoch_start(self) -> None:
```
titu1994, line 664: This is not good. Very easy to forget to add this in so many model classes (hybrid for example, ASR TTS for another). 

How about instead checking the decoding training type inside of greedy decoding itself and use that to decide whether to turn off or on cuda graphs.

titu1994: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py#L372

artbataev: At first I thought this is an excellent place to toggle CUDA graphs.
But after reviewing the changes, I think it's better to use explicit `on_train_epoch_start`:
- disabling CUDA graphs in `on_train_epoch_start` is better, since this operation frees up the memory. This is critical for training, and the amount of memory can be large, especially if the alignments are preserved. This is impossible to do in the decoder when the training starts, but only after some training steps are done and we request decoding for the batch.
- hybrid CTC-RNNT is not a problem since it inherits `EncDecRNNTModel` (same for `EncDecRNNTBPEModel`)
- I fixed the ASR-TTS model. Thanks for catching the bug!

titu1994: > fixed the ASR-TTS model. Thanks for catching the bug!

See this is exactly the problem. I or future reviewers won't always remember every possible model that can be affected by this. 

Lets either put this inside the place i mentioned or put this in ASR Model and make it generic to all ASR models. Doing it per class is not future proof.

nemo/collections/asr/models/asr_model.py
```diff
@@ -171,6 +173,51 @@ def on_after_backward(self):
                 logging.warning(f'detected inf or nan values in gradients! Setting gradients to zero.')
                 self.zero_grad()
 
+    @classmethod
```
titu1994, line 176: These two methods should be class methods of WithOptionalCYDAGraphs

nemo/collections/asr/models/asr_model.py
```diff
@@ -171,6 +173,51 @@ def on_after_backward(self):
...
+                and hasattr(submodule.decoding, "decoding")
+                and isinstance(submodule.decoding.decoding, WithOptionalCudaGraphs)
+            ):
+                submodule.decoding.decoding.maybe_enable_cuda_graphs()
+
+    def on_train_epoch_start(self) -> None:
+        """Decoder with CUDA graphs does not release memory, thus we disable it for training epoch"""
+        self.disable_cuda_graphs_in_decoder(self)
+
+    def on_train_epoch_end(self) -> None:
```
titu1994, line 209: What about cleanup of on val test predict epoch end ?

nemo/collections/asr/models/asr_model.py
artbataev: I think, generally, this is not a good idea to introduce a 2-way dependency `WithOptionalCudaGraphs` <-> `ASRModel` (actually, `EncDecRNNTModel`, since decoding is only in this model).
I made the method more abstract to separate the logic, separating the path in the model and the lookup logic.

nemo/collections/asr/models/asr_model.py
artbataev: - `Training`: We want to disable graphs only in training. So for training, we enable it when training starts and disable it when training ends.
- `Val`: validation can be done separately or inside the training loop. We need to disable CUDA graphs after validation to avoid them in training https://lightning.ai/docs/pytorch/stable/common/lightning_module.html#hooks
- `Test`: We do not need to disable CUDA graphs; we need to enable them only since `test` cannot be used in training.
- `Predict`: Same as `test`

I added comments about this logic to the hooks.

titu1994 (approved):
Looks good to me, @galv for final approval

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -618,20 +620,74 @@ def __init__(
...
-
-                self._greedy_decode = RNNTGreedyDecodeCudaGraph(max_symbols_per_step, self)
             else:
-                # previous algo: loop over frames
-                self._greedy_decode = self._greedy_decode_blank_as_pad_loop_frames
+                # Frame-Looping algorithm
+                if not self.use_cuda_graph_decoder:
+                    self._greedy_decode = self._greedy_decode_blank_as_pad_loop_frames
+                else:
+                    if self.preserve_alignments:
```
galv, line 630: I'm surprsied that you would silently change the behavior  on lines 630 to 639 rather than throw an exception in these cases, to be honest. 

Meanwhile, the situation where we set the symbols_per_step to 10 if it is None seems okay because it is unlikely to change the results, since 10 is such a large number.

I'm not going to hold up merging this because of this concern, anyway, since it is a code path most people won't see.

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
```diff
@@ -618,20 +620,74 @@ def __init__(
...
+                    if not torch.cuda.is_available():
+                        self.use_cuda_graph_decoder = False
+
+                    if self.use_cuda_graph_decoder:
+                        try:
+                            from nemo.collections.asr.parts.submodules.cuda_graph_rnnt_greedy_decoding import (
+                                RNNTGreedyDecodeCudaGraph,
+                            )
+
+                            self._greedy_decode = RNNTGreedyDecodeCudaGraph(max_symbols_per_step, self)
```
galv, line 647: I am not certain, but it currently looks like we will throw an exception if max_symbols_per_step is None, rather than overriding it to 10 for the frame-loop decoder right now.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -612,12 +728,13 @@ def _before_inner_loop_get_joint_output(self):
...
         self.state.time_indices_current_labels.copy_(self.state.time_indices, non_blocking=True)
         if self.state.alignments is not None:
+            float_dtype = self.state.float_dtype
             self.state.alignments.add_results_masked_no_checks_(
                 active_mask=self.state.active_mask,
                 time_indices=self.state.time_indices_current_labels,
                 logits=logits if self.preserve_alignments else None,
                 labels=self.state.labels if self.preserve_alignments else None,
-                confidence=self._get_confidence_tensor(F.log_softmax(logits, dim=-1))
+                confidence=self._get_confidence_tensor(F.log_softmax(logits, dim=-1)).to(dtype=float_dtype)
```
galv, line 737: As a point of clarification, why did you need to add this to() call? It doesn't seem to be related to the rest of the changes.

galv (approved):
The issue you raised about memory usage when doing inference during training is a very good one. I believe we can figure out a solution that is less instrusive in the future if we point it out to the right people (follow up with me!).

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
artbataev: Yep, thanks for catching this. I will address this in a follow-up PR

nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py
artbataev: I made this fallback behavior to prevent crashes when the user wants to change some parameters since `use_cuda_graph_decoder` is `True` by default now. Since it's only about speed (not quality), it is acceptable to switch silently between implementations instead of requiring the user to understand all the nuances of the available parameter combinations.
`LoopLabelsComputer`(s) are designed to handle all situations without explicit errors (e.g., when cuda is unavailable, etc.).

nemo/collections/asr/models/asr_model.py
```diff
@@ -171,6 +172,52 @@ def on_after_backward(self):
...
+        WithOptionalCudaGraphs.enable_cuda_graphs_recursive(self, attribute_path="decoding.decoding")
+
+    def on_validation_epoch_end(self) -> Optional[dict[str, dict[str, torch.Tensor]]]:
+        """
+        After validation, we disable CUDA graphs, since `validation` can be called in training loop, and
+        training will continue after validation
+        EncDecRNNTModel.decoding.decoding is the inference class with CUDA graphs.
+        """
+        WithOptionalCudaGraphs.disable_cuda_graphs_recursive(self, attribute_path="decoding.decoding")
+        return super().on_validation_epoch_end()
```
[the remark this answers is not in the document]
artbataev: These hooks return `None` in the PyTorch-Lightning interface, and basically, there is no code in such hooks. But `ModelPT` defines the `on_validation_epoch_end` hook for all models with the customized return type, so I need to call it.

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
artbataev: I would be happy to avoid this, but without casting, the code will fail with mixed bf16 precision:
- `log_softmax` returns the value of `float32` type in bfloat16 mixed precision (amp)
- alignments storage is initialized with bf16 type, and adding confidence values inside `add_results_masked_no_checks_` will fail

This is the same issue I observed when reviewing Sasha's PR related to TDT confidence #8982 
Since I enabled computing confidence in tests for RNN-T, I caught this bug and fixed it here https://github.com/NVIDIA/NeMo/pull/8972/files#diff-d8ba9ce8e77769e06174cf0d16842d130debb4a289e92fea5296b081f5a4deabR133

PR #9001, opened 2024-04-22
Improved random seed configuration for Lhotse dataloaders with docs

pzelasko:
# What does this PR do ?

* Comprehensive documentation about random seed configuration, behavior, and effects on dataloading.
* Backward-compatible changes to random seed configuration in training.
* Allows more precise control over randomness; this work was triggered by the need to support mixed data-parallel (DP) and tensor-parallel (TP) training. Before this PR, it wasn't possible to configure the dataloader to yield identical data on identical DP ranks but different TP ranks, causing deadlocks. Now it's possible but the user has to be careful about manually setting different random seeds on training continuations (this is documented).
* Unit tests to ensure the desired behavior in DDP setups.

**Collection**: ASR

# Changelog 

# Usage

```python
```

# Jenkins CI

There's no need to comment `jenkins` on the PR to trigger Jenkins CI.
The GitHub Actions CI will run automatically when the PR is opened.
To run CI on an untrusted fork, a NeMo user with write access must click "Approve and run".

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [x] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


docs/source/asr/datasets.rst
```diff
@@ -823,6 +823,52 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+
+    * The ``randomized`` setting uses ``seed`` along with DDP ``rank`` and dataloading ``worker_id`` to set a unique but deterministic random seed in each dataloading process across all GPUs.
+
+    * Each node/GPU/dataloading worker yields data in a different order (no mini-batch duplication).
+
+    * On each training script run, the order of dataloader examples are **identical** as long as ``seed`` is the same.
+
+    * This setup guarantees 100% dataloading reproducibility.
+
+    * Resuming training without changing of the ``seed`` value will cause the model to train on data it has already seen. For large data setups, not managing the ``seed`` may cause the model to never be trained on a majority of data. This is why this mode is not the default.
```
zhehuaichen, line 860: what's the recommended practice on managing the "seed" on large data setup?

zhehuaichen (commented):
Looks good from my side.
Thanks for the quick fix!

docs/source/asr/datasets.rst
pzelasko: Generally every time you resume, you'd provide a different value to `model.train_ds.seed=<val>`. A true-enough random seed can be obtained on most systems by reading /dev/urandom, e.g. uint32 seed: `RSEED=$(od -An -N4 -tu4 < /dev/urandom | tr -d ' ')`. If you have some sort of "launcher script" that queues multiple jobs, this would be the right place to use this.

Ideally we'd be able to automate this seed management thing by keeping some state in the checkpoints, but at this point it'd be a scope creep.

docs/source/asr/datasets.rst
```diff
@@ -823,6 +823,54 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+Seeds and randomness
+~~~~~~~~~~~~~~~~~~~~
+
+In Lhotse dataloading configuration we have two parameters controlling randomness: ``seed`` and ``shard_seed``.
+Both of them can be either set to a fixed number, or one of two string options ``"randomized"`` and ``"trng"``.
+Their roles are:
+
+    * ``seed`` is the base random seed, and is one of several factors used to initialize various RNGs participating in dataloading.
+
+    * ``shard_seed`` controls the shard randomization strategy in distributed data parallel setups when using sharded tarred datasets.
```
erastorgueva-nv, line 836:
```suggestion
* ``seed`` is the base random seed, and is one of several factors used to initialize various RNGs participating in dataloading.

* ``shard_seed`` controls the shard randomization strategy in distributed data parallel setups when using sharded tarred datasets.
```

docs/source/asr/datasets.rst
```diff
@@ -823,6 +823,54 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+
+    * Generally the seed can be managed by the user by providing a different value each time the training script is launched. For example, for most models the option to override would be ``model.train_ds.seed=<value>``. If you're launching multiple tasks queued one after another on a grid system, you can generate a different random seed for each task, e.g. on most Unix systems ``RSEED=$(od -An -N4 -tu4 < /dev/urandom | tr -d ' ')`` would generate a random uint32 number that can be provided as the seed.
+
+Other, more exotic configurations:
+
+    * With ``shard_seed=<int>``, all dataloading workers will yield the same results. This is only useful for unit testing and maybe debugging.
+
+    * With ``seed="trng"``, the base random seed itself will be drawn using a TRNG. It will be different on each GPU training process. This setting is not recommended.
+
+    * With ``seed="randomized"``, the base random seed is set to Python's global RNG seed. It might be different on each GPU training process. This setting is not recommended.
```
erastorgueva-nv, line 872:
```suggestion
* With ``shard_seed=<int>``, all dataloading workers will yield the same results. This is only useful for unit testing and maybe debugging.

* With ``seed="trng"``, the base random seed itself will be drawn using a TRNG. It will be different on each GPU training process. This setting is not recommended.

* With ``seed="randomized"``, the base random seed is set to Python's global RNG seed. It might be different on each GPU training process. This setting is not recommended.
```

docs/source/asr/datasets.rst
```diff
@@ -823,6 +823,54 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+
+    * This setup guarantees 100% dataloading reproducibility.
+
+    * Resuming training without changing of the ``seed`` value will cause the model to train on data it has already seen. For large data setups, not managing the ``seed`` may cause the model to never be trained on a majority of data. This is why this mode is not the default.
+
+    * If you're combining DDP with model parallelism techniques (Tensor Parallel, Pipeline Parallel, etc.) you need to use ``shard_seed="randomized"``. Using ``"trng"`` will cause different model parallel ranks to desynchronize and cause a deadlock.
+
+    * Generally the seed can be managed by the user by providing a different value each time the training script is launched. For example, for most models the option to override would be ``model.train_ds.seed=<value>``. If you're launching multiple tasks queued one after another on a grid system, you can generate a different random seed for each task, e.g. on most Unix systems ``RSEED=$(od -An -N4 -tu4 < /dev/urandom | tr -d ' ')`` would generate a random uint32 number that can be provided as the seed.
```
erastorgueva-nv, line 864:
```suggestion
* If you're combining DDP with model parallelism techniques (Tensor Parallel, Pipeline Parallel, etc.) you need to use ``shard_seed="randomized"``. Using ``"trng"`` will cause different model parallel ranks to desynchronize and cause a deadlock.

* Generally the seed can be managed by the user by providing a different value each time the training script is launched. For example, for most models the option to override would be ``model.train_ds.seed=<value>``. If you're launching multiple tasks queued one after another on a grid system, you can generate a different random seed for each task, e.g. on most Unix systems ``RSEED=$(od -An -N4 -tu4 < /dev/urandom | tr -d ' ')`` would generate a random uint32 number that can be provided as the seed.
```

docs/source/asr/datasets.rst
```diff
@@ -823,6 +823,54 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+
+    * The ``randomized`` setting uses ``seed`` along with DDP ``rank`` and dataloading ``worker_id`` to set a unique but deterministic random seed in each dataloading process across all GPUs.
+
+    * Each node/GPU/dataloading worker yields data in a different order (no mini-batch duplication).
+
+    * On each training script run, the order of dataloader examples are **identical** as long as ``seed`` is the same.
+
+    * This setup guarantees 100% dataloading reproducibility.
+
+    * Resuming training without changing of the ``seed`` value will cause the model to train on data it has already seen. For large data setups, not managing the ``seed`` may cause the model to never be trained on a majority of data. This is why this mode is not the default.
```
erastorgueva-nv, line 860:
```suggestion
* The ``trng`` setting discards ``seed`` and causes the actual random seed to be drawn using OS's true RNG. Each node/GPU/dataloading worker draws its own unique random seed when it first needs it.

* Each node/GPU/dataloading worker yields data in a different order (no mini-batch duplication).

* On each training script run, the order of dataloader examples are **different**.

* Since the random seed is unpredictable, the exact dataloading order is not replicable.

Case 2: ``seed=<int>`` and ``shard_seed="randomized"``:

* The ``randomized`` setting uses ``seed`` along with DDP ``rank`` and dataloading ``worker_id`` to set a unique but deterministic random seed in each dataloading process across all GPUs.

* Each node/GPU/dataloading worker yields data in a different order (no mini-batch duplication).

* On each training script run, the order of dataloader examples are **identical** as long as ``seed`` is the same.

* This setup guarantees 100% dataloading reproducibility.

* Resuming training without changing of the ``seed`` value will cause the model to train on data it has already seen. For large data setups, not managing the ``seed`` may cause the model to never be trained on a majority of data. This is why this mode is not the default.
```

erastorgueva-nv (changes requested):
Recommend to get rid of the tabs before the bullet points. Currently, with the tabs, the bullet points are placed inside block quotes

pzelasko:
Thanks @erastorgueva-nv, for some reason I thought it's required by sphinx. It should look much better indeed!

erastorgueva-nv (approved):

erastorgueva-nv:
Docs LGTM

zhehuaichen (approved):
LGTM from my side

PR #9086, opened 2024-05-02
Draft: Fix the "cast ping pong" problem when we run AMP inference.

galv:
This has been tested only for Parakeet-CTC-1.1B right now. This problem certainly exists elsewhere.

It also is not ina  mergeable state. This is my initial hack to fix the problem.

Automatic mixed precision and inference do not play well together.

Frankly, I do not think we should be using AMP in inference in the first place.

First, automatic mixed precision was created back when neural networks were much simpler. In particular, they did not have softmax and layer norm as frequent operations. In the era of transformers, softmax and layer norm are very common. AMP will uncoditionally output fp32 outputs from these operations, even if their inputs are fp16. See here: https://pytorch.org/docs/stable/amp.html#cuda-ops-that-can-autocast-to-float32

This is no longer necessary, now that layer norm does accumulation in fp32 in pytorch, even if the input is fp16:
https://github.com/pytorch/pytorch/issues/66707

There is a second problem. It is elaborated on in my long comment in nemo/core/classes/module.py. Basically, setting requires_grad=False on a parameter, confusingly enough, disables AMP's caching mechanism, which means we uncoditionally waste time calling cache kernels. Read more here: https://github.com/NVIDIA/NeMo/pull/9086/files#diff-4f86565ac3601ea79395f9d2884c28115c5fd9fe3655894564d71539fe5d8042R58-R90

Both of these problems need to be fixed to prevent inference from constantly casting its intermediate tensors to and from float16 and float32, which is a waste of time that does nothing.

This hacky fix does not suppoer bfloat16 for now.

Again, I recommend simply running model.half() or model.bfloat16() for inference. AMP does not make sense for inference because it wastes memory.

Results from running:

```
batch_size=32
amp=true

echo "GALVEZ: Conformer CTC"
python examples/asr/speech_to_text_eval.py  pretrained_name=nvidia/parakeet-ctc-1.1b dataset_manifest=/home/dgalvez/scratch/data/test_other_sorted_downward.json  batch_size=$batch_size  output_filename=test_clean_decoded.jsonl  amp=$amp  amp_dtype=float16  use_cer=false num_workers=1
\# ctc_decoding.greedy.batched_inference=true
```

RTFx values before my change:

```
922.302270024654
1072.6919226908824
853.5230428014742
1114.2164255498367
1264.829085006686
```

RTFx values after my change:

```
1132.0813676953462
1255.840791955383
838.7445635257388
1389.7852550162668
1479.8061656119771
```

You can see that RTFx gets as high as 1480 after my change. Meanwhile, the highest RTFx before my change is 1264. So it's about a 15% speedup by removing the unnecessary cast kernels.

Forgive the large variance from run-to-run. It is because of the changes made in: https://github.com/NVIDIA/NeMo/pull/8521 There is a huge delay (at least 10 milliseconds) when there is a cache miss for the caching pinned host memory allocator. I'm planning to revert the change since I didn't realize initially how large the cache miss cost was (enough to undue the benefits on reasonably sized workloads).

### DIFF
--- a/examples/asr/transcribe_speech.py
+++ b/examples/asr/transcribe_speech.py
@@ -253,7 +253,10 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
 
     trainer = pl.Trainer(devices=device, accelerator=accelerator)
     asr_model.set_trainer(trainer)
+
+    # assert all(param.requires_grad for param in asr_model.parameters())
     asr_model = asr_model.eval()
+    # assert all(param.requires_grad for param in asr_model.parameters())
 
     # we will adjust this flag if the model does not support it
     compute_timestamps = cfg.compute_timestamps
@@ -407,7 +410,27 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
                 override_cfg.augmentor = augmentor
                 override_cfg.text_field = cfg.gt_text_attr_name
                 override_cfg.lang_field = cfg.gt_lang_attr_name
-                transcriptions = asr_model.transcribe(audio=filepaths, override_config=override_cfg,)
+                # assert all(param.requires_grad for param in asr_model.parameters())
+                for i in range(5):
+                    if i == 1:
+                        # import nvtx
+                        # pr = nvtx.Profile()
+                        # pr.enable()  # begin annotating function calls
+                        # ctx = torch.autograd.profiler.emit_nvtx()
+                        # ctx.__enter__()
+                        torch.cuda.cudart().cudaProfilerStart()
+                    import time
+                    start_time = time.time()
+                    # assert all(param.requires_grad for param in asr_model.parameters())
+                    transcriptions = asr_model.transcribe(audio=filepaths, override_config=override_cfg,)
+                    # assert all(param.requires_grad for param in asr_model.parameters())
+                    end_time = time.time()
+                    print(5.1 * 60 * 60 / (end_time - start_time))
+                    if i == 1:
+                        # pr.disable()
+                        # ctx.__exit__(None, None, None)
+                        torch.cuda.cudart().cudaProfilerStop()
+
 
     if cfg.dataset_manifest is not None:
         logging.info(f"Finished transcribing from manifest file: {cfg.dataset_manifest}")
--- a/examples/asr/transcribe_speech.py
+++ b/examples/asr/transcribe_speech.py
@@ -420,6 +420,7 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
                         # ctx.__enter__()
                         torch.cuda.cudart().cudaProfilerStart()
                     import time
+
                     start_time = time.time()
                     # assert all(param.requires_grad for param in asr_model.parameters())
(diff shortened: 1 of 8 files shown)


nemo/collections/asr/models/ctc_models.py
```diff
@@ -655,18 +656,19 @@ def test_dataloader(self):
     def _transcribe_on_begin(self, audio, trcfg: TranscribeConfig):
         super()._transcribe_on_begin(audio, trcfg)
 
-        # Freeze the encoder and decoure_exder modules
-        self.encoder.freeze()
-        self.decoder.freeze()
+        # Freeze the encoder and decoder modules
```
titu1994, line 659: Can you revert all the removed sections that call freeze and unfreeze / as_frozen please ? 

We had a memory leak there for RNNT, it's fixed via inference_mode() but more often than note users do not remember to use it and face the memory leak. It is better for the freeze and unfreeze logic to be inherent on call so that even by mistake such leak is not possible

titu1994 (changes requested):

nemo/collections/asr/models/ctc_models.py
galv: That is precisely the problem. If you don't want to use torch.inference_mode() or torch.no_grad(), but instead rely on setting the requires_grad field of a tensor to false, then AMP will *always* fail to use its "casted tensor cache". So it will repeatedly cast your float32 weights to float16, which is what makes inference slow for all of our models.  It is because of the args.requires_grad() condition in this code: https://github.com/pytorch/pytorch/blob/6f5f405b057c7de0f5fce0b1432cb74468f96f95/aten/src/ATen/autocast_mode.cpp#L121C61-L121C80

The code is a rough draft for a reason. This is one way to fix the problem, but I don't think it's right because it is so fragile. I think that simply casting the model to float16 or bfloat16 would be more appropriate.

titu1994: How about we avoid mixed precision then and require full fp16 or bf16 cast (apart from all BN uses, which need to be wrapped to always cast to 32 no matter what, similar to how it's done in metas ASR library)

galv: I think that is the right solution. I'm frankly surprised NeMo uses AMP for inference, because it increases weight memory usage and isn't intended to be used for inference (and no other toolkits use it for inference either).

Of course... a lot of code is written around the assumption of AMP being used for inference. Take a look at the do_amp flag in in transcribe_speech.py, etc. So it's a heavy lift.

examples/asr/transcribe_speech.py
```diff
@@ -407,7 +410,27 @@ def autocast(dtype=None):
...
+                    start_time = time.time()
+                    # assert all(param.requires_grad for param in asr_model.parameters())
+                    transcriptions = asr_model.transcribe(audio=filepaths, override_config=override_cfg,)
+                    # assert all(param.requires_grad for param in asr_model.parameters())
+                    end_time = time.time()
+                    print(5.1 * 60 * 60 / (end_time - start_time))
+                    if i == 1:
+                        # pr.disable()
+                        # ctx.__exit__(None, None, None)
+                        torch.cuda.cudart().cudaProfilerStop()
```
nithinraok, line 433: These modifications are not necessary for transcribe but to show RTFx and numbers are only applicable for LS

galv:
#9198 supersedes this.

PR #9099, opened 2024-05-02
Add Apex import Guard for MM collection

suiyoubi:


ericharper (changes requested):
Please capture the error and include in the ImportError so users will know exactly what's not being imported properly.

suiyoubi:
> Please capture the error and include in the ImportError so users will know exactly what's not being imported properly.

ImportError is raised at the model level not layer level. Most of the cases here are using Apex's `groupnorm`, I added the ImportError when instantiate the model, not for each layer we invoke groupnorm.

@ericharper do you think this is sufficient ?

PR #9100, opened 2024-05-02
Fix #8891 by supported GPU-side batched CTC Greedy Decoding

### LINKED ISSUE #8891: Consider refactoring CTC greedy decoding
**Is your feature request related to a problem? Please describe.**

As part of my work on #8673 , I have been profiling to make sure we have low latency and high throughput. I have observed that this line can sometimes take https://github.com/NVIDIA/NeMo/blob/57444ae9b63afc5419a16948f7d032a1a9f7dc7f/nemo/collections/asr/parts/submodules/ctc_greedy_decoding.py#L218 several milliseconds. To be concrete, running at batch size 10, the model is taking 167 milliseconds, while decoding is taking 29 milliseconds, with almost all of the time spent on this max() operation. This is documented below.

image

It is not acceptable to have model inference be bottlenecked on non-GPU work like this.

I uploaded a gzipped version of the nsight systems file if you would like to view it.

[report25.nsys-rep.gz](https://github.com/NVIDIA/NeMo/files/14950389/report25.nsys-rep.gz)

The most obvious way to handle this situation is to do the "max()" operation on the GPU. I previously did a refactor that sped up CTC decoding by improving copy from GPU to CPU https://github.com/NVIDIA/NeMo/pull/8521, but it seems that is better to copy only after the max() operation is done. I did not do this in my previous PR, because (1) I didn't think it would be a concern and (2) that change is much more intrusive. _greedy_decode_logprobs is built with running on only a single element at a time rather than a batch. Maybe vmap can save us...

BTW, why is this only sometimes slow? I'm not certain, but it probably has to do with the CPU cache. It's working directly on pinned memory. The cache line may be getting evicted only sometimes?

galv:
# What does this PR do ?

I add a new flag "batched_inference" to the CTC greedy decoder.

**Collection**: ASR

# Changelog 
- Support batched inference of greedy CTC decoding.
- Add support for batched inference for label inputs as well.

# Usage
Here is an example turning this on for transcribe_speech.py

```
batch_size=32
amp=true

python examples/asr/speech_to_text_eval.py  pretrained_name=nvidia/parakeet-ctc-1.1b \ 
dataset_manifest=/home/dgalvez/scratch/data/test_other_sorted_downward.json  \
batch_size=$batch_size  output_filename=test_clean_decoded.jsonl  \
amp=$amp  amp_dtype=bfloat16  use_cer=false num_workers=1 \
return_hypotheses=false ctc_decoding.greedy.batched_inference=true
```

**Pre checks**:
- [X] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

# Additional Information
* Related to #8891

Here are the speedups in terms of throughput. First, I applied this diff to transcribe_speech.py in order to run multiple times:

```
modified   examples/asr/transcribe_speech.py                                                                                                                                                                
@@ -407,7 +407,33 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis                                                                                                       
                 override_cfg.augmentor = augmentor                                                                                                                                                         
                 override_cfg.text_field = cfg.gt_text_attr_name                                                                                                                                            
                 override_cfg.lang_field = cfg.gt_lang_attr_name                                                                                                                                            
-                transcriptions = asr_model.transcribe(audio=filepaths, override_config=override_cfg,)                                                                                                      
+                                                                                                                                                                                                           
+                transcriptions = None                                                                                                                                                                      
+                                                                                                                                                                                                           
+                torch.cuda.cudart().cudaProfilerStart()                                                                                                                                                    
+                torch.cuda.nvtx.range_push("GALVEZ_START")                                                                                                                                                 
+                for i in range(5):                                                                                                                                                                         
+                        # import nvtx                                                                                                                                                                      
+                        # pr = nvtx.Profile()                                                                                                                                                              
+                        # pr.enable()  # begin annotating function calls                                                                                                                                   
+                        # ctx = torch.autograd.profiler.emit_nvtx()                                                                                                                                        
+                        # ctx.__enter__()                                                                                                                                                                  
+                        # torch.cuda.cudart().cudaProfilerStart()                                                                                                                                          
+                        pass                                                                                                                                                                               
+                    import time                                                                                                                                                                            
+                    del transcriptions                                                                                                                                                                     
+                    print("RTFx=", 5.1 * 60 * 60 / (end_time - start_time))                                                                                                                                
+                        # pr.disable()                                                                                                                                                                     
+                        # ctx.__exit__(None, None, None)                                                                                                                                                   
+                        # torch.cuda.cudart().cudaProfilerStop()                                                                                                                                           
+                        pass                                                                                                                                                                               
+                torch.cuda.nvtx.range_pop() # "GALVEZ_START"                                                                                                                                               
+                torch.cuda.cudart().cudaProfilerStop()                                                                                                                                                     
                                                                                                                                                                                                            
     if cfg.dataset_manifest is not None:                                                                                                                                                                   
         logging.info(f"Finished transcribing from manifest file: {cfg.dataset_manifest}")                                                                                                                  
```

Then I ran with and without this option turned on:

Without batched inference:

```
set -euo pipefail

batch_size=32
amp=true

echo "GALVEZ: Conformer CTC"
python examples/asr/speech_to_text_eval.py  pretrained_name=nvidia/parakeet-ctc-1.1b dataset_manifest=/home/dgalvez/scratch/data/test_other_sorted_downward.json  batch_size=$batch_size  output_filename\
=test_clean_decoded.jsonl  amp=$amp  amp_dtype=bfloat16  use_cer=false num_workers=1 return_hypotheses=false ctc_decoding.greedy.batched_inference=false
RTFx= 1128.651983273993
RTFx= 1248.7464311388146
RTFx= 1327.39890551856
RTFx= 1344.5921792715505
RTFx= 1346.41703283495
[NeMo I 2024-05-02 15:07:14 speech_to_text_eval:210] Dataset WER/CER 3.76%/1.41%
```

With batched inference:

```
set -euo pipefail

batch_size=32
amp=true

echo "GALVEZ: Conformer CTC"
python examples/asr/speech_to_text_eval.py  pretrained_name=nvidia/parakeet-ctc-1.1b dataset_manifest=/home/dgalvez/scratch/data/test_other_sorted_downward.json  batch_size=$batch_size  output_filename=test_clean_decoded.jsonl  amp=$amp  amp_dtype=bfloat16  use_cer=false num_workers=1 return_hypotheses=false ctc_decoding.greedy.batched_inference=true
RTFx= 1157.725053779039
RTFx= 1265.9191489003545
RTFx= 1361.67639098725
RTFx= 1354.9981484942944
RTFx= 1350.2222749634843
[NeMo I 2024-05-02 15:25:17 speech_to_text_eval:210] Dataset WER/CER 3.76%/1.41%
```

You can see that RTFx throughput improvements are only modest. However, in terms of inference latency, there is a large speed up at the P99 latency levels, because the sometimes long calls to max() on the CPU no longer happen. We don't have a great way to measure this right now with NeMo's current tooling, so you will just have to believe me for now. See https://github.com/NVIDIA/NeMo/issues/8891 for details about this problem.

### DIFF
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -254,6 +254,7 @@ class AbstractCTCDecoding(ConfidenceMixin):
                 compute_timestamps=self.compute_timestamps,
                 preserve_frame_confidence=self.preserve_frame_confidence,
                 confidence_method_cfg=self.confidence_method_cfg,
+                batched_inference=self.cfg.greedy.get('batched_inference', True),
             )
 
         elif self.cfg.strategy == 'beam':
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -247,6 +247,8 @@ class AbstractCTCDecoding(ConfidenceMixin):
             self.compute_timestamps |= self.preserve_frame_confidence
 
         if self.cfg.strategy == 'greedy':
+            logging.warning("CTC decoding strategy 'greedy' is slower than 'greedy_batched', which implements the same exact interface. Consider changing your strategy to 'greedy_batched' for a free performance improvement.",
+                            mode=logging_mode.ONCE)
 
             self.decoding = ctc_greedy_decoding.GreedyCTCInfer(
                 blank_id=self.blank_id,
@@ -1296,7 +1298,7 @@ class CTCBPEDecoding(AbstractCTCDecoding):
 
 @dataclass
 class CTCDecodingConfig:
-    strategy: str = "greedy"
+    strategy: str = "greedy_batched"
 
     # preserve decoding alignments
     preserve_alignments: Optional[bool] = None
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -213,7 +213,7 @@ class AbstractCTCDecoding(ConfidenceMixin):
         self.batch_dim_index = self.cfg.get('batch_dim_index', 0)
         self.word_seperator = self.cfg.get('word_seperator', ' ')
 
-        possible_strategies = ['greedy', 'beam', 'pyctcdecode', 'flashlight']
+        possible_strategies = ['greedy', 'greedy_vectorized', 'beam', 'pyctcdecode', 'flashlight']
         if self.cfg.strategy not in possible_strategies:
             raise ValueError(f"Decoding strategy must be one of {possible_strategies}. Given {self.cfg.strategy}")
 
@@ -254,7 +254,15 @@ class AbstractCTCDecoding(ConfidenceMixin):
                 compute_timestamps=self.compute_timestamps,
                 preserve_frame_confidence=self.preserve_frame_confidence,
                 confidence_method_cfg=self.confidence_method_cfg,
-                batched_inference=self.cfg.greedy.get('batched_inference', False),
+            )
+
+        elif self.cfg.strategy == "greedy_vectorized":
+            self.decoding = ctc_greedy_decoding.GreedyVectorizedCTCInfer(
+                blank_id=self.blank_id,
+                preserve_alignments=self.preserve_alignments,
+                compute_timestamps=self.compute_timestamps,
+                preserve_frame_confidence=self.preserve_frame_confidence,
+                confidence_method_cfg=self.confidence_method_cfg,
             )
 
         elif self.cfg.strategy == 'beam':
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -254,7 +254,7 @@ class AbstractCTCDecoding(ConfidenceMixin):
                 compute_timestamps=self.compute_timestamps,
                 preserve_frame_confidence=self.preserve_frame_confidence,
                 confidence_method_cfg=self.confidence_method_cfg,
-                batched_inference=self.cfg.greedy.get('batched_inference', True),
+                batched_inference=self.cfg.greedy.get('batched_inference', False),
             )
 
         elif self.cfg.strategy == 'beam':
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -247,8 +247,10 @@ class AbstractCTCDecoding(ConfidenceMixin):
             self.compute_timestamps |= self.preserve_frame_confidence
 
         if self.cfg.strategy == 'greedy':
-            logging.warning("CTC decoding strategy 'greedy' is slower than 'greedy_batched', which implements the same exact interface. Consider changing your strategy to 'greedy_batched' for a free performance improvement.",
-                            mode=logging_mode.ONCE)
+            logging.warning(
+                "CTC decoding strategy 'greedy' is slower than 'greedy_batched', which implements the same exact interface. Consider changing your strategy to 'greedy_batched' for a free performance improvement.",
+                mode=logging_mode.ONCE,
+            )
 
             self.decoding = ctc_greedy_decoding.GreedyCTCInfer(
                 blank_id=self.blank_id,
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -219,14 +219,14 @@ class AbstractCTCDecoding(ConfidenceMixin):
 
         # Update preserve alignments
         if self.preserve_alignments is None:
-            if self.cfg.strategy in ['greedy']:
+            if self.cfg.strategy in ['greedy', 'greedy_batched']:
                 self.preserve_alignments = self.cfg.greedy.get('preserve_alignments', False)
             else:
                 self.preserve_alignments = self.cfg.beam.get('preserve_alignments', False)
 
         # Update compute timestamps
         if self.compute_timestamps is None:
-            if self.cfg.strategy in ['greedy']:
+            if self.cfg.strategy in ['greedy', 'greedy_batched']:
                 self.compute_timestamps = self.cfg.greedy.get('compute_timestamps', False)
             elif self.cfg.strategy in ['beam']:
                 self.compute_timestamps = self.cfg.beam.get('compute_timestamps', False)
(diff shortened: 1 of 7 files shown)


galv:
To verify that latency does indeed fall, I ran the above prefixed by `nsys profile --env-var=NSYS_NVTX_PROFILER_REGISTER_ONLY=0 -t nvtx -c nvtx -p GALVEZ_START` with batch_inference=false and batch_inference=true. I added an nvtx range just for the decoder.



batch_inference=true case:
```
Time	Total Time	Instances	Avg	Med	Min	Max	StdDev	Style	Range
93.9%	71.801 s	1	71.801 s	71.801 s	71.801 s	71.801 s	0 ns	PushPop	GALVEZ_START
6.1%	4.683 s	460	10.180 ms	1.173 ms	796.420 μs	117.883 ms	19.312 ms	PushPop	decoder
```

batched_inference=false case:
```
Time	Total Time	Instances	Avg	Med	Min	Max	StdDev	Style	Range
86.7%	79.410 s	1	79.410 s	79.410 s	79.410 s	79.410 s	0 ns	PushPop	GALVEZ_START
13.3%	12.217 s	460	26.558 ms	3.800 ms	2.125 ms	171.248 ms	38.294 ms	PushPop	decoder
```

You can see that the decoder takes half the time it used to, and that its maximum runtime is 118ms, down from 171 ms.

nemo/collections/asr/parts/submodules/ctc_greedy_decoding.py
```diff
@@ -195,6 +209,7 @@ def forward(
 
             for ind in range(prediction_cpu_tensor.shape[0]):
                 out_len = decoder_lengths[ind] if decoder_lengths is not None else None
+                # Gross, why are we doing this on CPU one at a time?
```
titu1994, line 212: Remove

nemo/collections/asr/parts/submodules/ctc_greedy_decoding.py
```diff
@@ -203,6 +218,94 @@ def forward(
...
         return (packed_result,)
 
+    @torch.no_grad()
+    def _greedy_decode_logprobs_batched(self, x: torch.Tensor, out_len: torch.Tensor):
+        # x: [B, T, D]
+        # out_len: [B]
+
+        batch_size = x.shape[0]
+
+        predictions = x
```
titu1994, line 228: Pleasr add detailed comments inline per stage of operation similar to how CTC and RNNT decoding algo is documented.

nemo/collections/asr/parts/submodules/ctc_greedy_decoding.py
```diff
@@ -203,6 +218,94 @@ def forward(
...
+            if self.preserve_frame_confidence:
+                hypothesis.frame_confidence = self._get_confidence(predictions[i, : out_len[i], :])
+
+            hypotheses.append(hypothesis)
+
+        return hypotheses
+
+    @torch.no_grad()
+    def _greedy_decode_labels_batched(self, x: torch.Tensor, out_len: torch.Tensor):
+        # x: [B, T]
```
titu1994, line 272: Add docstring + inline comments for every section of code being executed to be perfectly clear what's happening at every stage

nemo/collections/asr/parts/submodules/ctc_decoding.py
```diff
@@ -254,6 +254,7 @@ def __init__(self, decoding_cfg, blank_id: int):
                 compute_timestamps=self.compute_timestamps,
                 preserve_frame_confidence=self.preserve_frame_confidence,
                 confidence_method_cfg=self.confidence_method_cfg,
+                batched_inference=self.cfg.greedy.get('batched_inference', False),
```
titu1994, line 257: Argument name is too generic, CTC inference code runs in batch mode (just not fully batched mode).

This can just be a different strategy altogether honestly so instead of "greedy" and "pyctcdecode" etc we can just call this "greedy_batch" algo similar to RNNT ?

If not, can you assign some other name than "batched_inference"

titu1994 (commented):
The new algorithm part looks great, but shouldn't be fused into CTC decoding but he treated as a strategy similar to how RNNT does it. 

@nithinraok upto you if you want to fuse the two together like this with a flag or use rnnt strategy pattern here in CTC

nemo/collections/asr/parts/submodules/ctc_decoding.py
galv: Making a new strategy seems like the right call. I'm called it "greedy_vectorized".

nemo/collections/asr/parts/submodules/ctc_greedy_decoding.py
galv: Just a note, _greedy_decode_labels is an implementation detail that was added only as a hack for initial streaming inference. For some reason, streaming inference did an argmax on its own inside of conformer_stream_step, so _greedy_decode_labels was added to accommodate that. But it is short-sighted because other decoders like RNN-T greedy search, RNN-T beam search, and CTC beam search cannot handle such an input. I don't think this code has much life left in this world.

nemo/collections/asr/parts/submodules/ctc_decoding.py
```diff
@@ -213,7 +213,7 @@ def __init__(self, decoding_cfg, blank_id: int):
         self.batch_dim_index = self.cfg.get('batch_dim_index', 0)
         self.word_seperator = self.cfg.get('word_seperator', ' ')
 
-        possible_strategies = ['greedy', 'beam', 'pyctcdecode', 'flashlight']
+        possible_strategies = ['greedy', 'greedy_vectorized', 'beam', 'pyctcdecode', 'flashlight']
```
titu1994, line 216: Hmm can't we call it greedy_batched similar to RNNT? I understand both are technically batched, and vectorized is more appropriate, however conformity is nice for both us and users

titu1994 (dismissed):
Looks far cleaner than before, minor comments but overall good enough to merge. 

1) Can we make this the default algo ? If users want inference features that don't work in this mode they can switch to slower code path. NeMo 2.0 is the time for switching defaults if we want to. 

2) can we name this greedy_batch similar to RNNT

nemo/collections/asr/parts/submodules/ctc_decoding.py
galv: Good observation. I changed to "greedy_batched", and generally speaking changed "vectorized" to "batched" based on your observation about conformity.

galv:
Will need to get back to you about what it takes to make this the default algo. But the main concern is that many existing ".nemo" files will have the "strategy" field of the relevant config set to "greedy". It's not really clear to me how we can convert those users using pretrained models to use the "greedy_batched" strategy. For what it's worth, my tests check that the interfaces are effectively the same, so I could just use GreedyBatchedCTCInfer whenever "greedy" strategy is specified, unless that distresses you.

pablo-garay:
I see CICD pipeline passed: https://github.com/NVIDIA/NeMo/actions/runs/8976148083
Let me know if you need/ready to merge

titu1994:
Hmm I see your point. Lets do this - 
1) Set default in config to greedy_batch
2) Add logging message to greedy (onetime log) that informs users that greedy_batch strategy is faster. 

I'm not average to replacing "greedy" with greedy_batch, but then we need a different flag name to allow users to fall back to older codepath. Maybe we can call it "greedy_legacy" then?

galv:
@titu1994 done.

titu1994 (approved):
Looks great ! Thanks for the refactor !

PR #9101, opened 2024-05-02
Add llama3 and distributed checkpoint support in NeVA

yaoyu-33:
# What does this PR do ?

Add llama3 and distributed checkpoint support in NeVA

**Collection**: [multimodal]

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/multimodal/multimodal_llm/neva/eval/vqa_science.py
```diff
@@ -169,7 +169,7 @@ def eval_model(args):
     parser.add_argument("--image-folder", type=str, default="")
     parser.add_argument("--question-file", type=str, default="tables/question.json")
     parser.add_argument("--answers-file", type=str, default="answer.jsonl")
-    parser.add_argument("--conv-mode", type=str, default="llava_v0")
+    parser.add_argument("--conv-mode", type=str, default="llava_v0")  # this flag has no use!
```
cuichenx, line 172: then should we get rid of it..?

nemo/collections/nlp/modules/common/text_generation_strategy.py
```diff
@@ -487,6 +544,7 @@ def __init__(self, model):
             is_multimodal=self.data_cfg.is_multimodal,
             sep_image_conv_front=self.data_cfg.sep_image_conv_front,
             conv_template=self.data_cfg.get("conv_template", "nvgpt"),
+            model_type=self.cfg.mm_cfg.llm.get("model_type", "nvgpt"),
```
cuichenx, line 547: should we rename this to nemotron?

yaoyu-33: that will break many old checkpoint, we might just keep this way for a while...

nemo/utils/callbacks/dist_ckpt_io.py
```diff
@@ -75,19 +80,24 @@ def load_checkpoint(
         else:
             sharded_strategy = None
 
+        if not strict:
+            for key in list(sharded_state_dict['state_dict'].keys()):
```
mikolajblaz, line 84: I assume this is a temporary implementation for a strict flag because we also need to notify the user which keys are skipped.

Also, this will work only with Zarr ckpt format, for PyT Distributed it will be different

nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py
```diff
@@ -235,6 +240,22 @@ def replace_media_embeddings(self, input_ids, inputs_embeds, media):
 
         return updated_input_embeds
 
+    def sharded_state_dict(self, prefix: str = '', sharded_offsets: tuple = (), **kwargs):
+        sharded_state_dict = super().sharded_state_dict(prefix=prefix, sharded_offsets=sharded_offsets, **kwargs)
+
+        state_dict = self.state_dict(prefix='', keep_vars=True)
+        state_dict.pop('weight')
```
mikolajblaz, line 247: Is `weight` not needed at all?

nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py
```diff
@@ -235,6 +240,22 @@ def replace_media_embeddings(self, input_ids, inputs_embeds, media):
...
+        # duplicate everything else
+        for layer_name in state_dict.keys():
+            tensor = state_dict[layer_name]
+            layer_key = f'{prefix}{layer_name}'
+            sharded_state_dict[layer_key] = make_sharded_tensor_for_checkpoint(
+                tensor,
+                layer_key,
+                prepend_offsets=sharded_offsets,
+            )
```
mikolajblaz, line 256: I think this can be replaced with 
```
from megatron.core.transformer.utils import make_sharded_tensors_for_checkpoint
     ...
     make_sharded_tensors_for_checkpoint(state_dict)
```

nemo/utils/callbacks/dist_ckpt_io.py
mikolajblaz: Also, I don't think it's correct, because we should use the sharded key, not the state dict key (also doesn't account for nested dicts).
Can you try something like this (I didn't run this code so there might be errors)?
This should work for all backends, nested dicts and use correct keys.
```

from megatron.core.dist_checkpointing.dict_utils import extract_matching_values
from megatron.core.dist_checkpointing.mapping import ShardedBase


    if not strict:
        sharded_state_dict = self.adjust_non_strict_load(path, sharded_state_dict)
    ...

    def adjust_non_strict_load(self, path: _PATH, sharded_state_dict: Dict[str, Any]):
        ckpt_sharded_metadata = dist_checkpointing.load_tensors_metadata(path)
        loaded_keys = []
        missing_keys = []
        unexpected_keys = []

        def should_remove_missing_sharded_base(x: Any):
            if isinstance(x, ShardedBase):
                if x.key in ckpt_sharded_metadata:
                    loaded_keys.append(x.key)
                    return False
                else:
                    unexpected_keys.append(x.key)
                    return True
            return False

        _, sharded_state_dict = extract_matching_values(sharded_state_dict, should_remove_missing_sharded_base)
        logging.info(f'The following keys are not in the checkpoint and will not be loaded: {unexpected_keys}')

        # TODO: compute missing_keys by:
        #  1. all_gather_object of loaded_keys
        #  2. missing_keys = ckpt_sharded_metadata.keys() - loaded_keys
        return sharded_state_dict

```

cuichenx (approved):

Issue #8235, opened 2024-01-24
OOM killed when diarizing 4 hours recording


tttalshaul:
I'm trying nemo_toolkit 1.22.0 on T4 (14G vRAM) with 40G cpu RAM to diarize with ClusteringDiarizer. 
After completing speaker embeddings, tqdm shows "clustering: 0%" and then my script is getting killed by OS with error 137 OOM (out of memory) because of RAM consumption greater then 40G..

Installed nemo using pip install nemo_toolkit==1.22.0 (python 3.10 x64) on premise K8S.

I'm using diar_infer_meeting.yaml with max_num_speakers=20, chunk_cluster_count=50,
Tried to lower embeddings_per_chunk from default 10000 to 500 but it didn't help.

nithinraok:
@tttalshaul can you share more info on length of audio and size of embeddings in pkl file you saved?

tttalshaul:
Audio length is 3:43 hours after cutting silence (using VAD). 
embeddings:
scale0 6.6M len 8920
scale1 7.9M len 10704
scale2 9.9M len 13385
scale3 14M len 17846
scale4 20M len 26772
scale5 40M len 53548

tttalshaul:
@nithinraok 
What tests can I make to understand the cause of this problem?

I'm moving ClusteringDiarizer(cfg=cfg).to(device), and I see that device is cuda:0. 
I understand that this only effects vad and embeddings extraction, but clustering occurs in cpu RAM anyway?

Thank you

nithinraok:
@tango4j what is the maximum audio duration we currently support for single pass offline diarzation

tango4j:
@tttalshaul 

Sorry for the late reply.
If you have visible cuda devices, both embedding extractions and clustering will be happening on GPU RAM.

```
clustering: 
    parameters:
....
      chunk_cluster_count: 50 # Number of forced clusters (overclustering) per unit chunk in long-form audio clustering.
      embeddings_per_chunk: 10000 # Number of embeddings in each chunk for long-form audio clustering. Adjust based on GPU memory capacity. (default: 10000, approximately 40 mins of audio) 
```
You should **increase** the `embeddings_per_chunk` value to consume less amount of memory.
Also note that even if you use long form audio clustering, there is a limit of audio length.
Thus, you need to make a compromise between performance and length. 
Increasing `embeddings_per_chunk` will consume less memory but could lead to deteriorated performance.

tttalshaul:
[...]

PR #9114, opened 2024-05-04
Fixed chokepoint in diarization for longer audios

AatikaNazneen:
# What does this PR do ?

Even after addition of long form clustering feature, the code was failing because get_argmin_mat acted as a chokepoint and consumed too much memory for a mathematical operation that was not that complex. The tiling function ended up creating n^2 size matrices. The current function gives a drastic change in memory consumption with minimal change in running time.
Fixes the problem described in [OOM Clustering Bug]( https://github.com/NVIDIA/NeMo/issues/8235 )
**Collection**: asr

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


AatikaNazneen:
I will modify the code so that all the test cases pass soon.

tango4j:
Thank you for suggesting this idea.
Currently I don't have bandwidth to test this so let me check on this as soon as I have bandwidth to take care of this.  
This needs to be tested with long-form and short-form audio clustering.

AatikaNazneen:
@tango4j Reminder to please review it

tango4j:
Hi. @AatikaNazneen Can you provide 4 hour long audio data for testing that is creating the OOM?
Also, there is a limit in the audio length size. Please be aware that the NeMo speaker diarization cannot handle infinite length audio. Even if it is long form clustering method, the length of audio is capped by GPU memory.

tttalshaul:
haven't tested this file but an example for 4 hours podcast that can used for testing..
https://www.youtube.com/watch?v=HoRMD9NtKYc

AatikaNazneen:
I used this one: https://www.youtube.com/watch?v=0OhTQnDp1yw
I was able to do this diarize this on sagemaker g4 instance which has 16 GB GPU RAM and 16 GB CPU RAM after this change.

tango4j:
@AatikaNazneen 
I was not able to find a way to download the wav file from your youtube link. Could you please extract wav file and upload it through Googledrive or dropbox account?
Also, reviewing this PR would take some time, since we need to test with other scenarios. 
Please provide us with the long form audio files you can share, then it will expedite the process to test your PR.

AatikaNazneen:
You can use [this ](https://traffic.libsyn.com/secure/schoolofmotion/School_of_Motion_EOY_Podcast-AUDIO-20221228.mp3?dest-id=485169) audio or any audio. The main purpose is that the long form clustering was taking up more RAM previously and that has been optimized so it can be tested of any long audio but you would have to limit your resources to test the before and after effect.

AatikaNazneen:
@tango4j Any updates?

AatikaNazneen:
@tango4j @nithinraok

AatikaNazneen:
@tango4j @nithinraok Reminder to review it please

tango4j:
@AatikaNazneen 
Sorry for the frustration. Let me review this today.

tango4j:
Hi, @AatikaNazneen.

I have run the original main version and your for-loop version.
First of all, your analysis that `torch.tile` can be taking excessive memory is True with.
However, we can let the `get_argmin_mat` by using `torch.expand`

```python
In [1]: import torch

In [2]: vec = torch.rand(1000, 1).T.to("cuda:0") # 556MB 

In [3]: vec_tile =  vec.tile(10000000,1) # 38704MB, 38148 MB was added 

In [4]: vec_expand =  vec.expand(10000000, 1000) # 0MB was added
``` 
Now, if we change the `get_argmin_mat` line to these:
```python
curr_mat = curr_scale_anchor.expand(base_scale_anchor.shape[0], -1)
base_mat = base_scale_anchor.expand(curr_scale_anchor.shape[0], -1).t()
 ```
 These are the time taken for three 10 min audio files.
 
> torch.expand: 0.0266 seconds
torch.tile: 0.0568 seconds
For Loop: 0.1970 seconds

Bottom line is, we need to modify this to use expand, instead of adding for loop.
For loop dramatically increases the Time complexity from O(S) to O(NS) where S is number of scales and N is the frame length in base-scale.

Let me push the `torch.expand` version of code and please check if there is still memory issue.
As you can see in the torch example I showed above, there should be no memory choke point with `torch.expand`.

tttalshaul:
@AatikaNazneen @tango4j are more checks waiting to be reviewed?

AatikaNazneen:
@tango4j Your modification did not do the work for me and i still get OOM error.
The reason is that torch.expand doesnt increase memory but when we do the subtraction operation, it occupies the double the memory for both the vectors. The for loop reduces the memory consumption by breaking the problem into a smaller chunk so the peak memory occupied is much lesser as compared to without the for loop

AatikaNazneen:
@tango4j If you're convinced then please revert it back to the for loop version. and the rest of the labels needed to run all the checks as well.

AatikaNazneen:
The time complexity is theoretically the same but using a for loop, the previous operation memory is freed (hence the increase in time) but it significantly reduces the space complexity. 
Using the expand function, we need  2 (n^2) sizes at a particular memory and using the for loop we use n * (2n) memory but at a particular time, we only need 2n memory so we dont get OOM.
Let me know if you need further clarification

tango4j:
@AatikaNazneen 
I still do not get the idea of for loop and expand function getting the same complexity.
Both in theory and experiments, for loop is increasing the running time.
It is array assignments versus sequential process so for loop is obviously takes more time.
and then if you implement it, it also shows slower speed.

I missed the part where expand-approach causes memory consumption when it performs subtractions.

So in short, 
I think we should create an option to use forloop to handle extremely long audio.
We cannot increase the time complexity to xN steps just for rarely handled extremely long audio files.

AatikaNazneen:
@tango4j We can go for that implementation when we use long-form-clustering pipeline.
> I missed the part where expand-approach causes memory consumption when it performs subtractions.
If you take a look at line 13 in the image I shared, the memory increment is twice the amount increment in line 9
Also, I tested the 5 hour audio of sagemaker with mlg4dn and it was failing with the expand solution

tango4j:
@AatikaNazneen 
I can see that line 13 in your screenshot shows the additional memory usage during the subtraction operation of the copied tensors. I neglected this part when I was testing.
This is why we are not saving much on memory peak usage compared to "tile" method.

So at the end of the day, I believe this is all about memory-time trade off.
I think, as you suggested, we can let the system use "for loop" for long form audio clustering cases. That way we can only value memory efficiency for long form audio clustering and reduce the peak memory usage. 

I will add this option shortly to long form audio clustering path.
- standard: expand (speed oriented)
- long form audio: for loop (memory oriented)

tttalshaul:
When do you think all of this will be completed? 
Thanks..
@tango4j @nithinraok

tango4j:
@AatikaNazneen @tttalshaul 
Sorry that I have been traveling overseas for conference. 
Will work on this PR starting today and finalize this week.

tttalshaul:
reminding 🙏🏿

tango4j:
@AatikaNazneen @tttalshaul 
I pushed the change to the @AatikaNazneen 's Repo.
This version is using expand for less than 10K embedding frames then use for-loop for more than 10K embeddings.

tttalshaul:
@tango4j Thank you.
Would it be an official option in main NeMo repo/pip install?

AatikaNazneen:
@tango4j Can you please finalize this?

nithinraok:
Removing stale, @tango4j fyi

Arno-Riverside:
@tango4j @AatikaNazneen can we reopen this ? The PR has unfortunately not been merged and this fix is needed

AatikaNazneen:
@tango4j @nithinraok

guilhermehge:
Hi, I'm here to attest that this PR should be merged into the NeMo project. I have built a transcription pipeline that used NeMo for speaker diarization and I was having problems when diarizing 2h, 3h or even 4h long áudios. I found this PR, changed the source code within my container and it started working, I managed to diarize a 4h10min long audio.

I'm using NeMo 2.0.0 and I changed the following files:
* online_diarizer.py
* longform_clustering.py
* offline_clustering.py

to the code of this PR, and it worked.

@tango4j, please take a look at this PR.

guilhermehge:
> @guilhermehge
> 
> Hi. We did not have bandwidth while we are working on Sortformer Diarizer. https://huggingface.co/nvidia/diar_sortformer_4spk-v1
> 
> We will be releasing long-form audio diarization with streaming sortformer also. I will consider merging this to NeMo main then.
> 
> The clustering diarizer based diarization will be deprecated in the future.

Thank you for your answer. I'll take a look at this project and perhaps apply it. The pipeline I've built, transcribes mainly call center calls, that's why NVIDIA NeMo is the best for me, since it was trained on telephone calls, but I'll give this one a try.

Cheers!

PR #9128, opened 2024-05-07
Revert "Add assert for max_steps to be positive in MegatronGPTSFTMode…

pablo-garay:
…l (#9110)"

This reverts commit 00023c42a8364a1468a1bb31c1f8f35bc7b0c11e.


ericharper (approved):
We can revert this since the check is already happening at the build data set time.

PR #9131, opened 2024-05-07
[do-not-merge] Illustration how to run audio+text dataloading

pzelasko:
# What does this PR do ?

The key config options to set are:

```
 +model.train_ds.input_cfg=/path/to/speech_text_input.yaml \
 +model.train_ds.use_multimodal_sampling=true \
 +model.train_ds.batch_tokens=65536 \
 +model.train_ds.token_equivalent_duration=0.08 \
 +model.train_ds.quadratic_factor=4096 \
 +model.train_ds.tarred_audio_filepaths=null \
 model.train_ds.manifest_filepath=null \
 model.train_ds.batch_size=null \
 model.train_ds.batch_duration=null \
 model.train_ds.use_lhotse=true \
 model.train_ds.use_bucketing=true \
 model.train_ds.num_buckets=30 \
```

The values of batch_tokens, token_equivalent_duration, and quadratic_factor need tuning.

where contents of `speech_text_input.yaml` are:

```yaml
 - type: group
    weight: 1.0
    input_cfg:
    - type: nemo_tarred
      manifest_filepath: /path/to/manifest__OP_0..127_CL_.json
      tarred_audio_filepaths: /path/to/audio__OP_0..127_CL_.tar
      tags:
        pnc: 'no'
        source_lang: en
        target_lang: en
        taskname: asr
      weight: 1.0
    - type: nemo_sft_jsonl
      paths: /data/nemo_megatron_sft_chat_qa_data.shuf.jsonl
      language: en
      weight: 1.0
```

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/asr/data/audio_to_text_lhotse_prompted.py
+++ b/nemo/collections/asr/data/audio_to_text_lhotse_prompted.py
@@ -16,10 +16,11 @@ from typing import Callable, Sequence
 import omegaconf
 import torch.utils.data
 from lhotse import CutSet
-from lhotse.cut import MixedCut, MonoCut
+from lhotse.cut import Cut, MixedCut, MonoCut
 from lhotse.dataset import AudioSamples
 from lhotse.dataset.collation import collate_vectors
 
+from nemo.collections.common.data.lhotse.text_adapters import NeMoSFTExample
 from nemo.collections.common.tokenizers import CanaryTokenizer, TokenizerSpec
 from nemo.collections.common.tokenizers.aggregate_tokenizer import TokenizerWrapper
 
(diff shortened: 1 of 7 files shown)


nemo/collections/common/data/lhotse/text_adapters.py
```diff
@@ -95,3 +116,416 @@ def __iter__(self) -> Iterator[TextPairExample]:
...
+
+        The tokenization is compatible with Megatron's :class:`GPTSFTChatDataset`.
+        """
+        special_tokens = {
+            "system_turn_start": "<extra_id_0>",
+            "turn_start": "<extra_id_1>",
+            "label_start": "<extra_id_2>",
+            "end_of_turn": "\n",
+            "end_of_name": "\n",
+        }
```
pzelasko, line 144: The special tokens are tricky: perhaps we need to ensure that they are included in the tokenizer's vocab.

nemo/collections/asr/data/audio_to_text_lhotse_prompted.py
```diff
@@ -54,24 +55,51 @@ def __init__(
...
+                context_ids=collate_vectors([e.context_ids for e in text_examples], padding_value=self.padding_value),
+                masks=collate_vectors([e.mask for e in text_examples], padding_value=self.padding_value),
+            )
         else:
-            prompt_tokens = None
-            prompt_token_lens = None
+            text_minibatch = {k: None for k in "input_ids answer_ids context_ids masks".split()}
 
-        return audio, audio_lens, tokens, token_lens, prompt_tokens, prompt_token_lens
+        return audio, audio_lens, tokens, token_lens, prompt_tokens, prompt_token_lens, text_minibatch
```
zhehuaichen, line 102: copy the discussion note here and some followups in the end

```
supports 3 different strategies of mixing speech and text data
1.current strategy (sampled together as if it's one modality by projecting duration to equivalent token counts)
2.zip sampler strategy (separate sampler for each modality with separately defined batch_duration and batch_tokens, both samplers zipped to yield a mini-batch together)
3.round-robin strategy (separate samplers as in zip but the mini-batches are alternately yielded from each)
```
do we expect all the three strategies will return audio and text batch separately  to the model like here?
if so, what do you think about also grouping the text-audio batch data in one dict like text-only batch (and in future, add audio-audio batch)

no need to do these in this PR but can be done on top of the canary-llm branch which has a speech SFT batch compatible with mtron

pzelasko: It's open for discussion. For the POC I separated text-only and audio-text portions of the mini-batch because it was easy to re-use the existing code and extend it with text-only branches. We could explore "fusing" those but we should design how the fused mini-batch would look like (e.g. do we introduce zero inputs in audio tensor for text-only examples? there will be a number of decisions to make here).

pzelasko: Initially, I wanted to return sth like:

```python
@dataclass
class MultimodalQABatch:
    audio_text: dict  # dict might become a nested AudioTextQABatch dataclass, similar for other fields
    text_text: dict
    # ... could be extend with sub fields
```

But GPT SFT model code that we inherit from expects batch to be a dict and explicitly scans it for some fields like "tokens", also discards every key that has a non-tensor value. If we want to proceed with separate dataclasses or even dicts, we would need to override some fwd/bwd functions in the parent class.

zhehuaichen: > It's open for discussion. For the POC I separated text-only and audio-text portions of the mini-batch because it was easy to re-use the existing code and extend it with text-only branches. We could explore "fusing" those but we should design how the fused mini-batch would look like (e.g. do we introduce zero inputs in audio tensor for text-only examples? there will be a number of decisions to make here).

I actually like this design. we can leave how to "fuse"  to the model code to decide, e.g. model code may want to concat speech and text batch together and do one fprop instead of two

PR #9041, opened 2024-04-25
GPU-based vectorized `SpecAug`

pzelasko:
# What does this PR do ?

Context: together with @galv we found feature normalization and specaug to take approx 30% of the total forward step time in Canary training, due to CPU-bottlenecked implementations. Feature normalization is addressed in #8964. 

This PR adds a GPU based SpecAugment. The original implementation loops over every example and every mask, then waits on CPU RNG to sample the numbers. The new fast implementation still applies masks sequentially, but is vectorized on batch size and uses GPU's RNG. We found approx. 5x speedup (70ms -> 17ms in profiling, but both numbers include profiler overhead). I also added a flag to be able to revert to the old implementation in case anybody encounters a compatibility issue.

I validated visually that the new impl behavior is as expected. 

Old:
image

New:
image

**Collection**: ASR

# Changelog 
- 5x faster SpecAugment to reduce typical forward step time by ~10%.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# Jenkins CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

There's no need to comment `jenkins` on the PR to trigger Jenkins CI.
The GitHub Actions CI will run automatically when the PR is opened.
To run CI on an untrusted fork, a NeMo user with write access must click "Approve and run".

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #9155, opened 2024-05-09
GPU-based vectorized Specaug Version 2

amorari-nvidia:
# What does this PR do ?

This PR proposes a faster version of the Spectrogram Augmentation module for time and frequency masking.


pzelasko (dismissed):
I visually inspected that the masks look correct on random spectrograms. I also ran a training with this variant in Nsight profiler and noticed the specaug computation mostly disappeared from the profile. It is estimated to be about 4-5x faster than the variant proposed in #9041 which would make it 8-10x faster than the original version (still available as the "legacy" setting). 

LGTM!

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -56,7 +60,14 @@ def output_types(self):
...
     def __init__(
-        self, freq_masks=0, time_masks=0, freq_width=10, time_width=10, rng=None, mask_value=0.0,
+        self,
+        freq_masks: int = 0,
+        time_masks: int = 0,
+        freq_width: int = 10,
+        time_width: int | float = 10,
+        rng: random.Random | None = None,
+        mask_value: float = 0.0,
+        fast: bool = True,
```
titu1994, line 70: Give flag a proper name

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -112,6 +130,82 @@ def forward(self, input_spec, length):
...
+                dtype=torch.long,
+            )
+            mask_width = torch.randint(
+                low=0, high=max(1, width), size=(batch_size, num_masks), device=input_spec.device, dtype=torch.long
+            )
+            mask_end = mask_start + mask_width
+
+        mask_end = mask_start + mask_width
+
+        indices = torch.arange(axis_length, device=input_spec.device)
```
titu1994, line 196: Can you inline document what's being done

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -112,6 +130,82 @@ def forward(self, input_spec, length):
...
+        mask_tensor = (indices >= mask_start.unsqueeze(-1)) & (indices < mask_end.unsqueeze(-1))
+        mask_tensor = mask_tensor.any(dim=1)
+
+        mask = torch.zeros_like(input_spec, dtype=torch.bool)
+        if axis == 2:
+            mask_ranges = mask_tensor[:, None, :]
```
titu1994, line 202: Raise an error for any other value of axis

titu1994 (commented):
Looks great ! Minor comments

nemo/collections/asr/parts/submodules/spectr_augment.py
pzelasko: ... vectorized?

amorari-nvidia: I changed it and it gives error on a test. Doing _grep -R rng_ inside the nemo/ folder shows that the _rng_ variable is used everywhere to mean random generator, so I think it is better to leave as it is for consistency with the rest of the project.

amorari-nvidia: Never mind, I realized that I was looking at the wrong flag.

amorari-nvidia: use_vectorized_version ?

nemo/collections/asr/parts/submodules/spectr_augment.py
amorari-nvidia: ok

nemo/collections/asr/parts/submodules/spectr_augment.py
amorari-nvidia: ok

nemo/collections/asr/parts/submodules/spectr_augment.py
amorari-nvidia: or use_vectorized_code

titu1994: I like use_vectorized_code

amorari-nvidia:
After this I think that we may also have to add the flag in the audio processing module:
https://github.com/amorari-nvidia/NeMo/blob/d26eab45d54e9e00e74c68ea4b0bbbf7b48a7a50/nemo/collections/asr/modules/audio_preprocessing.py#L494

And also we may have to add a test, adding the pytest run_only_on('GPU') flag the same it is done for the numba code:
#https://github.com/amorari-nvidia/NeMo/blob/d26eab45d54e9e00e74c68ea4b0bbbf7b48a7a50/tests/collections/asr/test_asr_modules.py#L71

nemo/collections/asr/parts/submodules/spectr_augment.py
amorari-nvidia: ok. I used that one.

nemo/collections/asr/parts/submodules/spectr_augment.py
```diff
@@ -142,40 +142,43 @@ def _forward_fast(self, input_spec: torch.Tensor, length: torch.Tensor) -> torch
...
+                width = torch.clamp(width * length, max = axis_length).unsqueeze(1)
 
+        # Generate [0-1) random numbers and then scale the tensors.
+        # Use float32 dtype for begin/end mask markers before they are quantized to long.
+        # Using x.dtype might cause us to encounter dtypes such as bf16 or smaller which
+        # wouldn't be able to represent every frame index leading to subtle bugs.
+        mask_width = torch.rand((batch_size, num_masks), device=input_spec.device, dtype=torch.float32) * width
+        mask_width = mask_width.long()
+        mask_start = torch.rand((batch_size, num_masks), device=input_spec.device, dtype=torch.float32) * (axis_length - mask_width)
+        mask_start.long()
```
pzelasko, line 180: should we assign the result? `mask_start = mask_start.long()`

pzelasko: for reference the version that I tested and confirmed that it masks properly (ie skipping the padding regions) had the following code in this bit:

```python
if isinstance(width, float):
            width = torch.min(width * length, torch.tensor(axis_length, dtype=torch.float32)).unsqueeze(1)
        mask_width = torch.rand((batch_size, num_masks), device=input_spec.device, dtype=torch.float32) * width
        mask_width = mask_width.long()
        mask_start = torch.rand((batch_size, num_masks), device=input_spec.device, dtype=torch.float32) * (length.unsqueeze(1) - mask_width)
        mask_start = mask_start.long()
```

I see you are using axis_length rather than length for mask_start so I'm suspecting this version has the padding mask issue.

amorari-nvidia: @pzelasko you are right let me change it to length

amorari-nvidia: I fixed both

pzelasko:
@amorari-nvidia As a last thing, could you remove the numba specaug implementation altogether in this PR? IIRC it will automatically kick in with numba 0.58 (or when numba+cuda versions are compatible for earlier versions), overriding the implementation here. CC @titu1994 pls double check if I'm right about this.

titu1994:
You can remove it entirely, or set the default flag of numba cuda to false. I think that's sufficient for the time being. 

If you can remove it cleanly, then that's fine too

amorari-nvidia:
I will just set the Numba version default flag to false for now

amorari-nvidia:
The Numba code is already set to false, so this is good to go.

> I will just set the Numba version default flag to false for now

The Numba SpecAugment version is already set to false, while the vectorized version is set to true and is the default code.

pzelasko:
great, re-triggered tests, LGTM as soon as it's green

pablo-garay:
> great, re-triggered tests, LGTM as soon as it's green

CI passed

pzelasko (approved):

pzelasko:
Congrats @amorari-nvidia on the first contribution!

PR #9182, opened 2024-05-13
add additonal condition for prev msdd cfgs

nithinraok:
# What does this PR do ?

TitaNet model was updated but some previous configurations were not. 
This PR provides support for prevevious msdd configuration versions.  

**Collection**: SD

# Changelog 
- Catch the `_target_` code in newer titanet versions, which is not available in previous versions. 

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


tango4j:
- Changed Speaker embedding extraction to avoid `torch.cat` in every for loop step for speed up. 
This does not make a lot of difference in general, but if CPU is busy with other tasks, this could make some difference in processing time.

- Now, device=null and if torch.cuda.is_cuda_available() == True, then MSDD is running on GPU. This means diarization default setting will now run on GPU if cuda is available.

tango4j (dismissed):

titu1994 (dismissed):

PR #9198, opened 2024-05-14
Use model-cast-to-bfloat16 rather than AMP-to-bfloat16 for inference.

galv:
# What does this PR do ?

I demonstrate, using transcribe_speech.py, that simply casting the entire model to bfloat16 gives about 15% higher performance than using automatic mixed precision. The reasons why are discussed in #9086

There are a few small modifications to our conformer encoder and multi head attention implementations to make this work. Basically, torch.float32 was unintentionally used at a few points.

I also disable updates to the batch norm statistics in this PR during inference. See the relevant comment in module.py.

Note that casting the preprocessor's input waveform to bfloat16 causes a serious accuracy degredation. I provide a warning about this. It is better to just do the preprocessor in float32, and then cast its output to bfloat16 (which is what I do).

I have verified that this works with Parakeet CTC 1.1B and Parakeet RNN-T 1.1B. I will upload a table showing WER and RTFx throughput when running transcribe_speech.py, using the new casting method and the old AMP method. RTFx improves by about 15%, and WER stays about the same.

**Collection**: ASR

# Changelog 

# Usage
* See transcribe_speech.py. Note that several parts of the code use AMP rather than simply casting the model.

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #9086


examples/asr/transcribe_speech.py
```diff
@@ -255,6 +257,14 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
     asr_model.set_trainer(trainer)
     asr_model = asr_model.eval()
 
+    if cfg.compute_dtype != "float32" and cfg.amp:
+        raise ValueError("amp=true is mutually exclusive with a compute_dtype other than float32")
```
nithinraok, line 261: use nemo `logging.error`

examples/asr/transcribe_speech.py
```diff
@@ -407,7 +421,27 @@ def autocast(dtype=None):
...
+                    # if i == 1:
+                    #     import nvtx
+                    #     pr = nvtx.Profile()
+                    #     pr.enable()  # begin annotating function calls
+                    #     ctx = torch.autograd.profiler.emit_nvtx()
+                    #     ctx.__enter__()
+                    #     torch.cuda.cudart().cudaProfilerStart()
+                    import time
+
+                    # import ipdb; ipdb.set_trace()
```
nithinraok, line 435: remove comments

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -282,14 +284,26 @@ def prepare_audio_data(cfg: DictConfig) -> Tuple[List[str], bool]:
...
+
+        with NamedTemporaryFile("w", suffix=".json", delete=False) as durations_f, open(cfg.dataset_manifest, "rt") as f:
+            for item in f:
+                item = json.loads(item)
+                item["audio_filepath"] = get_full_path(item["audio_filepath"], cfg.dataset_manifest)
+                if item.get("duration") is None:
+                    sound = sf.SoundFile(item["audio_filepath"])
+                    duration = sound.frames / sound.samplerate
+                    item["duration"] = duration
+                print(json.dumps(item), file=durations_f, flush=True)
```
nithinraok, line 297: why print is required here?

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -282,14 +284,26 @@ def prepare_audio_data(cfg: DictConfig) -> Tuple[List[str], bool]:
...
+        audio_key = cfg.get('audio_key', 'audio_filepath')
+
+        with NamedTemporaryFile("w", suffix=".json", delete=False) as durations_f, open(cfg.dataset_manifest, "rt") as f:
+            for item in f:
+                item = json.loads(item)
+                item["audio_filepath"] = get_full_path(item["audio_filepath"], cfg.dataset_manifest)
+                if item.get("duration") is None:
+                    sound = sf.SoundFile(item["audio_filepath"])
+                    duration = sound.frames / sound.samplerate
+                    item["duration"] = duration
```
nithinraok, line 296: soundfile reads full file when duration is not mentioned, is it chosen here for performance?

nemo/core/classes/module.py
```diff
@@ -58,6 +58,21 @@ def freeze(self) -> None:
...
+        # cause a denial of service attack, because an adversary can
+        # provide an input that causes the batch norm statistics to
+        # get set to NaN or infinity. Setting track_running_stats to
+        # False prevents updates of the batch norm statistics.
+        for sub_module in self.modules():
+            if isinstance(sub_module,
+                          (torch.nn.BatchNorm1d, torch.nn.LazyBatchNorm1d, torch.nn.BatchNorm2d,
+                           torch.nn.LazyBatchNorm2d, torch.nn.BatchNorm3d, torch.nn.LazyBatchNorm3d,
+                           torch.nn.SyncBatchNorm)):
+                sub_module.track_running_stats = False
```
nithinraok, line 74: I might be wrong. Isn't track running stats necessary for grad accumulation calculation?

nithinraok (changes requested):

nemo/collections/asr/parts/utils/transcribe_utils.py
galv: It's not. I coped it from: https://github.com/NVIDIA/NeMo/pull/9198/files/9a928206ec297f0b2652f4c382883c3bb29fd8a3#diff-f7b768bed0ebc1145787134787f17cc8dfb0defbf8758c8cb759b96e39fbbec5R361

In fact, there is a subtle bug with using print(). Flushing at a new line is not guaranteed. So the code I linked to actually chops off the last few lines of a manifest file. That was one of the reasons why I took a while to get this PR up. FYI @pzelasko , since I think you worked on pre-sorting before.

I will change from using print() to simply do json.write or json.save or w/e as an update to this PR.

https://docs.python.org/3/library/functions.html#print

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -219,10 +219,12 @@ def forward(self, query, key, value, mask, pos_emb, cache=None):
             q = q.transpose(1, 2)  # (batch, time1, head, d_k)
 
             n_batch_pos = pos_emb.size(0)
+
             p = self.linear_pos(pos_emb).view(n_batch_pos, -1, self.h, self.d_k)
             p = p.transpose(1, 2)  # (batch, head, time1, d_k)
 
             # (batch, head, time1, d_k)
+            # We are stuck casting this up to float32... ugh.
```
galv, line 227: remove

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -219,10 +219,12 @@ def forward(self, query, key, value, mask, pos_emb, cache=None):
             q = q.transpose(1, 2)  # (batch, time1, head, d_k)
 
             n_batch_pos = pos_emb.size(0)
+
```
galv, line 222: remove

nemo/collections/asr/parts/utils/transcribe_utils.py
galv: We need to get the duration of the audio file if we want to sort by duration. Sorting by duration can speed you up by a factor of two or so, IIRC. (I have the real data if you want to see it.) I'm not sure if I want to keep the change. It turns out that simply loading these audio files serially just to get their duration is slower than actually transcribing the data, annoyingly...

The duration fields are all "null" for the dataset I used here: https://bc.ngc.nvidia.com/datasets/1618966

So I threw this shim code in.

It is probably better to just update the duration field and make a new version of that dataset.

And throw an exception when someone wants to sort by duration but there is no duration field.

nemo/core/classes/module.py
galv: This in the freeze() method. freeze() sets requires_grad=False for all parameters. and sets self.eval(). Why would we need gradients in this situation?

Anyway, to answer your question, no you don't need to track running stats to calculate the gradient through batch norm. (Might be useful if you freeze a discriminator, but not a generator? IDK.)

galv:
FYI, the PR description has been updated to describe the benefit from using this mode of execution instead of AMP

examples/asr/transcribe_speech.py
```diff
@@ -255,6 +257,14 @@ def main(cfg: TranscriptionConfig) -> Union[TranscriptionConfig, List[Hypothesis
     asr_model.set_trainer(trainer)
     asr_model = asr_model.eval()
 
+    if cfg.compute_dtype != "float32" and cfg.amp:
+        raise ValueError("amp=true is mutually exclusive with a compute_dtype other than float32")
+
+    amp_dtype = torch.float16 if cfg.amp_dtype == "float16" else torch.bfloat16
```
titu1994, line 263: Lets change default to bf16 in another PR later. Fyi @nithinraok

examples/asr/transcribe_speech.py
```diff
@@ -407,7 +421,27 @@ def autocast(dtype=None):
...
+                    # import ipdb; ipdb.set_trace()
+
+                    start_time = time.time()
+                    transcriptions = asr_model.transcribe(audio=filepaths, override_config=override_cfg,)
+                    end_time = time.time()
+                    print("RTFx=", total_duration / (end_time - start_time))
+                    # if i == 1:
+                    #     pr.disable()
+                    #     ctx.__exit__(None, None, None)
+                    #     torch.cuda.cudart().cudaProfilerStop()
```
titu1994, line 444: Remove comments plus prints. Or replace print with logging

nemo/collections/asr/modules/audio_preprocessing.py
```diff
@@ -85,11 +85,28 @@ def __init__(self, win_length, hop_length):
...
+        # Normally, when you call to(dtype) on a torch.nn.Module, all
+        # floating point parameters and buffers will change to that
+        # dtype, rather than being float32. The AudioPreprocessor
+        # classes, uniquely, don't actually have any parameters or
+        # buffers from what I see. In addition, we want the input to
+        # the preprocessor to be float32, but need to create the
+        # output in appropriate precision. We have this empty tensor
+        # here just to detect which dtype tensor this module should
+        # output at the end of execution.
+        self.register_buffer("dtype_sentinel_tensor",
```
titu1994, line 97: Lol this is ingenious :D

nemo/collections/asr/parts/submodules/conformer_modules.py
```diff
@@ -348,7 +348,7 @@ def forward(self, x, pad_mask=None, cache=None):
             x = self.pointwise_activation(x)
 
         if pad_mask is not None:
-            x = x.float().masked_fill(pad_mask.unsqueeze(1), 0.0)
+            x = x.masked_fill(pad_mask.unsqueeze(1), 0.0)
```
titu1994, line 351: Hmm can we cast to appropriate dtype explicitly. Vahid added this for a particular reason which I forget

nemo/collections/asr/parts/utils/transcribe_utils.py
nithinraok: yeah reason I asked is, unfortunately reading and writing is where most of the time spent.

examples/asr/transcribe_speech.py
nithinraok: Make sense, @galv could you please also add this to the PR. Thanks!

nemo/core/classes/module.py
galv: Okay, sounds good. I will remove this change.

BTW, for the sake of good epistemics, I reran the test I did in the first comment with track_running_stats = False and with track_running_stats = True. The WER score does not change at all, for any dataset, up to 8 decimal places after the decimal point, for either the casting mode or AMP.

examples/asr/transcribe_speech.py
galv: just to clarify, `raise ValueError` is commonly done throughout this file. Are you sure?

examples/asr/transcribe_speech.py
galv: the default is float32 right now. bfloat16 can be made the default later.

nemo/collections/asr/parts/submodules/conformer_modules.py
galv: @VahidooX can you comment on why you added this cast? It is immediately followed by `self.depthwise_conv(x, cache=cache)`, which will end up casting the input `x` back down to fp16 when you run in automatic mixed precision. The only reason I can think of for why you might do this is if your cache is in float32, and adding a float16 value to the float32 cache caused a problem.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -943,15 +945,16 @@ class RelPositionalEncoding(PositionalEncoding):
...
 
-    def extend_pe(self, length, device):
+    def extend_pe(self, length, device, dtype):
         """Reset and extend the positional encodings if needed."""
         needed_size = 2 * length - 1
         if hasattr(self, 'pe') and self.pe.size(1) >= needed_size:
             return
         # positions would be from negative numbers to positive
         # positive positions would be used for left positions and negative for right positions
+        # fix this
```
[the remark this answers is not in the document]
galv: I talk to myself in comments often, sorry.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -997,15 +1000,15 @@ def __init__(self, att_context_size, **kwargs):
...
+    def extend_pe(self, length, device, dtype):
         """Reset and extend the positional encodings only at the beginning"""
         if hasattr(self, 'pe'):
             return
 
         positions = torch.arange(
             self.left_context, -self.right_context - 1, -1, dtype=torch.float32, device=device
         ).unsqueeze(1)
-        self.create_pe(positions=positions)
+        self.create_pe(positions=positions, dtype=dtype)
```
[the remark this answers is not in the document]
galv: Yes. I did a code search. You can see that squeezeformer_encoder.py is modified appropriately in this PR.

examples/asr/transcribe_speech.py
galv: To clarify, I don't want to change the default value right now. It would cause transcribe_speech.py to throw an error for all users who also turn on the amp=true.

galv:
This is ready for another round of review. Unfortunately the black + isort change made a lot of noise. But basically I removed a lot of cruft. This PR basically does only three things now:

1. Modify transcribe_speech.py to run using casting to float16/bfloat16 instead of AMP
1. Modify transcribe_speech.py to output RTFx scores, using the calculate_rtfx=True config. It is set to false by default (but I wouldn't be opposed to setting it to True by default later for the sake of educating users about their throughput results).
1. Make small changes to the positional encoding buffer and a call to masked_fill() to make sure that data is not accidentally casted up to float32.

examples/asr/transcribe_speech.py
nithinraok: Not sure of previous code, but usage of logging is the recommended way

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -17,8 +17,10 @@
 import re
 from dataclasses import dataclass
 from pathlib import Path
+from tempfile import NamedTemporaryFile
```
nithinraok, line 20: `NamedTemporaryFile` is unused

nemo/collections/asr/parts/utils/transcribe_utils.py
```diff
@@ -17,8 +17,10 @@
...
 from typing import List, Optional, Tuple, Union
 
+import soundfile as sf
```
nithinraok, line 23: `sf` is unused

nithinraok (approved):
Minor comments, LGTM otherwise

examples/asr/transcribe_speech.py
```diff
@@ -387,10 +401,22 @@ def autocast(dtype=None):
...
 
-    amp_dtype = torch.float16 if cfg.amp_dtype == "float16" else torch.bfloat16
+    if cfg.calculate_rtfx:
+        total_duration = 0.0
+
+        with open(cfg.dataset_manifest, "rt") as fh:
+            for line in fh:
+                item = json.loads(line)
+                if "duration" not in item:
+                    raise ValueError(
```
nithinraok, line 411: Revert this, don;t raise error but only warning to say that it is recommended to have duration in manifest for faster inference!

nithinraok (changes requested):
revert the duration required field and also make sure to compute RTF when duration is provided, its not mandatory to provide duration

examples/asr/transcribe_speech.py
galv: No. It's behind the calculate_rtfx flag, which defaults to false. If someone passes in a manifest without durations, transcribe_speech.py will still continue to run.

nemo/collections/asr/parts/submodules/conformer_modules.py
galv: Haven't heard back. Dropping this one.

examples/asr/transcribe_speech.py
galv: I'm not going to change to logging.error(). It's a critical error. Why shouls we continue execution in this case?

galv:
@nithinraok @pzelasko @titu1994 I think this change is good to go at this point.

nithinraok (dismissed):
LGTM!

nemo/collections/asr/parts/submodules/conformer_modules.py
titu1994: My bad, wasnt notified of ping. I am extremely wary of modifying any part of mha or relpos mha operations. I'd seriously like them to continue operating at fp32. Vahid worked a lot on this module, he has these explicit casts for a reason because he was unable to get stable training at all on fp16 and bf16 without these. 

The goal of this PR was to affect only inference, this affects training so id request to revert any changes to any file concerning MHA or RelPosMHA

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -650,13 +650,14 @@ def _compute_out_global_to_all(
...
-            is_index_masked.transpose(2, 3), torch.finfo(global_attn_scores.dtype).min,
+            is_index_masked.transpose(2, 3),
+            torch.finfo(global_attn_scores.dtype).min,
         )
 
         global_attn_scores = global_attn_scores.view(batch_size * self.h, max_num_global_attn_indices, seq_len)
 
         # compute global attn probs
-        global_attn_probs_float = nn.functional.softmax(global_attn_scores, dim=-1, dtype=torch.float32)
+        global_attn_probs_float = nn.functional.softmax(global_attn_scores, dim=-1)
```
titu1994, line 660: See above

nemo/collections/asr/parts/submodules/conformer_modules.py
titu1994: A middle ground is the following - check for self.training - if true, then use the explicit fp32 casts here and below, otherwise perform no cast (and we anyway explicitly cast to fp16/bf16 so it wont make a difference in inference)

titu1994 (changes requested):
Overall everything looks good, apart from chages to MHA and RelPosMHA. Please see comments.

nemo/collections/asr/parts/submodules/conformer_modules.py
VahidooX: I sent @galv the details on this casting. It is not used for train or inference in nemo. @borisfom added that, I guess to make it work with trt or onxx conversions. Let's check it out with Boris before dropping that.
Here is the PR for this specific one:
https://github.com/NVIDIA/NeMo/pull/3787/files

borisfom: Sure if that works without the cast, do remove it - both ONNX exporter and TRT went very far since then.

nemo/collections/asr/parts/utils/transcribe_utils.py
pzelasko: Indeed my bad; I forgot to add `f.flush()` at the end of line writing. `f.write()` would have the same issue. `print` is just fine IMO but if you want to change it I don't mind.

nemo/collections/asr/parts/utils/transcribe_utils.py
pzelasko: It's more complicated than that: if the format has a header, such as WAV or FLAC, soundfile will only read the header unless you call `.read()` on that object. But if it's something like MP3, it may read the whole thing.

pzelasko:
Looks good to me!

titu1994 (approved):
Thanks for the fixes !

vadimkantorov:
Also, does inference support torch.compile for Conformers (with no graph breaks)? Is there any test for zero graph breaks (for conformer and for frontend)?

galv:
@vadimkantorov 

> Also, does inference support torch.compile for Conformers (with no graph breaks)? 

> Is there any test for zero graph breaks (for conformer and for frontend)?

People have not adequately prioritized making nemo models run in torch.compile AFAIK. I remember having to change a few things to make dynamo accept a nemo model, but that was many months ago and things could be different. graph breaks aren't that bad, BTW.

vadimkantorov:
> graph breaks aren't that bad

Yeah, but likely for conformer / frontend it can even be made strictly 0 graph breaks, which is nice

PR #9242, opened 2024-05-17
Dgalvez/fix greedy batch strategy name

galv:
# What does this PR do ?

Change the name of greedy_batched CTC decoding strategy to greedy_batch.

Fix a spurious warning.

**Collection**: ASR

# Changelog 

1. Lazily warn about using greedy strategy instead of greedy_batch strategy.

Previously, the warning would often run spuriously, since several
existing code paths simply call "change_decoding_strategy()" after
having first initialized a Module, rather than changing the config
before initializing the Module. This can be confusing.

The only problem I can see with this is that using logging inside a
forward() method might interfer with some compiler toolkits like
Torchscript or thunder.compile. Presumably it would be easy to add a
conditional statement to avoid this statement in a compiler context if
necessary.

2. Change "greedy_batched" CTC decoding strategy to "greedy_batch

This will be consistent with the RNN-T decoding strategy named
"greedy_batch".

# Usage
N/A

# GitHub Actions CI

**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/parts/submodules/ctc_decoding.py
+++ b/nemo/collections/asr/parts/submodules/ctc_decoding.py
@@ -213,20 +213,20 @@ class AbstractCTCDecoding(ConfidenceMixin):
         self.batch_dim_index = self.cfg.get('batch_dim_index', 0)
         self.word_seperator = self.cfg.get('word_seperator', ' ')
 
-        possible_strategies = ['greedy', 'greedy_batched', 'beam', 'pyctcdecode', 'flashlight']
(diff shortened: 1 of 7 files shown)


nithinraok (dismissed):
LGTM

nithinraok:
pls send PR to `2.0.0rc0`

galv:
Closed in favor of https://github.com/NVIDIA/NeMo/pull/9243, so that this changes incorporated in r2.0 release.

PR #9247, opened 2024-05-17
[Nemo CICD] Increase time limit for Speech_Checkpoints_tests (#9186)

pablo-garay:
[Nemo CICD] Increase time limit for Speech_Checkpoints_tests (#9186)


titu1994:
What exactly is this checkpoint test ? Its way too expensive. Lets trim it down

PR #9259, opened 2024-05-20
add large model stable training fix and contrastive loss update for variable seq

nithinraok:
# What does this PR do ?

This PR brings two updates:
* Fix SSL contrastive loss to support loss computation for variable input length rather than computation based on subsampling factor
* Add an option to remove bias from Linear and Conv layers in Conformer layers to support scaling for multi billion parameter training [tested currently for 1B parameter using bf16] 

**Collection**: ASR

# Changelog 
- Updated ssl_loss 
- Added argument for conformerlayer to optionally remove bias, default is False

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [x] Bugfix
- [ ] Documentation


examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_ctc_bpe_streaming.yaml
```diff
@@ -80,6 +80,7 @@ model:
     feat_out: -1 # you may set it if you need different output size other than the default d_model
     n_layers: 17
     d_model: 512
+    remove_bias: False # whether to remove bias in the feedforward, MHA and convolution modules
```
titu1994, line 83: Isn't this double negative ? You have to disable bias to get better stability or add bias to get better stability ? If you have to remove bias, then instead call it use_bias: false. 

And set it as the default in all configs,small and large

titu1994 (commented):
Nice improvement, minor comment

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_ctc_bpe_streaming.yaml
nithinraok: `remove_bias: True` is informing to remove bias, and false of remove bias says no bias would be removed. I mentioned remove bias because having bias is native and hence removing is something out of common. 

I haven;t set it as default as this would be different from published model configs. What do you think?

titu1994: Published models have their own configs, so as long as code default has "use_bias" True, older models continue to work. Just newer models that train om new configs will have it set as false. 

Exactly because bias is the normal argument for pytorch, it's best to say use_bias True and setting it to false is new thing. 

Remove bias suggests to me bias was already added and now you will go and remove it from a build model

nithinraok: Sounds good, updated `remove_bias` to `use_bias`

titu1994 (approved):
Looks great !

PR #9270, opened 2024-05-21
TRT-LLM Export Code Cleanup

oyilmaz-nvidia:
# What does this PR do ?

Cleaning all of the unused functions, classes, etc, after the TRT-LLM API integration.


janekl (dismissed):

jiemingz:
[get_tensor_parallel_group](https://github.com/oyilmaz-nvidia/NeMo/blob/export-cleanup/nemo/export/trt_llm/tensor_utils.py#L47C5-L47C30) is not used anymore.

jiemingz:
I believe [cpu_map_location] (https://github.com/oyilmaz-nvidia/NeMo/blob/export-cleanup/nemo/export/trt_llm/utils.py#L63)
is duplicated with [this](https://github.com/oyilmaz-nvidia/NeMo/blob/export-cleanup/nemo/export/trt_llm/nemo/convert.py#L26) 

Also it seems to be `cpu_map_location` and `gpu_map_location` should be in `trt_llm/nemo/nemo.py` since thats where they're only used

jiemingz:
[This function ](https://github.com/oyilmaz-nvidia/NeMo/blob/export-cleanup/nemo/export/trt_llm/nemo/convert.py#L418) is only used by `build` and `refit` so it can be removed, eventually though we'll have to add it back as well when `build` and `refit` are moved onto the new API

jiemingz (changes requested):
Please check my comments above @oyilmaz-nvidia, thanks!

oyilmaz-nvidia:
Thanks @jiemingz for your review. Updating the code per your comments.

jiemingz (approved):

meatybobby (approved):

PR #9276, opened 2024-05-22
Refactor Quantizer for reusing in QAT

kevalmorabia97:
# What does this PR do ?

This PR refactors the Quantizer class used for PTQ so it can be reused for QAT as well (follow-up PR)

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


examples/nlp/language_modeling/conf/megatron_quantization.yaml
```diff
@@ -20,8 +20,13 @@ trainer:
   precision: bf16 # 16, 32, or bf16
   enable_checkpointing: false
 
+model:
```
kevalmorabia97, line 23: These are standard parameters for a nemo config hence adding them directly here instead of separately keeping them and then overwriting the model config

nemo/export/quantize/quantizer.py
```diff
@@ -72,89 +79,13 @@ class Quantizer:
     for TensorRT-LLM deployment. This is useful to getting baseline results for a full-precision model.
     """
 
-    def __init__(
```
kevalmorabia97, line 75: Move quantization and inference arguments to `quantize()` and trtllm export related arguments to `export` so depending on what function a user wants to use, they can use it without worrying about other one. Also duplicated `decoder_type` to quantization parameters since it is used there as well

nemo/export/quantize/quantizer.py
```diff
@@ -72,89 +79,13 @@ class Quantizer:
...
-                "num_bits": 8 if quantization_config.algorithm == "int8_sq" else (4, 3),
-                "axis": None,
-                "enable": enable_quant_kv_cache,
-            }
-
-            self.quant_cfg = quant_cfg
-        else:
-            self.quant_cfg = None
-
-    def _load_model(
```
kevalmorabia97, line 124: model/trainer loading may be different for different pipelines (e.g. SFT uses a different model and trainer class and has some other changes) so moved this part to the example script instead of this common place

nemo/export/quantize/quantizer.py
```diff
@@ -167,80 +98,73 @@ def dummy():
...
-        model_cfg = load_config(model_file)
-
-        with open_dict(model_cfg):
-            model_cfg.activations_checkpoint_method = None
-            model_cfg.activations_checkpoint_granularity = None
-            model_cfg.sequence_parallel = False
-            if tensor_model_parallel_size is not None:
-                model_cfg.tensor_model_parallel_size = tensor_model_parallel_size
-            if pipeline_model_parallel_size is not None:
-                model_cfg.pipeline_model_parallel_size = pipeline_model_parallel_size
```
kevalmorabia97, line 185: this part is automatically taken care of since it is moved to `model:` section in the yaml

nemo/export/quantize/quantizer.py
```diff
@@ -167,80 +98,73 @@ def dummy():
...
-            model_cfg.name = "modelopt"
-            model_cfg.apply_rope_fusion = False
+            model_config.name = "modelopt"
+            model_config.apply_rope_fusion = False
 
-        return model_cfg
+        return model_config
 
     @staticmethod
-    def _sample_output(model):
```
kevalmorabia97, line 196: This function is just moved down close to export where it is used for better readability

nemo/export/quantize/quantizer.py
```diff
@@ -167,80 +98,73 @@ def dummy():
...
-            length_params={
-                "max_length": 100,
-                "min_length": 100,
-            },
-        )
-
-        if torch.distributed.get_rank() == 0:
-            print(f'Example NeMo output after PTQ: {response["sentences"]}"')
-
     def quantize(
```
kevalmorabia97, line 119: Different quantization pipelines may have different way of using dataloader or writing the forward loop hence changed the arguments to take  forward loop directly instead of dataloader (mtq.quantize also requires forward loop from a user instead of dataloader).

nemo/export/quantize/quantizer.py
```diff
@@ -167,22 +98,14 @@ def dummy():
...
-    ):
-        model_cfg = load_config(model_file)
+        model.set_inference_config(OmegaConf.to_container(inference_config))
 
+    @staticmethod
+    def modify_model_config(model_cfg: DictConfig) -> DictConfig:
         with open_dict(model_cfg):
             model_cfg.activations_checkpoint_method = None
             model_cfg.activations_checkpoint_granularity = None
             model_cfg.sequence_parallel = False
```
kevalmorabia97, line 108: @janekl do you remember why sequence parallel is set to False? I tested for QAT and didn't have any issues with sequence parallel

nemo/export/quantize/quantizer.py
```diff
@@ -209,38 +135,60 @@ def _sample_output(model):
...
+        if "awq" in algorithm:
+            weight_quantizer = quant_cfg["quant_cfg"]["*weight_quantizer"]
+            if isinstance(weight_quantizer, list):
+                weight_quantizer = weight_quantizer[0]
+            weight_quantizer["block_sizes"][-1] = quantization_config.awq_block_size
+
+        # Always turn on FP8 kv cache to save memory footprint.
+        # For int8_sq, we use int8 kv cache.
+        # TODO: Investigate why enabling FP8 kv cache will cause accuracy regressions for Nemotron.
+        enable_quant_kv_cache = "int8" not in algorithm and decoder_type != "gptnext"
```
kevalmorabia97, line 175: @janekl Should this instead be like below:

```suggestion
        enable_quant_kv_cache = "fp8" in algorithm and decoder_type != "gptnext"
```

Since the comment says to enable only for fp8 but the code enable for non-int8 hence int4 also does the kv cache quantization which makes the results worse

examples/nlp/language_modeling/conf/megatron_quantization.yaml
kevalmorabia97, line 1: why do we need to explicitly set the inference config? I was able to run quantization and export even if I do not set the inference config

nemo/export/quantize/quantizer.py
```diff
@@ -167,23 +100,17 @@ def dummy():
...
-        model_cfg = load_config(model_file)
+        if inference_config is not None:
+            model.set_inference_config(OmegaConf.to_container(inference_config))
 
+    @staticmethod
+    def modify_model_config(model_cfg: DictConfig) -> DictConfig:
+        """Modify model config for quantization."""
         with open_dict(model_cfg):
             model_cfg.activations_checkpoint_method = None
             model_cfg.activations_checkpoint_granularity = None
```
kevalmorabia97, line 111: Can we remove this disable of activation checkpointing as well? For the SFT model there is no accuracy regression even if we enable activation checkpointing plus that reduces GPU memory requirement

nemo/export/quantize/quantizer.py
janekl: There is a functional issue during evaluation with sequence parallel=True. I double checked now and the code errors out here https://github.com/NVIDIA/Megatron-LM/blob/c4d12e26b2dc25a2eab7da92e2ac30338c0ed3de/megatron/core/model_parallel_config.py#L281. So it seems that single-GPU tests won't work with sequence parallelism enabled.

Related examples where it is set to False explicitly on evaluation (see also other settings):

1. https://github.com/NVIDIA/NeMo/blob/b0f3138a6be7fab3175deb8935f8492aeb1445bd/examples/nlp/language_modeling/megatron_gpt_eval.py#L226

2. https://github.com/EleutherAI/lm-evaluation-harness/blob/1060b68d35f6a4f1b571e7b5241049ead64b2936/lm_eval/models/nemo_lm.py#L52

nemo/export/quantize/quantizer.py
janekl: Do you mean model section from model_config.yaml or megatron_quantization.yaml modified here? Or please just link the piece of code that updates parallelism settings based on appropriate model section.

It's important to easily specify parallelism settings as this typically is going to be different than settings used in model_config.yaml for example, during training

examples/nlp/language_modeling/conf/megatron_quantization.yaml
janekl: I believe that megatron_quantization.yaml config was derived based on [megatron_gpt_inference.yaml config](https://github.com/NVIDIA/NeMo/blob/main/examples/nlp/language_modeling/conf/megatron_gpt_inference.yaml). I'm not entirely sure but it was the state of things when I first looked into this in ammo repository late '23 (you could ask around).

If that's the case I would suggest let's please keep this config in analogy to the megatron_gpt_inference.yaml config and that config specifies parallelism settings and model file at the top level.

nemo/export/quantize/quantizer.py
janekl: 1. Whatever we do it should be exactly the same as in [nemo_ptq.py](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/06a155388edd4a240051176a67a11886b15db082/llm_ptq/nemo_ptq.py#L237-L239) from modelopt
2. Could you @realAsma and/or @cjluo-omniml comment on this proposal?

examples/nlp/language_modeling/conf/megatron_quantization.yaml
janekl: It might be some relict, see also this comment https://github.com/NVIDIA/NeMo/pull/9276#discussion_r1624472661.

Do you mean that we don't really need the line below as it all works with default inference settings? https://github.com/NVIDIA/NeMo/blob/b0f3138a6be7fab3175deb8935f8492aeb1445bd/nemo/export/quantize/quantizer.py#L228

Even so my intuition is that we could keep the inference parameters in megatron_quantization.yaml explicitly anyway as defaults might change at some point.

nemo/export/quantize/quantizer.py
janekl: I'm not sure regarding this, see also related examples in https://github.com/NVIDIA/NeMo/pull/9276#discussion_r1624463595. Maybe you could ask internally?

janekl:
Could you demostrate/link how do you want to use that refactored `Quantizer`?

janekl:
Note that changes will need to be unstreamed to another project as well here https://github.com/NVIDIA/NeMo-Framework-Launcher/blob/main/launcher_scripts/conf/ptq/model/quantization.yaml. I think we'll need to chop this slightly

examples/nlp/language_modeling/megatron_quantization.py
```diff
@@ -64,7 +70,18 @@ def main(cfg) -> None:
     if not torch.cuda.is_available():
         raise EnvironmentError("GPU is required for the inference.")
 
-    quantizer = Quantizer(cfg.quantization, cfg.inference, cfg.export, cfg.trainer)
+    # Overwrite model config with the one from the model checkpoint and apply quantization modifications
+    model_cfg = load_config(cfg.model.restore_from_path)
+    with open_dict(model_cfg):
+        for key, val in cfg.model.items():
+            model_cfg[key] = val
+    model_cfg = Quantizer.modify_model_config(model_cfg)
```
janekl, line 78: Since L75-77 also modify model config, they could become a part of `modify_model_config` to keep it all in a single location

nemo/export/quantize/quantizer.py
janekl: This would be my major remark in general, other things should be more or less clear

examples/nlp/language_modeling/conf/megatron_quantization.yaml
kevalmorabia97: Yes with default inference config, it still works fine. Also note that ideally for PTQ we recommend using forward step and not predict step (as done in PTQ example) and hence there shouldn't be any need for inference config anyway

nemo/export/quantize/quantizer.py
realAsma: "comment says to enable only for fp8" -> That is not correct. Comments says to enable fp8 for all formats except int8_sq. So yes, int4 quantization will use fp8 kv cache quantization. For most of the models fp8 kv cache does not cause much accuracy degradation.

nemo/export/quantize/quantizer.py
realAsma: Sequence parallel is a technique used only for training, not for evaluation or inference (It requires the sequence length to be divisble by the TP size). This wont be true during token by token inference.

kevalmorabia97: Do we care how it is handled in evaluation though? Model can be PTQ/QAT on sequence parallel or with activations checkpointing enabled since it reduces memory requirement. When users run inference/eval on these models, its fine for these scripts to explicitly disable it. 
Plus users always have the option to disable sequence parallel if they want to by setting `model.sequence_parallel=False` in their config files.

kevalmorabia97:
> Could you demostrate/link how do you want to use that refactored `Quantizer`?

Its used in https://github.com/NVIDIA/NeMo/pull/9276

examples/nlp/language_modeling/conf/megatron_quantization.yaml
kevalmorabia97: I see. 

I only looked at pretraining and SFT-related scripts that use different format as I have updated here. 

For PTQ here we only need `model.tp`, `model.pp` so we just manually overwrite it. 

But for other training based scripts there are many more `model.` parameters that needs to be overwritten so they are directly defined under `model:` section in the yaml so we don't have to manually list all overwrites. I like this approach more as it is more cleaner

If you think this updated version doesn't look good, I can revert PTQ to the previous format. Only thing is that then QAT and PTQ will have different way of defining parameters which may not look good. Thoughts?

nemo/export/quantize/quantizer.py
kevalmorabia97: When model is loaded, it read `model:` section from `model_config.yaml` but then we overwrite values from `model:` section in `megatron_quantization.yaml` after loading the model so we can have the final configuration we want to use for PTQ/QAT.

If we look at training based example in https://github.com/NVIDIA/NeMo/blob/main/.github/workflows/cicd-main.yml#L4258 we can see how they define TP/PP, Activation checkpointing, model restore path etc under `model.` section.

For QAT since it uses the training pipeline, we need to use the parameters how they do it in training yaml files. The question is whether we want to have similar format for PTQ or have it different than QAT

examples/nlp/language_modeling/conf/megatron_quantization.yaml
kevalmorabia97: I would suggest for PTQ example as well we use `model.fwd_bwd_step(dataloader_iter, forward_only=True)` instead of `model.predict_step` since that's what ModelOpt docs recommend as well and that would completely remove `inference:` section in yaml

nemo/export/quantize/quantizer.py
kevalmorabia97: The only issue with that is `inference_config` and `export_config` are optional for QAT and we only need `quantization_config` in Quantizer. How do you feel about passing `quantization_config` to `Quantizer.__init__` and then `export_config` will be an argument to `Quantizer.export()` and we can remove `inference_config` all together as it is not required as I suggested in another comment?

nemo/export/quantize/quantizer.py
kevalmorabia97: INT4 has a little accuracy degradation if we enable KV Cache hence I though we didn't want to enable it. PTQ val lm loss with KV cache enabled is 1.35 vs 1.31 and QAT was around 1.2 instead of 1.17 (don't exactly remember QAT numbers)

examples/nlp/language_modeling/megatron_quantization.py
kevalmorabia97: SFTModel already has a utility function doing this called [`merge_cfg_with`](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/parts/mixins/nlp_adapter_mixins.py#L470) hence I kept it separate

nemo/export/quantize/quantizer.py
realAsma: sequence_parallel could break PTQ (if the calibration seq len is not divisble by TP size). Hence I recommend disabling it by default. 
activation_checkpointing only matters for QAT, but does not affect PTQ. It should be okay to enable it by default.

kevalmorabia97: Got it. Thanks for the clarification. I will keep sequence_parallel update here and remove activations checkpointing update so users can set it what they want to. I will keep them on by default in the QAT example yaml as well

examples/nlp/language_modeling/megatron_quantization.py
janekl: 1. Could we move overriding fields with `cfg.model` to ` Quantizer.modify_model_config` anyway? We modify config, let's do it in one place to keep things organized i.e. that `modify_model_config` method
2. [minor] Can we use https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#omegaconf-update instead of manual loop in L75-77

nemo/export/quantize/quantizer.py
janekl: I like the proposal, this is what I also was thinking of here https://github.com/kevalmorabia97/NeMo/compare/kmorabia/update-quantizer-for-qat...janekl:NeMo:review

Removing `inference_config` and using `forward_bwd_step` sounds good but we'll need to run accuracy check. There are two options:
* run this with [nemo_example.sh](https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/llm_ptq/scripts/nemo_example.sh)in the source modelopt repo
* run Lambada with [test_nemo_export.py](https://github.com/NVIDIA/NeMo/blob/main/tests/export/test_nemo_export.py) on quantized checkpoints (with a new quantized checkpoint from megatron_quantization.py after changes) - that one is also a part of internal CI

examples/nlp/language_modeling/conf/megatron_quantization.yaml
janekl: I see, introducing that `model:` section is fine

nemo/export/quantize/quantizer.py
janekl: OK. I only wonder if we can avoid specifying decoder type. Is there a use case when only `quantization_config` is used, but **not** `export_config`? If `quantization_config` always comes with `export_config`, we may resign from duplicating `decoder_type`

nemo/export/quantize/quantizer.py
janekl: OK

nemo/export/quantize/quantizer.py
janekl: Thanks for catching this. The change is fine but requires MMLU accuracy check for, say, Llama3 (8b, 70b) int4_awq.

It's quite easy for me to test so I'll handle that

examples/nlp/language_modeling/conf/megatron_quantization.yaml
janekl: As for `fwd_bwd_step`: fine, just needs accuracy checks

janekl:
We'll need to do follow up MRs to:
* https://github.com/NVIDIA/NeMo-Framework-Launcher/blob/main/launcher_scripts/conf/ptq/model/quantization.yaml 
* https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/llm_ptq/nemo_ptq.py
(I know it's duplication but it is what it is)

nemo/export/quantize/quantizer.py
janekl: ...but first let me double check: @realAsma I conclude from your comment that you suggest keeping the current behavior, right?

Only that comment "comment says to enable only for fp8" needs to be changed then.

nemo/export/quantize/quantizer.py
kevalmorabia97: Its possible that users just quantize in one script and separately export. Or they may skip export all together (currently int4 export is not working so we don't export in qat example). Also given that quantization post processing needs decoder type, quantization config should explicitly contain this parameter and not rely on some other config

kevalmorabia97:
@janekl should I rename `alpha` to `sq_alpha` so it is clear alpha is used only for SQ?

janekl:
> @janekl should I rename `alpha` to `sq_alpha` so it is clear alpha is used only for SQ?

OK, good idea (same applies to Launcher ofc.)

nemo/export/quantize/quantizer.py
```diff
@@ -171,22 +157,13 @@ def dummy():
...
         with open_dict(model_cfg):
-            model_cfg.activations_checkpoint_method = None
-            model_cfg.activations_checkpoint_granularity = None
-            model_cfg.sequence_parallel = False
-            if tensor_model_parallel_size is not None:
-                model_cfg.tensor_model_parallel_size = tensor_model_parallel_size
-            if pipeline_model_parallel_size is not None:
-                model_cfg.pipeline_model_parallel_size = pipeline_model_parallel_size
-            # Only custom ModelOpt spec is supported for PTQ: this custom spec is largely based on local Megatron-LM
+            if model_cfg.sequence_parallel:
```
janekl, line 163: Not sure if all models specify this - maybe for safety let's write `model_cfg.get("sequence_parallel", False)`

nemo/export/quantize/quantizer.py
```diff
@@ -212,38 +188,24 @@ def _sample_output(model):
...
+        assert self.quant_cfg is not None, "Quantization algorithm is not set"
 
-        model.set_inference_config(OmegaConf.to_container(self.inference_config))
-
-        def forward_loop(model):
-            print("Calibrating the model...")
-            for i, batch in enumerate(tqdm(dataloader)):
-                model.predict_step(batch, i)
+        logging.info(f"Quantizing model to {self.quantization_config.algorithm}...")
+        Quantizer._setup(model)
```
janekl, line 198: let's please write `Quantizer` -> `self` (I realize it's a static method, but it just looks odd; I want to make sure a user instantiates a `Quantizer`, OK?)

nemo/export/quantize/quantizer.py
```diff
@@ -254,22 +216,24 @@ def forward_loop(model):
...
         return model
 
-    def export(self, model, model_save: str):
+    def export(self, model: MegatronGPTModel):
         """Export model to '.qnemo' format for TensorRT-LLM engine build."""
+        assert self.export_config is not None, "Export config is not set"
         torch_dtype = torch_dtype_from_precision(self.export_config.dtype)
 
-        self._sample_output(model)
+        Quantizer._sample_output(model)
```
janekl, line 224: same here

examples/nlp/language_modeling/megatron_gpt_quantization.py
```diff
@@ -59,31 +65,45 @@ def get_calib_dataloader(data="cnn_dailymail", batch_size=64, calib_size=512, ma
...
-        dataloader = None
+        dataloader = [data for data in data_iter]
 
-    model = quantizer.quantize(
-        cfg.model_file, dataloader, cfg.tensor_model_parallel_size, cfg.pipeline_model_parallel_size
-    )
+        def forward_loop(model):
+            # NOTE: Alternatively you can also use `model.forward_bwd_step(data_iter, forward_only=True)`
+            # if your model is setup for training.
+            model.set_inference_config(OmegaConf.to_container(cfg.inference))
```
janekl, line 100: This inference config can / cannot go into Quantizer - why / why not?

nemo/export/quantize/quantizer.py
```diff
@@ -63,38 +69,40 @@ class Quantizer:
...
-        export_config: DictConfig,
-        trainer_config: DictConfig,
-    ):
+    def __init__(self, quantization_config: Optional[DictConfig], export_config: Optional[DictConfig]):
+        """Initialize Quantizer with quantization and export configurations.
+
+        Expected keys in `quantization_config`:
+            - algorithm: str
+            - decoder_type: str
+            - awq_block_size: int (only for awq algorithms)
```
janekl, line 87: `sq_alpha` as expected key as well now

examples/nlp/language_modeling/megatron_gpt_quantization.py
kevalmorabia97: Whether user needs an inference config depends on what forward loop they use. If they use predict_step then only they need inference config so makes sense to have that with forward loop configuration in the user code instead of quantizer

nemo/export/quantize/quantizer.py
janekl: I ran tests for enabling / disabling KV cache and it looks it doesn't really matter much for accuracy.
```
     MODEL  QFORMAT  GPUS DTYPE KV_CACHE  MMLU  LAMBADA  TOKENS_PER_SEC
 LLAMA3-8B int4_awq     1  fp16   Enable 0.628   0.662          2045.11
 LLAMA3-8B int4_awq     1  fp16  Disable 0.628   0.663          2037.57
LLAMA3-70B int4_awq     1  fp16   Enable 0.779   0.727           459.53
LLAMA3-70B int4_awq     1  fp16  Disable 0.778   0.729              NaN
```
benchmark.py OOM-ed for LLAMA3-70B with KV_CACHE = Disable

examples/nlp/language_modeling/megatron_gpt_quantization.py
janekl: Got it, makes sense then, thanks

nemo/export/quantize/quantizer.py
kevalmorabia97: Can you try non-AWQ `int4` as well please?

examples/nlp/language_modeling/conf/megatron_gpt_quantization.yaml
```diff
@@ -20,21 +20,22 @@ trainer:
...
+model:
+  tensor_model_parallel_size: 1
+  pipeline_model_parallel_size: 1
+  restore_from_path: llama2-7b-fp16.nemo # Nemo file path
+
 quantization:
-  quantize_bmm1: false
-  algorithm: fp8 # int8_sq, fp8, int8, int4_awq, null
+  decoder_type: ${export.decoder_type} # gptnext, gpt2, llama
+  algorithm: fp8 # null, int8, int8_sq, fp8, int4_awq, w4a8_awq, int4
```
janekl, line 30: Nice to have `int8`, `w4a8_awq`, `int4` but bear in mind these have never been tested.

I would opt for removing them from description until someone tests this first

nemo/export/quantize/quantizer.py
janekl: Let's postpone this so as to not to block this MR. You could optionally open some issue / ticket to track it.

Please also modify the comment "# Always turn on FP8 kv cache to save memory footprint." to match the implementation as we discussed

nemo/export/quantize/quantizer.py
janekl: So in the end it turns out that it's OK to remove setting these to `None`?:
```
model_cfg.activations_checkpoint_method = None
model_cfg.activations_checkpoint_granularity = None
```
I hope it will not blow up in some large tests

kevalmorabia97: Users always have the option to disable it in the yaml if they want to, but we should not restrict them to not be able to use it as it allows reducing GPU memory requirement

kevalmorabia97: I will set them as None in the yaml file so there is no difference than previous version

examples/nlp/language_modeling/conf/megatron_gpt_quantization.yaml
kevalmorabia97: Sure

nemo/export/quantize/quantizer.py
kevalmorabia97: As Asma mentioned, the comment is correct. KV cache is FP8 quantized regardless of the quantization algorithm. 
Whether we enable fp8 kv cache for int4 model - that I will bring up in the qat PR

janekl (approved):

PR #9299, opened 2024-05-23
Speeds up copying of necessary artifact files with SaveRestoreConnector

terrykong:
# What does this PR do ?

Previously, the SaveRestoreConnector would copy and untar entire
checkpoints just to copy out a tokenizer. For models in the >100GB, this
led to timeouts since only rank=0 did this work, while other ranks moved
on and waited at an all-gather barrier (observed NCCL timeout at 10min).

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/core/connectors/save_restore_connector.py
+++ b/nemo/core/connectors/save_restore_connector.py
@@ -19,7 +19,7 @@ import shutil
 import tarfile
 import tempfile
 import uuid
-from typing import Optional, Set, Union
+from typing import Callable, Optional, Set, Union
 
 import torch
 from omegaconf import DictConfig, OmegaConf
@@ -141,7 +141,7 @@ class SaveRestoreConnector:
                 else:
                     # Extract the nemo file into the temporary directory
                     self._unpack_nemo_file(
-                        path2file=restore_path, out_folder=tmpdir, extract_config_only=return_config is True
+                        path2file=restore_path, out_folder=tmpdir, extract_fn=return_config and (lambda name: '.yaml' in name)
                     )
 
                 # Change current working directory to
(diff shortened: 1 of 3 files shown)


nemo/core/connectors/save_restore_connector.py
```diff
@@ -479,37 +481,35 @@ def _handle_artifacts(self, model, nemo_file_folder):
...
-
-                        # Update artifacts registry
-                        new_artiitem = model_utils.ArtifactItem()
-                        new_artiitem.path = "nemo:" + artifact_uniq_name
-                        new_artiitem.path_type = model_utils.ArtifactPathType.TAR_PATH
-                        model.artifacts[conf_path] = new_artiitem
-                finally:
-                    # change back working directory
-                    os.chdir(cwd)
+            for conf_path, artiitem in tarfile_artifacts:
```
titu1994, line 484: These changes aren't required for the scope of just supporting different file from being extracted.

nemo/core/connectors/save_restore_connector.py
```diff
@@ -479,37 +481,35 @@ def _handle_artifacts(self, model, nemo_file_folder):
...
             # Get path where the command is executed - the artifacts will be "retrieved" there
             # (original .nemo behavior)
-            cwd = os.getcwd()
-            # Step into the nemo archive to try and find the file
-            # TemporaryDirectory context must always be outer to try-catch chdir otherwise it crashes on Windows
-            with tempfile.TemporaryDirectory() as archive_dir:
-                try:
-                    # unpack all restorations paths (nemo checkpoints)
-                    # in nemo checkpoints all resources contain hash in name, so there should be no collisions
-                    for path in restoration_paths:
```
titu1994, line 489: Continue with old logic, wherever extraction is required, only change that part to extract filtered filer rather than all checkpoint files.

titu1994 (changes requested):

nemo/core/connectors/save_restore_connector.py
terrykong: Could you elaborate on why this change isn't needed in this "scope"? Do you mean I should move this logic elsewhere? If so, could you advise.

Or are you saying this change isn't required at all? I tried outlining the issue I observed in the description. This change is to avoid naively copying huge checkpoints just to extract a few small artifact files.

nemo/core/connectors/save_restore_connector.py
terrykong: What is "old logic" here? I don't see the reason to create a tmp dir (potentially on a diff file system) just to pull one file into the nemo dir. In my experiments with a large model, b/c this work was only done on rank=0, it took >10min and then NCCL timed out.

I believe this is a general improvement, but also helps with a functional issue I am facing.

titu1994: Creation of temp dir is not the slowdown, it's the extraction of the entire weight file that's slow. You can keep the temporary directory extraction, and still extract just one tokenizer file or yaml file to it just as always, which will be a non issue in time. You do not need to modify the current code in it's entirety to achieve your primary goal of extracting particular file type without full extraction of tarfiles to any kind of directory. That is what I meant.

terrykong: > You do not need to modify the current code in it's entirety to achieve your primary goal of extracting particular file type without full extraction of tarfiles to any kind of directory.

If I copy to a temp dir and then copy to the final tmp nemo file dir, I'll copy twice. Are you proposing keeping two copies instead of this refactor which should only copy once b/c it reduces the diff?

PR #9326, opened 2024-05-28
Add ModelOpt QAT example for Llama2 SFT model

kevalmorabia97:
# What does this PR do ?
- Add an example for doing Quantization aware Training (QAT) for Llama SFT model.
- Requires https://github.com/NVIDIA/NeMo/pull/9276 to be merged first

**Collection**: nlp
- `nemo/collections/nlp/parts/nlp_overrides.py` is modified to save/restore model optimizer sharded state to the .nemo checkpoint

# Changelog 
- Add support to save model optimizer's quantization state to the `.nemo` checkpoint (similar support is already merged to Megatron-LM) to save/resume PyTorch Quantization state before TRT-LLM export is done
- Add an example to perform Quantization-aware training (QAT) for the Llama2 SFT Model
- Updated necessary documentation
- Add INT4 QAT CI test 
- Upgrade ModelOpt version from 0.11 to 0.13

# Usage
```bash
python examples/nlp/language_modeling/tuning/megatron_gpt_qat.py \
  quantization.algorithm=fp8 \
  quantization.num_calib_size=8 \
  trainer.devices=1 \
  trainer.num_nodes=1 \
  trainer.max_steps=4 \
  trainer.val_check_interval=4 \
  +trainer.limit_val_batches=2 \
  exp_manager.explicit_log_dir=llama2_qat_results \
  model.restore_from_path=/home/TestData/nlp/megatron_llama/llama_ci.nemo \
  model.tensor_model_parallel_size=1 \
  model.pipeline_model_parallel_size=1 \
  model.global_batch_size=2 \
  model.data.train_ds.file_names=[/home/TestData/nlp/megatron_sft/quarel.jsonl] \
  model.data.train_ds.concat_sampling_probabilities=[1.0]
  model.data.validation_ds.file_names=[/home/TestData/nlp/megatron_sft/quarel.jsonl] \
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [x] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


realAsma:
Can we enable the gradient/activation checkpointing from Nemo trainer? This will make the GPU memory requirement for QAT similar to that of original finetuning.

docs/source/index.rst
```diff
@@ -12,7 +12,7 @@ NVIDIA NeMo Framework is an end-to-end, cloud-native framework designed to build
 - Flash Attention
 - Activation Recomputation
 - Positional Embeddings and Positional Interpolation
-- Post-Training Quantization (PTQ) with Ammo
+- Post-Training Quantization (PTQ)  and Quantization Aware Training (QAT) with TensorRT Model Optimizer
```
realAsma, line 15:
```suggestion
- Post-Training Quantization (PTQ)  and Quantization Aware Training (QAT) with [TensorRT Model Optimizer](https://github.com/NVIDIA/TensorRT-Model-Optimizer)
```

docs/source/nlp/quantization.rst
```diff
@@ -133,15 +132,60 @@ Known issues
...
+Quantization-Aware Training (QAT)
+---------------------------------
 
-References
-----------
+QAT is the technique of fine-tuning a quantized model to recover model quality degradation due to quantization.
+While QAT requires much more compute resources than PTQ, it is highly effective in recovering model quality.
+To perform QAT on a calibrated model from PTQ, you need to further fine-tune the model on a downstream task using a small dataset before exporting to TensorRT-LLM.
+You can reuse your training pipeline for QAT.
+Since QAT is done after PTQ, the supported model families are the same as for PTQ.
```
realAsma, line 142: Can we redirect the users to https://github.com/NVIDIA/TensorRT-Model-Optimizer/tree/main/llm_qat#qat-example-workflow

or copy-paste the example workflow to here?

```suggestion
QAT is the technique of fine-tuning a quantized model to recover model quality degradation due to quantization. During QAT, the scaling factors inside quantizers are frozen and the model weights are fine-tuned.

Please see [QAT Example Workflow](https://github.com/NVIDIA/TensorRT-Model-Optimizer/tree/main/llm_qat#qat-example-workflow) for more details on QAT workflow.
```

docs/source/nlp/quantization.rst
```diff
@@ -133,15 +132,60 @@ Known issues
...
+
+
+Example
+^^^^^^^
+
+The example below shows how to perform PTQ and QAT on a Supervised Finetuned Llama2 7B model to INT4 precision using tensor parallelism of 8 on 8x A5000 24GB GPUs.
```
realAsma, line 148: For ModelOpt OSS QAT, the minimum hardware is 1 x 80GB with activation checkpointing enabled. Can we reduce the compute requirement here as well using activation checkpointing?

docs/source/nlp/quantization.rst
```diff
@@ -133,15 +132,60 @@ Known issues
...
+
+The example is a modified version of the `SFT with Llama 2 playbook <https://docs.nvidia.com/nemo-framework/user-guide/latest/playbooks/llama2sft.html>`_.
+Please refer to the playbook for more details on setting up a BF16 NeMo model and the ``databricks-dolly-15k`` instruction dataset.
+
+First we will run the SFT example command from the playbook as-is to train a Llama2 7B SFT model for 100 steps.
+Make sure to change ``trainer.max_steps=50`` to ``trainer.max_steps=100`` for the ``examples/nlp/language_modeling/tuning/megatron_gpt_finetuning.py`` script.
+This will produce a model checkpoint with validation loss approximately ``1.15`` that we will use for PTQ and QAT next.
+
+For Quantization, we use a modified version of the sft script and config file which includes the quantization and TRT-LLM export support.
+Along with the new parameters, make sure to pass the same parameters you passed for SFT training except the model restore path will be the SFT output ``.nemo`` file.
```
realAsma, line 159: WE recommend using a small learning rate for QAT. See here - https://github.com/NVIDIA/TensorRT-Model-Optimizer/tree/main/llm_qat#qat-example-workflow

docs/source/nlp/quantization.rst
```diff
@@ -133,15 +132,60 @@ Known issues
...
+        trainer.devices=8 \
+        trainer.precision=bf16 \
+        trainer.max_steps=100 \
+        model.restore_from_path=<llama2-7b-sft-nemo-path> \
+        model.global_batch_size=128 \
+        quantization.algorithm=int4 \
+        export.decoder_type=llama \
+        export.inference_tensor_parallel=2 \
+        export.save_path=llama2-7b-sft-int4-qnemo \
+        # other parameters from sft training
```
realAsma, line 175: nit:

Move these to a `launch.sh` script and provide SFT -> PTQ -> QAT workflow as in https://github.com/NVIDIA/TensorRT-Model-Optimizer/tree/main/llm_qat#end-to-end-qat-example

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
```diff
@@ -0,0 +1,203 @@
+name: llama2-7b
+
+trainer:
+  devices: 1
+  accelerator: gpu
+  num_nodes: 1
+  precision: bf16
+  logger: False # logger provided by exp_manager
+  enable_checkpointing: False
```
realAsma, line 9:
```suggestion
  enable_checkpointing: True
```

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
```diff
@@ -0,0 +1,203 @@
...
+  ## Activation Checkpoint
+  activations_checkpoint_granularity: null # 'selective' or 'full'
+  activations_checkpoint_method: null # 'uniform', 'block', not used with 'selective'
+  # 'uniform' divides the total number of transformer layers and checkpoints the input activation
+  # of each chunk at the specified granularity
+  # 'block' checkpoints the specified number of layers per pipeline stage at the specified granularity
+  activations_checkpoint_num_layers: null # not used with 'selective'
+  activations_checkpoint_layers_per_pipeline: null
+  answer_only_loss: True
+  gradient_as_bucket_view: False
```
realAsma, line 73: Can we update the parameters to use activation checkpointing by default?

examples/nlp/language_modeling/tuning/megatron_gpt_qat.py
```diff
@@ -0,0 +1,85 @@
...
+    if cfg.quantization.algorithm is not None:
+        num_batches = cfg.quantization.num_calib_size // cfg.model.global_batch_size
+        forward_loop = get_forward_loop(model.train_dataloader(), num_batches)
+        model = Quantizer.quantize(model, forward_loop, cfg.quantization)
+
+        logging.info("Validating model after PTQ...")
+        trainer.validate(model)
+
+    # Perform QAT on the PTQ Model
+    trainer.fit(model)
```
ChenhanYu, line 78: @kevalmorabia97 , have we checked if `modelopt_state` can be resumed with `model.resume_from_checkpoint` is set?

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
kevalmorabia97: @janekl do you remember why they are explicitly disabled in the Quantizer here: https://github.com/NVIDIA/NeMo/blob/main/nemo/export/quantize/quantizer.py#L179-L180

docs/source/nlp/quantization.rst
kevalmorabia97: I just followed similar to the previous PTQ format in this doc. I think that keeps things simple here as most default parameters are already set in the yaml file

docs/source/nlp/quantization.rst
kevalmorabia97: The way PTQ example was previously authored by @janekl is using a Quantizer class which abstracts usage of `modelopt` from the user so I would prefer to not expose nemo users to this doc as it may overwhelm them. Let me add the note on scaling factors and weight update as you suggested. Users can take a look at the github repository root already linked above and navigate to QAT if they are interested to have a deeper idea.

Wdyt?

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
kevalmorabia97: This needs to be set to False since `exp_manager.create_checkpoint_callback` is already set to True

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
kevalmorabia97: Enabling activation checkpoint works fine for QAT - Uses ~1.5GB less memory per GPU for Llama2-7b and no accuracy regression

docs/source/nlp/quantization.rst
kevalmorabia97: Tried 1e-6 instead of 5e-6 SFT learning rate but the model trains very very slowly, 100 steps only improves loss from 1.31 to 1.30 so I guess we need to use the same LR as the SFT LR likely because SFT LR is already very small

realAsma: The recommendation we give is to use a small learning rate such as "QAT for 1-10% original training duration and a small learning rate, e.g. 1e-5 for Adam optimizer"

In cases such as these SFT LR is already small - QAT happened to work with the original LR. But this does not need to be the case always.

Lets make this clear.  Instead of saying the training parameters are kept the same, can we say we adjust the QAT LR to be a small value.

kevalmorabia97: Updated as follows:

```
As a rule of thumb, we recommend QAT for 1-10% original training duration and a small learning rate, e.g. 1e-5 for Adam optimizer.
If you are doing QAT on an SFT model where learning rates and finetuning dataset size are already small, you can continue using the same SFT learning rate and dataset size as a starting point for QAT.
```

.github/workflows/cicd-main.yml
```diff
@@ -376,7 +376,7 @@ jobs:
         - uses: "NVIDIA/NeMo/.github/actions/cancel-workflow@main"
           if: "failure()"
 
-  L2_PTQ_Llama2_FP8:
+  L2_QAT_Llama2_FP8:
```
janekl, line 379: We should add a new test, not override PTQ, right?

kevalmorabia97:
@realAsma should we disable FP8 KV cache for INT4 weight only?

| FP8 KV Cache | PTQ | QAT |
| --------|-------|-------- |
| Enabled | 1.345 | 1.216 |
| Disabled | 1.305 | 1.172 |

BF16 baseline is 1.15

ChenhanYu:
LGTM;

docs/source/nlp/quantization.rst
realAsma: Lets update the documentation here itself.

realAsma:
> @realAsma should we disable FP8 KV cache for INT4 weight only?
> 
> FP8 KV Cache	PTQ	QAT
> Enabled	1.345	1.216
> Disabled	1.305	1.172
> BF16 baseline is 1.15

That depends on how easy it is to enable KV cache quantization from the config file. If it is easy to do, sure we can enable. However this might not have huge value if there are already good PTQ examples demonstrating KV cache quantization.

kevalmorabia97:
> That depends on how easy it is to enable KV cache quantization from the config file. If it is easy to do, sure we can enable. However this might not have huge value if there are already good PTQ examples demonstrating KV cache quantization.

We currently don't have KV cache quantization parameter in config and its enabled by default for all except int8_sq.

For INT4 having it enabled (default) makes QAT look less appealing as we are still a bit off from bf16 so I though it would be better to disable it.

Let me make if a config parameter and disable it for now -- maybe speedup may be lower

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -381,6 +390,14 @@ def save_checkpoint(
             checkpoint['state_dict'] = OrderedDict([])
 
             self.checkpoint_io.save_checkpoint(checkpoint, ckpt_to_dir(filepath), storage_options=storage_options)
+
+            if HAVE_MODELOPT:
```
ericharper, line 394: I'm not sure I understand this, why do we save the modelopt states just because we have modelopt installed?

kevalmorabia97: All these functions will not do anything if the model has not been modified by modelopt

examples/nlp/language_modeling/tuning/megatron_gpt_qat.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def get_forward_loop(dataloader, num_batches):
+    if len(dataloader) < num_batches:
+        logging.warning(
+            f"Dataloader has fewer batches ({len(dataloader)}) than required ({num_batches}) for calibration."
+        )
+
+    def forward_loop(model):
+        data_iter = iter(dataloader)
```
janekl, line 41: you could maybe use `https://docs.python.org/3/library/itertools.html#itertools.islice` for dataloader to avoid that `i >= num_batches` 2 lines later as a minor simplification

examples/nlp/language_modeling/tuning/megatron_gpt_qat.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
janekl, line 1: Now it would be good to rename megatron_gpt_quantization.py to megatron_gpt_ptq.py :p (separate mr, outside of scope here)

examples/nlp/language_modeling/tuning/megatron_gpt_qat.py
```diff
@@ -0,0 +1,88 @@
...
+        model = quantizer.quantize(model, forward_loop)
+
+        logging.info("Validating model after PTQ...")
+        trainer.validate(model)
+
+    # Perform QAT on the PTQ Model
+    trainer.fit(model)
+
+    # Export the quantized model for TensorRT-LLM inference
+    if cfg.quantization.algorithm != "int4":
```
janekl, line 83: Please add a comment in the code on this condition != "int4" i.e. not supported and why (brefily)

examples/nlp/language_modeling/conf/megatron_gpt_quantization.yaml
```diff
@@ -36,6 +36,7 @@ quantization:
   num_calib_size: 512 # number of samples used for calibration
   awq_block_size: 128 # block size for scaling factors (only used in AWQ algorithms)
   sq_alpha: 1.0 # alpha parameter (only used in SmoothQuant algorithms)
+  enable_kv_cache: null # Enable FP8 KV cache quantization. Set to null for automatic selection.
```
janekl, line 39: sounds more like boolean True/False (minor, we can change that later)

kevalmorabia97: Yes this is boolean but users can also set to null to let quantizer auto select

realAsma:
LGTM!

realAsma (approved):

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
```diff
@@ -0,0 +1,205 @@
...
+  sync_batch_comm: False
+  megatron_amp_O2: True
+
+  ## Sequence Parallelism
+  # Makes tensor parallelism more memory efficient for LLMs (20B+) by parallelizing layer norms and dropout sequentially
+  # See Reducing Activation Recomputation in Large Transformer Models: https://arxiv.org/abs/2205.05198 for more details.
+  sequence_parallel: False
+
+  ## Activation Checkpoint
+  activations_checkpoint_granularity: selective # 'selective' or 'full'
```
realAsma, line 65: nit: 
In Megatron activations_checkpoint_granularity: selective only applies activation checkpointing to certain layers such as attention layer. I run GPU compute limited experiments with `activations_checkpoint_granularity`: `full`

realAsma: Here is my arguments for megatron activation checkpointing: 
```
    --recompute-granularity full \
    --recompute-method uniform \
    --recompute-num-layers 1 \
```
If you decide to use full activation recomputation, you would need to update the equivalent arguments similar to above.

We dont need to address this suggestion. I am very sorry for not pointing this out earlier.

kevalmorabia97: I set it to what the [llama2 SFT example](https://docs.nvidia.com/nemo-framework/user-guide/latest/playbooks/llama2sft.html) suggests for config parameters

janekl (approved):

examples/nlp/language_modeling/conf/megatron_gpt_quantization.yaml
janekl: Right, makes sense!

examples/nlp/language_modeling/tuning/conf/megatron_gpt_qat_config.yaml
```diff
@@ -0,0 +1,205 @@
...
+      monitor: val_loss
+      reduce_on_plateau: false
+
+quantization:
+  decoder_type: ${export.decoder_type} # gptnext, gpt2, llama
+  algorithm: int4 # null, int8_sq, fp8, int4_awq, int4
+  num_calib_size: 512 # number of samples used for calibration
+  awq_block_size: 128 # block size for scaling factors (only used in AWQ algorithms)
+  sq_alpha: 1.0 # alpha parameter (only used in SmoothQuant algorithms)
+  enable_kv_cache: false # Enable FP8 KV cache quantization. Set to null for automatic selection.
```
janekl, line 198: Here you're using `false` which is different from https://github.com/NVIDIA/NeMo/pull/9326/files#r1658879860

kevalmorabia97: Right. For INT4 I was seeing noticeable lm loss drop with kv cache enabled so disabled it for now. Later we plan to change int4 to another algorithm in this example

ericharper (approved):
LGTM. Thanks!

PR #9332, opened 2024-05-28
HF_TOKEN param for the scripts

oyilmaz-nvidia:
# What does this PR do ?

Adds HF TOKEN to the scripts


akoumpa:
Hi, I wonder why passing HF-TOKEN via CLI is required instead of passing it via an environment variable? Can you please clarify for future readers? Thank you.

Issue #8487, opened 2024-02-22
LoRA training with FSDP has spike in train loss


zpx01:
**Describe the bug**

We'd like to use FSDP when doing LoRA fine-tuning with large LLMs. We noticed in our experiments that the train loss when using FSDP vs. no FSDP is very different. With FSDP,  the train loss is extremely high, but after turning off FSDP and doing a regular LoRA tuning, we see normal train loss convergence. We also noted that the outputs of the forward pass when using FSDP are different than those when not using FSDP. Specifically, we observed that the forward pass using FSDP + LoRA seems to have lower magnitude outputs than the forward pass with just LoRA. 

Train Loss Comparison:
<img width="431" alt="Screenshot 2024-02-22 at 3 41 45 PM" src="https://github.com/NVIDIA/NeMo/assets/51731412/da7478a8-0517-4f7d-b0b6-809afcba63ee">

**Steps/Code to reproduce bug**

Please list *minimal* steps or code snippet for us to be able to reproduce the bug.

1. Mount the changes in this [PR](https://github.com/NVIDIA/NeMo/pull/8414) to include the FSDP + LoRA integration. 
2. Use the following script to run an example fine-tuning job:
```
#!/bin/bash
set -x

HOME=/lustre/fsw/coreai_dlalgo_llm/zeeshanp
NEMO=$HOME/NeMo
MEGATRON=$HOME/Megatron-LM

PEFT_SCHEME='lora'
MODEL_SIZE=7b
MBS=1
TP=1
PP=1
NUM_DEVICES=2
GBS=8
SEQ_LEN=4096
DATASET=squad  # dolly, squad

if [ "${PEFT_SCHEME}" == "sft" ]; then
	PEFT_SCHEME_=none
	EXTRA_ARGS="model.optim.name=distributed_fused_adam \ # don't work
                model.optim.lr=5e-6
	           "
else
	PEFT_SCHEME_=${PEFT_SCHEME}
	EXTRA_ARGS=""
fi

EXTRA_ARGS+="
	+model.fp8=False \
	+model.fp8_e4m3=False \
	+model.fp8_hybrid=True \
	+model.fp8_margin=0 \
	+model.fp8_interval=1 \
	+model.fp8_amax_history_len=1024 \
	+model.fp8_amax_compute_algo=max
	"

MODEL=$HOME/pretrained_models/Llama-2-${MODEL_SIZE}-bf16-mcore
[...]

PR #9335, opened 2024-05-29
Fix FSDP gradient calculation with orig params

janEbert:
As observed by some users in #8487, FSDP causes differences in loss. This is because gradients are not correctly calculated when `fsdp_use_orig_params=True`. Currently, any non-`FlatParameter` is treated as being unsharded, when in reality they may be sharded when `use_orig_params=True`. This leads to a double reduction of sharded gradients. We thus simply adjust the parameters that are reduced manually. This requires use of a private variable in the FSDP module. Alternatively, the value of `self.kwargs['ignored_states']` in `nlp_overrides.py` could be set as an attribute on the model to avoid this private variable access.

Thanks to @ofivite for suggesting that `use_orig_params=True` could be the cause of the issue, which greatly helped with analysis.

# What does this PR do ?

Fix FSDP gradient calculation when `fsdp_use_orig_params=True`.

**Collection**: nlp

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to #8487 (issue)

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -999,8 +999,8 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
         """All-reduce gradients of FSDP-sharding-omitted parameters in sharding domain (data-parallel domain)."""
         assert isinstance(self.model, torch.nn.Module)
         grads = []
-        for param in self.model.parameters():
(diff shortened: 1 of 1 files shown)


akoumpa:
Nice, can you fix the DCO?

akoumpa (approved):

janEbert:
Done thanks :)

PR #9402, opened 2024-06-06
TRT-LLM 0.10 Update

oyilmaz-nvidia:
# What does this PR do ?

Update the trt-llm export code to the version 0.10


PeganovAnton:
@oyilmaz-nvidia could you please tell how to install the NeMo in a way that doesn't bring up all the dependencies? I don't see `requirements_export.txt` anymore.

PeganovAnton:
I cannot find `use_inflight_batching`, `enable_context_fmha` parameters of the `export()` method. What is the new way of setting them?

oyilmaz-nvidia:
@PeganovAnton I'll create a release branch for you.

So for the `use_inflight_batching`, `enable_context_fmha`, we no longer get those parameters since trt-llm also changed it. So, as long as the gemm plugin is enabled, remove padding is also true, inflight batching is enabled underneath. Basically, the default parameters should enable the IFB automatically.

meatybobby (approved):
LGTM

PR #9422, opened 2024-06-10
Wrap all fp8 extra states in LocalNonpersistentObject

jbaczek:
# What does this PR do ?

This PR generalizes FP8 extra state wrapping for all tensors.

**Collection**: nlp

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -1850,7 +1850,7 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
 
             # WAR: This is a temporary fix to skip loading FP8 parameters for Dot Product Attention
             def skip_fp8_load(x):
-                if isinstance(x, ShardedObject) and 'fused_attention' in x.key and '_extra_state' in x.key:
+                if isinstance(x, ShardedObject) and '_extra_state' in x.key:
                     x = LocalNonpersitentObject(x.data)  # use the FP8 state from initialization, not from ckpt
                 return x
 


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1850,7 +1850,7 @@ def sharded_state_dict(self, prefix: str = '') -> Dict[str, Any]:
 
             # WAR: This is a temporary fix to skip loading FP8 parameters for Dot Product Attention
             def skip_fp8_load(x):
-                if isinstance(x, ShardedObject) and 'fused_attention' in x.key and '_extra_state' in x.key:
```
mikolajblaz, line 1853: I'm concerned that this PR basically makes the GPT models ignore the FP8 state for all layers in the checkpoint.

In the meantime I prepared a thorough solution, with an almost merged [MCore branch](https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/1628) and a corresponding [NeMo branch](https://github.com/NVIDIA/NeMo/pull/9613).

@jbaczek could you check (in theory or in practice) if this would solve your problem?
The required flag to set would be `model.dist_ckpt_load_sctrictness=log_all`

jbaczek: I think that non-strict loading would solve the problem. I see that this branch is already merged to mcore. When should we expect the sync of repositories, so I could use the public implementation?

mikolajblaz: I asked for a public sync, ideally should be available later today, but we don't have an official ETA

PR #9431, opened 2024-06-10
Add python_requires

galv:
Prevents people from getting unexpected syntax errors when they install on a python version too old.


titu1994 (approved):
Thanks !

pzelasko (dismissed):

PR #9487, opened 2024-06-17
Use Mcore's Modelopt Spec When Exporting

suiyoubi:
# What does this PR do ?

Unified to use mcore's modelopt specs instead of the NeMo one


janekl (approved):

janekl:
Thanks, approved. Just please make sure that CI pipeline passes

janekl:
Hi @suiyoubi, do you think we could get this in? It just needs conflict resolution

janekl:
Hmm I think that @akoumpa extended https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron/gpt_layer_modelopt_spec.py to cover MoE.

So we can't really replace the spec unless we also enable MoE layer spec in https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/inference/gpt/model_specs.py

PR #9512, opened 2024-06-20
[Nemo-UX] IO fixes

marcromeyn:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/lightning/io/fdl_torch.py
```diff
@@ -0,0 +1,116 @@
...
+    return cst.Call(
+        func=cst.Attribute(value=convert_child(torch), attr=cst.Name("tensor")),
+        args=[
+            cst.Arg(convert_child(value.tolist())),
+            py_val_to_cst_converter.kwarg_to_cst("dtype", convert_child(value.dtype)),
+        ],
+    )
+
+
+def enable():
```
akoumpa, line 84: how about `register_pytorch_extensions()`

nemo/lightning/pytorch/strategies.py
```diff
@@ -167,6 +164,21 @@ def connect(self, model: pl.LightningModule) -> None:
...
+        has_optim = getattr(model, "optim", None)
+        if has_optim:
+            opt_config = getattr(model.optim, "config", None)
+            if isinstance(opt_config, OptimizerConfig):
+                mcore_opt_config: OptimizerConfig = cast(OptimizerConfig, opt_config)
+                if not self.ddp_config:
+                    raise ValueError("PyTorch DDP is not enabled for mcore optimizer")
+                ddp_config = cast(DistributedDataParallelConfig, self.ddp_config)
+
+                if mcore_opt_config.use_distributed_optimizer != ddp_config.use_distributed_optimizer:
```
akoumpa, line 176: does that mean we have two entry points for a single option?

marcromeyn: Yeah that's how mcore does it currently, I don't like that design that much. Hopefully we can improve mcore's design around this in a near future release.

nemo/lightning/io/mixin.py
```diff
@@ -321,3 +336,32 @@ def _get_connector(cls, ext, path=None, importer=True) -> ModelConnector:
...
+        encoded_data = base64.b64encode(pickled_data).decode('utf-8')
+        return (encoded_data,), None
+
+    return instance.__io__.__flatten__()
+
+
+def _io_unflatten_object(values, metadata):
+    if len(values) == 1:
+        encoded_data = values[0]
+        pickled_data = base64.b64decode(encoded_data.encode('utf-8'))
```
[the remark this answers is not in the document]
marcromeyn: So this is a fallback scenario for things we don't know how to serialize, we expect to not need this for the big majority of use-cases. I guess we could change the design by not putting the pickled things inside the json but in external binary formatted files?

nemo/lightning/pytorch/strategies.py
akoumpa: Hmm, how about the following: remove the patching here, convert it an assertion & make the ddp_config's use_distributed_optimizer value dependent on the mcore_opt_config? would that be possible>

marcromeyn: Yeah maybe we should remove [this](https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/pytorch/strategies.py#L135-L136) from the constructor and only initialize it here when we detect the mcore optimizer? Then the default code-path is easier to follow, when the user specifies a custom ddp_config we could as you said throw an exception when the values don't match. You think that makes sense?

akoumpa (approved):

PR #9527, opened 2024-06-24
Inflight nemo model export support

JimmyZhang12:
# What does this PR do ?
Adds inflight NeMo to TRTLLM v10 model conversion and engine refitting using device weights

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/export/tensorrt_llm.py
```diff
@@ -317,6 +325,115 @@ def export(
         if load_model:
             self._load()
 
+    def build(
```
oyilmaz-nvidia, line 328: How is this different then the export? It looks similar to export. If yes, can we merge this function with the export function? We might add a couple of parameters to control anything specific with refit.

nemo/export/trt_llm/converter/model_converter.py
```diff
@@ -83,30 +89,40 @@ def model_to_trtllm_ckpt(
...
-    )
-
-    world_size = tensor_parallel_size * pipeline_parallel_size
-
-    has_lm_head = "lm_head.weight" in weights_dict
-    if has_lm_head:
-        lm_head_weight = weights_dict["lm_head.weight"]
+    # If the model has been sharded with model parallelism, convert the model in a gpu-distributed manner
+    if use_distributed_convert:
+        weights_dict = dist_model_to_trt_llm_ckpt(
```
oyilmaz-nvidia, line 94: Is this for the old nemo ckpt type? Old one wasn't called the dist ckpt btw. And we no longer support the old ckpt format.

oyilmaz-nvidia (commented):
Why do you need the support for the old nemo ckpt format?

nemo/export/trt_llm/converter/model_converter.py
jiemingz: This converts an instantiated nemo model that may be sharded with model parallelism on device, and not a checkpoint

oyilmaz-nvidia: You mean this is for the models that are already in the GPU memory?

nemo/export/tensorrt_llm.py
jiemingz: For instance: https://github.com/NVIDIA/NeMo/blob/a973ef529f98c90be5825fac4bcaca869f25e294/nemo/export/tensorrt_llm.py#L243
In export a single rank creates the engines in a loop, but in build all gpus create engines simultaneously (because the distributed environment has already been created)

nemo/export/trt_llm/converter/model_converter.py
jiemingz: Yeah exactly.

jiemingz: It's also why we add new arguments like the vocab size, since we need to unpad the embedding layers, which NeMo pads upon creating the model from the checkpoint.

oyilmaz-nvidia: Can you please add this refit option to this script https://github.com/NVIDIA/NeMo/blob/main/scripts/deploy/nlp/deploy_triton.py and this test https://github.com/NVIDIA/NeMo/blob/main/tests/export/nemo_export.py ? Let me take a look how it works.

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
```diff
@@ -238,6 +239,246 @@ def handle_model_level_weights(model, tp_idx: int, pp_idx: int):
     return weights_dict
 
 
+def _get_layer_index(split_key):
```
oyilmaz-nvidia, line 242: There are too many duplicate code here. Can we reuse some of the existing functions and add parameters and code as needed?

nemo/export/trt_llm/converter/utils.py
```diff
@@ -414,3 +414,124 @@ def split(v, tp_size, idx, dim=0):
         return np.ascontiguousarray(np.split(v, tp_size)[idx])
     else:
         return np.ascontiguousarray(np.split(v, tp_size, axis=dim)[idx])
+
+
+# Similar to split_save_weight but done on GPU for performance
+@torch.no_grad()
+def split_save_weight_gpu(key, val, config):
```
oyilmaz-nvidia, line 421: There are also too many duplicate code here. Can we reuse some of the existing functions and add parameters and code as needed?

oyilmaz-nvidia (changes requested):

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
JimmyZhang12: Can you please carefully review this code? To my knowledge this function is not duplicated anywhere

oyilmaz-nvidia: 1. Didn't mean specifically _get_layer_index function.
2. Also, I don't mean the exact copy of the code by duplicate but rather code with similar logic.

JimmyZhang12: The convention export currently uses ([see here](https://github.com/NVIDIA/NeMo/blob/7fee8e7a0f576317e4113cd58282fa833358c574/nemo/export/trt_llm/converter/utils.py#L179)) is to hardcode index 1. For Nemo models the layer number is not guaranteed (for instance if the module has been wrapped in autocast) so we have unique logic to determine the layer index.

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
```diff
@@ -238,6 +239,246 @@ def handle_model_level_weights(model, tp_idx: int, pp_idx: int):
...
+
+
+def get_layer_num(param_name):
+    split_key = param_name.split(".")
+    layer_index = int(_get_layer_index(split_key))
+    return int(split_key[layer_index])
+
+
+@torch.no_grad()
+def dist_model_to_trt_llm_ckpt(
```
oyilmaz-nvidia, line 263: Can dist_model_to_trt_llm_ckpt function and convert_model_to_trt_llm_ckpt be merged into one function for instance?

oyilmaz-nvidia (approved):
Approving with the condition to work on this again after the release to provide one export path for all use cases.

jiemingz (approved):

PR #9558, opened 2024-06-27
Add tps and pps params to the export script

oyilmaz-nvidia:
# What does this PR do ?

Update tps and pps and depricate n_gpus param in the scripts.


pablo-garay:
CI passed (test that failed flaky & is now fixed): https://github.com/NVIDIA/NeMo/actions/runs/9702936067

scripts/deploy/nlp/deploy_triton.py
```diff
@@ -254,13 +263,14 @@ def get_trtllm_deployable(args):
                 nemo_checkpoint_path=args.nemo_checkpoint,
                 model_type=args.model_type,
                 n_gpus=args.num_gpus,
-                tensor_parallel_size=args.num_gpus,
-                pipeline_parallel_size=1,
+                tensor_parallel_size=args.tensor_parallelism_size,
```
janekl, line 266: How about using the same name for this param everywhere: `args.tensor_parallelism_size` -> `args.tensor_parallel_size` (same for `pipeline_parallel_size` below)

oyilmaz-nvidia: Good suggestion, thanks. Updated it.

janekl (approved):

nemo/export/tensorrt_llm.py
```diff
@@ -253,8 +255,8 @@ def export(
                     max_output_len=max_output_len,
                     max_batch_size=max_batch_size,
                     max_prompt_embedding_table_size=max_prompt_embedding_table_size,
-                    tensor_parallel_size=tensor_parallel_size,
-                    pipeline_parallel_size=pipeline_parallel_size,
+                    tensor_parallel_size=tensor_parallelism_size,
+                    pipeline_parallel_size=pipeline_parallelism_size,
```
janekl, line 259: still different naming is used.. (this is minor though).

I think tensor_parallel_size / pipeline_parallel_size would be more common options

PR #9576, opened 2024-07-01
Add "offline" data cache generation support

dimapihtar:
# What does this PR do ?

1. Adds "offline" data cache generation support. The training script will be stopped once data cache is generated.
2. Enable mcore's `num_dataset_builder_threads` param usage.

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -1662,6 +1663,12 @@ def setup(self, stage=None):
             # Override limit_val_batches to be a multiple of num microbatches to prevent val_step from exiting in between a step
             self._reconfigure_limit_batches(self.trainer.limit_val_batches, self._validation_dl, 'val')
 
+        # Data cache generation only
+        # Stops training after creating a data cache
+        if self.cfg.data.get('data_cache_generation_only', False):
+            self.trainer.num_sanity_val_steps = 0
+            self.trainer.should_stop = True
+
```
athitten, line 1671: @dimapihtar training wouldn't have started at all right to stop it here ? Since we are checking in the setup() method which is run before training begins.

dimapihtar: > @dimapihtar training wouldn't have started at all right to stop it here ? Since we are checking in the setup() method which is run before training begins.

yes, but trainer is already initialized at that moment so it quits the scripts.

dimapihtar: @athitten it quits the script before the first training step.

samskillman:
What is the reasoning behind overloading num_workers for the num_dataset_builder_threads, instead of passing num_dataset_builder_threads in as a separate config parameter? 

I could imagine a case where I set num_dataset_builder_threads to a very large number (64 or 128) but wouldn't want that many workers during the typical data loading. That would allow many workloads to specify a high num_dataset_builder_threads without needing to explicitly use the data_cache_generation_only offline step.

Instead I would suggest num_dataset_builder_threads as an explicit new parameter in megatron_gpt_config.yaml, and default it to 1 to mirror https://github.com/NVIDIA/Megatron-LM/blob/e33c8f78a35765d5aa37475a144da60e8a2349d1/megatron/core/datasets/blended_megatron_dataset_config.py#L48

dimapihtar:
> What is the reasoning behind overloading num_workers for the num_dataset_builder_threads, instead of passing num_dataset_builder_threads in as a separate config parameter?
> 
> I could imagine a case where I set num_dataset_builder_threads to a very large number (64 or 128) but wouldn't want that many workers during the typical data loading. That would allow many workloads to specify a high num_dataset_builder_threads without needing to explicitly use the data_cache_generation_only offline step.
> 
> Instead I would suggest num_dataset_builder_threads as an explicit new parameter in megatron_gpt_config.yaml, and default it to 1 to mirror https://github.com/NVIDIA/Megatron-LM/blob/e33c8f78a35765d5aa37475a144da60e8a2349d1/megatron/core/datasets/blended_megatron_dataset_config.py#L48

thank you for the review, fixed.

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
athitten: Thanks @dimapihtar, the comment `# Stops training after creating a data cache` made me think training would have started. So with `data_cache_generation_only` arg we just cache data and don't run or start any training at all ?

dimapihtar: @athitten yes, exactly. Probably not the best explanation from my side.

samskillman (approved):

athitten (approved):
LGTM, thanks @dimapihtar !

Issue #6257, opened 2023-03-20
Torch Multi Head Attention


grazder:
**Is your feature request related to a problem? Please describe.**

I've seen that you are using custom implementation of MHA block - https://github.com/NVIDIA/NeMo/blob/da6d825b8c2fec0a3b8e0d2c33a7600f242db4c1/nemo/collections/asr/parts/submodules/multi_head_attention.py#L108

I think it would be better to use `torch.nn.MultiheadAttention` which is meant to improve the performance of this block (there is FlashAttention now) and torch continues to maintain and update this block while you're doing basic implementation of it.

**Describe the solution you'd like**

I'd like to see an update of models with torch implementation of MHA

https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html

What do you think about it?

titu1994:
The pytorch MHA implementation is specific to absolutele pos embedding MHA used commonly in NLP. In ASR, we use relative positional embedding for transformer blocks - https://github.com/NVIDIA/NeMo/blob/da6d825b8c2fec0a3b8e0d2c33a7600f242db4c1/nemo/collections/asr/parts/submodules/multi_head_attention.py#L211

If you look at pytorch code, the mask shape is not directly same, but can be made similar (we use BxT1xT2, they use BxHxL1xL2) 

@SeanNaren can you try it to see if it works ?

titu1994:
That's pytorch 2.0, which we don't currently test against but will soon. Even then, it will not be the minimum version for some time so we cannot switch to it right now

desh2608:
Another constraint at the moment is that the implementation using flash attention does not support arbitrary attention masks (https://github.com/pytorch/pytorch/pull/81434#issuecomment-1478759658).

vadimkantorov:
I guess we should also raise this issue in xformers as it might be faster to evolve

SeanNaren:
@vadimkantorov we had an issue in xFormer: https://github.com/facebookresearch/xformers/issues/558

It was closed after support for an additive bias was added directly into xFormer. However we still need to create a model that uses an additive bias as apposed to relative bias/shift to be able to use it unfortunately...

raman-r4334:
Hi @SeanNaren , 

Please let me know if you make `additive bias` work for conformer based models

vadimkantorov:
[...]

PR #9590, opened 2024-07-02
Use torch sdpa implementation in ASR mha

WoodieDudy:
Hola. I changed the mha implementation for the ASR modules so that it uses `torch.nn.functional.scaled_dot_product_attention`.  
This accelerated forward in the mha by 27% and backward by 17% on the A100.   
Pytorch sdpa is continuously being optimized, ensuring that we benefit from the latest performance improvements.  
My code uses [memory efficient backend](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) in sdpa because flash attention doesn't support custom attention bias. There is ongoing work to contribute custom bias support in the flash-attention repository. [PR](https://github.com/Dao-AILab/flash-attention/pull/617).
  
What else do I need to do to merge this pr?  

## Usage
There is also my benchmark:

```python
import torch
import torch.nn as nn
import torch.utils.benchmark as benchmark
from nemo.collections.asr.parts.submodules.multi_head_attention import RelPositionMultiHeadAttention

torch.manual_seed(123)

device = "cuda"
batch_size = 32
seq_len = 1024
d_model = 512
n_head = 8

query = torch.rand(batch_size, seq_len, d_model, device=device, requires_grad=True)
key = torch.rand(batch_size, seq_len, d_model, device=device, requires_grad=True)
value = torch.rand(batch_size, seq_len, d_model, device=device, requires_grad=True)
mask = torch.ones(batch_size, seq_len, seq_len, device=device, requires_grad=False)
mask = torch.triu(mask, diagonal=1).bool() # mask: True - make zero, False - leave unchanged 
mask = None
pos_emb = torch.rand(batch_size, seq_len, d_model, device=device, requires_grad=True)

attention_sdpa = RelPositionMultiHeadAttention(n_head, d_model, 0.0, None, None, use_pytorch_sdpa=True).to(device)
attention_original = RelPositionMultiHeadAttention(n_head, d_model, 0.0, None, None, use_pytorch_sdpa=False).to(device)
for original_param, sdpa_param in zip(attention_original.parameters(), attention_sdpa.parameters()):
    original_param.data.copy_(sdpa_param.data)

# attention_sdpa = torch.compile(attention_sdpa)
# attention_original = torch.compile(attention_original)


def measure_time(attention, query, key, value, mask, pos_emb):
    timer = benchmark.Timer(
        stmt='attention(query, key, value, mask, pos_emb);torch.cuda.synchronize()',
        setup='torch.cuda.synchronize()',
        globals={'attention': attention, 'query': query, 'key': key, 'value': value, 'mask': mask, 'pos_emb': pos_emb}
    )

    with torch.no_grad():
        torch.cuda.synchronize()
        results = timer.blocked_autorange(min_run_time=10)
        forward_time = results.mean
        output = attention(query, key, value, mask, pos_emb)
    return forward_time, output


def measure_fwd_bwd_time(attention, query, key, value, mask, pos_emb):
    timer = benchmark.Timer(
        stmt='loss=attention(query, key, value, mask, pos_emb).sum();torch.cuda.synchronize();loss.backward();torch.cuda.synchronize()',
        globals={'attention': attention, 'query': query, 'key': key, 'value': value, 'mask': mask, 'pos_emb': pos_emb}
    )
    torch.cuda.synchronize()
    results = timer.blocked_autorange(min_run_time=10)
    fwd_bwd_time = results.mean
    return fwd_bwd_time


time_fwd_original, output_original = measure_time(attention_original, query, key, value, mask, pos_emb)
time_fwd_sdpa, output_sdpa = measure_time(attention_sdpa, query, key, value, mask, pos_emb)

print(f"Original implementation time: {time_fwd_original:.6f} seconds")
print(f"SDPA implementation time: {time_fwd_sdpa:.6f} seconds")
print(f"SDPA boost {(time_fwd_original - time_fwd_sdpa) / time_fwd_original * 100:.2f}%")

time_fwd_bwd_original = measure_fwd_bwd_time(attention_original, query, key, value, mask, pos_emb)
time_fwd_bwd_sdpa = measure_fwd_bwd_time(attention_sdpa, query, key, value, mask, pos_emb)
time_bwd_original = time_fwd_bwd_original - time_fwd_original
time_bwd_sdpa = time_fwd_bwd_sdpa - time_fwd_sdpa

print(f"Original implementation backward time: {time_bwd_original:.6f} seconds")
print(f"SDPA implementation backward time: {time_bwd_sdpa:.6f} seconds")
print(f"SDPA backward boost {(time_bwd_original - time_bwd_sdpa) / time_bwd_original * 100:.2f}%")

print(f"Outputs are {'the same' if torch.allclose(output_original, output_sdpa, atol=1e-5) else 'different'}")

# Original implementation time: 0.049075 seconds
# SDPA implementation time: 0.035598 seconds
# SDPA boost 27.46%
# Original implementation backward time: 0.127004 seconds
# SDPA implementation backward time: 0.104986 seconds
# SDPA backward boost 17.34%
# Outputs are the same
```

  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

cc @titu1994 @SeanNaren

## Additional Information
* Related to [(issue)](https://github.com/NVIDIA/NeMo/issues/6257)


WoodieDudy:
I also attempted to run the tests in the repository but encountered an issue. NaNs appear when a mask with a fully False row is passed to MHA. Because of such mask, filling the `matrix_bd` with `-inf` values using `matrix_bd.masked_fill_(mask.logical_not(), float("-inf"))` results in a row of only `-inf`, and after the softmax, this entire row becomes NaNs. I am unsure how to resolve this since the softmax and multiplication by value occur within `torch.nn.functional.scaled_dot_product_attention`, and I cannot intervene. In your implementation, this is handled by manually filling with zeros `attn = torch.softmax(scores, dim=-1).masked_fill(mask, 0.0)`.

What can be done about this? Should we write a separate test that doesn't use such masks as input?

vadimkantorov:
@SeanNaren @titu1994 An option of using SDPA is a good thing also because a Triton-based version of FAv2 with custom `attn_bias` support is being added into PyTorch core: https://github.com/pytorch/pytorch/pull/130250#issuecomment-2216318222, so Conformer attention can benefit in the future from the speed-ups and proper compilation of SDPA in core PyTorch developments

vadimkantorov:
@SeanNaren @titu1994 haha, and now that FAv3 is out, probably PyTorch would integrate it as well in some near term - for maximum brr on H100 :) so having Nemo's Conformer auto-benefitting from this new work would be awesome

WoodieDudy:
cc @redoctopus @jbalam-nv @okuchaiev

vadimkantorov:
stale bump

VahidooX:
@titu1994 please take a look at this PR as it looks like an interesting addition to speedup conformer models.

VahidooX:
Thanks for the contribution!
@titu1994 please take a look at this PR as it looks like an interesting addition to speedup conformer models.

Just some notes:

+ Please use -10000 instead of -inf if it is possible as -inf may cause NAN with some data types.

+ Please add it as a config to the config files somewhere like here to be able to control it from configs:
https://github.com/NVIDIA/NeMo/blob/07d536babf25f34d2866e7fa76c05eeef3b0086b/examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml#L147
Name is "use_pytorch_sdpa"?

+ I suggest to make it True as default if we can make sure it works in all cases? @titu1994 what do you think?

+ Please evaluate one of the pretrained models on NGC on a test-other LS to make sure that it produces the same exact output and accuracy.

+ You need to set the dropout to zero manually in non-training model as sdpa does not respect that and it always uses the dropout.

titu1994 (commented):
The PR looks great! Sorry for the delay in review. 

I would be ok with making the default use torch sdpa as True only if we can add a test that runs this function two times - setting the flag to true and false and comparing the output to have 1e-4 or lower mse difference. 

Could you add your example code as a unit test somewhere as a unit test ?

WoodieDudy:
Thanks @titu1994 
Okey, I'll try to add tests.

nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -1038,13 +1044,15 @@ def change_attention_model(
                         att_context_size=att_context_size,
                         pos_bias_u=None,
                         pos_bias_v=None,
+                        # use_pytorch_sdpa=self.use_pytorch_sdpa,
```
nithinraok, line 1047: why is this commented, what is effect of sdpa for LongFormer Attention?

nithinraok (approved):
Thanks much for the PR! 

LGTM, minor question. 

I evaluated the PR on HF-Leaderboard datasets, observed no difference in WER. And for LS test set on A6000, improved RTFx by 5%

examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml
```diff
@@ -145,6 +145,7 @@ model:
     xscaling: true # scales up the input embeddings by sqrt(d_model)
     untie_biases: true # unties the biases of the TransformerXL layers
     pos_emb_max_len: 5000
+    use_pytorch_sdpa: true
```
VahidooX, line 148: Please add one line comment on what it does in the config.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -240,18 +270,32 @@ def forward(self, query, key, value, mask, pos_emb, cache=None):
...
-            matrix_bd = matrix_bd[:, :, :, : matrix_ac.size(-1)]
 
-            scores = (matrix_ac + matrix_bd) / self.s_d_k  # (batch, head, time1, time2)
+            if self.use_pytorch_sdpa:
+                scale_factor = 1 / math.sqrt(q_with_bias_u.size(-1))
+                matrix_bd = matrix_bd[:, :, :, : k.size(-2)] * scale_factor
+
+                if mask is not None:
+                    mask = mask.unsqueeze(1)
+                    matrix_bd.masked_fill_(mask, -10000.0)
```
VahidooX, line 285: Is this necessary to mask out the matrix_bd here?

examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml
VahidooX: Also let's add it option to the ctc config too.

VahidooX (changes requested):

examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml
pzelasko: is it necessary? it's already turned on by default in `__init__`

nemo/collections/asr/modules/conformer_encoder.py
WoodieDudy: I didnt add torch sdpa in longformer. I would do it in new pr i guess

nemo/collections/asr/parts/submodules/multi_head_attention.py
WoodieDudy: Yes, because below in the code I use `matrix_bd` as `attn_mask`
```python
out = torch.nn.functional.scaled_dot_product_attention(
    q_with_bias_u, k, v, attn_mask=matrix_bd, dropout_p=dropout_rate, scale=scale_factor
)

File: examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml (hunk 2 above)
Comment (VahidooX) [reply examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml:148]: Let's have it here at least to make it exposed to users otherwise it would be hidden.

File: nemo/collections/asr/modules/conformer_encoder.py (hunk 1 above)
Comment (VahidooX) [reply nemo/collections/asr/modules/conformer_encoder.py:1047]: Would you please raise an error when use_pytorch_sdpa is enabled?

**Comment by VahidooX** [comment]
> I'm working with tests and faced a problem with `tests/collections/asr/test_conformer_encoder.py:test_stochastic_depth_forward`. But I think that in this test, the data in `random_length` is incorrect, because `random_length` must one dim with shape `batch`, not two dims. Am I right?
> 
> https://github.com/NVIDIA/NeMo/blob/6774fdcf23a915421135b223404b2213bf2c1b7a/tests/collections/asr/test_conformer_encoder.py#L84

Yes, that looks like to be incorrect.

**Review by nithinraok — DISMISSED** [review]

File: examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml (hunk 2 above)
Comment (WoodieDudy) [reply examples/asr/conf/fastconformer/fast-conformer_transducer_bpe.yaml]: done

File: nemo/collections/asr/modules/conformer_encoder.py (hunk 1 above)
Comment (WoodieDudy) [reply nemo/collections/asr/modules/conformer_encoder.py]: done

**Comment by WoodieDudy** [comment]
As I mentioned above [here](https://github.com/NVIDIA/NeMo/pull/9590#issuecomment-2205160174), I encountered an error caused by the mask for padding tokens being all False, which led to NaNs appearing after the softmax operation.

I resolved this issue by adding dummy columns and rows to the query, key, value, and mask https://github.com/NVIDIA/NeMo/blob/8d292d6ef218d5dc5f40878cde2052448e08aabb/nemo/collections/asr/parts/submodules/multi_head_attention.py#L160-L162
However, this may not be the most appropriate solution since it could affect the divisibility of dimensions by powers of two.

Do you think this approach is acceptable?

Recently, a similar issue was resolved for SDPA https://github.com/pytorch/pytorch/pull/131863. I guess when this fix is included in the global version of PyTorch, we can remove this workaround involving the additional columns.

**Comment by WoodieDudy** [comment]
Maybe we should run a training test with mem efficient backend before setting use_pytorch_sdpa=True by default. Does anyone have a ready script for this?

**Review by titu1994 — DISMISSED** [review]
This is fantastic, thanks for adding the equivalence checks between sdpa and the legacy code. Ready to merge

**Comment by titu1994** [comment]
> As I mentioned above [here](https://github.com/NVIDIA/NeMo/pull/9590#issuecomment-2205160174), I encountered an error caused by the mask for padding tokens being all False, which led to NaNs appearing after the softmax operation.
> 
> I resolved this issue by adding dummy columns and rows to the query, key, value, and mask
> 
> https://github.com/NVIDIA/NeMo/blob/8d292d6ef218d5dc5f40878cde2052448e08aabb/nemo/collections/asr/parts/submodules/multi_head_attention.py#L160-L162
> 
> 
> However, this may not be the most appropriate solution since it could affect the divisibility of dimensions by powers of two.
> Do you think this approach is acceptable?
> 
> Recently, a similar issue was resolved for SDPA [pytorch/pytorch#131863](https://github.com/pytorch/pytorch/pull/131863). I guess when this fix is included in the global version of PyTorch, we can remove this workaround involving the additional columns.

Hmm, did you try to use -10000 instead of -inf and then slice off the excess as suggested in the above pr as option 2? 

If that doesn't work, I'm wondering if we should merge this without the WAR and guard this code path for only after pytorch 2.4 is released ? We can do a version check and run the correct code for relposmha. 

We can fall back to older codebase for lower pytorch versions. @VahidooX what do you think ?

File: nemo/collections/asr/parts/submodules/multi_head_attention.py [hunk 4]
```diff
@@ -240,18 +279,43 @@ def forward(self, query, key, value, mask, pos_emb, cache=None):
...
+                    matrix_bd = torch.cat([matrix_bd, new_column.unsqueeze(-1)], dim=-1).to(matrix_bd.device)
+
+                dropout_rate = self.dropout_rate if self.training else 0
+                # add extra col for key and value to handle problem with nan after solfmax
+                extra_column = torch.zeros(k.shape[:-2] + (1, k.shape[-1])).to(k.device)
+                k = torch.cat([k, extra_column], dim=-2)
+
+                extra_column = torch.zeros(v.shape[:-2] + (1, v.shape[-1])).to(v.device)
+                v = torch.cat([v, extra_column], dim=-2)
+                out = torch.nn.functional.scaled_dot_product_attention(
```
nithinraok, line 308: I just noticed: isn;t this different from `matrix_ac + matrix_bd` for computation of attention scores? How are both equal?

WoodieDudy: Here is the SDPA code from the [pytorch documentation](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html).
I passed `matrix_bd` as `attn_mask`, which then becomes `attn_bias`:
```python
attn_bias = torch.zeros(L, S, dtype=query.dtype)
attn_bias += attn_mask
```

`attn_weight` is similar to `matrix_ac`:
```python
attn_weight = query @ key.transpose(-2, -1) * scale_factor
```

This line `attn_weight += attn_bias` is equivalent to `matrix_ac + matrix_bd`:
```python
# Efficient implementation equivalent to the following:
def scaled_dot_product_attention(query, key, value, attn_mask=None, dropout_p=0.0, is_causal=False, scale=None) -> torch.Tensor:
    L, S = query.size(-2), key.size(-2)
    scale_factor = 1 / math.sqrt(query.size(-1)) if scale is none else scale
    attn_bias = torch.zeros(L, S, dtype=query.dtype)
    if is_causal:
        assert attn_mask is none
        temp_mask = torch.ones(L, S, dtype=torch.bool).tril(diagonal=0)
        attn_bias.masked_fill_(temp_mask.logical_not(), float("-inf"))
        attn_bias.to(query.dtype)

    if attn_mask is not none:
        if attn_mask.dtype == torch.bool:
            attn_bias.masked_fill_(attn_mask.logical_not(), float("-inf"))
        else:
            attn_bias += attn_mask
    attn_weight = query @ key.transpose(-2, -1) * scale_factor
    attn_weight += attn_bias
    attn_weight = torch.softmax(attn_weight, dim=-1)
    attn_weight = torch.dropout(attn_weight, dropout_p, train=True)
    return attn_weight @ value
```

WoodieDudy:
So, @titu1994 , yes, it was possible to get rid of these NaNs after softmax by simply zeroing them out using a mask, which is what I just did.
```python
out = torch.nn.functional.scaled_dot_product_attention(q, k, v, attn_mask=mask, dropout_p=dropout_rate)

# this IF block can be deleted when https://github.com/pytorch/pytorch/pull/131863 is in the stable version
if mask is not None:
    all_masked_rows = torch.all(~mask, dim=-1)
    all_masked_rows.unsqueeze_(-1)
    out = out.masked_fill(all_masked_rows, 0.0)
```

In the end, it seems that before merging with the flag set to True by default, it would be good to run a training test to ensure everything works fine. Does anyone have the code ready for this?

pzelasko:
I'm on it - launched Canary-1B training with this, will keep you posted.

pzelasko:
Unfortunately the Canary-1B training on 32 GPUs with bf16 AMP is diverging after ~10 hours with SDPA. I launched a second run with a different seed to confirm and it replicates. I have a ~1 month old baseline run that works fine, and I launched another from today's main that also runs fine. Any ideas what may have caused this?

vadimkantorov:
@pzelasko I think the first step should be running SDPA with the most basic `math` algo selected:
```python
with torch.backends.cuda.sdp_kernel(enable_math=True, enable_flash=False, enable_mem_efficient=False, enable_cudnn=False):
   x = F.sdpa(...)
```

This would validate/benchmark that the basic SDPA impl works fine (should be more similar to the original Nemo's impl). I heard somewhere on PyTorch forum/issues that numerics of different SDPA methods are slightly different and this can lead to blow-ups during internal accumulation :(

orena1:
@pzelasko what pytorch version did you use? We found that the latest version where more stable.
Did you try with bf16_mixed?

WoodieDudy:
So, let's make `use_pytorch_sdpa` False by default and merge this PR? And then may be continue work in different PR

nithinraok:
That sounds fine with me. Can you make it False by default. @titu1994 wdyt?

titu1994:
Sounds ok to me

WoodieDudy:
I set the flag everywhere to false by default. And also added an argument `use_pytorch_sdpa_backends` in which you can set the list of backends for sdpa

WoodieDudy:
Let's merge? @titu1994

nithinraok (dismissed):
LGTM! @VahidooX @titu1994 @pzelasko for final review

titu1994:
It looks ok for inference at least, lets avoid if for training for now

WoodieDudy:
@VahidooX @pzelasko 👀

pzelasko:
Uh, sorry for the lack of responsiveness lately. I used bf16-mixed and PyTorch 2.4.0 with CUDA 12.5. I agree we should merge it disabled by default. Let's take a look again later with the newer backends.

pzelasko (approved):

nithinraok:
Just initiated the CI run: looks like some of them are failing like this: https://github.com/NVIDIA/NeMo/actions/runs/11128061705/job/30922390803?pr=9590 

address them and once CI passes this is good to go. Thanks!

examples/asr/conf/fastconformer/fast-conformer_ctc_bpe.yaml
```diff
@@ -136,6 +136,8 @@ model:
     xscaling: true # scales up the input embeddings by sqrt(d_model)
     untie_biases: true # unties the biases of the TransformerXL layers
     pos_emb_max_len: 5000
+    use_pytorch_sdpa: false # use torch sdpa instead of manual attention
+    use_pytorch_sdpa_backends: [torch.nn.attention.SDPBackend.MATH] # empty list means all backends https://pytorch.org/docs/stable/generated/torch.nn.attention.SDPBackend.html
```
VahidooX, line 140: Please set the backend list to empty.

nemo/collections/asr/parts/submodules/multi_head_attention.py
```diff
@@ -48,6 +49,8 @@
     'PositionalEncoding',
 ]
 
+inf_val = 10000.0
```
VahidooX, line 52: Please use "INF_VAL".

VahidooX (approved):
LGTM! Just left some minor comments.

WoodieDudy:
So, the CI is using torch==2.3.0a0+ebedce2 which has a bit different way to setting sdpa backends. What a good way to deal with it? Add condition on pytorch version in a MultiHeadAttention `__init__` and map backend names to bools
```
if backend == MATH:
    enable_math = True
if backend == ...:
    enable_mem_efficient = True
with torch.backends.cuda.sdp_kernel(enable_math=enable_math, enable_mem_efficient=enable_mem_efficient):
```

or update CI torch version?

pzelasko:
I suggest adding the condition so that NeMo can be used with earlier PyTorch versions.

nemo/collections/asr/parts/submodules/multi_head_attention.py
WoodieDudy: Done

examples/asr/conf/fastconformer/fast-conformer_ctc_bpe.yaml
WoodieDudy: Done

WoodieDudy:
@nithinraok can you please rerun CI. I think I fixed import issue

WoodieDudy:
Is it okey that you are using no-release pytorch version in CI i.e. `2.3.0a0+ebedce2`. It doesn't have all the features of 2.3.0

vadimkantorov:
@pzelasko If possible, I would propose to simply upgrade the CI to use the release version of PyTorch 2.3.0, it would make the life much easier for all these SDPA control knobs

ko3n1g:
> @pzelasko If possible, I would propose to simply upgrade the CI to use the release version of PyTorch 2.3.0 (currently the CI seems to use some alpha version of 2.3.0), it would make the life much easier for all these SDPA control knobs

We just (yesterday) upgraded the CI to [24.07](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel-24-07.html) which maybe already solves the issue?

WoodieDudy:
@nithinraok can you please rerun CI

WoodieDudy:
Let's merge

nithinraok:
Rerunning CI as lot of ASR tests failing

WoodieDudy:
Are the failed tests unrelated to the changes in this PR?

ko3n1g:
The relevant test `CICD NeMo / Nemo_CICD_Test (pull_request) ` has passed so once this has approvals it can get merged

WoodieDudy:
Merge?

nithinraok (approved):

PR #9593, opened 2024-07-02
Set n_gpu to None in nemo export

oyilmaz-nvidia:
# What does this PR do ?

Change ngpu to none to use tensor_parallelism_size param.


thomasdhc (approved):
LGTM

PR #9641, opened 2024-07-08
enables default data step in megatron parallel to operate on a wider variety of tensors

jomitchellnv:
…variety of tensors coming out of the dataloader

Add more flexibility to the default data step inside megatron parallel.

# Changelog 
- Megatron parallel's default data step can operate on wrapped tuples and unwrapped items coming directly out of the dataloader

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/lightning/megatron_parallel.py
+++ b/nemo/lightning/megatron_parallel.py
@@ -25,9 +25,11 @@ from typing import (
 
 import torch
 import torch.distributed
+from megatron.core import parallel_state
 from megatron.core.distributed import DistributedDataParallel as McoreDDP
 from megatron.core.distributed import DistributedDataParallelConfig
 from megatron.core.transformer.transformer_config import TransformerConfig
+from pytorch_lightning.utilities import move_data_to_device
 from torch import Tensor, nn
 from typing_extensions import override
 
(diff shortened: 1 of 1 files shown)


nemo/lightning/megatron_parallel.py
```diff
@@ -43,15 +45,37 @@
...
+    match next(dataloader_iter):
+        # If its wrapped in a tuple, unpack it.
+        case (batch, int(_), int(_)):
+            pass
+        # Canonical case.
+        case batch:
+            pass
 
-    return batch
+    return move_data_to_device(batch, torch.cuda.current_device())
```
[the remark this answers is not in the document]
jomitchellnv: @jstjohn  this might make it tricky to use match

nemo/lightning/megatron_parallel.py
```diff
@@ -43,15 +45,43 @@ def convert_output(self, output: torch.Tensor) -> torch.Tensor: ...
...
+        )
+
+    match next(dataloader_iter):
+        # If its wrapped in a tuple, unpack it.
+        case (batch, int(_), int(_)):
+            pass
+        # Canonical case.
+        case batch:
+            pass
+        # If the dataloader_iter is empty, return None.
```
jstjohn, line 77: This comment is not what happens anymore. Now in your `else` you say `raise ValueError(...):`. Are we sure we don't want to pass through a None on the iter? Otherwise we should prob change line 79 or 84 to `return None` and drop the `ValueError` piece.

jstjohn: Actually what I hope would happen is that the `match(dataloader_iter)` would trigger an `StopIteration` or whatever it is exception rather than returning None. The StopIteration exception ideally would raise through this function (what happens now on the match fail) and be handled elsewhere. So I _think_ your case _ should never get hit...

jstjohn: Maybe a better error message would be ``None` batch returned at the end of iteration rather than a `StopIteration` exception being raised, why?`

marcromeyn (approved):

nemo/lightning/megatron_parallel.py
jomitchellnv: So I only added this to be past the linter. I updated the comment so that its more accurate. My only concern would be how it plays out with drop_last=True since maybe that would dump some None's?

marcromeyn (approved):

PR #9689, opened 2024-07-11
NeVa::forward - remove device syncs (torch.where) and vectorize over batch dimensions

nikitaved:
The following patch modifies the forward of the NeVa model (the `replace_media_embeddings` method) by introducing the following improvements:

- Elimination of `torch.where` with a single argument. That call introduces device synchronizations when run on GPUs.
- Media index selection manipulation in the construction of `padded_media_indices` can be done  in a single vectorized call.
- Reduction in the overall number of kernel launches.
- Elimination of calls to `all`. This also causes device syncs. We can leave it, or run it in debug mode only. It is best to have a dedicated test instead of that assert, imho.

### DIFF
--- a/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py
+++ b/nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py
@@ -260,22 +260,32 @@ class NevaWordEmbeddingMixin(torch.nn.Module, adapter_mixins.AdapterModuleMixin)
         # flatten patches
         media_features = media_features.view(batch_size, -1, hidden_size)
 
-        # create an indices matrix used in torch.scatter
-        padded_media_indices = torch.ones(
-            (batch_size, num_images_per_sample), dtype=torch.long, device=input_ids.device
+        # create an indices matrix used in torch.scatter {
+        sorted_media_end_positions_mask, media_end_positions_mask_sort_idx = (
+            # NOTE: to(torch.long) is needed because PyTorch does not have sort for boolean tensors on CUDA
+            (input_ids == self.media_end_id).to(torch.long).sort(dim=-1, descending=True, stable=True)
         )
-        padded_media_indices *= sequence_length
-        for idx, input_id in enumerate(input_ids):
-            media_end_positions = torch.where(input_id == self.media_end_id)[0]
-            if self.use_im_start_end:
-                # locate the first media token positions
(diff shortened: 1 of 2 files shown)


nemo/collections/multimodal/models/multimodal_llm/neva/neva_model.py
```diff
@@ -232,22 +232,32 @@ def replace_media_embeddings(self, input_ids, inputs_embeds, media):
...
+            )
+        else:
+            padded_media_indices = torch.where(
+                sorted_media_end_positions_mask.to(torch.bool),
+                media_end_positions_mask_sort_idx - num_patches + 1,
+                sequence_length
+            )
+
+        # TODO: add asserts that were here below
+        # }
```
nikitaved, line 262: The assert I mean the call to `all`. These also cause device synchronizations. They are better be reserved for runs in a some kind of debug mode. It is best to have a dedicated test though. Where would one place such a test?

yaoyu-33: @nikitaved  can you add condition if env var `NEMO_TESTING` is True, still run assert?

nikitaved: @yaoyu-33 , done! Is there a way to get some CI output?

nikitaved:
@yaoyu-33 , could you please have another look and let the workflow be run?

tfogal:
FWIW, I ran our little test (a `neva_pretrain.py` invocation) with `NEMO_TESTING=1` and things ran fine there.

yaoyu-33 (approved):

nikitaved:
@yaoyu-33 , @ericharper , could you please run the workflow again? Thank you!

tfogal:
bump to kill 'stale'.
i checked that i have access to an internal cluster today, will check convergence soon

nikitaved:
Thanks you, @tfogal ! FWIW, the very last CI run turned out to be green

nikitaved:
Is it possible to move it forward? Commenting to remove the current stale status.

tfogal:
> Remove stale label or comment or update or this will be closed in 7 days.

comment

nikitaved:
"
Wait it out
Gonna wait it out
Wait it out (be patient)
" Patient, Tool

nikitaved:
Is there interest in reviving this PR? I assume it is rather low impact with perf benefits assuming the tests are green. And green they were the last time CIs were run.

PR #9722, opened 2024-07-12
Set TE flag in legacy -> mcore conversion script

terrykong:
Same as #9585, but into release branch


ericharper (approved):

cuichenx:
#9585 was merged to the main branch two weeks ago so the release branch should already have the commit?

terrykong:
@cuichenx You're right, I'm actually surprised github said we could make this MR. When I look at the `0ec7e90` commit history for that conversion script it actually doesn't even show this commit. So the git history on the conversion script looks okay:

image

PR #9726, opened 2024-07-14
Add option to convert PyTriton response to OpenAI format

athitten:
# What does this PR do ?

1. Adds option to convert responses from PyTriton server to OpenAI compatible format. 
2. Also adds `triton_request_timeout` and `openai_format_response` as user defined args which are then written to a json file along with `triton IP` and `triton port` to be accessed by `rest_model_api.py` (Deletes the existing `nemo/deploy/service/config.json` favor of these changes)

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/deploy/nlp/query_llm.py
+++ b/nemo/deploy/nlp/query_llm.py
@@ -15,6 +15,7 @@
 from abc import ABC, abstractmethod
 
 import numpy as np
+import time
 
 from nemo.deploy.utils import str_list2numpy
 
@@ -172,6 +173,7 @@ class NemoQueryLLM(NemoQueryLLMBase):
         compute_logprob: bool = None,
         end_strings=None,
         init_timeout=60.0,
+        openai_format_response: bool = False
     ):
         """
         Query the Triton server synchronously and return a list of responses.
@@ -259,7 +261,20 @@ class NemoQueryLLM(NemoQueryLLMBase):
                     return "Unknown output keyword."
 
                 sentences = np.char.decode(output.astype("bytes"), "utf-8")
-                return sentences
+                if openai_format_response:
+                    openai_response = {
+                        "id": f"cmpl-{int(time.time())}",
+                        "object": "text_completion",
+                        "created": int(time.time()),
+                        "model": self.model_name,
+                        "choices": [{
+                            "text": str(sentences)
+                        }
+                        ]
+                    }
+                    return openai_response
+                else:
+                    return sentences
             else:
                 return result_dict["outputs"]
 
--- a/nemo/deploy/nlp/query_llm.py
+++ b/nemo/deploy/nlp/query_llm.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
 from abc import ABC, abstractmethod
 
 import numpy as np
-import time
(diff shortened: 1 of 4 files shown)


nemo/deploy/service/rest_model_api.py
```diff
@@ -66,6 +76,23 @@ class CompletionRequest(BaseModel):
     frequency_penalty: float = 1.0
 
 
+@app.get("/triton_health")
+async def check_triton_health():
+    """
```
athitten, line 81: Added a test method `check_triton_health` for users to verify if Pytriton server is up and running since many times while using REST service, we encounter an error if Pytriton server is not accessible. This test method adds a check for that.

scripts/deploy/nlp/deploy_triton.py
```diff
@@ -73,6 +74,9 @@ def get_args(argv):
     parser.add_argument(
         "-tha", "--triton_http_address", default="0.0.0.0", type=str, help="HTTP address for the Triton server"
     )
+    parser.add_argument(
+        "-trt", "--triton_request_timeout", default=60, type=int, help="Timeout in seconds for Triton server"
+    )
```
athitten, line 79: Moved the args  `triton_request_timeout` from `nemo/service/deploy/config.json` to the argparser of `deploy_triton.py` so that users can pass it. Also added `openai_format_response` as an user defined option while running `deploy_triton.py` with REST service True

athitten: Also deleted `config.json` as triton IP, triton port, triton timeout args are passed by the user while running `deploy_triton.py` and they are stored in a json file when `deploy_triton.py` is run which are then accessed in `rest_model_api.py` via the stored config.json file.

nemo/deploy/nlp/query_llm.py
```diff
@@ -259,7 +261,17 @@ def query_llm(
                     return "Unknown output keyword."
 
                 sentences = np.char.decode(output.astype("bytes"), "utf-8")
-                return sentences
+                if openai_format_response:
```
oyilmaz-nvidia, line 264: Do you think we can do this openai response conversion in the here https://github.com/NVIDIA/NeMo/blob/athitten/openai_format_response/nemo/deploy/service/rest_model_api.py#L101 ?

Also, is there any advantage of supporting the openai response type at the class level? Maybe our default should be the openai response format (still not sure though).

athitten: @oyilmaz-nvidia yes we can do the openAI response conversion in `rest_model_api.py` after output is received [here](https://github.com/NVIDIA/NeMo/blob/athitten/openai_format_response/nemo/deploy/service/rest_model_api.py#L101). Do you recommend that ?

Also regarding having OpenAI format as the default, do any of the methods/functions in the deploy module or any downstream tasks rely on output being returned directly (non OpenAI format) i.e the way it is now ? I wasn't sure of this so added it as an option to not break anything.

oyilmaz-nvidia (approved):
LGTM.

PR #9763, opened 2024-07-17
OOMptimizer: bucketing batch size profiles to make GPUs go 🔥

pzelasko:
# What does this PR do ?

Major contributions:
* Canary-1B can be trained with 5x larger batch sizes compared to our earlier baseline. It maxes out GPU utilization (memory, compute, and power consumption wise). As a result the mean training step time is 2.75x longer, resulting in a training throughput of 5x / 2.75x ~= 180% of the original recipe. I managed to reproduce Canary-1B in about 40k training steps on the same number of GPUs, changing only bucketing/batch size settings using new features in this PR.
    * Update: actually reproduces Canary-1B in half of the training time with slightly improved WERs.
    * Update 2: also reproduced Canary-1B in the original training time using 4x less GPUs.
* Note: these tools are applicable to all ASR models and can easily be made applicable to any (audio|text)->(audio|text) model.
* OOMptimizer script that given a model config and bucket bins, finds the optimal batch sizes for each bucket bin. Optimal = maximum GPU utilization.
* 2D bucketing with a dedicated estimation script and dataloading support. Allows to stratify sampling by input and output sequence lengths, resulting in improved training throughput for encoder-decoder models.
* Ability to filter out examples exceeding a tokens-per-second threshold during training (e.g. some datasets have very severe outliers, 20x more tokens than median).
* Enables concurrent bucketing, speeding up the start of the training loop (in my experiments reduces the wait time from ~2min to ~20 seconds). With this setting, the bucketing buffer is filled asynchronously and lets the sampler draw batches when it's at least 10% filled, so the training can start faster.
* Documentation and examples of usage of the new features.
* Unit and integration for the new features.

**Collection**: ASR

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,283 @@
...
+    1) Run scripts/speech_recognition/estimate_duration_bins.py to get the duration distribution of your data.
+    2) Run OOMptimizer to find the optimal batch sizes for your specific model, optimizer, and GPU.
+    3) Use these optimal settings in your actual training script and enjoy optimal GPU utilization OOM-free.
+    """
+    logging.setLevel(logging.CRITICAL)
+    device = "cuda"
+
+    print("Intializing ASR model.")
+    # TODO(pzelasko): This currently only supports "from_pretrained".
+    #                 We need to be able to read a model training configuration and instantiate the model
```
titu1994, line 206: You can use restore_from(..., return_config=True)...

titu1994 (dismissed):
Very interesting approach to maximize the batch size, minor comments but it looks good.

pzelasko:
I realized we can make a post-processing pass on the max_batch_size list and merge buckets with identical batch sizes. Merging buckets will improve randomization. If the original num_buckets was large enough to trigger merging, that approach will lead us to an optimal number of buckets.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,283 @@
...
+        print(f"The current sequence lengths are: input={seq_len_in} output={seq_len_out}.")
+        gen.reset()
+        batch_idx = 0
+
+        def step():
+            batch = gen(seq_len_in, seq_len_out)
+            oom = False
+            try:
+                print(f"Current gap: {gen.current_rel_gap}. Attempting shapes: {[b.shape for b in batch]}", end=" ")
+                optimizer.zero_grad()
```
galv, line 256: It is theoretically possible to do these three lines in a cuda stream capture with "relaxed" mode to avoid doing any sort of GPU-side computation. However, it will work only for code that has no data-dependent shapes (like torch.nonzero). Note that I haven't run your code and don't know how slow it is right now.

pzelasko: It is surprisingly fast - for ~30 buckets the total runtime seems within 1-2 minutes. If CUDA graph "relaxed" mode would be "ok" with skipping NCCL ops then we might even incorporate this as a training time calibration (which we can't do now because these steps trigger NCCL syncs, if one GPU dies and other doesn't, it would hang). But even as-is I think this is a viable approach.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,283 @@
...
+    def max_batch_size(self) -> int | None:
+        if (
+            self._max_ok is not None
+            and self._min_err is not None
+            and (self.current_rel_gap <= self.rel_gap_thresh or self._min_err - self._max_ok <= 1)
+        ):
+            return self._max_ok
+        return None
+
+    @property
```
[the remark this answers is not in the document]
pzelasko: Added doc to explain

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,283 @@
...
+                # We haven't found a batch size that causes an error yet, keep going 2x higher
+                self._current *= 2
+            else:
+                # Try the middle-point between the known extremes.
+                self._current = round((self._max_ok + self._min_err) / 2)
+
+        return False
+
+
+class FloatList(click.Option):
```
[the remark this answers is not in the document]
pzelasko: Right, I went with click out of an old habit. This auto-parses bucket duration bins `[1,2,3,4]` to list of floats.

scripts/speech_recognition/oomptimizer.py
pzelasko: Ended up going with `--module-name` and `--config-path` like discussed offline. It works well.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,420 @@
...
+            batch = gen(seq_len_in, seq_len_out)
+            oom = False
+            try:
+                print(f"\tCurrent gap: {gen.current_rel_gap}. Attempting shapes: {[b.shape for b in batch]}", end=" ")
+                optimizer.zero_grad()
+                model.training_step(batch, batch_idx)
+                optimizer.step()
+            except torch.cuda.OutOfMemoryError as e:
+                print(f"- OOM!")
+                torch.cuda.memory.empty_cache()
```
galv, line 366: One notable thing is that, when you do empty_cache() here, some memory will still be occupied by "batch", since there are living references to it. `del batch` before empty_cache() will probably fix that. I'm not sure if this matters.

Secondarily, if I understand correctly, the intermediate activations used to do backpropagation will also "be alive" when you call empty_cache(). I'm not sure if this matters either, but if you are setting these to None anytime you change buckets in your actual training run, but not doing so here, then effectively you will a time during forward and back prop when intermediate activations from two different shapes are "alive".

Another thought is that I am not sure if torch.cuda.memory.empty_cache() is the correct approach here, but that requires more investigation on my part.

Note that I have not run this code at this time!

nemo/collections/common/tokenizers/canary_tokenizer.py
```diff
@@ -71,7 +71,7 @@ def text_to_ids(self, text, lang_id) -> list[int]:
             return self._tokenize_special_prompt(text)
         if text.endswith(CANARY_EOS):
             return super().text_to_ids(text[: -len(CANARY_EOS)], lang_id) + [self.eos_id]
-        return super().text_to_ids(text[-len(CANARY_EOS) :], lang_id)
+        return super().text_to_ids(text, lang_id)
```
pzelasko, line 74: drive-by fix

scripts/speech_recognition/oomptimizer.py
pzelasko: For lots of buckets (i.e. 100+) it takes a while. We should try the "relaxed" CUDA graph trick, and if it works, make a follow up PR.

galv:
"Optimal = maximum GPU utilization." in your original comment: Well, optimal here means that the maximum amount of GPU memory (the highest possible such that you won't get an OOM) is used right?

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -1064,6 +1074,26 @@ def predict_step(self, batch, batch_idx=0, dataloader_idx=0, has_processed_signa
...
+        return [
+            {"type": NeuralType(("B", "T"), AudioSignal()), "seq_length": "input"},
+            {"type": NeuralType(("B",), LengthsType()), "seq_length": "input"},
+            {
+                "type": NeuralType(("B", "T"), LabelsType()),
+                "seq_length": "output",
+                "vocab_size": self.tokenizer.vocab_size,
+            },
+            {"type": NeuralType(("B",), LengthsType()), "seq_length": "output"},
+            {"type": "dummy"},
```
galv, line 1093: Are these two lines still needed? It looks like "dummy" entries get filtered out. But I may be missing code in lhotse that might work with this.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,458 @@
...
+from numbers import Number
+
+import click
+import pytorch_lightning as pl
+import torch
+from lhotse import compute_num_samples
+from omegaconf import OmegaConf
+
+from nemo.core.classes.common import Model
+from nemo.core.neural_types import AudioSignal, LabelsType, LengthsType, NeuralType
```
galv, line 14: MaskType may also need to be supported at some point, as an alternative to LengthsType. I don't think it's a big deal, though.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -67,10 +69,12 @@ class LhotseDataLoadingConfig:
     quadratic_duration: float | None = None
     #   c. Lhotse bucketing.
     use_bucketing: bool = False
+    bucket_batch_size: list[int] | None = None
     num_buckets: int = 30
     num_cuts_for_bins_estimate: int = 10000
-    bucket_duration_bins: list[float] | None = None
+    bucket_duration_bins: Any = None  # list[float] | list[list[float]] | None = None
```
galv, line 75: Is there a reason why you must use Any instead of `list[float] | list[list[float]] | None`?

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -67,10 +69,12 @@ class LhotseDataLoadingConfig:
...
     bucket_buffer_size: int = 10000
+    concurrent_bucketing: bool = True  # fetches data in a background thread
```
galv, line 77: Do you have any concern that this might increase memory unreasonably?

e.g., is there a limit to how much the background thread will load data? e.g., in pytorch there is a prefetch_factor option to prevent unbounded memory loading if the data loader is faster than the consumer.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -186,16 +192,13 @@ def get_lhotse_dataloader_from_config(
...
-        ), "You must pass a tokenizer to `get_lhotse_dataloader_from_config` in order to read text-only datasets (enabled via use_multimodal_dataloading)"
+    if tokenizer is not None:
         from nemo.collections.asr.data.audio_to_text_lhotse import TokenizerWrapper
 
         if not isinstance(tokenizer, TokenizerWrapper):
             tokenizer = TokenizerWrapper(tokenizer)
-        # Note this code can also pre-tokenize the text in cuts, but for now we disable it with apply_fn.
-        cuts = cuts.map(partial(tokenize, tokenizer=tokenizer), apply_fn=is_text)
+        cuts = cuts.map(partial(tokenize, tokenizer=tokenizer), apply_fn=None)
+        cuts = cuts.filter(TokenPerSecondFilter(config.min_tps, config.max_tps))
```
galv, line 201: TokenPerSecondFilter needs to have the same min_tps and max_tps parameters as https://github.com/NVIDIA/NeMo/pull/9763/files#diff-7cf5386bc408c1b18dc8c78908e65d75e01aff39680de8fca6b37ac1350add4eR294 for expected performance, right? Assuming that you use estimate_durations_bins_2d.py to create your bins (which I don't see why you wouldn't).

scripts/speech_recognition/oomptimizer.py
galv: just curious how long "a while" is.

The relaxed cuda graph trick definitely won't always work unfortunately... I spoke with someone who works on end-to-end training and he told me that there is a cudaStreamSynchronize() is the torch.amp.GradScaler, which will prevent using relaxed stream capture for models that do gradient scaling in mixed precision training.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -449,6 +478,58 @@ def measure_length(self, example: Any) -> float:
         raise RuntimeError(f"Unsupported example type: {type(example)}")
 
 
+@dataclass
+class FixedBucketBatchSizeConstraint2D(FixedBucketBatchSizeConstraint):
```
galv, line 482: I'm just going to trust these and skip over reading them in depth.

galv (approved):
I think it's very obviously good. I'm certain there is more room for improvement elsewhere that is opened because of this work.

docs/source/asr/datasets.rst
```diff
@@ -818,6 +828,122 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+        input_cfg.yaml
+    Use the following options in your config:
+            num_buckets=30
+            bucket_duration_bins=[[1.91,10],[1.91,17],[1.91,25],...
+            max_duration=...
+            max_tps=...
+    <other diagnostic information about the dataset>
+
+Note that the output in ``bucket_duration_bins`` is a nested list, where every bin specifies
+the maximum duration and the maximum number of tokens that go into the bucket.
```
orena1, line 866: This assume that the buckets are in tar? maybe you can add some-kind of exmaple on how the data looks, it will significantly help to prepare data for using this tool.

pzelasko: It doesn't have any assumptions about the data. These can be regular NeMo manifests or lhotse manifests. You can see the existing examples above to see how to provide nemo/lhotse inputs or how to build an input config for this script.

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -682,9 +683,18 @@ def training_step(self, batch, batch_nb):
...
+        num_tokens = transcript_len.sum()
+        tot_frames = signal.numel()
+        tot_tokens = transcript.numel()
         tensorboard_logs = {
             'train_loss': audio_loss,
             'learning_rate': self._optimizer.param_groups[0]['lr'],
+            'batch_size': signal.shape[0],
+            'num_frames': num_frames,
+            'num_tokens': num_tokens,
+            'input_to_padding_ratio': num_frames / tot_frames,
```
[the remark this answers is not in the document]
pzelasko: Can you think of an alternative way to compute this without requiring the sync? I think it's a useful diagnostic.

nemo/collections/asr/models/aed_multitask_models.py
pzelasko: Yes, dummy will pass an empty tensor, but it still needs to be there to match the training_step function signature.

nemo/collections/common/data/lhotse/dataloader.py
pzelasko: OmegaConf doesn't allow unions of containers with explicit typing.

nemo/collections/common/data/lhotse/dataloader.py
pzelasko: Yeah, if you use oomptimizer but don't specify max_tps, and you e.g. change the training dataset to one that has longer utterances, you'll likely run into OOM. Both estimate_duration_bins[_2d] and oomptimizer print out suggested `max_tps`.

scripts/speech_recognition/oomptimizer.py
pzelasko: I think it's around 15-20 minutes for 150 buckets.

scripts/speech_recognition/estimate_duration_bins_2d.py
```diff
@@ -0,0 +1,319 @@
...
+
+def load_tokenizer(paths: list[str], langs: list[str] = None) -> TokenizerWrapper:
+    if len(paths) == 1:
+        tok = SentencePieceTokenizer(paths[0])
+    else:
+        assert langs is not None and len(paths) == len(
+            langs
+        ), f"Cannot create AggregateTokenizer; each tokenizer must have assigned a language via --langs option (we got --tokenizers={paths} and --langs={langs})"
+        tok = AggregateTokenizer({lang: SentencePieceTokenizer(p) for lang, p in zip(langs, paths)})
+    return TokenizerWrapper(tok)
```
krishnacpuvvada, line 220: A note: when working with speechLM models, sometimes we also need to support hugginface tokenizers (e.g. llama3.1). NeMo has a wrapper around hugging face tokenizer. https://github.com/NVIDIA/NeMo/blob/9f9a01affd16bb6a4f730cb312c73d728bbb2877/nemo/collections/common/tokenizers/huggingface/auto_tokenizer.py#L28

scripts/speech_recognition/estimate_duration_bins_2d.py
```diff
@@ -0,0 +1,319 @@
...
+        "-a", "--langs", nargs="+", help="Language names for each of AggregateTokenizer sub-tokenizers."
+    )
+    parser.add_argument(
+        "-b",
+        "--buckets",
+        type=int,
+        default=30,
+        help="The desired number of buckets (dim0 => covers input sequence length / audio duration).",
+    )
+    parser.add_argument(
```
krishnacpuvvada, line 79: curious, is there a lot of difference between 150 1-d buckets vs 30x5 2-d buckets (after filtering out outliers)

scripts/speech_recognition/estimate_duration_bins_2d.py
pzelasko: Good point. Let's add the support for that in a follow up PR (or speechlm branch) once this is merged.

scripts/speech_recognition/estimate_duration_bins_2d.py
pzelasko: I don't think 150 1D buckets would help you a lot vs 30. With 30, I'm already seeing only 5% of input padding, but there is typically ~40% of output padding. With 2D buckets the exact improvement in padding ratios is not conclusive yet; for 30x2 I've seen it drop to ~23.5%, but with 30x5 I'm seeing ~31% (although there are visually a lot more batches that are close to ~5%, so the distribution seems more bi-modal). I don't really want to try 30x30 because that might take a lot of time to finish, partiallu defeating the purpose.

Anyway, I am positive that 30x2 is a better setup than 30x1 both in terms of throughput (60->80%) / batch-size (4x -> 5x) and a speed-up in convergence.

vadimkantorov:
@pzelasko regarding more brrr for asr models, there is a low-hanging fruit of using PyTorch core's SDPA in conformer attention:
- https://github.com/NVIDIA/NeMo/pull/9590

pzelasko:
> @pzelasko regarding more brrr for asr models in nemo, there is also a low-hanging fruit of using PyTorch core's SDPA in conformer attention:
> 
> * [Use torch sdpa implementation in ASR mha #9590](https://github.com/NVIDIA/NeMo/pull/9590)

Thanks for bringing to my attention. I'll try to review this soon, also CC @galv you might find it interesting.

vadimkantorov:
> I'll try to review this soon

Thanks a bunch :)

Also, maybe https://pytorch.org/blog/flexattention/ will accelerate in the future SDPA with attn_bias :)

pzelasko:
Note: I will only merge this PR after #10086; there are enough changes to Canary's training workflow here that we need these new integration and unit tests to be certain that we introduce no regression.

nemo/collections/asr/data/audio_to_text_lhotse_prompted.py
```diff
@@ -102,126 +105,5 @@ def _collate_tokens(self, tokens: list[list[int]]) -> tuple[torch.Tensor, torch.
         return tokens, token_lens
 
 
-# Mapping from a string name to a known prompt formatter function.
```
pzelasko, line 105: Necessary refactor: split into two files and moved to `common/prompts/{fn,canary}.py`. Having it there allows us to measure the actual output sequence length using the complete tokenized + prompt formatted transcript before we put the training examples into buckets.

nemo/collections/common/data/lhotse/dataloader.py
pzelasko: This turned out to be where the bugs were 😅

nemo/collections/common/prompts/canary.py
```diff
@@ -1,4 +1,11 @@
+import torch
```
pzelasko, line 1: Unchanged, but moved to a separate file from `audio_to_text_lhotse_prompted.py`

nemo/collections/common/prompts/fn.py
```diff
@@ -0,0 +1,38 @@
+from typing import Callable, Sequence
```
pzelasko, line 1: Unchanged, but moved to a separate file from `audio_to_text_lhotse_prompted.py`

galv (commented):
Sorry got distracted by a few things today. Would like to fully review before approving, but let me know if I end up holding you up.

docs/source/asr/datasets.rst
```diff
@@ -818,6 +828,122 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+.. code-block:: bash
+
+    $ python scripts/speech_recognition/estimate_duration_bins_2d.py \
+        --tokenizer path/to/tokenizer.model \
+        --buckets 30 \
+        --sub-buckets 5 \
+        input_cfg.yaml
+    Use the following options in your config:
+            num_buckets=30
+            bucket_duration_bins=[[1.91,10],[1.91,17],[1.91,25],...
```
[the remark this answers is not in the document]
pzelasko: thanks, fixed

docs/source/asr/datasets.rst
```diff
@@ -818,6 +828,122 @@ For multi-dataset setups, one may provide multiple manifests and even their weig
...
+
+.. code-block:: bash
+
+    $ python scripts/speech_recognition/oomptimizer.py \
+        --config-path fast-conformer_aed.yaml \
+        --module-name nemo.collections.asr.models.EncDecMultiTaskModel \
+        --buckets '[[3.975,30],[3.975,48],[4.97,37],[4.97,60],[5.851,42],[5.851,71],[6.563,46],[6.563,79],[7.32,49],[7.32,88],[8.19,54],[8.19,99],[8.88,61],[8.88,107],[9.75,66],[9.75,117],[10.55,72],[10.55,127],[11.21,76],[11.21,135],[11.87,79],[11.87,143],[12.54,82],[12.54,151],[13.08,87],[13.08,157],[13.62,91],[13.62,164],[14.16,93],[14.16,170],[14.7,96],[14.7,177],[15.19,99],[15.19,183],[15.67,101],[15.67,189],[16.13,103],[16.13,194],[16.66,105],[16.66,200],[17.2,108],[17.2,207],[17.73,111],[17.73,213],[18.2,114],[18.2,219],[18.69,117],[18.69,225],[19.15,120],[19.15,230],[19.62,123],[19.62,236],[20.264,122],[20.264,244],[32.547,173],[32.547,391],[36.587,227],[36.587,440],[40.0,253],[40.0,480]]'
+    <output logs from the search>
+    The final profile is:
+            bucket_duration_bins=[[3.975,30],[3.975,48],[4.97,37],[4.97,60],[5.851,42],[5.851,71],[6.563,46],[6.563,79],[7.32,49],[7.32,88],[8.19,54],[8.19,99],[8.88,61],[8.88,107],[9.7
```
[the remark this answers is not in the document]
pzelasko: right, I merged the lines now

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -512,6 +512,7 @@ def _setup_dataloader_from_config(self, config: Optional[Dict]):
                 tokenizer=self.tokenizer,
                 prompt_format_fn=get_prompt_format_fn(self.prompt_format),
             ),
+            tokenizer=self.tokenizer,
```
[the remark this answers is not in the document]
pzelasko: Great point! Adding that. Will also add integration tests for 2D bucketing estimation, OOMptimizer script, and a training script with 2D buckets once my other PR with new integration tests is merged.

scripts/speech_recognition/estimate_duration_bins_2d.py
```diff
@@ -0,0 +1,319 @@
...
+        sizes.append(dur)
+        num_tokens.append(toks)
+    sizes = np.array(sizes, dtype=np.float32)
+    num_tokens = np.array(num_tokens, dtype=np.int32)
+    joint = np.rec.fromarrays([sizes, num_tokens])
+    joint.sort()
+    sizes = joint.f0
+    num_tokens = joint.f1
+
+    size_per_bucket = sizes.sum() / num_buckets
```
galv, line 163: I don't know if you ever said this explicitly, but it took me a while to understand that you are choosing the bucket lower and upper bounds on duration such that every bucket has approximately the same duration.

scripts/speech_recognition/estimate_duration_bins_2d.py
```diff
@@ -0,0 +1,319 @@
...
+                if tot_toks > tokens_per_subbucket:
+                    bins.append((size, num_toks))
+                    tot_toks = 0
+                tot_toks += num_toks
+            bins.append((size, math.ceil(size * max_tps)))
+            bin_indexes.append(binidx)
+            tot = 0.0
+        tot += size
+
+    num_tokens_bucket = num_tokens[bin_indexes[-1] : binidx]
```
galv, line 198: Consider somehow merging this into the loop above. Not sure what that would take. Up to you.

scripts/speech_recognition/estimate_duration_bins_2d.py
```diff
@@ -0,0 +1,319 @@
...
+    if prompt is not None:
+        turns = prompt.get_default_dialog_slots()
+        last_turn = {"role": prompt.OUTPUT_ROLE, "slots": prompt.get_slots(prompt.OUTPUT_ROLE)}
+        assert len(last_turn["slots"]) == 1  # TODO: not sure how to handle multi-slot for system output here
+        for key in last_turn["slots"]:
+            last_turn["slots"][key] = cut.supervisions[0].text
+        last_turn["slots"][prompt.PROMPT_LANGUAGE_SLOT] = cut.supervisions[0].language
+        turns.append(last_turn)
+        ans = prompt.encode_dialog(turns)
+        cut.supervisions[0].tokens = ans["input_ids"]
```
galv, line 233: Just to make sure I understand, `ans["input_ids"]` contains the entire token sequence that the model will train on to predict, not just the initial prompt (i.e., " <|startoftranscript|>" and whatnot), right?

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,525 @@
...
+            click.echo(f"Intializing ASR model from pretrained checkpoint {pretrained_name}.")
+            model = ASRModel.from_pretrained(pretrained_name, trainer=trainer).to(device)
+        else:
+            assert config_path is not None, "--module-name requires --config-path to be specified as well."
+            assert module_name is not None, "--config-path requires --module-name to be specified as well."
+            cfg = OmegaConf.load(config_path)
+            namespace, name = module_name.rsplit('.', maxsplit=1)
+            model_cls = getattr(importlib.import_module(namespace), name)
+            model = model_cls(cfg=cfg.model, trainer=trainer).to(device)
+        model_clones.append(model)
```
galv, line 361: I don't know about this for various reasons, but let's keep it for now, since I believe your choice here is appropriately conservative. I need to study DistributedDataParallel in more detail myself.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,525 @@
...
+                del batch
+                # Note: We could call empty_cache() to free up some more memory on the GPU,
+                #       but we have found out empirically that this causes a mismatched condition
+                #       between OOMptimizer and the actual training. During training, there is some
+                #       degree of memory fragmentation and it's better to simulate that in OOMptimizer.
+                # torch.cuda.memory.empty_cache()
+                torch.cuda.reset_max_memory_allocated()
+            return oom
+
+        with torch.autocast("cuda", getattr(torch, dtype)):
```
galv, line 473: Small FYI: the autocast's "cache cast" will be destroyed upon exiting the context manager's context. This should not affect the correctness of your work (since the new allocations will be the same size as the old ones and the allocator does a "best fit" search), but it does impose a slowdown because the cache must be reconstructed.

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,525 @@
...
+                    raise RuntimeError(f"Unexpected modality combination: {_}")
+
+        return [_determine_lens_for_bucket(bin) for bin in buckets]
+
+    click.echo("Starting profiling.")
+    max_seq_lens = get_max_seq_lens(buckets)
+    gen = ProfilingBatchGenerator(schema=schema, start_batch_size=start_batch_size, rel_gap_thresh=threshold)
+    profile = {}
+
+    # Iterate buckets from the largest to the smallest sequences. This usually ends up creating
```
galv, line 431: Wait, really? I would expect that going from small to large sequences creates more memory fragmentation. Did you write this comment down right?

scripts/speech_recognition/oomptimizer.py
```diff
@@ -0,0 +1,525 @@
...
+                click.echo(f"\tCurrent gap: {gen.current_rel_gap}... ", nl=False)
+                optimizer.zero_grad()
+                out = model.training_step(batch, batch_idx)
+                out['loss'].sum().backward()
+                optimizer.step()
+            except torch.cuda.OutOfMemoryError as e:
+                click.secho(f"OOM!", fg="yellow")
+                oom = True
+            except RuntimeError as e:
+                if "cuFFT error: CUFFT_INTERNAL_ERROR" not in str(e):
```
galv, line 454: Interesting. Unfortunately there is no way to get the stack of previous cuda errors when we encounter an error like this to ensure that the problem really is an OOM... Yikes.

BTW, I am considering replacing the STFT operation with 1D convolutions as discussed. So I wonder what error cudnn gives if it gets an OOM error, though I'm not sure where it might allocate internally other than for im2col-based GEMM, and even though I think you have to pass the workspace yourself...

scripts/speech_recognition/estimate_duration_bins_2d.py
pzelasko: That's right. This is the relevant discussion about it (although a lot of other facts mentioned there about lhotse are already outdated, as is the sampler we used there itself) https://github.com/lhotse-speech/lhotse/issues/364

galv (dismissed):

scripts/speech_recognition/oomptimizer.py
pzelasko: Yeah, I ran it both ways. I saw that the batch sizes for the first half of the buckets (shorter examples) are smaller if we start from the longest examples. I don't know how to explain that TBH.

scripts/speech_recognition/estimate_duration_bins_2d.py
pzelasko: Correct:  `input_ids == context_ids + answer_ids` (context=prompt, answer=model prediction)

scripts/speech_recognition/oomptimizer.py
pzelasko: We can disable this if we set `gradient_buckets_as_view=True` (or w/e that option was called). Since this is not the default in NeMo and most users don't know, I prefer to set the conservative choice as the default for now. I have a separate PR where I'm enabling the gradient view thingy where I can override this choice later.

scripts/speech_recognition/oomptimizer.py
pzelasko: cool, moved the autocast to cover the whole profiling loop in that case

scripts/speech_recognition/estimate_duration_bins_2d.py
pzelasko: Refactored using a closure

.github/workflows/cicd-main.yml
```diff
@@ -794,6 +794,66 @@ jobs:
...
+          --sub-buckets 2
+        # 2D buckets with prompt [AED/Canary, SpeechLM] / with aggregate tokenizer + prompt format 
+        python scripts/speech_recognition/estimate_duration_bins_2d.py \
+          /home/TestData/an4_dataset/an4_train_lang.json \
+          --tokenizer /home/TestData/asr_tokenizers/canary/canary_spl_tokenizer_v32/tokenizer.model \
+             /home/TestData/asr_tokenizers/canary/en/tokenizer_spe_bpe_v1024_max_4/tokenizer.model \
+             /home/TestData/asr_tokenizers/canary/es/tokenizer_spe_bpe_v1024_max_4/tokenizer.model \
+          --langs spl_tokens en es \
+          --prompt-format canary \
+          --prompt '[{"role":"user","slots":{"source_lang":"en","target_lang":"en","task":"asr","pnc":"yes"}}]' \
```
ko3n1g, line 823:
```suggestion
          --prompt \'[{"role":"user","slots":{"source_lang":"en","target_lang":"en","task":"asr","pnc":"yes"}}]\'  \
```

I think that should do the trick @pzelasko

.github/workflows/cicd-main.yml
```diff
@@ -794,6 +794,66 @@ jobs:
...
+    needs: [cicd-test-container-setup]
+    uses: ./.github/workflows/_test_template.yml
+    with:
+      RUNNER: self-hosted-azure
+      SCRIPT: |
+        # 1D bucketing
+        python scripts/speech_recognition/oomptimizer.py \
+          -c /home/TestData/oomptimizer/fast-conformer_ctc_bpe.yaml \
+          -m nemo.collections.asr.models.EncDecCTCModelBPE \
+          -b '[5.0,10.0]'
```
ko3n1g, line 838:
```suggestion
          -b "[5.0,10.0]"
```

.github/workflows/cicd-main.yml
```diff
@@ -794,6 +794,66 @@ jobs:
...
+        # 2D bucketing
+        python scripts/speech_recognition/oomptimizer.py \
+          -c /home/TestData/oomptimizer/fast-conformer_ctc_bpe.yaml \
+          -m nemo.collections.asr.models.EncDecCTCModelBPE \
+          -b '[[5.0,30],[5.0,45],[10.0,57],[10.0,71]]'
```
ko3n1g, line 843:
```suggestion
          -b "[[5.0,30],[5.0,45],[10.0,57],[10.0,71]]"
```

.github/workflows/cicd-main.yml
```diff
@@ -794,6 +794,66 @@ jobs:
...
+  L2_Speech_Batch_Size_OOMptimizer_Canary:
+    needs: [cicd-test-container-setup]
+    uses: ./.github/workflows/_test_template.yml
+    with:
+      RUNNER: self-hosted-azure
+      SCRIPT: |
+        python scripts/speech_recognition/oomptimizer.py \
+          -c /home/TestData/oomptimizer/fast-conformer_aed.yaml \
+          -m nemo.collections.asr.models.EncDecMultiTaskModel \
+          -b '[[5.0,30],[5.0,45],[10.0,57],[10.0,71]]'
```
ko3n1g, line 855:
```suggestion
          -b "[[5.0,30],[5.0,45],[10.0,57],[10.0,71]]"
```

ko3n1g (commented):
hope you don't mind my comments!

.github/workflows/cicd-main.yml
pzelasko: Thanks... I was seriously scratching my head with this one lol.

galv (approved):
Forgot to hit approve last time.

.github/workflows/cicd-main.yml
pzelasko: It also resulted with an error. I am disabling the check so this PR may go in. If we can figure out how to work around the quoting issue, I will enable the check in a follow up PR.

galv (approved):

ko3n1g (approved):
will merge this after pipeline pass

PR #9869, opened 2024-07-24
Update the test checking for cooperative kernels in conditional nodes.

galv:
Now we conditionally xfail only when a cuda driver version less than 12.6 is installed. CUDA 12.6 fixes this issue. Before it, cooperative kernels could not be used within the body of a conditional node.

We also provide a better error message for users to know that the fix is to upgrade to CUDA 12.6.

### DIFF
--- a/nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
@@ -24,8 +24,8 @@ from nemo.collections.asr.parts.utils import rnnt_utils
 from nemo.collections.asr.parts.utils.asr_confidence_utils import ConfidenceMethodMixin
 from nemo.collections.common.parts.optional_cuda_graphs import WithOptionalCudaGraphs
 from nemo.core.utils.cuda_python_utils import (
-    checked_graph,
     check_cuda_python_cuda_graphs_conditional_nodes_supported,
+    checked_graph,
     cu_call,
     run_nvrtc,
     with_conditional_node,
--- a/nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
+++ b/nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
@@ -24,6 +24,7 @@ from nemo.collections.asr.parts.utils import rnnt_utils
 from nemo.collections.asr.parts.utils.asr_confidence_utils import ConfidenceMethodMixin
 from nemo.collections.common.parts.optional_cuda_graphs import WithOptionalCudaGraphs
 from nemo.core.utils.cuda_python_utils import (
+    checked_graph,
     check_cuda_python_cuda_graphs_conditional_nodes_supported,
     cu_call,
     run_nvrtc,
@@ -630,7 +631,7 @@ class GreedyBatchedRNNTLoopLabelsComputer(WithOptionalCudaGraphs, ConfidenceMeth
         with (
             torch.cuda.stream(stream_for_graph),
             torch.inference_mode(),
-            torch.cuda.graph(
+            checked_graph(
                 self.separate_graphs.before_outer_loop, stream=stream_for_graph, capture_error_mode="thread_local"
             ),
         ):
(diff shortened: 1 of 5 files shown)


nemo/core/utils/cuda_python_utils.py
```diff
@@ -219,3 +241,19 @@ def run_nvrtc(kernel_string: str, kernel_name: bytes, program_name: bytes):
...
+    """
+    Wrapper around torch.cuda.graph that checks for common errors that are too vague for an end user to diagnose based on the error message.
+    """
+    try:
+        with torch.cuda.graph(*args, **kwargs):
+            yield
+    except RuntimeError as err:
+        if "CUDA error: invalid argument" in str(err):
+            raise RuntimeError(
+                "CUDA Graph capture failed. It is likely that you are calling a cooperative kernel in your RNN-T or TDT prediction network. Cooperative kernels are not allowed inside the bodies of CUDA Graph conditional nodes until CUDA 12.6. Please update to CUDA 12.6. File an issue if that still does not work."
```
nithinraok, line 257: We only support CUDA 12.5 with published pytorch containers see here: https://docs.nvidia.com/deeplearning/frameworks/support-matrix/index.html. 

Make sure to support running without cuda_graphs decoding by default and for later version we can make cuda_graphs on by default when containers support it.

nithinraok: Update https://github.com/NVIDIA/NeMo/blob/8880c37e210b83f712eef49b6d04e6b0b219e737/nemo/core/utils/cuda_python_utils.py#L21 the minimum version to 12.6

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
```diff
@@ -630,7 +631,7 @@ def _partial_graphs_compile(self):
         with (
             torch.cuda.stream(stream_for_graph),
             torch.inference_mode(),
-            torch.cuda.graph(
+            checked_graph(
```
artbataev, line 634: Do we need `checked_graph` for partial graphs (without conditional nodes)?

nemo/core/utils/cuda_python_utils.py
artbataev: Hmm, I think @galv wants to preserve `12.3` as a requirement, but fail in rare cases when cooperative kernels are selected.

I would suggest in this case applying fallback behavior like this

https://github.com/NVIDIA/NeMo/blob/8880c37e210b83f712eef49b6d04e6b0b219e737/nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py#L614
(also, the same code in `tdt_loop_labels_computer.py`)

```
if self.cuda_graphs_mode is self.CudaGraphsMode.FULL_GRAPH:
    try:
        self._full_graph_compile()
    except RuntimeError:
       # fallback to graphs without while loops
       self.cuda_graphs_mode = self.CudaGraphsMode.NO_WHILE_LOOPS
       self._partial_graphs_compile()
```

galv: Yeah, that's an interesting possibility.

One of the big challenges is that the error returned by torch is not very precise. It's just a RuntimeError corresponding to "invalid argument", or cudaErrorInvalidValue, which is not a precise enough error for us to tell that the problem specifically is that the code is using a cooperative kernel within a conditional node's body graph. And unfortunately we cannot check whether this is the case because conditional node API does not expose a way to get the body graph(s) of a conditional node, right now...

Anyway, I suppose if the error was not because of a cooperative kernel, but because of something else, then there is a good chance the error will get thrown by the partial graphs implementation. But it's still not a guarantee!

nemo/collections/asr/parts/submodules/rnnt_loop_labels_computer.py
galv: This is a very good point. No, we don't.

But I am considering doing your suggestion anyway for a fallback, in which case we wouldn't use checked_graph.

nemo/core/utils/cuda_python_utils.py
pzelasko: IMO sounds like it's worth a shot to be able to move forward here

PR #9896, opened 2024-07-25
Added defer wgrad support with mcore optim

sanandaraj5597:
Removed draining of weight gradient GEMM's from NeMo when using MCore optimizer.

### DIFF
--- a/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
+++ b/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
@@ -817,6 +817,7 @@ class MegatronGPTModel(MegatronBaseModel, TextGeneration):
         ):
             if (
                 self.cfg.get('defer_embedding_wgrad_compute', False) and self.mcore_gpt
+                and not self.use_mcore_dist_optim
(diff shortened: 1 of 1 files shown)


akoumpa (approved):

nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
```diff
@@ -839,7 +841,9 @@ def training_step(self, dataloader_iter):
             ignore_virtual=True
         ):
             if (
-                self.cfg.get('defer_embedding_wgrad_compute', False) and self.mcore_gpt
+                self.cfg.get('defer_embedding_wgrad_compute', False)
+                and self.mcore_gpt
+                and not self.use_mcore_dist_optim
```
deepakn94, line 846: Why does MCore distributed optimizer + defer wgrad support not work?

sanandaraj5597: It works. The functionality done in NeMo for Apex DistOpt is already done in MCore when using Mcore optim. So when using MCore Opt with NeMo, we are doing the same operation twice, so removing it from NeMo when using MCore Opt.

PR #9898, opened 2024-07-25
Updated Latest News and added Blog section

jgerh:
# What does this PR do ?

Updates the Latest News with new blogs and adds a Blog section at the end that provides the legacy blog news.


ericharper (approved):
LGTM. Thanks!

One comment, though, can we start archiving the older news somewhere? I think we should try to keep 3 bullets per domain maximum.

PR #9904, opened 2024-07-25
Temporarily disable cuda graph based RNN-T greedy inference for r2.0.0rc1

galv:
Temporarily disable cuda graph based RNN-T greedy inference for r2.0.0rc1.

For very rare input shapes, a cooperative kernel might be used by
pytorch for LSTM operations. This does not work within a cuda graph
conditional node until CUDA 12.6.

Unfortunately CUDA 12.6 is not part of the 24.07 pytorch container
release, which this release of nemo is intended for.


artbataev (approved):
Thanks!

nithinraok (approved):
Thanks

PR #9932, opened 2024-07-26
log TFLOPs

malay-nagda:
# What does this PR do ?

Log TFLOPs per sec per GPU (for LLMs)

**Collection**: [common, nlp, Run CICD]

# Changelog 
- `callbacks.append(FLOPsMeasurementCallback(self.cfg))`

# Usage

```python
from pytorch_lightning import Trainer
from nemo.collections.common.metrics.perf_metrics import FLOPsMeasurementCallback

callbacks = []
flops_measurement_callback =  FLOPsMeasurementCallback(self.cfg)
callbacks.append(flops_measurement_callback)

trainer = Trainer(..., callbacks=callbacks)
...
trainer.fit(model)
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to #9673

PR #10047, opened 2024-08-05
fix FLOPsMeasurementCallback on_train_end when experiment is None

ar-nowaczynski:
# What does this PR do ?

Fixes basic error in recently introduced #9932 FLOPsMeasurementCallback on_train_end method when `pl_module.logger.experiment` is None.

**Collection**: common

**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/common/metrics/perf_metrics.py
+++ b/nemo/collections/common/metrics/perf_metrics.py
@@ -84,7 +84,8 @@ class FLOPsMeasurementCallback(Callback):
             logging.error(f"Failed to calculate TFLOPs per sec per GPU.\n{exc}")
 
         logging.info(f"TFLOPs per sec per GPU={tflops_per_sec_per_gpu:.2f}")
-        pl_module.logger.experiment.add_scalar("tflops_per_sec_per_gpu", tflops_per_sec_per_gpu)
+        if pl_module.logger.experiment is not None:
+            pl_module.logger.experiment.add_scalar("tflops_per_sec_per_gpu", tflops_per_sec_per_gpu)
 
     def eval_tflops_per_sec_per_gpu(self, train_step_time: List | float | int) -> float:
         """


gautham-kollu:
@ar-nowaczynski  Can you check if https://github.com/NVIDIA/NeMo/pull/10125 fixes this issue.

nemo/collections/common/metrics/perf_metrics.py
```diff
@@ -84,7 +84,8 @@ def on_train_end(self, trainer, pl_module):
             logging.error(f"Failed to calculate TFLOPs per sec per GPU.\n{exc}")
 
         logging.info(f"TFLOPs per sec per GPU={tflops_per_sec_per_gpu:.2f}")
-        pl_module.logger.experiment.add_scalar("tflops_per_sec_per_gpu", tflops_per_sec_per_gpu)
+        if pl_module.logger.experiment is not None:
```
malay-nagda, line 87: I think the check should also include `pl_module.logger is not None in case [no logger](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/trainer/trainer.py#L1576) is initialized.

gautham-kollu: @ar-nowaczynski  Can you path the diff from @malay-nagda' PR #10125  to check if that fixes that issue ? 
and if so, close this ?

ar-nowaczynski: Shortly after creating this PR, I discovered the problem on my side can be easily fixed by adding to config:
```yaml
exp_manager:
  log_tflops_per_sec_per_gpu: False
```
I'm willing to close this PR.
Do you agree to close this PR?

gautham-kollu: Though that works, #10125 is a right fix. 

As you already have a workaround, you could wait for #10125 to be merged and check if fixes without that config. 

After #10125, this config should be a no-op. (One less config)

ar-nowaczynski: Agree. Closing.

PR #10061, opened 2024-08-06
Copy data from GPU to CPU only once in batched_hyps_to_hypotheses for each tensor of interest.

galv:
Previously, several small GPU->CPU copies were used, which caused excess latency linearly proportional to the batch size. For small copies, it is much more efficient to do a single memory copy.

There is still a remaining issue. pack_hypotheses() appears to be copying data from GPU to CPU for the dec_state tensor, one at a time. I would like to remove that, but that is an intrusive change that affects interfaces, so I will not do it for now.

RTFx before this change on librispeech test other: 1766.645297158031
RTFx after  this change on librispeech test other: 1795.3508087085368

### DIFF
--- a/nemo/collections/asr/parts/utils/rnnt_utils.py
+++ b/nemo/collections/asr/parts/utils/rnnt_utils.py
@@ -586,11 +586,15 @@ def batched_hyps_to_hypotheses(
     """
     assert batch_size is None or batch_size <= batched_hyps.scores.shape[0]
     num_hyps = batched_hyps.scores.shape[0] if batch_size is None else batch_size
+    scores = batched_hyps.scores.cpu()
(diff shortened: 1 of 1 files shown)


artbataev (approved):
Cool, thank you!

artbataev (changes requested):
Please, fix also failing tests [tests/collections/asr/decoding/test_batched_hyps_and_alignments.py::TestConvertToHypotheses`](https://github.com/NVIDIA/NeMo/blob/7b7d02fe5e000559875e3fe0e6cf6b5f48d48431/tests/collections/asr/decoding/test_batched_hyps_and_alignments.py#L381)

should be now 
```python
assert (hypotheses[0].y_sequence == torch.tensor([5, 2], device="cpu")).all()
...  # device="cpu" in all asserts
```

PR #10062, opened 2024-08-06
FSDP support of SALM and BESTOW

zhehuaichen:
# What does this PR do ?

Support FSDP in SALM and BESTOW. Thank @pyf98 for early trying

# Usage

Elaborated and tested here https://docs.google.com/document/d/1fr2UlqThlzqvqL_lpc_HIyrh0yCzL27A-QX1eCI_ojg/edit#heading=h.9o9g0iu9eu69

### DIFF
--- a/nemo/collections/multimodal/speech_llm/models/modular_models.py
+++ b/nemo/collections/multimodal/speech_llm/models/modular_models.py
@@ -1659,6 +1659,49 @@ class ModularAudioGPTModel(SpeechLLMAdapterMixin, MegatronGPTSFTModel):
         results.append(model)
         return results
 
+    def configure_sharded_model(self):
+        """Modified version from MegatronBaseModel.
+
+        1. exclude self.model.embedding
+        2. include speech encoder and modality adapter.
+        """
+
+        def find_frozen_submodules(model):
+            frozen_submodules = []
+            frozen_submodule_names = []
+            for name, module in model.named_modules():
+                if (
+                    isinstance(module, torch.nn.Module)
+                    and list(module.parameters())
+                    and all(not param.requires_grad for param in module.parameters())
+                ):
+                    frozen_submodule_names.append(name)
+                    frozen_submodules.append(module)
+            return frozen_submodule_names, frozen_submodules
+
+        if self.use_fsdp:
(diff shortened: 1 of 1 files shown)


pyf98:
A few notes:
- You might need to adapt this class to support additional modules: https://github.com/zhehuaichen/NeMo/blob/speechllm-develop/nemo/collections/nlp/parts/nlp_overrides.py#L626
  - E.g., adding new modules here: https://github.com/zhehuaichen/NeMo/blob/speechllm-develop/nemo/collections/nlp/parts/nlp_overrides.py#L668
- The above class has several methods related to "state dict". Not sure if those are correct for speech LM.
- I excluded the embedding layer of LLM to avoid errors when using it elsewhere. We'd better fix it since the embedding layer can be very large. https://github.com/zhehuaichen/NeMo/commit/44056f2d1f69261c47829181e7d8e4c0c1c077de#diff-e3fcc68c97812462e588465942b917f381808fc65501b32a431b8b311fab6fd3R1691

zhehuaichen:
> A few notes:
> 
> * You might need to adapt this class to support additional modules: https://github.com/zhehuaichen/NeMo/blob/speechllm-develop/nemo/collections/nlp/parts/nlp_overrides.py#L626
>   
>   * E.g., adding new modules here: https://github.com/zhehuaichen/NeMo/blob/speechllm-develop/nemo/collections/nlp/parts/nlp_overrides.py#L668
> * The above class has several methods related to "state dict". Not sure if those are correct for speech LM.
> * I excluded the embedding layer of LLM to avoid errors when using it elsewhere. We'd better fix it since the embedding layer can be very large. [zhehuaichen@44056f2#diff-e3fcc68c97812462e588465942b917f381808fc65501b32a431b8b311fab6fd3R1691](https://github.com/zhehuaichen/NeMo/commit/44056f2d1f69261c47829181e7d8e4c0c1c077de#diff-e3fcc68c97812462e588465942b917f381808fc65501b32a431b8b311fab6fd3R1691)

Thanks for the note.
SpeechLLM does not add extra nlp modules but only reusing them if that is what you mean. Locally I found the ckpt restore seems working. Will confirm with a cluster run

PR #10080, opened 2024-08-08
nemo-ux MixedPrecision fix

akoumpa:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/lightning/pytorch/plugins/mixed_precision.py
+++ b/nemo/lightning/pytorch/plugins/mixed_precision.py
@@ -30,30 +30,21 @@ class MegatronMixedPrecision(MixedPrecision):
     def __init__(
         self,
         precision: Literal["16-mixed", "bf16-mixed"],
-        amp_O2: bool = False,
         device="cuda",
     ) -> None:
+        assert precision in ["16-mixed", "bf16-mixed"], f"Got unknown precision = `{precision}`"
         if precision == "bf16-mixed":
             scaler = None
         else:
             scaler = GradScaler(init_scale=2**32, growth_interval=1000, hysteresis=2)
 
         super().__init__(precision, device, scaler)
-        self.amp_O2 = amp_O2
 
     def connect(
         self, model: Module, optimizers: List[Optimizer], lr_schedulers: List[Any]
     ) -> Tuple[Module, List[Optimizer], List[Any]]:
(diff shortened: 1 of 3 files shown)


cuichenx (dismissed):
verified working from my side, thanks for the fix!

ko3n1g:
stopping this now until the CI out of maintenance

nemo/lightning/pytorch/plugins/mixed_precision.py
```diff
@@ -84,14 +83,10 @@ def convert_optimizer(self, optimizer: Optimizer) -> Optimizer:
...
 
-        if isinstance(optimizer, MainParamsOptimizerWrapper) or not self.amp_O2:
-            return optimizer
-
-        return MainParamsOptimizerWrapper(
-            optimizer,
-            fp32_grad_accum=True,
-            contiguous_grad_bucket=True,
-        )
+        optim_config = get_optim_config(optimizer)
```
cuichenx, line 86: instead of asserts, could we change the optimizer args if self.precision is 16? this way we make the mixed precision plugin the only entry point
```suggestion
        if self.precision == "bf16-mixed":
          optimizer.mcore_optimizer.config.bf16 = True
        elif self.precision == "fp16-mixed":
          optimizer.mcore_optimizer.config.fp16 = True
```

cuichenx (commented):
The code works for me, but a few comments

nemo/lightning/pytorch/plugins/mixed_precision.py
akoumpa: I'm leaning towards keeping the current code. I want to avoid modifying configs that were used to create objects (in this case the optimizer). This is mostly to catch misconfigurations.

cuichenx: Yeah I'm not sure if those flags would be read before this function. @marcromeyn do you know? We could also do that in a followup PR

cuichenx (dismissed):

cuichenx (approved):
good catch

PR #10079, opened 2024-08-07
Drop PyTorch 2.1 version check from fabric strategies

farhadrgh:
# What does this PR do ?

Lightning has [Dropped PyTorch 2.0 from the test matrix](https://github.com/Lightning-AI/pytorch-lightning/pull/20009/files#diff-edeb9aa08130aab819bbadaf08ebf1f761433c177ecc38f546e0b19104f5a470L29) and `TORCH_GREATER_EQUAL_2_1` no longer exists. I bumped the version requirement instead of import/check in `NeMo/nemo/lightning/fabric/strategies.py`

**Collection**: [Note which collection this PR will affect]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #10086, opened 2024-08-08
Canary: CI integration tests; fix for training dataset and backcompat

pzelasko:
# What does this PR do ?

* Integration tests for 3 Canary's transcribe cases, Canary training, and unit test for Canary dataset.
* Allows the user to provide tasks via "task" slot instead of "taskname".
* Fixes an issue in Canary's training dataset class.

**Collection**: ASR

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -532,55 +532,45 @@ jobs:
       AFTER_SCRIPT: |
         rm -rf examples/asr/speech_to_text_results
 
-
-  # L2_Speech_to_Text_AED:
-  #   needs: [cicd-test-container-setup]
-  #   runs-on: self-hosted-azure-gpus-1
-  #   container:
-  #     image: nemoci.azurecr.io/nemo_container_${{ github.run_id }}
-  #     options: 
-  #       # --user 0:128
-  #       --device=/dev/nvidia0
-  #       --gpus all
-  #       --shm-size=8g 
-  #       --env TRANSFORMERS_OFFLINE=0 
-  #       --env HYDRA_FULL_ERROR=1
-  #       --volume /mnt/datadrive/TestData:/home/TestData
-  #   steps:
-  #       - name: Checkout repository
-  #         uses: actions/checkout@v4
-  #       - run: |
-  #           python examples/asr/speech_multitask/speech_to_text_aed.py \
-  #           model.prompt_format=canary \
-  #           model.model_defaults.asr_enc_hidden=256 \
-  #           model.model_defaults.lm_dec_hidden=256 \
-  #           model.encoder.n_layers=12 \
-  #           model.transf_encoder.num_layers=0 \
-  #           model.transf_decoder.config_dict.num_layers=12 \
-  #           model.train_ds.manifest_filepath=/home/TestData/asr/manifests/canary/an4_canary_train.json \
-  #           ++model.train_ds.is_tarred=false \
-  #           model.train_ds.batch_duration=60 \
-  #           +model.train_ds.text_field="answer" \
-  #           +model.train_ds.lang_field="target_lang" \
-  #           model.validation_ds.manifest_filepath=/home/TestData/asr/manifests/canary/an4_canary_val.json \
-  #           +model.validation_ds.text_field="answer" \
-  #           +model.validation_ds.lang_field="target_lang" \
-  #           model.test_ds.manifest_filepath=/home/TestData/asr/manifests/canary/an4_canary_val.json \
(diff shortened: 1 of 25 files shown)


.github/workflows/cicd-main.yml
```diff
@@ -817,6 +817,72 @@ jobs:
...
+        output_filename=preds.json \
+        batch_size=10 \
+        pretrained_name=nvidia/canary-1b \
+        num_workers=0 \
+        amp=false \
+        compute_dtype=bfloat16 \
+        matmul_precision=medium \
+        2>&1 | tee transcribe.log
+        actual_wer=$(grep -oP "(?<='wer': )[0-9.]*" transcribe.log)
+        (( $(echo "$actual_wer >= 0.031 && $actual_wer <= 0.032" | bc -l) )) && exit 0 || exit 1
```
titu1994, line 838: Don't use such a tight range.

.github/workflows/cicd-main.yml
```diff
@@ -817,6 +817,72 @@ jobs:
...
+        amp=false \
+        compute_dtype=bfloat16 \
+        matmul_precision=medium \
+        +prompt.source_lang="en" \
+        +prompt.target_lang="en" \
+        +prompt.task="asr" \
+        +prompt.pnc="no" \
+        2>&1 | tee transcribe.log
+        actual_wer=$(grep -oP "(?<='wer': )[0-9.]*" transcribe.log)
+        (( $(echo "$actual_wer >= 0.031 && $actual_wer <= 0.032" | bc -l) )) && exit 0 || exit 1
```
titu1994, line 863: Same as above. Keep it 0.03 +- 0.02.

requirements/requirements.txt
```diff
@@ -9,7 +9,7 @@ scikit-learn
 setuptools>=65.5.1
 tensorboard
 text-unidecode
-torch
+torch>=2.1.0
```
titu1994, line 12: Don't use > or any operator. It overrides container version. Revert

pzelasko: It's cherry picked from #10079 to unblock me (CI is broken otherwise and I need to check these checks actually work). I will revert this change later but FYI it looks like this will be merged in another PR

.github/workflows/cicd-main.yml
pzelasko: This is too wide, that would allow lots of variations in WER (1-5% absolute). For fixed data and fixed model the result should be 100% deterministic; and even with some oversights, there is no reason the WER should vary more than within 0.1%. Let me make it 0.03 - 0.032 (the current output is 0.03107 or so).

krishnacpuvvada (approved):
LGTM.

tbartley94 (changes requested):
Could you add tests for changing MultiTaskTranscription config? I've been encountering a few issues were the default assumption `target_lang==lang` breaks some workflows. (We want to preserve flexibility in cases like en-US and en-GB are dialects in the model but having different tokenizers wants to be compared with shared `en`

tbartley94 (dismissed):

krishnacpuvvada (approved):
LGTM

.github/workflows/cicd-main.yml
pzelasko: FWIW you were right (the CI yields 3.8% WER) and I broadened the range because I don't have the time to look into this right now. For the record I find this... scary to say the least that we can get such a wide WER variance depending on the environment.

titu1994: Ouch. Just between runs or between config changes ? Between same runs shouldn't have that wide a spread normally

pzelasko: Between my workstation and the CI environment. It is stable in the CI though (doesn't vary) which gives hopes for debugging it.

BTW I removed the WER check altogether, turning this test into "it can run" for now similarly to other tests in the suite. I was getting bash errors that didn't make sense in the CI env and I'd rather merge this ASAP since multiple people are experiencing these bugs. I'll try to restore these checks later when CI is more stable, letting me experiment a bit.

requirements/requirements.txt
pzelasko: I think they reverted it. I don't have those changes anymore btw.

titu1994: Ah ok that's good

PR #10096, opened 2024-08-09
Export fp8 te nemo to trt-llm

Laplasjan107:
# What does this PR do ?

Add support for exporting FP8 TE NeMo to TRT LLM.

**Collection**: nlp

# Changelog 

# Usage

```python
from nemo.export import TensorRTLLM

trt_llm_exporter = TensorRTLLM(model_dir="/opt/checkpoints/tmp_triton_model_repository/")
trt_llm_exporter.export(nemo_checkpoint_path="/opt/checkpoints/GPT-2B-001_bf16_tp1.nemo", model_type="gptnext", n_gpus=1)

# Autodetects when quantisation flags are not set
output = trt_llm_exporter.forward(["What is the best city in the world?"], max_output_token=17, top_k=1, top_p=0.0, temperature=1.0, fp8_quantized=True,  fp8_kvcache=True)
print("output: ", output)
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/export/trt_llm/nemo_ckpt_loader/nemo_file.py
```diff
@@ -85,24 +139,65 @@ def load_sharded_metadata_torch_dist(checkpoint_dir: Union[Path, TarPath], torch
     return state_dict
 
 
+def load_sharded_pickle_extra_state_scale(dir: Union[Path, TarPath]):
+    scales = []
+    layer_number = 0
+
+    while pt_file_list := list(dir.glob(f'shard_{layer_number}_*.pt')):
+        pt_file = pt_file_list[0]
+        with pt_file.open('rb') as checkpoint_file:
```
janekl, line 148: Hey @mikolajblaz. Could you please help us with the recommended way of reading extra states? For example, can we somehow avoid double `torch.load` and that `dictionary.seek(0)` call?

(Background is that we need to process extra states to get FP8 scaling factors for direct FP8 TRT-LLM deployment.)

nemo/export/trt_llm/nemo_ckpt_loader/nemo_file.py
```diff
@@ -56,25 +59,76 @@ class TarFileSystemReader(FileSystemReader):
...
+        state_dict.pop(key)
+    return state_dict
+
+
+def load_scaling_factors(state_dict: dict, basename: str, size: int):
+    scales = []
+    for layer in range(size):
+        keyname = f'{basename}/shard_{layer}_{size}'
+        extra_state = state_dict[keyname][0]
+        extra_state.seek(0)
```
janekl, line 94: [Same here](https://github.com/NVIDIA/NeMo/pull/10096/files#r1721540129) @mikolajblaz

nemo/export/tensorrt_llm.py
```diff
@@ -167,6 +167,8 @@ def export(
         multiple_profiles: bool = False,
         gpt_attention_plugin: str = "auto",
         gemm_plugin: str = "auto",
+        fp8_quantized: bool = False,
```
oyilmaz-nvidia, line 170: Is there way to detect if a ckpt is a fp8 ckpt by reading something from config?

nemo/export/trt_llm/converter/model_converter.py
```diff
@@ -38,10 +39,11 @@
 def get_config(decoder_type, config):
     if decoder_type == "llama":
         return LLaMAConfig(**config)
-    elif decoder_type == "gpt" or decoder_type == "gptnext":
+
+    if decoder_type in ["gpt", "gptnext"]:
```
oyilmaz-nvidia, line 43: Not sure if we need this.

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
```diff
@@ -118,37 +135,28 @@ def convert_model_to_trt_llm_ckpt(
 
     has_position_embedding = get_layer_name("position_embedding", prefix) in model_state_dict
     has_lm_head = get_layer_name("output_layer", prefix) in model_state_dict
```
oyilmaz-nvidia, line 137: Why are some of the values removed?

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
```diff
@@ -118,37 +135,28 @@ def convert_model_to_trt_llm_ckpt(
...
-        in ["swiglu", "geglu", "fast-swiglu", "fast-geglu"]
-        and (decoder_type == "gptnext" or is_mcore),
-        "num_attention_heads": num_attention_heads,
-        "num_kv_heads": num_kv_heads,
-        "kv_channels": kv_channels,
-        "use_attention_nemo_shape": True,
-        "transpose_weights": True,
-        "use_parallel_embedding": use_parallel_embedding,
-    }
+    export_config.update(
```
oyilmaz-nvidia, line 152: Same here, why did you remove some of the values?

oyilmaz-nvidia (changes requested):
Can we please remove the parts that are not related to the fp8 ckpt support? I saw some of the values in the config removed. We are in the process of moving main parts of the export into mcore and some of the code optimizations were done in the mcore already.

nemo/export/tensorrt_llm.py
```diff
@@ -202,6 +204,8 @@ def export(
             multiple_profiles: (bool): enables multiple profiles feature of TRT-LLM. Default = False
             gpt_attention_plugin (str): enable the gpt attention plugin. Default = "auto"
             gemm_plugin (str): enable the gpt plugin. Default = "auto"
+            fp8_quantized (bool): enables exporting to FP8 TRT-LLM checkpoints
+            fp8_kvcache (bool): enables FP8 KV-cache quantization
```
JimmyZhang12, line 208: Following up on Onur's comment: What happens if these are not set correctly?
Perhaps if would be more user friendly to make these optional, so if these are not set then we try to detect the data type format of the checkpoint. The function signature of this function is already pretty huge

nemo/export/trt_llm/converter/model_converter.py
```diff
@@ -203,7 +229,7 @@ def model_to_trtllm_ckpt(
         return weights_dicts, model_configs
 
     pp_key = {
-        "transformer.vocab_embedding.weight",
+        vocab_embedding_key,
```
JimmyZhang12, line 232: is this because the name of the embedding weight changes between TE FP8 and mcore? Or just to make the code cleaner?

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
JimmyZhang12: [I believe these were refactored into here](https://github.com/NVIDIA/NeMo/pull/10096/files#diff-4a09132034fd9a231315e092fa057c1ad7a7db379985c97d776a8dd1ea995b25R83)

nemo/export/trt_llm/converter/model_converter.py
```diff
@@ -78,6 +80,21 @@ def prompt_convert(prompt_config, prompt_weights):
     return vtokens_embeddings
 
 
+def create_common_export_config(nemo_model_config, decoder_type, fp8_quantized=False, fp8_kvcache=False):
+    is_mcore = nemo_model_config.get("mcore_gpt", False)
```
JimmyZhang12, line 84: I dont think we actually support legacy non-mcore nemo models. Maybe you can set this to false and eventually this can get cleaned up.

nemo/export/trt_llm/converter/model_converter.py
```diff
@@ -242,14 +267,12 @@ def model_to_trtllm_ckpt(
 
         if mapping.is_first_pp_rank():
             embedding_weight = (
-                np.ascontiguousarray(
-                    split(weights_dict["transformer.vocab_embedding.weight"], mapping.tp_size, mapping.tp_rank)
-                )
+                np.ascontiguousarray(split(weights_dict[vocab_embedding_key], mapping.tp_size, mapping.tp_rank))
```
JimmyZhang12, line 270: There are alot of `weights_dict[vocab_embedding_key]` I wonder if you could just reference the weight directly

JimmyZhang12 (dismissed):
In general most of the changes are very much well needed code cleanups that had needed to happen for a while.
@shanmugamr1992 is simultaneously working to port this code to megatron core, and he also has many changes to cleanup the code. As @oyilmaz-nvidia has mentioned, we should coordinate with him to see which of these are needed to land in this PR [For instance he also refactors the giant messy weight name dict here.](https://github.com/NVIDIA/NeMo/pull/10096/files#diff-ca09d6b4056ac09a60896695f946518fc696a61477eaa327843d4eaeb8a65bffR40)

Can we make sure these refactors do not break the NeMo-Aligner's existing code path? I believe the CI isn't running for aligner. cc @terrykong

terrykong:
@JimmyZhang12 I'm not sure I will be able to test this, but I'd like to try to manually test the NeMo-Aligner TRTLLM integration with this PR. It looks like this PR is rooted after @oyilmaz-nvidia upgraded to v11, so I can try a one-off build of (TRTLLM v10) + (this Nemo PR) + (mcore ToT) and hopefully with a few hacks I can validate

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
Laplasjan107: Indeed, I've moved the common parts to a single place

nemo/export/tensorrt_llm.py
Laplasjan107: Yes, we can read if the checkpoint is in fp8 from `nemo_config`.

The idea behind the flags is:
- preserving the old behaviour,
- enabling deployment of fp8 checkpoints on non-fp8 hardware,
- in some [cases](https://github.com/NVIDIA/NeMo/blob/753c70e5c5cfc6acb7ecfb416374aed59d5d233c/nemo/export/quantize/quantizer.py#L120-L125) using FP8 KV-cache leads to accuracy regression,
- from the developmental point of view, it will be easier to isolate problems when some issues arise.

I've changed the default behaviour to auto-detection, that can be overridden by the flags.

nemo/export/tensorrt_llm.py
Laplasjan107: For FP8 checkpoints:
* `fp8_quantized=False`, creates a non-quantised checkpoint
* `fp8_kvcache=False`, disables KV-cache quantisation

For non-FP8 checkpoints:
* `fp8_quantized=True`, TRT-LLM fails to build with `Required, but not provided` error regarding scaling factors
* `fp8_kvcache=True`, exports the engine with quantised KV-cache

As suggested, I've made them optional with a new commit (and by default auto-detect)

nemo/export/trt_llm/converter/model_converter.py
Laplasjan107: Just to make it cleaner

nemo/export/trt_llm/converter/model_converter.py
Laplasjan107: I suppose it will be refactored once the [mcore export](https://github.com/NVIDIA/NeMo/pull/10238) module is completed.

nemo/export/trt_llm/converter/model_converter.py
Laplasjan107: I agree, but since the PR is already quite big and there is an ongoing migration to Megatron, I'd leave it this way not to create further confusion.

nemo/export/trt_llm/nemo_ckpt_loader/nemo_file.py
janekl: We discussed this and it is fine to import from `megatron.core` here, especially in the light of https://github.com/NVIDIA/NeMo/pull/10238, that's a minor detail anyway.

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
Laplasjan107: Same as above, refactored into a single place

terrykong (changes requested):
Blocking until I can verify aligner doesn't obviously break (ETA this week)

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
terrykong, line 1: @Laplasjan107 So I see on this line:

https://github.com/NVIDIA/NeMo/pull/10096/files#diff-5f67eef240773c3352ea1d3ad2275be15a011ddc37836ad7213433d4be042393R434

we run into an error b/c the starmap doesn't include scaling_factors, which is a positional arg. Can you update to make sure either scaling factors is optional, or `load_scaling_factors` is called in `dist_model_to_trt_llm_ckpt`?

terrykong (changes requested):

nemo/export/tensorrt_llm.py
terrykong, line 1: Now that export_config needs to be passed in, could you make sure

https://github.com/NVIDIA/NeMo/pull/10096/files#diff-3a3bd49eed8a038989253f6f2e0a01cf494d1de391863d4ab2377a1f7d4cf832R526

Doesn't error as well?

terrykong (changes requested):

nemo/export/trt_llm/converter/model_to_trt_llm_ckpt.py
terrykong: fixed in https://github.com/NVIDIA/NeMo/pull/10096/commits/e2a3139f759995034acc9283d209093d3cbd7425

nemo/export/tensorrt_llm.py
Laplasjan107: Should be fine [now](https://github.com/NVIDIA/NeMo/pull/10096/commits/bcf85e40ac9dc7d4ad901dd7fca16dde5927a638)

terrykong: Thanks for addressing!

terrykong (approved):
From a smoke test, the Aligner code path looks okay

oyilmaz-nvidia (approved):

janekl (approved):

Issue #9967, opened 2024-07-30
Adding an option to only retain optimizer states for the last N checkpoints


Kipok:
**Is your feature request related to a problem? Please describe.**

Currently when training LLMs, the checkpoints are saved with all of the optimizer states, which generally increases the checkpoint size about 6 times. When the LLM itself is very large (e.g. > 30B parameters) and we are running many training jobs at the same time, this requires having a very large storage on the cluster. In case of SFT experiments where we almost never need to restart training from the middle of the job, saving optimizer states for intermediate checkpoints is really wasteful and it would be great to have some control over this.

**Describe the solution you'd like**

Have a config parameter "save_optimizer_states_for_last_checkpoints" (probably a less verbose name is needed). I'd default it to 2 since we always need a last checkpoint state in case job fails or is preempted and it's good to keep another one in case the last state is corrupted for some reason. Then we can set it to -1 to keep the current behavior of saving all optimizer states, set it to 0 in case we don't want to save any, or anything in between. 

The way I'd implement it is by modifying PTL/megatron logic in checkpoint saving and whenever we save a new checkpoint, we would then simply remove the optimizer states from one of the old checkpoints based on the config parameter. 

**Describe alternatives you've considered**

A current alternative that we have to use is basically to manually remove the optimizer states when we run out of space, which is of course very inconvenient. Another potential way to solve this is by having a script that runs alongside the training jobs and monitors the training folder and whenever a new checkpoint is saved, it does the cleaning. This way we don't have to modify nemo code, but it does not feel like a good solution.

**Additional context**

n/a

dimapihtar:
PR is merged: https://github.com/NVIDIA/NeMo/pull/10098

PR #10098, opened 2024-08-09
add save_last_n_optim_states flag

dimapihtar:
# What does this PR do ?

1. Adds save_last_n_optim_states to save optimizer states only for last n checkpoints. This approach allows to save the disk space. Checkpoint file without optimizer states weighs about 7 times less
2. Now it's also possible to load checkpoint without optimizer states for the inference.

# Changelog 

# Usage
* To save optimizer states for last 2 checkpoints, set this config value:

```python
exp_manager.checkpoint_callback_params.save_last_n_optim_states=2 
```
`save_last_n_optim_states=-1` is set by default. It means that optimizer states will be saved for each checkpoint. In order to drop optimizer states for each training checkpoint, you can set `save_last_n_optim_states=0`.

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


ericharper:
closes #9967

.github/workflows/cicd-main.yml
```diff
@@ -2776,6 +2776,43 @@ jobs:
...
+        trainer.log_every_n_steps=1 \
+        trainer.val_check_interval=2 \
+        trainer.limit_val_batches=2 \
+        trainer.accumulate_grad_batches=1 \
+        trainer.max_steps=6 \
+        trainer.precision=bf16 \
+        trainer.gradient_clip_val=1.0 \
+        exp_manager.exp_dir=examples/nlp/language_modeling/gpt_pretrain_results \
+        model.save_last_n_optim_states=1 \
+        model.dist_ckpt_format='torch_dist' \
```
ko3n1g, line 2797:
```suggestion
        model.dist_ckpt_format="torch_dist" \
```

ko3n1g (changes requested):
one small required change to make the gpu go burr burr and then it should be fine

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -410,6 +412,9 @@ def save_checkpoint(
             if self.is_global_zero or app_state.data_parallel_rank == 0:
                 self.checkpoint_io.save_checkpoint(checkpoint, filepath, storage_options=storage_options)
 
+        if self.save_last_n_optim_states >= 0 and '-last' in filepath:
```
mikolajblaz, line 415: This creates a big entanglement in the code.

`-last` is very specific to ModelCheckpoint callbacks, the strategy shouldn't know about stuff like this and it shouldn't happen during `save_checkpoint` call

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -584,6 +589,51 @@ def _integrate_original_checkpoint_data(self, checkpoint: Dict[str, Any]) -> Dic
 
         return checkpoint
 
+    def _drop_optimizer_states(self, filepath: Union[str, Path], storage_options: Optional[Any] = None) -> None:
```
mikolajblaz, line 592: most of the code below conceptually fits into ModelCheckpoint class (except for state dict construction, see [here](https://github.com/NVIDIA/NeMo/pull/10241/files#r1736080010))

I would suggest moving it all there except for this part which I guess can stay at `save_checkpoint` and we can just decide whether we want to include optimizer state in `storage_options`
 ```
            checkpoint['optimizer_states'] = [None]
            checkpoint['state_dict'] = OrderedDict([])
            checkpoint['sharded_state_dict'] = self.lightning_module.sharded_state_dict()

            # Save the checkpoint without optimizer states
            self.checkpoint_io.save_checkpoint(checkpoint, checkpoint_path, storage_options=storage_options)

            if HAVE_MODELOPT and hasattr(self.lightning_module, "get_model_module_list"):
                save_sharded_modelopt_state(
                    self.lightning_module.get_model_module_list(),
                    checkpoint_path,
                    self.unwrapped_checkpoint_io.save_sharded_strategy,
                    prefix="model.",
                )
```

**Review by mikolajblaz — COMMENTED** [review]
I think this approach is correct, although quite inefficient (after a few saves, we will load and save old ckpt on **each** new ckpt save).

For now I think it's fine for users who need this functionality, long term we can separate the model and optimizer state inside the checkpoint so that it's easy to remove the opt state without load/save.

**Comment by mikolajblaz** [comment]
> I think this approach is correct, although quite inefficient (after a few saves, we will load and save old ckpt on **each** new ckpt save).
> 
> For now I think it's fine for users who need this functionality, long term we can separate the model and optimizer state inside the checkpoint so that it's easy to remove the opt state without load/save.

For a more efficient implementation of dropping optimizer state, I see 2 alternatives:
- One is NeMo specific and described here: https://gitlab-master.nvidia.com/dl/JoC/nemo-ci/-/issues/578#note_24602416
- second option is adding a capability to separate different parts of the checkpoint **within** the `torch_dist` format so that it's easy to drop one of them in-place. This would be implemented inside MCore and NeMo would just specify that `optimizier_states` is the part to separate during saving and drop during cleanup.

File: nemo/collections/nlp/parts/nlp_overrides.py [hunk 4]
```diff
@@ -584,6 +589,51 @@ def _integrate_original_checkpoint_data(self, checkpoint: Dict[str, Any]) -> Dic
...
+        # Drop optimizer states
+        checkpoint_index = len(checkpoints) - self.save_last_n_optim_states - 1
+        if len(checkpoints) > self.save_last_n_optim_states:
+            checkpoint_path = checkpoints[checkpoint_index]
+
+            logging.info(f"Loading '{checkpoint_path}' checkpoint to drop optimizer states...")
+            checkpoint = self.load_checkpoint(checkpoint_path=checkpoint_path)
+
+            # Remove the checkpoint version with optimizer states
+            self.remove_checkpoint(checkpoint_path)
```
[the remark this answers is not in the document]
dimapihtar: @mikolajblaz but in this case we'll get an error that checkpoint exitst and not empty or should we change the name for checkpoint w/o optimizer states?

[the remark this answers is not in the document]
mikolajblaz: We probably have to save to a temp dir and rename.
Imagine the following case: at iter 1000 we have 3 checkpoints from iters 800, 900, 1000 and we want to remove the optimizer state from iter 80. If during removal we have an error we lose the iter 800 checkpoint forever (we can't even retrain the model for a while) and it could be e.g. checkpoint with Top1 val_loss

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -545,7 +556,9 @@ def load_checkpoint(self, checkpoint_path: Union[str, Path]) -> Dict[str, Any]:
 
             # after dist_checkpointing.load, sharded tensors will be replaced with tensors
             checkpoint['state_dict'] = sharded_state_dict
-            checkpoint['optimizer_states'] = [self.optimizer_sharded_state_dict(is_loading=True)]
+            # Check whether to load optim states
+            if map_location.get("load_optim_states", True):
```
mikolajblaz, line 560: We shouldn't use `map_location` for that purpose, it's not even in the base strategy interface?

For the record, this approach will work even if we don't restrict optimizer loading, because we won't save the optimizer state. But restricted load will be way faster because optimizer state is the majority of the checkpoint.

I would suggest adding a new argument to `load_checkpoint`, e.g. `load_optimizer_state: bool = True`

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -391,7 +394,12 @@ def save_checkpoint(
             sharded_optim_state = self.optimizer_sharded_state_dict(
                 unsharded_optim_state=checkpoint['optimizer_states'][0]
             )
-            checkpoint['optimizer_states'] = [sharded_optim_state]
+
+            # Check whether to save optim states
+            if storage_options.get('drop_optim_states', False):
```
mikolajblaz, line 399: Even though these are different strategies, maybe let's decide on a single name to reduce confusion, see parallel PR for NeMo 2.0: https://github.com/NVIDIA/NeMo/pull/10241/files#diff-78f81f4094cfea056c177e87c0d527b9ce27cee11813138e5a2a69370b922c19R610
I'm OK with both.

cc @ashors1

dimapihtar: okay, I'm fine with the name @ashors1 used. Not a big deal I think.

mikolajblaz (approved):

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -198,6 +201,10 @@ def setup(self, trainer, pl_module, stage: str) -> None:
         self.last_model_path = trainer.strategy.broadcast(self.last_model_path)
 
     def on_save_checkpoint(self, trainer, pl_module, checkpoint):
+        # Load the correct state_dict for current checkpoint.
+        # Temporary solution.
+        if self.save_last_n_optim_states >= 0:
+            self._load_current_state_dict(trainer, checkpoint)
```
mikolajblaz, line 207: why is this in `on_save_checkpoint`?

mikolajblaz: Even if for some reason we need to put it here, let's make sure the `._load_current_state_dict` is not executed for "regular" saves

nemo/utils/callbacks/nemo_model_checkpoint.py
```diff
@@ -361,6 +364,78 @@ def _ema_callback(self, trainer: 'pytorch_lightning.Trainer') -> Optional[EMA]:
...
+                trainer.strategy.remove_checkpoint(checkpoint_path)
+                shutil.move(f"{checkpoint_path}{suffix}", checkpoint_path)
+
+            if torch.distributed.is_initialized():
+                torch.distributed.barrier()
+
+            # Load the correct state_dict for current checkpoint.
+            # Temporary solution.
+            checkpoint = trainer.strategy.load_checkpoint(
+                checkpoint_path=ckpt_to_dir(filepath), load_optimizer_states=True
```
mikolajblaz, line 404: Did you test with load_optimizer_states=False maybe? It will be way faster

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -391,7 +391,14 @@ def save_checkpoint(
             sharded_optim_state = self.optimizer_sharded_state_dict(
                 unsharded_optim_state=checkpoint['optimizer_states'][0]
             )
-            checkpoint['optimizer_states'] = [sharded_optim_state]
+
+            # Check whether to save optim states
```
mikolajblaz, line 395: since we are not modifying storage_options maybe let's not turn it into a dict (just check if it's not None line below)

nemo/collections/nlp/parts/nlp_overrides.py
```diff
@@ -470,6 +477,8 @@ def _check_param_groups_mismatch(self, checkpoint_path: Union[str, Path], sharde
         # @akoumparouli: check if it contains an mcore dist opt
         if common_state_dict.get('optimizer_states', [{}])[0].get('param_groups', None) is None:
             return False
+        if sharded_state_dict.get('optimizer_states') is None:
```
mikolajblaz, line 480: [Minor] If you put it before the previous check, you can simplify the `common_state_dict.get('optimizer_states', [{}])` to `common_state_dict['optimizer_states']`

nemo/utils/callbacks/nemo_model_checkpoint.py
dimapihtar: yes, it works fine with load_optimizer_states=False

mikolajblaz (approved):

pablo-garay (approved):

PR #10166, opened 2024-08-15
Add MemoryProfileCallback

ShriyaPalsamudram:
# What does this PR do ?

This callback enables recording a timeline of memory allocations during training. 

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/lightning/pytorch/callbacks/memory_profiler.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Args:
+        dir (Optional[str]): Directory to store the memory profile dump
+
+    Example:
+        >>> callback = MemoryProfileCallback(dir="/mem_profile")
+        >>> trainer = Trainer(callbacks=[callback])
+    """
+
+    def __init__(self, dir: str = "/mem_profile"):
```
hemildesai, line 27: Can we add an option to enable [`warn_tensor_cycles`](https://pytorch.org/blog/understanding-gpu-memory-2/) in this profiler as well?

maanug-nv:
Currently, this will dump the snapshot for all ranks. Should we add an option to only save a snapshot from rank 0? or a specifiable rank?

nemo/lightning/pytorch/callbacks/memory_profiler.py
maanug-nv: The blog post recommends adding that to model init. Is it possible to do this via a Callback? Could we just make this a flag for GPTModel?

ShriyaPalsamudram: Addressed in latest commit

ShriyaPalsamudram:
> Currently, this will dump the snapshot for all ranks. Should we add an option to only save a snapshot from rank 0? or a specifiable rank?

Addressed in latest commit

nemo/lightning/pytorch/callbacks/memory_profiler.py
hemildesai: I think we can just do it in the `on_train_start` method.

ShriyaPalsamudram: Also might help catch issues before on_train_start

maanug-nv (commented):
Callback LGTM. I have no concerns about memory profiling, since this is almost identical to the one callback we've been using. For memory cycles, we should test by introducing an intentional cycle. Also curious if this produces warnings about any cycles not related to our code (eg in PTL).

maanug-nv:
also, can we allow disabling the profiling, so users can only run the cycle detector?

nemo/lightning/pytorch/callbacks/memory_profiler.py
```diff
@@ -0,0 +1,78 @@
+import os
+
+import torch
+from pytorch_lightning.callbacks.callback import Callback
+from torch.utils.viz._cycles import warn_tensor_cycles
```
hemildesai, line 5: Can we move this import inside the callback to be safe?

hemildesai (approved):
LGTM, just one minor comment.

nemo/lightning/pytorch/callbacks/memory_profiler.py
ShriyaPalsamudram: Isn't it better to fail early if there is an issue with this import than fail inside the callback?

hemildesai:
Yeah I was just being cautious to avoid side effects of the import but I don't think this import should have any, so it's probably fine.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #10213, opened 2024-08-20
Lazy import tokenizers

akoumpa:
# What does this PR do ?

Had this on my local disk and wanted to push it here as well. 

The goal is to avoid importing all tokenization libraries when the user will only use one.

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/common/tokenizers/en_ja_tokenizers.py
+++ b/nemo/collections/common/tokenizers/en_ja_tokenizers.py
@@ -36,6 +36,7 @@ class EnJaProcessor:
 
     def __init__(self, lang_id: str):
         from sacremoses import MosesDetokenizer, MosesPunctNormalizer, MosesTokenizer
+
         self.lang_id = lang_id
         self.moses_tokenizer = MosesTokenizer(lang=lang_id)
         self.moses_detokenizer = MosesDetokenizer(lang=lang_id)
--- a/nemo/collections/common/tokenizers/en_ja_tokenizers.py
+++ b/nemo/collections/common/tokenizers/en_ja_tokenizers.py
@@ -81,6 +81,7 @@ class JaMecabProcessor:
 
     def detokenize(self, text: List[str]) -> str:
         from pangu import spacing
+
         RE_WS_IN_FW = re.compile(
             r'([\u2018\u2019\u201c\u201d\u2e80-\u312f\u3200-\u32ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff00-\uffef])\s+(?=[\u2018\u2019\u201c\u201d\u2e80-\u312f\u3200-\u32ff\u3400-\u4dbf\u4e00-\u9fff\uf900-\ufaff\uff00-\uffef])'
         )
--- a/nemo/collections/common/tokenizers/en_ja_tokenizers.py
+++ b/nemo/collections/common/tokenizers/en_ja_tokenizers.py
@@ -15,7 +15,6 @@ import re
 from typing import List
 
 from pangu import spacing
-from sacremoses import MosesDetokenizer, MosesPunctNormalizer, MosesTokenizer
 
 try:
     import ipadic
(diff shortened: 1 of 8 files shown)


nemo/collections/nlp/modules/common/tokenizer_utils.py
```diff
@@ -96,46 +77,55 @@
...
     if special_tokens is None:
         special_tokens_dict = {}
     else:
         special_tokens_dict = special_tokens
 
     if 'megatron' in tokenizer_name:
-        if not HAVE_MEGATRON_CORE:
+        try:
+            from nemo.collections.nlp.modules.common.megatron.megatron_utils import get_megatron_tokenizer
+            from nemo.collections.nlp.modules.common.megatron.megatron_utils import get_megatron_vocab_file, get_megatron_merges_file
```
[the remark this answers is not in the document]
akoumpa: Need to figure this out, do not have time now.

nemo/collections/nlp/modules/common/tokenizer_utils.py
```diff
@@ -193,26 +190,32 @@ def get_nmt_tokenizer(
             trust_remote_code=trust_remote_code,
         )
     elif library == 'sentencepiece':
+        from nemo.collections.common.tokenizers.sentencepiece_tokenizer import SentencePieceTokenizer
```
aklife97, line 193: whats the idea behind having these lazy imports? do we want to make some packages optional?

nemo/collections/common/tokenizers/en_ja_tokenizers.py
```diff
@@ -15,7 +15,6 @@
 from typing import List
 
 from pangu import spacing
```
aklife97, line 17: why not lazy import pangu and other lang specific utilities as well then?

nemo/collections/nlp/modules/common/tokenizer_utils.py
```diff
@@ -16,28 +16,8 @@
...
-from nemo.collections.nlp.parts.nlp_overrides import HAVE_MEGATRON_CORE
 from nemo.utils import logging
 
-try:
-    from nemo.collections.nlp.modules.common.megatron.megatron_utils import get_megatron_tokenizer
-
-    HAVE_MEGATRON_CORE = True
-
-except (ImportError, ModuleNotFoundError):
-    HAVE_MEGATRON_CORE = False
```
aklife97, line 38: why are we removing the import guard?

nemo/collections/nlp/modules/common/tokenizer_utils.py
akoumpa: I profiled and looks like we spend a lot of time importing tokenizers, but eventually we only use one (at least in most cases)

nemo/collections/nlp/modules/common/tokenizer_utils.py
akoumpa: since the import is move, why is the import guard still needed?

nemo/collections/common/tokenizers/en_ja_tokenizers.py
akoumpa: good point, it did not show up in the profiling, I can move it

nemo/collections/nlp/modules/common/tokenizer_utils.py
aklife97: I see, the scattered imports make code look unclean do you think there is a solution to that?

maybe having a common function to do the imports or maybe doing some importlib lazyimport stuff?
I dont mind how this PR changes the import, but would definitely prefer if we can somehow clean it, not sure about how though

akoumpa: I echo the concern over the scattered imports and tbh I don't see a substantially better solution offered in python. We could do importlib but would that be that much better than the current solution? It seems hard to beat the cleanness of a few imports at the top of the file, which unfortunately bear a considerable loading time cost.

The change is isolated so I'm wondering if we could permit it here. This lazy importing pattern is used frequently in NeMo-UX btw.

aklife97 (approved):
LGTM, thank you!

in the broader horizon it would be great if we can come up with cleaner way to lazy import that doesn't make us embed the import calls into individual code blocks, but since this speeds up init, we should push this in current form quickly

PR #10241, opened 2024-08-23
[NeMo-UX] checkpointing improvements

ashors1:
# What does this PR do ?
- saves weights and artifacts to separate directories
- provides the option to skip saving optimizer states on train end

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -48,13 +48,16 @@ class ModelCheckpoint(PTLModelCheckpoint):
...
+        always_save_artifacts: Whether to dump the artifacts needed to reinintialize the current
+            model, trainer, and dataloader to allow for reproducibility of experiments.
+        save_artifacts_on_train_end: Whether to dump the artifacts on_train_end regardless of whether
+            ``always_save_artifacts`` is ``True``.
         async_save: Whether to enable asynchronous checkpointing.
         try_restore_best_ckpt: Whether to restore the best model path.
     """
 
     UNFINISHED_CHECKPOINT_SUFFIX = "-unfinished"
+    MODEL_WEIGHTS_PATH = "model_weights"
```
mikolajblaz, line 60: in .nemo checkpoints it's called `model_weights` because it's just the model. During training this is also optimizer state and the whole training state (PTL loops, etc.).

Is it going to be additionally converted into a final checkpoint after training, or is the plan to keep the checkpoints as-is after training?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -251,14 +264,16 @@ def setup(self, trainer, *args, **kwargs) -> None:
...
-
     def on_train_end(self, trainer, pl_module):
+        from nemo.utils.get_rank import is_global_rank_zero
+
         if trainer.fast_dev_run:
             return None
 
+        if not self.save_optim_on_train_end:
+            ## Do not include optimizer states in final checkpoint
+            trainer.strategy.ckpt_include_optimizer = False
```
mikolajblaz, line 275: This flag is not used anywhere?
I think this shouldn't be a modifiable attribute because this could lead to confusion. How about `storage_options` argument to `trainer.strategy.save_checkpoint`?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -445,18 +471,70 @@ def _save_checkpoint(self, trainer: 'pytorch_lightning.Trainer', filepath: str)
...
+            checkpoint_path = checkpoints[checkpoint_index]
+
+            logging.info(f"Loading '{checkpoint_path}' checkpoint to drop optimizer states...")
+            ## TODO: clean
+            checkpoint = trainer.strategy.load_checkpoint(Path(checkpoint_path) / ModelCheckpoint.MODEL_WEIGHTS_PATH)
+
+            # Remove the checkpoint version with optimizer states
+            self._remove_checkpoint(trainer, checkpoint_path)
+
+            checkpoint['optimizer'] = [None]
```
mikolajblaz, line 500: this code can't live in ModelCheckpoint class because it's very specific to a MegatronStrategy.

I would suggest handling the "no optimizer" version in `strategy.save_checkpoint`

nemo/lightning/resume.py
```diff
@@ -37,6 +37,8 @@ class AutoResume(Resume, io.IOMixin):
     checkpoints in NeMo.
     """
 
+    MAYBE_MODEL_WEIGHTS_PATH = "model_weights"
```
mikolajblaz, line 40: why maybe?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
ashors1: I reverted this change. Support for `save_last_n_optim_states` will be added in a subsequent PR.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -48,13 +48,16 @@ class ModelCheckpoint(PTLModelCheckpoint):
             ``every_n_epochs`` or ``every_n_train_steps``.
         save_best_model: When ``True``, reloads and saves the best checkpoint.
         save_on_train_epoch_end: Whether to run checkpointing at the end of the training epoch
-        enable_nemo_ckpt_io: Whether to dump the current model model state, including the
-            config file, to allow for reproducibility of experiments.
+        always_save_artifacts: Whether to dump the artifacts needed to reinintialize the current
```
marcromeyn, line 51: I am wondering since we have `io.load_context`, we should call this `always_save_context`.

nemo/lightning/pytorch/strategies/fsdp_strategy.py
blahBlahhhJ, line 1: Noticed I got the default value wrong for fsdp `ckpt_include_optimizer`. Could you change it to True to match megatron strategy? (line 61) This should make it work the same as megatron strategy.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -272,26 +273,11 @@ def on_train_end(self, trainer, pl_module):
                     logging.debug(f'Last checkpoint {self.last_model_path} already saved')
                 else:
                     super()._save_last_checkpoint(trainer, monitor_candidates)
+            if self.save_context_on_train_end and not self.always_save_context and is_global_rank_zero():
+                TrainerContext.from_trainer(trainer).io_dump(ckpt_to_dir(self.last_model_path) / "artifacts")
```
marcromeyn, line 277: I would prefer a folder called `context`. In there we could have a folder artifacts where we can put things like tokenizer etc.

nemo/lightning/resume.py
```diff
@@ -169,6 +171,10 @@ def nemo_path(self, model=None) -> Optional[Path]:
         if checkpoint:
             if self.adapter_path:
                 return AdapterPath(checkpoint, adapter_path=Path(self.adapter_path))
+
+            model_weights_path = Path(checkpoint) / AutoResume.WEIGHTS_PATH
+            if os.path.isdir(model_weights_path):
```
mikolajblaz, line 176: Why is it conditional? What's the case where we don't have the `weights` directory?

mikolajblaz (approved):
Thanks, only one small question, otherwise approved.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -420,17 +409,22 @@ def _save_checkpoint(self, trainer: 'pytorch_lightning.Trainer', filepath: str)
...
+                rank_zero_info(f"Saving EMA weights to separate checkpoint {ckpt_filepath}")
+                ckpt_filepath = self._ema_format_filepath(ckpt_filepath)
                 if self.verbose:
-                    rank_zero_info(f"Saving EMA weights to separate checkpoint {filepath}")
-                super()._save_checkpoint(trainer, filepath)
+                    rank_zero_info(f"Saving EMA weights to separate checkpoint {ckpt_filepath}")
+                super()._save_checkpoint(trainer, ckpt_filepath)
             self.remove_checkpoint_unfinished_marker(filepath, barrier_before=True)
         else:
+            ## Do not include optimizer states in final checkpoint
```
mikolajblaz, line 423: Actually, since we have an opportunity to do things "properly" in NeMo 2.0, I would suggest using the flag that was designed to skip optimizer state save, i.e.  the second argument of `trainer.save_checkpoint`. That would require slight changes in the `save_checkpoint` function, i.e. we would have to rely on PTL native mechanisms to create optimizer state (`strategy.optimizer_state` function).

Let me know what you think

nemo/lightning/resume.py
ashors1: I wanted to make `AutoResume` as flexible as possible and wanted to try to account for the case where the user is not using our `ModelCheckpoint`  class. But that codepath has not been tested and would likely have other bugs, so I can remove this conditional to be more explicit.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
ashors1: Good point! I'll look into this today

jstjohn (dismissed):

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -272,26 +274,11 @@ def on_train_end(self, trainer, pl_module):
                     logging.debug(f'Last checkpoint {self.last_model_path} already saved')
                 else:
                     super()._save_last_checkpoint(trainer, monitor_candidates)
+            if self.save_context_on_train_end and not self.always_save_context and is_global_rank_zero():
```
akoumpa, line 277: I'm confused with the docstring and this code. Docstring says save_context_on_train_end does not depend on always_save_context, but looks like it is?

save_context_on_train_end: Whether to dump the artifacts on_train_end **regardless** of whether ``always_save_context`` is ``True``.
            
`if self.save_context_on_train_end **and** not self.always_save_context`

ashors1: That's just because if `always_save_context` is true, the context is already saved to the final checkpoint. That check avoids unnecessarily saving the context twice.

akoumpa: Thanks.

PR #10295, opened 2024-08-28
Add option to selectively restore model weights and optimizer states in AutoResume and MegatronStrategy

hemildesai:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/llm/gpt/data/fine_tuning.py
+++ b/nemo/collections/llm/gpt/data/fine_tuning.py
@@ -54,10 +54,7 @@ class FineTuningDataModule(pl.LightningDataModule):
         self.seq_length = seq_length
         self.seed = seed
         self.dataset_root = Path(dataset_root)
-
-        from nemo.collections.nlp.modules.common.tokenizer_utils import get_nmt_tokenizer
-
-        self.tokenizer = tokenizer or get_nmt_tokenizer("megatron", "GPT2BPETokenizer")
+        self.tokenizer = tokenizer
         self.memmap_workers = memmap_workers
         self.num_workers = num_workers
         self.pin_memory = pin_memory
--- a/nemo/collections/llm/recipes/llama3_8b.py
+++ b/nemo/collections/llm/recipes/llama3_8b.py
@@ -97,7 +97,7 @@ def pretrain_recipe(
 def hf_resume() -> Config[nl.AutoResume]:
     return Config(
         nl.AutoResume,
-        selective_restore_config=Config(nl.SelectiveRestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"),
+        restore_config=Config(nl.RestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"),
     )
 
 
--- a/nemo/collections/llm/recipes/llama3_8b.py
+++ b/nemo/collections/llm/recipes/llama3_8b.py
@@ -96,7 +96,7 @@ def pretrain_recipe(
 def hf_resume() -> Config[nl.AutoResume]:
     return Config(
         nl.AutoResume,
-        restore_config=Config(nl.RestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"),
+        restore_config=Config(nl.RestoreConfig, path="hf://meta-llama/Meta-Llama-3-8B"),
     )
 
 
--- a/nemo/collections/llm/recipes/llama3_8b.py
+++ b/nemo/collections/llm/recipes/llama3_8b.py
@@ -96,7 +96,7 @@ def pretrain_recipe(
 
 
 def hf_resume() -> Config[nl.AutoResume]:
-    return Config(nl.AutoResume, import_path="hf://meta-llama/Meta-Llama-3.1-8B")
+    return Config(nl.AutoResume, restore_path="hf://meta-llama/Meta-Llama-3.1-8B")
 
 
 def finetune_recipe(name: str, ckpt_dir: str, num_nodes: int, num_gpus_per_node: int) -> Partial:
--- a/nemo/collections/llm/recipes/llama3_8b.py
+++ b/nemo/collections/llm/recipes/llama3_8b.py
@@ -95,7 +95,10 @@ def pretrain_recipe(
 
 
 def hf_resume() -> Config[nl.AutoResume]:
-    return Config(nl.AutoResume, selective_restore_config=Config(nl.SelectiveRestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"))
+    return Config(
+        nl.AutoResume,
+        selective_restore_config=Config(nl.SelectiveRestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"),
+    )
 
 
 def finetune_recipe(name: str, ckpt_dir: str, num_nodes: int, num_gpus_per_node: int) -> Partial:
--- a/nemo/collections/llm/recipes/llama3_8b.py
+++ b/nemo/collections/llm/recipes/llama3_8b.py
@@ -45,7 +45,6 @@ def trainer(
         context_parallel_size=context_parallelism,
         sequence_parallel=sequence_parallelism,
         gradient_as_bucket_view=True,
-        ckpt_include_optimizer=True,
         ckpt_async_save=True,
         ckpt_parallel_load=True,
     )
@@ -96,7 +95,7 @@ def pretrain_recipe(
 
 
 def hf_resume() -> Config[nl.AutoResume]:
-    return Config(nl.AutoResume, restore_path="hf://meta-llama/Meta-Llama-3.1-8B")
+    return Config(nl.AutoResume, selective_restore_config=Config(nl.SelectiveRestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"))
 
 
 def finetune_recipe(name: str, ckpt_dir: str, num_nodes: int, num_gpus_per_node: int) -> Partial:
--- a/nemo/collections/llm/recipes/mistral.py
+++ b/nemo/collections/llm/recipes/mistral.py
@@ -32,7 +32,7 @@ def trainer(devices=8) -> nl.Trainer:
 
 @factory(name=NAME + "_hf")
 def hf_resume() -> nl.AutoResume:
-    return nl.AutoResume(selective_restore_config=nl.SelectiveRestoreConfig(path="hf://mistralai/Mistral-7B-v0.3"))
+    return nl.AutoResume(restore_config=nl.RestoreConfig(path="hf://mistralai/Mistral-7B-v0.3"))
 
 
 @factory(name=NAME, for_task="llm.pretrain")
--- a/nemo/collections/llm/recipes/mistral.py
+++ b/nemo/collections/llm/recipes/mistral.py
@@ -32,7 +32,7 @@ def trainer(devices=8) -> nl.Trainer:
 
 @factory(name=NAME + "_hf")
 def hf_resume() -> nl.AutoResume:
-    return nl.AutoResume(import_path="hf://mistralai/Mistral-7B-v0.3")
+    return nl.AutoResume(restore_path="hf://mistralai/Mistral-7B-v0.3")
 
 
 @factory(name=NAME, for_task="llm.pretrain")
--- a/nemo/collections/llm/recipes/mistral.py
+++ b/nemo/collections/llm/recipes/mistral.py
@@ -32,7 +32,7 @@ def trainer(devices=8) -> nl.Trainer:
 
 @factory(name=NAME + "_hf")
 def hf_resume() -> nl.AutoResume:
-    return nl.AutoResume(restore_path="hf://mistralai/Mistral-7B-v0.3")
+    return nl.AutoResume(selective_restore_config=nl.SelectiveRestoreConfig(path="hf://mistralai/Mistral-7B-v0.3"))
 
 
 @factory(name=NAME, for_task="llm.pretrain")
--- a/nemo/lightning/pytorch/strategies/megatron_strategy.py
+++ b/nemo/lightning/pytorch/strategies/megatron_strategy.py
@@ -77,6 +77,7 @@ class ParallelismConfig:
     moe_extended_tp: bool
     pipeline_dtype: torch.dtype
 
+
 class MegatronStrategy(DDPStrategy, io.IOMixin):
     """Megatron plugin for Pytorch Lightning.
 
--- a/nemo/lightning/pytorch/strategies/megatron_strategy.py
+++ b/nemo/lightning/pytorch/strategies/megatron_strategy.py
@@ -615,7 +615,9 @@ class MegatronStrategy(DDPStrategy, io.IOMixin):
         self.checkpoint_io.save_checkpoint(checkpoint, filepath, storage_options=storage_options)
 
     @override
-    def load_checkpoint(self, checkpoint_path: Union[str, Path], include_optimizer_states: bool = False) -> Dict[str, Any]:
+    def load_checkpoint(
+        self, checkpoint_path: Union[str, Path], include_optimizer_states: bool = False
+    ) -> Dict[str, Any]:
         """PTL method which we override to integrate distributed checkpoints for model parallel models.
         In order to load distributed checkpoints we need to provide the sharded_state_dict to
         the distributed load function. We get the sharded_state_dict from self.lightning_module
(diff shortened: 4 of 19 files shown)


nemo/lightning/pytorch/strategies.py
```diff
@@ -628,17 +637,39 @@ def load_checkpoint(self, checkpoint_path: Union[str, Path]) -> Dict[str, Any]:
         sharded_state_dict = {}
         sharded_state_dict["state_dict"] = self.megatron_parallel.sharded_state_dict()
 
-        if self.ckpt_include_optimizer and self.trainer.state.fn == TrainerFn.FITTING:
+        if (self.ckpt_include_optimizer or include_optimizer_states) and self.trainer.state.fn == TrainerFn.FITTING:
```
ashors1, line 640: I feel like it's a bit confusing to have two different arguments with nearly the same name. What about instead we add `ckpt_load_optimizer` and `ckpt_save_optimizer` attributes to the strategy?

nemo/lightning/resume.py
```diff
@@ -1,107 +1,81 @@
...
+    resume_past_end: bool = False
+    resume_ignore_no_checkpoint: bool = False
+    restore_optimizer_states: bool = False
 
     def setup(self, trainer: Union[pl.Trainer, fl.Fabric], model=None):
         if isinstance(trainer, fl.Fabric):
             raise NotImplementedError("Fabric is not supported yet.")
 
-        ckpt_path = self.nemo_path(model)
+        ckpt_path, restore_path = self.nemo_path(model)
```
ashors1, line 60: could we rename these to make it clear that one path is for the optimizer and the other is for the model?

nemo/lightning/pytorch/strategies.py
hemildesai: Yeah I think that should work. I'm ok with that.

nemo/lightning/resume.py
hemildesai: The `ckpt_path` is for `trainer.ckpt_path` whereas the `restore_path` is for selective restore. I can rename `ckpt_path` to `trainer_ckpt_path`.

nemo/lightning/pytorch/strategies.py
ashors1: Good point. What about just renaming `include_optimizer_states` to `load_optimizer_states_from_ckpt` or something similar?

nemo/lightning/resume.py
```diff
@@ -1,107 +1,81 @@
...
+    resume_past_end: bool = False
+    resume_ignore_no_checkpoint: bool = False
+    restore_optimizer_states: bool = False
 
     def setup(self, trainer: Union[pl.Trainer, fl.Fabric], model=None):
         if isinstance(trainer, fl.Fabric):
             raise NotImplementedError("Fabric is not supported yet.")
 
-        ckpt_path = self.nemo_path(model)
+        ckpt_path, selective_restore_path = self.nemo_path(model)
```
yaoyu-33, line 60: can we add a comment explain what are the two returns here?

nemo/lightning/resume.py
```diff
@@ -148,18 +140,52 @@ def nemo_path(self, model=None) -> Optional[Path]:
...
+                    if base_model_path != Path(metadata['model_ckpt_path']):
+                        raise ValueError(
+                            f"When trying to resume a PEFT training run, found mismatching values: "
+                            f"your specified restore_path points to {base_model_path}, "
+                            f"but the PEFT checkpoint was trained with "
+                            f"model_ckpt_path={metadata['model_ckpt_path']}"
+                        )
+
+                    # return AdapterPath(base_model_path, adapter_path=checkpoint), None
+                    return AdapterPath(checkpoint, base_model_path=base_model_path), None
```
hemildesai, line 172: Can you move this to a separate private method?

hemildesai (commented):
@cuichenx Two small comments for the PEFT changes

nemo/lightning/resume.py
```diff
@@ -1,107 +1,97 @@
...
+            restore_path = model.import_ckpt(self.restore_path)
+        except (ValueError, AttributeError):
+            restore_path = self.restore_path
 
-class AutoResume(Resume, io.IOMixin):
-    """Class that handles the logic for setting checkpoint paths and restoring from
-    checkpoints in NeMo.
-    """
+        if self.adapter_path:
+            restore_path = AdapterPath(Path(self.adapter_path), adapter_path=restore_path)
```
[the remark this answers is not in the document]
cuichenx: yes sorry, missed this

nemo/lightning/resume.py
cuichenx: good idea

nemo/collections/llm/gpt/data/fine_tuning.py
```diff
@@ -54,10 +54,7 @@ def __init__(
         self.seq_length = seq_length
         self.seed = seed
         self.dataset_root = Path(dataset_root)
-
-        from nemo.collections.nlp.modules.common.tokenizer_utils import get_nmt_tokenizer
-
-        self.tokenizer = tokenizer or get_nmt_tokenizer("megatron", "GPT2BPETokenizer")
+        self.tokenizer = tokenizer
```
ashors1, line 57: What's the reason for this change?

cuichenx: I don't think it made sense to put a default tokenizer here. Can you tell me your use case? In practice, it was downloading a megatron-345m model every time, so I wanted to prevent that

tests/lightning/test_nemo_logger.py
```diff
@@ -82,21 +82,21 @@ def test_resume(self, trainer, tmp_path):
         # Error because explicit_log_dir does not exist
         with pytest.raises(NotFoundError):
             nl.AutoResume(
-                dirpath=str(tmp_path / "test_resume"),
+                resume_from_directory=str(tmp_path / "test_resume"),
```
yaoyu-33, line 85: why do I only see `resume_from_directory` change in this test file?

nemo/collections/llm/recipes/llama3_8b.py
```diff
@@ -96,7 +95,10 @@ def pretrain_recipe(
 
 
 def hf_resume() -> Config[nl.AutoResume]:
-    return Config(nl.AutoResume, import_path="hf://meta-llama/Meta-Llama-3.1-8B")
+    return Config(
+        nl.AutoResume,
+        selective_restore_config=Config(nl.SelectiveRestoreConfig, path="hf://meta-llama/Meta-Llama-3.1-8B"),
```
cuichenx, line 100: This didn't originate from this PR, but this hf path should be `hf://meta-llama/Meta-Llama-3-8B` to match the `Llama3Config8B` above. (Alternatively change above to `Llama31Config8B`, and same for other recipe files.

cuichenx: let me know if you think this is okay, i can also make the change

hemildesai: Oh yeah feel free to change, I must've missed this.

nemo/collections/llm/gpt/data/fine_tuning.py
ashors1: Makes sense, was just wondering why this was added to this PR. The tokenizer is no longer optional now, right? Should we change the type hint too?

nemo/lightning/pytorch/strategies/utils.py
```diff
@@ -21,6 +22,14 @@
 from nemo.utils.callbacks.dist_ckpt_io import AsyncFinalizableCheckpointIO
 
 
+@dataclass(kw_only=True)
+class SelectiveRestoreConfig:
```
akoumpa, line 26: `RestoreConfig` seems like a better option, "Config" conveys the selective part by definition.

nemo/lightning/pytorch/strategies/utils.py
```diff
@@ -21,6 +22,14 @@
...
+    path: str
+    adapter_path: Optional[str] = None
+    model_weights: bool = True
```
akoumpa, line 29: How about `load_model_state` & `load_optim_state`?

I want to avoid a verb here but also make sure the options aren't confused with the actual thing.

nemo/collections/llm/recipes/mistral.py
```diff
@@ -32,7 +32,7 @@ def trainer(devices=8) -> nl.Trainer:
 
 @factory(name=NAME + "_hf")
 def hf_resume() -> nl.AutoResume:
-    return nl.AutoResume(import_path="hf://mistralai/Mistral-7B-v0.3")
+    return nl.AutoResume(selective_restore_config=nl.SelectiveRestoreConfig(path="hf://mistralai/Mistral-7B-v0.3"))
```
akoumpa, line 35: why isn't this `Config(nl.SeclectiveRestoreConfig ...` ?

nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -602,13 +611,19 @@ def save_checkpoint(
...
             checkpoint["sharded_state_dict"] = self.megatron_parallel.sharded_state_dict()
-        if self.trainer.state.fn == TrainerFn.FITTING and self.ckpt_include_optimizer:
+        if self.trainer.state.fn == TrainerFn.FITTING and self.ckpt_save_optimizer:
             checkpoint["optimizer"] = [self.optimizer_sharded_state_dict()]
 
         self.checkpoint_io.save_checkpoint(checkpoint, filepath, storage_options=storage_options)
 
+    def should_restore_optimizer_states(self, selective_restore: bool = False) -> bool:
+        if selective_restore:
+            return self.selective_restore_config.optimizer_states if self.selective_restore_config else False
```
akoumpa, line 621: Would that cause any trouble? We are using m argument and a class attribute's value to make a decision. 

Is there any way to make this only use the argument?

nemo/lightning/resume.py
```diff
@@ -1,183 +1,209 @@
...
     BasePath = WindowsPath
 else:
     BasePath = PosixPath
 
 
-class Resume(IOMixin):
-    def nemo_path(self, model=None) -> Optional[Path]:
-        """Returns the checkpoint to resume from."""
+@dataclass(kw_only=True)
+class AutoResume:
```
akoumpa, line 24: we need to rename this class, but I'm not sure it should happen in this PR.

nemo/lightning/resume.py
```diff
@@ -1,183 +1,209 @@
...
+            trainer.strategy.selective_restore_config = self.selective_restore_config
+
+    def _try_import_model(
+        self, model: Optional[io.ConnectorMixin], path: str, adapter_path: Optional[str] = None
+    ) -> BasePath:
+        if model is None:
+            raise ValueError("Model is needed to import checkpoint from HF or other non-NeMo checkpoint format.")
+        try:
+            new_path = model.import_ckpt(path)
+        except (ValueError, AttributeError):
```
akoumpa, line 83: what triggers the ValueError / AttributeError here?

nemo/lightning/resume.py
```diff
@@ -1,183 +1,209 @@
...
-                    checkpoint = last_checkpoints[0]
-                    checkpoint = uninject_model_parallel_rank(checkpoint)
-                else:
-                    # Select the checkpoint with the latest modified time
-                    checkpoint = sorted(last_checkpoints, key=lambda pth: pth.lstat().st_mtime, reverse=True)[0]
-                    logging.warning(
-                        f"Multiple checkpoints {last_checkpoints} matches *last.ckpt. Selecting one with the latest modified time."
-                    )
+        elif len(end_checkpoints) > 0:
+            if self.resume_past_end:
```
akoumpa, line 149: make this `if not self.resume_past_end`, raise the exception and move the nested block one level up.

nemo/lightning/resume.py
```diff
@@ -1,183 +1,209 @@
...
+            True and a checkpoint matching ``*end.ckpt`` indicating a previous training run
+            fully completed. Setting resume_past_end=True disables this behavior and loads the
+            last checkpoint.
+        resume_ignore_no_checkpoint (bool): AutoResume throws an error if resume_if_exists is
+            True and no checkpoint could be found. Setting resume_ignore_no_checkpoint=True
+            disables this behavior, in which case exp_manager will print a message and
+            continue without restoring.
+    """
+
+    selective_restore_config: Optional[SelectiveRestoreConfig] = None
```
akoumpa, line 48: Would it be possible to rename this `restore_config` (prev. comment) and also allow passing a string, in which case a `SelectiveRestoreConfig` is created in the constructor and restores everything the checkpoint contains?

hemildesai: I had named it `resotre_` before but @ashors1 suggested to use `selective_restore` to make it more explicit. I don't have a strong preference for either, but I don't think being explicit hurts in this case.

nemo/lightning/resume.py
hemildesai: It's for when the connector is not found for the particular path.

nemo/lightning/resume.py
```diff
@@ -1,183 +1,209 @@
...
-        self.resume_if_exists = resume_if_exists
-        self.resume_past_end = resume_past_end
-        self.resume_ignore_no_checkpoint = resume_ignore_no_checkpoint
-
-    def nemo_path(self, model=None) -> Optional[Path]:
+        trainer_ckpt_path = self.get_trainer_ckpt_path(model)
+        if trainer_ckpt_path:
+            trainer.ckpt_path = trainer_ckpt_path
+            trainer.checkpoint_callback.last_model_path = trainer_ckpt_path
+        elif self.selective_restore_config:
```
akoumpa, line 63: I think there's a bug here, my suggestion is to change `elif` to `if`.

akoumpa (changes requested):

nemo/lightning/resume.py
hemildesai: This is intended as you cannot have both together. Basically, you want to skip selective restore if checkpoint path is found.

nemo/lightning/pytorch/strategies/megatron_strategy.py
hemildesai: It's not possible to only use the argument since the helper is called by methods like `load_checkpoint` which are invoked by internal PTL APIs without any args. The arg `selective_restore` just tells the helper whether you're in the selective restore stage or not. Open to suggestions, but couldn't find a better way to accomplish this.

nemo/lightning/pytorch/strategies/utils.py
hemildesai: Yeah I'm fine with the rename, will do.

nemo/lightning/resume.py
hemildesai: Raising an error would not support the following usecase:
1. Start finetuning from SelectiveRestoreConfig
2. Resume finetuning from the last saved checkpoint by 1.

nemo/lightning/resume.py
akoumpa: @hemildesai thanks, can you add include your GH comment as a code comment?

nemo/lightning/pytorch/strategies/utils.py
akoumpa: thank you.

nemo/lightning/resume.py
akoumpa: I don't see the value of `Selective` here.

hemildesai: @ashors1 @yaoyu-33 Are you fine with just `RestoreConfig`?

nemo/collections/llm/gpt/data/fine_tuning.py
cuichenx: A user doesn't need to specify this, because the llm.finetune API will link the tokenizer from the imported model

nemo/collections/llm/recipes/mistral.py
hemildesai: Looks like the recipes in this file are not updated. MIght require a separate PR to update all recipes here.

nemo/lightning/resume.py
hemildesai: Updated to `RestoreConfig` in https://github.com/NVIDIA/NeMo/pull/10295/commits/f240efad57edf08aeaa2d9b2bfc73724b865ca6d

nemo/collections/llm/recipes/llama3_8b.py
hemildesai: Updated.

akoumpa (dismissed):
LGTM, thanks.

akoumpa (approved):

PR #10511, opened 2024-09-17
Add slice_with_offset and dry_run Support for Tar Dataset Creation; New Script for Partial Conversion

ssh-meister:
# What does this PR do ?

1. `--slice_with_offset`:
Added support to create tar datasets from audio sample segments when the slice_by_offset flag is enabled and the manifest contains an offset field. This feature ensures that audio segments are correctly sliced according to specified offsets.
2. `--only_manifests`:
If set, only creates manifests for each shard without creating the actual tar files. This allows you to verify the output structure and content before committing to the full tarball creation process.
3. `--dry_run`:
Run in simulation mode: calculate and display the number of shards and estimated data per shard without reading audio files or writing any output.
4. New Script - `partial_conversion_to_tarred_audio_dataset.py`:
This script facilitates the creation of tarred and sharded audio datasets from existing tarred manifests. It allows you to select specific shards from a manifest file and then tar them separately. 
This is useful in several scenarios:
- When you only need to process a specific subset of shards (e.g., for debugging or incremental dataset preparation).
- When you want to parallelize shard creation across multiple SLURM jobs to accelerate the dataset generation process and overcome per-job time limits.

# Changelog 
- Functions:
-- `create_shards`: Updated to support the `slice_with_offset` flag for segment-based tar creation and `--only_manifests` functionality.
-- Added new functionality to handle sharded manifests when `--only_manifests` is enabled.
- New Script:
-- `partial_conversion_to_tarred_audio_dataset.py:` Processes dry-run manifests to generate individual shards.
  
**PR Type**:
- [v] New Feature

# Additional Information
* These changes enhance flexibility and control over the tar dataset creation process, allowing for more granular and controlled dataset management.


scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -644,17 +568,28 @@ def _read_manifest(self, manifest_path: str, config: ASRTarredDatasetConfig):
...
             audio, sampling_rate = soundfile.read(audio_filepath, dtype=np.float32)
+            # Calculate start and end points for slicing
+            start_sample = int(offset * sampling_rate)
+            end_sample = int((offset + duration) * sampling_rate) if duration else None
+            audio = audio[start_sample:end_sample]
+            # Transcode and write to tar
             encoded_audio = BytesIO()
-            if codec == "opus":
-                kwargs = {"format": "ogg", "subtype": "opus"}
+            if codec is not None:
```
pzelasko, line 586: Actually we could do a fast-path here: `if codec=None and offset=0 and duration=None`, instead of reading+writing with soundfile (triggering encoding and decoding) just read the raw audio file bytes and write them to tar file directly. Will likely speed up data tarring by a non-trivial factor. Up to you.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -644,17 +568,28 @@ def _read_manifest(self, manifest_path: str, config: ASRTarredDatasetConfig):
 
         return entries, total_duration, filtered_entries, filtered_duration
 
-    def _write_to_tar(self, tar, audio_filepath: str, squashed_filename: str) -> None:
-        if (codec := self.config.force_codec) is None or audio_filepath.endswith(f".{codec}"):
+    def _write_to_tar(
+        self, tar, audio_filepath: str, squashed_filename: str, duration: float = None, offset: float = 0
+    ) -> None:
+        if ((codec := self.config.force_codec) is None or audio_filepath.endswith(f".{codec}")) and not duration:
```
pzelasko, line 574: There are so many if/else here now that I think we should have some unit test coverage for this script (or just the function).

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -798,27 +791,151 @@ def create_tar_datasets(min_duration: float, max_duration: float, target_dir: st
...
+        "--force_codec",
+        type=str,
+        default=None,
+        help=(
+            "If specified, transcode the audio to the given format. "
+            "Supports libnsndfile formats (example values: 'opus', 'flac')."
+        ),
+    )
+    parser.add_argument(
+        "--dry_run",
```
pzelasko, line 932: `dry_run` is not an accurate name since the script will create some output: the manifests without tar files. I suggest renaming it to `--only_manifest` instead. 

BTW `dry_run` would also be a nice option, telling you very quickly how many shards with how much data per shard are going to be created without actually reading any audio or writing anything.

scripts/speech_recognition/partial_convertion_to_tarred_audio_dataset.py
```diff
@@ -0,0 +1,196 @@
...
+import hydra
+from convert_to_tarred_audio_dataset import ASRTarredDatasetBuilder, ASRTarredDatasetMetadata
+from hydra.core.config_store import ConfigStore
+from joblib import Parallel, delayed
+from omegaconf import MISSING
+
+"""
+# Partial Tarred Audio Dataset Creator
+
+## Overview
```
pzelasko, line 29: I'm not sure I understand the purpose of this script. Why would you want to only tar specific shards? Can you add some comments / docs about the intended use cases?

Jorjeous:
@ssh-meister

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -644,39 +601,69 @@ def _read_manifest(self, manifest_path: str, config: ASRTarredDatasetConfig):
...
+
+        tar_filepath = os.path.join(target_dir, f'audio_{shard_id}.tar')
+        if not only_manifests:
+            tar = tarfile.open(tar_filepath, mode='w', dereference=True)
 
         count = dict()
-        for entry in entries:
+        for entry in tqdm(entries, desc="Creating shard.."):
             # We squash the filename since we do not preserve directory structure of audio files in the tarball.
             if os.path.exists(entry["audio_filepath"]):
```
karpnv, line 661: Please add or only_manifests
`if os.path.exists(entry["audio_filepath"]) or only_manifests:`

scripts/speech_recognition/partial_convertion_to_tarred_audio_dataset.py
```diff
@@ -0,0 +1,199 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
pzelasko, line 1: Typo in script's name: `convertion` -> `conversion`

scripts/speech_recognition/partial_convertion_to_tarred_audio_dataset.py
```diff
@@ -0,0 +1,199 @@
...
+                if entry['shard_id'] not in entries_to_shard:
+                    entries_to_shard[entry['shard_id']] = []
+
+                if slice_with_offset:
+                    if 'abs_audio_filepath' not in entry or 'source_audio_offset' not in entry:
+                        raise KeyError(
+                            f"`slice_with_offset` is enabled, but `abs_audio_filepath` and/or `source_audio_offset` are not found in the entry:\n{entry}."
+                        )
+                    entry['audio_filepath'] = entry['abs_audio_filepath']
+                    entry['offset'] = entry['source_audio_offset']
```
pzelasko, line 105: should we delete `abs_audio_filepath` and `source_audio_offset` keys from the entry, as they are no longer useful at this point?

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -644,39 +601,69 @@ def _read_manifest(self, manifest_path: str, config: ASRTarredDatasetConfig):
...
-            encoded_audio = BytesIO()
+            return
+
+        # Standard processing: read, trim, and transcode the audio file
+        audio, sampling_rate = soundfile.read(audio_filepath, dtype=np.float32)
+
+        # Trim audio based on offset and duration.
+        start_sample = int(offset * sampling_rate)
+        end_sample = int((offset + duration) * sampling_rate) if duration else None
+        audio = audio[start_sample:end_sample]
```
pzelasko, line 622: Use `frames` and `start` arguments to `soundfile.read` to avoid reading the entire audio into memory. It will drastically speed up the script for long-form audio with offsets. 

https://github.com/bastibe/python-soundfile/blob/5069b8cb828eb2bb26aba0620e1f7ea2ae37f189/soundfile.py#L253-L257

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -798,27 +856,159 @@ def create_tar_datasets(min_duration: float, max_duration: float, target_dir: st
...
+
+    parser.add_argument(
+        "--keep_files_together",
+        action='store_true',
+        help="Whether or not to keep entries from the same file (but different offsets) together when sorting before tarring/sharding.",
+    )
+    parser.add_argument(
+        "--slice_with_offset",
+        action='store_true',
+        help="If set, only slices the audio based on `duration` and `offset` entry parameters.",
```
pzelasko, line 950: and what happens when it is unset?

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -686,14 +673,51 @@ def _create_shard(self, entries, target_dir, shard_id, manifest_folder):
...
+            if self.config.slice_with_offset:
+                if squashed_filename not in count:
+                    count[squashed_filename] = 1
+
+                entry_offset = str(entry['offset']).split('.')
+                if len(entry_offset) == 1:
+                    entry_offset.append('0')
+                elif len(entry_offset) > 2:
+                    raise ValueError(
+                        f"The offset for the entry with audio_filepath '{entry['audio_filepath']}' is incorrectly provided ({entry['offset']})."
```
pzelasko, line 686: What is the correct format? A comment here with an example + clarifying message in the exception would be helpful.

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -798,27 +856,159 @@ def create_tar_datasets(min_duration: float, max_duration: float, target_dir: st
...
+            "If specified, transcode the audio to the given format. "
+            "Supports libnsndfile formats (example values: 'opus', 'flac')."
+        ),
+    )
+    parser.add_argument(
+        "--only_manifests",
+        action='store_true',
+        help=(
+            "If set, only creates manifests for each shard without creating the actual tar files. "
+            "This allows you to verify the output structure and content before committing to the full tarball creation process."
```
pzelasko, line 1002: Please document the extra fields that are added to the manifest when this option is specified (eg ~abs_filepath)

pzelasko (changes requested):
Thanks! I left some suggestions to improve the documentation and speed

pzelasko (approved):
LGTM

scripts/speech_recognition/convert_to_tarred_audio_dataset.py
```diff
@@ -686,14 +670,56 @@ def _create_shard(self, entries, target_dir, shard_id, manifest_folder):
...
+                del entry['offset']
+            else:
+                if squashed_filename not in count:
+                    if not only_manifests:
+                        self._write_to_tar(tar, audio_filepath, squashed_filename)
+                    to_write = squashed_filename
+                    count[squashed_filename] = 1
+                else:
+                    to_write = base + "-sub" + str(count[squashed_filename]) + ext
+                    count[squashed_filename] += 1
```
nithinraok, line 718: Do we need this now? Previously we used -sub for writing offset audio files, and now you are giving the script an explicit option which was handled in above case.

nithinraok: The reason for sub is, in nemo tarred datasets, I believe we use this to check if we need to segment the file with  an offset or not. I would recommend you to add a test with a tarred set created with offset with updated set and make sure dataloading is working as intedted with both lhotse and nemo tarred dataset

scripts/speech_recognition/partial_conversion_to_tarred_audio_dataset.py
```diff
@@ -0,0 +1,199 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
nithinraok, line 1: update year.

pzelasko (approved):

PR #10548, opened 2024-09-20
[NeMo-UX] Support `save_last="link"`

ashors1:
# What does this PR do ?
Adds support for creating a symlink for `-last` checkpoints. Implementation is compatible with synchronous and asynchronous checkpointing.

**Collection**: llm

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -395,6 +402,27 @@ def _monitor_candidates(self, trainer: "pl.Trainer") -> Dict[str, torch.Tensor]:
...
 
+    def _link_checkpoint(self, trainer: "pl.Trainer", filepath: str, linkpath: str, override_async=False) -> None:
+        ## linking will happen as part of the finalize fn
+        if self.async_save and not override_async:
+            self.ckpts_to_link[str(filepath)] = str(linkpath)
+            return
+
+        filepath = ckpt_to_dir(filepath)
+        linkpath = ckpt_to_dir(linkpath)
+        if trainer.is_global_zero:
```
mikolajblaz, line 413: Can we call super()._link_checkpoint for everything below 413?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -450,10 +484,42 @@ def _save_checkpoint(self, trainer: 'pytorch_lightning.Trainer', filepath: str)
                 TrainerContext.from_trainer(trainer).io_dump(ckpt_to_dir(filepath) / "context")
 
             if self.async_save:
+                self._last_checkpoint_saved = filepath
                 logging.info(f'Scheduled async checkpoint save for {filepath}')
             else:
                 finalize_fn()
 
+    def _save_last_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, torch.Tensor]) -> None:
```
mikolajblaz, line 492: I'm wondering if there is some way to avoid overriding the whole method. This is always risky, since we lose touch with the upstream.

How is our flow different from the one in PTL which makes us add `saved_current_step` logic and also not rely on `self.last_model_path`?
Is it because PTL links to _any_ available last checkpoint (not necessarily from the last iteration)?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -81,11 +81,18 @@ def __init__(
         self.save_context_on_train_end = save_context_on_train_end
         self.save_optim_on_train_end = save_optim_on_train_end
 
+        ## stores the previously saved -last checkpoint, used only when save_last = 'link'
+        ## this is needed because when using symlinks, we need to update last_model_path when
+        ## saving the non-last checkpoint. Storing "previous" is needed to keep us from overwriting the
+        ## previous "last_model_path"
+        self.previous = ""
```
mikolajblaz, line 88: Can you give a more descriptive name?
E.g. `previous_last_model_path`

nemo/lightning/pytorch/callbacks/model_checkpoint.py
ashors1: Yeah, I made two changes:
1. Made sure to add a symlink only when the current step was actually saved. As you suggested, PTL always links to the last checkpoint saved, which might not correspond to the latest step
2. Added [these lines](https://github.com/NVIDIA/NeMo/blob/200cae3a3794f539db80186c1c5fc40ca9723995/nemo/lightning/pytorch/callbacks/model_checkpoint.py#L505-L508) which fix `last_model_path` saved to the `*-last` checkpoint state dict when using symlinks

I'll think about whether we can make these fixes without overwriting the entire `_save_last_checkpoint` method

mikolajblaz: Thanks. Maybe overwriting save_last_checkpoint is inevitable in which case current version is ok

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -395,6 +408,25 @@ def _monitor_candidates(self, trainer: "pl.Trainer") -> Dict[str, torch.Tensor]:
...
+
+        ## linking will happen as part of the finalize fn
+        if self.async_save and not override_async:
+            self.ckpts_to_link[str(filepath)] = str(linkpath)
+            return
+
+        filepath = ckpt_to_dir(filepath)
+        linkpath = ckpt_to_dir(linkpath)
+        super()._link_checkpoint(trainer, filepath, linkpath)
+
```
athitten, line 429: Is there a way to avoid overriding PTL's `_link_checkpoint` method ? We want to avoid overriding PTL's private methods to have stable code.

ashors1: I can think about it, but it might be challenging to support linking with async checkpointing without overwriting this method. Also, the addition of `saved_current_step` is needed to fix a bug that seems to exist in PTL's link implementation in which the `-last` checkpoint gets linked to the most recently saved checkpoint, even if that corresponds to a different step.

athitten: I see, then maybe we should file an issue with [PTL issues](https://github.com/Lightning-AI/pytorch-lightning/issues) and ask them to fix this bug ? That way it can save us from overriding private methods.

ashors1: Yeah it would be great if PTL could fix that issue! But we'd still have to figure out how to handle async checkpointing, and I do think that would require us to overwrite either `_link_checkpoint` or `_save_last_checkpoint` (where `_link_checkpoint` is invoked)

mikolajblaz: Agree that for handling async save we have to override `_link_checkpoint` anyway.
But since we call `super()._link_checkpoint` I don't think there is too much risk connected with that

nemo/lightning/pytorch/callbacks/model_checkpoint.py
mikolajblaz: Great, thanks, this is great.

Do you know how `last_model_path` is used during restart? I'm wondering if it will be valid if e.g. failure happens between the regular and "last" ckpt save

ashors1: I think `last_model_path` is only used when removing the previous `-last` model to ensure we only retain a single `-last` checkpoint. If failure happens between the regular and last checkpoint save, I **don't** think the state dict will be valid, but I also don't think this is a concern, because we'd end up restoring from the previously saved `-last` checkpoint which does have the correct state dict.

tests/lightning/pytorch/callbacks/test_model_checkpoint.py
```diff
@@ -0,0 +1,228 @@
...
+
+        with reset_megatron_parallel_state():
+            tmp_path = tmpdir / "link_ckpt_test"
+            data, model, trainer = setup_test(tmp_path, async_save=False)
+
+            trainer.fit(model, data)
+
+            checkpoint_dir = Path(tmp_path / "default" / "checkpoints")
+            dist_checkpoints = [d for d in list(checkpoint_dir.glob("*")) if d.is_dir()]
+            last_checkpoints = [d for d in dist_checkpoints if d.match("*last")]
```
athitten, line 202: Thank you @ashors1 for adding the test. I think this does the job. Can we also add an assert line ensuring there are only `save_top_k` number of checkpoints in the dir.

ashors1: Yep, done

mikolajblaz (approved):

athitten (approved):
LGTM, thank you @ashors1

PR #10552, opened 2024-09-20
fix type error in llm collection

stevehuang52:
# What does this PR do ?

fix type error in llm collection


nemo/lightning/io/mixin.py
```diff
@@ -326,6 +326,7 @@ def _get_connector(cls, ext, path=None, importer=True, **kwargs) -> ModelConnect
                         when required.
         """
         _path = None
+        ext = str(ext)
```
maanug-nv, line 329: Docstring already suggests this should be a string, but no harm in this I guess. Can we add type hints for the args as well?

stevehuang52: Added

maanug-nv: sorry for nitpick, can you update docstring type of `path` to match signature typehint (or vice versa), for consistency.

stevehuang52: Done~

maanug-nv (approved):
All concerns resolved

PR #10644, opened 2024-09-26
fix codec/asr/mos model loading for validation

ehsanvidia:
# What does this PR do ?

When using tensor parallelism, model assumes Nemo codec/ASR/MOS checkpoints are parallelized. This PR will fix loading checkpoints when tp is enabled 

## Who can review?

@zhehuaichen

### DIFF
--- a/nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
+++ b/nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
@@ -284,7 +284,7 @@ class S2sModularAudioGPTModel(ModularAudioGPTModel):
         trainer: Optional[Trainer] = None,
     ):
 
-        model, codec_model, asr_model, mos_model = super().restore_from_pretrained_models(cfg, trainer)
+        # model, codec_model, asr_model, mos_model = super().restore_from_pretrained_models(cfg, trainer)
         model, codec_model, asr_model = super().restore_from_pretrained_models(cfg, trainer)
         if cfg.model.get('salm_model_path') is not None:
             torch_state_dict = torch.load(cfg.model.get('salm_model_path'))['state_dict']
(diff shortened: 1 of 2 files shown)


nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
```diff
@@ -283,17 +283,23 @@ def restore_from_pretrained_models(
...
+        model, codec_model, asr_model = super().restore_from_pretrained_models(cfg, trainer)
         if cfg.model.get('salm_model_path') is not None:
             torch_state_dict = torch.load(cfg.model.get('salm_model_path'))['state_dict']
             model.setup_complete = False
             model.load_state_dict(torch_state_dict, strict=False)
             logging.info(f"loading from {cfg.model.get('salm_model_path')}: {torch_state_dict.keys()}")
 
         model.padded_vocab_size = cfg.model.s2s_vocab_size
-        model.model.extend_embedding(model.padded_vocab_size)
+        model.model.module.extend_embedding(model.padded_vocab_size)
```
zhehuaichen, line 296: this will only work for amp
can u wrap these using if self.megatron_amp_O2 and keep the original code for False case
same for the following

nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
```diff
@@ -414,7 +420,7 @@ def parse_decoder_outputs(
             speech_tokens = decoder_output[:, 1:]
 
         # Get speech token ids
-        n_speech_codebooks = self.model.n_proj_heads - 1
+        n_speech_codebooks = self.model.module.n_proj_heads - 1
```
zhehuaichen, line 423: same here

nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
```diff
@@ -816,7 +806,7 @@ def write_predictions_to_file(self, outputs, output_file_path_prefix, output_dir
     def de_concat_multiproj_logits(self, logits):
         logits_list = []
         prev = 0
-        for i in self.model.proj_head_dims:
+        for i in self.model.module.proj_head_dims:
```
zhehuaichen, line 809: same here.
can u confirm you change does not break amp=False case

ehosseiniasl: confirmed

nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
ehosseiniasl: confirmed

nemo/collections/multimodal/speech_llm/models/modular_s2s_models.py
ehosseiniasl: fixed

ehosseiniasl:
for using MOS model, a container with torchaudio should be used. otherwise, comment MOS model loading and inference

PR #10758, opened 2024-10-04
ci: Improve VM maintenance

ko3n1g:
# What does this PR do ?

* Infers the list of VMs to check via the rest api. Another benefit: We can filter for runners that are online!
* Runs a check to decide if a reboot is necessary or not
* Transposes the execution order:  
  Prev. we worked in stages (first: shutdown of _all_ runners, second: re-install drivers) that all runners had to pass before the next stage could start. So all runners were delayed by the slowest runner, and so were just idle. Now, we do this on a one-by-one basis, so each runner will be maintained independently of other runners.
* Run this healthy check every 4hrs, b/c why not? It finishes super quickly (less than 5minutes) and is good for body and soul 

🚨 This PR is in draft because of `environment: ko3n1g/ci/improve-vm-maintenance`. This was necessary to safely test this workflow. After an approval (just write a comment), I will change this to `environment: main` and merge it.

PS: This workflow requires sensitive secrets that we want to expose only to protected branches. So I added a branch protection rule for this branch (`ko3n1g/ci/improve-vm-maintenance`) such that I can safely test this feature.


ko3n1g:
Successful run: https://github.com/NVIDIA/NeMo/actions/runs/11178779025/job/31077046171

.github/workflows/node-reboot-single-vm.yml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - name: Check status
+        id: status
+        run: |
+          docker run --rm --runtime=nvidia --gpus all ubuntu nvidia-smi
+
+          NUM_GPUS=$(nvidia-smi --query-gpu=name --format=csv,noheader | wc -l)
+
+          if [[ $NUM_GPUS -ne ${{ inputs.n_gpus }} ]]; then
+            echo "Issues with GPU detected, will attempt a reboot & re-install."
```
pablo-garay, line 30: I guess this message needs to be changed now since not rebooting

.github/workflows/node-reboot-single-vm.yml
```diff
@@ -0,0 +1,54 @@
...
+        type: string
+        description: Name of VM
+        required: true
+      n_gpus:
+        type: string
+        description: Number of GPUs this VM has
+        required: true
+        
+jobs:
+  check-status-and-maybe-reboot:
```
pablo-garay, line 16: I guess this job name needs to be changed since not rebooting

pablo-garay (changes requested):

PR #10759, opened 2024-10-04
multiturn training support for SALM 

zhehuaichen:
## Manifest format
```
{
  “id”: “some_id_14323452”,
  "conversations": [
    {
      "value": "Can you help summarize the following?",
      "from": "User",
      "type": "text",
    },
    {
      "value": "123.wav",
      "from": "User",
      "type": "audio",
      “duration”: 5.73,
    },
    {
      "value": "I'm glad to assist you with your request. The information can be summarized as the following: ….",
      "from": "Assistant",
      "type": "text",
    },
    {
      "value": "123_answer.wav",
      "from": "Assistant",
      "type": "audio",
      “duration”: 7.11,
    },
    {
      "value": "Can you further shorten it?",
      "from": "User",
      "type": "text",
    },
    …
}
```
## Input CFG format
```
 - type: group
   input_cfg:
    - manifest_filepath: /media/data/datasets/LibriSpeech/dev_clean_10.conv.json
      type: multimodal_conversation
      audio_locator_tag: "<unused0>"
```
- Note:  we need seprate ICFG for single turn and multi turn data mixing because we need to measure length / calibrate batch size differently.
##  Example usage
- check here https://docs.google.com/document/d/15BppG_IyP7DWRh5Lw20bRVHdOXt1uvfhJZhIVmzh8AQ/edit#bookmark=id.5i9bljw374yo mainly needs to add the following flags
```
    ++model.data.validation_ds.force_finite=True \
    ++model.data.train_ds.prompt_format="gemma" \
    ++model.data.validation_ds.prompt_format="gemma" \
    ++model.data.validation_ds.drop_last=False \
    ++model.data.train_ds.drop_last=False \
    ++model.data.train_ds.input_cfg=manifests/train2.conv2.yaml \
    ++model.data.validation_ds.input_cfg=manifests/train2.conv2.yaml
```
## TODO
- Tar/sharding the audios
- BESTOW support https://github.com/zhehuaichen/NeMo/tree/speechllm-develop-prompt

### DIFF
--- a/examples/multimodal/speech_llm/conf/bestow/modular_audio_gpt_config_cross_llama_lhotse_multi.yaml
+++ b/examples/multimodal/speech_llm/conf/bestow/modular_audio_gpt_config_cross_llama_lhotse_multi.yaml
@@ -1,0 +1,314 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: megatron_audio_gpt_bestow_lhotse
+
+trainer:
+  devices: 1
+  accelerator: gpu
+  num_nodes: 1
+  precision: 16
+  logger: False # logger provided by exp_manager
+  enable_checkpointing: False
+  use_distributed_sampler: False
+  max_epochs: 9999
+  max_steps: 1000000 # consumed_samples = global_step * micro_batch_size * data_parallel_size * accumulate_grad_batches
+  limit_train_batches : 1000
+  log_every_n_steps: 10 # frequency with which training steps are logged 
+  val_check_interval: 1000 # If is an int n > 1, will run val every n training steps, if a float 0.0 - 1.0 will run val every epoch fraction, e.g. 0.25 will run val every quarter epoch
+  gradient_clip_val: 1.0
+  accumulate_grad_batches: 1
+
+model_target: nemo.collections.multimodal.speech_llm.models.modular_models.CrossAttendModularAudioGPTModel
+
+exp_manager:
+  # explicit_log_dir: null
+  exp_dir: null
+  name: ${name}
+  create_wandb_logger: False
+  wandb_logger_kwargs:
+    project: null
+    name: null
+  resume_if_exists: True
+  resume_ignore_no_checkpoint: True
+  create_checkpoint_callback: True
+  checkpoint_callback_params:
+    monitor: validation_${model.data.validation_ds.metric.name}
+    save_top_k: 1
+    mode: min
+    save_nemo_on_train_end: True
+    filename: '${name}--{${exp_manager.checkpoint_callback_params.monitor}:.3f}-{step}-{epoch}'
+    model_parallel_size: ${model.tensor_model_parallel_size}
+    always_save_nemo: False
+    save_best_model: True
+  create_early_stopping_callback: False
+  early_stopping_callback_params:
+    monitor: "val_loss"
+    mode: "min"
+    min_delta: 0.001
+    patience: 10
+    verbose: True
+    strict: False # Should be False to avoid a runtime error where EarlyStopping says monitor is unavailable, which sometimes happens with resumed training.
+
+
+model:
+  seed: 1234
+  tensor_model_parallel_size: 1 # intra-layer model parallelism
+  pipeline_model_parallel_size: 1 # inter-layer model parallelism
+  
+  pretrained_audio_model: stt_en_fastconformer_transducer_large
+  freeze_llm: True
+  freeze_audio_encoder: False
+  freeze_modality_adapter: False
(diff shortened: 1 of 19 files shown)


examples/multimodal/speech_llm/conf/bestow/modular_audio_gpt_config_cross_llama_lhotse_multi.yaml
```diff
@@ -0,0 +1,314 @@
...
+        shuffle: true
+        shuffle_buffer_size: ???
+        text_field: answer
+        pin_memory: true
+        max_duration: 40  # it is set for LibriSpeech, you may need to update it for your dataset
+        min_duration: 0.1
+      multiturn:
+        input_cfg: ???
+        use_multimodal_sampling: true
+        batch_size: 1
```
pzelasko, line 258: 1 seems radical, wouldn't it work with anything larger than that?

nemo/collections/common/prompts/gemma.py
```diff
@@ -37,7 +37,7 @@ class GemmaPromptFormatter(PromptFormatter):
 
 
 @registered_prompt_format_fn
-def gemma1(cuts: CutSet, tokenizer: TokenizerSpec):
+def gemma(cuts: CutSet, tokenizer: TokenizerSpec):
     prompt = GemmaPromptFormatter(tokenizer)
     ans = defaultdict(list)
     for cut in cuts:
```
pzelasko, line 43: below (l.54) there's a system prompt support that should be removed - this was copied from llama2, but gemma doesn't have a system turn.

nemo/collections/multimodal/speech_llm/data/lhotse_dataset.py
```diff
@@ -124,9 +132,96 @@ def __getitem__(self, all_cuts: CutSet) -> dict[str, torch.Tensor | list[str] |
...
+                if truncation_length < len(ans.answer_ids):
+                    ans.answer_ids = ans.answer_ids[:-truncation_length]
+                else:
+                    logging.warning(
+                        f'Input ids length {len(ans.input_ids)} exceed max sequence length {self.max_seq_length} {truncation_length} > {len(ans.answer_ids)} may cause losing audio context'
+                    )
+                    ans.answer_ids = ans.answer_ids[: -min(truncation_length, len(ans.answer_ids))]
+                    ans.context_ids = ans.context_ids[: -min(truncation_length, len(ans.context_ids))]
+            # add eos
+            eos_id = self.text_processor.tokenizer.eos_id
```
pzelasko, line 150: do we need to `assert eos_id is not None and eos_id > -1, f"The tokenizer does not have a defined EOS ID (we got {eos_id})"`

nemo/collections/multimodal/speech_llm/data/lhotse_dataset.py
```diff
@@ -124,9 +132,96 @@ def __getitem__(self, all_cuts: CutSet) -> dict[str, torch.Tensor | list[str] |
             )
             ans.update(text_minibatch)
 
+        def post_process_seq(ans):
+            max_seq_length = self.max_seq_length - 1  # reserve for eos
+            if len(ans.input_ids) > max_seq_length:
```
pzelasko, line 137: I believe this truncation is a very, very bad practice and can possibly break your model without you knowing it. Consider the following scenario:

* 50% of training data exceeds the max_seq_length threshold
* you truncate the answers in 50% of cases
* after every such case, you add an <eos> prediction
* therefore, you are training the model to output incomplete answers and stop the prediction after random tokens.

In my opinion this should only be enabled if you set an option `allow_unsafe_answer_truncate` that is always `False` by default. In EMMeTT branch I have a filter for `max_tokens` that allows you to filter out these examples instead.

nemo/collections/multimodal/speech_llm/models/modular_models.py
```diff
@@ -345,6 +347,146 @@ def _get_text_embeddings(self, text_tokens, position_ids):
...
+                input_length=input_length,
+                loss_mask=loss_mask,
+                audio_locator_ids=audio_locator_ids,
+                remove_bos_or_eos=True,
+            )
+        )
+
+        return encoder_input, attention_mask, labels, loss_mask, encoder_length
+
+    def inject_perception_input_conv(
```
pzelasko, line 382: I didn't review this part in-depth; however, I think it would be best if you factored it out as a separate utility function and added unit test coverage for it for no-audio, single-audio, and multi-audio cases.

nemo/collections/multimodal/speech_llm/data/lhotse_dataset.py
```diff
@@ -124,9 +132,96 @@ def __getitem__(self, all_cuts: CutSet) -> dict[str, torch.Tensor | list[str] |
...
+                    )
+                    ans.answer_ids = ans.answer_ids[: -min(truncation_length, len(ans.answer_ids))]
+                    ans.context_ids = ans.context_ids[: -min(truncation_length, len(ans.context_ids))]
+            # add eos
+            eos_id = self.text_processor.tokenizer.eos_id
+            ans.input_ids = torch.cat([ans.input_ids, torch.tensor([eos_id])])
+            ans.answer_ids = torch.cat([ans.answer_ids, torch.tensor([eos_id])])
+            ans.mask = torch.cat([ans.mask, torch.tensor([1])])
+
+        multimodal_convo_examples = all_cuts.filter(lambda c: isinstance(c, NeMoMultimodalConversation))
```
pzelasko, line 155: Would be good to add a unit test that creates a single `NeMoMultimodalConversation` (you can re-use the fixtures from my unit tests), runs it through this dataset, and checks that we got the expected output.

pzelasko (commented):
Left some comments. Generally looks good, but we need some unit test coverage

PR #10794, opened 2024-10-08
Fix ASR tests

artbataev:
# What does this PR do ?

Fix ASR tests after upgrading to `nvidia/pytorch:24.07` container:
- merge RNN-T alignment tests with unit tests
- avoid files in RNN-T alignment tests that cause issues with the precision, which leads to different results for batched/non-batched decoding
- support non-CUDA inputs for CUDA graph decoder with frame-looping algorithm (fallback to default algorithm in this case)
- temporary fix for failing tests with broken `torch.load` (should be debugged and fixed in the follow-up PR)

**Collection**: [ASR]

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


.github/workflows/cicd-main.yml
```diff
@@ -131,16 +131,16 @@ jobs:
         ### \'\'
   
   # L0: GPU unit tests
-  OPTIONAL_L0_Unit_Tests_GPU_ASR:
+  L0_Unit_Tests_GPU_ASR:
      needs: [cicd-test-container-setup]
```
ko3n1g, line 135: Please make sure to also re-add them to the final job `Nemo_CICD_Test` (into the `needs` list)

ko3n1g (changes requested):

.github/workflows/cicd-main.yml
artbataev: Ok, thanks, I will do this! 
I'm currently trying to reproduce the issue on CI.
For RNN-T alignments, I do not see failures when running locally.
For some other tests, there are strange issues with serialization. If reproducible, I will try to fix it.

ko3n1g: Maybe there’s a cuda driver mismatch? If you cannot reproduce it locally?

artbataev: The problem is with the precision of calculations. I reduced the number of files. See details in Slack.

nemo/collections/asr/parts/submodules/cuda_graph_rnnt_greedy_decoding.py
```diff
@@ -293,6 +293,13 @@ def __call__(
         device: torch.device,
         partial_hypotheses: Optional[List[rnnt_utils.Hypothesis]] = None,
     ):
+        if x.device.type != "cuda":
```
artbataev, line 296: @galv I added this to automatically support non-cuda inputs when the CUDA graph decoder + frame-looping algorithm is enabled. If you think there is a better way, please let me know.

artbataev: Also, it is possible to avoid some tests for frame-looping algorithm instead of this solution.

galv: To be honest, if I were doing this, I would probably just throw an exception (what torch does if you pass a tensor on the wrong device to an op that runs on a different device). Well, I suppose the code already throws an exception and fails :P 

I'm just not a fan of "falling back" to something else when we cannot do what the user requests, but I know that decision has already been made a few places regarding the cuda graphs decoding work. So I think this is fine! Let me know if you need me to approve. Otherwise, I will let someone else more familiar with the rest approve.

ko3n1g (commented):
QQ: Why did you remove `L2_Transducer_alignment_Running_pytest` completely?

artbataev:
> QQ: Why did you remove L2_Transducer_alignment_Running_pytest completely?

@ko3n1g Initially, this stage was separated because it was tied to the Jenkins machine.
I fixed this in the previous PR https://github.com/NVIDIA/NeMo/pull/9862, and now it is just a unit test. This saves computing time since checkpoints used in different tests are cached. 

The tests passed now. You can see in the log https://github.com/NVIDIA/NeMo/actions/runs/11371567918/job/31634824387?pr=10794

```
tests/collections/asr/decoding/test_rnnt_alignments.py::test_rnnt_alignments[True-True-cuda] PASSED [ 28%]
tests/collections/asr/decoding/test_rnnt_alignments.py::test_rnnt_alignments[True-False-cuda] SKIPPED [ 28%]
tests/collections/asr/decoding/test_rnnt_alignments.py::test_rnnt_alignments[False-True-cuda] PASSED [ 28%]
tests/collections/asr/decoding/test_rnnt_alignments.py::test_rnnt_alignments[False-False-cuda] PASSED [ 28%]
```

ko3n1g (approved):
Note that my review is only from perspective of `cicd-main.yml`. For everthing else, I'm not the right guy please collect feedback from others as you prefer. Theoretically we're good to merge if you're comfortable with it

pablo-garay (approved):

PR #10808, opened 2024-10-08
Migrate SpeechLM to NeMo 2.0

stevehuang52:
# What does this PR do ?

Migrate SpeechLM to NeMo 2.0

**Collection**: [asr,llm,speechlm,multimodal]

# Changelog 

### Major design in `nemo.collections.speechlm`: 

@titu1994 @pzelasko @yaoyu-33 

#### Data
- `nemo/collections/speechlm/data/audio_to_text_module.py`: integrates all previous speechlm datasets
- `nemo/collections/speechlm/data/data_sampler.py`: customized sampler that skip adding for iterable datasets
- `nemo/collections/speechlm/data/dataset/audio_text_dataset.py`: refactored speechlm dataset to use prompt formatter from lhotse
- `nemo/collections/speechlm/data/text_processing.py`: unified text processor for both lhotse and non-lhotse datasets

#### Model
- `nemo/collections/speechlm/models/base.py`: a basic PTL class that maintain some key functions from `ModelPT` in nemo 1.x
- `nemo/collections/speechlm/models/speech_to_text_llm_model.py` the actual model class implementation

#### Module
- `nemo/collections/speechlm/modules/*.py`: wrapper classes for ASR encoder and modality adapter, such that all ASR modules can be used.

#### Entry point
- `nemo/collections/speechlm/recipes/pipeline.py`: main entrypoint for training/validation, support hydra and yaml config as in nemo 1.x

#### Utils
- `nemo/collections/speechlm/utils/hydra_utils.py`: util funcs for hydra
- `nemo/collections/speechlm/utils/model_transform.py`: customized PEFT class for multi-modal LLM (FYI @yaoyu-33)
- `nemo/collections/speechlm/utils/resume.py`: customized resume class for multi-modal LLM such that RestoreConfig is no longer required.

#### Text generation
- `nemo/collections/speechlm/utils/text_generation/*.py`: adapt from `nemo.collections.multimodal.speech_llm` such that it works with nemo 2.0

#### Megatron strategy
`nemo/collections/speechlm/strategies/megatron_strategy.py`: need to overwrite a function to make resuming PEFT jobs work with SpeechLM.

#### Examples
- `examples/speechlm/speech_to_text_llm_train.py` : main python entrypoint for training
- `examples/speechlm/conf/salm/salm_llama3-8b_fc_fc_peft.yaml`: example configs for SALM and similar models

### Changes in `nemo.lightning`: 

@hemildesai @marcromeyn 

- `nemo/lightning/io/mixin.py`: add class method to support importing checkpoint without instantiating an object from the class.
- `nemo/lightning/megatron_parallel.py`: add back dataloader index for metric calculation on multiple validation datasets.
- `nemo/lightning/pytorch/callbacks/progress_printer.py`: hard-code total length of the progress bar for iterable dataset that don't have finite length (e.g., lhotse, FYI @pzelasko )
- ~`nemo/lightning/pytorch/strategies/megatron_strategy.py`:  allow empty state dict such that multi-modal PEFT can work properly.  (FYI @yaoyu-33 )~ Moved change to a new class in `nemo/collections/speechlm/strategies/megatron_strategy.py`
- `nemo/utils/callbacks/dist_ckpt_io.py`: add class info in error message.

### Changes in `nemo.collections` not related to `speechlm`

`nemo.collections.asr`: @titu1994 
- add `add_blank` param to conv asr decoder so that we can skip adding blank class.

`nemo.collections.common`: @pzelasko 
- `data`: minor change, replace raising error with logging.warning when mismatched items found.
- `prompts` : add llama3 prompt formatter
- `tokenizers`: add `TokenizerWrapper` to `__all__`

### CI tests
- `tests/collections/speechlm/speech_to_text_llm_train.py`
- `.github/workflows/cicd-main.yml`

### Minor changes that don't need much attention
`nemo.collections.multi_modal.speech_llm`:
- minor improvements, should be fine

`nemo/collections/nlp/modules/common/megatron/utils.py`
- add more detailed logging info when batch size not divisible


nemo/lightning/megatron_parallel.py
```diff
@@ -1327,14 +1327,21 @@ def get_data_iterator_and_seq_length(self) -> Tuple[List[Iterator[DataT]], Optio
...
+
             # finetuning can have dynamic sequence lengths
             seq_length = batch['tokens'].size(1) if 'tokens' in batch else None
             from nemo.collections.nlp.modules.common.megatron.utils import get_iterator_k_split
 
             data = get_iterator_k_split(batch, self.num_microbatches, True)
+
+            if batch_idx is not None and dataloader_idx is not None:
+                packed_data = [(d, batch_idx, dataloader_idx) for d in data]
+                return [itertools.chain(packed_data)], seq_length
```
yaoyu-33, line 1344: this seems to change the behavior of all models?
previously it it will just return the batch, now it's a packed thing

nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -840,7 +840,14 @@ def load_model_state_dict(self, checkpoint: Mapping[str, Any], strict: bool = Tr
         """loads model state dict"""
         assert self.megatron_parallel is not None
 
-        strict = strict if self.ckpt_load_strictness is None else self.ckpt_load_strictness
+        if "state_dict" not in checkpoint or len(checkpoint["state_dict"]) == 0:
```
yaoyu-33, line 843: what case you are loading a checkpoint without "state_dict"?

nemo/lightning/megatron_parallel.py
stevehuang52: I've tested it with a LLM SFT script and it works. The returned types remain the same as before, which are `List[Iterator]` and `int`

nemo/lightning/pytorch/strategies/megatron_strategy.py
stevehuang52: In https://github.com/NVIDIA/NeMo/blob/heh/speechlm_nemo2.0/nemo/collections/speechlm/utils/model_transform.py#L224

The first pass of PEFT resume needs a base model, but for SpeechLM we have two base models (ASR and LLM), thus the current PEFT resume method won't work for SpeechLM. I use an empty state_dict to avoid the error during resuming from PEFT jobs.

nemo/collections/speechlm/models/base.py
```diff
@@ -0,0 +1,507 @@
...
+
+__all__ = ['SpeechLanguageModel']
+
+
+class SpeechLanguageModel(pl.LightningModule, io.IOMixin, io.ConnectorMixin, fn.FNMixin):
+    def __init__(self):
+        super().__init__()
+        app_state = AppState()
+
+        self._set_model_guid()
```
titu1994, line 39: Do we need the guid infrastructure for some reason?

nemo/collections/speechlm/models/base.py
```diff
@@ -0,0 +1,507 @@
...
+from nemo.collections.llm import fn
+from nemo.lightning import io
+from nemo.lightning.megatron_parallel import MaskedTokenLossReductionWithLossMask
+from nemo.utils import logging
+from nemo.utils.app_state import AppState
+
+__all__ = ['SpeechLanguageModel']
+
+
+class SpeechLanguageModel(pl.LightningModule, io.IOMixin, io.ConnectorMixin, fn.FNMixin):
```
titu1994, line 34: Awesome work on this, looks like a modernized version of ASRModel

nemo/collections/speechlm/models/base.py
```diff
@@ -0,0 +1,507 @@
...
+            for _ in range(self._num_test_dl):
+                self._test_step_outputs.append([])
+
+        return self._test_step_outputs
+
+    @test_step_outputs.setter
+    def test_step_outputs(self, value):
+        self._test_step_outputs = value
+
+    def on_validation_epoch_end(self) -> Optional[Dict[str, Dict[str, torch.Tensor]]]:
```
titu1994, line 235: This.. can be refactored out of ModelPT as a mixin and shared between ModelPT and SpeechLanguageModel probably. But for now this is fine

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+
+
+def set_input_tensor(self, tensor: torch.Tensor):
+    """
+    Placeholder function for pipeline parallel, not implemented yet.
+    """
+    pass
+
+
+def speech_to_text_llm_data_step(dataloader_iter) -> Dict[str, Any]:
```
titu1994, line 78: Can you document where this is used and in what context?

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+
+    # slice batch along sequence dimension for context parallelism
+    output = get_batch_on_this_context_parallel_rank(_batch)
+
+    output["dataloader_idx"] = dataloader_idx
+    output["batch_idx"] = batch_idx
+    return output
+
+
+def speech_to_text_llm_forward_step(model: L.LightningModule, batch: Dict[str, Any]) -> torch.Tensor:
```
titu1994, line 134: I know LLM does it this style, but if you feel its better to fuse this with MCoreSpeechToTextLLM, feel free to do so.

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+    forward_step_fn: Callable = speech_to_text_llm_forward_step
+    data_step_fn: Callable = speech_to_text_llm_data_step
+
+    text_generation_strategy: SpeechToTextGenerationStrategy = SpeechToTextGenerationStrategy
+
+    inference_config: Optional[Dict[str, Any]] = None
+
+    data_config: Optional[DictConfig] = None
+
+    def freeze_module(self, module: nn.Module):
```
titu1994, line 184: duplicate since SpeechLanguageModel has this?

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+
+        model = MCoreSpeechToTextLLM(
+            config=self,
+            language_model=language_model,
+            speech_model=speech_model,
+            modality_adapter=modality_adapter,
+            tokenizer=tokenizer,
+        )
+
+        if self.freeze_language_model:
```
titu1994, line 237: Can be replaced by making the SpeechLanguageModel.freeze into a static method - its not really dependent on anything

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+                emb1.shape[0] + emb2.shape[0], emb1.shape[-1], device=emb1.device, dtype=emb1.dtype
+            )
+            padded_new_emb[:new_len, ...] = new_emb
+            concat_emb.append(padded_new_emb)
+            concat_len.append(new_len)
+        concat_emb = torch.stack(concat_emb, dim=0)
+        concat_len = torch.stack(concat_len, dim=0)
+        return concat_emb, concat_len
+
+    def _concat_multi_features(
```
titu1994, line 305: Doc on what the multi part is

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+        ):
+            position_embeddings = lm_embedding.position_embeddings(position_ids)
+            combined_embed = combined_embed + position_embeddings
+
+        encoder_max_length = combined_embed.shape[1]
+        # Data format change to avoid explicit tranposes : [b s h] --> [s b h].
+        combined_embed = combined_embed.transpose(0, 1).contiguous()
+
+        # If the input flag for fp32 residual connection is set, convert for float.
+        if lm_embedding.config.fp32_residual_connection:
```
titu1994, line 379: ... how did you even find out that you had to do this step i wonder

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+
+    def training_step(self, batch, batch_idx=None) -> torch.Tensor:
+        # In mcore the loss-function is part of the forward-pass (when labels are provided)
+        return self.forward_step(batch)
+
+    def validation_step(self, batch, batch_idx=None) -> torch.Tensor:
+        return self.inference_step(batch, mode='validation')
+
+    @property
+    def _metrics_require_string2category_map(self):
```
titu1994, line 642: f1 and accuracy for speech metrics?

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+
+            metric_name = data_cfg.metric.name
+            metric_cls = MetricStringToTorchMetric[metric_name]
+            if metric_name not in TextMetricsSet:
+                metric = [metric_cls(**data_cfg.metric)]
+            else:
+                metric = [metric_cls()]
+        return metric, metric_name
+
+    def inference_step(self, batch, mode):
```
titu1994, line 692: Thats a LOT of code just to calculate metrics - what is it necessary for? Do we track this during training for checkpointing?

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+                inference_config['inputs'] = (
+                    batch['contexts'].cuda(),
+                    batch['context_lengths'].cuda(),
+                    batch['audio_signal'].cuda(),
+                    batch['audio_signal_length'].cuda(),
+                )
+            response = generate(self, **inference_config)
+
+        app_state = AppState()
+        reconfigure_num_microbatches_calculator(
```
titu1994, line 828: Why do we need to reconfigure micro batch size on the fly ? Document in one-two sentences

nemo/collections/speechlm/models/speech_to_text_llm_model.py
```diff
@@ -0,0 +1,1127 @@
...
+            reconfigure_num_microbatches_calculator(
+                rank=app_state.global_rank,
+                rampup_batch_size=None,
+                global_batch_size=data_cfg.global_batch_size,
+                micro_batch_size=data_cfg.micro_batch_size,
+                data_parallel_size=parallel_state.get_data_parallel_world_size(),
+            )
+        return averaged_loss, averaged_metric
+
+    def gather_and_maybe_write_predictions(self, data_cfg, output, averaged_metric, mode, dataloader_idx):
```
titu1994, line 943: ?? why are we writing files under distributed programming setup, that to across multi nodes?

nemo/collections/speechlm/recipes/pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+from nemo.collections.speechlm.modules.asr_module import ASRModuleConfig
+from nemo.collections.speechlm.modules.modality_adapter import ModalityAdapterConfig
+from nemo.collections.speechlm.utils import SpeechToTextLLMPEFT, get_object_list_from_config
+from nemo.core.classes.common import Serialization, typecheck
+from nemo.lightning.pytorch.callbacks import PreemptionCallback
+from nemo.utils import logging
+from nemo.utils.exp_manager import StatelessTimer, TimingCallback
+
+
+def speech_to_text_llm_train(cfg: DictConfig):
```
titu1994, line 32: Pretty cool ! If i read this correctly, we can reuse this one function for most SpeechToTextLLM while swapping out the underlying model with hydra target?

nemo/collections/speechlm/recipes/pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+    # 3. setup the optimizer
+    optim = Serialization.from_config_dict(cfg['optim'])
+
+    # 4. setup trainer
+    callbacks = get_object_list_from_config(cfg['callbacks'])
+    if cfg.get('max_time_per_run', None):
+        if cfg['strategy'].get('ckpt_async_save', True):
+            logging.warning(
+                f"`strategy.ckpt_async_save` must be `False` to save ckpt when `max_time_per_run` is set,",
+                f"got {cfg['strategy']['ckpt_async_save']}. `max_time_per_run` will not work in this case!",
```
titu1994, line 67: Better to error out here. I dont want a run to go on assuming it will stop at max time, only to get a corrupted checkpoint after 4 hours

nemo/collections/speechlm/utils/text_generation/audio_text_generation_strategy.py
```diff
@@ -0,0 +1,427 @@
...
+                    f"The end string '{end_string}' has no associated special token: this may slow down "
+                    "generation (consider using a different tokenizer or modifying `end_strings`)"
+                )
+                self._end_of_generation_cache["end_strings_to_check"].add(end_string)
+                end_strings_to_check.append(end_string)
+
+        return end_tokens, end_strings_to_check
+
+
+class SpeechToTextGenerationStrategy(TextGenerationStrategy):
```
titu1994, line 272: Is this tested? Or placeholder till llm.generate API works?

titu1994:
Incredible work in making this work !

nemo/collections/speechlm/data/audio_to_text_module.py
```diff
@@ -0,0 +1,466 @@
...
+            augmentor = None
+
+        # Notably, the data weights are controlled by either bucketing_weights
+        # or concat_sampling_probabilities depending on the dataset type.
+        if data_cfg.get("use_lhotse"):
+            logging.info(f"Creating Lhotse dataset for {mode}")
+            if mode != 'train':
+                setattr(self, f"_{mode}_names", self._parse_lhotse_data_name(mode))
+
+            return LhotseAudioQuestionAnswerDataset(
```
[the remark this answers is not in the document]
stevehuang52: It is better for users to overwrite the whole `_create_dataset` for new datasets, since they will have very different `__init__`. Also, the current `LhotseAudioQuestionAnswerDataset` dataset class is flexible enough for all speech-to-text tasks we have explored so far, and it also supports pure text training.

nemo/collections/speechlm/data/audio_to_text_module.py
```diff
@@ -0,0 +1,466 @@
...
+        elif hasattr(dataset, 'collate_fn'):
+            collate_fn = dataset.collate_fn
+        elif hasattr(dataset.datasets[0], 'collate_fn'):
+            # support datasets that are lists of entries like ChainDataset
+            collate_fn = dataset.datasets[0].collate_fn
+        else:
+            # support datasets that are lists of lists
+            collate_fn = dataset.datasets[0].datasets[0].collate_fn
+
+        if isinstance(dataset, torch.utils.data.IterableDataset):
```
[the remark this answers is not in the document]
stevehuang52: done

nemo/collections/speechlm/data/audio_to_text_module.py
```diff
@@ -0,0 +1,466 @@
...
+        Args:
+            state_dict: the datamodule state returned by ``state_dict``.
+
+        """
+        try:
+            from megatron.core.num_microbatches_calculator import update_num_microbatches
+
+        except (ImportError, ModuleNotFoundError):
+            logging.warning("Megatron num_microbatches_calculator not found, using Apex version.")
+            from apex.transformer.pipeline_parallel.utils import update_num_microbatches
```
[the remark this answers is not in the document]
stevehuang52: Dropped. I was following the fine_tuning.py data modulde

nemo/collections/speechlm/data/dataset/audio_text_dataset.py
```diff
@@ -0,0 +1,1039 @@
...
+        self.trim = trim
+        self.channel_selector = channel_selector
+
+    def get_manifest_sample(self, sample_id):
+        """
+        return the manifest item of the given index
+        """
+        return self.collection[sample_id]
+
+    def __getitem__(self, index):
```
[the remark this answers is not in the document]
stevehuang52: I asssume you mean the one in `nemo.collections.multimodal.speech_llm`, this one is used to share the same text processor with the lhotse dataset. As for why we keep the nemo dataset while having the lhotse one, my answer is we always need something to fall back to and have full control on, same reason as we want nemo lite since we don't want to go inside megatron-core to modify model arch.

nemo/collections/speechlm/data/dataset/audio_text_dataset.py
```diff
@@ -0,0 +1,1039 @@
...
+            [datasets.append(dataset) for _ in range(bucketing_weights[dataset_idx])]
+        else:
+            datasets.append(dataset)
+
+    with open_dict(config):  # patch for bucketing tarred datasets
+        config['batch_size'] = config.get("micro_batch_size", 1)
+    return get_chain_dataset(datasets=datasets, ds_config=config, rank=global_rank)
+
+
+def get_concat_tarred_audio_text_dataset(
```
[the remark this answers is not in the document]
stevehuang52: Haha....I was just following the ASR collection. In the end only
```python
__all__ = [
    'AudioTextDataset',
    'TarredAudioTextDataset',
    'get_tarred_audio_text_dataset_from_config',
    'get_audio_text_dataset_from_config',
]
```
are exposed to other scripts

nemo/collections/speechlm/models/base.py
stevehuang52: yeah I agree this can be a mixin, but leave it as it is for now to avoid more changes to ModelPT

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: added doc. in nemo2 each model config needs to have its own data step func and forward func...

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: You're right, thanks for spotting this~

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: there's a freeze() function in nemo/collections/llm/fn/mixin.py:FNMixin (and SpeechLanguageModel has FNMixin), I try not to overwrite it...

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: added

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: by looking into the implementation in nlp and mcore....I did spend some time to figure it out

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: those were kept following the GPT SFT model in nlp collection, can be used for multiple choice QA

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: if the logging metric is val_loss, the text decoding and metric calculation parts are skipped to save time. I'm going to simplify it to use the inference pipeline in mcore but it will take quite some time to do. For now the current method works fine.

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: I think it's only needed when the last batch has different batch size from the previous batches. This line was copied from nemo1.x implementation of speechlm.

nemo/collections/speechlm/models/speech_to_text_llm_model.py
stevehuang52: writing only happens on global rank 0, others are just aggregating the results.

nemo/collections/speechlm/recipes/pipeline.py
stevehuang52: updated to raise error

nemo/collections/speechlm/utils/text_generation/audio_text_generation_strategy.py
stevehuang52: It's tested to work correctly. Will update to llm.generate API in future PRs.

ko3n1g (approved):

.github/workflows/cicd-main.yml
```diff
@@ -5023,6 +5023,40 @@ jobs:
         rm -rf /tmp/nemo2_ckpt
         rm -rf /tmp/vllm_from_nemo2
 
+  L2_SpeechLM_LoRA_TP1PP1_MBS2:
+    needs: [cicd-test-container-setup]
```
ko3n1g, line 5027:
```suggestion
    needs: [pre-flight, cicd-test-container-build]
```

nemo/collections/common/data/lhotse/nemo_adapters.py
```diff
@@ -429,7 +435,10 @@ def basename(d: dict) -> str:
                     del raw_audio
                     yield from cuts_for_recording
             except tarfile.ReadError:
-                logging.warning(f"Skipping tar file due to read errors (unstable storage or bad file?): {tar_path=}")
+                logging.warning(
+                    f"Skipping tar file due to read errors (unstable storage or bad file?): {tar_path=}",
+                    mode=logging_mode.ONCE,
```
pzelasko, line 440: Please revert, if we have a lot of issues with data access, it should be clear from the logs.

nemo/collections/common/data/lhotse/nemo_adapters.py
```diff
@@ -360,6 +361,11 @@ def _iter_sequential(self, tar_path, shard_manifest, manifest_path) -> Generator
                     yield data, raw_audio, tar_info
                 except KeyError as e:
                     if self.skip_missing_manifest_entries:
+                        logging.warning(
```
pzelasko, line 364: Please revert, the user set a flag to silently omit missing manifest entries, we shouldn't spam them with warnings.

nemo/collections/speechlm/data/text_processing.py
```diff
@@ -0,0 +1,110 @@
...
+@dataclass
+class MockCutSupervision:
+    """
+    A dummy class for MockCut to support behavior of Cut.supervisions[0].text
+    """
+
+    text: str = ""
+
+
+class MockCut(Cut):
```
pzelasko, line 35: What is this? It will only work with a small subset of PromptFormatters for which `@registered_prompt_format_fn` reads from `context` / `question` / `answer`.

pzelasko: This shouldn't be needed, please add a data type following `nemo/collections/common/data/lhotse/text_adapters.py` and register a prompt format fn for it. You can follow the tutorial at https://github.com/NVIDIA/NeMo/blob/main/tutorials/multimodal/Prompt%20Formatter%20Tutorial.ipynb

nemo/collections/speechlm/data/dataset/audio_text_dataset.py
```diff
@@ -0,0 +1,1039 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
pzelasko, line 1: IMO we should only support Lhotse based dataset for this new collection to reduce the amount of complexity and maintenance. Supporting two completely different dataloader implementations will split our resources and waste our energy going forward. I suggest that this whole file and references to it throughout the collection should be removed.

nemo/collections/speechlm/data/data_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    def transform_dataloader(self, dataloader: DataLoader, consumed_samples: int = 0) -> DataLoader:
+        """
+        Overwrites the MegatronDataSampler.transform_dataloader() function
+        """
+        sampler = getattr(dataloader, 'sampler', None)
+
+        if sampler is not None and "lhotse." in str(type(sampler)):
+            logging.info(f"Using Lhotse sampler for dataloader {dataloader}. Skipping Megatron data sampler.")
```
pzelasko, line 32: `MegatronDataSampler` seems to be implementing the logic for splitting batches into micro batches based on some predefined settings. If we completely skip it, and don't add our own logic for the splitting, does that mean lhotse based dataloading doesn't work with any model parallelism at all? Was that tested?

Again if we maintain two dataloader implementations, we will have to test both of them against the full feature set we want to support. AFAIK at the very least TP should be supported for SpeechLM in NeMo 2.

nemo/collections/speechlm/data/audio_to_text_module.py
```diff
@@ -0,0 +1,463 @@
...
+        if data_cfg.get('prompt_format', None):
+            text_processor = TextProcesserWithPromptFormatter(
+                self.tokenizer,
+                prompt_format=data_cfg['prompt_format'],
+                max_seq_length=data_cfg["max_seq_length"],
+                audio_locator=data_cfg.get('audio_locator', None),
+            )
+            return text_processor
+
+        text_processor = TextProcessing(
```
pzelasko, line 111: This `TextProcessing` class is an orthogonal API to `PromptFormatter` with largely overlapping functionality (see arguments such as `prompt_template`, `add_bos/eos/sep`, `separate_prompt_and_response_with_newline`, `context/answer_key`, etc).

Why do we need two APIs for this in a new, clean version of SpeechLM? Let's just choose one. `PromptFormatter` supports EMMETT-style bucketing and OOMptimizer for both text and audio data, so I don't see the reason to keep supporting the older `TextProcessing` API especially if it's applied on top of `PromptFormatter`...

nemo/collections/speechlm/modules/asr_module.py
```diff
@@ -0,0 +1,135 @@
...
+
+import nemo.collections.asr as nemo_asr
+from nemo.collections.speechlm.utils import to_dict_config
+from nemo.core.classes.common import Serialization
+from nemo.core.classes.module import NeuralModule
+from nemo.lightning import io
+from nemo.utils import logging, model_utils
+
+
+class MCoreASRModule(MegatronModule):
```
pzelasko, line 33: Can you add some documentation what is the purpose of this class?

nemo/collections/speechlm/modules/asr_module.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(
+        self,
+        encoder: NeuralModule,
+        preprocessor: Optional[nn.Module] = None,
+        spec_augment: Optional[nn.Module] = None,
+    ):
+        super().__init__(config=TransformerConfig(num_layers=1, hidden_size=1, num_attention_heads=16))
```
pzelasko, line 40: What is this `TransformerConfig` with hard-coded options? Can you add a comment explaining this?

nemo/collections/speechlm/modules/modality_adapter.py
```diff
@@ -0,0 +1,64 @@
...
+from megatron.core.transformer.transformer_config import TransformerConfig
+
+from nemo.collections.asr.modules import ConformerEncoder
+from nemo.collections.speechlm.utils import to_dict_config
+from nemo.core.classes.common import Serialization
+from nemo.core.classes.module import NeuralModule
+from nemo.lightning import io
+
+
+class MCoreModalityAdapterModule(MegatronModule):
```
pzelasko, line 30: Can you add a docstring describing what this is?

pzelasko (commented):
Leaving an initial review. Lots of great work navigating NeMo 2.0 APIs (which I'm still trying to wrap my head around)! I have a few concerns about supporting multiple dataloader backends and multiple prompt formatting implementations that are at odds with each other. I'm afraid this will lead to fragmentation within the team which backends are used by whom and we won't be able to maintain feature parity. I left some comments indicating which parts/approaches would be best removed IMO.

nemo/collections/speechlm/data/audio_to_text_module.py
titu1994: TextProcessing is the manifest processor logic no ? That's independent of Prompt formatting?

nemo/collections/speechlm/data/data_sampler.py
```diff
@@ -0,0 +1,49 @@
...
+    Overwrite the default MegatronDataSampler to not add batch sampler when iteralbe dataset is used.
+    """
+
+    def transform_dataloader(self, dataloader: DataLoader, consumed_samples: int = 0) -> DataLoader:
+        """
+        Overwrites the MegatronDataSampler.transform_dataloader() function
+        """
+        sampler = getattr(dataloader, 'sampler', None)
+
+        if sampler is not None and "lhotse." in str(type(sampler)):
```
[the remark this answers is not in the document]
stevehuang52: Done~

nemo/collections/speechlm/data/data_sampler.py
stevehuang52: TP support is verified with both lhotse as well, we did the same thing in NeMo 1.x by not using Megatron sampler, and it worked.

nemo/collections/speechlm/modules/asr_module.py
stevehuang52: done

nemo/collections/speechlm/modules/asr_module.py
stevehuang52: done

nemo/collections/speechlm/modules/modality_adapter.py
stevehuang52: done

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -585,6 +588,8 @@ def read_nemo_manifest(config) -> tuple[CutSet, bool]:
                 cutsets.append(CutSet(nemo_iter))
                 weights.append(weight)
         # Finally, we multiplex the dataset streams to mix the data.
+        if len(cutsets) == 1:
+            return cutsets[0], is_tarred
```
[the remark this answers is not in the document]
stevehuang52: moved it to `mux`. the change is mainly to avoid the error when single manifest is used, since `CutSet.mux` has to get at least two cutsets.

nemo/collections/speechlm/data/text_processing.py
stevehuang52: This is necessary for using the prompt formatter in both lhotse and non-lhotse datasets. It's not creating a new prompt format or adapter.

nemo/collections/common/data/lhotse/nemo_adapters.py
stevehuang52: done

nemo/collections/speechlm/data/dataset/audio_text_dataset.py
stevehuang52: I think we should keep the non-lhotse dataset as well. It only needs to have the basic functions so that people not familiar with lhotse can still know what's happning in the dataset. For more advanced features we will just ask users to use lhotse dataloader.

nemo/lightning/io/mixin.py
```diff
@@ -415,6 +415,39 @@ def import_ckpt(self, path: str, overwrite: bool = False, base_path: Optional[Pa
         connector.on_import_ckpt(self)
         return ckpt_path
 
+    @classmethod
+    def import_ckpt(cls, path: str, overwrite: bool = False, base_path: Optional[Path] = None, **kwargs) -> Path:
```
yaoyu-33, line 419: how is it different from `https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/io/api.py#L111`

what's the use case of this?

stevehuang52: Ah, thanks for the pointer~I'll remove the classmethod

nemo/collections/speechlm/utils/text_generation/audio_text_generation_strategy.py
```diff
@@ -0,0 +1,427 @@
...
+
+
+def switch(val1, val2, boolean):
+    """switch between two values based on the boolean tensor"""
+    boolean = boolean.type_as(val1)
+    boolean = boolean.unsqueeze(0).unsqueeze(-1)
+    return (1 - boolean) * val1 + boolean * val2
+
+
+class TextGenerationStrategy:
```
zhehuaichen, line 49: can we inherit this from text-only GPT or do we have to create our own from scratch?

tests/collections/speechlm/conf/salm_ci_config.yaml
```diff
@@ -0,0 +1,208 @@
+name: megatron_audio_gpt_peft
```
zhehuaichen, line 1: can u remind me which .py decides the ci test process with this .yaml file?

nemo/collections/multimodal/speech_llm/data/audio_text_dataset.py
```diff
@@ -282,6 +283,9 @@ def __init__(
         context_file: Optional[Union[List[str], str]] = None,
         sample_alpha: Optional[float] = None,
         audio_locator: Optional[str] = None,
+        add_boa_eoa: Optional[bool] = False,
+        boa_string: Optional[str] = "<BOA>",
```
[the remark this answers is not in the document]
stevehuang52: want to see if they can help model locate the audios

nemo/collections/speechlm/data/audio_to_text_module.py
```diff
@@ -0,0 +1,463 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
stevehuang52: for now, yes

nemo/collections/speechlm/recipes/pipeline.py
```diff
@@ -0,0 +1,191 @@
...
+from nemo.collections.speechlm.modules.modality_adapter import ModalityAdapterConfig
+from nemo.collections.speechlm.utils import SpeechToTextLLMPEFT, get_object_list_from_config
+from nemo.core.classes.common import Serialization, typecheck
+from nemo.lightning.pytorch.callbacks import PreemptionCallback
+from nemo.utils import logging
+from nemo.utils.exp_manager import StatelessTimer, TimingCallback
+
+
+def speech_to_text_llm_train(cfg: DictConfig, tokenizer: Optional[AutoTokenizer] = None):
+    """Train the model using provided config."""
```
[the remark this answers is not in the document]
stevehuang52: can do later, but not in this PR

nemo/collections/speechlm/utils/model_transform.py
```diff
@@ -0,0 +1,226 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
stevehuang52: please refer to nemo2.0 docs for more details

nemo/collections/speechlm/utils/resume.py
```diff
@@ -0,0 +1,58 @@
...
+from nemo.lightning import io
+from nemo.lightning.pytorch.callbacks.peft import ADAPTER_META_FILENAME
+from nemo.lightning.resume import AdapterPath, AutoResume
+from nemo.utils.app_state import AppState
+
+
+@dataclass(kw_only=True)
+class SpeechLMAutoResume(AutoResume):
+    """
+    Wrapper for AutoResume to get rid of requirement on restore_config.
```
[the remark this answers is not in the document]
stevehuang52: it's difficult to explain in a PR comment, please refer to nemo2.0 docs for more details.

nemo/collections/speechlm/utils/text_generation/audio_text_generation_strategy.py
stevehuang52: this is just for now, later we should use the text generation pipeline in mcore and these will be removed. we can borrow some code from the pipeline in mcore, but still need to overwrite a few things

tests/collections/speechlm/conf/salm_ci_config.yaml
stevehuang52: it's under `tests/collections/speechlm/` as well

nemo/lightning/megatron_parallel.py
yaoyu-33: refactor a bit not do a early return

nemo/lightning/pytorch/strategies/megatron_strategy.py
stevehuang52: @cuichenx Could you please take a look and see if the change is fine or there is a better solution to it?

nemo/lightning/megatron_parallel.py
stevehuang52: done~

nemo/collections/speechlm/recipes/pipeline.py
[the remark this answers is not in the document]
stevehuang52: Just cleaned up as you suggested.

nemo/lightning/pytorch/strategies/megatron_strategy.py
cuichenx: hmm I think this change feels dangerous.. is it possible to create a custom checkpoint io for speechlm to handle this?

stevehuang52: I tried but failed, since PTL will always try to load the checkpoint before adding PEFT parameters to the network. I'll be better if PEFT transform can happen before PTL resumes from previous jobs.

cuichenx: how do you load two checkpoints with PTL without peft?

stevehuang52: @cuichenx  Is there a way to return current parameters in the network as they are? For example, in this line https://github.com/NVIDIA/NeMo/blob/heh/speechlm_nemo2.0/nemo/collections/speechlm/utils/model_transform.py#L219 , if I can do `return  {'state_dict': sharded_state_dict['state_dict']}` it will not need the change in megatron_strategy.py. However if I do that I'll get some errors saying that "unexpected checkpoint format".

stevehuang52: > how do you load two checkpoints with PTL without peft?

I load the two base models in the SpeechLM model's `configure_model()` function

cuichenx: I don't think it's possible to get model params from the checkpoint io. 

> I load the two base models in the SpeechLM model's configure_model() function
After loading, what happens when PTL calls `trainer._checkpoint_connector._restore_modules_and_callbacks`? Do you do anything to skip loading checkpoints?

stevehuang52: > After loading, what happens when PTL calls trainer._checkpoint_connector._restore_modules_and_callbacks? Do you do anything to skip loading checkpoints?

when PTL calls `trainer._checkpoint_connector._restore_modules_and_callbacks`, it will try to load the saved checkpoint from previous PEFT jobs, which contains only ASR and PEFT params for LLM, which means the loaded checkpoint doesn't not match with the current params in the model which has ASR and LLM (no PEFT).

I cannot skip the loading process, since it's a PTL default and I do need it to load the previous checkpoint in the second pass of PEFT transform.

> I don't think it's possible to get model params from the checkpoint io.

I was able to hack it by saving the current `sharded_state_dict` and then loading it and return, which worked but not a good practice especially for multi-node training. So I wonder if I can directly convert `sharded_state_dict` into the state dict format returned by `dist_checkpointing.load`?

cuichenx: sorry my question is still when not running PEFT. What happens =when PTL calls trainer._checkpoint_connector._restore_modules_and_callbacks during normal pretraining/finetuning resume?
(Basically PEFT tries to mimic everything that happens during full finetuning, and right before training it adds adapters and loads adapters.)

cuichenx: > So I wonder if I can directly convert sharded_state_dict into the state dict format returned by dist_checkpointing.load?
this should be possible

stevehuang52: > What happens =when PTL calls trainer._checkpoint_connector._restore_modules_and_callbacks during normal pretraining/finetuning resume? Basically PEFT tries to mimic everything that happens during full finetuning, and right before training it adds adapters and loads adapters.

I don't need the change if just using SFT, only need this for PEFT due to the fact that PEFT is added after PTL tries to resume from previous jobs. PEFT doesn't try to mimic what PTL dose, in fact it performs a two-pass loading when resuming from previous PEFT jobs, and that's causing the issue in SpeechLM where we have more than one base model to load while the first pass in PEFT only supports one base model: https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/pytorch/callbacks/peft.py#L485

cuichenx: The first pass of PEFT mimics what happens during SFT resume

stevehuang52: Yes but it does't work when we have two base models.....Anyway, it's better to explain in meeting, but given you're super busy with deepseek, I just subclass MegatronStrategy in a new class in speechlm to make the change, so that there's no change here.

ko3n1g (dismissed):

nemo/collections/common/data/lhotse/cutset.py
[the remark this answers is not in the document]
pzelasko: thanks, I'll fix that in next lhotse release

nemo/collections/common/data/lhotse/nemo_adapters.py
pzelasko: Looking at the implementation of `logging_mode.ONCE`:

```python
    def _logged_once(self, msg, mode):
        PREFIX_LEN = 12
        if mode == LogMode.ONCE:
            if msg[PREFIX_LEN:] in self.once_logged:
                return True
            self.once_logged.add(msg[PREFIX_LEN:])
        return False
```

It only considers the first 12 characters, in that case we will only identify the first corrupted file. If it was considering the full message, I would have been OK with that approach, but as-is it's not good enough.

nemo/collections/speechlm/data/audio_to_text_module.py
pzelasko: I think both of you missed my point so let me re-iterate.

These are some of the options of `TextProcessing`:
> (see arguments such as prompt_template, add_bos/eos/sep, separate_prompt_and_response_with_newline, context/answer_key, etc).

`PromptFormatter` was literally written to replace this class by taking care of all these options - it provides an explicit prompt template and a function that let's you "tell" how exactly to apply these formats to different data structures (audio, text, single turn, multi turn, etc).

Since you are applying **both** to your data, you are bound to run into issues.

nemo/collections/common/data/lhotse/nemo_adapters.py
```diff
@@ -457,7 +461,13 @@ def make_cut_with_subset_inmemory_recording(
         return cut
 
     # Otherwise, apply the memory optimization.
-    cut = cut.truncate(offset=offset, duration=duration, preserve_id=True)
+    try:
+        cut = cut.truncate(offset=offset, duration=duration, preserve_id=True)
+    except Exception as e:
+        raise RuntimeError(
+            f"Lhotse cut.truncate failed with offset={offset}, duration={duration}, recording={recording}: {e}"
+        )
```
pzelasko, line 469:
```suggestion
        ) from e
```

to provide more detail. Can you share what was the error? We should debug whether it's a data issue or code issue.

stevehuang52: it's because the manifest has "duration=0" but the actual audio has length > 0. It'll be better to provide the error details here to users know where went wrong

nemo/collections/common/data/lhotse/nemo_adapters.py
stevehuang52: I see, just reverted the change

nemo/collections/speechlm/data/audio_to_text_module.py
stevehuang52: It's not going to use both at runtime, only one will be used in class instantiation.

> PromptFormatter was literally written to replace this class
Then  PromptFormatter should work with both lhotse and non-lhotse datasets, but it's not. That's why I made a lot of changes. Also PromptFormatter doesn't support using arbitrary free-form prompt during inference, so TextProcessing still has its unique feature by letting users play with different prompt formats during inference.

nemo/collections/speechlm/modules/modality_adapter.py
stevehuang52: just to follow nemo2 convention

nemo/collections/common/data/lhotse/nemo_adapters.py
pzelasko: +1, my point is use the `raise RuntimeError(...) from e` syntax to provide the original exception details as well.

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -545,19 +547,20 @@ def read_nemo_manifest(config) -> tuple[CutSet, bool]:
...
+            f"Initializing Lhotse CutSet from multiple NeMo manifest (is_tarred={is_tarred}) sources with a weighted multiplexer. "
+            f"We found the following sources and weights: "
         )
         cutsets = []
         weights = []
         tar_paths = config.tarred_audio_filepaths if is_tarred else repeat((None,))
         # Create a stream for each dataset.
         for manifest_info, tar_path in zip(config.manifest_filepath, tar_paths):
-            if isinstance(tar_path, (list, tuple, ListConfig)):
+            if is_tarred and isinstance(tar_path, (list, tuple, ListConfig)):
```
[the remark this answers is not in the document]
pzelasko: It already does? if you have non-tarred manifests, why specify tar_paths as a list rather than null? anyway if all dataloader unit tests pass, it should be OK

nemo/collections/speechlm/data/audio_to_text_module.py
pzelasko: >Then PromptFormatter should work with both lhotse and non-lhotse datasets, but it's not.

PromptFormatter itself does not have any lhotse dependency, we specifically designed it with @titu1994 that way. It can be made to work with non-lhotse datasets as well.

Integration of PromptFormatter with lhotse is done via registered prompt fn callbacks which you can, but don't need to use with non-lhotse data - use `prompt.encode_dialog()` API directly.

> Also PromptFormatter doesn't support using arbitrary free-form prompt during inference, so TextProcessing still has its unique feature by letting users play with different prompt formats during inference.

I don't understand - you want to switch e.g. between gemma and llama prompts during inference? Why?

The current prompt formatter absolutely supports multi-turn / free form prompting during inference - again, see `encode_dialog` API.

stevehuang52: >PromptFormatter itself does not have any lhotse dependency, we specifically designed it with @titu1994 that way. It can be made to work with non-lhotse datasets as well.

but all the prompts require `Cut` as input, which also means it's specific to lhotse, and it would be a duplicated work if we need to create prompt functions for non-lhotse datasets as well.

> you want to switch e.g. between gemma and llama prompts during inference? Why?

No specific reason, users can do whatever they want, maybe they just want to slight tweak the prompt format.

Anyway let's leave it as it's for now. Just figured that the current LhotseAudioQuestionAnswerDataset dataset and PromptFormatterTextProcessing doesn't support multi-modal conversation, so there'll be a lot of changes in the future I believe.

nemo/collections/common/data/lhotse/nemo_adapters.py
stevehuang52: done~

ko3n1g:

yaoyu-33 (dismissed):

nemo/collections/speechlm/recipes/pipeline.py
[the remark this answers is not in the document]
zhehuaichen: still see lots of overlap between speech_to_text_llm_train and speech_to_text_llm_validate

nemo/collections/speechlm/utils/text_generation/audio_text_generation_strategy.py
zhehuaichen: can you document these as TODOs in the code

.github/workflows/cicd-main.yml
```diff
@@ -5148,6 +5148,40 @@ jobs:
...
+          --devices 2 \
+          --max_steps 600 \
+          --experiment_dir /tmp/nemo2_speechlm_lora/${{ github.run_id }} \
+          --peft lora \
+          --tp_size 1 \
+          --pp_size 1 \
+          --mbs 2
+
+      AFTER_SCRIPT: |
+        rm -rf /tmp/nemo2_speechlm_lora/${{ github.run_id }}
```
pablo-garay, line 5183: This is not needed to do, but OK doesnt hurt

pablo-garay (approved):

chtruong814 (approved):

chtruong814:
After discussing with the team, we opted to merge this into main but we don't need to cherry-pick to the release branch for 2.2.0.  Will go ahead and force merge this since it passed CI tests earlier and we established that speech domain should not be blocked from merging on linter issues.  We'll address the lint configuration in a separate MR.

PR #10820, opened 2024-10-09
Rreduce storage usage in favor of local storage (part3)

pablo-garay:
Rreduce storage usage in favor of local storage (part3)


ko3n1g (approved):
This is great

PR #10883, opened 2024-10-14
perf recipes and Mcore DistOpt params

malay-nagda:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/collections/llm/gpt/model/base.py
+++ b/nemo/collections/llm/gpt/model/base.py
@@ -195,6 +195,7 @@ class GPTConfig126M(GPTConfig):
     ffn_hidden_size: int = 3072
     num_attention_heads: int = 12
 
+
 @dataclass
 class GPTConfig5B(GPTConfig):
     seq_length: int = 2048
--- a/nemo/collections/llm/gpt/model/base.py
+++ b/nemo/collections/llm/gpt/model/base.py
@@ -195,6 +195,9 @@ class GPTConfig126M(GPTConfig):
     ffn_hidden_size: int = 3072
     num_attention_heads: int = 12
 
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
+
 
 @dataclass
 class GPTConfig5B(GPTConfig):
@@ -203,6 +206,9 @@ class GPTConfig5B(GPTConfig):
     hidden_size: int = 4096
     ffn_hidden_size: int = 16384
     num_attention_heads: int = 32
+
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
 
 
 @dataclass
@@ -222,6 +228,9 @@ class GPTConfig20B(GPTConfig):
     ffn_hidden_size: int = 24576
     num_attention_heads: int = 48
 
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
+
 
 @dataclass
 class GPTConfig40B(GPTConfig):
@@ -231,6 +240,9 @@ class GPTConfig40B(GPTConfig):
     ffn_hidden_size: int = 32768
     num_attention_heads: int = 64
 
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
+
 
 @dataclass
 class GPTConfig175B(GPTConfig):
@@ -239,6 +251,9 @@ class GPTConfig175B(GPTConfig):
     hidden_size: int = 12288
     ffn_hidden_size: int = 49152
     num_attention_heads: int = 96
+
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
 
 
 class GPTModel(L.LightningModule, io.IOMixin, io.ConnectorMixin, fn.FNMixin):
--- a/nemo/collections/llm/gpt/model/base.py
+++ b/nemo/collections/llm/gpt/model/base.py
@@ -141,8 +141,6 @@ class GPTConfig(TransformerConfig, io.IOMixin):
     cross_entropy_loss_fusion: bool = True
     gradient_accumulation_fusion: bool = _grad_accum_fusion_available
     deallocate_pipeline_outputs = True
-    bias_activation_fusion: bool = True
-    bias_dropout_add_fusion: bool = True
 
     transformer_layer_spec: Union[ModuleSpec, Callable[["GPTConfig"], ModuleSpec]] = default_layer_spec
     forward_step_fn: Callable = gpt_forward_step
@@ -197,7 +195,6 @@ class GPTConfig126M(GPTConfig):
     ffn_hidden_size: int = 3072
     num_attention_heads: int = 12
 
-
 @dataclass
 class GPTConfig5B(GPTConfig):
     seq_length: int = 2048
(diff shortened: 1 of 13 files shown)


nemo/collections/llm/recipes/gpt3_175b.py
```diff
@@ -0,0 +1,239 @@
...
+from megatron.core.distributed import DistributedDataParallelConfig
+from pytorch_lightning.callbacks.callback import Callback
+
+from nemo import lightning as nl
+from nemo.collections.llm.api import pretrain
+from nemo.collections.llm.gpt.data.mock import MockDataModule
+from nemo.collections.llm.gpt.model import GPTConfig175B, GPTModel
+from nemo.collections.llm.recipes.log.default import default_log, default_resume, tensorboard_logger
+from nemo.collections.llm.recipes.optim.adam import distributed_fused_adam_with_cosine_annealing
+from nemo.collections.llm.recipes.precision.mixed_precision import bf16_mixed, bf16_with_fp8_mixed
```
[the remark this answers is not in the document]
malay-nagda: @erhoo82 should the default recipe be bf16 or fp8?

nemo/collections/llm/gpt/model/base.py
```diff
@@ -204,6 +207,9 @@ class GPTConfig5B(GPTConfig):
     ffn_hidden_size: int = 16384
     num_attention_heads: int = 32
 
+    bias_activation_fusion: bool = True
+    bias_dropout_add_fusion: bool = True
+
```
JimmyZhang12, line 212: Can these be removed since you already set these in base.py?

nemo/collections/llm/recipes/gpt3_175b.py
```diff
@@ -0,0 +1,242 @@
...
+    recipe = pretrain_recipe(name=name, dir=dir, num_nodes=num_nodes, num_gpus_per_node=num_gpus_per_node, fn=fn)
+
+    recipe.trainer.callbacks.append(
+        run.Config(
+            MegatronCommOverlapCallback,
+            tp_comm_overlap=True,
+            tp_comm_overlap_cfg=userbuffers_fp8_h100_h12288_tp4_mbs1_seqlen2048,
+            defer_embedding_wgrad_compute=True,
+            wgrad_deferral_limit=50,
+            overlap_param_gather_with_optimizer_step=True,
```
JimmyZhang12, line 237: Can you add a comment about why these two are important for the user? and that they already get set automatically by the plugin

nemo/collections/llm/recipes/mixtral_8x22b.py
```diff
@@ -217,7 +220,11 @@ def pretrain_recipe_performance(
     recipe = pretrain_recipe(name=name, dir=dir, num_nodes=num_nodes, num_gpus_per_node=num_gpus_per_node, fn=fn)
     recipe.trainer.callbacks.extend(
         [
-            run.Config(MegatronTokenDropCallback),
+            run.Config(
+                MegatronTokenDropCallback,
+                overlap_param_gather_with_optimizer_step=True,
```
JimmyZhang12, line 225: Thiese are attached to the wrong plugin?

nemo/collections/llm/gpt/model/base.py
malay-nagda: This is in base.py. Do you mean some other file where this is set?

JimmyZhang12: Oh my mistake, could you set these in `GPTConfig`  instead?

malay-nagda: Ok. I'll set them here. But, also we have to set them to False in `NemotronConfig` then.

JimmyZhang12: Do they cause errors if you set them in Nemotron?

malay-nagda: Haven't tried. It is set to False for all nemotron models in Launcher.

nemo/collections/llm/recipes/gpt3_175b.py
malay-nagda: Added

JimmyZhang12 (dismissed):

nemo/collections/llm/recipes/gpt3_175b.py
```diff
@@ -0,0 +1,245 @@
...
+        ),
+        data=run.Config(MockDataModule, seq_length=2048, global_batch_size=2048, micro_batch_size=2),
+        log=default_log(dir=dir, name=name, tensorboard_logger=tensorboard_logger(name=name)),
+        optim=distributed_fused_adam_with_cosine_annealing(max_lr=0.9e-4),
+        resume=default_resume(),
+    )
+
+
+@run.cli.factory(target=pretrain, name=NAME + "_performance")
+def pretrain_recipe_performance(
```
marcromeyn, line 191: I would be a bit afraid that this will blow up the number of recipes that we show in the CLI. Since we allow the user to call the recipe with arguments, what about adding a new arg to the pretrain recipes that will dictate if we use the performance one or not?

malay-nagda: I agree @marcromeyn . But, should we do this as a follow-up PR as this would involve performance recipes for all models (I assume) and the interface for perf recipe/perf mode would also change. Reviewing would be difficult/erroneous?

Also, cc: @erhoo82 @JimmyZhang12

erhoo82: @marcromeyn We need this performance recipe for every key model training because the baseline recipe is missing many performance features and it is hard to communicate with users without having a recipe that has all the important perf features. There is a high risk that a user can miss out many of them without asking us for the actual CLI command line every time.

erhoo82: @ericharper As we discussed, the baseline recipe is for debugging and we need a performance recipe for all model training.

BTW, are we supposed to have different recipes for different sequence lengths? 
Also, one question on the global batch size here.
image

ericharper: Sequence lengths have their own recipe, for example: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/recipes/llama3_70b_64k.py

erhoo82: @ericharper 
Yes, I see that it uses different a CP size for the target seqlen. So, I think it is only TP and CP mapping that needs to be changed. Do we need a full recipe only for this change? Also, I think the global batch size there is wrong.

ericharper: We decided previously to include the full recipe for this but we could revisit it. If there's a bug please send a fix.

@malay-nagda we can merge this as is, but please send a follow up PR with @marcromeyn 's suggestion

ericharper (dismissed):
LGTM. Thanks!

nemo/collections/llm/gpt/model/base.py
malay-nagda: This breaks other model tests in CI. Reverting to model specific fusion params.

JimmyZhang12 (approved):

PR #10943, opened 2024-10-18
Add model_downsampling_factor to NFA cfg

ssh-meister:
The `model_downsampling_factor` parameter was missing in the configuration, which caused the `use_buffered_chunked_streaming` option to malfunction. This PR adds the missing parameter and updates the documentation accordingly.

### DIFF
--- a/tools/nemo_forced_aligner/align.py
+++ b/tools/nemo_forced_aligner/align.py
@@ -88,6 +88,9 @@ Arguments:
     total_buffer_in_secs: float  Length of buffer (chunk + left and right padding) in seconds
     chunk_batch_size: int batch size for buffered chunk inference,
                       which will cut one audio into segments and do inference on chunk_batch_size segments at a time
+    model_downsample_factor: int interval at which the model processes overlapping segments of audio input, impacting 
+                      its ability to capture temporal features in speech (8 for Citrinet and FasConformer models and 
+                      4 for Conformer models).
 
     simulate_cache_aware_streaming: False, if set True, using cache aware streaming to do get the logits for alignment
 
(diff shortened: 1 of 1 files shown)


erastorgueva-nv (changes requested):
Thanks for this!
Actually, the story is that I used to require model_downsample_factor to be in the config, but then I made some logic to calculate it, so that the user doesn't need to worry about it. The logic is here: https://github.com/NVIDIA/NeMo/blob/b6c3c0d5d41d2c695e6a1b23aa573e867b0ec729/tools/nemo_forced_aligner/utils/data_prep.py#L822

But I forgot to update buffered case/buffered case was merged without realizing this was gone.
Could you please update the code so that we calculate model_downsample_factor for the buffered case also?

ssh-meister:
> Thanks for this!
> 
> Actually, the story is that I used to require model_downsample_factor to be in the config, but then I made some logic to calculate it, so that the user doesn't need to worry about it. The logic is here: https://github.com/NVIDIA/NeMo/blob/b6c3c0d5d41d2c695e6a1b23aa573e867b0ec729/tools/nemo_forced_aligner/utils/data_prep.py#L822
> 
> 
> 
> But I forgot to update buffered case/buffered case was merged without realizing this was gone.
> 
> Could you please update the code so that we calculate model_downsample_factor for the buffered case also?

In this case, it’s unlikely we can calculate the `model_downsampling_factor` for the bucketed chunked ASR scenario. The reason is that we need to know the number of tokens generated from inference, but to run the inference, we first need the `model_stride_in_sec`, which is computed as the window stride multiplied by the `model_downsampling_factor`. Since the inference requires this value upfront, it seems unavoidable to include `model_downsampling_factor` in the configuration for the buffered case.

Let me know your thoughts!

PR #11009, opened 2024-10-23
Log_props output in TRT-LLM

oyilmaz-nvidia:
# What does this PR do ?

Adds the functionality to return log_props from TRT-LLM using PyTriton.


scripts/deploy/nlp/query.py
```diff
@@ -38,6 +38,7 @@ def get_args(argv):
     parser.add_argument("-tpp", "--top_p", default=0.0, type=float, help="top_p")
     parser.add_argument("-t", "--temperature", default=1.0, type=float, help="temperature")
     parser.add_argument("-ti", "--task_id", type=str, help="Task id for the prompt embedding tables")
+    parser.add_argument("-lp", "--log_probs", default=False, action='store_true', help="Returns log_probs")
```
athitten, line 41: @oyilmaz-nvidia maybe unrelated to this PR, but curious what's the difference between this script and [nemo/deploy/nlp/query_llm.py](https://github.com/NVIDIA/NeMo/blob/e9d73742dccfc3f6b77946bd0440d4787c73f7ce/nemo/deploy/nlp/query_llm.py) ?

nemo/deploy/nlp/query_llm.py
```diff
@@ -170,7 +170,7 @@ def query_llm(
         repetition_penalty: float = None,
         add_BOS: bool = None,
         all_probs: bool = None,
-        compute_logprob: bool = None,
+        log_probs: bool = False,
         end_strings=None,
```
athitten, line 174: Also curious how `all_probs` is different from `log_probs`. It will be good if we can add some docstrings explaining the difference between these two args to avoid confusion.

oyilmaz-nvidia: I'll actually remove the all_probs.

scripts/deploy/nlp/query.py
oyilmaz-nvidia: Script is for those who would like to use the script and it doesn't require nemo installation. And the class is there for nemo users who want to use python classes.

oyilmaz-nvidia:
@athitten If this PR works for you, I'll do some cleanup and start the CI.

nemo/deploy/nlp/query_llm.py
oyilmaz-nvidia: Removed.

athitten:
> @athitten If this PR works for you, I'll do some cleanup and start the CI.

@oyilmaz-nvidia sorry I dint get a chance to test it out yet. I will test it tomorrow. But the PR looks good otherwise.

athitten (approved):
Tested it locally and it works, although the logProbs of the input tokens are 0.0 and logProbs of output tokens are as expected. It's probably some flag or setting on TRTLLM side to get logProbs of input tokens as well.
Thank you @oyilmaz-nvidia

oyilmaz-nvidia:
Closing this since it'll be included here: https://github.com/NVIDIA/NeMo/pull/11233

PR #11058, opened 2024-10-28
New extended prompt format for Canary, short utterances inference fix, and training micro-optimizations

pzelasko:
# What does this PR do ?

* New prompt format 
  * Prompt formatter definition
  * Special tokenizer build script 
  * Unit tests for the new Canary prompt format.
* Micro-optimizations (16% faster training step)
* Option to enable loss mask on prompt tokens
* Fix for inference of very short utterances

**Collection**: ASR

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/common/prompts/canary2.py
```diff
@@ -0,0 +1,206 @@
...
+            prompt_template=prompt_template, expected_slots=expected_slots, slot_values=slot_values
+        )
+
+
+def map_manifest_values_to_special_tokens(slot_values: dict[str, str]) -> dict[str, str]:
+    slot_values = slot_values.copy()
+
+    any_special_token_present = False
+
+    lang_dict_compat = {"en": "en-US", "es": "es-ES", "fr": "fr-FR", "de": "de-DE"}
```
bonham79, line 113: es-US

nemo/collections/common/prompts/canary2.py
```diff
@@ -0,0 +1,206 @@
...
+        # first, validate the utterance
+        expected_slots = {"source_lang", "target_lang", "pnc"}
+        missing_keys = expected_slots - set(cut.custom)
+        if missing_keys:
+            raise RuntimeError(
+                f"We found cut with ID {cut.id} that is missing the following keys: {missing_keys}"
+                f"Please ensure that every utterance in the input manifests contains these keys."
+            )
+
+        optional_slots = {
```
bonham79, line 169: Is there a check to maintain consistency across slots or are you allowing variable slot sizes? (e.g. sample 1 has `itn` sample 2 does not have that slot.)

nemo/collections/common/tokenizers/canary_tokenizer.py
```diff
@@ -69,22 +71,34 @@ def pad_id(self) -> int:
...
+        Tokenize the input special prompt of Canary family of models.
 
         Required because otherwise self.text_to_ids() returns a different result than what Canary had been trained with.
         """
         ans = []
-        assert text.count('>') == 5, f"Expected exactly 5 special tokens in Canary's prompt, got: {text}."
-        assert text.startswith(CANARY_BOS), text
-        for _ in range(5):
+
+        if text.startswith(CANARY2_BOCTX):
```
bonham79, line 87: Wait, can you explain this? I don't remember this in tokenizer discussions why this needs to be the case.

scripts/speech_recognition/canary/build_canary_1b_special_tokenizer.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
```
bonham79, line 1: I don't believe this script is necessary. It can just be an example in the canary tutorial instead.

scripts/speech_recognition/canary/build_canary_2_special_tokenizer.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
bonham79, line 1: ditto

bonham79 (commented):
So I'm a bit hesitant to create a separate Canary2 prompt class. I think it sets a bad precedent for future Canary releases that could lead to bloat. Is there any motivated reason the original Canary tokenizer class can't just redefined and then there's a separate conversion script to write up that converts the old Canary model into the new version? There's only one anyhow, so it's not like it's going to create absurd cascade problems for users.

nemo/collections/common/prompts/canary2.py
```diff
@@ -0,0 +1,206 @@
...
+                    "nodiarize",
+                    "<|diarize|>",
+                    "<|nodiarize|>",
+                ),
+            },
+        },
+        # User prompt.
+        # This role is used for emotion / LID inference only - use it for two just two decoder inference steps
+        # to retrieve the recognized emotion and language tokens.
+        "user_partial": {
```
krishnacpuvvada, line 79: Leaving a note that this doesn't cover the cases  like "identify language and transcribe with ITN" (which is fine for this version). should put sufficient guards regarding what users are allowed to ask in transcribe fn.

nemo/collections/common/prompts/canary2.py
```diff
@@ -0,0 +1,206 @@
...
+    for cut in cuts:
+        if isinstance(cut, MixedCut):
+            cut = cut._first_non_padding_cut
+        if not isinstance(cut, MonoCut):
+            raise TypeError(
+                f"Expected input audio to have a single channel (required MonoCut/MixedCut, but we received: {cut=})"
+            )
+
+        # first, validate the utterance
+        expected_slots = {"source_lang", "target_lang", "pnc"}
```
krishnacpuvvada, line 161: how about `expected_slots = formatter.expected_slots` 

should we make all the slots expected ie. source_lang, target_lang, pnc, itn, timestamp, diarize and we will set the default tags in data config yaml if something is missing in manifest file ?

nemo/collections/common/prompts/canary2.py
krishnacpuvvada: we are combining different datasets annotated for different things, so the flexibility is required atleast the level of each manifest if not per sample.

nemo/collections/common/tokenizers/canary_tokenizer.py
krishnacpuvvada: this is for decoding with context. the way canary currently does long-form transcripttion is to divide audio into non-overlapping chunks and run inference as independent examples. this is to be able to feed previous chunks inference into current chunks inference (and potentially explore better recognition of names if they were given in context)

nemo/collections/common/tokenizers/canary_tokenizer.py
```diff
@@ -69,22 +71,34 @@ def pad_id(self) -> int:
...
+            # Canary 2 prompt format. It starts with decoder context, which should be tokenized using
+            # a different tokenizer than spl_tokens. We don't really know what it is, so we'll use the
+            # following HACK solution: look up 5th token which is target_lang and tokenize this part
```
krishnacpuvvada, line 90: limitation: this will error out if we are decoding with unk_lang as target_lang

nemo/collections/common/tokenizers/canary_tokenizer.py
bonham79: Oh that makes way more sense

nemo/collections/common/prompts/canary2.py
pzelasko: That will be complicated to implement in transcribe fn in general. We should think about whether a new API specifically for Canary would be cleaner.

nemo/collections/common/prompts/canary2.py
tbartley94: Hmm, most of these are binary prompts. What about just having default values for the prompts instead? Else I believe y'all are adding the extra difficulty of the model needing to learn variable length prompt input.

nemo/collections/common/prompts/canary2.py
tbartley94: +1 to new API, even already the prompt setup can be a bit bulky. Putting thought into API changes before it an actual problem could save some heartache.

scripts/speech_recognition/canary/build_canary_1b_special_tokenizer.py
pzelasko: The people building the new tokenizer for Canary 3 or community Canary forks will have an easier time and won't need to guess how the previous special tokenizers were built :)

tbartley94: Those people will be using the tutorial anyhow no? From my understanding, Canary use hasn't picked up that much this year, so there's not a significant community. It's more likely you'll have greater growth with the current model, so providing backend support may not be crucial.

nemo/collections/common/prompts/canary2.py
pzelasko: it will require rebuilding manifests (or at least data configs) for existing data, a bit of extra headache that I'm not sure is worth the time. WDYT?

nemo/collections/common/prompts/canary2.py
pzelasko: There is no variable length (except if you add decoder context). We indeed have default values for slots here.

nemo/collections/common/tokenizers/canary_tokenizer.py
pzelasko: Yes. Better use single tokenizer in these cases...

scripts/speech_recognition/canary/build_canary_1b_special_tokenizer.py
pzelasko: I'm re-doing the Canary 2 tokenizer right now and am grateful I put these scripts in here. Not removing them :)

nemo/collections/asr/modules/transformer/transformer_modules.py
```diff
@@ -58,27 +58,7 @@ def _build_pos_enc(self, hidden_size, max_sequence_length, device=None):
         self.register_buffer('pos_enc', pos_enc)
 
     def forward(self, position_ids):
-        max_pos_id = position_ids.max()
-        # update positional encoding if needed
-        if max_pos_id >= self._max_sequence_length:
```
pzelasko, line 63: This check is super costly as it triggers a DtoH transfer and CUDA sync on every call to transformer decoder forward, and the proposed solution doesn't work anyway.

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -67,7 +67,8 @@
 
 def lens_to_mask(lens, max_length):
     batch_size = lens.shape[0]
-    mask = torch.arange(max_length).repeat(batch_size, 1).to(lens.device) < lens[:, None]
+    arange = torch.arange(max_length, device=lens.device)
+    mask = arange.expand(batch_size, max_length) < lens.unsqueeze(1)
```
pzelasko, line 71: micro-optimization, removes some copies and memory movement

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -673,24 +674,33 @@ def training_step(self, batch: PromptedAudioToTextMiniBatch, batch_nb):
             return torch.tensor([0.0])
 
         input_ids, labels = batch.get_decoder_inputs_outputs()
+        input_ids_lens = batch.prompted_transcript_lens - 1
```
pzelasko, line 677: fixing off-by-one issue that included an extra padding frame in decoder masks

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -673,24 +674,33 @@ def training_step(self, batch: PromptedAudioToTextMiniBatch, batch_nb):
...
+
+        num_frames = batch.audio_lens.sum().float()
+        num_tokens = batch.prompted_transcript_lens.sum().float()
+        tot_frames = torch.as_tensor(batch.audio.numel(), device=num_frames.device, dtype=torch.float)
+        tot_tokens = torch.as_tensor(batch.prompted_transcript.numel(), device=num_frames.device, dtype=torch.float)
```
pzelasko, line 682: micro-optimizations

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -673,24 +674,33 @@ def training_step(self, batch: PromptedAudioToTextMiniBatch, batch_nb):
...
-        num_frames = batch.audio_lens.sum()
-        num_tokens = batch.prompted_transcript_lens.sum()
-        tot_frames = batch.audio.numel()
-        tot_tokens = batch.prompted_transcript.numel()
         tensorboard_logs = {
             'train_loss': audio_loss,
-            'learning_rate': self._optimizer.param_groups[0]['lr'],
-            'batch_size': batch.audio.shape[0],
+            'learning_rate': torch.as_tensor(self._optimizer.param_groups[0]['lr']),
+            'batch_size': torch.as_tensor(batch.audio.shape[0]),
```
pzelasko, line 703: micro-optimizations (the PTL logger turned out to have an inefficient way of converting scalars to tensors)

nemo/collections/common/prompts/canary2.py
krishnacpuvvada: it shouldn't require us to rebuild manifests. it should be doable using the `tags` we have in train_ds yaml config? 

atleast may be the first part i.e, `expected_slots = formatter.expected_slots` ?

krishnacpuvvada (commented):
LGTM overall; left a minor comment. 
Great work!

nemo/collections/common/prompts/canary2.py
pzelasko: discussed offline - keeping the current code to hide the extra complexity of new canary2 slots for users that are not interested in them

pzelasko: after further conversation - made pnc optional too

krishnacpuvvada (approved):
LGTM

PR #11106, opened 2024-10-30
Fix yaml serialization in io mixin

hemildesai:
Uses __class__ only when data.__qualname__ or similar fails

### DIFF
--- a/nemo/lightning/io/mixin.py
+++ b/nemo/lightning/io/mixin.py
@@ -81,13 +81,29 @@ def _partial_representer_with_defaults(dumper, data):
 
 
 def _safe_object_representer(dumper, data):
+    """
+    Represent a given object as YAML using the specified dumper.
+
+    This function is a fallback for objects that don't have specific representers.
+    If the object has __qualname__ attr, the __target__ is set to f"{inspect.getmodule(obj).__name__}.{obj.__qualname__}".
+    If the object does not have a __qualname__ attr, the __target__ is set from its __class__ attr.
+    The __call__ key is used to indicate whether the target should be called to create an instance.
+
+    Args:
+        dumper (yaml.Dumper): The YAML dumper to use for serialization.
+        data (Any): The data to serialize. This can be any Python object,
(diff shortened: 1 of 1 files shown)


nemo/lightning/io/mixin.py
```diff
@@ -81,15 +81,17 @@ def _partial_representer_with_defaults(dumper, data):
...
 def _safe_object_representer(dumper, data):
-    if not inspect.isclass(data):
-        cls = data.__class__
-        call = True
-    else:
+    try:
         cls = data
+        target = f"{inspect.getmodule(cls).__name__}.{cls.__qualname__}"
         call = False
+    except Exception:
```
cuichenx, line 88: is it possible to specify the exception types here? or is it difficult to enumerate

cuichenx (commented):
Would be good if you could add some comments or doc strings to this function too, thanks!

nemo/lightning/io/mixin.py
hemildesai: I can change it to `AttributeError`

hemildesai:
> Add docstring

Updated in https://github.com/NVIDIA/NeMo/pull/11106/commits/bd903a0c86b8ba847903f9ba8e1aad37f1f5ea6a

nemo/lightning/io/mixin.py
hemildesai: Changed in https://github.com/NVIDIA/NeMo/pull/11106/commits/bd903a0c86b8ba847903f9ba8e1aad37f1f5ea6a

oyilmaz-nvidia (approved):

pablo-garay:
CI passed : https://github.com/NVIDIA/NeMo/actions/runs/11602055172

pablo-garay:
There was a further commit after the triggered CI run but it was a docstring name + name change which shouldnt affect result:
https://github.com/NVIDIA/NeMo/pull/11106/commits/bd903a0c86b8ba847903f9ba8e1aad37f1f5ea6a

PR #11108, opened 2024-10-30
Add scripts for importing a ckpt and running a forward step on it for nemo.collections.llm

hemildesai:
Jupyter notebook with pure python is not possible as of now because of 
```
MegatronStrategy object at 0x70030c368ac0>)` is not compatible with an interactive environment
```

### DIFF
--- a/examples/llm/import_and_forward/forward_step.py
+++ b/examples/llm/import_and_forward/forward_step.py
@@ -1,0 +1,37 @@
+import torch
+from nemo import lightning as nl
+from nemo.lightning import io
+from nemo.lightning.ckpt_utils import ckpt_to_context_subdir
+
+def run_forward(model_path: str):
+    strategy = nl.MegatronStrategy(
+        tensor_model_parallel_size=1,
+    )
+    trainer = nl.Trainer(
+        devices=1,
+        accelerator="gpu",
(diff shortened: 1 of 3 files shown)


examples/llm/import_and_forward/forward_step.py
```diff
@@ -0,0 +1,56 @@
...
+        output = model(
+            input_ids=input_ids,
+            position_ids=position_ids,
+            attention_mask=None,
+        )
+        print(output)
+
+
+if __name__ == "__main__":
+    run_forward("/workspace/starcoder2_3b_nemo2")
```
yaoyu-33, line 56: can you change this to "/path/to/llm_model_nemo2" and add comments saying the user need to insert their model

yaoyu-33, line 56: Can you add comment say the user need to run import ckpt first, or need to replace the path to their own checkpoint

hemildesai: Added in https://github.com/NVIDIA/NeMo/pull/11108/commits/0252d0937fdb1046b64df3fd497e47a5e493315f

yaoyu-33 (approved):

PR #11166, opened 2024-11-05
Integrate TE for HF models

oyilmaz-nvidia:
# What does this PR do ?

Accelerates HF models with TE.


oyilmaz-nvidia:
Closing since the code was moved to this PR: https://github.com/NVIDIA/NeMo/pull/11261

PR #10643, opened 2024-09-26
Update mcore parallelism initialization in nemo2

yaoyu-33:
# What does this PR do ?

Update to use encoder_tp_size and encoder_pp_size in MCore API. This enables more flexible parallelism in encoder-decoder models and VLMs.

**Collection**: [common]

# Changelog 
- Add specific line by line info of high level changes in this PR.

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"
**Pre checks**:
- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #11232, opened 2024-11-08
Use Mcore ModelParallelConfig in strategy parallelism property

hemildesai:
This uses Mcore's ModelParallelConfig directly in `strategy.parallelism`. This should also enable backwards compatibility with different Mcore versions as it relies on the Mcore class now.

### DIFF
--- a/nemo/lightning/fabric/strategies.py
+++ b/nemo/lightning/fabric/strategies.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from contextlib import ExitStack, contextmanager
+from dataclasses import fields
 from datetime import timedelta
 from typing import (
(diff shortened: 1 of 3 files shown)


akoumpa (dismissed):
This makes more sense than the ParallelismConfig, thank you.

JimmyZhang12 (approved):

maanug-nv (approved):

yaoyu-33:
@hemildesai   can you hold on this, I think my previous pr might get affected, since we are deleting encoder_pp_size and encoder_tp_size. https://github.com/NVIDIA/NeMo/pull/10643

marcromeyn (approved):

PR #11241, opened 2024-11-09
Fix: Data from AIStore

monica-sekoyan:
# What does this PR do ?

Changing path patterns to also take into account paths starting with `s3://` and adding cache manifests for NeMoLazyIterator.

**Collection**: common

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/core/classes/modelPT.py
```diff
@@ -902,17 +902,15 @@ def setup(self, stage: Optional[str] = None):
         if stage in ('fit', 'validate'):
             val_deferred_setup = (
                 'validation_ds' in self._cfg
-                and self._cfg.validation_ds is not None
+                and not self._cfg.validation_ds
```
stevehuang52, line 905: Why do we need this change? the change `not self._cfg.validation_ds` seems to be the opposite of the previous `self._cfg.validation_ds is not None`

monica-sekoyan: oh, sorry, the condition should be changed for `self.test_dataloader()`, as it is not None but [], corrected now

nemo/collections/common/data/lhotse/nemo_adapters.py
```diff
@@ -92,6 +92,8 @@ def __init__(
         self.shuffle_shards = shuffle_shards
         self.shard_seed = shard_seed
         paths = expand_sharded_filepaths(path)
+        cache_datastore_manifests(paths)
```
stevehuang52, line 95: Is this necessary? Wondering since this is not in `LazyNeMoTarredIterator`, and that lhotse should be able to automatically fetch and cache the data? @pzelasko what do you think?

pzelasko: Lhotse supports URLs but doesn't perform caching. If we tried caching tarred data we would run out of local space in many cases. For non-tarred data it is probably fine as these datasets would typically be rather small.

I am not familiar with NeMo's original `cache_datastore_manifests` function, does it gracefully handle datasets larger than available local disk space? @monica-sekoyan  @stevehuang52

nemo/utils/data_utils.py
```diff
@@ -46,18 +46,16 @@ def is_datastore_path(path) -> bool:
     """Check if a path is from a data object store.
     Currently, only AIStore is supported.
     """
-    return path.startswith('ais://')
+    return path.startswith('ais://') or path.startswith('s3://')
```
pzelasko, line 49: can we change to `is_valid_url(path)` instead? You can use `from lhotse.utils import is_valid_url` or just paste

```python
def is_valid_url(value: str) -> bool:
    try:
        result = urlparse(value)
        return bool(result.scheme) and bool(result.netloc)
    except AttributeError:
        return False
```

nemo/utils/data_utils.py
```diff
@@ -178,11 +173,11 @@ def get_datastore_object(path: str, force: bool = False, num_retries: int = 5) -
         path: path to an object
         force: force download, even if a local file exists
         num_retries: number of retries if the get command fails
-    
+
     Returns:
         Local path of the object.
     """
-    if path.startswith('ais://'):
+    if path.startswith('ais://') or path.startswith('s3://'):
```
pzelasko, line 180: ditto

nemo/utils/data_utils.py
```diff
@@ -155,7 +150,7 @@ def datastore_path_to_local_path(store_path: str) -> str:
     Returns:
         Path to the same object in local cache.
     """
-    if store_path.startswith('ais://'):
+    if store_path.startswith('ais://') or store_path.startswith('s3://'):
```
pzelasko, line 153: ditto

nemo/utils/data_utils.py
```diff
@@ -298,7 +290,7 @@ def datastore_path_to_webdataset_url(store_path: str):
     Returns:
         URL which can be directly used with WebDataset.
     """
-    if store_path.startswith('ais://'):
+    if store_path.startswith('ais://') or store_path.startswith('s3://'):
```
pzelasko, line 293: ditto

nemo/utils/data_utils.py
```diff
@@ -46,7 +47,11 @@ def is_datastore_path(path) -> bool:
     """Check if a path is from a data object store.
     Currently, only AIStore is supported.
     """
-    return path.startswith('ais://') or path.startswith('s3://')
+    try:
+        result = urlparse(path)
+        return result.scheme in ["ais", "s3"]
```
pzelasko, line 52: Let's make this future-proof - right now it will break if the user provides GCP or Azure URL schema here. You can use: `return bool(result.scheme) and bool(result.netloc)` (see urlparse docs to understand why https://docs.python.org/3/library/urllib.parse.html#url-parsing )

pzelasko (changes requested):
Left a few more comments.

stevehuang52:
@monica-sekoyan could you please address the conflicts so that we can merge this PR?

nemo/collections/common/parts/preprocessing/manifest.py
```diff
@@ -278,12 +279,14 @@ def get_full_path(
             # assume audio_file path is relative to data_dir
             audio_file_path = os.path.join(data_dir, audio_file)
 
-            if is_datastore_path(audio_file_path):
+            if is_datastore_path(audio_file_path) and cached:
                 # If audio was originally on an object store, use locally-cached path
```
stevehuang52, line 283: what is the expected behavior if `is_datastore_path(audio_file_path) == True` and `cached==False`?

nemo/utils/data_utils.py
```diff
@@ -46,10 +47,12 @@ def resolve_cache_dir() -> pathlib.Path:
...
 
 def is_datastore_path(path) -> bool:
-    """Check if a path is from a data object store.
-    Currently, only AIStore is supported.
-    """
-    return path.startswith('ais://')
+    """Check if a path is from a data object store."""
+    try:
+        result = urlparse(path)
+        return bool(result.scheme) and bool(result.netloc)
```
shuoyangd, line 53: nit: Can you explain a little why you are only checking the scheme and the netloc here? It does not seem to be exclusively checking for datastore path. For example, I could throw in an url and still pass the test.

stevehuang52: I think `is_datastore_path` is used to determine whether the filepath is a remote file rather than a local file, so it's not restricted to aistore. It's the other functions that use this `is_datastore_path` assumes that we're only using aistore as the backend, which is not perfect but is also what we only support at the moment, so I think it's fine.

shuoyangd: Sounds good. That's the only reservation I had. Will approve.

stevehuang52:
@pzelasko could you please take another look and see if it's good to merge?

pzelasko (approved):
LGTM

PR #11261, opened 2024-11-12
TE acceleration using callbacks

oyilmaz-nvidia:
# What does this PR do ?

Accelerates HF models with TE.


nemo/lightning/pytorch/accelerate/transformer_engine.py
```diff
@@ -0,0 +1,101 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from types import MethodType
+
+import torch
+from transformer_engine import pytorch as te
+
+
+def accelerate(model, fp8_autocast=False):
```
ananthsub, line 21: why not make the name explicit that this is for transformer engine? accelerate otherwise is very generic

ananthsub, line 21: could you also add a test for this function to check that the expected layers are swapped?

oyilmaz-nvidia: Good point, it was in a class before. Actually it's in a transformer_engine.py file. Let me think about this.

oyilmaz-nvidia: Sure, we have a example and will copy that to the test.

nemo/lightning/pytorch/accelerate/transformer_engine.py
```diff
@@ -0,0 +1,90 @@
...
+    _accelerate(model)
+    if fp8_autocast:
+        apply_fp8_autocast(model)
+
+
+def _accelerate(model):
```
akoumpa, line 27: I would call this something like `replace_nn_modules_with_te_modules` or similar, WDYT @oyilmaz-nvidia ?

nemo/lightning/pytorch/accelerate/transformer_engine.py
```diff
@@ -0,0 +1,90 @@
...
+    for name, module in model.named_modules():
+        if isinstance(module, torch.nn.Linear):
+            has_bias = module.bias is not None
+            if any(p % 16 != 0 for p in module.weight.shape):
+                continue
+            te_module = te.Linear(
+                module.in_features, module.out_features, bias=has_bias, params_dtype=module.weight.dtype
+            )
+            with torch.no_grad():
```
akoumpa, line 36: I see there's torch.no_grad here but not in layernorm/rmsnorm, I wonder if it would be possible to move the torch.no_grad as a function decorator, i.e.:
```
@torch.no_grad
def _accelerate(model):
....
```

nemo/lightning/pytorch/accelerate/transformer_engine.py
```diff
@@ -0,0 +1,90 @@
...
+            te_module.bias.copy_(module.bias)
+            setattr(module, name.split(".")[-1], te_module)
+        elif isinstance(module, torch.nn.RMSNorm):
+            te_module = te.RMSNorm(module.normalized_shape[0], eps=module.eps, dtype=module.weight.dtype)
+            te_module.weight.copy_(module.weight)
+            te_module.bias.copy_(module.bias)
+            setattr(module, name.split(".")[-1], te_module)
+
+
+def te_accelerated(model):
```
akoumpa, line 54: I would call this `is_te_module` WDYT @oyilmaz-nvidia ?

nemo/lightning/pytorch/accelerate/transformer_engine.py
akoumpa: +1 , WDYT about `apply_te_optimizations` ?

nemo/lightning/pytorch/accelerate/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
akoumpa, line 1: @oyilmaz-nvidia any objections to deprecating the accelerate directory?

nemo/lightning/pytorch/accelerate/transformer_engine.py
```diff
@@ -0,0 +1,90 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from types import MethodType
+
+import torch
+from transformer_engine import pytorch as te
```
akoumpa, line 18: Searched in NeMo/requirements and did not find transformer_engine, so I'm thinking importing here might break other paths (e.g. ASR) that don't have TE installed; can we use safe_import instead?
```
from nemo.utils.import_utils import safe_import_from
te, HAVE_TE = safe_import_from("transformer_engine", "pytorch")
```
And then later add an assert statement for HAVE_TE wherever it's used?

nemo/lightning/pytorch/accelerate/__init__.py
oyilmaz-nvidia: I'd rather keep it since there will be more files and possibly other model accelerators in this folder. Having a callback and the code for model acceleration should be separated IMO.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: If we install mcore, TE should be included actually. Let me check this.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: Changed it to te_accelerate. I think it makes sense to use accelerate since that's the reason we use TE. But if you guys can think of a better function name, please suggest.

te_optimization or apply_te_optimization sound like an optimizer. So, that's why I don't think we should use the optimization keyword.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: Changed it to ```_apply_basic_module_replacement``` since we'll have more advanced module replacements.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: Ok, added the safe import.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: Done.

nemo/lightning/pytorch/accelerate/transformer_engine.py
oyilmaz-nvidia: Added a test

akoumpa (dismissed):
LGTM, thank you.

.github/workflows/cicd-main.yml
```diff
@@ -4440,7 +4453,7 @@ jobs:
       - L2_NeMo_2_GPT_Pretraining_no_transformer_engine
       - L2_NeMo_2_GPT_DDP_Param_Parity_check
       - L2_NeMo_2_HF_MODEL_IMPORT
-      - L2_NeMo_2_llama3_pretraining_recipe 
+      - L2_NeMo_2_llama3_pretraining_recipe
```
ko3n1g, line 4456:
```suggestion
      - L2_NeMo_2_llama3_pretraining_recipe
      - L2_HF_Transformer_SFT_TE_Acceleration
```

ko3n1g: please add your test here too

ko3n1g (changes requested):

akoumpa (approved):

ko3n1g (approved):

PR #11324, opened 2024-11-19
build: Move dependencies to `requirements*.txt` files

ko3n1g:
# What does this PR do ?

This PR pushes dependencies from Dockerfile to down to requirements files. This is done for all dependencies except of Apex, because Apex stricly requires `--no-build-isolation`. This is a bad practice and not required for other dependencies, so I decided to separate it.

What's missing to this PR are two extensions:

1. During/after split of release branch, we replace git-dependencies with pypi-dependencies (and version for that specific release)
2. Before pushing to pypi, we comment out/delete all git-dependencies

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -48,7 +48,7 @@ RUN \
 pip install --no-cache-dir --no-build-isolation "apex @ git+https://github.com/NVIDIA/apex.git@${APEX_TAG}"
 pip install --no-cache-dir --extra-index-url https://pypi.nvidia.com \
 -r tools/ctc_segmentation/requirements.txt \
-".[nlp]"
+".[all]"
 EOF
 
 # Copy over M-LM stuff
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -33,47 +33,26 @@ EOF
 
 WORKDIR /workspace
 
-# Install Mamba Dependancy
-ARG CAUSAL_CONV_TAG=v1.2.2.post1
-
-RUN <<"EOF" bash -ex
-# Mamba dependancy installation
-
-git clone --depth 1 --branch ${CAUSAL_CONV_TAG} https://github.com/Dao-AILab/causal-conv1d && \
-  cd causal-conv1d && \
-  python setup.py install && \
-  cd .. && \
-  rm -rf causal-conv1d
-
-EOF
-
-RUN pip install hatchling   # needed to install nemo-run
-ARG NEMO_RUN_TAG=34259bd3e752fef94045a9a019e4aaf62bd11ce2
-RUN pip install nemo_run@git+https://github.com/NVIDIA/NeMo-Run.git@${NEMO_RUN_TAG}
-
 # Install NeMo requirements
+ARG NEMO_RUN_TAG=34259bd3e752fef94045a9a019e4aaf62bd11ce2
 ARG TE_TAG=7d576ed25266a17a7b651f2c12e8498f67e0baea
 ARG MODELOPT_VERSION=0.19.0
 ARG MCORE_TAG=2f67f35b4a9e5596034e93eddeed07009bcf630c
-
 ARG APEX_TAG=810ffae374a2b9cb4b5c5e28eaeca7d7998fca0c
 RUN \
   --mount=type=bind,source=requirements,target=requirements \
   --mount=type=bind,source=tools,target=tools \
   --mount=type=bind,source=setup.py,target=setup.py \
   --mount=type=bind,source=nemo/package_info.py,target=nemo/package_info.py \
-  --mount=type=bind,source=nemo/__init__.py,target=nemo/__init__.py <<"EOF" bash -ex
-pip install --no-cache-dir --no-build-isolation --extra-index-url https://pypi.nvidia.com \
-"transformer-engine @ git+https://github.com/NVIDIA/TransformerEngine.git@${TE_TAG}" \
-"megatron_core @ git+https://github.com/NVIDIA/Megatron-LM.git@${MCORE_TAG}" \
-"nvidia-modelopt[torch]~=${MODELOPT_VERSION}" \
-"apex @ git+https://github.com/NVIDIA/apex.git@${APEX_TAG}" \
-"unstructured==0.14.9" \
-"llama-index==0.10.43" \
-"onnxscript @ git+https://github.com/microsoft/onnxscript" \
+  --mount=type=bind,source=nemo/__init__.py,target=nemo/__init__.py <<"EOF" bash -ex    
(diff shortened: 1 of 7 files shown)


requirements/requirements_nlp.txt
```diff
@@ -8,10 +9,10 @@ gdown
 h5py
 ijson
 jieba
-mamba-ssm==2.2.2; sys_platform == 'linux'
+mamba_ssm @ git+https://github.com/state-spaces/mamba.git@v2.2.2
```
ericharper, line 12: does mamba_ssm have a pypi package we can replace this line with?

requirements/requirements_nlp.txt
```diff
@@ -1,5 +1,6 @@
 accelerated-scan
 boto3
+causal_conv1d @ git+https://github.com/Dao-AILab/causal-conv1d.git@v1.2.2.post1
```
ericharper, line 3: Does causal_conv1d have a pypi package?

requirements/requirements_nlp.txt
ko3n1g: they exist, but installing doesn't really work. We still need to build them, and so far all attempts of building failed because of some missing c-headers. There are many discussions in their project issues around this, and all solutions fall back to git-installs

requirements/requirements_nlp.txt
ko3n1g: same as for mamba

requirements/requirements.txt
ericharper, line 1: Can we add inline comments for optional deps that can't be included in pypi?

ericharper: We should also add default values for the env variables.

Dockerfile.ci
```diff
@@ -33,47 +33,26 @@ EOF
...
+  --mount=type=bind,source=requirements,target=requirements \
+  --mount=type=bind,source=tools,target=tools \
+  --mount=type=bind,source=setup.py,target=setup.py \
+  --mount=type=bind,source=nemo/package_info.py,target=nemo/package_info.py \
+  --mount=type=bind,source=nemo/__init__.py,target=nemo/__init__.py <<"EOF" bash -ex    
+pip install --no-cache-dir --no-build-isolation "apex @ git+https://github.com/NVIDIA/apex.git@${APEX_TAG}"
+pip install --no-cache-dir --extra-index-url https://pypi.nvidia.com \
 -r tools/ctc_segmentation/requirements.txt \
-".[all]"
+".[nlp]"
```
titu1994, line 51: This will break ci tests for all other domains, no?

Dockerfile.ci
```diff
@@ -83,13 +62,12 @@ git checkout ${MCORE_TAG} && \
...
 popd
 export PYTHONPATH="${PYTHONPATH}:/workspace/Megatron-LM"
-
 EOF
 
 # Copy over NeMo code
 COPY ./ ./
 RUN <<"EOF" bash -ex
-pip install --no-cache-dir --no-build-isolation ".[all]"
+pip install --no-cache-dir ".[nlp]"
```
titu1994, line 70: Needs to be reverted to all

requirements/requirements.txt
```diff
@@ -1,7 +1,12 @@
+hatchling 
 huggingface_hub>=0.24
+llama-index==0.10.43
```
titu1994, line 3: Why are these being added to core dependencies ?
Put them in another file and add it to setup.py. 
Core requirements are only for that which all domains require - llama-index is not used by all domains.

requirements/requirements.txt
```diff
@@ -1,7 +1,12 @@
+hatchling
```
titu1994, line 1: Why do we need to switch build toolchain? Please revert.

requirements/requirements.txt
```diff
@@ -1,7 +1,12 @@
+hatchling 
 huggingface_hub>=0.24
+llama-index==0.10.43
+nemo_run @ git+https://github.com/NVIDIA/NeMo-Run.git@${NEMO_RUN_TAG:-main}
 numba
 numpy>=1.22
+nvidia-modelopt[torch]==${MODELOPT_VERSION:-0.19.0}
```
titu1994, line 7: Remove from core requirements. You may put it for export requirements.

requirements/requirements.txt
```diff
@@ -1,7 +1,12 @@
...
 onnx>=1.7.0
+onnxscript @ git+https://github.com/microsoft/onnxscript
```
titu1994, line 9: onnx is not a core requirement

requirements/requirements.txt
```diff
@@ -10,5 +15,7 @@ tensorboard
 text-unidecode
 torch
 tqdm>=4.41.0
+transformer-engine @ git+https://github.com/NVIDIA/TransformerEngine.git@${TE_TAG}
```
titu1994, line 18: Remove TE from core. No compiled dependencies can added at all to pip requirements.txt file, especially core requirements

requirements/requirements.txt
```diff
@@ -10,5 +15,7 @@ tensorboard
...
+unstructured==0.14.9
```
titu1994, line 19: Not a core requirement. Its for processing unstructured data - https://pypi.org/project/unstructured/

requirements/requirements_nlp.txt
titu1994, line 12: Set it to sys_platform == linux

titu1994 (changes requested):

requirements/requirements.txt
ko3n1g: Do you have a suggestion for a better requirements txt? I added all into the top-level file since I'm not familiar with dependencies of our collections.

requirements/requirements.txt
ko3n1g: is vllm a good place?

requirements/requirements.txt
ko3n1g: same: is vllm a good place?

requirements/requirements.txt
ko3n1g: Makes sense. But which file is the best place for it?

requirements/requirements_nlp.txt
ko3n1g: we actually have to move this into `reinstall.sh` since the Pypi package doesn't build all c-headers when done in CPU-only environments. We learned that it's safer to always build from source and so we think its best to have this in the `reinstall.sh` script.

ko3n1g:
Thanks for your detailed review @titu1994! This PR lost a bit traction b/c of the delayed Mcore bump. But it's now ready for another review!

PR #11356, opened 2024-11-21
Fix linear layer replacement

oyilmaz-nvidia:
# What does this PR do ?

Fixes the bug with TE linear layer replacement


akoumpa (approved):
LGTM, but since this slipped in we need a test @oyilmaz-nvidia

PR #11362, opened 2024-11-21
Remove logic to skip checkpoint save if checkpoint exists

ashors1:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/lightning/io/pl.py
+++ b/nemo/lightning/io/pl.py
@@ -155,9 +155,6 @@ class MegatronCheckpointIO(AsyncCompatibleCheckpointIO, IOMixin):
         checkpoint_dir = ckpt_to_weights_subdir(path, is_saving=True)
 
         fs = get_filesystem(checkpoint_dir)
-        if fs.isdir(checkpoint_dir) and dist_checkpointing.check_is_distributed_checkpoint(checkpoint_dir):
-            logging.info(f'Distributed checkpoint at path {checkpoint_dir} already exists, skipping saving')
-            return
         fs.makedirs(checkpoint_dir, exist_ok=True)
 
         validate_sharding_integrity = not (self.validated_consistency and self.assume_constant_structure)
--- a/nemo/lightning/pytorch/strategies/fsdp_strategy.py
+++ b/nemo/lightning/pytorch/strategies/fsdp_strategy.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 import shutil
 from collections import OrderedDict
 from pathlib import Path
--- a/tests/collections/llm/test_mnist_model_nemo2_fsdp.py
+++ b/tests/collections/llm/test_mnist_model_nemo2_fsdp.py
@@ -573,7 +573,7 @@ def run_train_mnist_litautoencoder_with_fsdp_strategy_single_gpu():
                     global_batch_size=2,
                     output_log=False,  # Disable logs to support predict_step
                 ),
-                ckpt_load_optimizer=False
+                ckpt_load_optimizer=False,
             )
             predict_trainer = nl.Trainer(
                 accelerator="gpu",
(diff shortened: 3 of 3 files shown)


mikolajblaz (dismissed):

maanug-nv:
based on current test failure, looks like this requires changes in MCore as well to handle and existing dist ckpt dir. i'd say we should make that change first before merging this

mikolajblaz:
> based on current test failure, looks like this requires changes in MCore as well to handle and existing dist ckpt dir. i'd say we should make that change first before merging this

How can I check the failing tests? The only failing job I see is this one but it doesn't have any useful info: https://github.com/NVIDIA/NeMo/actions/runs/11958751953/job/33339441430?pr=11362

Regardless of the failure, I would recommend ensuring that on the NeMo side the checkpointing logic ensures that we don't save twice to the same directory, otherwise that could lead to silent errors.

We did it for NeMo 1.0 and currently all such circumstances (saving twice to the same dir) are explicitly treated as an error which I believe leads to a better experiments handling from a user perspective

maanug-nv:
from that link, go to the Summary, and see the 'Failed jobs'. 
> I would recommend ensuring that on the NeMo side the checkpointing logic ensures that we don't save twice to the same directory, otherwise that could lead to silent errors.

You're right, I completely agree. Currently on main, async save will crash if the dir exists. Non-async save will just print warning and skip saving. Looks like this change is exposing that problem in one of the tests. 

What I don't understand is why we aren't running the PTL logic that appends a version counter ([1](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/callbacks/model_checkpoint.py#L661-L665), [2](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/callbacks/model_checkpoint.py#L685-L689))

ashors1:
>Currently on main, async save will crash if the dir exists. Non-async save will just print warning and skip saving. Looks like this change is exposing that problem in one of the tests.

With this PR, both async and synchronous checkpointing will error if we try to save to an existing directory, which I think is what we want, right? In that case, maybe we can just update the test to save to different directories

maanug-nv:
ran the failing test locally with your changes, I think FSDPStrategy is missing `import os` so removing previous last checkpoint will fail.

ashors1:
> ran the failing test locally with your changes, I think FSDPStrategy is missing `import os` so removing previous last checkpoint will fail: https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/pytorch/strategies/fsdp_strategy.py#L198-L205

Hm yeah, I noticed that and tried to update but looks like there was a problem with pushing my changes. Let me try again

maanug-nv (approved):

maanug-nv (approved):

nemo/lightning/io/pl.py
```diff
@@ -155,9 +155,6 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], path: _PATH, storage_optio
         checkpoint_dir = ckpt_to_weights_subdir(path, is_saving=True)
 
         fs = get_filesystem(checkpoint_dir)
-        if fs.isdir(checkpoint_dir) and dist_checkpointing.check_is_distributed_checkpoint(checkpoint_dir):
-            logging.info(f'Distributed checkpoint at path {checkpoint_dir} already exists, skipping saving')
-            return
```
ananthsub, line 160: @ashors1 @maanug-nv n00b question: could you describe why this is a fix? is it unexpected behavior that we want to raise more visibly? or something else?

maanug-nv: > Currently on main, async save will crash if the dir exists. Non-async save will just print warning and skip saving. Looks like this change is exposing that problem in one of the tests.
> 
> With this PR, both async and synchronous checkpointing will error if we try to save to an existing directory, which I think is what we want, right? In that case, maybe we can just update the test to save to different directories

This change makes the error visible for synchronous and makes the error message more clear for both. The MCore dist_checkpointing library will raise the error now for both of these cases.

ashors1: Yes, @mikolajblaz removed this in NeMo 1 and asked us to do the same in NeMo 2. We should not attempt to save two checkpoints to the same directory, and if a user ever tries, we'd like to throw an error rather than just skipping saving.

PR #11367, opened 2024-11-21
Fix deploy conflicts in llm.api

hemildesai:
In `llm.api.deploy`, the function is also named `deploy` and the import is also named `deploy` as `from nemo.collections.llm import deploy`. This PR fixes the imports to avoid the name conflicts.

Currently, python gives an attribute error when doing `deploy.` because of the naming conflict.

### DIFF
--- a/nemo/collections/llm/api.py
+++ b/nemo/collections/llm/api.py
@@ -521,7 +521,7 @@ def evaluate(
         model_name, url, tokenizer, max_tokens_to_generate, temperature, top_p, top_k, add_bos
     )
     results = evaluator.simple_evaluate(
-        model=model, tasks=eval_task, limit=limit, num_fewshot=num_fewshot, bootstrap_iters=bootstrap_iters
+        model=model, tasks=eval_task, limit=limit, num_fewshot=num_fewshot, bootstrap_iters=bootstrap_iters, batch_size=8
     )
 
     print("score", results["results"][eval_task])
--- a/nemo/collections/llm/api.py
+++ b/nemo/collections/llm/api.py
@@ -377,7 +377,7 @@ def deploy(
         output_generation_logits (bool): If True builds trtllm engine with gather_generation_logits set to True.
         generation_logits are used to compute the logProb of the output token. Default: True.
     """
-    from nemo.collections.llm.deploy.base import unset_environment_variables
+    from nemo.collections.llm.deploy.base import get_trtllm_deployable, unset_environment_variables
     from nemo.deploy import DeployPyTriton
 
     unset_environment_variables()
@@ -392,7 +392,7 @@ def deploy(
         os.environ['OPENAI_FORMAT_RESPONSE'] = str(openai_format_response)
         os.environ['OUTPUT_GENERATION_LOGITS'] = str(output_generation_logits)
 
-    triton_deployable = deploy.get_trtllm_deployable(
+    triton_deployable = get_trtllm_deployable(
         nemo_checkpoint,
         model_type,
         triton_model_repository,
--- a/nemo/collections/llm/api.py
+++ b/nemo/collections/llm/api.py
@@ -41,7 +41,6 @@ from nemo.lightning.pytorch.callbacks import PEFT, ModelTransform
 from nemo.utils import logging
 from nemo.utils.get_rank import is_global_rank_zero
 
-
 if TYPE_CHECKING:
     from megatron.core.inference.common_inference_params import CommonInferenceParams
     from megatron.core.inference.inference_request import InferenceRequest
@@ -526,7 +525,6 @@ def evaluate(
         limit=limit,
         num_fewshot=num_fewshot,
         bootstrap_iters=bootstrap_iters,
-        batch_size=8,
     )
 
     print("score", results["results"][eval_task])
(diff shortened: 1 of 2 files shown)


nemo/collections/llm/evaluation/base.py
```diff
@@ -74,7 +76,7 @@ def tokenizer_type(self, tokenizer):
         elif isinstance(tokenizer, SentencePieceTokenizer):
             return "SentencePieceTokenizer"
         else:
-            raise ValueError(
+            warnings.warn(
                 "Tokenizer type is not one of SentencePieceTokenizer or HF's AutoTokenizer. Please check "
                 "how to handle special tokens for this tokenizer"
             )
```
ananthsub, line 82: Why is a warning being logged instead of raising an error now?

nemo/collections/llm/evaluation/base.py
```diff
@@ -167,7 +169,7 @@ def generate_until(self, inputs: list[Instance]):
         return results
 
 
-def wait_for_rest_service(rest_url, max_retries=60, retry_interval=2):
+def wait_for_rest_service(rest_url, max_retries=600, retry_interval=2):
```
ananthsub, line 172: could you explain why the max retries need to be increased 10x? if increasing the number of retries, are there downstream dependencies regarding expectations for how long it will take for a service to initialize?

nemo/collections/llm/api.py
```diff
@@ -311,22 +311,22 @@ def deploy(
         output_generation_logits (bool): If True builds trtllm engine with gather_generation_logits set to True.
         generation_logits are used to compute the logProb of the output token. Default: True.
     """
-    from nemo.collections.llm import deploy
+    from nemo.collections.llm.deploy.base import get_trtllm_deployable, unset_environment_variables
```
ananthsub, line 314: is this the main fix, to be specific about imports? this seems to have come up in a few other issues, so should nemo move entirely to be as specific as possible for such imports?

ananthsub (commented):
it'd be really helpful if you updated the PR summary with a description of why this solves deploy conflicts for others to learn, or if there are related issues this fixes. is it purely about imports? if so, why are the other changes needed?

nemo/collections/llm/evaluation/base.py
hemildesai: In my experience the TRTLLM server always took more than 1 minute to be ready. 10 minutes is just a conservative default that I decided to use.

nemo/collections/llm/evaluation/base.py
hemildesai: The tokenizer type is needed here https://github.com/NVIDIA/NeMo/blob/c62cb7d1dc1a23aed47010916a1a6b5e8af14643/nemo/collections/llm/evaluation/base.py#L91-L96 only to decide special kwargs. In case of other tokenizers, we assume that no special kwargs are required unless explicitly added.

nemo/collections/llm/api.py
hemildesai: Yes this is the main fix. I updated the PR description to reflect this.

nemo/collections/llm/evaluation/base.py
athitten: I agree that it's better to keep a conservative defualt. considering PyTriton takes some time to be ready. If 10 min is too large, maybe we can bring it down to 5 min or so. This should cover the base.

athitten (commented):
Thank you very much for taking care of this @hemildesai! . PR LGTM just some minor comments.

nemo/collections/llm/evaluation/base.py
hemildesai: Cool I can add it back.

hemildesai: I think 10 is good for now as anyways it will proceed once the server is ready.

nemo/collections/llm/evaluation/base.py
hemildesai: Cool I can add this back.

nemo/collections/llm/api.py
```diff
@@ -513,18 +513,23 @@ def evaluate(
...
         model_name, url, tokenizer, max_tokens_to_generate, temperature, top_p, top_k, add_bos
     )
     results = evaluator.simple_evaluate(
-        model=model, tasks=eval_task, limit=limit, num_fewshot=num_fewshot, bootstrap_iters=bootstrap_iters
+        model=model,
+        tasks=eval_task,
+        limit=limit,
+        num_fewshot=num_fewshot,
+        bootstrap_iters=bootstrap_iters,
+        batch_size=8,
```
ananthsub, line 529: is adding a batch size intentional for this change too? or a leftover from testing?

hemildesai: Good catch, removed

athitten: That's a good find. I overlooked it. Actually `batch_size` is not supported yet with our code (as we are using a custom model to interface with lm-eval-harness) so even if this is passed it does nothing. So good to remove it to avoid confusion and I will add it with the batching support.

athitten: Oops sorry hadn't seen the latest comment.

athitten (approved):
LGTM, thank you again @hemildesai !

PR #11376, opened 2024-11-22
Add missing test to CICD needed list

pablo-garay:
Add missing test to CICD needed list


HuiyingLi (approved):
Thanks will remember to add next time!

PR #11454, opened 2024-12-03
clean write_on_batch_end

orena1:
No need to repeat all the lines
@pzelasko I think relevant to you


orena1:
@pzelasko did you have a check to look?

PR #11457, opened 2024-12-03
Add nvidia/low-frame-rate-speech-codec-22khz model on docs

Edresson:
# What does this PR do ?

Add [nvidia/low-frame-rate-speech-codec-22khz](https://huggingface.co/nvidia/low-frame-rate-speech-codec-22khz) model on docs


docs/source/tts/data/ngc_models_codec.csv
```diff
@@ -4,3 +4,4 @@ mel_codec_22khz_medium,LibriVox and Common Voice,22050Hz,nemo.collections.tts.mo
...
 mel_codec_22khz_fullband_medium,LibriVox and Common Voice,22050Hz,nemo.collections.tts.models.AudioCodecModel,`mel_codec_22khz_fullband_medium <https://ngc.nvidia.com/catalog/models/nvidia:nemo:mel_codec_22khz_fullband_medium>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/mel_codec_22khz_fullband_medium/versions/v1/files/mel_codec_22khz_fullband_medium.nemo``
 mel_codec_44khz_fullband_medium,LibriVox and Common Voice,44100Hz,nemo.collections.tts.models.AudioCodecModel,`mel_codec_44khz_fullband_medium <https://ngc.nvidia.com/catalog/models/nvidia:nemo:mel_codec_44khz_fullband_medium>`_,``https://api.ngc.nvidia.com/v2/models/nvidia/nemo/mel_codec_44khz_fullband_medium/versions/v1/files/mel_codec_44khz_fullband_medium.nemo``
+audio_codec_low_frame_rate_22khz,LibriVox and Common Voice,22050Hz,nemo.collections.tts.models.AudioCodecModel,`audio_codec_low_frame_rate_22khz <https://huggingface.co/nvidia/Low-Frame-rate-Speech-Codec>`_,``https://huggingface.co/nvidia/Low-Frame-rate-Speech-Codec/resolve/main/Low_Frame-rate_Speech_Codec_21Hz.nemo``
```
rlangman, line 7: Should we remove "ngc" from the name of this csv file? Or create a `huggingface_models_codec.csv`? It looks like this file is only referenced in `checkpoints.rst`.

nemo/collections/tts/models/audio_codec.py
```diff
@@ -684,4 +684,11 @@ def list_available_models(cls) -> List[PretrainedModelInfo]:
         )
         models.append(model)
 
+        model = PretrainedModelInfo(
+            pretrained_model_name="audio_codec_low_frame_rate_22khz",
+            location="https://huggingface.co/nvidia/Low-Frame-rate-Speech-Codec/resolve/main/Low_Frame-rate_Speech_Codec_21Hz.nemo",
+            description="For details about this model please refer to the model card: https://huggingface.co/nvidia/Low-Frame-rate-Speech-Codec",
```
rlangman, line 690: I think our convention would be to name the .nemo file `audio_codec_low_frame_rate_22khz.nemo`

docs/source/tts/data/ngc_models_codec.csv
Edresson: I removed "ngc" from the csv file.

nemo/collections/tts/models/audio_codec.py
Edresson: Unfortunately, We can use this convention for hugging face models because currently, Nemo tries to load the checkpoint with the name Low-Frame-rate-Speech-Codec.nemo (name of the hugging face model card + ".nemo").

Edresson: I think to keep the Hugging face model conventions we should keep the model name as the name of the model card (" Low Frame-rate Speech Codec"). Because it would make it easier for the users to find the model. What do you think guys @anteju @rlangman ?

rlangman: In the definition of `from_pretrained()` https://github.com/NVIDIA/NeMo/blob/main/nemo/core/classes/common.py#L710 the logic is:

1. If there is a "/" in the name like f"{repo_name}/{model_name}", then attempt to download from HF. Otherwise, attempt to download from NGC.
2. If downloading from NGC, download using the information in the `list_available_models()`
3. If downloading from HF, then download file at hardcoded path f"{repo_name}/{model_name}/{model_name}.nemo".

The only models I saw on HF using this logic are Parakeet and Canary. I feel like the downloading logic should be more configurable.

But if we don't want to modify the logic ourselves, I would suggest following their convention of naming the model card all lowercase like "low-frame-rate-speech-codec-22khz". Though I would be alright with uppercase.

Edresson: Ok done, we opted to rename the model card to "low-frame-rate-speech-codec-22khz" to make it consistent with all codecs released and that we will release in the future.

Edresson:
I think now the PR is ready to be merged :). Can you guys please review it? @rlangman @anteju  @blisc

rlangman (approved):

ko3n1g (approved):

PR #11464, opened 2024-12-03
Whisper model support in Lite

oyilmaz-nvidia:
# What does this PR do ?

Adds AutoModelSeq2Seq HF support


nemo/collections/llm/gpt/data/hf_dataset.py
```diff
@@ -61,7 +61,10 @@ def pad_within_micro(batch, pad_token_id):
             max_len = max(map(len, batch))
             return [item + [pad_token_id] * (max_len - len(item)) for item in batch]
 
-        keys = list(filter(lambda x: x in batch[0], ['tokens', 'labels', 'position_ids', 'loss_mask']))
+        print(batch[0])
+        keys = list(
+            filter(lambda x: x in batch[0], ['input_features', 'tokens', 'labels', 'position_ids', 'loss_mask'])
```
akoumpa: any way to avoid this? this change will break LLM peft

oyilmaz-nvidia: Yeah, that was mistakenly pushed. Reverted already.

nemo/collections/llm/gpt/data/hf_dataset.py
```diff
@@ -18,7 +18,7 @@
 from nemo.lightning.pytorch.plugins import MegatronDataSampler
 
 
-class HfDatasetDataModule(pl.LightningDataModule):
+class HFDatasetDataModule(pl.LightningDataModule):
```
pzelasko, line 21: FYI this rename breaks imports

pzelasko:
I made some changes to enable the training to converge (commit https://github.com/NVIDIA/NeMo/pull/11464/commits/c605d12aa89f8674e6a11141a0e7fdd4c9e80c5c):
* Preserve HF tokenizer's Whisper special tokens (prompt)
* Change the precision to bf16 (fp16 results in nan after the first weight update)
* Fix decoder_input_ids and labels for training step

titu1994 (approved):
Looks good !

PR #11537, opened 2024-12-10
Update for NEST release

stevehuang52:
# What does this PR do ?

1. Add script for extracting embeddings from different layers of NEST model (also works for other Conformer-based models). 
2. Fixes ModelPT.py when defer_setup is True, since the current code will skip builing validation dataloaders as they are lists and not None.
3. Update ASR manifest processor to allow `duration=None`
4. Minor documentation update.

**Collection**: [asr,core]


nithinraok (dismissed):
LGTM

examples/asr/conf/ssl/nest/nest_fast-conformer.yaml
```diff
@@ -2,22 +2,22 @@
...
-#  +--------------+---------+--------+-----------+----------------+--------------+------------+
-#  | XXLarge(1.2B)|   1024  |    8   |    42     |        9       |     1e-3     |    False   |
-#  +--------------------------------------------------------------+--------------+------------+ 
-
-
-name: "NEST-FastConformer-SSL"
+#  +--------------+---------+---------+----------+----------------+--------------+------------+---------+
+#  | Model        | d_model | n_heads | n_layers |conv_kernel_size| weight_decay |  xscaling  | use_bias|
+#  +==============+=========+========+===========+================+==============+============+=========+
+#  | Small  (14M) |   176   |    4   |    16     |        9       |     0.0      |    True    |   True  |
```
titu1994, line 8: Have we ever tried just disabling this by default for all model sizes? Something to try in future models

scripts/ssl/extract_features.py
```diff
@@ -0,0 +1,285 @@
...
+    Check if the cache is full and save features to disk
+    """
+    if len(results) == 0 or max_cache < 0 or len(results) < max_cache:
+        return
+    os.makedirs(output_dir, exist_ok=True)
+    logging.info(f"Saving {len(results)} features to {output_dir}")
+
+    for sample_id, audio_file, features_np in tqdm(results, desc="Saving features", total=len(results)):
+        filename = str(audio_file).replace("/", "_").replace(".", "_")
+        if len(filename) > 256:
```
titu1994, line 194: Hmm this shouldnt be silent - how about making it a param of the script ?

scripts/ssl/extract_features.py
```diff
@@ -0,0 +1,285 @@
...
+        pin_memory=True,
+        drop_last=False,
+    )
+
+    # Extract features
+    indices = set()
+    results = []
+    amp_dtype = torch.float16 if args.amp_dtype == "float16" else torch.bfloat16
+    logging.info(f"Extracting features using AMP: {args.use_amp}, dtype: {amp_dtype}")
+    with torch.amp.autocast('cuda' if torch.cuda.is_available() else 'cpu', dtype=amp_dtype, enabled=args.use_amp):
```
titu1994, line 247: Uh why make it a lightning module if you're manually doing forward pass anyway? Its fine by me, just odd.

titu1994 (approved):
Overall looks ok, minor comments

examples/asr/conf/ssl/nest/nest_fast-conformer.yaml
stevehuang52: yeah we should try disabling it by default in in the future

scripts/ssl/extract_features.py
stevehuang52: the 256 limit is from filesystem, so I hard-coded it there

PR #7812, opened 2023-10-26
Add safetensors

coldwaterq:
# What does this PR do ?

By default, state_dicts are being saved as pytorch files which contain pickles. This is a security concern which this PR attempts to start the mitigation for while preserving backwards compatability.

**Collection**: all, but none specifically (core functionality)

# Changelog 
- add nemo.utils.secure.torch_load and nemo.utils.secure.torch_save
- torch.load and torch.save removed from save_restore_connector and alternates of it
- secure.torch_load and torch_save use safetensors when available, however if safe=False torch.load and torch.save are used to allow for backwards compatability
- safe is a named parameter and passed through from save_to and restore_from. The default is safe=False to preserve backwards compatability
- added unit tests to double check backwards compatability and secure functions work correctly.

# Usage
* When using nemo with untrusted .nemo files this will greatly reduce the potential for the user to be attacked.

```python
# Existing usage should be preserved. although the result will include a .safetensor inside the .nemo now as well
model.save_to(save_path=model_save_path)
model.save_to(save_path=model_save_path, safe=False)

# the .nemo will only include the .safetensor.
model.save_to(save_path=model_save_path, safe=True)

# Existing usage should be preserved. although if the .nemo contains a .safetensor, it will be used instead of the pytorch version
model.restore_from(model_save_path)
model.restore_from(save_path=model_save_path, safe=False)

# The restore will fail if a .safetensor isn't available to prevent an untrusted .nemo file from resulting in an exploit
model.restore_from(save_path=model_save_path, safe=False)
```

# Before your PR is "Ready for review"
**Pre checks**:
- [X] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [X] Did you write any new necessary tests?
- [X] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [X] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #11549, opened 2024-12-11
Add safetensor option when saving and restoring models

stevehuang52:
# What does this PR do ?

Re-do changes in a previously closed PR https://github.com/NVIDIA/NeMo/pull/7812 to add safetensor options when saving and restoring models. as customers requested.

CC @galv 

**Collection**: [core,asr,nlp,common]

# Changelog 
- Re-do changes in https://github.com/NVIDIA/NeMo/pull/7812 for current main branch
- add nemo.utils.secure.torch_load and nemo.utils.secure.torch_save
- torch.load and torch.save removed from save_restore_connector and alternates of it
- secure.torch_load and torch_save use safetensors when available, however if safe=False torch.load and torch.save are used to allow for backwards compatability
 - safe is a named parameter and passed through from save_to and restore_from. The default is safe=False to preserve backwards compatability
- added unit tests to double check backwards compatability and secure functions work correctly.

# Usage

Following usage is copied from https://github.com/NVIDIA/NeMo/pull/7812

When using nemo with untrusted .nemo files this will greatly reduce the potential for the user to be attacked.
```
# Existing usage should be preserved. although the result will include a .safetensor inside the .nemo now as well
model.save_to(save_path=model_save_path)
model.save_to(save_path=model_save_path, safe=False)

# the .nemo will only include the .safetensor.
model.save_to(save_path=model_save_path, safe=True)

# Existing usage should be preserved. although if the .nemo contains a .safetensor, it will be used instead of the pytorch version
model.restore_from(model_save_path)
model.restore_from(save_path=model_save_path, safe=False)

# The restore will fail if a .safetensor isn't available to prevent an untrusted .nemo file from resulting in an exploit
model.restore_from(save_path=model_save_path, safe=True)
```

**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/collections/asr/models/clustering_diarizer.py
```diff
@@ -508,6 +511,7 @@ def restore_from(
         override_config_path: Optional[str] = None,
         map_location: Optional[torch.device] = None,
         strict: bool = False,
+        safe: bool = False,
```
stevehuang52, line 514: @tango4j this won't work for diarization models, since the vad and speaker models are restored in the `_init_vad_model` and `_init_speaker_model` methods that don't take `safe` as additional param. Do you think we should fix it or maybe plan in the future?

FredSRichardson:
I just wanted to chime in that if there is any way to get this pull request merged into the main branch and a release branch that would be hugely helpful.  I can hobble along using my clone of this branch, but that's obviously not ideal.  Thank you guys for you great work and for this really helpful contribution in particular!

galv:
Reopened. It got stale over the holidays.

nemo/utils/secure.py
```diff
@@ -0,0 +1,73 @@
...
+    Returns:
+        the loaded tensors
+    """
+    try:
+        return storch.load_file(filename + SAFE_EXTENSION, device=map_location)
+    except FileNotFoundError as e:
+        if safe:
+            raise e
+        else:
+            logging.info(e)
```
galv, line 43: My opinion is that this is going to create a lot of confusingly noise. Basically every time that someone does an unsafe load on a model that wasn't saved with the safe=True option (which is frequent because it's the default), they will get an unhelpful log message of a file not found error, even though the application is otherwise working correctly.

nemo/collections/asr/models/clustering_diarizer.py
galv: My opinion is to leave it for the future. Enabling safe tensors unblocks an important use case for us.

nemo/utils/secure.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Args:
+        tensors: the tensors to save
+        filename: the file to save to
+        safe: whether to save with safetensors
+    Returns:
+        None
+    """
+    try:
+        storch.save_file(tensors, filename + SAFE_EXTENSION)
```
galv, line 63: Why use the safetensors format if the user does not request it? You will basically double the size of the model on the disk this way. This does not seem like a good idea to me.

FredSRichardson:
I just wanted to say I fully endorse this pull request.  I tested it and it works well for my use case.  It will greatly enable my work if some form of this safetensor PR can make it into the main branch.  Thank you!!!

nemo/utils/secure.py
stevehuang52: The old PR enables safetensor by default, but yes I agree that it should be disabled by default, and only enable when required.

FredSRichardson:
Hello!  Any chance of resurrecting this pull request?  Thank you!!!  -Fred

FredSRichardson:
Can this pull request be resurrected?  It would be great if this could make it into the released code some time in the near future.

stevehuang52:
> It would be great if this could make it into the released code some time in the near future.

Thanks for your patience on this matter. This PR still needs a lot of modifications to put it into the future releases, while we currently don't have enough bandwidth to work on it yet. It's on our list for sure, and we will update once we have made progress.

FredSRichardson:
Could you provide some feedback on when this pull request will get merged?  We are at a  point where we need to import NeMo with safetensor support into our client's environment.  Thank you!

stevehuang52:
@nithinraok Do you think someone else can take over and finish the PR? I think the most important part is to switch to only using safetensor if the user desires to enable it.

FredSRichardson:
@nithinraok @stevehuang52 - I'm a little confused by something.  The HuggingFace Canary 1b Flash model at https://huggingface.co/nvidia/canary-1b-flash appears to be a safetensor model.  How is that supported by the NeMo toolkit without this patch?

nithinraok:
I think this needs to be clarified. 

Current safetensors in `https://huggingface.co/nvidia/canary-1b-flash` are for hf transformers support through https://github.com/huggingface/transformers/pull/39062 

@stevehuang52 what is the plan with this PR? How should we proceed with now addition of FC support to transformers?

stevehuang52:
@nithinraok I think the current PR needs some improvement, but the most important one is to only store weights as safetensor if user specifies `safe=true`, while current way is to store both torch ckpt and safetensor therefore doubling the checkpoint size. 

I think the minimum function is to allow users to convert existing nemo checkpoint to use safetensor, then load safetensor checkpoints when doing inference.

nithinraok:
what is the use of it when performing inference with NeMo with added safetensors support with `safe=true`?

nithinraok:
As if we are inferring with NeMo toolkit why not just use torch ckpt as we are doing? If for security reasons then we should drop torch ckpt and just save safetensor in .nemo and not both

stevehuang52:
> As if we are inferring with NeMo toolkit why not just use torch ckpt as we are doing? If for security reasons then we should drop torch ckpt and just save safetensor in .nemo and not both

I'm afraid I'm not the right person to answer. 

@FredSRichardson could you please explain your use case?

FredSRichardson:
@stevehuang52  In our case, we are only able to deliver safetensor models to our client.

To clarify - NeMo without safetensor support is a non-starter for our client.  That is we will have to recommend a different solution.  Relying on a version of NeMo from a pull request branch is not really a sustainable option.

PR #11564, opened 2024-12-12
Adding documentation for packed dataset preparation with context para…

tomlifu:
# What does this PR do ?

This PR adds some documentation in the code for packed dataset preparation with context parallel.
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [x] Documentation

### DIFF
--- a/scripts/nlp_language_modeling/prepare_packed_ft_dataset.py
+++ b/scripts/nlp_language_modeling/prepare_packed_ft_dataset.py
@@ -50,6 +50,16 @@ python scripts/nlp_language_modeling/prepare_packed_ft_dataset.py \
    +output_dir=/path/to/output_folder \
    +pack_sizes=[2048,4096,8192]
    
+when using context parallelism (CP) with packed dataset, CP size needs to be set in the command:
+
+python scripts/nlp_language_modeling/prepare_packed_ft_dataset.py \
+    model.data.train_ds.file_names=[/path/to/training.jsonl] \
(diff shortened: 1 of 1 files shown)


xrennvidia (dismissed):
LGTM. Thanks.

scripts/nlp_language_modeling/prepare_packed_ft_dataset.py
```diff
@@ -50,6 +50,16 @@
    +output_dir=/path/to/output_folder \
    +pack_sizes=[2048,4096,8192]
    
+when using context parallelism (CP) with packed dataset, CP size needs to be set in the command:
+
+python scripts/nlp_language_modeling/prepare_packed_ft_dataset.py \
+    model.data.train_ds.file_names=[/path/to/training.jsonl] \
+    model.data.train_ds.max_seq_length=4096 \
+    +model.context_parallel_size=2 \
```
ashors1, line 58: Maybe it would be safer to make this 
```
++model.context_parallel_size=2 \
```
because if `context_parallel_size` is already defined in the model config, attempting to append `context_parallel_size` to the config will fail. What do you think?

ashors1 (approved):

scripts/nlp_language_modeling/prepare_packed_ft_dataset.py
tomlifu: I think that's better. I've added it in the latest commit.

xrennvidia (approved):

PR #11619, opened 2024-12-17
Add more fine-grained performance metrics

ashors1:
# What does this PR do ?

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -0,0 +1,264 @@
...
+        # use config params only when NOT provided explicitly
+        self.model = model_name
+
+        self.gbs = self.data_cfg.global_batch_size
+        self.enc_seq_len = self.model_cfg.seq_length
+        self.hs = self.model_cfg.hidden_size
+        self.layers = self.model_cfg.num_layers
+        self.ffn_hs = self.model_cfg.ffn_hidden_size
+        self.attention_heads = self.model_cfg.num_attention_heads
+        self.moe_router_topk = self.model_cfg.moe_router_topk
```
cspades, line 73: Nit: When adding a new model formula, this was a bit annoying to deal with so I ended up just getting everything from the `model_config` directly! Do we expect most of these params to be standardized in NeMo / Megatron going forward?

nemo/lightning/pytorch/callbacks/progress_printer.py
```diff
@@ -138,6 +144,7 @@ def on_train_batch_end(self, trainer, pl_module, *_, **__):
             prefix = self.train_description + f" epoch {trainer.current_epoch}, iteration {n-1}/{self.total-1}"
             log_string = self.format_string(prefix, self.average_metrics_dict)
             print(log_string)
+            print(megatron_log_string, flush=True)
```
cspades, line 147: Getting a bunch of unexpected `None`:
```
Training epoch 0, iteration 98/99 | lr: 1.176e-05 | consumed_samples: 792 | global_batch_size: 8 | global_step: 98 | reduced_train_loss: 1.41 | train_step_timing in s: 2.174 | tflops_per_sec_per_gpu: 51.42 | val_loss: 1.472
None  # where does this come from?
Training epoch 0, iteration 99/99 | lr: 1.188e-05 | consumed_samples: 800 | global_batch_size: 8 | global_step: 99 | reduced_train_loss: 1.448 | train_step_timing in s: 2.172 | tflops_per_sec_per_gpu: 51.45 | val_loss: 1.472
None
```

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def _gpt3(self):
+        """Model FLOPs for GPT3 family"""
+
+        vocab_size = LLM_VOCAB_SIZE_MAP["gpt3"]
+
+        return (
+            24 * self.gbs * self.enc_seq_len * self.hs * self.hs
+            + 4 * self.gbs * self.enc_seq_len * self.enc_seq_len * self.hs
+        ) * (3 * self.layers) + (6 * self.gbs * self.enc_seq_len * self.hs * vocab_size)
```
cspades, line 174: Just a reminder to split all model formulas into a separate module and import them here as well as in the NeMo1 implementation: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/common/metrics/perf_metrics.py#L26

nemo/lightning/pytorch/callbacks/progress_printer.py
```diff
@@ -201,3 +208,8 @@ def on_test_batch_end(
 
     def should_log(self, n):
         return n % self.log_interval == 0
+
+    def log_megatron_timers(self, timers):
+        output_string = timers.get_all_timers_string(names=None, normalizer=self.log_interval)
+        if output_string is not None:
+            return output_string + "\n"
```
cspades, line 215: Ah, this is the `None` issue, this function will return either a string or `None`, I just added:
```
    def log_megatron_timers(self, timers):
        output_string = timers.get_all_timers_string(names=None, normalizer=self.log_interval)
        if output_string is not None:
            return output_string + "\n"
        else:
            return ""
```

nemo/lightning/pytorch/callbacks/progress_printer.py
cspades: Found it: https://github.com/NVIDIA/NeMo/pull/11619/files#r1902024482

nemo/lightning/pytorch/callbacks/flops_callback.py
ashors1: I think if we want to unify the flops formulas across nemo1 and nemo2, we need to stick with pulling out the relevant hparams here rather than getting them from the config, because nemo1 and nemo2 have different ways of getting hparams from the config. Does this sound okay to you?

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -0,0 +1,158 @@
...
+            Defaults to None.
+    """
+
+    higher_is_better = True
+
+    def __init__(
+        self,
+        model_config: Dict[str, Any],
+        data_config: pl.LightningDataModule,
+        model_name: Optional[str],
```
maanug-nv, line 56: Seems like this callback won't really work if not given a `model_name`? also does it make sense to have an arg for a custom formula?

ashors1: Good point. I'll update the type hint to make `model_name` required.

> also does it make sense to have an arg for a custom formula?

Yeah, we could to that. In that case, we'd expect exactly one of `model_name` or `custom_formula` to be provided. Does that sound good to you?

maanug-nv: yeah that works. what do you think about having the user explicitly specify the function from `flops_formula.py`? then it could be 1 arg.
i don't have a preference, so whatever makes sense to you

nemo/lightning/pytorch/callbacks/flops_callback.py
cspades: No problem, also hypothetically if we have configs that need to be initialized either here or in on_train_start, they'd have to go here anyway to avoid unnecessary training time latency, so my original point was a bit hopeless.

cspades (approved):
LGTM overall, thanks!

maanug-nv (approved):
LGTM, thanks!

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        if self.model is not None:
+            model_matches = [model for model in model_flops_map if model in self.model]
+            self.model = model_matches[0] if len(model_matches) > 0 else self.model
+        if self.model not in model_flops_map:
+            logging.info(f"FLOPs measurement supported for {list(model_flops_map.keys())}")
+            raise KeyError(f"Failed to extract valid model name from or missing FLOPs calculations for {self.model}")
+
+        total_flops = model_flops_map[self.model](self)
```
ananthsub, line 149: this requires passing the Callback instance to the model flops calculator functions. instead, why not pass a dataclass or config object that specifies the requires fields to calculate this? that would also make writing unit tests easier for those functions

ashors1: That makes sense to me. I updated the code to use a dataclass. Let me know if you have any outstanding concerns.

nemo/utils/flops_formulas.py
```diff
@@ -0,0 +1,135 @@
...
+    enc_seq_len: int
+    hs: int
+    layers: int
+    ffn_hs: int
+    attention_heads: int
+    moe_router_topk: int
+    query_groups: int
+
+
+def gpt3(mdl):
```
ananthsub, line 33: this now takes a config object right? if so, could you add typehints to make it easier for readers to tell?
```suggestion
def gpt3(config: FLOPSConfig) -> int:
```

ananthsub (dismissed):

nemo/utils/flops_formulas.py
ashors1: Yep, done

maanug-nv (approved):

PR #11694, opened 2024-12-20
r2.1.0 ptl fix

pablo-garay:
r2.1.0 ptl fix

### DIFF
--- a/requirements/requirements_lightning.txt
+++ b/requirements/requirements_lightning.txt
@@ -1,7 +1,7 @@
 cloudpickle
 fiddle
 hydra-core>1.3,<=1.3.2
-lightning>2.2.1
+lightning>2.2.1,<=2.4.0
 omegaconf<=2.3
 torchmetrics>=0.11.0
 transformers>=4.45.0


requirements/requirements_lightning.txt
```diff
@@ -1,7 +1,7 @@
 cloudpickle
 fiddle
 hydra-core>1.3,<=1.3.2
-lightning>2.2.1
+lightning>2.2.1,<=2.4.0
```
jstjohn, line 4: Are you addressing the progress printer issue with 2.5? You can fix that directly by calling `on_train_epoch_start` from inside of a new `on_fit_start` in the progressprinter class.

jstjohn: Eg you could add:

```python
    @override
    def on_fit_start(self, trainer: Trainer, pl_module: LightningModule) -> None:
        self.on_train_epoch_start(trainer, pl_module)
```

Right before https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/pytorch/callbacks/progress_printer.py#L114-L120

jstjohn: that got 2.5 working for me with job resumption without the `AttributeError: 'ProgressPrinter' object has no attribute 'total'` error.

pablo-garay: Virtually all tests are failing. Thats why . Needs workaround before 2.5 upgrade

pablo-garay:
This change made tests pass again, so merging it

PR #11712, opened 2024-12-24
Lora ckpt in HF format for NeMo AutoModel

oyilmaz-nvidia:
# What does this PR do ?

Adds support to save Lora ckpt in HF format for NeMo automodel.


nemo/lightning/pytorch/strategies/utils.py
```diff
@@ -117,7 +117,15 @@ def ckpt_to_dir(filepath: Union[str, Path]) -> Path:
 
 
 def create_checkpoint_io(wrapping_ckpt_io=None, **kwargs):
```
akoumpa, line 119: can you add a test to make a checkpoint with NeMo and restore it in huggingface? We have tests now for LLM & VLM.

Also, right now checkpoint saving is disabled in the tests, can you turn it on (minor change in the test command)?

ko3n1g (approved):

nemo/lightning/pytorch/strategies/utils.py
oyilmaz-nvidia: Enabled the ckpt savings in those tests. I'll need to address the restore in a separate PR using AutoResume right after this PR.

pablo-garay (approved):

.github/workflows/cicd-main.yml
```diff
@@ -3675,7 +3676,7 @@ jobs:
     with:
       RUNNER: self-hosted-azure
       SCRIPT: |
-        TRANSFORMERS_OFFLINE=1 python tests/collections/llm/hf/peft.py --model /home/TestData/nlp/hf_gemma/hf_gemma_2b --max-steps 10 --devices 2 --strategy ddp --disable-ckpt
+        TRANSFORMERS_OFFLINE=1 python tests/collections/llm/hf/peft_hf.py --model /home/TestData/nlp/hf_gemma/hf_gemma_2b --max-steps 10 --devices 2 --strategy ddp --disable-ckpt
```
akoumpa, line 3679: @oyilmaz-nvidia thus still has --disable-ckpt

PR #11793, opened 2025-01-08
Add "_skipme" option to Lhotse Dataloading

monica-sekoyan:
# What does this PR do ?

Lhotse Dataloading will ignore samples where "_skipme" is set to 1 if not instructed otherwise with `ds.ignore_skipme=True`

**Collection**: [common]

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -176,6 +176,8 @@ class LhotseDataLoadingConfig:
     # Enables iteration of NeMo non-tarred manifests that don't have a "sampling_rate" key without performing any I/O.
     # Note that this will not allow actual dataloading; it's only for manifest iteration as Lhotse objects.
     metadata_only: bool = False
+    # Forces to ignore the "_skipme" key in data entry resulting into not skipping loading of entries with "_skipme" set to 1.
+    ignore_skipme: bool = False
```
nithinraok, line 180: Prefer better naming here. double negation -> easy to confuse. 

`drop_skipme`? -> if true drops the samples for samples if _skipme key in them set to true.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -443,6 +445,14 @@ def get_lhotse_sampler_from_config(config, global_rank, world_size, tokenizer=No
     cuts, use_iterable_dataset = read_cutset_from_config(config)
     use_iterable_dataset = determine_use_iterable_dataset(use_iterable_dataset, config)
 
+    if not config.ignore_skipme:
+        cuts = cuts.filter(lambda cut: not cut.custom.pop("_skipme", 0))
```
nithinraok, line 449: Don;t we need to check for if cut.custom is None case here for robustness?

monica-sekoyan: they are always added [here ](https://github.com/NVIDIA/NeMo/blob/5d8baa4ada1313c704d4c51bfea6eb7a988faa58/nemo/collections/common/data/lhotse/nemo_adapters.py#L52
), but for further possible changes, added the check as well

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -176,6 +176,8 @@ class LhotseDataLoadingConfig:
     # Enables iteration of NeMo non-tarred manifests that don't have a "sampling_rate" key without performing any I/O.
     # Note that this will not allow actual dataloading; it's only for manifest iteration as Lhotse objects.
     metadata_only: bool = False
+    # If set to True entries with "_skipme: 1" will be skipped during data loading.
+    drop_skipme: bool = True
```
pzelasko, line 180: Remove the option - simply always ignore these entries.

pzelasko (commented):
Please add a unit test

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -443,6 +445,14 @@ def get_lhotse_sampler_from_config(config, global_rank, world_size, tokenizer=No
     cuts, use_iterable_dataset = read_cutset_from_config(config)
     use_iterable_dataset = determine_use_iterable_dataset(use_iterable_dataset, config)
 
+    if config.drop_skipme:
+        cuts = cuts.filter(lambda cut: not cut.custom.pop("_skipme", False) if hasattr(cut, "custom") else True)
```
pzelasko, line 449:
```suggestion
        cuts = cuts.filter(lambda cut: "_skipme" not in cut.custom if hasattr(cut, "custom") else True)
```

I suggest the following change: skip whenever this key is present. This would let us adopt the following convention: `{"_skipme": "reason for skipping"}`

monica-sekoyan: @pzelasko , for consistency in manifest entries wouldn't it be better to have "_skipme" keys for every entry with values "False" or "0" for entries we do not need to skip and "reason for skipping" for entries we need to skip?

nemo/collections/common/data/lhotse/sampling.py
```diff
@@ -315,3 +315,20 @@ def _measure_tokens(cut: Cut) -> int:
...
     return num_tokens / cut.duration
+
+
+class SkipmeFilter:
+    """
+    Callable, returns ``True`` if a cut's "_skipme" is set and ``False`` otherwise.
+    Acts as a pass-through for objects of other type than Cut.
+    """
+
+    def __init__(self) -> None:
```
pzelasko, line 326: you can remove init if it's empty, it will be auto-created

nemo/collections/common/data/lhotse/sampling.py
```diff
@@ -315,3 +315,20 @@ def _measure_tokens(cut: Cut) -> int:
 def _measure_tps(cut: Cut) -> float:
     num_tokens = _measure_tokens(cut)
     return num_tokens / cut.duration
+
+
+class SkipmeFilter:
```
pzelasko, line 320: Could we name this `PlaceholderFilter` instead?

tests/collections/common/test_lhotse_dataloading.py
```diff
@@ -37,16 +37,18 @@
 
 @pytest.fixture(scope="session")
 def cutset_path(tmp_path_factory) -> Path:
-    """10 utterances of length 1s as a Lhotse CutSet."""
+    """12 utterances of length 1s as a Lhotse CutSet."""
```
pzelasko, line 40: Could you avoid modifying existing fixtures and create a new one, dedicated specifically to this feature? There are many tests depending on those, I don't want to further increase test setup complexity.

To clarify: my point is that with this change, every test in this file is testing the "skipme" functionality. It is simpler to debug if there is just one test that's testing it instead.

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -365,7 +366,7 @@ def read_lhotse_manifest(config) -> tuple[CutSet, bool]:
             logging.info(f"Initializing Lhotse Shar CutSet (tarred) from a single data source: '{config.shar_path}'")
             cuts = CutSet.from_shar(
                 **_resolve_shar_inputs(config.shar_path, metadata_only), shuffle_shards=True, seed=shard_seed
-            )
+            ).filter(SkipmeFilter())
```
pzelasko, line 369: Instead of applying these in multiple branches, please apply this filter in a single place: just before `return` statement in `read_cutset_from_config` definition.

pzelasko (changes requested):

nemo/collections/common/data/lhotse/cutset.py
monica-sekoyan: I did so to filter the entries before they can be mixed with some weights
(because filtering before or after applying weights can yield different results)

nithinraok (approved):
LGTM. @pzelasko for final approval

stevehuang52:
A general discussion, as we're adding more and more functionalities to `nemo/collections/common/data/lhotse/cutset.py` and `nemo/collections/common/data/lhotse/dataloader.py`, can we have a more modular design such that each downstream task can customize the behavior of manifest reading and data augmentation? For example in nemo map dataset we have different collections for different tasks.

Also, the flag `_skipme` seems to ambiguous and doesn't provide enough information when reading the manifests. What is the reason that these entries need to be skipped? @monica-sekoyan could you please remind me of the reason? If we're skipping entries whenever there's a `_skipme` flag in it, it seems to mean that they should not be used in any application, and it might be better to create a "cleaned" version of the entries that don't need to be skipped.

If there're certain attributes that determine which entries to be skipped, we can use different filters for them. For example, ASR/AST might skip samples that don't have ground-truth text, so we can use a filter that skips the entries that don't have `text` field. Or if we only want to include clean audios, we can use a filter to skip entries that have `"noisy": true`.

pzelasko:
> can we have a more modular design such that each downstream task can customize the behavior of manifest reading and data augmentation? 

Yes, great point. For manifest reading this is already possible using `@data_type_parser` auto-registering decorator but I need to document it better (it has code docs and is mentioned in multimodal lhotse dataloading tutorial but it needs to be gathered into one place and explained together with its purpose).

For data augmentation we need to develop this - will follow up.

monica-sekoyan:
> Also, the flag `_skipme` seems to ambiguous and doesn't provide enough information when reading the manifests. What is the reason that these entries need to be skipped? @monica-sekoyan could you please remind me of the reason?
> 
@stevehuang52 I have added a section in docs to better explain this key. So, for example I have large amount of audios in tar files and the corresponding sharded_manifests. At some point I want to filter out some samples from this dataset based on some criteria, let's say because of the low character-rate. For me, not to copy the whole data and retar it again, it is more convenient to remove these entries from the manifests. But for not causing misalignment with tar file's audios and sharded manifest, I can not fully remove these entries (because we are starting the iteration from tar file content itself). Technically, I can currently with ``skip_missing_manifest_entries`` keys from Lhotse. But what if after filtration all the entries of manifest_n.json are filtered out, then we would not have that manifest, and this will cause another issue. So the solution, we are currently proposing is to add ``_skipme`` key to the entries in the manifests we want to skip. Its value can be either ``True`` or the reason for skipping (e.g. ``low char-rate``). 

with the proposal of @pzelasko not to complexify Lhotse data loading params, we are not adding an option to ignore ``_skipme`` keys, and for the use-cases when you need all the data in that manifests again, you need to have 2 versions: 1 with the ``_skipme`` keys set and representing filtered subset of the data and 1 without ``_skipme`` keys with all the data available in tar files.

CC: @nithinraok

pzelasko (approved):
Fantastic work @monica-sekoyan, LGTM!

PR #11852, opened 2025-01-14
Removing qk layernorms for DiT diffusion attention modules as qk layernorm is not supported for CrossAttention

Sy-Zhang:
* Docker Image: `nvcr.io/nvidia/nemo:24.12`

* Command: `torchrun --nproc-per-node 8 nemo/collections/diffusion/train.py --yes --factory pretrain_xl`

* Error: `CrossAttentionSubmodules.__init__() got an unexpected keyword argument q_layernorm`

* Reason: `megatron.core.transformer.attention.CrossAttentionSubmodules` does not have argument `q_layernorm` and `k_layernorm`.


ethanhe42:
hi @Sy-Zhang we have a change to add qk LN to cross attention. it hasn't been merged to megatron-lm yet. you can try the container nvcr.io/nvidia/nemo:cosmos.01

PR #11870, opened 2025-01-16
fix convert_llama_hf_to_nemo.py llama31 argument

j40903272:
# What does this PR do ?

Fix conversion script arguments. The original script using type=bool convert empty strings to False and non-empty strings to True. Might not be an ideal way of using it.

**Collection**: NLP

# Changelog 
- Fix convert_llama_hf_to_nemo script --llama31 arguments.

# Usage

```shell
$ scripts/checkpoint_converters/convert_llama_hf_to_nemo.py --llama31  # when it is llama31
$ scripts/checkpoint_converters/convert_llama_hf_to_nemo.py                  # when it is not llama31
```

### DIFF
--- a/scripts/checkpoint_converters/convert_llama_hf_to_nemo.py
+++ b/scripts/checkpoint_converters/convert_llama_hf_to_nemo.py
@@ -64,9 +64,7 @@ def get_args():
     )
     parser.add_argument(
         "--llama31",
-        type=bool,
-        default=True,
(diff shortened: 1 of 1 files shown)


j40903272:
@MaximumEntropy, @ericharper, @ekmb, @yzhang123, @VahidooX, @vladgets, @okuchaiev

Can anyone help with the review?

PR #11911, opened 2025-01-21
Add New Transformer Backbone for TTS Models

blisc:
# What does this PR do ?

Add new transformer backbone for TTS models.

**Collection**: tts

# Changelog 
- Add new transformer backbone for TTS models
  
**PR Type**:
- [x] New Feature

### DIFF
--- a/nemo/collections/tts/modules/transformer_2501.py
+++ b/nemo/collections/tts/modules/transformer_2501.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/nemo/collections/tts/modules/transformer_2501.py
+++ b/nemo/collections/tts/modules/transformer_2501.py
@@ -681,6 +681,12 @@ class Transformer(torch.nn.Module):
             output <torch tensor> (B, T1, C): Output tensor
             attn_probabilities <list>: Attention probabilities of each layer
         """
+        if isinstance(cond, list) and len(self.layers) < len(cond):
+            raise ValueError(
+                f"Insufficient Transformer layers for multiple conditionals. Each layer must cross-attend one conditional."
+                f"Found {len(self.layers)} layers for {len(cond)} conditionals."
+            )
+
         if self.use_learnable_pos_emb:
             positions = torch.arange(x.size(1), device=x.device).unsqueeze(0)
             x = x + self.position_embeddings(positions)
--- a/nemo/collections/tts/modules/transformer_2501.py
+++ b/nemo/collections/tts/modules/transformer_2501.py
@@ -630,7 +630,7 @@ class Transformer(torch.nn.Module):
 
     @staticmethod
     def _init_weights_gpt2(module):
-        if isinstance(module, (torch.nn.Linear, torch.nn.Embedding)):
+        if isinstance(module, (torch.nn.Linear, torch.nn.Embedding, torch.nn.Conv1d)):
             torch.nn.init.normal_(module.weight, mean=0.0, std=0.02)
 
         if isinstance(module, torch.nn.Linear) and module.bias is not None:
--- a/nemo/collections/tts/modules/transformer_2501.py
+++ b/nemo/collections/tts/modules/transformer_2501.py
@@ -1,0 +1,704 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


nemo/collections/tts/modules/transformer_2501.py
```diff
@@ -0,0 +1,704 @@
...
+
+        self.conv = torch.nn.Conv1d(
+            in_channels,
+            out_channels,
+            kernel_size=kernel_size,
+            stride=stride,
+            padding=padding,
+            dilation=dilation,
+            bias=bias,
+        )
```
paarthneekhara, line 83: Adding this initialization helps stabilize training (after this line) as per my observations `torch.nn.init.normal_(self.conv.weight, mean=0.0, std=0.02) `
This was there is `experimentalt5tts`

blisc: So, we will not add this here. o_net is handled in Transformer init, but proj was not. I will add this init to the projection layer.

nemo/collections/tts/modules/transformer_2501.py
```diff
@@ -0,0 +1,704 @@
...
+                )
+        else:
+            return cond, cond_mask, attn_prior
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        x_mask: torch.Tensor,
+        cond: Optional[Union[torch.Tensor, List[torch.Tensor]]] = None,
+        cond_mask: Optional[Union[torch.Tensor, List[torch.Tensor]]] = None,
```
XuesongYang, line 664: nit: we now support the types of `cond` either `torch.Tensor` or a list of `torch.Tensor`. Is it better to unify the two as only a list of `torch.Tensor`? in this case, a list of a single tensor corresponds to a single encoder, and a list of multipe tensors corresponds to multiple encoders?

nemo/collections/tts/modules/transformer_2501.py
```diff
@@ -0,0 +1,704 @@
...
+            cond <torch tensor> (B, T2, C) or list of such tensors (from different encoders)
+            cond_mask <bool mask> (B, T2): Multiplicative mask where True means we keep the input, False we zero it
+                out or list of such tensors (from different encoders) output <torch tensor> (B, T1, C)
+            multi_encoder_mapping <list> <int>: None or Same size as n_layers, value indicates which cond input to use
+                for this layer
+
+        Returns dict with keys:
+            output <torch tensor> (B, T1, C): Output tensor
+            attn_probabilities <list>: Attention probabilities of each layer
+        """
```
XuesongYang, line 683: we should catch exceptions earlier here where the number of layers should be at least 2 when applying multiple conditions.

```suggestion
        """
        if isinstance(cond, list) and len(self.layers) < 2:
            raise ValueError(
                f"The number of Transformer layers should be no less than 2 when applying multiple conditionals. "
                f"Observed {len(self.layers)} layers on {len(cond)} conditionals."
            )
        
```

nemo/collections/tts/modules/transformer_2501.py
blisc: Fixed

tests/collections/tts/modules/test_transformer_2501.py
```diff
@@ -0,0 +1,164 @@
...
+            has_xattn=self.has_xattn,
+            is_causal=self.is_causal,
+            max_length_causal_mask=self.max_length_causal_mask,
+        )
+
+        # Check model init
+        assert torch.isclose(torch.mean(model.layers[0].pos_ff.proj.weight), 0.0)
+        assert torch.isclose(torch.std(model.layers[0].pos_ff.proj.weight), 0.02)
+        assert torch.isclose(torch.mean(model.layers[0].pos_ff.o_net.weight), 0.0)
+        assert torch.isclose(torch.std(model.layers[0].pos_ff.o_net.weight), 0.02 / math.sqrt(2.0))
```
XuesongYang, line 81: bug: *** AttributeError: 'ConvolutionLayer' object has no attribute 'weight'.

```suggestion
        assert torch.isclose(torch.mean(model.layers[0].pos_ff.proj.conv.weight), 0.0)
        assert torch.isclose(torch.std(model.layers[0].pos_ff.proj.conv.weight), 0.02)
        assert torch.isclose(torch.mean(model.layers[0].pos_ff.o_net.conv.weight), 0.0)
        assert torch.isclose(torch.std(model.layers[0].pos_ff.o_net.conv.weight), 0.02 / math.sqrt(2.0))
```

XuesongYang: fixed.

nemo/collections/tts/modules/transformer_2501.py
XuesongYang: @paarthneekhara what is your thoughts on that?

XuesongYang (approved):
LGTM. Thank you guys!

@paarthneekhara there is a nit-pick regarding `cond` to discuss, but not a blocker to merge the code.

PR #11965, opened 2025-01-27
Add onnxruntime-gpu dependency

oyilmaz-nvidia:
# What does this PR do ?

Adding onnxruntime-gpu dependency since it's needed to run onnx and trt based inference.

### DIFF
--- a/nemo/export/tensorrt_lazy_compiler.py
+++ b/nemo/export/tensorrt_lazy_compiler.py
@@ -291,8 +291,8 @@ def parse_groups(
     """
     groups: Tuple[Union[torch.Tensor, List[torch.Tensor]], ...] = tuple()
     cur = 0
-    for l in range(len(output_lists)):
-        gl = output_lists[l]
+    for i in range(len(output_lists)):
+        gl = output_lists[i]
         assert len(gl) == 0 or len(gl) == 1
         if len(gl) == 0 or gl[0] == 0:
             groups = (*groups, ret[cur])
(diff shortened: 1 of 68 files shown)


requirements/requirements.txt
```diff
@@ -2,6 +2,7 @@ huggingface_hub>=0.24
 numba
 numpy>=1.22
 onnx>=1.7.0
+onnxruntime-gpu>=1.20.1
```
borisfom, line 5: Well we kind of kept onnxruntime optional - should not be in requirements.

nemo/export/tensorrt_lazy_compiler.py
```diff
@@ -189,7 +189,7 @@ def set_inputs(self, feed_dict, stream):
 
         def try_set_inputs():
             for binding in self.input_names:
-                t = feed_dict.get(self.input_table[binding], None)
+                t = feed_dict.get(self.input_table[binding], None).cuda()
```
borisfom, line 192: Well we assumed inputs should be on GPU already, is it really necessary ?

oyilmaz-nvidia: Actually the inputs in my example are already on GPU and without cuda() call, it's giving the memory error. I guess if the tensor is already on GPU, that code shouldn't do anything anyways, no?

requirements/requirements.txt
oyilmaz-nvidia: We need the onnx runtime. Moving it to the requirements_nlp.txt. This file is also used for building the container.

nemo/export/tensorrt_lazy_compiler.py
borisfom: Well, In PyTorch, the `cuda()` method moves a tensor to the GPU associated with the currently active CUDA device. That's apparently why you need it - are you using multiple GPUs ? 
We may do better by setting that device current - can you check with your example ?

oyilmaz-nvidia: Actually you are right, the model wasn't on the GPU memory and the inputs were moved based on where the model is on. So, this update is not needed. Removing this update from the PR. We only need to add the onnx runtime requirement.

PR #11969, opened 2025-01-28
Adding TFLOPs callback for Multimodal models and NeVA calculator

parthmannan:
# What does this PR do ?

Adds TFLOPs callback for multimodal models. Works for any model architecture which has more than one model config in the architecture.

# Changelog 
- Adds new class for FLOPs calculation for Multimodal models which takes in a dictionary of model configs.
- Dictionary defines model name with model config.
- Also adds a few non-text modality specific config args for FLOPs calculation.

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo/lightning/pytorch/callbacks/flops_callback.py
+++ b/nemo/lightning/pytorch/callbacks/flops_callback.py
@@ -22,7 +22,7 @@ from lightning.pytorch.callbacks import Callback
 from nemo.lightning.pytorch.callbacks import PEFT
 from nemo.utils import flops_formulas, logging
 
-__all__ = ["FLOPsMeasurementCallback"]
+__all__ = ["FLOPsMeasurementCallback", "MM_FLOPsMeasurementCallback"]
 
 
 class FLOPsMeasurementCallback(Callback):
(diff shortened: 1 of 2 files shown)


nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -162,3 +162,126 @@ def eval_model_flops(self):
...
+            self.flops_config_dict[model_name] = flops_formulas.FLOPSConfig(**kwargs)
+
+        self.avg_train_step_time = 0
+
+    def eval_model_flops(self):
+        """
+        Calculate model FLOPs for a given model
+        """
+
+        model_flops_map = {
```
ashors1, line 265: It looks like this `model_flops_map` is a superset of `FLOPsMeasurementCallback`'s `model_flops_map`. Is there a way to unify `FLOPsMeasurementCallback`  and `MM_FLOPsMeasurementCallback` into a single class? Or would that be too much effort?

nemo/utils/flops_formulas.py
```diff
@@ -13,21 +13,31 @@
...
 
-    gbs: int
-    enc_seq_len: int
-    hs: int
-    layers: int
-    ffn_hs: int
-    attention_heads: int
-    moe_router_topk: int
-    query_groups: int
+    gbs: Optional[int] = None
```
ashors1, line 25: Are some of these attributes required? For example, I believe `gbs` should be specified regardless of which model we're using. Can we be explicit about which of these are required?

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -162,3 +162,126 @@ def eval_model_flops(self):
...
+        data_config: pl.LightningDataModule,
+    ):
+        self.data_cfg = data_config
+        self.flops_config_dict = dict()
+
+        gbs = self.data_cfg.global_batch_size
+        for model_name, model_cfg in model_name_config_dict.items():
+            hs = model_cfg.hidden_size
+            if model_name in ["hf_clip_vit_l"]:
+                layers = model_cfg.num_hidden_layers
```
malay-nagda, line 193: can we update kwargs (`kwrgs['layers']=model_cfg.num_hidden_layers`, etc) directly here in this if-else block and remove `kwargs.update` in the if-else block below (L225-L255)? 

this will avoid extra/repetitive code.

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -162,3 +162,126 @@ def eval_model_flops(self):
...
+        model_name_config_dict (dict):
+            Dictionary containing all the individual model configs that make up the multi-modal model.
+        data_config (pl.LightningDataModule): Data module being used in the experiment.
+    """
+
+    higher_is_better = True
+
+    def __init__(
+        self,
+        model_name_config_dict: dict(),
```
malay-nagda, line 183:
```suggestion
        model_name_config_dict: dict,
```

nemo/lightning/pytorch/callbacks/flops_callback.py
```diff
@@ -162,3 +162,126 @@ def eval_model_flops(self):
...
+                attention_heads = model_cfg.num_attention_heads
+                moe_router_topk = model_cfg.moe_router_topk
+
+            try:
+                query_groups = model_cfg.num_query_groups
+                if query_groups is None:
+                    query_groups = attention_heads
+            except:
+                # Multi-modal models use HF model configs which may/may not define num_query_groups
+                pass
```
malay-nagda, line 219:
```suggestion
            if hasattr(model_cfg, "num_query_groups"):
                query_groups = model_cfg.num_query_groups
                if query_groups is None:
                    query_groups = attention_heads
```

nemo/lightning/pytorch/callbacks/flops_callback.py
parthmannan: I have made a single global `model_flops_map` for both classes. Let me know what you guys think.

nemo/lightning/pytorch/callbacks/flops_callback.py
parthmannan: Done. Thanks for the suggestion.

nemo/lightning/pytorch/callbacks/flops_callback.py
ashors1: That's fine with me, but won't `FLOPsMeasurementCallback` not work with some of these models without changes because we don't specify some of the attributes we need in the config? For example, the flops computation for vit requires `patch_dim`, which isn't specified from `FLOPsMeasurementCallback`.

If making `FLOPsMeasurementCallback` support all flops formulas is too much work, I'm okay with keeping the implementations separate.

parthmannan: From an effort perspective, it shouldn't be too complex and I can make that change. The only difference is that for multiple models, we need to pass a dict of `model_name: model_config` whereas normally `FLOPsMeasurementCallback` expects just the `model_config`

So unless we either
a) Change the usage of `FLOPsMeasurementCallback` everywhere to pass a dict OR
b) Add some complexity/extra args to the code where user can either send a dict or just a config...

I think the choice is whether we prefer higher code complexity vs higher code bloat. I am okay either way. Have a slight preference for code bloat as this is just the first multimodal model. We need more time to see what complexity other multimodal models bring. We can also review again in 3 months if we can unify the code.

malay-nagda: Can we have two sub-keys- `llm` and `mm` to `_model_flops_map` and use these sub-keys in respective classes? This way both callbacks work only for intended models. 

I am open to both as well. But, I'd prefer code bloat too 😅, coz over time as we add stuff, it usually becomes unwieldy.

malay-nagda (approved):

malay-nagda:
LGTM!

malay-nagda:
@ericharper can you help review this?

parthmannan:
@malay-nagda @ashors1 Made another change to address a concern with the combined model map. Split the model map into two and Multimodal map now inherits from the standard map and adds the multimodal models to it.

ashors1 (approved):

ashors1:
> @malay-nagda @ashors1 Made another change to address a concern with the combined model map. Split the model map into two and Multimodal map now inherits from the standard map and adds the multimodal models to it.

Looks good! Thanks!

malay-nagda (approved):

malay-nagda (approved):

PR #11993, opened 2025-01-29
Speechllm develop gen duplex

kevinhu-nv:
# What does this PR do ?

Predict user text at the end of user speech stream instead of in parallel to agent text.

Enable by:
++model.predict_source_text=True

Script:
/home/kevinhu/s2s/s2s_salm_1ab_d1r_12fps_duplex4_data4b_mistral_1a_srctext_front.sh


kevinhu-nv:
@zhehuaichen An old change - can a take a look?

PR #12016, opened 2025-02-01
build: Improve installer

ko3n1g:
# What does this PR do ?

This PR improves the dependency management as well the build-process of our container.

Dependencies are manages inside a newly introduces `manifest.json` file. This is parsed and fed into the build process so that we can benefit from caching.

The docker build process is kept efficient by using build-stages that produce wheels; the final stage consumes and installs the rest of the dependencies. 

# Changelog 

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:
  
**PR Type**:
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information
* Related to # (issue)


reinstall.sh
```diff
@@ -1,107 +1,272 @@
...
+  fi
+    
+}
+
+te() {
+  local mode="$1"
+  cd /opt
+  DIR="TransformerEngine"
+
+  if [ ! -d "$DIR/.git" ]; then
```
thomasdhc, line 88: Why is it necessary to check the existence of DIR?

Dockerfile.ci
```diff
@@ -15,9 +15,64 @@
...
 FROM ${BASE_IMAGE}
 ARG IMAGE_LABEL
+ARG MLM_REPO
+ARG MLM_TAG
+ARG TE_REPO
+ARG TE_TAG
+ARG APEX_REPO
+ARG APEX_TAG
+ARG NEMO_REPO
+ARG NEMO_TAG
```
thomasdhc, line 75: These ARGs aren't consumed at this layer of the container right?

reinstall.sh
```diff
@@ -1,107 +1,272 @@
...
+  else
+      pip install /tmp/wheels/mcore/*.whl
+  fi
+
+  cd /opt
+  DIR="mamba"
+
+  if [ ! -d "$DIR/.git" ]; then
+    rm -rf "$DIR" &&
+      git clone https://github.com/state-spaces/mamba.git
```
thomasdhc, line 50: Is mamba and causal-conv1d specific to mcore?

reinstall.sh
ko3n1g: we clone this repo during `build` and copy it from the build-stage to `final`. Avoid some unnecessary duplicated network calls

reinstall.sh
ko3n1g: Yes, ideally those would be hidden by Mcore install itself. Let's tackle that in a follow-up PR

reinstall.sh
```diff
@@ -1,107 +1,272 @@
...
+    git fetch origin $NEMO_TAG && 
+    git checkout -f $NEMO_TAG
+      
+  ${PIP} install --no-cache-dir virtualenv &&
+    virtualenv /opt/venv &&
+    /opt/venv/bin/pip install --no-cache-dir --no-build-isolation \
+      -r requirements/requirements_vllm.txt \
+      -r requirements/requirements_deploy.txt
+
+  pip install --no-cache-dir . "llama-index==0.10.43" "unstructured==0.14.9" "triton==3.1.0"
```
chtruong814, line 147: Should we pull these out too to manifest.json or just out of scope for now?

reinstall.sh
thomasdhc: Gotcha, I'm not completely caught up on the GHA process, are build and final different stages of CI or different layers of the dockerfile?

ko3n1g: Oh I was just referring to the stages in [Dockerfile.ci](https://github.com/NVIDIA/NeMo/blob/ko3n1g/build/improve-installer/Dockerfile.ci#L90), where we copy over the src-code+wheels from build-stages into the final stage

Dockerfile.ci
thomasdhc: Ok I see now!

reinstall.sh
thomasdhc: Got it, got it!

Dockerfile.ci
```diff
@@ -32,19 +87,26 @@ apt-get clean
...
+    --mount=type=bind,from=apex-wheel,source=/tmp/wheels/apex/,target=/tmp/wheels/apex/ \
+    --mount=type=bind,source=requirements,target=requirements \
+    --mount=type=bind,source=tools/ctc_segmentation/requirements.txt,target=tools/ctc_segmentation/requirements.txt \
+    --mount=type=bind,source=reinstall.sh,target=reinstall.sh \
+    --mount=type=bind,source=setup.py,target=setup.py \
+    --mount=type=bind,source=nemo/package_info.py,target=nemo/package_info.py \
+    --mount=type=bind,source=nemo/__init__.py,target=nemo/__init__.py <<"EOF" bash -ex
+  ls -al /tmp/wheels/
+  ls -al /tmp/wheels/te/
+  bash reinstall.sh --library all --mode install
```
thomasdhc, line 106: To me it seems like we can just install all whls that are available, instead of having a separated logic for checking whether we want to reinstall a package and having logic to check for cloning repos again.

What we want to install or not should be handled in one stage the build stage. The final should simply consume.

thomasdhc: I think that can help simplify some of the logic. In the build stage of the reinstall, build all whl and store under a directory. In the install stage of the reinstall, consume all whls under that directory and install any additional packages we need.

This way we no longer need to pass the ARGs to the final stage and we can skip some of the logic to determine if we need to clone, or build whl.

Let me know what you think!

PR #12076, opened 2025-02-06
Fix dataclass field in some asr examples.

ego:
# What does this PR do ?

Fix dataclass field in some asr examples.
  
**PR Type**:

- [*] Bugfix


nithinraok (approved):
Thanks for the PR

PR #12090, opened 2025-02-07
Fix multi-GPU in-framework deployment

oyilmaz-nvidia:
# What does this PR do ?

Fixes the multi-gpu generation in in-framework deployment.

Fix description: The current code is only using the main process (rank 0) even though multiple ranks were spawned by the lightning. And, since only the rank0 is calling the generate function, the mcore generation code is only called by rank0 and as a result, rank0 keeps waiting for other ranks to call the genrate function. 

With the fix, all the ranks now are calling the generate function.


athitten:
@oyilmaz-nvidia thank you! Could you please add some details on how the issue is being fixed either in the PR desc or in the code itself ? It makes it easier for the review and also anyone else who is looking at this.

athitten:
Tested the PR and it works fine. Based on a limited test case of running locally on 2 GPUs and works in each of the following cases: TP=2, PP=2 and EP=2.

oyilmaz-nvidia:
@athitten Thanks, added some info on how the issue was fixed.

ko3n1g:

athitten (approved):

PR #12094, opened 2025-02-07
Save and Restore ModelOpt state in NeMo 2.0

AAnoosheh:
> [!IMPORTANT]  

# What does this PR do ?

NeMo 1.0 allows [Nvidia Model Optimizer](https://github.com/NVIDIA/TensorRT-Model-Optimizer) state to be saved and restored in checkpoints, and the feature has been ported to 2.0

Note this is the equivalent of these lines in NeMo 1:
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/parts/nlp_overrides.py#L418
https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/parts/nlp_overrides.py#L557

**Collection**: LLM

# Changelog

- Modify save and load functions in `MegatronStrategy` to handle possible ModelOpt states for NeMo 2.0 models.

# Usage

- Save and restore any model transformed by `modelopt` package (i.e. as done in `scripts/llm/ptq.py`)

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -777,6 +805,14 @@ def load_checkpoint(self, checkpoint_path: Union[str, Path], selective_restore:
         """
         torch.cuda.empty_cache()
 
+        if HAVE_MODELOPT:
+            # If present, first restore and modify the model according to the ModelOpt state.
+            restore_sharded_modelopt_state(
+                [unwrap_model(self.megatron_parallel)],
+                ckpt_to_weights_subdir(checkpoint_path, is_saving=False),
+            )
+            logging.info("Restored Model-Optimizer state from checkpoint.")
```
kevalmorabia97, line 814: we should log this message only if there was modelopt state in the checkpoint i.e. (`if ModeloptStateManager.is_converted(core_model)`)

nemo/lightning/pytorch/strategies/megatron_strategy.py
kevalmorabia97, line 1: Does the intermediate checkpoints saved by trainer.fit during training also use this file for saving? For QAT resume, we need the modelopt state being stored in the per-iter checkpoints as well. Previously in NeMo1, they used different functions for that hence we needed 2 places each for saving and restoring modelopt state

AAnoosheh: Yes Nemo1 had distinct `.nemo` files and lightning checkpoints, but they're unified in Nemo2

nemo/collections/llm/quantization/utils.py
```diff
@@ -73,6 +73,7 @@ def load_with_modelopt_layer_spec(
     nemo_checkpoint_path: str,
     tensor_model_parallel_size: int = 1,
     pipeline_model_parallel_size: int = 1,
+    num_nodes: int = 1,
```
kevalmorabia97, line 76: I think we also need a separate parameter for devices as I think it refers to num gpus per node

scripts/llm/ptq.py
```diff
@@ -68,8 +75,10 @@ def get_args():
     parser.set_defaults(generate_sample=False)
 
     args = parser.parse_args()
+
     if args.export_path is None:
         args.export_path = f"./qnemo_{args.algorithm}_tp{args.inference_tp}_pp{args.inference_pp}"
```
kevalmorabia97, line 80: nemo or qnemo depending on `args.save_as_torch`

scripts/llm/ptq.py
```diff
@@ -30,7 +33,11 @@ def get_args():
     parser.add_argument("-cpp", "--calibration_pp", "--calib_pp", type=int, default=1)
     parser.add_argument("-itp", "--inference_tp", "--tensor_parallelism_size", type=int, default=1)
     parser.add_argument("-ipp", "--inference_pp", "--pipeline_parallelism_size", type=int, default=1)
+    parser.add_argument("-nodes", "--num_nodes", type=int, default=1)
     parser.add_argument('-out', '--export_path', '--output_path', type=str, help='Path for the exported engine')
+    parser.add_argument(
+        "--save_as_torch", action="store_true", help="Save `-out` as a Torch checkpoint instead of TRT-LLM"
```
kevalmorabia97, line 39: Thoughts on changing to `--export_format` which is default as `trtllm` but can also be `nemo` or in future `hf`

scripts/llm/ptq.py
```diff
@@ -97,16 +105,27 @@ def main():
...
+    if args.save_as_torch:
+        # Standard NeMo 2.0 checkpoint format
+        trainer.save_checkpoint(args.export_path)
+        if is_global_rank_zero():
+            TrainerContext.from_trainer(trainer).io_dump(
+                ckpt_to_context_subdir(args.export_path), yaml_attrs=["model"]
+            )
+    else:
+        # TRT-LLM
+        quantizer.export(model, args.nemo_checkpoint)
```
kevalmorabia97, line 128: Can you update one of the CI tests to save nemo instead of default trtllm and also check `modelopt_state` folder is present?

nemo/collections/llm/quantization/utils.py
```diff
@@ -73,6 +73,8 @@ def load_with_modelopt_layer_spec(
     nemo_checkpoint_path: str,
     tensor_model_parallel_size: int = 1,
     pipeline_model_parallel_size: int = 1,
+    devices_per_node: int = 1,
```
kevalmorabia97, line 76: nit: lets name this `devices` as that's the standard name being used everywhere in nemo

scripts/llm/ptq.py
```diff
@@ -68,8 +77,16 @@ def get_args():
...
+
     if args.export_path is None:
-        args.export_path = f"./qnemo_{args.algorithm}_tp{args.inference_tp}_pp{args.inference_pp}"
+        if args.export_format == "trtllm":
+            args.export_path = f"./qnemo_{args.algorithm}_tp{args.inference_tp}_pp{args.inference_pp}"
+        else:
+            args.export_path = f"./nemo_{args.algorithm}"
+
+    if args.num_devices is None:
+        args.num_devices = "auto"  # PL Trainer can auto-detect number of devices
```
kevalmorabia97, line 88: how about we just set this as `default` in the argparse?

scripts/llm/ptq.py
```diff
@@ -30,7 +34,12 @@ def get_args():
     parser.add_argument("-cpp", "--calibration_pp", "--calib_pp", type=int, default=1)
     parser.add_argument("-itp", "--inference_tp", "--tensor_parallelism_size", type=int, default=1)
     parser.add_argument("-ipp", "--inference_pp", "--pipeline_parallelism_size", type=int, default=1)
+    parser.add_argument("-devices", "--num_devices", type=int, default=None, help="Number of devices per node")
```
kevalmorabia97, line 37:
```suggestion
    parser.add_argument("--devices", type=int, default="auto", help="Number of GPUs to use per node")
```

.github/workflows/cicd-main.yml
```diff
@@ -5062,6 +5062,21 @@ jobs:
         rm -rf /tmp/nemo2_ckpt
         rm -rf /tmp/nemo2_ptq_engine
 
+  L2_NeMo_2_PTQ_Llama2_FP8_nemo:
+    needs: [pre-flight, cicd-test-container-build]
+    uses: ./.github/workflows/_test_template.yml
+    if: contains(fromJSON(needs.pre-flight.outputs.test_to_run), 'L2_NeMo_2_PTQ_Llama2_FP8_nemo')
+    with:
+      RUNNER: self-hosted-azure
+      SCRIPT: |
```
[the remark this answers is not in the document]
AAnoosheh: I believe the quantization is simulated precision so it's fine, hence the test above it (which I copied and modified from) does same.

(Running an exported TRTLLM engine will be the real FP8)

ko3n1g (dismissed):

scripts/llm/ptq.py
```diff
@@ -28,9 +32,28 @@ def get_args():
...
+        default=1,
+        help="TRT-LLM engine TP size. (Only used when `--export_format` is 'ttllm')",
+    )
+    parser.add_argument(
+        "-ipp",
+        "--inference_pp",
+        "--pipeline_parallelism_size",
+        type=int,
+        default=1,
+        help="TRT-LLM engine PP size. (Only used when `--export_format` is 'ttllm')",
```
kevalmorabia97, line 49:
```suggestion
        help="TRT-LLM engine TP size. (Only used when `--export_format` is 'trtllm')",
    )
    parser.add_argument(
        "-ipp",
        "--inference_pp",
        "--pipeline_parallelism_size",
        type=int,
        default=1,
        help="TRT-LLM engine PP size. (Only used when `--export_format` is 'trtllm')",
```

scripts/llm/ptq.py
```diff
@@ -28,9 +32,28 @@ def get_args():
...
+    )
+    parser.add_argument("--devices", type=int, default="auto", help="Number of GPUs to use per node")
```
kevalmorabia97, line 51: Actually when I tested it, argparse gives an error if we don't pass value and it uses default `auto` as its a string and not int. Perhaps we just make it default to `1` and let people configure it?

kevalmorabia97 (dismissed):
LGTM

ko3n1g:

scripts/llm/ptq.py
Laplasjan107: Now it will be needed to pass all 4 flags. Maybe we could preserve the current default behaviour? (like setting the value to -1 by default and overriding it if not set)

AAnoosheh: @kevalmorabia97 what do you think about leaving it as None again as before and manually setting to "auto" after?

kevalmorabia97: Sounds good

scripts/llm/ptq.py
```diff
@@ -97,16 +127,29 @@ def main():
...
+        tensor_model_parallel_size=args.calibration_tp,
+        pipeline_model_parallel_size=args.calibration_pp,
+        devices=args.devices,
+        num_nodes=args.num_nodes,
         ckpt_load_strictness=args.ckpt_load_strictness,
     )
     model = quantizer.quantize(model)
-    quantizer.export(model, args.nemo_checkpoint)
+
+    if args.export_format == "nemo":
```
janekl, line 142: Could you move extending the export logic to `export(...)` (and adapt the call to `quantizer.export(model, nemo_checkpoint, trainer)`?

To this end I think we can add `args.export_format` as another field `ExportConfig` dataclass (defined [here](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/quantization/quantizer.py#L80-L92)).

Some benefits:
* keeping ptq.py script still very simple and high-level 
* having export logic in one place and easily reuse in, for example, [api.py](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/api.py#L406)

scripts/llm/ptq.py
janekl: How about having `devices` and `num_nodes` default to `None`, and in this case overridden to `calibration_tp` and `calibration_pp`, respectively? This would keep the current behavior w/o the need to specify explicitly all 4 params.

nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -777,6 +788,23 @@ def save_checkpoint(
...
+                ckpt_io = self.checkpoint_io
+                if isinstance(ckpt_io, _WrappingCheckpointIO):
+                    ckpt_io = ckpt_io.checkpoint_io
+                save_sharded_modelopt_state(
+                    [core_model],
+                    ckpt_to_weights_subdir(filepath, is_saving=True),
+                    sharded_strategy=ckpt_io.save_sharded_strategy,
+                )
+                logging.info("Saved Model-Optimizer state into checkpoint.")
+
```
janekl, line 807: FYI @mikolajblaz @ananthsub @marcromeyn there are updates to checkpoint saving logic. Let us know if you have any remarks

kevalmorabia97 (dismissed):

scripts/llm/ptq.py
AAnoosheh: IMO, I don't like it when code tries to assume things behind my back to "simplify" my user experience, making future debuggings worse. Also less logic the better – for refactoring and error-proneness.

nemo/lightning/pytorch/strategies/megatron_strategy.py
AAnoosheh: I will assume they're all good at this point and run CICD now

scripts/llm/ptq.py
janekl: Fair point. The intention here is that the script is used at least in docs https://docs.nvidia.com/nemo-framework/user-guide/latest/model-optimization/quantization/quantization.html#use-the-ptq-script-with-torchrun-or-slurm and our internal CI system (outside Github) so the changes need to be propagated accordingly. Hence we suggested to keep the current behavior. But OK we can handle this

janekl (dismissed):

ko3n1g (dismissed):

.github/workflows/cicd-main.yml
```diff
@@ -2792,6 +2792,21 @@ jobs:
...
+        python tests/collections/llm/test_hf_import.py --hf_model /home/TestData/nlp/megatron_llama/llama-ci-hf --output_path /tmp/nemo2_ckpt
+
+        python scripts/llm/ptq.py -nc /tmp/nemo2_ckpt -algo fp8 -out /tmp/nemo2_ptq_ckpt --export_format nemo
+
+      AFTER_SCRIPT: |
+        rm -rf /tmp/nemo2_ckpt
```
pablo-garay, line 2807: these are strictly not necessary to remove

pablo-garay (approved):

kevalmorabia97 (approved):
Re-approving to unblock

ko3n1g:
After the last failure, we'll need to merge it manually. I spoke with @AAnoosheh and will take care of it once he reports the successful pass of previously failing tests

ko3n1g:
CI passing: https://github.com/NVIDIA/NeMo/actions/runs/13528653924/job/37818530901

PR #12242, opened 2025-02-18
ONNX exporter 

oyilmaz-nvidia:
# What does this PR do ?

Adding ONNX exporter for LLM models. This class will be also used for embedding models.


nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+        self.model_input_names = None
+        self.model_output_names = None
+        self.onnx_runtime_session = None
+        self.calibration_data = None
+        self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        self.quant_max_batch_size = None
+
+        if self.model_name_or_path is not None:
+            if model is not None:
+                warnings.warn("A model was also passed but it will be overridden.")
```
fy-nv, line 115: Should we raise an error instead to remove ambiguity?

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+                else:
+                    self._load_hf_model()
+
+        if load_runtime:
+            self._load_runtime()
+
+    def _load_runtime(self):
+        if use_onnxruntime:
+            if Path(self.onnx_model_path).exists():
+                self.onnx_runtime_session = onnxruntime.InferenceSession(self.onnx_model_path)
```
fy-nv, line 129: We discovered that this is expected to fail if the onnx model is the product of ptq. If that's still the case, I think catching that early and throwing a more informative error would be helpful.

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+            onnx_model_dir="/path/for/onnx_model/files",
+            model_name_or_path="/path/for/model/files",
+        )
+
+        onnx_llm_exporter.export(
+            input_names=["input_ids", "attention_mask", "dimensions"],
+            output_names=["embeddings"],
+        )
+
+        output = trt_llm_exporter.forward(["Hi, how are you?", "I am good, thanks, how about you?"])
```
fy-nv, line 78: `trt_llm_exporter` is not defined.

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+        """
+        Initializes the ONNX Exporter.
+
+        Args:
+            onnx_model_dir (str): path for storing the ONNX model files.
+            model (Optional[torch.nn.Module]): torch model.
+            tokenizer (HF or NeMo tokenizer): tokenizer class.
+            model_name_or_path (str): a path for ckpt or HF model ID
+            load_runtime (bool): load ONNX runtime if there is any exported model available in
+                                 the onnx_model_dir folder.
```
fy-nv, line 99: nit: It's a bit confusing that an exporter can also load runtime. I suggest that we remove this option, instead create a new ONNX runtime class, and have something like 

```
runtime = onnx_llm_exporter.get_runtime(...)
output = runtime.forward(["Hi, how are you?", "I am good, thanks, how about you?"])
```

nemo/collections/llm/gpt/model/hf_llama_embedding.py
```diff
@@ -188,3 +191,138 @@ def forward(
...
+    model = AutoModel.from_pretrained(
+        model_name_or_path, torch_dtype=torch_dtype, trust_remote_code=trust_remote_code
+    ).eval()
+
+    # configure pooling
+    pooling_module = Pooling(pooling_mode=pooling_mode)
+
+    # NV-Embed-v1 model has seperate embedding model and a built-in pooling module
+    if (
+        model.__class__.__name__ == "NVEmbedModel"
```
fy-nv, line 320: Should we add a safeguard here and throw an error if `model.__class__.__name__ != "NVEmbedModel"`?

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+use_onnxruntime = True
+try:
+    import onnxruntime
+except Exception:
+    logging.warning("onnxruntime is not available.")
+    use_onnxruntime = False
+
+
+# pylint: disable=line-too-long
+class OnnxLLMExporter(ITritonDeployable):
```
fy-nv, line 61: The name is confusing since we are exporting to ONNX as well as TRT.

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+                self.model_output_names = [output.name for output in self.onnx_runtime_session.get_outputs()]
+                self.tokenizer = AutoTokenizer.from_pretrained(
+                    Path(self.onnx_model_dir) / "tokenizer", trust_remote_code=True
+                )
+
+    def _load_hf_model(self):
+        self.model = AutoModel.from_pretrained(
+            self.model_name_or_path,
+            trust_remote_code=True,
+        ).eval()
```
fy-nv, line 140: Can we add a safeguard here and check that the model is one of the supported models (e.g. `LlamaBidirectionalHFAdapter`)?

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,512 @@
...
+            input_names=input_names,
+            example_inputs=example_inputs,
+            output_names=output_names,
+            opset=opset,
+            dynamic_axes_input=dynamic_axes_input,
+            dynamic_axes_output=dynamic_axes_output,
+            export_dtype=export_dtype,
+            verbose=verbose,
+        )
+        self._load_runtime()
```
fy-nv, line 179: nit: see comment above regarding `load_runtime`

fy-nv (changes requested):

nemo/export/onnx_llm_exporter.py
oyilmaz-nvidia: This is a pattern we use also for other exporter classes. If the engine files are available in the model folder, than it's loaded automatically and without exporting, users can run inference with it.

nemo/export/onnx_llm_exporter.py
oyilmaz-nvidia: We are not directly exporting it TRT actually. Converting ONNX to TRT. So, everything is about ONNX in this class. If we have a direct TRT exporter, then we'd name it as TRTLLMExporter.

nemo/collections/llm/gpt/model/hf_llama_embedding.py
oyilmaz-nvidia: Added assertion.

nemo/export/onnx_llm_exporter.py
oyilmaz-nvidia: This is for the base model actually. Not sure what kind of check we can add. User can directly point to a local path or and model ID on HF.

fy-nv: Is ONNXLLMExporter intended to be compatible with `LlamaBidirectionalHFAdapter` only or more than that?

In the notebook, we have the following. The `model` here is of class `LlamaBidirectionalHFAdapter`. 

```
# Adapt the model first
model, tokenizer = get_llama_bidirectional_hf_model(
    model_name_or_path=hf_model_path,
    normalize=normalize,
    pooling_mode=pooling_mode,
    trust_remote_code=True,
)
...

onnx_exporter = OnnxLLMExporter(
    onnx_model_dir=onnx_export_path, 
    model=model,
    tokenizer=tokenizer,
)

```

nemo/collections/llm/gpt/model/hf_llama_embedding.py
oyilmaz-nvidia: But, this should only work if the model class is NVEmbedModel, right? I am checking the nvembed v2, and the class name is LlamaBidirectionalModel. Looks like we don't need to run this code for v2, no?

nemo/export/onnx_llm_exporter.py
oyilmaz-nvidia: The intention is to have compatibility with other models as well.

fy-nv: My concern is that a user may download a Llama model, use this exporter, and finds out that it doesn't work. In `LlamaBidirectionalHFAdapter`, we did several things that are not necessarily standard across all LLMs

1. `model.__class__.__name__ == "NVEmbedModel"`
2. Pooling logic
3. Clipped dimensions

fy-nv: If you intend to keep `OnnxLLMExporter` generic, then I'd suggest creating a subclass with the correct defaults for `LlamaBidirectionalHFAdapter`. For example, we should predefine `dynamic_axes_input` and other parameters for the `export` [method](https://github.com/NVIDIA/NeMo/blob/42416c7bef504f7c1327dfca4348d63613003e25/nemo/export/onnx_llm_exporter.py#L144) for the   `LlamaBidirectionalHFAdapter` model

oyilmaz-nvidia: So, the `OnnxLLMExporter` is a generic class and the params in the export method is generic enough IMO. It will work for most of the LLMs if it supports ONNX. `LlamaBidirectionalHFAdapter` is specific to the class that Ao created and having specific functions in the `LlamaBidirectionalHFAdapter` should be Ok.

nemo/export/onnx_llm_exporter.py
```diff
@@ -0,0 +1,453 @@
...
+try:
+    from pytriton.decorators import batch
+except Exception:
+    logging.warning("PyTriton is not available.")
+    use_pytriton = False
+
+
+use_onnxruntime = True
+try:
+    import onnxruntime
```
ko3n1g, line 52: Where do we install this?

ko3n1g: Should we add it as optional dependency to the CI so that we test the paths?

janekl (approved):

nemo/export/onnx_llm_exporter.py
oyilmaz-nvidia: Will add the tests and optional dependencies right after this PR.

PR #12244, opened 2025-02-18
fix speechlm import ckpt on slurm

stevehuang52:
Fix speechlm import ckpt on slurm multi-node setting, by adding the option to skip `importer.on_import_ckpt`.

### DIFF
--- a/nemo/lightning/io/api.py
+++ b/nemo/lightning/io/api.py
@@ -109,7 +109,11 @@ def model_exporter(target: Type[ConnectorMixin], ext: str) -> Callable[[Type[Con
 
 
 def import_ckpt(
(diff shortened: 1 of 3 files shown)


nemo/lightning/io/api.py
```diff
@@ -164,7 +171,8 @@ def import_ckpt(
 
     importer: ModelConnector = model.importer(source)
     ckpt_path = importer(overwrite=overwrite, output_path=output_path)
-    importer.on_import_ckpt(model)
+    if import_tokenizer:
+        importer.on_import_ckpt(model)
```
yaoyu-33, line 175: can you make import_tokenizer a arg to `on_import_ckpt` as well? `on_import_ckpt` is supposed to be a general hook, you can change inside `on_import_ckpt` instead of here.

yaoyu-33 (approved):

PR #12245, opened 2025-02-18
Energon ckpt multimodal

abhinavg4:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/multimodal/data/energon/base.py
```diff
@@ -101,6 +101,17 @@ def __init__(
         **kwargs: Additional keyword arguments. Will be passed to get_train_dataset() of Energon
         """
 
+        if not parallel_state.is_initialized():
+            logging.info(
+                f"Muiltimodal data loader parallel state is not initialized,"
+                f"Please ensure no data loader is being created on pipeline parallel ranks >0"
+            )
+        else:
+            pp_rank = parallel_state.get_pipeline_model_parallel_rank()
```
yaoyu-33, line 110: what about other parallel like TP and EP?

requirements/requirements_multimodal.txt
```diff
@@ -4,9 +4,9 @@ clip
 decord; sys_platform == 'linux'
 diffusers>=0.19.3
 einops_exts
+git+https://github.com/NVIDIA/Megatron-Energon.git
```
yaoyu-33, line 7: don't think use main is a good idea, at least pin to one of commit?

abhinavg4: So I think energon folks will release a new version soonish (Within next 24 hrs) and then I will update this to that version. I agree we should not use main.

nemo/collections/multimodal/data/energon/base.py
```diff
@@ -294,10 +295,18 @@ def state_dict(self) -> Dict[str, Any]:
 
         if self.trainer:
             dataloader_obj = self.trainer.train_dataloader
-            state = dataloader_obj.save_state_global(dst_rank=0)
+
+            state = []
+            if torch.distributed.get_rank() == parallel_state.get_model_parallel_src_rank():
```
yashaswikarnati, line 300: I would write a comment explaining the logic here for future purposes. Why we call get_model_parallel_src_rank and what energon is expected to return

yashaswikarnati, line 300: if lets say tp =2, dp =2, pp =2. Lets say 0, 1, 2, 3 global ranks has MP group 0 and DP group 0. 4, 5, 6, 7 has MP group 1 and DP group . I'm assuming above condition with  get_model_parallel_src_rank() is True for rank 0 and 4. Is my understanding accurate? As they are different DP ranks I assume them to have different state but what is save_state_global doing? is it not doing gather from DP ranks?

abhinavg4: I agree, I will add that

abhinavg4: Yes you are right. Your understanding is correct. save_state_global saves the states from all the ranks.

yashaswikarnati: https://github.com/NVIDIA/Megatron-LM/blob/e1586c218827fd686a7f88cb389fa0aff2fdb57e/megatron/core/dist_checkpointing/strategies/common.py#L40

yashaswikarnati: nit; if rank 0 and rank 4 in the above example has same state, based on how megatron is saving common.pt it would be enough to return state from rank 0 I think.

abhinavg4: Yes that's exacty what is ahppening but we need to make this call on every rank that has a unique state. 

`save_state_global` will return all the states to rank 0 and return None to all other ranks

yashaswikarnati: Ah ok. thanks for the clarification. Thats a bit counter intuitive to call on multiple ranks but only one rank returns the full state. Please add a comment with this information.

PR #12246, opened 2025-02-18
Fixed normalization of feature vector and weight vector

penguinwang96825:
# What does this PR do ?

Fixes the incorrect implementation of the **nemo.collections.asr.modules.conv_asr.SpeakerDecoder** by ensuring that feature vector and weight normalization is applied correctly. Previously, normalization was applied inside the loop on individual parameters, which led to incorrect cosine similarity values. Now, normalization is applied directly to the feature vector and weight matrix to maintain correct scaling.  

**Collection**: ASR

# Changelog

- Fixed incorrect normalization of weights inside the loop.  
- Applied normalization to the feature vector and weight matrix before computing cosine similarity.  
- Ensured cosine similarity values do not exceed 1. 

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas. @titu1994, @redoctopus, @jbalam-nv, or @okuchaiev


penguinwang96825:
We’ve actually tried this on our in-house framework with the same loss function and noticed that the gradients were erratic. Upon investigation, we found that the logits being passed to the loss function weren’t constrained between -1 and 1, which hinted at a missing or incorrect normalisation step. That’s what prompted us to dig into the code and realise that the original normalisation implementation was applied incorrectly.

Hence this PR: it ensures that both the feature vector and the weight matrix are normalised properly before the cosine similarity (or angular) loss calculation, so that the logits always remain within the valid [−1,1] range.

Let me know if you have any other questions or need more details!

nithinraok:
Thank you for the explanation. To merge this I will need to verify couple of things:
* Performance with published model
* If it doesn;t affect with published model (unlikely), I will need to retrain and push a newer version. Will keep this PR open untill then.

nithinraok:
Could you signoff the commit following this steps please: https://github.com/NVIDIA/NeMo/pull/12246/checks?check_run_id=37698120855

penguinwang96825:
@nithinraok Thanks for the update and for verifying the performance concerns. From our side, unfortunately, we didn't run on the same TitaNet model, but after fixing the incorrect implementation other models (e.g. ECAPA-TDNN, Gemini-Df-ResNet, and Conformer) worked. However, please take all the time you need to retrain and confirm on your end, I appreciate you keeping this PR open until then!

nithinraok (approved):
Apologies for the delay. 

Able to test this for both inference and training.

nithinraok:
@penguinwang96825 to merge this PR:
1. Could you sign off the PR: https://github.com/NVIDIA/NeMo/pull/12246/checks?check_run_id=42728111895

nithinraok (approved):

nithinraok:
Update on this PR:

Since we’re modifying the weights, the output of the pretrained model no longer matches the previous labels for the same input, even though embs are changed minimally (which I tested before). 

Example failure:
`
FAILED tests/collections/speaker_tasks/test_speaker_label_models.py::TestEncDecSpeechLabelModel::test_pretrained_titanet_embeddings - AssertionError: assert 11144 == tensor([8796], device='cuda:0')
`

To merge this, we’d need to fine-tune or retrain the Titanet model and publish a new version.

We’ll revisit this based on interest. Thanks again for the PR, @penguinwang96825!

PR #12289, opened 2025-02-20
Call default factory in dataclasses when saving yaml via nemo.lightning.io

hemildesai:


cuichenx (approved):
thank you, verified working

PR #12327, opened 2025-02-22
Add nemo.tron.pretrain

hemildesai:


nemo/tron/train.py
```diff
@@ -408,6 +408,7 @@ def train(
 
     # If any exit conditions (signal handler, duration, iterations) have been reached, exit.
     if should_exit:
+        maybe_finalize_async_save(ckpt_cfg=config.checkpoint_config, blocking=True, terminate=True)
```
maanug-nv, line 411: why is this necessary? it's called just a couple lines above?

nemo/tron/checkpointing.py
```diff
@@ -637,9 +637,9 @@ def train_state_finalize_fn():
                     )
 
         else:
+            train_state_dict = train_state.state_dict()
```
maanug-nv, line 640: just wondering why this move was needed?

hemildesai: This is needed for async saves. As the train_state object changes by the time you reach the finalize function, we need to preserve it for the particular checkpoint

maanug-nv (approved):

nemo/tron/train.py
hemildesai: Oh sorry, missed the above call.

maanug-nv: i just merged but feel free to change on `mlm-pretrain-loop` directly

PR #12421, opened 2025-02-28
Add more flexibility to peft

stevehuang52:
Current PEFT in nemo only saves parameters that require gradients. However, there are other params that don't require gradients but still need to be saved, such as the running mean/var of batch norm. Therefore, this PR separates setting `self.trainable_params` to a new member function `set_trainable_params()`, so that it can be customized for different use cases.

Another change is putting `partial(WrappedAdapterIO, peft=self)` into another member function `get_wrappped_io`, so that it can be customized by other use cases in SpeechLM.

The PR doesn't change the default behavior of the original code.


cuichenx (commented):
Looks good, but you need to remove
```
self.trainable_params = set(
            name for name, param in trainer.lightning_module.named_parameters() if param.requires_grad
        )
``` in line 213 right?

cuichenx (approved):
LGTM, thanks!

PR #12445, opened 2025-03-03
remove PytorchOptimizerModule connect

lifeiteng:
> [!IMPORTANT]  

# What does this PR do ?

Fix a bug that make lr be consistent value, will not be updated with training step

**Collection**: llm hf_auto_model_for_causal_lm.py

```
optim=pytorch_adam_with_cosine_annealing(max_lr=3e-4),
```


akoumpa:
Hi @lifeiteng , thanks for bringing this up. 

I was thinking users could change the optimizer with the `llm.api.finetune` and specify the optimizer of interest. I wonder if the change could be placed there?

lifeiteng:
It seems that has been fixed in newest main branch.

PR #12453, opened 2025-03-03
Make causal conv import not fail if you are not using the hyena model.

jstjohn:
**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


ko3n1g (approved):
That's a clever way of handling it! Our SOP handles it a bit differently, but I don't mind it: https://docs.nvidia.com/nemo-framework/user-guide/latest/best-practices.html#import-guarding

ko3n1g:
fast-tracking this since CI wouldn't test this path

PR #12457, opened 2025-03-03
[ModelOpt] Shared code in `nemo.collections.llm.modelopt` for ptq, prune, distill + QAT fixes

kevalmorabia97:
> [!IMPORTANT]  

- [x] Refactor ptq, distillation and pruning code in `nemo.collections.llm.modelopt`
- [x] Add `nvfp4` as a supported choice for PTQ and allow user to configure uneven pipeline parallel layers in first and last stage
- [x] Automatically use modelopt gpt spec if nemo checkpoint has `weights/modelopt_state` folder (enables QAT with distillation or generation on quantized ckpt)
- [x] Move pruning example core logic into llm.prune API and reuse in the example
- [x] Reuse API functions in example scripts instead of duplicating code

**Collection**: [llm]

# Changelog

- Refactor shared ptq and pruning code in `nemo.collections.llm.modelopt`

# Usage

- N/A

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/lightning/megatron_parallel.py
```diff
@@ -587,7 +587,7 @@ def init_model_parallel(self):
         from megatron.core.tensor_parallel.layers import set_defaults_if_not_set_tensor_model_parallel_attributes
 
         for model_module in self:
-            if not self._cpu and (not HAVE_CUSTOM_FSDP or not self.ddp_config.use_custom_fsdp):
+            if not self._cpu and not (HAVE_CUSTOM_FSDP and self.ddp_config and self.ddp_config.use_custom_fsdp):
```
kevalmorabia97, line 590: I had to add this locally where I am using NeMo from my branch and Mcore ToT and hitting `self.ddp_config` as None. But in CI this line was not hit (either maybe `HAVE_CUSTOM_FSDP=False`).

Regardless I think its better to check if `self.ddp_config` is not None before checking its attributed hence added the change

nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -849,12 +844,14 @@ def load_checkpoint(self, checkpoint_path: Union[str, Path], selective_restore:
...
         if HAVE_MODELOPT:
             # If present, first restore and modify the model according to the ModelOpt state.
+            # Avoid quantizers being added to teacher model if model is a distillation model.
             core_model = unwrap_model(self.megatron_parallel)
-            restore_sharded_modelopt_state(
-                [core_model],
-                ckpt_to_weights_subdir(checkpoint_path, is_saving=False),
-            )
-            if ModeloptStateManager.is_converted(core_model):
+            with core_model.hide_teacher_model() if hasattr(core_model, "hide_teacher_model") else nullcontext():
```
kevalmorabia97, line 849: This change enables us to load quantized student for distillation (QAT) otherwise teacher will also have quantizers added but no state dict for it.

This happens because quantization config has regex `*output_layer*` causing `_teacher_model.*output_layer*` to also match

scripts/llm/ptq.py
```diff
@@ -14,19 +14,19 @@
...
     """Parses PTQ arguments."""
-
     parser = argparse.ArgumentParser(
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
-        description="NeMo PTQ argument parser",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter, description="NeMo PTQ argument parser"
     )
     parser.add_argument("-nc", "--nemo_checkpoint", type=str, help="Source NeMo 2.0 checkpoint")
+    parser.add_argument(
+        "--tokenizer", type=str, help="Tokenizer to use. If not provided, model tokenizer will be used"
```
kevalmorabia97, line 28: Added the ability to override tokenizer in the model or support a case where nemo ckpt doesn't have a tokenizer by default. 
This is consistent with pruning and distillation.

Main use case is for llama ci model where model's vocab size is 50k but doesn't have tokenizer hence we use default nmt_tokenizer with 32k vocab size and we need to overwrite the tokenizer to gpt2 tokenizer.

I think the issue doesn't occur in ptq because we don't use tokenized CI dataset instead take HF text dataset (cnn_dailymail)

nemo/lightning/megatron_parallel.py
kevalmorabia97, line 590: I had to add this locally where I am using NeMo from my branch and Mcore ToT and hitting `self.ddp_config.use_custom_fsdp` but since for pruning we don't use DDP, the config is None. But in CI this line was not hit (maybe `HAVE_CUSTOM_FSDP=False`).

Regardless I think its better to check `if self.ddp_config is not None` before checking its attribute hence added the change

nemo/collections/llm/modelopt/prune/pruner.py
```diff
@@ -0,0 +1,137 @@
...
+    "hidden_size",
+    "num_attention_heads",
+    "num_query_groups",
+    # Depth pruning
+    "num_layers",
+}
+
+
+@dataclass
+class PruningConfig:
```
kevalmorabia97, line 43: new pruning config for better passing of arguments between functions

nemo/collections/llm/modelopt/prune/pruner.py
kevalmorabia97, line 1: functions moved here from `gpt_prune.py` example so they can be used in `llm.prune` API

nemo/collections/llm/modelopt/model_utils.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def _set_gpt_modelopt_spec(model_cfg: llm.GPTConfig) -> llm.GPTConfig:
+    """Set model.config.transformer_layer_spec to modelopt spec."""
+    logging.info("Setting model.config.transformer_layer_spec to gpt_modelopt_spec")
+    assert isinstance(model_cfg, llm.GPTConfig), "model_cfg must be a GPTConfig"
+    try:
+        from functools import partial
+
+        from megatron.core.post_training.modelopt.gpt.model_specs import get_gpt_modelopt_spec
```
kevalmorabia97, line 38: This is the new modelopt spec function that also supports DeepSeek and added to Mcore ToT hence added the support (verified locally with this spec on deepseek)

nemo/collections/llm/modelopt/model_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    except ImportError:
+        # Older spec: Will be deprecated, doesnt support DeepSeek
+        from megatron.core.inference.modelopt_support.gpt.model_specs import get_gpt_layer_modelopt_spec
+
+        modelopt_spec = get_gpt_layer_modelopt_spec(num_experts=model_cfg.num_moe_experts, remap_te_layernorm=True)
+    model_cfg.transformer_layer_spec = modelopt_spec
+    return model_cfg
+
+
+def set_modelopt_spec_if_exists_in_ckpt(model: L.LightningModule, path: str) -> bool:
```
kevalmorabia97, line 50: new utility function to allow automatically using modelopt spec if restoring a ckpt that has `modelopt_state` in it (i.e. quantized ckpt)

nemo/collections/llm/modelopt/model_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        if isinstance(model, llm.GPTModel):
+            _set_gpt_modelopt_spec(model.config)
+        else:
+            logging.warning(f"{type(model)} is not a GPTModel. Modelopt state will not be loaded.")
+            return False
+        return True
+    return False
+
+
+def setup_trainer_and_restore_model_with_modelopt_spec(
```
kevalmorabia97, line 63: Unified logic from ptq and pruning model loading part into this function

nemo/collections/llm/modelopt/distill/utils.py
```diff
@@ -114,7 +113,9 @@ def adjust_distillation_model_for_mcore(
     """Extra modifications to ``mtd.DistillationModel`` required for Megatron-Core."""
     # Get rid of ModelOpt Distillation state
     # NOTE: If re-placed, above losses need modifcation as `TransformerConfig` has non-pickleable elements.
-    mto.ModeloptStateManager(model)._state.pop()
+    existing_state = mto.ModeloptStateManager(model).state_dict()
+    assert len(existing_state) == 1 and existing_state[0][0] == "kd_loss", f"{existing_state=}"
+    mto.ModeloptStateManager.transfer_state_dict(model, torch.nn.Module())
```
kevalmorabia97, line 118: This is a small bug fix: Previously modelopt state for distillation was not removed but just set to empty hence we cannot again load ptq ckpt for student.

This fix now allows us to load student as a quantized ckpt hence allow QAT with distillation

nemo/collections/llm/inference/base.py
```diff
@@ -130,6 +131,12 @@ def _setup_trainer_and_restore_model(
     """
     assert isinstance(trainer.strategy, MegatronStrategy), "Only MegatronStrategy is supported for trainer.strategy."
     assert trainer.strategy.context_parallel_size <= 1, "Context parallelism is not supported for inference."
+
+    # [ModelOpt]: If modelopt_state exists, overwrite transformer_layer_spec to modelopt spec
+    from nemo.collections.llm.modelopt import set_modelopt_spec_if_exists_in_ckpt
+
+    set_modelopt_spec_if_exists_in_ckpt(model, path)
```
kevalmorabia97, line 138: This allows us to automatically use modelopt spec if restoring a ckpt that has modelopt_state in it (example use case - `generate.py` on a ptq nemo ckpt)

nemo/collections/llm/api.py
```diff
@@ -316,11 +410,18 @@ def distill(
...
+    _teacher_model = io.load_context(ckpt_to_context_subdir(teacher_model_path), subpath="model")
     assert isinstance(_student_model, GPTModel), "Only models based on `llm.GPTModel` are supported currently."
     assert isinstance(_teacher_model, GPTModel), "Only models based on `llm.GPTModel` are supported currently."
 
+    is_quantized_student = set_modelopt_spec_if_exists_in_ckpt(_student_model, student_model_path)
+    set_modelopt_spec_if_exists_in_ckpt(_teacher_model, teacher_model_path)
+
+    # Need to disable gradient accumulation fusion for QAT with distillation
+    if is_quantized_student:
+        _student_model.config.gradient_accumulation_fusion = False
```
kevalmorabia97, line 423: A small fix required to allow QAT with distillation. Otherwise getting this error on backward:

```python
[rank0]:   File "/opt/megatron-lm/megatron/core/tensor_parallel/layers.py", line 510, in backward
[rank0]:     if weight.main_grad.dtype == torch.float32:
[rank0]:        ^^^^^^^^^^^^^^^^
[rank0]: AttributeError: 'Tensor' object has no attribute 'main_grad'
```

nemo/collections/llm/api.py
```diff
@@ -267,6 +282,85 @@ def validate(
     return app_state.exp_dir
 
 
+@run.cli.entrypoint(name="prune", namespace="llm")
+def prune(
```
kevalmorabia97, line 286: Moved main logic from `gpt_prune.py` example to this new `llm.prune` API and directly use the API in the example to avoid code duplication

nemo/collections/llm/api.py
```diff
@@ -149,6 +159,11 @@ def pretrain(
         PosixPath('/path/to/log_dir')
     """
     _validate_config(model, data, trainer, log=log, resume=resume, optim=optim)
+
+    # [ModelOpt]: If modelopt_state exists, overwrite transformer_layer_spec to modelopt spec
+    if resume and resume.restore_config and resume.restore_config.path:
+        set_modelopt_spec_if_exists_in_ckpt(model, resume.restore_config.path)
```
kevalmorabia97, line 165: This hack will allow us to use modelopt spec if trying to run `llm.train` or `llm.finetune` on a ptq nemo ckpt (i.e. QAT)

kevalmorabia97 (commented):
Added comments to main changed parts that you can review. Rest all is just refactoring of existing logic.

Laplasjan107 (commented):
Thanks a lot for this refactor! It significantly improves the code

nemo/collections/llm/modelopt/model_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    model_path: str,
+    tensor_model_parallel_size: int = 1,
+    pipeline_model_parallel_size: int = 1,
+    devices: int = 1,
+    num_nodes: int = 1,
```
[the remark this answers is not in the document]
kevalmorabia97: Maybe user only wants to use a subset of gpus? If a model is small enough that user don't need TP/PP - then also they may use a subset of gpus.

Overall don't have strong opinions on either way but saw that all nemo llm examples follow similar structure that they require users to pass devices and nodes as cmdline arguments

nemo/collections/llm/modelopt/model_utils.py
```diff
@@ -0,0 +1,142 @@
...
+"""Utility functions for loading models with modelopt layer spec."""
+
+import lightning.pytorch as L
+import torch
+from megatron.core.dist_checkpointing.validation import StrictHandling
+
+from nemo import lightning as nl
+from nemo.collections import llm
+from nemo.collections.llm.inference.base import _setup_trainer_and_restore_model
+from nemo.collections.nlp.modules.common.tokenizer_utils import get_tokenizer
```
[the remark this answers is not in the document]
kevalmorabia97: Interesting. So is there any alternative function to load tokenizer?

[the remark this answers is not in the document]
kevalmorabia97: I do see `nemo.collections.nlp` imports in https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/data/pre_training.py#L226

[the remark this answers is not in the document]
Laplasjan107: I don't know if there is any alternative. I suppose that by default the tokenizer is stored inside the checkpoint. cc: @janekl @marcromeyn @dimapihtar

nemo/collections/llm/modelopt/model_utils.py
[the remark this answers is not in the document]
Laplasjan107: Sure, the `torch.cuda.device_count()` would be wrong, but I think `os.environ.get("LOCAL_WORLD_SIZE", 1)` needs to match the number of used devices per node.

Anyway, that's probably a question for another PR, cc: @janekl

nemo/collections/llm/modelopt/model_utils.py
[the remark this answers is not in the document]
kevalmorabia97: By default yes. Main use case for this was in prune/distillation cicd test where llama-ci-hf doesn't have a tokenizer and default nmt tokenizer's vocab size doesn't match the vocab size of the tokenized pertaining test dataset hence we needed to overwrite tokenizer to gpt2 to fix it.

nemo/collections/llm/api.py
```diff
@@ -95,6 +105,10 @@ def train(
         >>> llm.train(model, data, trainer, tokenizer="data")
         PosixPath('/path/to/log_dir')
     """
+    # [ModelOpt]: If modelopt_state exists, overwrite transformer_layer_spec to modelopt spec
+    if resume and resume.restore_config and resume.restore_config.path:
```
jenchen1398, line 109: If user just passes a path to the checkpoint directly using `AutoResume.resume_from_path` ([here](https://github.com/NVIDIA/NeMo/blob/main/nemo/lightning/resume.py#L72-L73)) this line doesn't get run. the logic should also handle `resume.resume_from_path` 

something like 
```
if resume:
    if resume.restore_config and resume.restore_config.path:
        set_modelopt_spec_if_exists_in_ckpt(model, resume.restore_config.path)
    elif resume.resume_from_path:   
         set_modelopt_spec_if_exists_in_ckpt(model, resume.resume_from_path)
```

other edge case is if user sets `resume_from_directory` in which case there may be multiple checkpoints ... not sure how to best to handle that

jenchen1398: also you might need to disable gradient accumulation here .. I ran into this error
```
4: [rank4]:   File "/opt/megatron-lm/megatron/core/tensor_parallel/layers.py", line 510, in backward
4: [rank4]:     if weight.main_grad.dtype == torch.float32:
4: [rank4]:        ^^^^^^^^^^^^^^^^
4: [rank4]: AttributeError: 'Tensor' object has no attribute 'main_grad'
```

```
is_quantized_model = set_modelopt_spec_if_exists_in_ckpt(model, resume.restore_config.path)
if is_quantized_model:
            model.config.gradient_accumulation_fusion = False
```

nemo/collections/llm/api.py
```diff
@@ -95,6 +105,13 @@ def train(
         >>> llm.train(model, data, trainer, tokenizer="data")
         PosixPath('/path/to/log_dir')
     """
+    # [ModelOpt]: If modelopt_state exists, overwrite transformer_layer_spec to modelopt spec
+    if resume:
+        if resume.restore_config and resume.restore_config.path:
+            set_modelopt_spec_if_exists_in_ckpt(model, resume.restore_config.path)
+        elif resume.resume_from_path:
+            set_modelopt_spec_if_exists_in_ckpt(model, resume.resume_from_path)
```
kevalmorabia97, line 113: @yashaswikarnati is there any utility function we can use to get path based on `resume.restore_config.path`, `resume.resume_from_path` or `resume.resume_from_dir`?

scripts/llm/gpt_prune.py
```diff
@@ -112,116 +91,41 @@ def get_data_module(args, tokenizer):
         data_module_cls = llm.MockDataModule
     data_module = data_module_cls(
         seq_length=args.seq_length,
-        tokenizer=tokenizer,
         micro_batch_size=args.mbs,
-        global_batch_size=args.gbs,
+        global_batch_size=args.mbs,
```
yashaswikarnati, line 95: is this because for pruning, we aggregate metrics based on mbs. Could we add a comment or log that gbs is not used

scripts/llm/gpt_prune.py
```diff
@@ -112,116 +91,41 @@ def get_data_module(args, tokenizer):
...
-        pipeline_dtype=torch.bfloat16,
-        sequence_parallel=False,
-        ckpt_load_optimizer=False,
-        ckpt_load_strictness=StrictHandling.LOG_ALL if args.legacy_ckpt else None,
-        ckpt_parallel_save_optim=False,
-        setup_optimizers=False,
-        ddp="pytorch",
-        replace_progress_bar=False,
+    pruning_config = PruningConfig(
+        ffn_hidden_size=args.prune_ffn_hidden_size,
```
yashaswikarnati, line 105: wonder if this is size to be prunued? or the final size? wonder if target_ffn_hidden size is better here

scripts/llm/gpt_prune.py
```diff
@@ -63,41 +63,20 @@
 import argparse
 import os
 
-import modelopt.torch.prune as mtp
-import torch
-from megatron.core import dist_checkpointing
-from megatron.core.dist_checkpointing.validation import StrictHandling
-from megatron.core.inference.modelopt_support.gpt.model_specs import get_gpt_layer_modelopt_spec
+# Import modelopt first to avoid circular import causing pruning to fail
+# TODO: This can be removed in modelopt 0.27 once modelopt pruning imports nemo inside a function
```
yashaswikarnati, line 67: curious why modelopt code would import nemo

kevalmorabia97: Had some checks that the model to be pruned is one of the supported types else will raise an error.
Generally all nemo related files are in `modelopt/torch/.../plugins/...` and we don't auto-import plugin files unless invoked so this isn't a problem. But one of the other pruning files had the nemo import which we will move to a function to avoid the circular dependency

scripts/llm/gpt_prune.py
kevalmorabia97: Good idea. Will rename config keys to `target_...`

scripts/llm/gpt_prune.py
kevalmorabia97: Right. There is no backward so GBS doesn't matter. Plus making GBS same as MBS means the progress bar / logging will be more request giving a better user experience.

Yes I'll add a note that  GBS is not used

nemo/collections/llm/modelopt/prune/pruner.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+
+    return model
+
+
+def save_pruned_model(model: llm.GPTModel, trainer: nl.Trainer, save_path: str):
+    """Save pruned model nemo checkpoint."""
+    logging.info(f"Saving pruned model to {save_path}...")
+    if hasattr(trainer.model, "__io__") and hasattr(trainer.model.tokenizer, "__io__"):
+        trainer.model.__io__.tokenizer = trainer.model.tokenizer.__io__
```
yashaswikarnati, line 125: why is this needed?

kevalmorabia97: Followed the nemo1_to_nemo2 script's ckpt saving logic: https://github.com/NVIDIA/NeMo/blob/main/scripts/checkpoint_converters/convert_nemo1_to_nemo2.py#L256-L257

Should I not need it?

kevalmorabia97: Removed the tokenizer updating logic as it works without as well

scripts/llm/gpt_prune.py
yashaswikarnati: got it, thanks! Wonder if its better to move the logic for type checking to nemo itself before invoking modelopt api's. I'm not sure if its a good idea that nemo imports modelopt and some function of modelopt try to import nemo

scripts/llm/gpt_distillation.py
```diff
@@ -87,26 +85,6 @@ def get_args():
         plugins=nl.MegatronMixedPrecision(precision=args.precision),
     )
 
-    ## Load both models and combine into an aggregate module
-    _student_model = nl.io.load_context(path=ckpt_to_context_subdir(args.student_path), subpath="model")
-    _teacher_model = nl.io.load_context(path=ckpt_to_context_subdir(args.teacher_path), subpath="model")
-    assert isinstance(_student_model, llm.GPTModel), "Only models based on `llm.GPTModel` are supported currently."
```
yashaswikarnati, line 93: We dont need this assert anymore?

kevalmorabia97: All this logic is moved to api.py https://github.com/NVIDIA/NeMo/blob/kmorabia/modelopt-refactoring/nemo/collections/llm/api.py#L418

nemo/collections/llm/modelopt/prune/pruner.py
```diff
@@ -0,0 +1,138 @@
...
+
+        logging.info("Pruning model...")
+        export_config = {
+            k: getattr(pruning_config, f"target_{k}")
+            for k in SUPPORTED_PRUNING_HPARAMS
+            if getattr(pruning_config, f"target_{k}") is not None
+        }
+        mtp.prune(
+            model,
+            mode="mcore_gpt_minitron",
```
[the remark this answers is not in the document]
kevalmorabia97: Currently this is the only one supported for NeMo. Once we have multiple modes then we can make this a cmdline argument

scripts/llm/gpt_prune.py
kevalmorabia97: ModelOpt can work on M-LM models as well. But regardless main motivation for nemo import is to identify how to extract model config and update its values to pruned choices. For example in NeMo 1, its `model.cfg` while for NeMo 2 its `model.config`.

The import for pruning case is here for your reference (will be moved to a function in next release): https://github.com/NVIDIA/TensorRT-Model-Optimizer/blob/main/modelopt/torch/prune/mcore_gpt_minitron.py#L41-L53

chtruong814 (approved):

PR #12471, opened 2025-03-04
fix auto resume for peft

stevehuang52:
# What does this PR do ?

Fix two bugs in AutoResume when PEFT ckpt is loaded:
- `restore_config` tries to load tokenizer from the PEFT ckpt instead of the base LLM.
- `resume_from_path` doesn't load the PEFT ckpt properly

**Collection**: [llm]


cuichenx (approved):
LGTM, thanks!

PR #12551, opened 2025-03-11
Add packed sequences for HFDatasetDataModule

athitten:
> [!IMPORTANT]  

# What does this PR do ?

Adds support for sequence packing in HFDatasetDataModule class for automodel.
1) Creates a method `pack()` in `HFDatasetDataModule` class that calls the `pack` method from `HFDatasetPackedSequence` class which implements the packing logic and contains all the util functions used by `pack()` method.
2) Modifies `collate_fn` of `HFDatasetDataModule` class to also get the `attention_mask` from the packed batch which is the arg of the `collate_fn`.
3) Creates its own `collate_fn` for HFMockDataModule class without the packing related code.
 
# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/llm/gpt/data/hf_dataset.py
```diff
@@ -297,19 +303,29 @@ def __init__(
...
+        self.packed_sequence_size = packed_sequence_size
 
     @staticmethod
     def from_dict(dataset_dict, split, **kwargs):
         """wraps Dataset's from_dict method"""
         dataset = Dataset.from_dict(dataset_dict)
         return HFDatasetDataModule(path_or_dataset=dataset, split=split, **kwargs)
 
-    @staticmethod
-    def collate_fn(batch, pad_token_id=0, pad_seq_len_divisible=None):
```
athitten, line 311: Make `collate_fn` a class method (instead of static) to be able to access instance variables like `packed_sequence_size` inside the `collate_fn`.

nemo/collections/llm/gpt/data/hf_dataset.py
```diff
@@ -284,7 +290,7 @@ def __init__(
         self.dataset_splits = make_dataset_splits(dataset, split, split_aliases)
 
         if collate_fn is None:
-            self._collate_fn = lambda x: HFDatasetDataModule.collate_fn(
+            self._collate_fn = lambda x: self.collate_fn(
```
athitten, line 293: `self.collate_fn` since `collate_fn` is made an instance method and not a static method.

nemo/collections/llm/gpt/data/hf_dataset.py
```diff
@@ -574,7 +605,7 @@ def __init__(
         self.pin_memory = pin_memory
         self.persistent_workers = persistent_workers
         self.create_attention_mask = create_attention_mask
-        self.collate_fn = lambda x: HFDatasetDataModule.collate_fn(x, pad_token_id=0)
+        self.collate_fn = lambda x: HFMockDataModule.collate_fn(x, pad_token_id=0)
```
athitten, line 608: Create an explicit and simple `collate_fn` for `HFMockDataModule` as the one from `HFDatasetDataModule` is not static anymore

BoxiangW (approved):

examples/llm/sft/automodel.py
```diff
@@ -67,12 +69,17 @@ def formatting_prompts_func(example):
...
         pad_seq_len_divisible=16 if fp8 else None,  # FP8 training requires seq length to be divisible by 16.
+        packed_sequence_size=packed_sequence_size,
     )
+    ## tokenization is happening here
     datamodule.map(
         formatting_prompts_func,
         batched=False,
         remove_columns=["id", "title", "context", "question", 'answers'],
     )
+    # Pack the sequences in the dataset if packed_sequence_size > 0
```
akoumpa, line 80: I'm wondering if there's a way to hide the `.pack` call within the `HFDatasetDataModule` class? For example, by moving this check just before the creation of the dataloaders (therefore all the .map calls have been applied) or similarly? The scenario I have in mind is a user wants to use packed sequences and they set the `packed_sequence_size` parameter accordingly but for some reason they forgot to call `.pack`. I'm still reading the code so I may have missed something. Please let me know what you think @athitten .

nemo/collections/llm/gpt/data/hf_dataset_packed_sequence.py
```diff
@@ -0,0 +1,297 @@
...
+import torch
+from datasets import Dataset
+from torch.nn import functional as F
+from tqdm import tqdm
+
+CROSS_ENTROPY_IGNORE_IDX = -100
+PACK_TYPE = Dict[str, Union[torch.Tensor, List[int]]]
+
+
+class HFDatasetPackedSequence:
```
akoumpa, line 26: I'm wondering if it would be better to subclass HFDataset here instead, WDYT @athitten ?

akoumpa:
Overall this looks great and implements an important feature, congrats @athitten .

I made a few minor comments, apart from the those, if there's time it would be great to include some unit tests (e.g. via `pytest`), to ensure that future changes do not break any code.

examples/llm/sft/automodel.py
athitten: Thanks for the review @akoumpa ! This a good point. Thinking of it, it should be possible to call packing method [here](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/data/hf_dataset.py#L342) if `packed_sequence_size > 0`. Let me try to make this change and check.

nemo/collections/llm/gpt/data/hf_dataset_packed_sequence.py
athitten: The intention of `HFDatasetPackedSequence` is just to have instance variables so that they can be accessed within all the util methods of the main pack method like `_should_stop_packing`, `_split_and_add_pack` etc., Having each of these methods as separate functions outside of a class can make the function definitions very long leading to a not very clean code. With this being the only intention, having `HFDatasetPackedSequence` subclass `HFDatasetDataModule` won't be of any help. Also the main pack method is already part of `HFDatasetDataModule` class [here](https://github.com/NVIDIA/NeMo/blob/athitten/packed_sequence_automodel/nemo/collections/llm/gpt/data/hf_dataset.py#L418). So `HFDatasetPackedSequence` is more of a helper class here. Let me know if this is okay. Maybe I can mention it as `helper` in the class name so that its not confusing ?

athitten:
> Overall this looks great and implements an important feature, congrats @athitten .
> 
> I made a few minor comments, apart from those, if there's time it would be great to include some unit tests (e.g. via `pytest`), to ensure that future changes do not break any code.

Yes unit tests and docs were going to be added in a subsequent PR soon.

examples/llm/sft/automodel.py
athitten: Added packing in [_make_dataloader](https://github.com/NVIDIA/NeMo/pull/12551/files#diff-ec2cdd55745a4a56dff9696f648b829d311ee8d7676a73d761adc0764153f011R447) method of `HFDatasetDataModulePacked` class.

nemo/collections/llm/gpt/data/hf_dataset_packed_sequence.py
athitten: I did end up creating a [HFDatasetDataModulePacked](https://github.com/NVIDIA/NeMo/pull/12551/files#diff-ec2cdd55745a4a56dff9696f648b829d311ee8d7676a73d761adc0764153f011R411) class that overrides certain methods from HFDatasetDataModule class to support packing. This class gets instantiated in `examples/llm/sft/automodel.py`  if packed_sequence_size > 0

akoumpa (dismissed):
LGTM, thank you.

akoumpa (approved):

nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
```diff
@@ -291,9 +291,16 @@ def training_step(self, batch, batch_idx=None, context_parallel=False):
             batch['input_ids'] = batch['tokens']
 
         # TODO(@boxiangw): Refractor. Needed for SP support
-        batch["position_ids"] = torch.arange(0, batch['input_ids'].shape[1]).unsqueeze(0).to(self.model.device)
+        # We cant have this for packed seq, attn_mask not in batch indicates regular sequence (not packed)
+        if 'attention_mask' not in batch:
+            batch["position_ids"] = torch.arange(0, batch['input_ids'].shape[1]).unsqueeze(0).to(self.model.device)
```
athitten, line 296: @akoumpa your commits have removed the attention_mask so this line can potentially override the position_ids in case of packed sequence which is incorrect.

examples/llm/sft/automodel.py
```diff
@@ -36,7 +36,14 @@
...
 def make_squad_hf_dataset(
-    tokenizer, micro_batch_size, seq_length, limit_dataset_samples=None, fp8=False, num_replicas=1, rank=0
+    tokenizer,
+    micro_batch_size,
+    seq_length,
+    packed_sequence_size,
+    limit_dataset_samples=None,
+    fp8=False,
+    num_replicas=1,
+    rank=0,
```
athitten, line 46: Also previously the reason for getting a matching loss curve with PS and without was because of seq-len default being 2048 and also the packed sequence testing was done with packed-sequence-size 2048. So all the seqs in squad were getting padded upto 2048 leading to no packing at all and it ran similar to running with no packing. We should disable padding while running packed-sequences.

nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
BoxiangW: This line is needed for SP support actually. Also there is another CP-related change for `batch['position_ids']` as well. Will pack-seq input position_ids then? But this change is just for batch["position_ids"] to match the shape of `batch['input_ids']`. I think for pack-seq, we can address that instead

athitten: Yes `PS` uses `position_ids` and the batch will contain `position_ids` while using `PS`. I was thinking just not do this line below if `"position_ids"` already exists in `batch` which is the case for `PS`. Does that sound okay ?
`batch["position_ids"] = torch.arange(0, batch['input_ids'].shape[1]).unsqueeze(0).to(self.model.device)`

Can you please share what code this is referring to ?
`Also there is another CP-related change for batch['position_ids'] as well.`

BoxiangW: https://github.com/NVIDIA/NeMo/pull/12821/files#diff-a1007f10a48739b42901a65bd9baf0ccd0a4ae2f42bdf083eef8a684cba28360R280 This is the one for CP. Can you try run it with TP/SP and CP after ignoring them then? Loss curve would be affected if the shape does not match

athitten: Do you mean run PS with TP/SP and CP ? If yes for PS we are just focusing on minimal functionality atm without any model parallelisms.

BoxiangW: I see, thanks for the info

examples/llm/sft/automodel.py
athitten: Disabled padding if packed_sequence_size > 0: [here](https://github.com/NVIDIA/NeMo/pull/12551/commits/26bd344948111711d404364dea346c19194b8822#diff-6ca255a65c7e092b5f5720b3ab66bd9dd17cdc9d98ebea5f37789e1a48ec4c41R60)

nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
athitten: Added check to not override position_ids if it already exists in the batch: [here](https://github.com/NVIDIA/NeMo/pull/12551/commits/26bd344948111711d404364dea346c19194b8822#diff-a1007f10a48739b42901a65bd9baf0ccd0a4ae2f42bdf083eef8a684cba28360R296)

akoumpa (approved):

chtruong814:
CI passed here except for an optional test.  Merging this now
https://github.com/NVIDIA/NeMo/actions/runs/14593384135/attempts/1

PR #12565, opened 2025-03-11
Remove adapter_path from base AutoResume and refactor PEFT checkpoint handling

hemildesai:
This commit simplifies the AutoResume class by:
- Removing the `adapter_path` attribute from the base class
- Refactoring PEFT checkpoint resume logic to be more robust
- Improving base model path extraction for adapter checkpoints
- Removing redundant code paths in checkpoint resolution

The changes ensure more consistent and straightforward handling of adapter and base model checkpoints during resume operations.

> [!IMPORTANT]  


nemo/lightning/resume.py
```diff
@@ -165,19 +162,17 @@ def _extract_path(
...
-        base_model_path = self._extract_path(model, self.restore_config.path)
-        if base_model_path not in [Path(metadata['model_ckpt_path']), Path(metadata['model_ckpt_path']).parent]:
-            logging.warning(
-                f"⚠️ When trying to resume a PEFT training run, found mismatching values: "
-                f"your specified restore_path points to {base_model_path}, "
-                f"but the PEFT checkpoint was trained with "
-                f"model_ckpt_path={metadata['model_ckpt_path']}"
-            )
+        # Use the model_ckpt_path from metadata directly
+        base_model_path = Path(metadata["model_ckpt_path"])
```
stevehuang52, line 190: `metadata["model_ckpt_path"]` stores the path of the base LLM used in training, but this path won't exist if we run inference on a different machine that has the LLM stored in a difference location. Have you tested this use case?

cuichenx: I have the same question. Should we keep this warning message?

nemo/lightning/pytorch/strategies/utils.py
```diff
@@ -51,7 +51,6 @@ class RestoreConfig:
     """
 
     path: str
-    adapter_path: Optional[str] = None
```
cuichenx, line 54: remove `adapter_path` in docstring as well :)

nemo/lightning/resume.py
```diff
@@ -123,13 +141,10 @@ def setup(self, trainer: Union[pl.Trainer, fl.Fabric], model=None):
             new_path = self._extract_path(
                 model=model,
                 path=self.restore_config.path,
-                adapter_path=self.restore_config.adapter_path,
+                adapter_path=None,
```
cuichenx, line 144: can we just remove this arg from self._extract_path?

nemo/lightning/resume.py
```diff
@@ -153,7 +168,6 @@ def _extract_path(
             new_path = path
 
         if adapter_path:
```
cuichenx, line 170: can we remove this entire if block

nemo/lightning/resume.py
cuichenx: Never mind, I understand that previously there were two ways to specify the base model -- 1) self.restore_config.path and 2) adapter_metadata.jsonI. Now we only want to keep 2) 

I think it is a clean interface and we should keep it this way. We should add a check to see whether Path(metadata["model_ckpt_path"]) exists, and let user know to update the path if it doesn't exist. I have this also noted in the [docs](https://docs.nvidia.com/nemo-framework/user-guide/latest/sft_peft/peft_nemo2.html#run-peft-inference-in-nemo-2-0), so that should be enough

stevehuang52:
@cuichenx Could you please review?

cuichenx (approved):
LGTM!

PR #12659, opened 2025-03-18
[automodel] fix validation_step

soluwalana:
# What does this PR do ?

The default `hf_auto_model_for_causal_lm` recipe fails to run due to the model being loaded on CPU, the validation loss being logged twice with different values, the validation_loss not being returned from the hf_auto_model_for_causal_lm validation step, and the validation_step passing the wrong parameters to forward.

**Collection**: llm

# Changelog

- Removes the hardcoded "cpu" device for loading the model
- Fixes the validation loss in hf_auto_model_for_causal to ensure that it calls forward with the correct parameter
- Returns loss from valdiation_step in hf_auto_model_for_causal
- Adds a check on loss in FSDP2 to ensure that None losses are not logged.

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
+++ b/nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
@@ -322,19 +322,19 @@ class HFAutoModelForCausalLM(pl.LightningModule, io.IOMixin, fn.FNMixin):
                 group = device_mesh.get_group()
 
             def reduce_item(val, op, device, group, dtype):
-                 """util function"""
-                 divide_by_world_size = False
-                 if torch.distributed.get_backend(group) == "gloo" and op == dist.ReduceOp.AVG:
-                     # GLOO does not support the `ReduceOp.AVG` operation
-                     op = dist.ReduceOp.SUM
-                     divide_by_world_size = True
+                """util function"""
+                divide_by_world_size = False
+                if torch.distributed.get_backend(group) == "gloo" and op == dist.ReduceOp.AVG:
+                    # GLOO does not support the `ReduceOp.AVG` operation
+                    op = dist.ReduceOp.SUM
+                    divide_by_world_size = True
 
-                 val = torch.tensor([val], device=device, dtype=dtype).detach()
-                 dist.all_reduce(val, group=group, op=op)
-                 val = val.item()
-                 if divide_by_world_size:
-                     val /= dist.get_world_size(group)
-                 return val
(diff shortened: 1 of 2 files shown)


nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
```diff
@@ -176,7 +176,7 @@ def _configure_model(self, attn_implementation):
             return auto_cls.from_pretrained(
                 self.model_name,
                 torch_dtype=torch.bfloat16,
-                device_map="cpu",
+                device_map=self.device,
```
akoumpa, line 179: Hi, I want to keep the "cpu" value since setting this to self.device may result in OOM for models that don't fit completely on the GPU.

nemo/lightning/pytorch/strategies/fsdp2_strategy.py
```diff
@@ -434,15 +435,8 @@ def lightning_module_state_dict(self) -> Dict[str, Any]:
         Only returns a non-empty state dict on rank 0 if ``save_distributed_checkpoint=False``.
 
         """
-        from nemo.lightning.pytorch.strategies.utils import to_cpu
-
         assert self.lightning_module is not None
         state_dict = self.lightning_module.state_dict()
-
-        module_names = list(state_dict.keys())
```
akoumpa, line 442: I wonder what's the motivation for this case? Do you recall if is there a value which break `to_cpu` or if it was something else?

nemo/collections/llm/gpt/model/hf_auto_model_for_causal_lm.py
soluwalana: This seems pretty unreasonable because it isn't overridable and is the default way this model gets created when using the recipe.

nemo/lightning/pytorch/strategies/fsdp2_strategy.py
soluwalana: I have not experienced any issues with this *not* being on CPU.  I think that this was set this way due to the fact that the device map was hardcoded to "CPU"

akoumpa:
I will split this PR into two, one to fix the validation_step and another to optimize the lightning_module_state_dict function. I want to make sure lightning_module_state_dict does not break anything before proceeding.

akoumpa (approved):
LGTM, thanks.

PR #12695, opened 2025-03-19
[DEPRECATED] Add logs for checkpoint saving start and finalization - see https://github.com/NVIDIA/NeMo/pull/12697 instead

lepan-google:
- Add logs to the start of checkpoint save with timestamp.
- Add logs to the checkpoint finalization function with timestamp.
- This helps evaluate the time taken to write a checkpoint into storage.

### DIFF
--- a/nemo/collections/llm/tools/auto_configurator/core/calculate_performance.py
+++ b/nemo/collections/llm/tools/auto_configurator/core/calculate_performance.py
@@ -14,12 +14,10 @@
 
 import os
 import re
-import pandas as pd
 from typing import Optional
 
-
+import pandas as pd
 from tensorboard.backend.event_processing import event_accumulator
-
 
 GPT_BASED_MODELS = [
     "gpt3",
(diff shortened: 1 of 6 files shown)


nemo/collections/llm/tools/auto_configurator/core/calculate_performance.py
```diff
@@ -14,13 +14,11 @@
 
 import os
 import re
-import pandas as pd
 from typing import Optional
 
-
+import pandas as pd
```
polydier1, line 19: why we need this dependency?

nemo/collections/llm/tools/auto_configurator/runner.py
```diff
@@ -15,13 +15,14 @@
 import copy
 import os
 import re
-import numpy as np
```
polydier1, line 18: the changes seem not needed in this file

polydier1 (approved):

PR #11972, opened 2025-01-28
[checkpoint] Log timings for checkpoint IO save and load

ananthsub:
# What does this PR do ?

Adds timings around checkpoint save/load IO calls within Nemo1 and Nemo2 Lightning strategies.

To test, sanity checked logs from existing checkpointing tests:
```
pytest -v --log-cli-level=INFO --capture=no -s tests/core/test_dist_ckpt.py::TestAsyncSave::test_async_save_produces_same_checkpoints_as_sync
```

**Collection**: nlp, llm

# Changelog 
Adds timings around checkpoint save/load IO calls within Nemo1 and Nemo2 Lightning strategies

# Usage
* You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"
**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed. 
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information
* Related to # (issue)

PR #12697, opened 2025-03-19
Add logs for checkpoint saving start and finalization

lepan-google:
- Add logs to the start of checkpoint save with timestamp.
- Add logs to the checkpoint finalization function with timestamp.
- This helps evaluate the time taken to write a checkpoint into storage.


polydier1 (approved):

ananthsub:
for checkpoint timing, are you looking at the time required to generate the state dict + the time to write to storage (or if async saving is used, the time to schedule an async request), or just the latter? if the latter, see https://github.com/NVIDIA/NeMo/pull/11972 - it would be more precise to add the timings to the CheckpointIO classes vs. the ModelCheckpoint callback.

lepan-google:
> for checkpoint timing, are you looking at the time required to generate the state dict + the time to write to storage (or if async saving is used, the time to schedule an async request), or just the latter? if the latter, see #11972 - it would be more precise to add the timings to the CheckpointIO classes vs. the ModelCheckpoint callback.

Hey @ananthsub, we use async saving, and we hope to include all the durations which are introduced by checkpoint saving, especially if the steps block GPU. This will help the user determine the checkpoint saving frequency.

Based on my current understanding, between the [checkpoint saving start time we added](https://github.com/NVIDIA/NeMo/blob/8341116325124cd8a0e5bdacc5a74a3c5fd32610/nemo/lightning/pytorch/callbacks/model_checkpoint.py#L625) and the [start of  CheckpointIO saving](https://github.com/NVIDIA/NeMo/blob/b180686ea888ebfafa4f8f68b5690641a28a34ba/nemo/utils/callbacks/dist_ckpt_io.py#L287), there are several steps:
- Convert the optimizer states to their sharded equivalents: [link](https://github.com/NVIDIA/NeMo/blob/208e0da28e2ada8da84d8f7ddff8623efe1ff01c/nemo/collections/nlp/parts/nlp_overrides.py#L412)
- Determine the shard strategy:
   - [link](https://github.com/NVIDIA/NeMo/blob/0366d9d41188a10d1c47382f35407250f9a9031a/nemo/utils/callbacks/dist_ckpt_io.py#L291)
   - [link](https://github.com/NVIDIA/NeMo/blob/0366d9d41188a10d1c47382f35407250f9a9031a/nemo/utils/callbacks/dist_ckpt_io.py#L468)

And based on our observations, we saw there are gaps between [checkpoint saving start time we added](https://github.com/NVIDIA/NeMo/blob/8341116325124cd8a0e5bdacc5a74a3c5fd32610/nemo/lightning/pytorch/callbacks/model_checkpoint.py#L625) and [shard strategy determination](https://github.com/NVIDIA/NeMo/blob/0366d9d41188a10d1c47382f35407250f9a9031a/nemo/utils/callbacks/dist_ckpt_io.py#L468) for the first checkpoint saving: 
image. 
And for the following checkpoint savings, we saw: 
image
The second gap should include in the [CheckpointIO saving](https://github.com/NVIDIA/NeMo/blob/b180686ea888ebfafa4f8f68b5690641a28a34ba/nemo/utils/callbacks/dist_ckpt_io.py#L287).

We are wondering whether these steps block the training or not, if it does, we hope to include that in the checkpoint write time. Please let me know if there are any other steps besides the two listed above.

We are also wondering whether the finalization functions block the training, for now, we consider these finalization steps are not async too.

Therefore, we currently calculate the checkpoint write time with this [start time](https://github.com/NVIDIA/NeMo/blob/8341116325124cd8a0e5bdacc5a74a3c5fd32610/nemo/lightning/pytorch/callbacks/model_checkpoint.py#L625) and [end time](https://github.com/NVIDIA/NeMo/blob/8341116325124cd8a0e5bdacc5a74a3c5fd32610/nemo/lightning/pytorch/callbacks/model_checkpoint.py#L657). Please let me know if our understanding is not correct or there are other places which make the checkpoint write time more accurate.

Thanks for your help!

BoxiangW (approved):

ko3n1g:
Looks like a minimal change, will force merge.

PR #12777, opened 2025-03-25
Refactor init and optim to take in individual args in nemo/tron

hemildesai:


nemo/tron/init.py
```diff
@@ -30,16 +30,24 @@
...
+from nemo.tron.config import DistributedInitConfig, RerunStateMachineConfig, RNGConfig
 from nemo.tron.utils.common_utils import get_local_rank_preinit, get_rank_safe, get_world_size_safe
 
 
 def initialize_megatron(
-    # extra_args_provider=None,
-    # args_defaults={},
-    # ignore_unknown_args=False,
-    cfg: ConfigContainer,
+    model_config: GPTConfig | T5Config,
```
ananthsub, line 40: this is likely to keep changing with new knobs added, and each required arg would be considered a breaking API change. while the changes to `setup_optimizer` are much more self contained, `initialize_megatron` could get really unwieldy over time. do you think its reasonable to accept the top-level config here?

ananthsub (approved):

ananthsub (approved):

PR #12830, opened 2025-04-01
Adds support for SFT and Llama + Qwen converters

hemildesai:


nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+
+    config.use_cpu_initialization = _orig_use_cpu_initialization
+
+
+@contextmanager
+def temporary_distributed_context():
+    if "MASTER_ADDR" not in os.environ:
+        os.environ["MASTER_ADDR"] = "localhost"
+    if "MASTER_PORT" not in os.environ:
+        os.environ["MASTER_PORT"] = "12355"
```
ananthsub, line 62: if these were added as env vars, they should also be removed after control is yielded back

nemo/tron/config.py
```diff
@@ -104,8 +106,8 @@ class TokenizerConfig:
     padded_vocab_size: Optional[int] = None
 
 
-@dataclass
-class GPTDatasetConfig(MCoreGPTDatasetConfig):
+@dataclass(kw_only=True)
+class DataloaderConfig:
```
ananthsub, line 110: do we also need these settings to be configurable?

https://github.com/NVIDIA/NeMo/blob/edc87f0e6b36257960dd76a75e93610f55cb9dc6/nemo/collections/llm/gpt/data/fine_tuning.py#L73-L74

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+    assert isinstance(dtype, str)
+    if dtype in ["float16", "fp16", "16", "16-mixed"]:
+        return torch.float16
+    elif dtype in ["bfloat16", "bf16-mixed"]:
+        return torch.bfloat16
+    else:
+        return torch.float32
+
+
+def dtype_from_hf(config):
```
ananthsub, line 111:
```suggestion
def dtype_from_hf(config) -> torch.dtype
```

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+    """Save HF tokenizer to the imported NeMo model"""
+    from transformers import AutoTokenizer
+
+    tok = AutoTokenizer.from_pretrained(tokenizer_name_or_path, trust_remote_code=True)
+    # Save tokenizer assets to save_path.
+    tok.save_pretrained(save_path)
+    return save_path
+
+
+def dtype_from_str(dtype):
```
ananthsub, line 98:
```suggestion
def dtype_from_str(dtype: str) -> torch.dtype:
```

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    Convert a str precision to equivalent torch dtype.
+    """
+    assert isinstance(dtype, str)
+    if dtype in ["float16", "fp16", "16", "16-mixed"]:
+        return torch.float16
+    elif dtype in ["bfloat16", "bf16-mixed"]:
+        return torch.bfloat16
```
ananthsub, line 106:
```suggestion
    if dtype in ("float16", "fp16", "16", "16-mixed"):
        return torch.float16
    elif dtype in ("bfloat16", "bf16-mixed"):
        return torch.bfloat16
```

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+                dataset_config=None,
+                logger_config=LoggerConfig(),
+                tokenizer_config=None,
+                checkpoint_config=CheckpointConfig(
+                    async_save=False, save=str(self.output_path), save_optim=False, ckpt_format="torch_dist"
+                ),
+                dist_config=None,
+                ft_config=None,
+                straggler_config=None,
+                profiling_config=None,
```
ananthsub, line 215: these are None by default - why explicitly include them this way?

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+        Returns:
+            Path: Path to the saved NeMo model
+        """
+        source = self.init_hf_model()
+
+        with temporary_distributed_context():
+            target = self.init_tron_model(self.tron_config)
+            self.convert_state(source, target[0])
+            state = GlobalState()
+            state.cfg = ConfigContainer(
```
ananthsub, line 200: this will be kind of tedious to update for each sub-level config we add. if this is all to save the model state dict, should we add an explicit API for that which takes in just what's needed?

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+
+        return _HuggingFaceTokenizer(
+            save_hf_tokenizer_assets(str(self.input_path), str(self.output_path / HF_ASSETS_DIR))
+        )
+
+    def init_tron_model(self, cfg: GPTConfig | T5Config):
+        with megatron_cpu_init_context(cfg):
+            model = cfg.configure_model(tokenizer=self.tokenizer)
+        return [model]
```
ananthsub, line 169: n00b question: why is this returned as a list of model parts? is this intended to be accessed anywhere outside `apply` ?

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def to(self, dtype):
+        # pylint: disable=C0115,C0116
+        for k, v in self._state_dict.items():
+            if v.dtype != dtype:
+                logger.warning(f"Converting {k} from {v.dtype} (source model) to {dtype} (target model)")
+            self._state_dict[k] = v.to(dtype)
+
+
+class BaseImporter:
```
ananthsub, line 145: nit: should this be an abstract base class for the unimplemented methods?

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        from transformers import AutoModelForCausalLM
+        from transformers.modeling_utils import no_init_weights
+
+        with no_init_weights(True):
+            return AutoModelForCausalLM.from_config(self.hf_config, torch_dtype=dtype)
+
+    def init_tron_model(self) -> tuple[dict, dict]:
+        """
+        This function loads the state dict directly from a distributed checkpoint, and modify the state dict
```
ananthsub, line 279:
```suggestion
        This function loads the state dict directly from a distributed checkpoint, and modifies the state dict
```

nemo/tron/converter/common.py
```diff
@@ -0,0 +1,346 @@
...
+        Returns
+        -------
+            tuple[Dict, Dict]: The loaded state dict and the yaml config dict.
+        """
+        tron_yaml = self.input_path / "run_config.yaml"
+        assert tron_yaml.exists(), f"Tron config file {tron_yaml} does not exist"
+        with open(tron_yaml, "r") as stream:
+            _config = yaml.safe_load(stream)
+        config = _config["model_config"]
+        config = instantiate(config)
```
ananthsub, line 295: easy to mix up `_config` and `config` in this method, especially below when the tokenizer config is being read

```suggestion
        model_config = _config["model_config"]
        model_config = instantiate(model_config)
```

nemo/tron/converter/llama.py
```diff
@@ -0,0 +1,112 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ananthsub, line 1: should this and qwen go under llm/ ?

nemo/tron/data/hf_dataset.py
```diff
@@ -0,0 +1,261 @@
...
+        split: Which split(s) of the dataset to use (e.g., "train", ["train", "validation"])
+        tokenizer: The tokenizer to use for preprocessing text
+        collate_fn: Function to collate samples into batches
+        seq_length: Maximum sequence length
+        is_built_on_rank: Function that returns True if dataset should be built on current rank
+        seed: Random seed for dataset operations
+        train_aliases: list of aliases for train split
+        test_aliases: list of aliases for test split
+        val_aliases: list of aliases for validation split
+        **kwargs: Additional arguments to pass to load_dataset or Dataset.from_dict
```
ananthsub, line 169: docs are off

nemo/tron/llm/utils.py
```diff
@@ -76,7 +94,7 @@ def _broadcast(item):
             dtype=torch.float32,
             device=torch.cuda.current_device(),
         )
-        if cfg.dataset_config.create_attention_mask:
+        if isinstance(cfg.dataset_config, FinetuningDatasetConfig) or cfg.dataset_config.create_attention_mask:
```
ananthsub, line 97: n00b question: could you share where in nemo this is currently handled?

nemo/tron/setup.py
```diff
@@ -164,11 +173,14 @@ def setup(
 
     # Data stuff.
     timers("train/valid/test-data-iterators-setup", log_level=0).start(barrier=True)
+    if "tokenizer" in inspect.signature(train_valid_test_datasets_provider).parameters:
```
ananthsub, line 176: could you describe this check more?

nemo/tron/tokenizers/tokenizer.py
```diff
@@ -8,15 +8,39 @@
...
+from megatron.core.datasets.megatron_tokenizer import MegatronTokenizer as MegatronTokenizerCore
 
 from nemo.tron.config import TokenizerConfig
 from nemo.tron.tokenizers.bert_tokenization import FullTokenizer as FullBertTokenizer
 from nemo.tron.tokenizers.gpt2_tokenization import GPT2Tokenizer
+from nemo.tron.tokenizers.multimodal_tokenizer import MultimodalTokenizer
 from nemo.tron.utils.common_utils import get_rank_safe, print_rank_0
 
 
+class MegatronTokenizer(MegatronTokenizerCore):
```
ananthsub, line 20: n00b question: why do we need this? in order to pass a tokenizer as a callable?

nemo/tron/utils/config_utils.py
```diff
@@ -0,0 +1,197 @@
...
+from omegaconf import OmegaConf
+
+from nemo.tron.utils.instantiate_utils import InstantiationMode, instantiate
+from nemo.tron.utils.yaml_utils import safe_yaml_representers
+
+T = TypeVar("T", bound="ConfigContainer")
+
+
+@dataclass(kw_only=True)
+class ConfigContainer:
```
ananthsub, line 32: unit tests for this will be extremely helpful

nemo/tron/utils/config_utils.py
```diff
@@ -0,0 +1,197 @@
...
+        Convert the config container to a dictionary.
+
+        Also converts any nested dataclasses (both ConfigContainer and regular dataclasses)
+        to dictionaries recursively.
+
+        Returns:
+            Dictionary representation of this config
+        """
+        result = {}
+        result["_target_"] = f"{self.__class__.__module__}.{self.__class__.__name__}"
```
ananthsub, line 118: should this use name or qualname?

nemo/tron/utils/config_utils.py
```diff
@@ -0,0 +1,197 @@
...
+            The converted value
+        """
+        if isinstance(value, ConfigContainer):
+            return value.to_dict()
+        elif is_dataclass(value) and not isinstance(value, type):
+            # Handle regular dataclasses
+            result = {}
+
+            # Add _target_ field for instantiation
+            result["_target_"] = f"{value.__class__.__module__}.{value.__class__.__name__}"
```
ananthsub, line 154: same question here

nemo/tron/utils/instantiate_utils.py
```diff
@@ -0,0 +1,358 @@
...
+            output_args = input_args
+    else:
+        raise InstantiationException(
+            f"Unsupported _args_ type: '{type(config_args).__name__}'. value: '{config_args}'"
+        )
+
+    return output_args, kwargs
+
+
+def instantiate(
```
ananthsub, line 199: nit: helpful to have the main API be at the top of the file for readers so they read this first

nemo/collections/llm/gpt/data/core.py
```diff
@@ -32,7 +33,8 @@
 )
 from nemo.core.classes import Dataset
 from nemo.lightning.base import NEMO_DATASETS_CACHE
-from nemo.utils import logging
```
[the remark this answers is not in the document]
hemildesai: NeMo utils logging doesn't log the file name, so it's really difficult to find out where the logs are coming from.

nemo/collections/llm/gpt/model/base.py
```diff
@@ -314,7 +316,7 @@ def configure_model(self, tokenizer, pre_process=None, post_process=None) -> "MC
         if not isinstance(transformer_layer_spec, ModuleSpec):
             transformer_layer_spec = transformer_layer_spec(self)
 
-        if hasattr(self, 'vocab_size'):
+        if hasattr(self, "vocab_size") and self.vocab_size is not None:
```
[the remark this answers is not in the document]
hemildesai: There are instances when you don't want to use the `tokenizer`. For eg, running a simple forward pass on dummy data. In those instances, there was no way to configure the model. Adding `vocab_size` as an attr to the config solves this.

nemo/tron/config.py
```diff
@@ -123,6 +128,17 @@ def __post_init__(self) -> None:
         assert self.eod_mask_loss is not None
 
 
+@dataclass(kw_only=True)
+class FinetuningDatasetConfig(DataloaderConfig):
+    dataset_root: Optional[Union[str, Path]] = None
+    seq_length: int = 1024
```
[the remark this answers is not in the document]
hemildesai: Yeah I can make it required.

nemo/tron/config.py
```diff
@@ -521,7 +546,7 @@ class ConfigContainer:
     straggler_config: Optional[StragglerDetectionConfig] = None
     profiling_config: Optional[ProfilingConfig] = None
 
-    def __post_init__(self):
+    def validate(self):
```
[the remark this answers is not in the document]
hemildesai: The code depends on `world_size`, which is not always available. For eg, when converting a checkpoint.

nemo/tron/utils/log_utils.py
```diff
@@ -14,53 +14,72 @@
...
+    modules_to_filter: Optional[list[str]] = None,
+    set_level_for_all_loggers: bool = False,
+) -> None:
+    """Sets the logging level and filters.
 
     Precedence:
-    1. Command line argument `--logging-level`
-    2. Env var `MEGATRON_LOGGING_LEVEL`
+    1. Config argument `logger_config.logging_level`
+    2. Env var `TRON_LOGGING_LEVEL`
```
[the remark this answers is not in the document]
hemildesai: Yeah I can reuse it. Do you know what its called?

nemo/tron/data/finetuning_dataset.py
```diff
@@ -0,0 +1,272 @@
...
+        print_rank_0(f"Building FinetuningDatasetBuilder with root={self.dataset_root}")
+
+        if self.packed_sequence_size > 0:
+            print_rank_0(f"Using packed sequences with size {self.packed_sequence_size}")
+
+    def prepare_data(self) -> None:
+        """Prepare data if needed."""
+        self.prepare_packed_data()
+
+    def prepare_packed_data(self) -> None:
```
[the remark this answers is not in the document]
hemildesai: This is so that child classes can overwrite them separately.

nemo/tron/data/finetuning_dataset.py
```diff
@@ -0,0 +1,272 @@
...
+        """Prepare packed sequence data if needed."""
+        if self.packed_sequence_size > 0:
+            from nemo.collections.llm.gpt.data.packed_sequence import prepare_packed_sequence_data
```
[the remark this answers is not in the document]
hemildesai: You mean the entire file? Or just duplicating this function in `nemo/tron/llm/data`?

nemo/tron/data/finetuning_dataset.py
```diff
@@ -0,0 +1,272 @@
...
+            self.validation_path if self.packed_sequence_size <= 0 else self.validation_path_packed,
+            pack_metadata_path=None if self.packed_sequence_size <= 0 else self.pack_metadata,
+            is_test=True,
+            **self.dataset_kwargs,
+        )
+
+        test_ds = (
+            self._create_dataset(
+                self.test_path,
+                tokens_to_generate=32,
```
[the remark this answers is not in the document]
hemildesai: Good catch, will update.

nemo/tron/converter/common.py
hemildesai: Good point, will update.

nemo/tron/config.py
hemildesai: Good catch, will update.

nemo/tron/converter/common.py
hemildesai: Good catch, will remove

nemo/tron/converter/common.py
hemildesai: Yeah this is a weird edge case. We can make an explicit API, but would like to do that in a follow up PR after this is merged.

nemo/tron/converter/common.py
hemildesai: Yeah I can make this an abstract base class.

nemo/tron/llm/utils.py
hemildesai: NeMo currently always sets the dataset on all ranks and doesn't broadcast. Ref: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/llm/gpt/model/base.py#L57

nemo/tron/setup.py
hemildesai: The pretraining dataset provider doesnt require tokenizer since its already preprocessed.

nemo/tron/tokenizers/tokenizer.py
hemildesai: Yes also to pass as a callable.

nemo/tron/converter/common.py
hemildesai: The `save_checkpoint` method expects a list

PR #12870, opened 2025-04-03
feat: support memmap dataset with object store via Multi-Storage Client

jayya2:
> [!IMPORTANT]  

# What does this PR do ?

Add support for using [multi-storage-client](https://github.com/NVIDIA/multi-storage-client) to load data from object store and construct memmap-based datasets.  

In multi-storage client we provide adaptors such as` msc.numpy.*` and `msc.pickle.*` which can handle both object store URLs (starting with `msc://...`) and POSIX file path (`/path/to/your/file`), so in theory, we can directly replace 
existing usages, for example`numpy.memmap(...)` with `msc.numpy.memmap(...)`.  

However multis-storage client is not a required dependency at the moment, so I pre-check the module availability to play it safe in this first introduction PR to multi-storage client library.

**Collection**: [NLP]

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/nlp/data/language_modeling/text_memmap_dataset.py
+++ b/nemo/collections/nlp/data/language_modeling/text_memmap_dataset.py
@@ -19,7 +19,7 @@ import os
 import pickle
 import time
 from functools import lru_cache, partial
-from typing import Callable, List, Optional, Type
+from typing import Callable, List, Optional, Type, TYPE_CHECKING
 
 import numpy as np
 import torch
@@ -34,6 +34,9 @@ try:
 except (ImportError, ModuleNotFoundError):
     MULTISTORAGECLIENT_AVAILABLE = False
 
+if TYPE_CHECKING:
+    from nemo.collections.common.tokenizers.tokenizer_spec import TokenizerSpec
+    
 __all__ = ["TextMemMapDataset", "CSVMemMapDataset", "build_index_files"]
 __idx_version__ = "0.2"  # index file version
 __idx_suffix__ = "idx"  # index file suffix
@@ -123,7 +126,7 @@ class TextMemMapDataset(Dataset):
         if sort_dataset_paths:
             self._files_list = sorted(self._files_list)
 
-        logging.info(f"Building data files")
+        logging.info("Building data files")
         # load all files into memmap
         is_distributed = torch.distributed.is_available() and torch.distributed.is_initialized()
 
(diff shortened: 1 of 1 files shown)


cuichenx (commented):
Looks good, just one question about MSC: if `multistorageclient` is a package in the container (so that `MULTISTORAGECLIENT_AVAILABLE =True`, but the environment is a normal file system, is there any impact to run multistorageclient functions

jayya2:
> Looks good, just one question about MSC: if `multistorageclient` is a package in the container (so that `MULTISTORAGECLIENT_AVAILABLE =True`, but the environment is a normal file system, is there any impact to run multistorageclient functions

@cuichenx  valid point,  two things here:
- we have adaptor functions in MSC to work with both `msc://` and normal file path starting with `/...`, example [here](https://github.com/NVIDIA/multi-storage-client/blob/main/src/multistorageclient/contrib/numpy.py#L33), so it shouldn't be a problem here (MSC adaptors should just pass through normal file path to original function, so overhead is expected to be minimal)
- Although, it's also reasonable to let users have the ability to explicitly enable MSC due to various reasons (less familiarity/confidence in a new dependency, etc) so that MSC dep introduced in some other places won't affect SFT training implicitly.  It can be either via YAML config or via environment variables.  Is this aligned with what you are thinking?

cuichenx:
> > Looks good, just one question about MSC: if `multistorageclient` is a package in the container (so that `MULTISTORAGECLIENT_AVAILABLE =True`, but the environment is a normal file system, is there any impact to run multistorageclient functions
> 
> @cuichenx valid point, two things here:
> 
> * we have adaptor functions in MSC to work with both `msc://` and normal file path starting with `/...`, example [here](https://github.com/NVIDIA/multi-storage-client/blob/main/src/multistorageclient/contrib/numpy.py#L33), so it shouldn't be a problem here (MSC adaptors should just pass through normal file path to original function, so overhead is expected to be minimal)
> * Although, it's also reasonable to let users have the ability to explicitly enable MSC due to various reasons (less familiarity/confidence in a new dependency, etc) so that MSC dep introduced in some other places won't affect SFT training implicitly.  It can be either via YAML config or via environment variables.  Is this aligned with what you are thinking?

Ah, having the first bullet point is perfect. (I realized you also put this in the description but I missed that earlier). 
If going through MSC adaptors has minimal risk or overhead, then I think this is fine. I don't think there is a need to add a config argument to allow user to enable/disable MSC. If they really wanted to they can easily modify the code to achieve that.

cuichenx (approved):
LGTM

PR #12887, opened 2025-04-04
Support huggingface export to tensorrtllm

pthombre:
Export huggingface models to trtllm so that they can be deployed to triton for inference

### DIFF
--- a/scripts/deploy/nlp/deploy_triton.py
+++ b/scripts/deploy/nlp/deploy_triton.py
@@ -18,10 +18,10 @@ import logging
 import os
 import sys
 from pathlib import Path
-from transformers import AutoModelForCausalLM, AutoConfig, AutoTokenizer
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


scripts/deploy/nlp/deploy_triton.py
```diff
@@ -52,6 +53,8 @@ def get_args(argv):
         description=f"Deploy nemo models to Triton",
     )
     parser.add_argument("-nc", "--nemo_checkpoint", type=str, help="Source .nemo file")
+    parser.add_argument("-hfp", "--hf_model_path", type=str, help="Huggingface model path")
```
oyilmaz-nvidia, line 56: Can we use only one parameter for the hf model like here? https://github.com/NVIDIA/NeMo/blob/onur/hf-inframework-deploy/scripts/deploy/nlp/deploy_inframework_hf_triton.py#L67

PR #12889, opened 2025-04-04
Support huggingface export to tensorrtllm

pthombre:
Support for exporting huggingface models to trtllm, and deploying it on a triton server


oyilmaz-nvidia (dismissed):

chtruong814 (approved):
Approved per @oyilmaz-nvidia 

The CI tests passed. The final call to upload code coverage failed because it included a `-branch` param which is not consistent with how other tests are invoked.

PR #12902, opened 2025-04-07
Use nvidia-lm-eval for evaluation

marta-sd:
> [!IMPORTANT]  

# What does this PR do ?

This PR replaces upstream lm-eval with a NVIDIA Evals Factory version: https://pypi.org/project/nvidia-lm-eval/

**Collection**: llm

# Changelog

- Updates evaluation code to use NVIDIA Evals Factory
- Sets new method as default evaluation path in `nemo/collections/llm/api.py`
- Updates docs and `scripts/llm/evaluation.py` to use the new method

# Usage

Deploy the model:

```python
from nemo.collections.llm import deploy

if __name__ == "__main__":
    deploy(
            nemo_checkpoint='/workspace/hf_llama3_8b_nemo2.nemo',
            max_input_len=4096,
            max_batch_size=4,
            num_gpus=1,)
```
...and launch evaluation:
```python
from nemo.collections.llm import evaluate
from nemo.collections.llm.evaluation.api import EvaluationConfig, ApiEndpoint, EvaluationTarget, ConfigParams

api_endpoint = ApiEndpoint()
eval_target = EvaluationTarget(api_endpoint=api_endpoint)
eval_params = ConfigParams(top_p=1, temperature=1, limit_samples=2, parallelism=4)
eval_config = EvaluationConfig(type='mmlu', params=eval_params)

if __name__ == "__main__":
    evaluate(target_cfg=eval_target, eval_cfg=eval_config)
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/docs/source/evaluation/evaluation_doc.rst
+++ b/docs/source/evaluation/evaluation_doc.rst
@@ -1,15 +1,33 @@
 Evaluate NeMo 2.0 Checkpoints
 ==============================
 
-This guide provides detailed instructions on evaluating NeMo 2.0 checkpoints using the integrated `lm-evaluation-harness
-<https://github.com/EleutherAI/lm-evaluation-harness>`__ within the NeMo Framework. Supported benchmarks include
-``MMLU``, ``GSM8k``, ``lambada_openai``, ``winogrande``, ``arc_challenge``, ``arc_easy``, and ``copa``.
+This guide provides detailed instructions on evaluating NeMo 2.0 checkpoints using the integrated `nvidia-lm-eval
+<https://pypi.org/project/nvidia-lm-eval/>`__ within the NeMo Framework. Supported benchmarks include:
+* Arc-challenge,
+* BBH,
+* GPQA,
+* GSM8K,
+* Hellaswag,
+* IFEval,
+* MGSM,
+* MMLU,
+* MMLU-Pro,
+* MUSR,
+* TruthfulQA,
+* Wikilingua, and
+* Winogrande.
+
 
 Introduction
 --------------
 The evaluation process employs a server-client approach, comprising two main phases. In Phase 1, the NeMo 2.0
 checkpoint is deployed on a PyTriton server by exporting it to TRT-LLM. Phase 2 involves running the evaluation
 on the model using the deployed URL and port.
+
+Some of the benchmarks (e.g. GPQA) use a gated dataset. In order to use them, you must authenticate to 
+`HuggingFace Hub <https://huggingface.co/docs/huggingface_hub/quick-start#authentication>`__
+before launching the evaluation.
+
 
 Run Evaluations without NeMo-Run
 ---------------------------------
@@ -42,14 +60,13 @@ the processes from being killed and aborting the runs.
     from nemo.collections.llm import evaluate
     from nemo.collections.llm.evaluation.api import EvaluationConfig, ApiEndpoint, EvaluationTarget, ConfigParams
 
-    nemo_checkpoint = '/workspace/hf_llama3_8b_nemo2.nemo/'
-    api_endpoint = ApiEndpoint(nemo_checkpoint_path=nemo_checkpoint)
+    api_endpoint = ApiEndpoint()
     eval_target = EvaluationTarget(api_endpoint=api_endpoint)
-    eval_params = ConfigParams(top_p=1, temperature=1, top_k=1, limit_samples=2, num_fewshot=5)
+    eval_params = ConfigParams(top_p=1, temperature=1, limit_samples=2, parallelism=4)
     eval_config = EvaluationConfig(type='mmlu', params=eval_params)
 
     if __name__ == "__main__":
-    evaluate(target_cfg=eval_target, eval_cfg=eval_config)
+        evaluate(target_cfg=eval_target, eval_cfg=eval_config)
 
 .. note::
 
@@ -97,7 +114,38 @@ Below is an example command:
 .. code-block:: bash
 
     python scripts/llm/evaluation.py --nemo_checkpoint='/workspace/hf_llama3_8b_nemo2.nemo' --slurm --nodes 1 
-    --devices 8 --container_image "nvcr.io/nvidia/nemo:25.02" --tensor_parallelism_size 8
+    --devices 8 --container_image "nvcr.io/nvidia/nemo:25.04" --tensor_parallelism_size 8
 
 By following these commands, you can successfully run evaluations using NeMo-Run on both local and Slurm-based
 environments.
+
+
+
+Run Legacy Evaluations with `lm-evaluation-harness <https://github.com/EleutherAI/lm-evaluation-harness>`__
+-----------------------------------------------------------------------------------------------------------
+
+You can also run evaluations of NeMo 2.0 checkpoints using the integrated `lm-evaluation-harness
+<https://github.com/EleutherAI/lm-evaluation-harness>`__ within the NeMo Framework. Supported benchmarks include
+``MMLU``, ``GSM8k``, ``lambada_openai``, ``winogrande``, ``arc_challenge``, ``arc_easy``, and ``copa``.
+Please note that this path is deprecated and will be removed in NeMo 25.06 release.
+
+
+The ``evaluate`` method defined in ``nemo/collections/llm/api.py`` supports the legacy way of evaluating the models.
+To run evaluations on the deployed model, use the following command. Make sure to pass `nemo_checkpoint_path` and
+`url` parameters as they are needed to use the legacy evaluation code. Make sure to open a new terminal within the
+same container to execute it. For longer evaluations, it is advisable to run both the deploy and evaluate commands
+in tmux sessions to prevent the processes from being killed and aborting the runs.
+
+.. code-block:: python
+
+    from nemo.collections.llm import evaluate
+    from nemo.collections.llm.evaluation.api import EvaluationConfig, ApiEndpoint, EvaluationTarget, ConfigParams
+
+    nemo_checkpoint = '/workspace/hf_llama3_8b_nemo2.nemo/'
+    api_endpoint = ApiEndpoint(nemo_checkpoint_path=nemo_checkpoint, url="http://0.0.0.0:8000")
+    eval_target = EvaluationTarget(api_endpoint=api_endpoint)
+    eval_params = ConfigParams(top_p=1, temperature=1, top_k=1, limit_samples=2, num_fewshot=5)
+    eval_config = EvaluationConfig(type='mmlu', params=eval_params)
+
+    if __name__ == "__main__":
+        evaluate(target_cfg=eval_target, eval_cfg=eval_config)
--- a/docs/source/evaluation/evaluation_doc.rst
+++ b/docs/source/evaluation/evaluation_doc.rst
@@ -1,7 +1,7 @@
 Evaluate NeMo 2.0 Checkpoints
 ==============================
 
-This guide provides detailed instructions on evaluating NeMo 2.0 checkpoints using the integrated `nvidia-lm-eval
+This guide provides detailed instructions on evaluating NeMo 2.0 checkpoints using the `NVIDIA Evals Factory
 <https://pypi.org/project/nvidia-lm-eval/>`__ within the NeMo Framework. Supported benchmarks include:
     * GPQA,
     * GSM8K,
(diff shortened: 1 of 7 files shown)


docs/source/evaluation/evaluation_doc.rst
```diff
@@ -1,16 +1,30 @@
...
+    * MMLU-Pro,
+    * MUSR, and
+    * Wikilingua.
+
 
 Introduction
 --------------
 The evaluation process employs a server-client approach, comprising two main phases. In Phase 1, the NeMo 2.0
 checkpoint is deployed on a PyTriton server by exporting it to TRT-LLM. Phase 2 involves running the evaluation
 on the model using the deployed URL and port.
```
athitten, line 21: Can we move this to legacy evaluation section ? And add the below here:

``` Python
The evaluation process employs a server-client approach, comprising two main phases. In Phase 1, the NeMo 2.0
checkpoint is deployed in-framework on a PyTriton server by exposing OpenAI API (OAI) compatible endppoints. Phase 2 involves running the evaluation
on the model using the OAI endpoint and port.
```

docs/source/evaluation/evaluation_doc.rst
```diff
@@ -1,16 +1,30 @@
...
+    * BBH,
+    * GPQA,
+    * GSM8K,
+    * IFEval,
+    * MGSM,
+    * MMLU,
+    * MMLU-Pro,
+    * MUSR, and
+    * Wikilingua.
+
```
athitten, line 15: I believe we should be able to run all chat benchmarks, and completions benchmarks that don't need logprobs in this version right ? Does this list cover all of them ?

nemo/collections/llm/api.py
```diff
@@ -752,9 +754,9 @@ def deploy(
         nm.stop()
 
 
-def evaluate(
+def _legacy_evaluate(
```
athitten, line 757: Can we move this method to `nemo/collections/llm/evaluation/base.py` ? Want to have all the util/helper method in the evaluation folder's `base.py` and keep `api.py` only for main methods. Thanks!

nemo/collections/llm/evaluation/api.py
```diff
@@ -59,22 +62,39 @@ class ConfigParams(BaseModel):
         description="Limit evaluation to `limit` samples. Default: use all samples", default=None
     )
     num_fewshot: Optional[int] = Field(
-        description="Number of examples in few-shot context. Default: None which means no few_shots are used.",
+        description="Number of examples in few-shot context. Default: None.",
         default=None,
+        deprecated="This parameter is deprecated and not used for evaluations with core_evals.",
```
athitten, line 67: How is `num_fewshot` passed with core-evals then ? I thought it would be part of `ConfigParams`.

nemo/collections/llm/evaluation/api.py
```diff
@@ -59,22 +62,39 @@ class ConfigParams(BaseModel):
...
+    )
+    task: Optional[str] = Field(description="Name of the task", default=None)
+    request_timeout: Optional[int] = Field(description="REST response timeout", default=300)
+    extra: Optional[Dict[str, Any]] = Field(
+        description="Framework specific parameters to be used for evaluation", default_factory=dict
+    )
+    batch_size: Optional[int] = Field(
+        description="batch size to use for evaluation",
+        default=1,
+        deprecated="This parameter is deprecated and not used for evaluations with core_evals.",
```
athitten, line 82: Should we replace all `core_evals` usage with `nvidia-lm-eval` to avoid confusion. Since its published as `nvidia-lm-eval`.

nemo/collections/llm/api.py
```diff
@@ -813,6 +817,64 @@ def evaluate(
...
+        raise RuntimeError("Server not ready for evaluation")
+
+    results = evaluate.evaluate_accuracy(
+        target_cfg=target_cfg,
+        eval_cfg=eval_cfg,
+    )
+    results_dict = results.model_dump()
+
+    print("========== RESULTS ==========")
+    print(yaml.dump(results_dict))
```
athitten, line 873: We want to avoid using print statements directly. Can we replace `print` with logging.info like [here](https://github.com/NVIDIA/NeMo/blob/main/nemo/deploy/service/fastapi_interface_to_pytriton.py#L162)

nemo/collections/llm/evaluation/api.py
```diff
@@ -59,22 +62,39 @@ class ConfigParams(BaseModel):
...
     )
-    max_new_tokens: Optional[int] = Field(description="max tokens to generate", default=256)
-    batch_size: Optional[int] = Field(description="batch size to use for evaluation", default=1)
     top_k: Optional[int] = Field(
         description="Limits to a certain number (K) of the top tokens to consider",
         default=1,
+        deprecated="This parameter is deprecated and not used for evaluations with core_evals.",
```
athitten, line 87: Don't we want to have `top_k` as a user configurable param ?

nemo/collections/llm/evaluation/api.py
```diff
@@ -59,22 +62,39 @@ class ConfigParams(BaseModel):
...
     )
     add_bos: Optional[bool] = Field(
         description="whether a special bos token should be added when encoding a string",
         default=False,
+        deprecated="This parameter is deprecated and not used for evaluations with core_evals.",
     )
     bootstrap_iters: int = Field(
         description="Number of iterations for bootstrap statistics",
         default=100000,
+        deprecated="This parameter is deprecated and not used for evaluations with core_evals.",
```
athitten, line 97: Same question as above. This should still be relevant and end user configurable right ?

athitten (commented):
Thank you @marta-sd ! Left some comments.

nemo/deploy/service/fastapi_interface_to_pytriton.py
```diff
@@ -77,8 +77,8 @@ class CompletionRequest(BaseModel):
     messages: list[dict] = [{}]
     max_tokens: int = 512
     temperature: float = 1.0
-    top_p: float = 0.0
-    top_k: int = 1
+    top_p: float = 1.0
+    top_k: int = 0
```
athitten, line 81: Shouldn't the default `top_k` be 1 ? So that single most probable token is chosen at each step.

nemo/collections/llm/evaluation/api.py
marta-sd: This parameter is specific for lm-eval-harness and not available for other evaluation frameworks. Framework-specific params are passed through `extra`. I'll add this information to the description of both params.

nemo/collections/llm/evaluation/api.py
marta-sd: You're right but I wonder if we should use more generic name - now we only have nvidia-lm-eval, but we expect to have more frameworks available soon. They will all come from `NVIDIA Evals Factory`, so maybe I'll use this term instead?

nemo/collections/llm/evaluation/api.py
marta-sd: It's not used - `top_k` is not available in [OpenAI API](https://platform.openai.com/docs/api-reference/chat/create) so benchmarks don't pass it to the server

nemo/collections/llm/evaluation/api.py
marta-sd: This is specific to lm-eval-harness, other eval frameworks don't have such param / call it differently. From what I see it's not exposed to the user in Evals Factory

nemo/deploy/service/fastapi_interface_to_pytriton.py
marta-sd: I thought so, but if I specify top_p (which is send in the request by benchmarks and present in OpenAI API), I'm getting an error that top_p and top_k can't be non-zero at the same time

docs/source/evaluation/evaluation_doc.rst
marta-sd: I updated the list and tasks in `evaluation.py` script

docs/source/evaluation/evaluation_doc.rst
```diff
@@ -97,7 +107,41 @@ Below is an example command:
...
+
+Run Legacy Evaluations with `lm-evaluation-harness <https://github.com/EleutherAI/lm-evaluation-harness>`__
+-----------------------------------------------------------------------------------------------------------
+
+You can also run evaluations of NeMo 2.0 checkpoints using the integrated `lm-evaluation-harness
+<https://github.com/EleutherAI/lm-evaluation-harness>`__ within the NeMo Framework. Supported benchmarks include
+``MMLU``, ``GSM8k``, ``lambada_openai``, ``winogrande``, ``arc_challenge``, ``arc_easy``, and ``copa``.
+Please note that this path is deprecated and will be removed in NeMo 25.06 release.
+
```
athitten, line 124: Can copy paste the deployment section for before here, but the code looks like this. Just adds `backend="trtllm"` arg.

```Python
from nemo.collections.llm import deploy

if __name__ == "__main__":
        deploy(
         nemo_checkpoint='/workspace/nemo-llama-3-8B',
         max_batch_size=4,
         num_gpus=1,
         tensor_parallelism_size=1,
         backend="trtllm")
```

docs/source/evaluation/evaluation_doc.rst
athitten: Looks like the snippet shared above isnt added under introduction, just FYI to add it in case it got missed.

docs/source/evaluation/evaluation_doc.rst
athitten: Thanks Marta, I think it will be good if we can also mention/distinguish in the docs which of them are completion and chat completion benchmarks as docs are the main entry point for the users.

nemo/collections/llm/evaluation/api.py
athitten: Yeah NVIDIA Evals Factory sounds okay. Can we add that nvidia-lm-eval is from NVIDIA Evals Factory  [here](https://github.com/NVIDIA/NeMo/pull/12902/files#diff-a6c3603d5a0520f521eabf29ff6386e852847f80c93ddfd7e66998d4b2521126R4) so that there is context ?

nemo/collections/llm/evaluation/api.py
athitten: I see, thanks! Do you think we should make it user configurable on the server side then ?

nemo/deploy/service/fastapi_interface_to_pytriton.py
athitten: Can we make the default as greedy decoding (i.e top_k=1, top_p=0) and add a check in triton_infer_fn as discussed on slack ?

marta-sd: I ended up setting defaults here to None, this way we know inside triton_infer_fn if the param was provided by the user or not. See https://github.com/NVIDIA/NeMo/pull/12902/commits/6e69e6c1909a4bab9c8418e9412e053716bc62e3

docs/source/evaluation/evaluation_doc.rst
marta-sd: Sorry, I somehow missed that. Thanks for pointing it out!

nemo/collections/llm/evaluation/api.py
marta-sd: I think they should be configurable.

First of all, benchmarks send them so removing top_p and temperature could lead to unexpected behavior. Secondly, for some benchmarks it makes sense to generate multiple responses in a non-greedy manner and calculate pass@k (often used for coding benchmarks). Also, it can happen that recommended params for a given model are different (recently we had a case of reasoning model that required different setting)

athitten: Sorry I am a bit confused wrt to these params: `top_p`, `top_k`, `temp`. So all of these are sent in the request to the server and like you said it should be user configurable. In that case we shouldn't be deprecating `top_k` [here](https://github.com/NVIDIA/NeMo/pull/12902/files#diff-39a14475dee052cbe5b6b94e9a3168894e284ab298745abaf60467820c3ab243R88) right ? since this is the way user would modify `top_k` if they want to try with a different value ?

nemo/deploy/nlp/megatronllm_deployable.py
```diff
@@ -310,19 +310,24 @@ def triton_infer_fn(self, **inputs: np.ndarray):
...
-        top_k = inputs.pop("top_k")[0][0] if "top_k" in inputs else 1
-        top_p = inputs.pop("top_p")[0][0] if "top_p" in inputs else 0.0
+        if ("top_k" not in inputs) and ("top_p" not in inputs):
+            # if neither top_k nor top_p are provided -> default to greedy decoding
+            top_k = 1
+            top_p = 0.0
+        else:
+            # otherwise use the provided params
+            top_k = inputs.pop("top_k")[0][0] if "top_k" in inputs else 0
+            top_p = inputs.pop("top_p")[0][0] if "top_p" in inputs else 0.0
```
athitten, line 320: @marta-sd how are we fixing the  `Cannot have top-p and top-k both greater than zero error` error with this change ? We still hit this error if top_p=0.9999 and top_k=1. Lmk if I am missing something

athitten: Oh sorry I saw your message offline that you intended for the code to error out.

nemo/collections/llm/evaluation/api.py
marta-sd: The reason is that top_k not part of OAI API and not supported by any evals. Only top_p and temperature are

nemo/deploy/nlp/megatronllm_deployable.py
marta-sd: Also as discussed offline I'll revert my changes in this method and set top_p=0.0 and top_k=0 as defaults in FastAPI.

nemo/deploy/service/fastapi_interface_to_pytriton.py
marta-sd: As discussed offline I'll set defaults here to top_p=0.0 and top_k=0 and revert changes in triton_infer_fn

athitten (approved):
Thank you so much @marta-sd, LGTM!

docs/source/evaluation/evaluation_doc.rst
```diff
@@ -97,7 +128,57 @@ Below is an example command:
...
     python scripts/llm/evaluation.py --nemo_checkpoint='/workspace/hf_llama3_8b_nemo2.nemo' --slurm --nodes 1 
-    --devices 8 --container_image "nvcr.io/nvidia/nemo:25.02" --tensor_parallelism_size 8
+    --devices 8 --container_image "nvcr.io/nvidia/nemo:25.04" --tensor_parallelism_size 8
 
 By following these commands, you can successfully run evaluations using NeMo-Run on both local and Slurm-based
 environments.
+
+
+
+Run Legacy Evaluations with `lm-evaluation-harness <https://github.com/EleutherAI/lm-evaluation-harness>`__
```
jgerh, line 138: remove link from heading, use plain text here

Run Legacy Evaluations with lm-evaluation-harness

jgerh (commented):
Completed review of docs/source/evaluation/evaluation_doc.rst and provided copyedits and some text revisions.

PR #12905, opened 2025-04-07
WebDataset URL refactoring

monica-sekoyan:
> [!IMPORTANT]  

# What does this PR do ?

Removes the part where there is a mandatory conversion to webdataset url format in `expand_sharded_filepaths` (`pipe: ais get...`). This was a bottleneck for the LhotseDataloader usage. As it only concern webdataset and only tar files!, make this conversion only for these cases in the code.

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/utils/data_utils.py
+++ b/nemo/utils/data_utils.py
@@ -144,6 +144,10 @@ def ais_binary() -> str:
     else:
         raise RuntimeError('AIS binary not found.')
 
+def check_ais_binary_installed() -> bool:
+    """Check if ais binary is installed."""
+    return ais_binary() is not None
+
 
 def datastore_path_to_local_path(store_path: str) -> str:
     """Convert a data store path to a path in a local cache.
@@ -295,6 +299,7 @@ def datastore_path_to_webdataset_url(store_path: str):
         URL which can be directly used with WebDataset.
     """
     if is_datastore_path(store_path):
+        check_ais_binary_installed()
         url = f'pipe:ais get {store_path} - || true'
     else:
         raise ValueError(f'Unknown store path format: {store_path}')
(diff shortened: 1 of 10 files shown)


nemo/utils/data_utils.py
```diff
@@ -295,6 +300,7 @@ def datastore_path_to_webdataset_url(store_path: str):
         URL which can be directly used with WebDataset.
     """
     if is_datastore_path(store_path):
+        check_ais_binary_installed()
```
pzelasko, line 303: the result of this check is ignored right now

pzelasko (changes requested):
I don't think this is the way to go. IMO we should just drop the `pipe:ais get {path} - || true` thing and simply return `{path}` since we now directly support AIStore via Python client. This way we can avoid all the changes introduced in this PR

monica-sekoyan:
> I don't think this is the way to go. IMO we should just drop the `pipe:ais get {path} - || true` thing and simply return `{path}` since we now directly support AIStore via Python client. This way we can avoid all the changes introduced in this PR

@pzelasko but in this way webdataset datapipeline won't work with the tar files coming from AIStore

nithinraok:
its changing lot of files unneccesarily. use skipdocs label for this PR

pzelasko (approved):
Thanks @monica-sekoyan! My only concern is that every dataset class here has to manually monkey-patch WDS, maybe it would be a bit more elegant if we had a `import nemo.utils.webdataset as wds` which monkey-patches it once and returns the original `wds` namespace. It can probably be implemented as:

`nemo/utils/webdataset.py`
```python
from webdataset import *
from nemo.utils.data_utils import wds_lhotse_url_opener
tariterators.url_opener = wds_lhotse_url_opener
```

But I'm not 100% sure -- would need some testing.

pzelasko (approved):

ko3n1g:
We need to revert this since it doesn’t provide import guards against lothse for collections like nlp.

PR #13103, opened 2025-04-18
[automodel]Set GA based on other parallel params.

yuanzhedong:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/llm/sft/automodel.py
+++ b/examples/llm/sft/automodel.py
@@ -188,7 +188,11 @@ def main():
     parser.add_argument('--use-te-optimizer', action='store_true', help='Use TE optimizer')
     parser.add_argument('--grad-clip', type=float, default=1.0, help='Grad clip value')
     parser.add_argument(
-        '--accumulate-grad-batches', '--accumulate_grad_batches', type=int, default=1, help='Number of batches to accumulate gradient over.'
+        '--accumulate-grad-batches',
+        '--accumulate_grad_batches',
+        type=int,
+        default=1,
+        help='Number of batches to accumulate gradient over.',
     )
     parser.add_argument('--max-steps', type=int, default=100, help='Maximum number of training steps')
     parser.add_argument('--log-every-n-steps', type=int, default=1, help='Log every n steps')
@@ -279,9 +283,7 @@ def main():
         # Faster convergence but may lead to memory issues
         optimizer = fdl.build(llm.adam.te_adam_with_flat_lr(lr=args.lr))
     else:
-        optimizer = fdl.build(
-            llm.adam.pytorch_adam_with_flat_lr(lr=args.lr)
-        )  # foreach need to be False for TP
+        optimizer = fdl.build(llm.adam.pytorch_adam_with_flat_lr(lr=args.lr))  # foreach need to be False for TP
 
     if args.fp8:
         from nemo.lightning.pytorch.accelerate.transformer_engine import TEConfig
--- a/examples/llm/sft/automodel.py
+++ b/examples/llm/sft/automodel.py
@@ -188,7 +188,7 @@ def main():
     parser.add_argument('--use-te-optimizer', action='store_true', help='Use TE optimizer')
     parser.add_argument('--grad-clip', type=float, default=1.0, help='Grad clip value')
     parser.add_argument(
-        '--accumulate-grad-batches', type=int, default=1, help='Number of batches to accumulate gradient over.'
+        '--accumulate-grad-batches', '--accumulate_grad_batches', type=int, default=1, help='Number of batches to accumulate gradient over.'
     )
     parser.add_argument('--max-steps', type=int, default=100, help='Maximum number of training steps')
     parser.add_argument('--log-every-n-steps', type=int, default=1, help='Log every n steps')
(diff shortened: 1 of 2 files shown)


examples/llm/sft/automodel.py
```diff
@@ -244,6 +244,22 @@ def main():
...
         args.use_te_optimizer = False
 
+    try:
+        args.accumulate_grad_batches = calculate_valid_accumulate_grad_batches(
+            global_batch_size=args.global_batch_size,
+            micro_batch_size=args.micro_batch_size,
+            devices=args.devices,
+            num_nodes=args.num_nodes,
+            tp_size=args.tp_size,
+            pp_size=1,  # Default to 1 as it's not in args
```
akoumpa, line 254: Hi, right now we don't have support for PP, so I would drop it for now and introduce it in the future as necessary.

examples/llm/sft/automodel.py
```diff
@@ -188,7 +188,7 @@ def main():
     parser.add_argument('--use-te-optimizer', action='store_true', help='Use TE optimizer')
     parser.add_argument('--grad-clip', type=float, default=1.0, help='Grad clip value')
     parser.add_argument(
-        '--accumulate_grad_batches', type=int, default=1, help='Number of batches to accumulate gradient over.'
+        '--accumulate-grad-batches', type=int, default=1, help='Number of batches to accumulate gradient over.'
```
akoumpa, line 191: Thank for this change I wanted it to make it myself, now since the underscore version is already used, I would add an alias 

```
    parser.add_argument(
        '--accumulate_grad_batches',  '--accumulate-grad-batches',  type=int, default=1, help='Number of batches to accumulate gradient over.'
```
So that we don't break anyones work that already uses the underscore version, maybe I'm being too careful.

nemo/automodel/misc_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    Returns:
+        int: The calculated gradient accumulation steps
+
+    Raises:
+        ValueError: If the parameters result in invalid configuration
+    """
+    # Validate inputs
+    if not all(
+        isinstance(x, int)
+        for x in [global_batch_size, micro_batch_size, devices, num_nodes, tp_size, pp_size, cp_size]
```
akoumpa, line 46: I'd leave this to argparse's type=int, WDYT?

examples/llm/sft/automodel.py
athitten: Curious if we still need this arg accumulate-grad-batches if we are calculating the correct value and overriding it ?

examples/llm/sft/automodel.py
```diff
@@ -244,6 +244,22 @@ def main():
...
+            num_nodes=args.num_nodes,
+            tp_size=args.tp_size,
+            pp_size=1,  # Default to 1 as it's not in args
+            cp_size=args.cp_size,
+        )
+    except ValueError as e:
+        print(f"Error calculating gradient accumulation steps: {e}")
+        print("Using default value of 1 for accumulate_grad_batches")
+        args.accumulate_grad_batches = 1
+    print(f"Accumulate grad batches: {args.accumulate_grad_batches}")
```
athitten, line 261: Can we use logging.info instead of using print directly ?

athitten: I think this script also has 1-2 other places where print is being used directly, probably good to replace them as well

examples/llm/sft/automodel.py
yuanzhedong: Yeah I was thinking the other option is to delete it the exposure of GA if we are going to calculate it and override. Or we keep it and give the freedom to the users and rely on the users to give the right value, what do you guys think?

examples/llm/sft/automodel.py
yuanzhedong: yeah I'm using print since I saw there're other places doing it, we can do the replacing in separate PR.

examples/llm/sft/automodel.py
yuanzhedong: sure I will remove pp

nemo/automodel/misc_utils.py
yuanzhedong: sure I will remove the type check here if that's what you mean.

examples/llm/sft/automodel.py
athitten: We should definitely calculate the correct GA and override it. They did this in case of nemo1, megatron automatically calculated it: [megatron_gpt_config.yaml#L21](https://github.com/NVIDIA/NeMo/blob/v2.2.0/examples/nlp/language_modeling/conf/megatron_gpt_config.yaml#L21). 
I think we can keep the arg for now with the default of 1 and maybe remove it in the next version or keep it and add description that its automatically calculate like it was mentioned in  megatron_gpt_config.yaml ?

yuanzhedong: sounds good, we can keep both and revisit this later.

akoumpa (approved):
LGTM, thanks.

PR #12773, opened 2025-03-25
Use first_value decorator for triton_infer_fn plus minor changes

janekl:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

* Using [PyTriton first_value](https://triton-inference-server.github.io/pytriton/0.5.0/reference/decorators/#pytriton.decorators.first_value) decorator to easily access repeated values in requests
* Aligning the length of "erroneous" response `[err_msg]` with the prompts list length to avoid PyTriton shape mismatch errors
* Adding `--legacy_ckpt` flag for in-framework deployment with older checkpoints
* Several minor changes

**Collection**: NLP

# Changelog

- Add specific line by line info of high level changes in this PR.

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information

- Related to # (issue)

PR #13177, opened 2025-04-22
Support TRTLLM Pytorch backend Deployment

meatybobby:
> [!IMPORTANT]  

# What does this PR do ?

Support TRTLLM Pytorch backend Deployment and minor changes for query

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,221 @@
...
+    def get_triton_output(self):
+        return (Tensor(name="sentences", shape=(-1,), dtype=bytes),)
+
+    @batch
+    def triton_infer_fn(self, **inputs: np.ndarray):
+        output_infer = {}
+
+        try:
+            prompts = str_ndarray2list(inputs.pop("prompts"))
+            temperature = inputs.pop("temperature")[0][0] if "temperature" in inputs else None
```
janekl, line 188: Could you please use `first_value` decorator here for consistency/simplicity? See https://triton-inference-server.github.io/pytriton/0.5.14/inference_callables/decorators/#first-value and the recent update in https://github.com/NVIDIA/NeMo/pull/12773

scripts/deploy/nlp/query_trtllm_pytorch.py
```diff
@@ -0,0 +1,98 @@
...
+        top_k=args.top_k,
+        top_p=args.top_p,
+        temperature=args.temperature,
+        init_timeout=args.init_timeout,
+    )
+    print(outputs)
+
+
+if __name__ == '__main__':
+    query(sys.argv[1:])
```
janekl, line 98: NIT no need to pass `sys.argv[1:]` around, `query()` should work (after removing `argv` from `get_args` etc.)

nemo/deploy/nlp/query_llm.py
```diff
@@ -527,3 +527,85 @@ def query_llm_streaming(
                     yield sentences
                 else:
                     yield partial_result_dict["outputs"]
+
+
+class NemoQueryTRTLLMPytorch(NemoQueryLLMBase):
```
janekl, line 532: All the classes here in this file -- `NemoQueryTRTLLMPytorch`, `NemoQueryLLM`, etc -- are very similar.

Maybe we should work out a way to unify them. CC @oyilmaz-nvidia 

Besides, I think we don't need the constructor here. Let's remove it -- perhaps everywhere here -- ok?

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,221 @@
...
+            attn_backend=attn_backend,
+            enable_overlap_scheduler=enable_overlap_scheduler,
+            torch_compile_enabled=torch_compile_enabled,
+            use_cuda_graph=use_cuda_graph,
+        )
+
+        self.model = LLM(
+            model=hf_model_id_path,
+            tokenizer=hf_model_id_path if tokenizer is None else tokenizer,
+            tensor_parallel_size=tensor_parallel_size,
```
janekl, line 89: Suggestion: how about passing all arguments of kind `param=param` via `kwargs` here? We could just link `LLM` docs from TRT-LLM instead on available parameters.

Pros are shorter code, no need to update params on renames or future extensions.

On the cons side maybe readability, though things like `tensor_parallel_size` or `dtype` should be clear

scripts/deploy/nlp/query_trtllm_pytorch.py
```diff
@@ -0,0 +1,98 @@
...
+    parser.add_argument("-tk", "--top_k", type=int, help="top_k")
+    parser.add_argument("-tpp", "--top_p", type=float, help="top_p")
+    parser.add_argument("-t", "--temperature", type=float, help="temperature")
+    parser.add_argument("-it", "--init_timeout", default=60.0, type=float, help="init timeout for the triton server")
+
+    args = parser.parse_args(argv)
+    return args
+
+
+def query_llm(
```
janekl, line 41: Do we really need that helper function here? It is defined within a script here, so won't be importable anyway. I'd suggest to write directly in the main block

scripts/deploy/nlp/deploy_trtllm_pytorch_triton.py
```diff
@@ -0,0 +1,125 @@
...
+    except Exception as error:
+        LOGGER.error("Error message has occurred during deploy function. Error message: " + str(error))
+        return
+
+    LOGGER.info("Model serving will be stopped.")
+    nm.stop()
+
+
+if __name__ == '__main__':
+    trtllm_deploy(sys.argv[1:])
```
janekl, line 125: Same comment as in https://github.com/NVIDIA/NeMo/pull/13177/files#r2066576452

janekl:
I would be great to fill in the MR description with usage example as a form of documenting what happens here.

nemo/deploy/nlp/trtllm_pytorch_deployable.py
meatybobby: I think we still want to keep some parameters for readability

meatybobby:
@janekl Applied several changes. Could you review it again?

nemo/deploy/nlp/query_llm.py
janekl: WDYT @oyilmaz-nvidia?

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,212 @@
...
+        return (Tensor(name="sentences", shape=(-1,), dtype=bytes),)
+
+    @batch
+    @first_value("temperature", "top_k", "top_p", "max_length")
+    def triton_infer_fn(self, **inputs: np.ndarray):
+        output_infer = {}
+
+        try:
+            prompts = str_ndarray2list(inputs.pop("prompts"))
+            temperature = inputs.pop("temperature") if "temperature" in inputs else None
```
janekl, line 179: Sorry to bother you again but this can be just `inputs.pop("temperature", None)` (same for `top_k`, etc.)

scripts/deploy/nlp/query_trtllm_pytorch.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def get_args():
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description=f"Queries Triton server running a TensorRT-LLM PyTorch backend model",
+    )
+    parser.add_argument("-u", "--url", default="0.0.0.0", type=str, help="url for the triton server")
+    parser.add_argument("-mn", "--model_name", required=True, type=str, help="Name of the triton model")
+    prompt_group = parser.add_mutually_exclusive_group(required=True)
```
janekl, line 27: Nice 👍

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,209 @@
...
+    @property
+    def get_triton_input(self):
+        inputs = (
+            Tensor(name="prompts", shape=(-1,), dtype=bytes),
+            Tensor(name="max_length", shape=(-1,), dtype=np.int_, optional=True),
+            Tensor(name="max_batch_size", shape=(-1,), dtype=np.int_, optional=True),
+            Tensor(name="top_k", shape=(-1,), dtype=np.int_, optional=True),
+            Tensor(name="top_p", shape=(-1,), dtype=np.single, optional=True),
+            Tensor(name="temperature", shape=(-1,), dtype=np.single, optional=True),
+            Tensor(name="random_seed", shape=(-1,), dtype=np.int_, optional=True),
```
janekl, line 161: We don't use `"random_seed"` in `triton_infer_fn`.

Is it used anywhere at all?

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,209 @@
...
+        try:
+            prompts = str_ndarray2list(inputs.pop("prompts"))
+            temperature = inputs.pop("temperature", None)
+            top_k = int(inputs.pop("top_k")) if "top_k" in inputs else None
+            top_p = inputs.pop("top_p", None)
+            max_length = inputs.pop("max_length", 256)
+
+            if torch.distributed.is_initialized():
+                if torch.distributed.get_world_size() > 1:
+                    torch.distributed.broadcast(torch.tensor([0], dtype=torch.long, device="cuda"), src=0)
```
janekl, line 183: Could you please explain why we need this `torch.distributed.broadcast` call here? It works for a on-the-fly tensor that is not used at all (discarded at all ranks).

If the only purpose here is to intruduce a barrier, should be perhaps use explicitely `torch.distributed.barrier()`?

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,209 @@
...
+                top_k=top_k,
+                top_p=top_p,
+                max_length=max_length,
+            )
+
+            output_infer = {"sentences": cast_output(output, np.bytes_)}
+
+        except Exception as error:
+            err_msg = "An error occurred: {0}".format(str(error))
+            output_infer["sentences"] = cast_output([err_msg], np.bytes_)
```
janekl, line 207: This error handling is buggy. If the `try` block errors out and `len(prompts) > 1`, then Pytriton will complain about wrong-sized output: `1` vs `len(prompts)`. This clutters real issues a lot, making debugging really hard.

Suggestion:
* **recommended**: let's get rid of try / except blocks here: its intent is not clear anyway
* WAR: let's make sure to return correctly sized "sentences" in the `except` block (see [here](https://github.com/NVIDIA/NeMo/blob/9cc90fc7f44ea44738ee1ece1aee89458f4d70e8/nemo/export/tensorrt_llm.py#L1591) from https://github.com/NVIDIA/NeMo/pull/12773)

janekl:
We will need a functional test for this new feature. Please have a look at some examples from `tests/functional_tests` for inspirations.

nemo/deploy/nlp/trtllm_pytorch_deployable.py
meatybobby: Removed this since this should be always launched in single process.

nemo/deploy/nlp/trtllm_pytorch_deployable.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @batch
+    @first_value("temperature", "top_k", "top_p", "max_length")
+    def triton_infer_fn(self, **inputs: np.ndarray):
+        output_infer = {}
+
+        try:
+            prompts = str_ndarray2list(inputs.pop("prompts"))
+            temperature = inputs.pop("temperature", None)
+            top_k = int(inputs.pop("top_k")) if "top_k" in inputs else None
```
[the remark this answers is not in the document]
meatybobby: Actually on get_triton_input we defined int, so we will always get int here. Don't need to cast, removed the cast.

Laplasjan107:
* If I'm not mistaken, it should be possible to use `tensorrt_llm.llmapi.LLM` with `LLM(..., backend='pytorch')`, instead of `tensorrt_llm._torch.LLM`, right?
* Would it be possible to make this class a general wrapper of `LLM` for any backed? That would be particularly useful for deploying quantized HF models in trtllm

PR #13328, opened 2025-04-29
fix bug from prior commit (#13264)

blisc:
# What does this PR do ?

Push #13264 on r2.3.0.
Bugfix(tts): Move tuple call to correct place

**Collection**: tts

# Changelog

- Bugfix(tts): Move tuple call in dataset to correct place


ko3n1g (approved):
trivial change

PR #13395, opened 2025-05-02
Hugging face model deployment with Ray Serve

pthombre:
# What does this PR do ?
Set up framework for using Ray Serve for model deployments. Currently this only supports HuggingFace models

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [X ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


oyilmaz-nvidia (commented):
So, overall, this looks good, and thanks for following the existing design patter for deploying models with PyTriton.

The only question I have in my mind is that is there any common functions that you think we should force users to implement to deploy anything with Ray? For instance, you can ask users to implement completion function as an example, or anything you think it should be a must.

scripts/deploy/nlp/deploy_ray_hf.py
```diff
@@ -0,0 +1,181 @@
...
+import logging
+import signal
+import sys
+from typing import Optional
+
+from nemo.deploy.deploy_ray import DeployRay
+from nemo.deploy.nlp.hf_deployable_ray import HFRayDeployable
+
+LOGGER = logging.getLogger("NeMo")
+
```
oyilmaz-nvidia, line 25: Do you think it makes sense to merge the ray and pytriton based deployment in one script and make the deploy option an argument?

Same thing can be done for the query as well.

nemo/deploy/deploy_ray.py
```diff
@@ -0,0 +1,166 @@
...
+            # Try to connect to existing Ray cluster
+            ray.init(address=address, ignore_reinit_error=ignore_reinit_error, runtime_env=runtime_env)
+        except ConnectionError:
+            # If no cluster exists, start a local one
+            LOGGER.info("No existing Ray cluster found. Starting a local Ray cluster...")
+            ray.init(
+                num_cpus=num_cpus,
+                num_gpus=num_gpus,
+                include_dashboard=include_dashboard,
+                ignore_reinit_error=ignore_reinit_error,
```
athitten, line 117: Just curious: it sounds like we want to have `ignore_reinit_error` as False by default. Is it not the case ?

nemo/deploy/nlp/hf_deployable.py
```diff
@@ -255,65 +349,58 @@ def triton_infer_fn(self, **inputs: np.ndarray):
...
-                    for r in output["logits"]:
-                        lp = torch.tensor(r).cpu().detach().numpy()
-                        if len(lp) == 0:
-                            output_logits.append([0])
-                        else:
-                            output_logits.append(lp)
-                    output_infer["logits"] = np.array(output_logits).transpose(1, 0, 2)
-            else:
-                output_infer = {"sentences": cast_output(output, np.bytes_)}
+    def ray_infer_fn(self, inputs: Dict[Any, Any]):
```
athitten, line 367: @pthombre not sure if this out of scope for the PR but can we unify `ray_Infer_fn` and `triton_infer_fn` ? They share the same logic.

scripts/deploy/nlp/deploy_ray_hf.py
```diff
@@ -0,0 +1,189 @@
...
+    parser.add_argument(
+        "--num_gpus_per_replica",
+        type=float,
+        default=1,
+        help="Number of GPUs per model replica",
+    )
+    parser.add_argument(
+        "--num_cpus_per_replica",
+        type=float,
+        default=8,
```
athitten, line 113: Should we make this as 1 ?

athitten:
@pthombre sorry for the delay. Thank you for this work, this looks great! Left some comments. Few follow ups:

1) Do you see any perf difference between using Pytriton v/s Ray for MegatronLLMDeployable? Its probably unlikely, but don't want to exclude it also.
2) Also when we use multi instance or replicas where in the code is output from all replicas get collated ? Or does this not exist because each replica works as an independent module and sends the outputs asynchronously ?
3) If its not difficult, can we add some perf numbers on some small model by doing multi instance query v/s single instance ?

PR #13436, opened 2025-05-05
Migrate Hyena to Megatron inference_context.

cspades:
# What does this PR do ?

- Megatron inference engine has [deprecated `inference_params` in favor of `inference_context`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/inference/model_inference_wrappers/abstract_model_inference_wrapper.py#L160), and it breaks our Evo2 unit tests. This performs the migration so we can fast-forward NeMo to pick up Evo2 fixes.

**Collection**: `nemo.collections.llm.gpt.models.hyena(.megatron.hyena)`

# Changelog

- Simply change `inference_params` to `inference_context`, although it would be better to use the new dataclasses: [`StaticInferenceContext`](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/inference/contexts/static_context.py#L10) but currently we don't need that extra tooling.
- Also adds a dummy argument for `runtime_gather_output` which can be used in transformer layers, but we just use the default. For cleanliness, this could be propagated through the model layers that do use this argument. (Should I?)

# Additional Information

- Related to...
  - BioNeMo Version Bump: https://github.com/NVIDIA/bionemo-framework/pull/855
  - Evo2: https://github.com/NVIDIA/bionemo-framework/pull/798


jstjohn (approved):

marcromeyn (approved):

cspades:
@ko3n1g Oh shoot just saw this, yeah go ahead and merge, but also if there is an option, do not delete the branch, BioNeMo is currently pointing to this branch! 💀 

We'll update to point to main after it's merged!

PR #13437, opened 2025-05-05
Add CallbackGroup & Metadata factory function

PytLab:
> [!IMPORTANT]  

# What does this PR do ?

Add global callback group & metadata factory function for NeMo

Will keep updating
- asr
- multimodal
- nlp

# Changelog

- New `CallbackGroup` and `Callback` ABC in NeMo/nemo/lightning/pytorch/callbacks/callback_group.py

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/lightning/pytorch/callbacks/callback_group.py
+++ b/nemo/lightning/pytorch/callbacks/callback_group.py
@@ -1,0 +1,114 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Callable
+
+from lightning.pytorch.callbacks import Callback
+
+
+class CallbackGroup:
+    """A class for hosting a collection of callback objects.
+
+    It is used to execute callback functions of multiple callback objects with the same method name.
+    When callbackgroup.func(args) is executed, internally it loops through the objects in
+    self._callbacks and runs self._callbacks[0].func(args), self._callbacks[1].func(args), etc.
+    The method name and arguments should match.
+
+    Attributes:
+        _callbacks (list[Callback]): List of callback objects.
+    """
+    _instance = None
+
+    @classmethod
+    def get_instance(cls) -> 'CallbackGroup':
+        """Get the singleton instance of the CallbackGroup.
+        Args:
+            cls (CallbackGroup): The class of the CallbackGroup.
+        Returns:
+            CallbackGroup: The singleton instance of the CallbackGroup.
+        """
+        if cls._instance is None:
+            cls._instance = CallbackGroup()
+        return cls._instance
+
+    def __init__(self, callbacks: list[Callback] | None) -> None:
+        """Initializes the list of callback objects.
+
+        Args:
+            callbacks (list[Callback]): List of callbacks.
+        """
+        self._callbacks = callbacks or []
+
+    def __getattr__(self, method_name: str) -> Callable:
+        """Loops through the callback objects to call the corresponding callback function.
+
+        Args:
+            method_name (str): Callback method name.
+        """
+
+        def multi_callback_wrapper(*args, **kwargs) -> None:
+            for callback in self._callbacks:
+                assert hasattr(callback, method_name)
+                method = getattr(callback, method_name)
+                assert callable(method)
+                _ = method(*args, **kwargs)
+
+        return multi_callback_wrapper
+
+    @property
+    def callbacks(self):
+        """ Return callbacks in order.
+
+        Returns:
+            list: callback objects
+        """
+        return self._callbacks
+
+
+class Callback(Callback):
+    """The base class for all callbacks. It inherits the pytorch lightning callback so the callback can be also passed to PTL trainer to reuse.
+    Below list extra callback functions in NeMo.
+    """
+    def on_dataloader_init_start(self):
+        """Called at the start of the data loading."""
+
+    def on_dataloader_init_end(self):
+        """Called at the end of the data loading."""
+
+    def on_model_init_start(self):
+        """Called at the start of the model initialization."""
+
+    def on_model_init_end(self):
+        """Called at the end of the model initialization."""
+
+    def on_optimizer_init_start(self) -> None:
+        """Called at the beginning of optimizer initialization."""
+
+    def on_optimizer_init_end(self) -> None:
+        """Called at the end of optimizer initialization."""
+
+    def on_load_checkpoint_start(self) -> None:
+        """Called at the beginning of loading checkpoint."""
+
+    def on_load_checkpoint_end(self) -> None:
+        """Called at the end of loading checkpoint."""
+
+    def on_save_checkpoint_start(self, iteration: int = 0) -> None:
+        """Called when start saving a checkpoint."""
+
+    def on_save_checkpoint_end(self, iteration: int = 0) -> None:
+        """Called when saving checkpoint (sync part) call ends."""
+
+    def on_save_checkpoint_success(self, iteration: int = 0) -> None:
+        """Called when checkpoint is saved successfully."""


nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,170 @@
...
+        self.one_logger.on_train_batch_end()
+
+    def on_validation_start(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> None:
+        """Called when validation begins.
+
+        Args:
+            trainer: The PyTorch Lightning trainer
+            pl_module: The PyTorch Lightning module
+        """
+        self.one_logger.on_validation_start(val_iterations=trainer.global_step)
```
dorotat-nv: it doesnt seem to be compatible with on_validation_start of OneLoggerUtils

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def on_validation_end(self, trainer: pl.Trainer, pl_module: pl.LightningModule) -> None:
```
dorotat-nv: this fails if the batch specific hooks for validation are not provided

```
    def on_validation_batch_start(self, trainer, pl_module, batch, batch_idx, dataloader_idx=0) -> None:
        """Called when validation batch begins.

        Args:
            trainer: The PyTorch Lightning trainer
            pl_module: The PyTorch Lightning module
        """
        self.one_logger.on_validation_batch_start()

    def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx=0) -> None:
        """Called when validation batch ends.

        Args:
            trainer: The PyTorch Lightning trainer
            pl_module: The PyTorch Lightning module
        """
        self.one_logger.on_validation_batch_end()
```

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,147 @@
...
+OneLogger callback for NeMo training.
+
+This module provides a callback that integrates OneLogger telemetry with NeMo training.
+"""
+
+import functools
+import logging
+from typing import Any, Dict, List, Optional, Type
+
+import nv_one_logger.training_telemetry.api.callbacks as CB
```
PeiyuanQi: Eric: it is better to publish to public index, if we have difficulties, make sure this won't fail with import guard, and maybe ref to the open sourced one logger repo.

nemo/utils/exp_manager.py
```diff
@@ -439,6 +442,85 @@ def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx)
...
 
+def configure_onelogger(cfg: OmegaConf, trainer: Optional[pl.Trainer] = None) -> None:
+    """Configure OneLogger v2 callback with training telemetry and exporters.
+
+    Args:
+        cfg: The configuration object
+        trainer: Optional trainer instance for checkpoint configuration
+    """
+    try:
+        from nv_one_logger.exporter.wandb_exporter import WandbExporter
```
PeiyuanQi: Saju: we need to change this import to be the config based exporter import. so we could change the exporter destination by simple onelogger package upgrade.

nemo/collections/llm/api.py
```diff
@@ -56,6 +56,7 @@
 )
 from nemo.lightning.base import NEMO_MODELS_CACHE
 from nemo.lightning.ckpt_utils import ckpt_to_context_subdir
+from nemo.lightning.one_logger_callback import get_onelogger_callbacks, update_one_logger_config
```
PeiyuanQi: get_one_logger_callbacks.

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,138 @@
...
+        # - is_log_throughput_enabled_or_fn: defaults to False (requires additional config)
+        "save_checkpoint_strategy": "async",  # TODO: need to remove this in nv-one-logger side
+        # - summary_data_schema_version_or_fn: defaults to "1.0.0"
+    }
+
+    return init_config
+
+
+def get_onelogger_training_loop_config(
+    trainer: Any,
```
PeiyuanQi: type should be trainer instead of Any, otherwise the type check is voided?

nemo/collections/llm/fn/mixin.py
```diff
@@ -50,6 +50,21 @@ class FNMixin:
...
+    def __init_subclass__(cls, **kwargs):
+        # Add OneLogger timing hooks for LightningModule subclasses
+        try:
+            import lightning.pytorch as pl
+
+            if issubclass(cls, pl.LightningModule):
+                from nemo.lightning.one_logger_callback import hook_class_init_with_callbacks
+
+                hook_class_init_with_callbacks(cls, "on_model_init_start", "on_model_init_end")
+        except (ImportError, AttributeError, Exception):
```
PytLab: @liquor233 is `ImportError` + `AttrbuteError` sufficient? or we only need generic `Exception` here? we should keep either `Exception` or specific Error type here.

nemo/utils/exp_manager.py
```diff
@@ -292,6 +292,8 @@ class ExpManagerConfig:
     fault_tolerance: Optional[FaultToleranceParams] = field(default_factory=FaultToleranceParams)
     # logs TFLOPs per sec per gpu
     log_tflops_per_sec_per_gpu: Optional[bool] = True
+    # OneLogger configuration
+    enable_onelogger: Optional[bool] = True
```
PytLab: @liquor233 let's rename all `onelogger` -> `one_logger`

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+    if not HAVE_ONELOGGER:
+        return _noop
+    if hasattr(CB, name):
+        return getattr(CB, name)
+    else:
+        raise AttributeError(f"OneLogger has no attribute {name}")
+
+
+# Wrapper functions for OneLogger callbacks
+def get_one_logger_callbacks(name: str, *args, **kwargs):
```
PytLab: @liquor233 the func name is confusing. we are getting and call one single cb function here, but it shows only `get` and `callbacks`

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+    TrainingTelemetryConfig = None
+    OneLoggerNeMoCallback = None
+
+from lightning.pytorch import Trainer
+
+from nemo.utils.meta_info_manager import (
+    enable_onelogger,
+    get_nemo_v1_callback_config,
+    get_nemo_v2_callback_config,
+    get_onelogger_init_config,
```
PytLab: @liquor233  need renaming

nemo/collections/llm/fn/mixin.py
liquor233: Replaced with a generic Exception.

nemo/lightning/one_logger_callback.py
liquor233: Done.

nemo/lightning/one_logger_callback.py
liquor233: Renamed.

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def _noop(*args, **kwargs):
+        pass
+
+    if not HAVE_ONE_LOGGER:
+        return _noop
+    if hasattr(CB, name):
+        return getattr(CB, name)
+    else:
+        raise AttributeError(f"OneLogger has no attribute {name}")
```
PytLab: @liquor233 it's better have more accurate err message e.g., `f'Invalid callback name: {name}'`

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+        return
+
+    original_init = cls.__init__
+
+    # Check if already wrapped to avoid double wrapping
+    if getattr(original_init, '_one_logger_wrapped', False):
+        return
+
+    @functools.wraps(original_init)
+    def wrapped_init(self, *args, **kwargs):
```
PytLab: @liquor233 we should also make it robust for adding thread-safety unless the `__init__` is only used in thread-safe mode in NeMo. Let's review this with the code owner

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+        # Check if this instance has already been initialized to prevent duplicate callbacks
+        # in inheritance chains
+        if hasattr(self, '_one_logger_init_started'):
+            # This instance is already being initialized, skip the callbacks
+            return original_init(self, *args, **kwargs)
+
+        # Mark this instance as being initialized
+        self._one_logger_init_started = True
```
PytLab: @liquor233 is this for guard the multi-thread case?

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+        call_one_logger_callback(end_callback, finish_time_msec=get_current_time_msec())
+        return result
+
+    # Mark as wrapped to prevent double wrapping
+    wrapped_init._one_logger_wrapped = True
+    cls.__init__ = wrapped_init
+
+
+def init_one_logger() -> None:
+    """Initialize OneLogger with configuration from OneLoggerMetaInfoManager.
```
PytLab: @liquor233 the meta info manager is not for only OneLogger right? this seems inaccurate

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+    - Exporters are configured via entry points
+    """
+    global _ONE_LOGGER_CALLBACK
+    global HAVE_ONE_LOGGER
+
+    if not HAVE_ONE_LOGGER:
+        return
+
+    # Check if OneLogger is enabled
+    if not enable_one_logger:
```
PytLab: @liquor233 @sajup-oss can you explain what the flag is for? I see we are setting this to `True` in meta info manager. it means if we import the meta_info_manager correctly, it will always be `True` right?

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # Add the OneLogger callback to the trainer if provided
+    if trainer is not None:
+        # Check if OneLoggerNeMoCallback is already in the trainer's callbacks
+        has_one_logger_callback = any(isinstance(callback, OneLoggerNeMoCallback) for callback in trainer.callbacks)
+
+        if not has_one_logger_callback and _ONE_LOGGER_CALLBACK is not None:
+            # Create the callback with metadata
+            one_logger_callback = _ONE_LOGGER_CALLBACK
+            trainer.callbacks.append(one_logger_callback)
```
PytLab: @liquor233 let's insert the one_logger callback before other callbacks

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+    generating and applying a TrainingTelemetryConfig based on the current training context.
+    The configuration is constructed using the appropriate NeMo v1 or v2 telemetry config
+    generator, depending on the provided version.
+
+    Args:
+        nemo_version: 'v1' or 'v2', selects which telemetry config generator to use.
+        kwargs: Additional keyword arguments to pass to the telemetry config generator.
+    """
+    # Check if TrainingTelemetryProvider is already configured
+    if not HAVE_ONE_LOGGER or not TrainingTelemetryProvider.instance().one_logger_ready:
```
PytLab: @liquor233 we are having a dedicated global var `HAVE_ONE_LOGGER`, shall it be sufficient to use it only to know if one-logger is available?

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+    if not HAVE_ONE_LOGGER:
+        return
+
+    # Check if OneLogger is enabled
+    if not enable_one_logger:
+        return
+
+    try:
+        # Check if OneLogger is already configured
+        if TrainingTelemetryProvider.instance().one_logger_ready:
```
PytLab: @liquor233 init_one_logger should be called once right ? in which case, we can reach here?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -390,9 +391,14 @@ def on_train_end(self, trainer, pl_module):
                 else:
                     super()._save_last_checkpoint(trainer, monitor_candidates)
             if self.save_context_on_train_end and not self.always_save_context and is_global_rank_zero():
-                TrainerContext.from_trainer(trainer).io_dump(
-                    ckpt_to_dir(self.last_model_path) / "context", yaml_attrs=["model"]
-                )
+                try:
```
PytLab, line 394: @liquor233 is code change is not one-logger relevant. why are we changing original NeMo code in this PR?

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+try:
+    import nv_one_logger.training_telemetry.api.callbacks as CB
+    from nv_one_logger.api.config import OneLoggerConfig
+    from nv_one_logger.training_telemetry.api.config import TrainingTelemetryConfig
+    from nv_one_logger.training_telemetry.api.training_telemetry_provider import TrainingTelemetryProvider
+    from nv_one_logger.training_telemetry.integration.pytorch_lightning import (
+        TimeEventCallback as OneLoggerNeMoCallback,
+    )
+
+    HAVE_ONE_LOGGER = True
```
PytLab: @liquor233 seems this var is only used within `one_logger_callback` module. it's better to add underscore as prefix in the name

nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -625,13 +635,22 @@ def _save_checkpoint(self, trainer: 'lightning.pytorch.Trainer', filepath: str)
...
+                    logging.warning(f"Failed to dump training context for checkpoint {filepath}: {e}")
 
             if self.async_save:
                 self._last_checkpoint_saved = filepath
                 logging.info(f'Scheduled async checkpoint save for {filepath}')
             else:
                 finalize_fn()
+                # Call OneLogger checkpoint success callback for sync checkpointing
+                call_one_logger_callback("on_save_checkpoint_success", global_step=trainer.global_step)
+            # Call OneLogger checkpoint end callback
```
PytLab: @liquor233 let's clean up these comments. we should clarify the "why" behind the code rather than the "what"

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+"""MetaInfoManager module for handling experiment metadata configuration."""
+
+import os
+from typing import Any, Dict
+
+enable_one_logger = True
+
+
+def get_one_logger_init_config() -> Dict[str, Any]:
```
PytLab: @liquor233 @sajup-oss the original purpose of adding meta info module in NeMo is to have a generic module for generating metadata for not only one-logger usage. but seems we are only having one_logger related functions. we may need to give a better name here.

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        session_tag = os.environ.get("EXP_NAME")  # For NeMo v1
+    else:
+        session_tag = os.environ.get("SLURM_JOB_NAME", "nemo-run")
+
+    world_size = int(os.environ.get('WORLD_SIZE', 1))
+
+    # Minimal configuration - required fields only
+    init_config = {
+        # Required fields (from OneLoggerConfig) - no defaults
+        "application_name": "nemo-application",
```
PytLab: @liquor233 let's just use `nemo` ?

nemo/lightning/one_logger_callback.py
liquor233: Done.

nemo/lightning/one_logger_callback.py
liquor233: I think it is sufficient -- we have already added a try-catch in init onelogger now. I removed the second condition.

nemo/lightning/one_logger_callback.py
liquor233: This will be only called only once as long as users do not call this function explicitly in their code. I suggest to keep this to avoid such cases.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
liquor233: The CICD get failed due to the checkpointing issue -- I have double checked and this is not related to OneLogger code change, this fix is needed if we want to pass the CICD pipeline.

nemo/lightning/one_logger_callback.py
liquor233: Changed.

nemo/lightning/pytorch/callbacks/model_checkpoint.py
liquor233: Done.

nemo/utils/exp_manager.py
liquor233: I think this is done. Could you check and resolve?

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,230 @@
...
+        # Get initialization configuration
+        init_config = get_one_logger_init_config()
+        one_logger_config = OneLoggerConfig(**init_config)
+
+        # Configure the provider with entry-point exporters (automatically calls on_app_start)
+        TrainingTelemetryProvider.instance().with_base_config(
+            one_logger_config
+        ).with_export_config().configure_provider()
+        _ONE_LOGGER_CALLBACK = OneLoggerNeMoCallback(TrainingTelemetryProvider.instance())
+    except Exception:
```
PytLab: @liquor233 is this try...except... block to wide? we are hiding possible configuration error even when user is trying to create one-logger callback but failed because of some unknown issue.

nemo/lightning/one_logger_callback.py
liquor233: Done.

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        trainer: PyTorch Lightning trainer instance
+        global_batch_size: Global batch size (calculated by version-specific function)
+        seq_length: Sequence length (calculated by version-specific function)
+
+    Returns:
+        Dictionary containing base training callback configuration
+    """
+    # Extract values from trainer
+    # Get job name from multiple sources in order of reliability
+    if "EXP_NAME" in os.environ:
```
PytLab: @liquor233 there is duplication in another function: https://github.com/NVIDIA-NeMo/NeMo/pull/13437/files#diff-b856b1c15015d3cfa5f428104fa2c6b4dbc39f6d0739d535f80f1e406a185ae5R34

nemo/lightning/one_logger_callback.py
liquor233: No, this is for guard such case: A  model class is inherited from multiple base classes where we have already called one_logger_init in the former base class.

nemo/lightning/one_logger_callback.py
liquor233: No, currently it is only for OneLogger.

nemo/lightning/one_logger_callback.py
liquor233: We need this because users will be able to disable onelogger; it's currently hard-coded as True, but we can set it via command-line arguments or environment variables. We can discuss this with code owners.

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        job_name = os.environ.get("EXP_NAME")  # For NeMo v1
+    else:
+        job_name = os.environ.get("SLURM_JOB_NAME", "nemo-run")
+
+    world_size = int(os.environ.get('WORLD_SIZE', 1))
+    max_steps = getattr(trainer, 'max_steps', 1)
+    # Use hardcoded value for log_every_n_steps instead of getting from trainer
```
PytLab: @liquor233 need to clean up this comment?

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+    log_every_n_steps = getattr(trainer, 'log_every_n_steps', 10)
```
PytLab: Need review this freq with NeMo owner to avoid possible high freq data posting

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+    micro_batch_size = global_batch_size // world_size
+    # Get PERF_VERSION_TAG from environment
+    perf_version_tag = os.environ.get('PERF_VERSION_TAG', '0.0.0')
+
+    # Calculate performance tag
+    perf_tag = f"{job_name}_{perf_version_tag}_bf{global_batch_size}_se{seq_length}_ws{world_size}"
```
PytLab: Need review with NeMo owner

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+    # Calculate train samples target
+    train_samples_target = max_steps * global_batch_size
+
+    # Fallback values
+    is_save_checkpoint_enabled = False
+    is_validation_iterations_enabled = False
+    save_checkpoint_strategy = "sync"
+
+    checkpoint_callbacks = [cb for cb in trainer.callbacks if 'Checkpoint' in type(cb).__name__]
```
PytLab: @liquor233 we are using string matching here, can it be more specific in NeMo? or is there any exception where the checkpoint callback in NeMo doesn't have `'Checkpoint'` keyword in the name

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+    is_save_checkpoint_enabled = len(checkpoint_callbacks) > 0
+
+    val_check_interval = getattr(trainer, 'val_check_interval', -1)
+    is_validation_iterations_enabled = val_check_interval > 0
+
+    # Check for async_save in trainer strategy (handle both dict and object cases)
```
PytLab: @liquor233 is this the difference between v1 & v2?

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+    if hasattr(trainer, 'strategy') and trainer.strategy is not None:
+        if isinstance(trainer.strategy, dict):
+            if trainer.strategy.get('async_save', False):
+                save_checkpoint_strategy = "async"
+        else:
+            if hasattr(trainer.strategy, 'async_save') and trainer.strategy.async_save:
+                save_checkpoint_strategy = "async"
+
+    for callback in checkpoint_callbacks:
```
PytLab: Need NeMo owner review about the logic of detecting async_save mode

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        hasattr(trainer, 'lightning_module')
+        and trainer.lightning_module is not None
+        and hasattr(trainer.lightning_module, 'cfg')
+    ):
+        model_cfg = trainer.lightning_module.cfg
+        if hasattr(model_cfg, 'train_ds') and hasattr(model_cfg.train_ds, 'batch_size'):
+            micro_batch_size = model_cfg.train_ds.batch_size
+            global_batch_size = micro_batch_size * int(os.environ.get('WORLD_SIZE', 1))
+        elif hasattr(model_cfg, 'train_ds') and hasattr(model_cfg.train_ds, 'bucket_batch_size'):
+            # For ASR with bucketing, use the average batch size
```
PytLab: @liquor233 Need to add a NOTE here for we are having this as temporary solution for ASR case that using average can lead possible inaccurate progress tracking

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Uses environment variables instead of torch.distributed to avoid circular imports.
+    In distributed training, typically only rank 0 (or the last rank) should
+    enable OneLogger to avoid duplicate telemetry data.
+
+    Returns:
+        True if OneLogger should be enabled for the current rank, False otherwise
+    """
+    try:
+        rank = int(os.environ.get('RANK', 0))
```
PytLab: @liquor233 need to confirm if NeMo has any module or function to get this info. if yes, we should reuse

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+        world_size = int(os.environ.get('WORLD_SIZE', 1))
+
+        # Enable for rank 0 or the last rank (common pattern)
+        return rank == 0 or rank == world_size - 1
```
PytLab: @sajup-oss @liquor233 so this means one-logger can be enabled on both rank_0 or rank_N-1?

nemo/utils/meta_info_manager.py
```diff
@@ -0,0 +1,243 @@
...
+    except (ValueError, TypeError):
+        # Default to True on invalid values (as expected by tests)
```
PytLab: @liquor233 @sajup-oss in what scenario, we can reach here?

nemo/collections/llm/fn/mixin.py
PytLab: @liquor233 let's remove the try catch block

nemo/lightning/one_logger_callback.py
PytLab: @liquor233 to remove the `_one_logger_init_started` flag

nemo/lightning/one_logger_callback.py
PytLab: @liquor233 move the meta_info_manager to one_logger_callbacks

nemo/lightning/pytorch/callbacks/model_checkpoint.py
PytLab: Need NeMo owner to help resolve this

nemo/utils/meta_info_manager.py
liquor233: Ack -- changed.

nemo/lightning/one_logger_callback.py
liquor233: Fixed.

nemo/utils/meta_info_manager.py
liquor233: Fixed.

nemo/utils/meta_info_manager.py
liquor233: Fixed with type check -- [cb for cb in trainer.callbacks if isinstance(cb, ModelCheckpoint)]

nemo/utils/meta_info_manager.py
liquor233: Added "# This is a temporary fix to support the bucketing"

nemo/utils/meta_info_manager.py
liquor233: This is incorrect -- I have fixed, but will launch a test to explore why we did not capture this issue before.

nemo/utils/meta_info_manager.py
liquor233: Fixed in last conversation.

nemo/utils/meta_info_manager.py
liquor233: Discussed with @PytLab , will move the one-logger specific code to one_logger_callback.py, will change the metadata manager in the future.

nemo/lightning/one_logger_callback.py
liquor233: Discussed with @PytLab , we can remove this.

nemo/lightning/one_logger_callback.py
liquor233: Discussed with @PytLab , will remove the flag one_logger_init_start but keep _one_logger_wrapped

nemo/lightning/one_logger_callback.py
liquor233: Removed.

nemo/lightning/one_logger_callback.py
liquor233: Discussed with @PytLab , no need to change the code. Please resolve.

nemo/collections/llm/fn/mixin.py
liquor233: Replaced.

liquor233:
Discussed with @PytLab , need to add the callbackgroup abstract class.

nemo/utils/meta_info_manager.py
PytLab: @liquor233 also need to update test case

nemo/collections/llm/api.py
```diff
@@ -135,6 +136,9 @@ def train(
 
     trainer.fit(model, data)
 
+    # Track app end for NeMo v2 recipe-based applications
+    CallbackGroup.get_instance().on_app_end()
```
stef1927: Where is the matching `on_app_start()` called? Why not consider a function decorator to call both?

nemo/collections/llm/api.py
```diff
@@ -1255,11 +1259,19 @@ def _setup(
...
         task_config=getattr(train, "__io__", None),
     )
+
+    # Configure telemetry via CallbackGroup
+    CallbackGroup.get_instance().update_config(nemo_version='v2', trainer=trainer, data=data)
+
     if resume is not None:
+        CallbackGroup.get_instance().on_load_checkpoint_start()
         resume.setup(trainer, model)
+        CallbackGroup.get_instance().on_load_checkpoint_end()
```
stef1927, line 1272: Consider introducing a context manager for cases similar to this, so that the end function is always called, and we can log any exception as an error event. For Heimdall error events are required.

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,79 @@
...
+    update_config computes TrainingTelemetryConfig and applies it.
+    """
+
+    def __init__(self) -> None:
+        # Configure provider from meta info
+        init_config = get_one_logger_init_config()
+        one_logger_config = OneLoggerConfig(**init_config)
+        TrainingTelemetryProvider.instance().with_base_config(
+            one_logger_config
+        ).with_export_config().configure_provider()
```
stef1927, line 286: How can users select different exporters?

stef1927 (commented):
I've left  a couple of comments after a quick review as I've noticed that the code is not yet stable, i.e. the callback group is back now. Let me know once the code is stable, reviewed by the Nemo team, and ready for the Heimdall signoff.

Here are the Heimdall requirements that may be missing at the moment:

* The exporters need to be configurable, for example with a command line argument or environment variable that the users can set.
* Exporters may be different according to the rank number, for example rank 0 may log to file whereas all ranks send data to an open telemetry endpoint.
* Exceptions need to be captured, and recorded as error events.
* The end functions should be called even in the presence of exceptions.
* We need to capture distributed initialization and model forward too.

Not an Heimdall requirement, but for you to consider:

* Generally, if it's possible, you should try to avoid modifying all model and data loader implementations, and hook only in one place. You might need the help of the Nemo team to determine if this is possible for things like model initialization, data loading and checkpoints.

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def init_one_logger() -> None:
+    """Initialize OneLogger callback if enabled.
+
+    Registers the OneLoggerNeMoCallback with the CallbackGroup singleton
+    if OneLogger is enabled in the configuration.
+    """
+    if enable_one_logger:
+        CallbackGroup.get_instance().register(OneLoggerNeMoCallback())
```
PytLab: @liquor233 we missed the `on_app_start` here

nemo/collections/llm/api.py
PytLab: @stef1927 we want the `on_app_start` to be invoked as early as possible, so in nv-one-logger we call it automatically when the callback is initialzed. in current NeMo, it is initialized when it is imported.

https://github.com/NVIDIA/nv-one-logger/blob/main/nv_one_logger/one_logger_pytorch_lightning_integration/src/nv_one_logger/training_telemetry/integration/pytorch_lightning.py#L71

nemo/lightning/one_logger_callback.py
PytLab: @stef1927 the exporter behavior is controlled by one-logger's [exporter manager](https://github.com/NVIDIA/nv-one-logger/blob/main/nv_one_logger/one_logger_core/src/nv_one_logger/exporter/export_config_manager.py) we added recently. it can be configured via:
1. installing configuure package
2. local yml file
3. coding explicitly.

nemo/lightning/callback_group.py
```diff
@@ -0,0 +1,130 @@
...
+    if not hasattr(cls, '__init__'):
+        return
+
+    original_init = cls.__init__
+
+    if getattr(original_init, '_callback_group_wrapped', False):
+        return
+
+    @functools.wraps(original_init)
+    def wrapped_init(self, *args, **kwargs):
```
PytLab, line 159: @liquor233 we need one more flag to check if we are already in a wrapped_init function e.g., `_in_wrapped_init` to avoid repeated hooking in hierarchical inheritance

nemo/collections/llm/api.py
PytLab: > The exporters need to be configurable, for example with a command line argument or environment variable that the users can set.

this is has been supported in nv-one-logger package [exporter_config_manager](https://github.com/NVIDIA/nv-one-logger/tree/main/nv_one_logger/one_logger_core/src/nv_one_logger/exporter#file-discovery) where one-logger can be configured via package or file or explicit config code. 

> Exporters may be different according to the rank number, for example rank 0 may log to file whereas all ranks send data to an open telemetry endpoint.

Current one-logger is designed to be enabled in single-rank. so this will need upgrading one-logger's exporter config to support rank-specific exporting @sajup-oss 

> Exceptions need to be captured, and recorded as error events.

@stef1927 as you mentioned you also use PTL callback here, do you mean you want to track the exceptions for the events those are not covered by PTL callbacks? 

> The end functions should be called even in the presence of exceptions

this is not implemented in current PR. will add this 

> We need to capture distributed initialization and model forward too.

Ok got it. this will also need upgrade in nv-one-logger firstly as new APIs

PytLab: @liquor233 we need to add exception handling logic here to guarantee the `on_app_end` is called finally. let's try to find what is the best way in current NeMo repo

nemo/lightning/callback_group.py
```diff
@@ -0,0 +1,141 @@
...
+    # Idempotence guard: avoid wrapping the same __init__ multiple times (e.g., in multiple inheritance)
+    if getattr(original_init, '_init_wrapped_for_callbacks', False) or getattr(
+        original_init, '_callback_group_wrapped', False
+    ):
+        return
+
+    @functools.wraps(original_init)
+    def wrapped_init(self, *args, **kwargs):
+        # Reentrancy guard: avoid double-emitting hooks across super().__init__ chains
+        if getattr(self, '_callback_group_emitting_init_hooks', False):
```
PytLab: @liquor233 where did we set this flag? I'm only seeing this flag is being used once here. isn't it `_in_wrapped_init`?

nemo/lightning/callback_group.py
```diff
@@ -0,0 +1,141 @@
...
+            # If we're already inside a wrapped __init__, just call the original
+            return original_init(self, *args, **kwargs)
+
+        setattr(self, '_in_wrapped_init', True)
```
PytLab, line 165: @liquor233 this is set, but not used

nemo/lightning/callback_group.py
```diff
@@ -0,0 +1,141 @@
...
+        end_callback (str): CallbackGroup method to call after __init__.
+    """
+    if not hasattr(cls, '__init__'):
+        return
+
+    original_init = cls.__init__
+
+    # Idempotence guard: avoid wrapping the same __init__ multiple times (e.g., in multiple inheritance)
+    if getattr(original_init, '_init_wrapped_for_callbacks', False) or getattr(
+        original_init, '_callback_group_wrapped', False
```
PytLab: @liquor233 what is the difference between `_init_wrapped_for_callbacks` & `_callback_group_wrapped`?

nemo/lightning/one_logger_callback.py
stef1927: Where can the yaml file be passed in? I don't see `generate_export_config()` called anywhere. It's important to be able to change exporters without having to change the code, so we need either a command line argument or env variable pointing to the yaml file path.

nemo/lightning/callback_group.py
liquor233: Oh, sorry, I pushed the wrong change, it's actually a naming issue, fixed now.

nemo/collections/llm/api.py
stef1927: > we want the on_app_start to be invoked as early as possible, so in nv-one-logger we call it automatically when the callback is initialzed. in current NeMo, it is initialized when it is imported.

Agreed that we want to call it as soon as possible, but right now we are calling `on_app_start()` in the one-logger repository and `on_app_end()` in the Nemo repository, did I understand correctly? IMO, this makes it impossible to understand this code unless you are familiar with both repositories. Think of future maintainers of either repository, how are they going to cope with code split this way? What about moving `on_app_start()` in `OneLoggerNeMoCallback`? 

> do you mean you want to track the exceptions for the events those are not covered by PTL callbacks?

I meant the exceptions not handled by the user code, so a try/finally in this function, or context, or function decorator if you can access the callbacks in the decorator implementation

nemo/lightning/one_logger_callback.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+OneLogger callback for NeMo training.
+
+This module provides a callback that integrates OneLogger telemetry with NeMo training.
+"""
+from lightning.pytorch import Trainer
+from nv_one_logger.api.config import OneLoggerConfig
+from nv_one_logger.training_telemetry.api.config import TrainingTelemetryConfig
+from nv_one_logger.training_telemetry.api.training_telemetry_provider import TrainingTelemetryProvider
+from nv_one_logger.training_telemetry.integration.pytorch_lightning import TimeEventCallback as OneLoggerPTLCallback
```
stef1927, line 29: How much does this callback buy us, it seems to be a very thin wrapper of the training callback, why not use the training callback directly in `OneLoggerNeMoCallback` below, so that we can have `on_app_start()` called in this file?

nemo/lightning/one_logger_callback.py
PytLab: @stef1927 in the exporter_config_manager, we implemented the config file search sequence as below:
1. direct path provided in parameter of function
2. check the env var `ONE_LOGGER_EXPORTER_CONFIG_PATH`
3. search in CWD
the env var should work for you, but let me know if it works for all your cases?

nemo/collections/llm/api.py
PytLab: @stef1927 added try...finally for guarantee `on_app_end` call

nemo/utils/meta_info_manager.py
PytLab: @liquor233 to move this and corresponding test cases to `one_logger_callback.py`

nemo/collections/llm/api.py
PytLab: @stef1927 as we are adding `CallbackGroup` layer in NeMo, the change to support context manager hook is not trivial:
1. need to add new interface in `CallbackGroup` for constructing context manager
2. need to add exception catch logic to get one-logger callback out of callback list and report error to exporter
3. update model_init/dataload_init/load_checkpoint related code in NeMo PR
4. add tests to cover all above changes
So we are suggesting to have this in another PR to support ctx. for on_app_end try...finally..., we've already added as a quick change.

nemo/lightning/one_logger_callback.py
PytLab: yes it makes sense. we added it in one-logger v2 for compatibility with v1. we are adding controlling of implicit `on_app_start` call in the initializer so that we can have `on_app_start` called in this file with the parameter flag to be `False`. we are pushing new changes into this PR and include the nv-one-logger changes in the next 2.1.0 release

nemo/collections/llm/api.py
PytLab: Mark as TODO in next PR (@stef1927 )

nemo/collections/llm/api.py
PytLab: Mark as TODO in next PR (@stef1927 )

stef1927: Next PR sounds good, thanks.

nemo/collections/llm/api.py
stef1927: Next PR sounds good, thanks.

nemo/lightning/one_logger_callback.py
stef1927: Env var and CWD should work, thanks

nemo/lightning/one_logger_callback.py
stef1927: that's great, thanks!

nemo/collections/llm/api.py
stef1927: Great, thanks!

nemo/lightning/callback_group.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Returns:
+            CallbackGroup: The singleton instance.
+        """
+        if cls._instance is None:
+            cls._instance = CallbackGroup()
+        return cls._instance
+
+    def __init__(self) -> None:
+        self._callbacks: List[BaseCallback] = [OneLoggerNeMoCallback()]
```
leoleoasd, line 46: Why is OneLoggerNeMoCallback enabled by default?

PytLab: @leoleoasd nv-one-logger is a NeMo dependency and integrated as default tool. but it is not enabled by default and have no-op if user doesn't configure it explicitly. 

the warning message is the nv-one-logger issue which the team will fix in next release soon.

leoleoasd: I still don't believe it's the correct way to enabling something that will produce tons of warning message on each training step. I could not find any documentation about how to disable or "configure it explicitly" so it won't make warnings. 

Why don't add this callback only after user configure it explicitly?

nemo/core/classes/modelPT.py
```diff
@@ -1344,6 +1374,8 @@ def maybe_init_from_pretrained_checkpoint(self, cfg: OmegaConf, map_location: st
                 f"Found : {[args[idx] for idx, arg_present in enumerate(arg_matches) if arg_present]}"
             )
 
+        CallbackGroup.get_instance().on_load_checkpoint_start()
```
weiqingw4ng, line 1377: Can this function be called multiple times? It is called here and then called in function `self.restore_from`. This raises an error for my training when I try to load a checkpoint for finetuning, and I can start training after removing the `on_load_checkpoint_start` and `on_load_checkpoint_end` function here. 

```
  File "/code/examples/asr/asr_transducer/speech_to_text_rnnt_bpe.py", line 81, in main                                                                                                     
    asr_model.maybe_init_from_pretrained_checkpoint(cfg)                                                                                                                                    
  File "/opt/venv/lib/python3.12/site-packages/lightning_utilities/core/rank_zero.py", line 41, in wrapped_fn                                                                               
    return fn(*args, **kwargs)                                                                                                                                                              
           ^^^^^^^^^^^^^^^^^^^                                                                                                                                                              
  File "/code/nemo/core/classes/modelPT.py", line 1396, in maybe_init_from_pretrained_checkpoint                                                                                            
    restored_model = self.restore_from(                                                                                                                                                     
                     ^^^^^^^^^^^^^^^^^^                                                                                                                                                     
  File "/code/nemo/core/classes/modelPT.py", line 482, in restore_from                                                                                                                      
    CallbackGroup.get_instance().on_load_checkpoint_start()                                                                                                                                 
  File "/code/nemo/lightning/callback_group.py", line 120, in dispatcher                                                                                                                    
    method(*args, **kwargs)                                                                                                                                                                 
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/training_telemetry/integration/pytorch_lightning.py", line 174, in on_load_checkpoint_start                                   
    on_load_checkpoint_start()                                                                                                                                                              
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/core/internal/safe_execution.py", line 80, in wrapper                                                                         
    return exception_guard(func)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/core/internal/safe_execution.py", line 46, in wrapper
    raise e
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/core/internal/safe_execution.py", line 35, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/training_telemetry/api/callbacks.py", line 143, in on_load_checkpoint_start
    return _recorder().on_load_checkpoint_start(start_time)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/training_telemetry/api/training_recorder.py", line 553, in on_load_checkpoint_start
    self._training_state.multi_iteration_timers[StandardTrainingJobSpanName.CHECKPOINT_LOAD].start(start_time)
  File "/usr/local/lib/python3.12/dist-packages/nv_one_logger/core/internal/multi_window_timer.py", line 70, in start
    raise OneLoggerError("Cannot start timer since it is already active")
nv_one_logger.core.exceptions.OneLoggerError: Cannot start timer since it is already active
```

tango4j: @PytLab 
Hi. This error is blocking our model release. 
Will there be a issue if we disable this for "restore_from" function?
Clearly, this fine-tuning case (restoring a checkpoint) is not tested, and causing errors.

PytLab: @tango4j removing the one-logger callback pairs in `restore_from` should resolve this issue and unblock the run. one question for long-term solution, is there a case where only `restore_from` function is used for loading checkpointing?

PytLab: > Will there be a issue if we disable this for "restore_from" function?

should be fine. it wont break one-logger instrumentation. but if there is a case where use only calls the `restore_from` for ckpt loading, with this removed, one-logger may lose some trace metrics for checkpoint_loading overheads.

PytLab: @tango4j we are working on a new hotfix PR for resolving the duplicated callback for ckpt_load

nemo/lightning/callback_group.py
PytLab: > I still don't believe it's the correct way to enabling something that will produce tons of warning message on each training step

@leoleoasd agree. this is a bug from nv-one-logger we are fixing and testing internally. 

> Why don't add this callback only after user configure it explicitly?

nv-one-logger is NVIDIA tools for cluster productivity tracking, it is designed to be enabled by default for nvidia internal user, for external users, it is still one of the dependencies and is imported for some initialization check but is disabled by default unless it is configured explicitly.

PR #13526, opened 2025-05-10
Multi token bleu

tbartley94:
> [!IMPORTANT]  

# What does this PR do ?

- Adds functionality to asr metrics so multiple bleu tokenizers can be used on evaluation. This allows for multilingual cases where a single SacreBLEU tokenizer isn't able to properly tokenize. (e.g. Roman scripts + CJK languages).
- Fixed bug in BLEU metric that improperly nested references, leading to only partial bleu calculations..
- Changed references in aed_models that were not properly referencing test_dataloader attributes.
- Added training_batch logging for bleu and wer metrics gathering during training so behavior is consistent with other asr models.

- asr

# Changelog

- Changed bleu.py to now allow passing dicts of multiple bleu tokenizers
- changed bleu.update to switch tokenizers 'on the fly' for target evaluation
- Changed aed_model logging to pass strings of `target_lang` to bleu.update to allow per sample evaluation
- Added util function to lhotse.cutset to flatten mixed set to allow easier reference to custom attributes
- Replaced references to validation attributes in test dataloder to appropriate keys.
- changed prefix of wer on validation pass to now allow referencing test step
- added intratraining bleu and wer update for training batches to catch training level bleu updates.
- moved metric logging in validation to standalone function to reduce boiletplate between validation and training steps.

# GitHub Actions CI

**Pre checks**:

- [y ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ y] Did you write any new necessary tests?
- [ y] Did you add or update any necessary documentation?
- [ n] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  
**PR Type**:

- [ y] New Feature
- [ y] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


tbartley94:
@pzelasko only adding you as a reviewer in case you have a better way of quickly getting the custom keyword out. (can't use supervision.lang since that's for the tokenizer)

nemo/collections/asr/metrics/bleu.py
```diff
@@ -82,7 +82,10 @@ def on_validation_epoch_end(self):
     def __init__(
         self,
         decoding: Union[AbstractCTCDecoding, AbstractRNNTDecoding, AbstractMultiTaskDecoding],
-        tokenize: Literal["none", "13a", "zh", "intl", "char"] = "13a",
+        tokenize: (
```
pzelasko, line 85: Please update the docstring above

nemo/collections/asr/metrics/bleu.py
```diff
@@ -82,7 +82,10 @@ def on_validation_epoch_end(self):
...
+            Literal["none", "13a", "zh", "intl", "char", "ja-mecab", "ko-mecab", "flores101", "flores200"]
+            | Dict[str, Literal["none", "13a", "zh", "intl", "char", "ja-mecab", "ko-mecab", "flores101", "flores200"]]
+        ) = "13a",
```
nithinraok, line 88: Could you update it something like:
```python
TokenizeType = Union[
    Literal["none", "13a", "zh", "intl", "char", "ja-mecab", "ko-mecab", "flores101", "flores200"],
    Dict[str, Literal["none", "13a", "zh", "intl", "char", "ja-mecab", "ko-mecab", "flores101", "flores200"]]
]
```
and use it in init. Readability is not good with current way.

nemo/collections/asr/metrics/bleu.py
```diff
@@ -91,16 +94,25 @@ def __init__(
...
+        self.log_prediction = log_prediction
+        self.batch_dim_index = batch_dim_index
+        self.decoding = decoding
+        self.decode = None
+        self.multi_token = type(tokenize) is dict
+        self._tokenizer = {}
+
         super().__init__(
-            tokenize=tokenize,
+            tokenize=tokenize["default"] if self.multi_token else tokenize,
```
nithinraok, line 105: why `default` here? Based on above input it can have any string in Dict for key ?

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -698,9 +703,54 @@ def forward(
...
+        self.wer.reset()
+
+        self.bleu.update(
+            predictions=encoded_states,
+            predictions_lengths=encoded_len,
+            targets=batch.transcript,
+            targets_lengths=batch.transcript_lens,
+            predictions_mask=encoded_mask,
+            input_ids=batch.prompt,
+            langs=[c.custom["target_lang"] for c in flatten_mixed(batch.cuts)],
```
nithinraok, line 744: I have been thinking about this as well. Currently we compute wer and bleu score irrespective of task. I thought about this and added the change based on tokenizer type. As this information can be grabbed from prompt and this `target_lang` key might change later based on tokenizer type (canary,canar2). See this branch code: https://github.com/NVIDIA/NeMo/blob/selectively_run_metrics/nemo/collections/asr/models/aed_multitask_models.py#L775-L831 

So instead adding this to cut we can grab from prompt which can be added to tokenizer type class itself.

nemo/collections/asr/modules/transformer/transformer_modules.py
```diff
@@ -44,22 +44,37 @@ def __init__(self, hidden_size, max_sequence_length=512):
...
+        """
+        Forward pass for pos_emb.
+        Checks if position_ids are in range.
+        If yes -> use lookup table
+        If no -> recalculate positions.
+        """
+        start_pos, seq_length = position_ids[0], position_ids.shape[1]
+        if start_pos + seq_length > self.max_sequence_length:
+            return FixedPositionalEncoding._calc_pos_enc(
+                self._hidden_size, seq_length, start_pos=start_pos, device=position_ids.device
```
nithinraok, line 75: Fix codeQL error below. Also make sure pretrained loading works as is.

nemo/collections/asr/metrics/bleu.py
tbartley94: Need to enforce a fallback tokenizer. Don't want to specify a unique tokenizer for every language.

nemo/collections/asr/models/aed_multitask_models.py
tbartley94: Oh I like that, reduces the amount of logging. I'll add this to the PR.

I also spoke with piotr about this offline and we're just going to create a new attribute called 'bleu_token' that you can specify in the config. Less mucking about with prompting.

nemo/collections/asr/modules/transformer/transformer_modules.py
tbartley94: This should have been removed from PR. See the other abs_emb pr and i'll delete this.

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -39,6 +39,15 @@
 from nemo.collections.common.parts.preprocessing.manifest import get_full_path
 
 
+def flatten_mixed(cutset: CutSet) -> CutSet:
```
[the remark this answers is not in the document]
tbartley94: Removed due to offline discussion.

nemo/collections/asr/metrics/bleu.py
tbartley94: How about this:

```
type SacreBLEUToken = Literal["none", "13a", "zh", "intl", "char", "ja-mecab", "ko-mecab", "flores101", "flores200"]
```
```
tokenize: Union[SacreBLEUToken, Dict[str, SacreBLEUToken]] = "13a",
```

tbartley94: want to keep the dict or single tokenizer option clear in the init function

nemo/collections/asr/modules/transformer/transformer_modules.py
tbartley94: resolved as this is removed from pr

nemo/collections/asr/metrics/bleu.py
tbartley94: Actually changed this to no longer use default. Explicitly assigns the default tokenize to attribute.

nemo/collections/asr/models/aed_multitask_models.py
tbartley94: Actually, let's add this code in a separate PR. I want to add a wrapper around it so it can extend to new evaluation cases as they come up (speaker asr and the like) and that's going to add more locs on this pr.

nemo/collections/asr/metrics/bleu.py
tbartley94: Actually turns out each SacreBLEU metric instantiates all tokenizers in backend, see `_get_tokenizer` for new behavior.

nemo/collections/asr/metrics/bleu.py
```diff
@@ -147,30 +134,48 @@ def update(
...
 
-        if self.log_prediction:
-            logging.info("\n")
-            logging.info(f"reference:{references[0]}")
-            logging.info(f"predicted:{hypotheses[0]}")
-
-        super().update(
-            [h.text for h in hypotheses], [references]
-        )  # Note: [references] since BLEU allows multiple references.
+                self.preds_len, self.target_len = _bleu_score_update(
```
nithinraok, line 165: Don;t you need to call super.__update__ here?

tbartley94: Nah, since this is now using the full super code: https://github.com/Lightning-AI/torchmetrics/blob/4c42cf2c29f48f0024bbceb612672c20006a88ec/src/torchmetrics/text/sacre_bleu.py#L125

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -91,6 +92,20 @@ def _config_check(cfg):
...
+    """
+
+    def _get_lang(c):
+        return c.custom.get("bleu_tokenizer", None)
+
+    # Dataloader passes multiple types of cuts. Need to diambiguate to access custom.
+    # TODO: resolve in lhotse backend.
+    return [_get_lang(c.first_non_padding_cut) if isinstance(c, MixedCut) else _get_lang(c) for c in cuts]
+
+
```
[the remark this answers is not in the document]
tbartley94: 1. I'm thinking it's probably better to keep in code since it allows development to pass langs in a custom manner. If you make it a util function there's more an assumption of universality. If you can't see a case where people would depart from this form of access I'm fine moving into utils. (Probably a lot of this header code should be in utils too.)

2. Can you clarify?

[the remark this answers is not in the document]
nithinraok: 2. I meant adding to here instead: https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/asr/data/audio_to_text_lhotse_prompted.py#L77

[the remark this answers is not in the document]
tbartley94: mmm, that works. If another dataloader is used it can be expanded in the future.

@pzelasko any strong opinions about expanding cases for the minibatch class?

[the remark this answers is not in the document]
tbartley94: @nithinraok moved behavior just to bleu to keep things more localized

nemo/collections/asr/models/aed_multitask_models.py
tbartley94: @nithinraok check the new `multitask` metric class. The functions at the bottom should pull off the behavior now.

nemo/collections/asr/models/aed_multitask_models.py
```diff
@@ -225,9 +227,12 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
         self.wer = WER(self.decoding, log_prediction=self.cfg.get("log_prediction"))
-        self.bleu = BLEU(
-            self.decoding, tokenize=self.cfg.get('bleu_tokenizer', "13a"), log_prediction=False
-        )  # Wer is handling logging
+
+        # Resolves bleu_tokenizer config if multiple configs are passed.
+        bleu_tokenizer = self.cfg.get("bleu_tokenizer", "13a")
+        if type(bleu_tokenizer) is DictConfig:
+            bleu_tokenizer = OmegaConf.to_container(bleu_tokenizer)
+        self.bleu = BLEU(self.decoding, tokenize=bleu_tokenizer, log_prediction=False)  # WER is handling logging
```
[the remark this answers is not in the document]
tbartley94: Check yaml file and docs in `multitask` file to see if example is comprehensive

nemo/collections/asr/metrics/bleu.py
```diff
@@ -147,30 +155,48 @@ def update(
                 ``[Time, Batch]`` (if ``batch_dim_index == 1``). Required for MultiTaskDecoding.
             input_ids: an int torch.Tensor of shape ``[Batch, Time]`` (if ``batch_dim_index == 0``) or
                 ``[Time, Batch]`` (if ``batch_dim_index == 1``). Required for MultiTaskDecoding.
+            CutSet: an optional sequence of strings of ``length == batch size``. If `self.check_cuts`, inspects each element
```
pzelasko: arg name should be `cuts:`, and it's a sequence of Cuts not strings

nemo/collections/asr/metrics/bleu.py
```diff
@@ -210,3 +236,39 @@ def _compute_bleu(
...
+            )
+        elif isinstance(self.decoding, AbstractMultiTaskDecoding):
+            self.decode = lambda predictions, prediction_lengths, predictions_mask, input_ids: self.decoding.decode_predictions_tensor(
+                encoder_hidden_states=predictions,
+                encoder_input_mask=predictions_mask,
+                decoder_input_ids=input_ids,
+                return_hypotheses=False,
+            )
+        else:
+            raise TypeError(f"WER metric does not support decoding of type {type(self.decoding)}")
```
pzelasko: WER -> BLEU

nemo/collections/asr/metrics/multitask.py
```diff
@@ -0,0 +1,237 @@
...
+from nemo.collections.asr.data.audio_to_text_lhotse_prompted import PromptedAudioToTextMiniBatch
+from nemo.collections.common.prompts.canary import TASK_TRANSLATE
+from nemo.core.classes import Serialization
+
+__all__ = ['MultiTaskMetric']
+
+# TODO: add decoding metric updater
+
+
+class MultiTaskMetric(Serialization):
```
pzelasko, line 172: Love this, I could re-use this for speech lm

nemo/collections/asr/metrics/multitask.py
```diff
@@ -0,0 +1,237 @@
...
+    values that determines the conditions for the metric to apply. (Currently only the keyword `task` with vals `transcribe` and `translate` is supported.) All defined
+    metrics in `asr.collections.metrics` are supported.
+
+    Similar to `input_cfg` for dataloading, extra defined properties assume soft inheritance. All properties defined outside
+    the `metrics` dict will be autopopulated into all metrics. Meanwhile, properties defined within a metrics entry will only populate that metric.
+
+    """
+
+    # trick from torch metrics `SacreBLEUTokenizer`
+    _INDEX_FN = {
```
pzelasko: Couldn't we instead read `cut.custom["task"]` for generic, prompt-format-invariant way of mapping example to task? `.update()` method already accepts CutSet.

nemo/collections/asr/parts/submodules/multitask_beam_decoding.py
```diff
@@ -232,6 +235,9 @@ def forward(
                 packed_result = pack_hypotheses(hypotheses, best_hypo, beam_scores)
                 self.format_hypotheses(packed_result, decoder_input_ids)
 
+        self.transformer_decoder.train()
+        self.log_softmax_module.train()
```
pzelasko, line 239: nit-pick, but perhaps we should only restore `.train()` if the initial state was `self.is_training == True`

nemo/collections/common/prompts/canary.py
```diff
@@ -29,6 +29,18 @@
     CANARY_SPECIAL_TOKENIZER,
 )
 
+# Use global variables to import slot values in other modules.
+BOOL_TRUE = {"yes", "Yes", "true", "True", "1"}
+BOOL_FALSE = {"no", "No", "false", "False", "0"}
+PNC_TRUE = {"yes", "true", "True", "1", "pnc", "<|pnc|>"}
```
pzelasko:
```suggestion
PNC_TRUE = BOOL_TRUE | {"pnc", "<|pnc|>"}
```
re-uses defined variables

nemo/collections/common/prompts/canary.py
```diff
@@ -38,13 +50,9 @@ class CanaryPromptFormatter(PromptFormatter):
             "template": f"{CANARY_BOS}|source_lang||task||target_lang||pnc|",
             "slots": {
                 "source_lang": Modality.Text,
-                "task": Modality.TextLiteral(
-                    "asr", "ast", "translate", "transcribe", "s2t_translation", "<|transcribe|>", "<|translate|>"
-                ),
+                "task": Modality.TextLiteral(*[t for t in (TASK_TRANSLATE | TASK_TRANSCRIBE)]),
```
pzelasko:
```suggestion
                "task": Modality.TextLiteral(*(TASK_TRANSLATE | TASK_TRANSCRIBE)),
```
should work too, right?

nemo/collections/common/prompts/canary.py
```diff
@@ -38,13 +50,9 @@ class CanaryPromptFormatter(PromptFormatter):
...
                 "target_lang": Modality.Text,
-                "pnc": Modality.TextLiteral(
-                    "yes", "no", "true", "True", "false", "False", "1", "0", "pnc", "nopnc", "<|pnc|>", "<|nopnc|>"
-                ),
+                "pnc": Modality.TextLiteral(*[p for p in (BOOL_TRUE | BOOL_FALSE | PNC_TRUE | PNC_FALSE)]),
```
pzelasko:
```suggestion
                "pnc": Modality.TextLiteral(*(PNC_TRUE | PNC_FALSE)),
```

nemo/collections/common/prompts/canary2.py
```diff
@@ -29,6 +30,17 @@
     CanaryTokenizer,
 )
 
+# Use global variables to import slot values in other modules.
+ITN_TRUE = {
```
pzelasko: similar changes to `canary.py`, remember to add `BOOL_TRUE | ...`

pzelasko (commented):
Great work! I left a bunch of comments, I think we can simplify the multi-metric thing a bit.

nemo/collections/asr/models/aed_multitask_models.py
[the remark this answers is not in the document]
pzelasko: We're starting to see it's useful to pass CutSet around in the model/metric code and read various properties from it (tokenizer lang, task, etc), I'd keep it the way Travis added it.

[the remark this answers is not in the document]
tbartley94: @pzelasko the previous version where you pass the string of langs or the current version where the cuts are passed and the bleu metric manages it?

nemo/collections/asr/metrics/multitask.py
tbartley94: Yeah i just noticed that y'all reimplemented a few metrics. We should probably unify the boiler plate in a later update before we get too in the weeds with multitasks.

nemo/collections/asr/metrics/multitask.py
tbartley94: We can but that assumes `task` is univeral across all future prompts. Canary2 doesn't have a task prompt, for instance. So passing through config seems the most agnostic.

nemo/collections/asr/parts/submodules/multitask_beam_decoding.py
tbartley94: mmm, this mirrors the setup in the rnnt versions of this. i'd rather retain boilerplate since it doesn't have effect any effects. can we think of a place where we'd accidentally cause weird behavior by default toggle for train?

nemo/collections/common/prompts/canary.py
tbartley94: Thanks for catching that

tbartley94: actually, do we want the bools defined with the OBJECT instantiation or just keep them separate until included in the template? i can see arguments for both

nemo/collections/common/prompts/canary.py
tbartley94: ha, you're right

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
```diff
@@ -42,6 +42,22 @@ model:
   prompt_format: ???   # Options supported: ["canary", "canary2"]
   prompt_defaults: null
 
+  # Sub-config for specifying multiple metrics for multitask modeling.
+  # See `asr.collections.metrics.multitask` for detsils.
```
nithinraok:
```suggestion
  # See `asr.collections.metrics.multitask` for details.
```

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
```diff
@@ -42,6 +42,22 @@ model:
...
+  multitask_metrics_cfg:
+    log_predictions: true
+    metrics:
+      -   name: wer
+          _target_: nemo.collections.asr.metrics.WER
+          slots: 
+              task: "transcribe"
+      -   name: bleu
+          tokenize: flores101
```
nithinraok: for current published models, doesn;t this need to be `bleu_tokenizer` ? You can add other options as a comment?

nemo/collections/asr/metrics/multitask.py
nithinraok: This might not work when prompt scheme doesn;t have task but rather based on other keys like target and source lang?

nemo/collections/asr/metrics/multitask.py
```diff
@@ -0,0 +1,237 @@
...
+            # 1 -> `source_lang` in canary, 3 -> 'target_lang. Use these instead of task ID to avoid lookup.
+            condition_met = prompt_ids[:, 1] != prompt_ids[:, 3]
+        else:  # default to transcribe
+            condition_met = prompt_ids[:, 1] == prompt_ids[:, 3]
+        indices = torch.nonzero(condition_met, as_tuple=False)
+        # reshape in case 0 dim
+        return indices.reshape(indices.numel())
+
+    # TODO: Add properties to `Canary2` to simply return `task` idx.
+    def _canary2_index_fn(self, prompt_ids: torch.Tensor, slots: Dict[str, str | bool]) -> torch.Tensor:
```
nithinraok: This definitely needs good test cases for all possible cases

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
tbartley94: remind me the serialization flow? does it pull the config from the model and then load as a checkpoint or is everything pickled?

tbartley94:
@nithinraok 
> This might not work when prompt scheme doesn;t have task but rather based on other keys like target and source lang?
The index function? That's just prompt based. 

If you're talking about the general flow, the task limitation can be removed with the assert case. Then individual processing of slot constraints can be updated in the respective index function as need requires. We can add it now but it would just be an empty loop over the `slots` constraints for each element of prompt. It's a follow up PR but want to avoid feature creep in this current one.

nemo/collections/asr/metrics/multitask.py
tbartley94: What ya feeling unit test wise? Just assert indexes line up with a dummy lhotse dataloader?

nemo/collections/asr/metrics/multitask.py
tbartley94: nono, task isn't referring to the prompt itself, but just a general `task` keyword that can be used by the indexing function. Even if the prompt doesn't use a task keyword, we'd still want users to simply refer to an assumed task to make config notation easier.

tbartley94:
> tests are missing.

yeah, once the main code is deemed acceptable I can throw in unit tests. (saves rewriting them in case y'all have a fundamental issue with structure)

nemo/collections/asr/models/aed_multitask_models.py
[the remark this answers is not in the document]
tbartley94: @pzelasko @nithinraok see new processing behavior in bleu metrics and multitask. Here it is more coherrent to rely on directly passing cuts.

nemo/collections/asr/metrics/multitask.py
tbartley94: @pzelasko @nithinraok see new behavior for multitask. The config parser is a bit extra but it now allows us to directly pass constraints via config. (Added example string there.) Also with support for not/and/or, it should cover all boolean behavior. I didn't leave in comparatives like <, and >, but technically can be added in if _really_ desired.

nemo/collections/asr/parts/submodules/multitask_beam_decoding.py
pzelasko: no, it's OK

nemo/collections/common/prompts/canary.py
pzelasko: OK to include them

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
```diff
@@ -43,20 +43,19 @@ model:
...
+          constraint: ".source_lang==.target_lang"
       -   name: bleu
-          tokenize: flores101
-          multi_tokenize: true
           _target_: nemo.collections.asr.metrics.BLEU  
-          slots: 
-              task: "translate"
+          constraint: ".source_lang!=.target_lang"
+          bleu_tokenizer: 13a
+          check_cuts_for_tokenizers: true  # Indicates if metric should check batch for per sample SacreBLEU tokenizer.
```
pzelasko: This option is a bit confusing on first sight. Should this be configurable? When would you ever want to set this to `false`?

nemo/collections/asr/metrics/bleu.py
```diff
@@ -183,6 +189,7 @@ def update(
 
                 # TODO: the backend implementation of this has a lot of cpu to gpu operations. Should reimplement
                 # for speedup.
+                print(hypotheses[idx].text, reference)
```
pzelasko: a wild print

nemo/collections/asr/metrics/bleu.py
```diff
@@ -169,7 +171,11 @@ def update(
 
         with torch.no_grad():
             # get predictions
-            hypotheses = self.decode(predictions, predictions_lengths, predictions_mask, input_ids)
+            hypotheses = (
+                self.decode(predictions, predictions_lengths, predictions_mask, input_ids)
+                if predictions.numel() > 0
```
pzelasko, line 176: when can `predictions` be empty?

bonham79: It's an edge case if your data is so lopsided that one rank is logging metrics that another rank never sees. It causes the backend metric sync to hang while waiting for a nonexistent update. It 'shouldnt' happen in most cases but it's not a well logged fail state so best to avoid.

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
bonham79: To run the check I have to iterate over batch again. So adds on some extra overhead per eval step. So useful for user to know if their validations seem to creep and it's unnecessary. 

Should add a doc line in config to  help though.

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
```diff
@@ -42,6 +42,21 @@ model:
...
 
+  # Sub-config for specifying multiple metrics for multitask modeling.
+  # Each metric allows a custom bool constraint to determine conditions for evaluation. 
+  # See `asr.collections.metrics.multitask` for further details.
+  multitask_metrics_cfg:
+    log_predictions: true
+    metrics:
+      -   name: wer
+          _target_: nemo.collections.asr.metrics.WER
+          constraint: ".source_lang==.target_lang"
```
nithinraok: Can we add multiple constraints or only one?

nemo/collections/asr/metrics/bleu.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
nithinraok: minor: No need to change year.

nemo/collections/asr/metrics/bleu.py
```diff
@@ -147,56 +157,76 @@ def update(
...
+                the initial tokenizer type from ``BLEU.__init__`` is used. If ``cuts == None`` then all elements
+                in batch are tokenized with initial tokenizer type.
         """
-        references = []
+        tokenizers = None
+        if self.check_cuts:
+            assert (
+                len(cuts) == targets_lengths.shape[0]
+            ), f"BLEU metrics configured for multiple tokenizers, but got only '{len(cuts)}' samples for '{targets_lengths.shape[0]}' predictions."
+            tokenizers = _get_bleu_tokenizers_from_cuts(cuts)
```
nithinraok, line 170: Does this mean we cannot use BLEU metric without lhotse cuts? It would be better IMO metric to not depend on dataset type.

tbartley94: Only if you're passing to the config `check_cuts_for_bleu_tokenizer`. So you had to make the conscious decision to rely on passing cuts to BLEU.

examples/asr/conf/speech_multitask/fast-conformer_aed.yaml
tbartley94: One constraint but you can use boolean and/or/xor parsing and parentheses nesting. It should be a full propositional logic.

nithinraok:
Please add this info to NeMo docs

nemo/collections/asr/metrics/bleu.py
tbartley94: For multitask we are lhotse bound though. But it's either lhotse bound or prompt bound. We only have lhotse at the moment so safer assumption (Is anyone still using webdataset for Canary?)

nithinraok: No but we would be blocking Bleu metric usage elsewhere

tbartley94: Correct, so they're still free to use bleu. Just need this lhotse setup for the multitokenization deal.  And if they're working at the level of multilingual modeling that they need the multitokenizer setup, they're likely scaling data to the level that NeMo better supports them with lhotse than webdataset. And if they'd rather use another dataloader, they're likely tinkering enough with the code they can swap out a dict like object in the place of the lhotse batch.

tbartley94:
> Please add this info to NeMo docs

done

tbartley94:
@nithinraok Looks like upstream people didn't manage dependencies well and it's causing unrelated failures in CICD tests. Can we just skip it if I screenshot pytests running on my end?

nithinraok:
> @nithinraok Looks like upstream people didn't manage dependencies well and it's causing unrelated failures in CICD tests. Can we just skip it if I screenshot pytests running on my end?

@chtruong814 can you please have a look

chtruong814:
> > @nithinraok Looks like upstream people didn't manage dependencies well and it's causing unrelated failures in CICD tests. Can we just skip it if I screenshot pytests running on my end?
> 
> @chtruong814 can you please have a look

@nithinraok @tbartley94 Is there a certain CI pipeline you're referring to?  The last failure was due to the pipeline getting canceled from a more recent push.

https://github.com/NVIDIA/NeMo/actions/runs/15602808643/job/43990771775

nithinraok (approved):
LGTM. Great work~

chtruong814:
CI passed here.  Will go ahead and merge
https://github.com/NVIDIA/NeMo/actions/runs/15718713187

PR #13545, opened 2025-05-12
Add  Féchet codec distance metric

rfejgin:
This PR adds the Fréchet Codec Distance metric. This is a new metric we are experimenting with which is based on the Fréchet Inception Distance (FID).

We compute the distance between the distribution of real codec frames and generated codec frames. The distance is computed in the (dequantized) embedding space of the codec.

Note that FD distances are computed at the dataset level. The number of real and generated frames should large for a reliable metric.

The metric is now called from `infer_and_evaluate.py` and included in our standard set of reported metrics.


rfejgin:
Hitting issues due to lack of "Signed-off by" string in some of the commits; will try to fix and recreate this PR. Closing.

PR #13553, opened 2025-05-13
Squashed commit of the following:

rfejgin:
This PR adds the Fréchet Codec Distance metric. This is a new metric we are experimenting with which is based on the Fréchet Inception Distance (FID).

We compute the distance between the distribution of real codec frames and generated codec frames. The distance is computed in the (dequantized) embedding space of the codec.

Note that FD distances are computed at the dataset level. The number of real and generated frames should large for a reliable metric.

The metric is now called from `infer_and_evaluate.py` and included in our standard set of reported metrics.

### DIFF
--- a/nemo/collections/tts/modules/fcd_metric.py
+++ b/nemo/collections/tts/modules/fcd_metric.py
@@ -1,0 +1,263 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import warnings
+
+import torch
+import torch.nn.functional as F
+from torch import nn, Tensor
+from torchmetrics import Metric
+
+from nemo.collections.tts.models import AudioCodecModel
+from nemo.collections.tts.parts.utils.tts_dataset_utils import _read_audio
+
+
+class CodecEmbedder(nn.Module):
+    """
+    Embeds audio codec codes into the codec's continuous embedding space.
+    Accepts as input either a batch of codes or a path to an audio file.
+    """
+    def __init__(self, codec: AudioCodecModel):
+        super().__init__()
+        self.codec = codec
+
+    def codes_to_emedding(self, x: Tensor, x_len: Tensor) -> Tensor:
+        """
+        Embeds a batch of audio codec codes into the codec's continuous embedding space.
+        """
+        # x: (B, T, C)
+        # x_len: (B,)
+        return self.codec.dequantize(tokens=x, tokens_len=x_len)
+
+    def encode_from_file(self, audio_path: str) -> Tensor:
+        """
+        Encodes an audio file into audio codec codes.
+        """
+        print(f"Encoding audio {audio_path}")
+        audio_segment = _read_audio(
+            audio_filepath=audio_path, sample_rate=self.codec.sample_rate, offset=0, duration=0
+        )
+        samples = samples = torch.tensor(audio_segment.samples, device=self.codec.device).unsqueeze(0)
+        audio_len = torch.tensor(samples.shape[1], device=self.codec.device).unsqueeze(0)
+        codes, codes_len = self.codec.encode(audio=samples, audio_len=audio_len)
+        return codes, codes_len
+
+
+class FrechetCodecDistance(Metric):
+    """
+    Computes the Frechet Codec Distance between two distributions of audio codec frames (real and generated).
+    This is done in codec embedding space, one frame at a time. We name this metric the Frechet Codec Distance (FCD).
+    """
+
+    """
+    Parts of this are based on the following implementation of FID (Frechet Inception Distance) on images:
+    
+        https://github.com/pytorch/torcheval/blob/main/torcheval/metrics/image/fid.py
+
+        # Copyright (c) Meta Platforms, Inc. and affiliates.
+        # All rights reserved.
+        #
+        # This source code is licensed under the BSD-style license found in the
+        # LICENSE file in the root directory of this source tree.
+
+    Contents of original LICENSE file:
+
+        #  BSD License
+        #                
+        #  For torcheval software
+        #
+        #  Copyright (c) Meta Platforms, Inc. and affiliates. All rights reserved.
+        #
+        #  Redistribution and use in source and binary forms, with or without modification,
+        #  are permitted provided that the following conditions are met:
+        #         
+        #  * Redistributions of source code must retain the above copyright notice, this
+        #  list of conditions and the following disclaimer.
+        #
+        #  * Redistributions in binary form must reproduce the above copyright notice,
+        #  this list of conditions and the following disclaimer in the documentation
+        #  and/or other materials provided with the distribution.
+        #
+        #  * Neither the name Meta nor the names of its contributors may be used to
+        #  endorse or promote products derived from this software without specific
+        #  prior written permission.
+        #
+        #  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+        #  ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+        #  WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+        #  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+        #  ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+        #  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+        #  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+        #  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+        #  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+        #  SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+    """
+    is_differentiable = False
+    higher_is_better = False
+    full_state_update = False
+
+    def __init__(
+        self,
+        codec,
+        feature_dim: int,
+    ) -> None:
+        """
+        Computes the Frechet Codec Distance between two distributions of audio codec codes (real and generated).
+        The original paper (FID on images): https://arxiv.org/pdf/1706.08500.pdf
+
+        Args:
+            codec (AudioCodecModel): The codec model to use.
+            feature_dim (int): The number of features in the codec embedding space (usually 4*num_codebooks)
+        """
+        super().__init__()
+
+        # Set the model and put it in evaluation mode
+        self.model = CodecEmbedder(codec)
+        self.model.eval()
+        self.model.requires_grad_(False)
+
+        # Initialize state variables used to compute FCD
+        self.add_state("real_sum", default=torch.zeros(feature_dim), dist_reduce_fx="sum")
+        self.add_state("real_cov_sum", default=torch.zeros((feature_dim, feature_dim)), dist_reduce_fx="sum")
+        self.add_state("fake_sum", default=torch.zeros(feature_dim), dist_reduce_fx="sum")
+        self.add_state("fake_cov_sum", default=torch.zeros((feature_dim, feature_dim)), dist_reduce_fx="sum")
+        self.add_state("num_real_frames", default=torch.tensor(0).int(), dist_reduce_fx="sum")
+        self.add_state("num_fake_frames", default=torch.tensor(0).int(), dist_reduce_fx="sum")
+
+    def update_from_audio_file(self, audio_path: str, is_real: bool) -> Tensor:
+        """
+        Takes a path to an audio file, embeds it, and updates the FCD metric.
+        """
+        codes, codes_len = self.model.encode_from_file(audio_path)
+        self.update_from_codes(codes, codes_len, is_real)
+
+    def update(self, codes: Tensor, codes_len: Tensor, is_real: bool):
+        # The torchmetrics.Metric interface requires an `update()` method which we override here.
+        # Alias for `update_from_codes()`
+        self.update_from_codes(codes, codes_len, is_real)
+
+    def update_from_codes(self, codes: Tensor, codes_len: Tensor, is_real: bool):
+        """
+        Update the states with a batch of real and fake codes.
+        Takes pre-computed codec codes, embeds them, and updates the FCD metric.
+
+        Args:
+            codes (Tensor): A batch of codec frames of shape (B, C, T).
+            is_real (Boolean): Denotes if images are real or not.
+        """
+        assert codes.ndim == 3
+        codes = codes.to(self.device)
+
+        # Dequantize the codes to a continuous representation
+        embeddings = self.model.codes_to_emedding(
+            codes, codes_len
+        )  # B, E, T where E is the codec's embedding dimension, usually 4*num_codebooks
+
+        # keep only the valid frames
+        valid_frames = []
+        for i in range(codes.shape[0]):
+            valid_frames.append(embeddings[i, :, : codes_len[i]].T)  # T', E
+        embeddings = torch.cat(valid_frames, dim=0)  # total_valid_frames, E
+        valid_frame_count = embeddings.shape[0]
+
+        # Update the state variables used to compute FCD
+        if is_real:
+            self.num_real_frames += valid_frame_count
+            self.real_sum += torch.sum(embeddings, dim=0)
+            self.real_cov_sum += torch.matmul(embeddings.T, embeddings)
+        else:
+            self.num_fake_frames += valid_frame_count
+            self.fake_sum += torch.sum(embeddings, dim=0)
+            self.fake_cov_sum += torch.matmul(embeddings.T, embeddings)
+
+        return self
+
+    def compute(self) -> Tensor:
+        """
+        Compute the FCD.
+
+        Returns:
+            tensor: The FCD.
+        """
+
+        # If the user has not already updated with at lease one
+        # image from each distribution, then we raise an Error.
+        if (self.num_real_frames == 0) or (self.num_fake_frames == 0):
+            warnings.warn(
+                "Computing FD requires at least 1 real image and 1 fake image,"
+                f"but currently running with {self.num_real_frames} real images and {self.num_fake_frames} fake images."
+                "Returning 0.0",
+                RuntimeWarning,
+            )
+
+            return torch.tensor(0.0)
+
+        # Compute the mean activations for each distribution
+        real_mean = (self.real_sum / self.num_real_frames).unsqueeze(0)
+        fake_mean = (self.fake_sum / self.num_fake_frames).unsqueeze(0)
+
+        # Compute the covariance matrices for each distribution
+        real_cov_num = self.real_cov_sum - self.num_real_frames * torch.matmul(real_mean.T, real_mean)
+        real_cov = real_cov_num / (self.num_real_frames - 1)
+        fake_cov_num = self.fake_cov_sum - self.num_fake_frames * torch.matmul(fake_mean.T, fake_mean)
+        fake_cov = fake_cov_num / (self.num_fake_frames - 1)
+
+        # Compute the Frechet Distance between the distributions
+        fd = self.calculate_frechet_distance(real_mean.squeeze(), real_cov, fake_mean.squeeze(), fake_cov)
+        # FD should be non-negative but due to numerical errors, it can be slightly negative
+        # Have seen -0.0011 in the past
+        assert fd >= -0.005
+        return torch.max(torch.tensor(0.0), fd)
+
+    def calculate_frechet_distance(
+        self,
+        mu1: Tensor,
+        sigma1: Tensor,
+        mu2: Tensor,
+        sigma2: Tensor,
+    ) -> Tensor:
+        """
+        Calculate the Frechet Distance between two multivariate Gaussian distributions.
+
+        Args:
+            mu1 (Tensor): The mean of the first distribution.
+            sigma1 (Tensor): The covariance matrix of the first distribution.
+            mu2 (Tensor): The mean of the second distribution.
+            sigma2 (Tensor): The covariance matrix of the second distribution.
+
+        Returns:
+            tensor: The Frechet Distance between the two distributions.
+        """
+
+        # Compute the squared distance between the means
+        mean_diff = mu1 - mu2
+        mean_diff_squared = mean_diff.square().sum(dim=-1)
+
+        # Calculate the sum of the traces of both covariance matrices
+        trace_sum = sigma1.trace() + sigma2.trace()
+
+        # Compute the eigenvalues of the matrix product of the real and fake covariance matrices
+        sigma_mm = torch.matmul(sigma1, sigma2)
+        eigenvals = torch.linalg.eigvals(sigma_mm)
+
+        # Take the square root of each eigenvalue and take its sum
+        sqrt_eigenvals_sum = eigenvals.sqrt().real.sum(dim=-1)
+
+        # Calculate the FCD using the squared distance between the means,
+        # the sum of the traces of the covariance matrices, and the sum of the square roots of the eigenvalues
+        fcd = mean_diff_squared + trace_sum - 2 * sqrt_eigenvals_sum
+
+        return fcd
+
--- a/scripts/magpietts/evaluate_generated_audio.py
+++ b/scripts/magpietts/evaluate_generated_audio.py
@@ -8,9 +8,11 @@ import torch
 
 import nemo.collections.asr as nemo_asr
 from nemo.collections.asr.metrics.wer import word_error_rate_detail
+from nemo.collections.tts.modules.fcd_metric import FrechetCodecDistance
+from nemo.collections.tts.models import AudioCodecModel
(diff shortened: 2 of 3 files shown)


nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        Args:
+            codes (Tensor): A batch of codec frames of shape (B, C, T).
+            is_real (Boolean): Denotes if images are real or not.
+        """
+        assert codes.ndim == 3
+        codes = codes.to(self.device)
+
+        # Dequantize the codes to a continuous representation
+        embeddings = self.model.codes_to_emedding(
+            codes, codes_len
```
XuesongYang, line 165: needs clarifications on the shapes: 
* the input `codes` of `CodeEmbedder.codes_to_embedding()` is defined in shape of `(B, T, C)` as
described in its docstring, while
here `codes` has the shape of `(B, C, T)` in docstring. I believe here `C` denotes the number of codebooks, which is
equivalent to vector quantizer's `D`, e.g. in `GroupFiniteScalarQuantizer`(https://github.com/NVIDIA/NeMo/blob/556d059953ce349c4d088fac341e4b00f91fd5d2/nemo/collections/tts/modules/audio_codec_modules.py#L1440). 
* again, `CodeEmbedder.codes_to_embedding()` directly calls `self.codec.dequantizer()` which expects `codes` in shape of
  `(B, C, T)`.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        Takes a path to an audio file, embeds it, and updates the FCD metric.
+        """
+        codes, codes_len = self.model.encode_from_file(audio_path)
+        self.update_from_codes(codes, codes_len, is_real)
+
+    def update(self, codes: Tensor, codes_len: Tensor, is_real: bool):
+        # The torchmetrics.Metric interface requires an `update()` method which we override here.
+        # Alias for `update_from_codes()`
+        self.update_from_codes(codes, codes_len, is_real)
```
XuesongYang, line 149: nit: why not directly move the codes of `update_from_codes` here?

XuesongYang (changes requested):
LGTM in general, but i haven't tested running eval with it yet.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        real_cov = real_cov_num / (self.num_real_frames - 1)
+        fake_cov_num = self.fake_cov_sum - self.num_fake_frames * torch.matmul(fake_mean.T, fake_mean)
+        fake_cov = fake_cov_num / (self.num_fake_frames - 1)
+
+        # Compute the Frechet Distance between the distributions
+        fd = self.calculate_frechet_distance(real_mean.squeeze(), real_cov, fake_mean.squeeze(), fake_cov)
+        # FD should be non-negative but due to numerical errors, it can be slightly negative
+        # Have seen -0.0011 in the past
+        assert fd >= -0.005
+        return torch.max(torch.tensor(0.0), fd)
```
XuesongYang, line 222: some optimization: Ensured that `torch.tensor(0.0)` is created on the correct device.
```suggestion
        return torch.max(torch.tensor(0.0, device=fd.device), fd)
```

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -184,7 +198,10 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
                 pred_context_ssim_alternate = torch.nn.functional.cosine_similarity(pred_speaker_embedding_alternate, context_speaker_embedding_alternate, dim=0).item()
                 gt_context_ssim_alternate = torch.nn.functional.cosine_similarity(gt_speaker_embedding_alternate, context_speaker_embedding_alternate, dim=0).item()
 
-
+        # update FCD metric for all generated codes
+        if fcd_metric is not None:
+            for i in range(predicted_codes.shape[0]):
+                fcd_metric.update_from_codes(predicted_codes[i:i+1], predicted_codes_lens[i:i+1], False)
```
XuesongYang, line 204: this code block loops to update FCD metric one sample a time, which seems not necessary. `FCD.update_from_codes` already support batched input, doesn't it?

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -112,7 +115,15 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
     speaker_verification_model_alternate = speaker_verification_model_alternate.to(device)
     speaker_verification_model_alternate.eval()
 
-
+    if codecmodel_path is not None:
+        codec = AudioCodecModel.restore_from(codecmodel_path, strict=False)
+        codec = codec.to(device)
+        codec.eval()
```
XuesongYang, line 121: No need to run `eval()` here again? it would be run inside `FCD` class initialization.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        self.model.eval()
+        self.model.requires_grad_(False)
+
+        # Initialize state variables used to compute FCD
+        self.add_state("real_sum", default=torch.zeros(feature_dim), dist_reduce_fx="sum")
+        self.add_state("real_cov_sum", default=torch.zeros((feature_dim, feature_dim)), dist_reduce_fx="sum")
+        self.add_state("fake_sum", default=torch.zeros(feature_dim), dist_reduce_fx="sum")
+        self.add_state("fake_cov_sum", default=torch.zeros((feature_dim, feature_dim)), dist_reduce_fx="sum")
+        self.add_state("num_real_frames", default=torch.tensor(0).int(), dist_reduce_fx="sum")
+        self.add_state("num_fake_frames", default=torch.tensor(0).int(), dist_reduce_fx="sum")
```
XuesongYang, line 137: there is a waste to cast to int32 by doing `Python 0 -> torch.int64 scalar tensor -> torch.int32 scalar tensor`. Instead, we can directly `torch.tensor(0, dytpe=torch.int)`

XuesongYang:
One question about why we want to implement our own Frechet distance between two multivariate Gaussian distribution, instead of re-using existing one. TorchAudio has implemented one as in `https://docs.pytorch.org/audio/2.7.0/generated/torchaudio.functional.frechet_distance.html?highlight=frechet#torchaudio.functional.frechet_distance`. We just need to feed in the dequantized codec embeddings as inputs. 

Frechet Distance applied to speech/audio domain is not new. There are two work has applied such concept,
1. Frechet Speech Distance in voicebox: https://proceedings.neurips.cc/paper_files/paper/2023/file/2d8911db9ecedf866015091b28946e15-Paper-Conference.pdf
2. Frechet Audio Distance in audio: https://www.isca-archive.org/interspeech_2019/kilgour19_interspeech.pdf

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        # image from each distribution, then we raise an Error.
+        if (self.num_real_frames == 0) or (self.num_fake_frames == 0):
+            warnings.warn(
+                "Computing FD requires at least 1 real image and 1 fake image,"
+                f"but currently running with {self.num_real_frames} real images and {self.num_fake_frames} fake images."
+                "Returning 0.0",
+                RuntimeWarning,
+            )
+
+            return torch.tensor(0.0)
```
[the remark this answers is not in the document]
rfejgin: I didn't want the FCD to be the cause of inference scripts breaking in case people do end up running them the scripts without a ground truth available, which would result in `num_real_frames`. I think it should be clear enough that if the FCD is 0.0 it's not a real value. So I suggest to keep this as is.

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Good catch! this is a typo in the comments of `CodeEmbedder.codes_to_embedding()`. Fixing the comment in the upcoming commit.

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Heh, I though it would be clearer this way to distinguish between the two flavors this metric can be used with (update from codes or from audio file). But I guess it is not clearer. I will update as suggested

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Yes; it would work without this eval() as well. I am following the template of how we handle all the other models (ASR, speaker embedding, etc.) in `evaluate_generated_audio()` - moving them to device and setting to eval mode. I didn't want to make too many assumptions about where this codec instance would be used; e.g. if anyone ends up using it directly in evaluate_generated_audio() it might not be in eval mode if the metric isn't being used. So I'd prefer to keep this in for safety.

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Interesting. I'll change it as suggested

rfejgin:
> One question about why we want to implement our own Frechet distance (eigenvalues of covariances, traces) between two multivariate Gaussian distribution, instead of re-using existing one. TorchAudio has implemented one as in https://docs.pytorch.org/audio/2.7.0/generated/torchaudio.functional.frechet_distance.html?highlight=frechet#torchaudio.functional.frechet_distance. We just need to feed in the dequantized codec embeddings as inputs.
> 
> Frechet Distance applied to speech/audio domain is not new. There are two work has applied such concept. The only difference is to feed in different kinds of embedding vectors.
> 
> 1. Frechet Speech Distance in voicebox: https://proceedings.neurips.cc/paper_files/paper/2023/file/2d8911db9ecedf866015091b28946e15-Paper-Conference.pdf
> 2. Frechet Audio Distance in audio: https://www.isca-archive.org/interspeech_2019/kilgour19_interspeech.pdf

Definitely, Frechet distance is not new in audio -- FAD (using `VGGish` as the embedder) in particular has been widely used. What is less common (although also not entirely new, e.g. in https://arxiv.org/html/2407.10387v1) is to use an audio codec's embedding space as the FD embedding. I think that makes sense for our use case because we are in specifically interested in making sure that MaskGit's various sampling methods (parallel in particular) don't end up generating codec frames that are out-of-distribution for the codec, which could throw its synthesis off.

Regarding our implementation: yes, there are a few lines of the core distance computation (`calculate_frechet_distance()`) that we could reuse from torch audio. It would amount to [4 lines](https://github.com/pytorch/audio/blob/ea5de17755d657508c84c4dce8970b614008adcf/src/torchaudio/functional/functional.py#L2532) plus some safety checks so not sure if it's worth it...

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+class CodecEmbedder(nn.Module):
+    """
+    Embeds audio codec codes into the codec's continuous embedding space.
+    Accepts as input either a batch of codes or a path to an audio file.
+    """
+    def __init__(self, codec: AudioCodecModel):
+        super().__init__()
+        self.codec = codec
+
+    def codes_to_emedding(self, x: Tensor, x_len: Tensor) -> Tensor:
```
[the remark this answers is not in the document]
rfejgin: fixed in latest commit

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        # x: (B, T, C)
+        # x_len: (B,)
+        return self.codec.dequantize(tokens=x, tokens_len=x_len)
+
+    def encode_from_file(self, audio_path: str) -> Tensor:
+        """
+        Encodes an audio file into audio codec codes.
+        """
+        print(f"Encoding audio {audio_path}")
```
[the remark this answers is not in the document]
rfejgin: Agreed, removed in latest commit

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        audio_segment = _read_audio(
+            audio_filepath=audio_path, sample_rate=self.codec.sample_rate, offset=0, duration=0
+        )
```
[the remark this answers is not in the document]
rfejgin: In the latest commit I've changed this to directly call `AudioSegment.from_file()`. I kept it with `int_value=True` for now until we hear the FLAC issue is no longer present.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,263 @@
...
+        samples = samples = torch.tensor(audio_segment.samples, device=self.codec.device).unsqueeze(0)
```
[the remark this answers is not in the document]
rfejgin: Good eye :) Fixed in latest commit.

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Fixed in latest commit

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Fixed in latest commit

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Fixed in latest commit

nemo/collections/tts/modules/fcd_metric.py
rfejgin: Fixed in latest commit

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -112,7 +115,15 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
...
+        codec_feature_dim = codec.vector_quantizer.codebook_dim
+        fcd_metric = FrechetCodecDistance(codec=codec, feature_dim=codec_feature_dim).to(device)
```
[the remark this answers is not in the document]
rfejgin: This is a somewhat confusing aspect of the codec API. The `codebook_dim` is the codec's embedding dimension (combined across all codebooks). So e.g. if there are 8 codebooks and 4 groups, then `codebook_dim == 8*4 == 32`.

rfejgin:
I believe I've addressed or commented on all review comments - this is ready for re-review.

scripts/magpietts/evaluate_generated_audio.py
XuesongYang: ok. let's keep as-is for model evaluation. 

please keep in mind that all the other models (like ASR, speaker embedding, etc.) run  .eval() once followed by inference. The script run .eval() once after instantiating codec model and run codec model .eval() again when initializing `FrechetCodecDistance`, which is still not necessary.

scripts/magpietts/evaluate_generated_audio.py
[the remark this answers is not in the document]
XuesongYang: thanks for adding the comment for clarification.

XuesongYang:
add @blisc for the final walkthrough.

nemo/collections/tts/modules/fcd_metric.py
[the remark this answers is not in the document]
rfejgin: In that case I guess I can rid of reading as integers. I've changed that in the latest commit and added some safety checks to make sure the read audio looks valid.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,269 @@
...
+    def compute(self) -> Tensor:
+        """
+        Compute the FCD.
+
+        Returns:
+            tensor: The FCD.
+        """
+
+        # If the user has not already updated with at lease one
+        # image from each distribution, then we raise an Error.
```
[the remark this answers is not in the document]
rfejgin: yes - fixed in latest commit

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,269 @@
...
+        real_cov_num = self.real_cov_sum - self.num_real_frames * torch.matmul(real_mean.T, real_mean)
+        real_cov = real_cov_num / (self.num_real_frames - 1)
+        fake_cov_num = self.fake_cov_sum - self.num_fake_frames * torch.matmul(fake_mean.T, fake_mean)
+        fake_cov = fake_cov_num / (self.num_fake_frames - 1)
+
+        # Compute the Frechet Distance between the distributions
+        fd = self.calculate_frechet_distance(real_mean.squeeze(), real_cov, fake_mean.squeeze(), fake_cov)
+        # FD should be non-negative but due to numerical errors, it can be slightly negative
+        # Have seen -0.0011 in the past
+        assert fd >= -0.005
```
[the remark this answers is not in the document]
rfejgin: It would be hard to ensure the unit test encounters numerical issues like this, if they even exist, so I'd prefer to keep the check in the code. I prefer to keep an assert rather than log because we would very easily miss such a warning message that potentially happens rarely.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,269 @@
...
+        return torch.max(torch.tensor(0.0, device=self.device), fd)
```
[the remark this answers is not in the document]
rfejgin: Thanks, yeah, that would be cleaner. Fixed in latest commit.

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -277,6 +290,9 @@ def run_inference(
                 language=language,
                 sv_model_type=sv_model,
                 asr_model_name=asr_model_name,
+                predicted_codes=predicted_codes,
+                predicted_codes_lens=predicted_codes_lens,
```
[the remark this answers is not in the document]
rfejgin: Crucial catch, thank you!! Indeed, it looks like the calling loop in `infer_and_evaluate.py` doesn't accumulate the predicted codes across batches. Might explain the higher-than-expected FCD variability I've been seeing... 😅.

I'll look into how to accumulate the codes in infer_and_evaluate (or write them to file).

rfejgin:
Putting back in draft mode to address important comments.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: I think running eval() a second time on a model that's already in eval mode should be very cheap. I prefer to keep this in place also in `FrechetCodecDistance` so that the module doesn't need to make the assumption that it's already been done before it was given to it.

scripts/magpietts/infer_and_evaluate.py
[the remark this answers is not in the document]
rfejgin: I have fixed this now. We now write the generated code to file, like we do for the audio wav files. I didn't want to accumulate them in memory both to limit memory usage on a large set (probably not a big issue) and also to make the association between the `.wav` and codec code file easier to keep track of.

After evaluation we delete the individual codec files to reduce clutter.

And indeed, from some initial testing I am seeing much reduced variability (between runs) in FCD scores now! Nice.

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -184,7 +216,9 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
                 pred_context_ssim_alternate = torch.nn.functional.cosine_similarity(pred_speaker_embedding_alternate, context_speaker_embedding_alternate, dim=0).item()
                 gt_context_ssim_alternate = torch.nn.functional.cosine_similarity(gt_speaker_embedding_alternate, context_speaker_embedding_alternate, dim=0).item()
 
-
+        # update FCD metric for all generated codes
+        if fcd_metric is not None:
+            fcd_metric.update(predicted_codes, predicted_codes_lens, False)
```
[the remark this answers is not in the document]
rfejgin: No longer an issue in latest implementation now that I've switched to saving and loading codec files from disk.

rfejgin:
Undoing draft state since key issues have been addressed.

nemo/collections/tts/modules/fcd_metric.py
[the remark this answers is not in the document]
blisc: I agree with Ryan. We should not have this assert statement in the code. Whenever we raise an error, we should inform the user why this error occurs via a log message. If it's not fixable by the user, then we should not have this error pop up when the code is run.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,270 @@
...
+class CodecEmbedder(nn.Module):
+    """
+    Embeds audio codec codes into the codec's continuous embedding space.
+    Accepts as input either a batch of codes or a path to an audio file.
+    """
+    def __init__(self, codec: AudioCodecModel):
+        super().__init__()
+        self.codec = codec
+
+    def codes_to_embedding(self, x: Tensor, x_len: Tensor) -> Tensor:
```
rlangman, line 47: A long time ago, after finding that codec research itself has no consistency in terminology, we decided NeMo code should use `tokens` to refer to quantized indexes and `codes` to refer to their corresponding numerical values/embedding. Should we try to maintain this convention?

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -163,6 +201,12 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
         gt_texts.append(gt_text)
         gt_audio_texts.append(gt_audio_text)
 
+        # update FCD metric
+        if fcd_metric is not None:
+            predicted_codes = torch.load(pred_codes_filepath).unsqueeze(0)
+            predicted_codes_lens = torch.tensor([predicted_codes.size(1)], dtype=torch.int, device=device)
+            fcd_metric.update(predicted_codes, predicted_codes_lens, False)
```
rlangman, line 208: `update` requires `predicted_codes` be `(B, C, T)` but this is using `size(1)` as the length?

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -8,16 +22,21 @@
...
+def find_generated_files(audio_dir, file_type):
+    assert file_type in ["audio", "codes"], f"Invalid file type: {file_type}"
+    extension = ".wav" if file_type == "audio" else ".pt"
+    prefix = "predicted_audio" if file_type == "audio" else "predicted_codes"
     file_list = []
     for f in os.listdir(audio_dir):
-        if "predicted_audio" in f and f.endswith(".wav"):
-            audio_number = int(f.split("_")[-1].split(".wav")[0])
+        if prefix in f and f.endswith(extension):
+            audio_number = int(f.split("_")[-1].split(extension)[0])
```
rlangman, line 39: Consider using a pattern like:

```
find_gneerated_files(audio_dir, prefix, extension):
    ...

find_generated_audio_files(audio_dir):
    return find_generated_files(audio_dir, "predicted_audio", ".wav")

find_generated_codec_files(audio_dir):
    return find_generated_files(audio_dir, "predicted_codes", ".pt")
```

blisc (approved):
Synced offline. Should be good to merge once tests pass

scripts/magpietts/evaluate_generated_audio.py
rfejgin: You are right! Fixed.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,270 @@
...
+        Encodes an audio file into audio codec codes.
+        """
+        audio_segment = AudioSegment.from_file(
+            audio_path, target_sr=self.codec.sample_rate, offset=0, duration=0
+        )
+        assert np.issubdtype(audio_segment.samples.dtype, np.floating)
+        audio_min = audio_segment.samples.min()
+        audio_max = audio_segment.samples.max()
+        if audio_min <= -1.0 or audio_max >= 1.0:
+            logging.warning(f"Audio samples are not normalized to [-1, 1]: min={audio_min}, max={audio_max}")
```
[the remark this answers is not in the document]
rfejgin: The main goal here wasn't to catch samples slightly out of range but to make sure `AudioSegment.from_file()` returned normalized samples as opposed to a floating point version of the signed ints usually find in WAV files, e.g.`32768.0`. I've updated the warning to be a bit less specific about the exact range, just that it is normalized, and also allowed a small tolerance since certain ways of normalizing samples can result in floats slightly smaller than -1.0.

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,270 @@
...
+        # x_len: (B,)
+        return self.codec.dequantize(tokens=x, tokens_len=x_len)
+
+    def encode_from_file(self, audio_path: str) -> Tensor:
+        """
+        Encodes an audio file into audio codec codes.
+        """
+        audio_segment = AudioSegment.from_file(
+            audio_path, target_sr=self.codec.sample_rate, offset=0, duration=0
+        )
```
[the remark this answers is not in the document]
rfejgin: Removed in latest commit

nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -0,0 +1,270 @@
...
+        self.update(codes, codes_len, is_real)
+
+    def update(self, codes: Tensor, codes_len: Tensor, is_real: bool):
+        """
+        Update the states with a batch of real and fake codes.
+        Takes pre-computed codec codes, embeds them, and updates the FCD metric.
+
+        Args:
+            codes (Tensor): A batch of codec frames of shape (B, C, T).
+            is_real (Boolean): Denotes if samples are real or not.
```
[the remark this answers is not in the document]
rfejgin: Good catch, added in latest commit

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Thanks for the suggestion; I was trying to encapsulate the knowledge of the prefixes inside the function; but I do like the style your suggestion better – updated accordingly now.

nemo/collections/tts/modules/fcd_metric.py
rfejgin: hmm... I find "codes" to be a strange name for the continuous values (embeddings)... Maybe it's just me? 🤔

rlangman: Whether it seems strange is probably dependent on past experience. We had decided on this terminology for the quantized values:

(1) Raw encoder output/embedding with no quantization noise (we call "encoded')
(2) Quantized embeddings with quantization noise (we call "dequantized" or "codes")
(3) The discrete integer index (we call "tokens", which is consistent with other domains like NLP).

But as mentioned, the terminology has little to no consistency across research groups/papers. I think most commonly people refer to all of these values as "codes" while randomly using the term "tokens" sometimes. The problem is just when writing/reading code, the code becomes hard to read without clearly distinguishing them.

No need to address this hear, but worth thinking about if we were to merge this to main.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Unit tests were run by CI and passed. I think we can merge now unless we care about the CI warning regarding lack of sign-off of one of the commits.

PR #13594, opened 2025-05-15
BPE char tokenizer

shehzeen:
Work in progress

### DIFF
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -1118,7 +1118,11 @@ class AggregatedTTSTokenizer:
         self.tokens = tokens
         self.tokenizer_names = tokenizer_names
         self.toknizer_offsets = toknizer_offsets
-        self.pad = self.tokenizers[tokenizer_names[0]].pad  # Use the first tokenizer's pad token
+        first_tokenizer = self.tokenizers[tokenizer_names[0]]
+        if hasattr(first_tokenizer, "pad_token_id"):
+            self.pad = first_tokenizer.pad_token_id
+        else:
+            self.pad = self.tokenizers[tokenizer_names[0]].pad  # Use the first tokenizer's pad token
 
     def encode(self, text: str, tokenizer_name: str) -> List[int]:
         tokenizer = self.tokenizers[tokenizer_name]
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -48,7 +48,8 @@ from nemo.collections.tts.parts.utils.tts_dataset_utils import stack_tensors
 from nemo.core.classes import ModelPT
 from nemo.core.classes.common import PretrainedModelInfo
 from nemo.utils import logging
-
+from torch import Tensor
+from nemo.core.classes.module import NeuralModule
 
 def worker_init_fn(worker_id):
     # For mp.set_start_method("spawn", force=True)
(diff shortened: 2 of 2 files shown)


nemo/collections/tts/models/magpietts.py
```diff
@@ -62,6 +63,90 @@ def worker_init_fn(worker_id):
...
+    return subword_id_to_char_ids, char_vocab, subword_padding_idx
+
+
+def sequence_mask(lengths: Tensor, max_length: int | None = None) -> Tensor:
+    if max_length is None:
+        max_length = int(lengths.max().item())
+    x = torch.arange(max_length, dtype=lengths.dtype, device=lengths.device)
+    return x.unsqueeze(0) < lengths.unsqueeze(1)
+
+class CharAwareSubwordEncoder(NeuralModule):
```
blisc, line 91: Can you move this to https://github.com/NVIDIA/NeMo/blob/magpietts_2503/nemo/collections/tts/modules/magpietts_modules.py?
And add a docstring

nemo/collections/tts/models/magpietts.py
```diff
@@ -62,6 +63,90 @@ def worker_init_fn(worker_id):
     dataset.text_tokenizer = tokenizer
     dataset.text_conditioning_tokenizer = text_conditioning_tokenizer
 
+def build_vocabs(subword_vocab_items):
```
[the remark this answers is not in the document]
shehzeen: Done.

nemo/collections/tts/models/magpietts.py
```diff
@@ -153,7 +239,20 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+            
+            if self.use_bpe_char_tokenizer:
+                # BPE char tokenizer
+                assert len(self.tokenizer.tokenizers) == 1, "BPE char tokenizer should only be used with one tokenizer"
+                tokenizer_name = self.tokenizer.tokenizer_names[0]
+                tokenizer = self.tokenizer.tokenizers[tokenizer_name]
+                subword_vocab = tokenizer.get_vocab()
+                subword_vocab['<BOS>'] = self.bos_id
+                subword_vocab['<EOS>'] = self.eos_id
+                self.cas_encoder = CharAwareSubwordEncoder(cfg.embedding_dim, subword_vocab.items())
```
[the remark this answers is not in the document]
shehzeen: I have now handled this, by making a separate special token vocab. Each special token will be added to the char vocab. So now the BPE special token will be mapped to a single char token (itself).

nemo/collections/tts/models/magpietts.py
```diff
@@ -62,6 +63,90 @@ def worker_init_fn(worker_id):
...
+
+    assert max(subword_id_to_char_ids) == len(subword_id_to_char_ids) - 1
+    # add padding_idx
+    subword_padding_idx = len(subword_id_to_char_ids)
+    subword_id_to_char_ids[subword_padding_idx] = (len(char_vocab),)
+
+    return subword_id_to_char_ids, char_vocab, subword_padding_idx
+
+
+def sequence_mask(lengths: Tensor, max_length: int | None = None) -> Tensor:
```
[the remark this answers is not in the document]
shehzeen: Reused nemo's get_mask_from_lengths.

nemo/collections/tts/models/magpietts.py
shehzeen: Done.

shehzeen:
> To run this, do we need to change these instances as well in the config? `+train_ds_meta.libri100train.tokenizer_names="[english_phoneme]"`

Yes. this should be changed to +train_ds_meta.libri100train.tokenizer_names="[qwen]"

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1118,7 +1118,11 @@ def __init__(self, tokenizers: List[Union[BaseTokenizer, PreTrainedTokenizerBase
         self.tokens = tokens
         self.tokenizer_names = tokenizer_names
         self.toknizer_offsets = toknizer_offsets
-        self.pad = self.tokenizers[tokenizer_names[0]].pad  # Use the first tokenizer's pad token
+        first_tokenizer = self.tokenizers[tokenizer_names[0]]
+        if hasattr(first_tokenizer, "pad_token_id"):
+            self.pad = first_tokenizer.pad_token_id
+        else:
+            self.pad = self.tokenizers[tokenizer_names[0]].pad  # Use the first tokenizer's pad token
```
blisc, line 1125:
```suggestion
        # Define aggregated token's pad value from the first tokenizer's pad value
        first_tokenizer = self.tokenizers[tokenizer_names[0]]
        if hasattr(first_tokenizer, "pad_token_id"):  # Defined in PreTrainedTokenizerBase subclasses
            self.pad = first_tokenizer.pad_token_id
        elif hasattr(first_tokenizer, "pad"):  # Defined in BaseTokenizer subclasses
            self.pad = first_tokenizer.pad
        else:
           raise ValueError("AggregatedTTSTokenizer could not find a padding token in the first tokenizer")
```

nemo/collections/tts/models/magpietts.py
```diff
@@ -153,7 +146,29 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+                }
+                self.cas_encoder = CharAwareSubwordEncoder(
+                    d_embed=cfg.embedding_dim,
+                    llm_tokenizer_vocab=subword_vocab,
+                    subword_padding_idx=self.tokenizer.pad,
+                    special_vocab=special_vocab
+                )
+            else:
+                # Regular text embedding
+                self.text_embedding = nn.Embedding(num_tokens, cfg.embedding_dim)
```
blisc, line 170: Should we have different naming for these two embedding algorithms? Why not default both to self.text_embedding?

shehzeen: I think we should use a different name because cas_encoder has a different forward interface than text_embedding. Also _embedding sound like a simple nn.Embedding layer, which can be confusing for someone reading the code, since there is a transformer also being used in CharAwareSubwordEncoder.

blisc (approved):

PR #13595, opened 2025-05-15
Magpietts 2503 s multi lingual

shehzeen:
* Updates Multilingual yaml to match the English yaml hyperparams
* Minor updates to support IPA chars in text field - computes CER/WER against a different text key in the manifest if available 'original_text'
* Bug fix in Whisper transcribe function for multilingual GRPO
* GRPO updates - added DR. GRPO loss function also

### DIFF
--- a/examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
+++ b/examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
@@ -17,7 +17,7 @@ sample_rate: 22050
 
 model:
   model_type: "multi_encoder_context_tts" # single_encoder_sv_tts, multi_encoder_context_tts, decoder_context_tts or decoder_pretrain_synthesizer
-  use_text_conditioning_encoder: false # If true, distilbert will be used to encode context_text if provided.
+  use_text_conditioning_encoder: true # If true, distilbert will be used to encode context_text if provided.
   transcript_decoder_layers: [3,4,5,6,7] # ONLY used in multi_encoder_context_tts, Transcript goes to these layer ids, context goes to the rest. In single_encoder_sv_tts, all layers are used for transcript.
   context_decoder_layers: [8,9] # ONLY used in multi_encoder_context_tts
   context_duration_min: 3.0
@@ -28,13 +28,14 @@ model:
   prior_end_step: 12000
   prior_scaledown_start_step: 8000
   indefinite_prior_prob: 0.0 # If > 0, then prior will be applied after prior_end_step with this probability.
-  alignment_loss_scale: 0.0
+  alignment_loss_scale: 0.002
   embedding_dim: 768
   codecmodel_path: ???
   max_epochs: ${max_epochs}
   steps_per_epoch: ${weighted_sampling_steps_per_epoch}
 
   sample_rate: ${sample_rate}
+  cfg_unconditional_prob: 0.1
 
   # Alignment encoder parameters, to binarize the prior
   # This is used for attention-constrained training and inference
@@ -130,13 +131,14 @@ model:
       weighted_sampling_steps_per_epoch: ${weighted_sampling_steps_per_epoch}
       sample_rate: ${sample_rate}
       # speaker_path: ${speaker_path}
-      min_duration: 0.5
+      min_duration: 0.2
       max_duration: 20.0
 
     dataloader_params:
       batch_size: ${batch_size}
       num_workers: 4
       drop_last: true
+      pin_memory: true
 
   validation_ds:
     dataset:
@@ -144,12 +146,13 @@ model:
       dataset_meta: ${val_ds_meta}
       sample_rate: ${sample_rate}
       # speaker_path: ${speaker_path}
-      min_duration: 0.5
+      min_duration: 0.2
       max_duration: 20.0
 
     dataloader_params:
       batch_size: ${batch_size}
-      num_workers: 0
+      num_workers: 4
+      pin_memory: true
 
   encoder:
     n_layers: 6
@@ -160,7 +163,7 @@ model:
     p_dropout: 0.1
     p_dropout_out: 0.0
     has_xattn: false
-    is_causal: False
+    is_causal: true
     apply_norm_out: true
     max_length_causal_mask: 2048
     use_learnable_pos_emb: true
@@ -184,12 +187,12 @@ model:
     d_model: 768
     d_ffn: 3072
     sa_n_heads: 12
-    kernel_size: 3
+    kernel_size: 1
     p_dropout: 0.1
     p_dropout_out: 0.0
     has_xattn: true
     xa_d_memory: 768
-    xa_n_heads: 12
+    xa_n_heads: 1
     is_causal: true
     apply_norm_to_cond: true
     apply_norm_out: true
@@ -197,9 +200,8 @@ model:
     use_learnable_pos_emb: true
 
   optim:
-    _target_: torch.optim.Adam
-    lr: 2e-4
-    betas: [0.8, 0.99]
+    _target_: torch.optim.AdamW
+    lr: 1e-4
 
     sched:
       name: ExponentialLR
@@ -210,14 +212,14 @@ trainer:
   devices: -1
   accelerator: gpu
   strategy: ddp_find_unused_parameters_true
-  precision: 32
+  precision: bf16-mixed
   max_epochs: ${max_epochs}
   accumulate_grad_batches: 1
   enable_checkpointing: False # Provided by exp_manager
   logger: false # Provided by exp_manager
   log_every_n_steps: 100
-  val_check_interval: 500
-  # check_val_every_n_epoch: 10
+  check_val_every_n_epoch: 1
+  num_sanity_val_steps: 0
   benchmark: false
 
 exp_manager:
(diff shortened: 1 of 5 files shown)


nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -606,7 +606,12 @@ def __getitem__(self, index):
             align_prior = torch.tensor(align_prior, dtype=torch.float32)
             example["align_prior"] = align_prior
 
-        example['raw_text'] = data.text
+        if "original_text" in data.manifest_entry:
+            # For Manifests in which text field is IPA.
+            example['raw_text'] = data.manifest_entry['original_text']
+        else:
+            example['raw_text'] = data.text
```
blisc: We should refactor this since this is a non-standard change to our manifest structure and dataset class

nemo/collections/tts/models/magpietts_preference_optimization.py
```diff
@@ -435,6 +435,11 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             # import ipdb; ipdb.set_trace()
             assert HAVE_TORCHAUDIO, "torchaudio is required for PESQ reward"
             self.squim_objective_model = SQUIM_OBJECTIVE.get_model()
+        
+        self.loss_type = self.cfg.get('loss_type', 'grpo')
+        assert self.loss_type in ['grpo', 'dr_grpo'], f"Unknown loss type: {self.loss_type}"
```
blisc:
```suggestion
        if self.loss_type not in ['grpo', 'dr_grpo']:
            raise ValueError(f"Received loss_type of {self.loss_type}, but the model only accepts one of ['grpo', 'dr_grpo']")
```
Try to avoid assert statements, and raise Exceptions instead. If possible, mention how the user can fix this error as well.

scripts/magpietts/README_magpie_po.md
```diff
@@ -136,42 +136,65 @@ python scripts/magpietts/dpo/create_text_contextpairs.py \
```
blisc, line 136: In addition to a readme, can we add a simple test to make sure that the online policy path works on our released checkpoints? This would mean that we have to support a path to allow for older checkpoints to work on our current branch. To address issues from other teams.

examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
```diff
@@ -14,7 +14,7 @@ val_ds_meta: ???
 
 model:
   model_type: "multi_encoder_context_tts" # single_encoder_sv_tts, multi_encoder_context_tts, decoder_context_tts or decoder_pretrain_synthesizer
-  use_text_conditioning_encoder: false # If true, distilbert will be used to encode context_text if provided.
+  use_text_conditioning_encoder: true # If true, distilbert will be used to encode context_text if provided.
```
blisc, line 17: Is there a reason to change this parameter?

examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
```diff
@@ -123,24 +124,27 @@ model:
...
       num_workers: 4
       drop_last: true
+      pin_memory: true
 
   validation_ds:
     dataset:
       _target_: nemo.collections.tts.data.text_to_speech_dataset.MagpieTTSDataset
       dataset_meta: ${val_ds_meta}
-      min_duration: 0.5
+      sample_rate: ${sample_rate}
```
blisc: We removed sample_rate from our configs, can you change this here too?

blisc (changes requested):

nemo/collections/tts/data/text_to_speech_dataset.py
paarthneekhara: What is raw_text being used for?

nemo/collections/tts/models/magpietts.py
```diff
@@ -293,7 +293,10 @@ def load_state_dict(self, state_dict, strict=True):
         if strict == False:
             super().load_state_dict(state_dict, strict=False)
         for name, child in self.named_children():
-            if name in ['_speaker_verification_model', '_codec_model']:
+            if name in ['_speaker_verification_model', '_codec_model', '_reference_model',
```
paarthneekhara, line 313: Should we create a list or a set like self.ignore_modules_when_loading = [], and then update it in the individual model classes as and when needed?

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -71,9 +71,10 @@ def process_text(input_text):
     return single_space_text
 
 def transcribe_with_whisper(whisper_model, whisper_processor, audio_path, language, device):
+    print("Transcribing with Whisper...")
```
paarthneekhara: Remove print

examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
shehzeen: 'use_text_conditioning_encoder' is true in magpietts_en.yaml and the 'use_text_conditioning_encoder' is overridden in all baseline english sub files as true.

blisc: Should these parameters exist in the inference yaml? Shouldn't we inherit them from the base checkpoint instead?

scripts/magpietts/evaluate_generated_audio.py
blisc: Inside of scripts/, I care less about print(). Within the main nemo/ directory (aka the library portion of the repo, and what actually gets installed when you pip install), we require nemo.logging instead of print()

nemo/collections/tts/data/text_to_speech_dataset.py
shehzeen: @paarthneekhara  `raw_texts` and  is used in preference optimization for CER and WER computation in DPO data generation and GRPO. This is when `text` field contains the IPA string (used as input to the model), but we need the original text for computing CER and WER of the generated audio.

@blisc by refactoring do you mean I should add the original_text field in DatasetSample?

```
@dataclass
class DatasetSample:
    dataset_name: str
    manifest_entry: Dict[str, Any]
    audio_dir: Path
    feature_dir: Path
    text: str
    original_text: str = None # Used for CER/WER computation when text field contains phonemized/processed text
    speaker: str
    speaker_index: int = None
    tokenizer_names: List[str] = None
   ```

scripts/magpietts/README_magpie_po.md
shehzeen: I added a test. `tests/functional_tests/L2_TTS_Fast_dev_runs_Magpietts_OnlinePO.sh` 

It now works with the old multi-encoder checkpoint with overrides, but I had to add `update_ckpt` function in magpietts, to rename `t5_encoder` to `encoder` and `t5_decoder` to `decoder`, similar to what we do in infer_and_evaluate.py, so that it works correctly with the old checkpoint.

examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
shehzeen: I think, we can eventually get rid of the inference yamls, and just keep the training ones. We mainly need to remove train_ds and val_ds and add test_ds instead, which should be doable from the command line. We can make a separate PR for that. It will avoid unnecessary updates to all set of yamls and keep the defaults the same between training and inference.

examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
shehzeen: Done.

nemo/collections/tts/models/magpietts.py
shehzeen: I think that will make it more difficult to track which keys are being ignored and where the list is being updated, especially if we update the ignore list in individual classes and the load_state_dict is only in the base class.

nemo/collections/tts/data/text_to_speech_dataset.py
blisc: > @paarthneekhara raw_texts and is used in preference optimization for CER and WER computation in DPO data generation and GRPO. This is when text field contains the IPA string (used as input to the model), but we need the original text for computing CER and WER of the generated audio.

This should be explained in our code. Ideally, our dataset or model would have this docstringed somewhere. For this PR, we should add this explanation here.

scripts/magpietts/README_magpie_po.md
blisc: Can you add this test to the CI suite? You need to add this script to github/workflows/cicd-main.yml . See https://github.com/NVIDIA/NeMo/pull/13831 for an example.

nemo/collections/tts/data/text_to_speech_dataset.py
shehzeen: Added comments about this.

scripts/magpietts/README_magpie_po.md
shehzeen: I missed that earlier. Thanks, I have added it now. It must be mentioned somewhere, but is there a way to run CI/CD or the tests locally before pushing?

blisc: Can you also add a reference to your new test on line 1828 on .github/workflows/cicd-main.yml?
> but is there a way to run CI/CD or the tests locally before pushing?

Yes. However I haven't found an easy to do so yet, so not yet.

blisc (approved):

PR #13616, opened 2025-05-16
Compatibility between TensorBoard and W&B logging

rfejgin:
We currently don't have a good way for W&B and TensorBoard (TB) to work together because
- if we don't explicitly set `exp_manager.version` then Lightning ends up creating a separate TB directory for each run which results in different-colored plots per run. We don't usually want that, especially when running multiple 4-hour jobs which are really the same experiment.
- if we do set `exp_manager.version=0` it ends up confusing W&B which appears to end up using the provided version as the ID and then all experiments (even ones with different names) end up overwriting each other.

The solution suggested here is to add a config flag named `tensorboard_resume`. If set to true (I suggest to make that the default in the `ExpManagerConfig` class but we don't have to) then the **TB** version will get forced to `0`, resulting in all TB logs getting written to the same directory, but W&B not affected.

We could go fancier and try to detect the most recent TB version but that would mean we have to make assumptions on how Lightning names those, so not sure it's worth it. I gave an example of that under `if False` [here](https://github.com/NVIDIA/NeMo/compare/magpietts_2503...rfejgin:NeMo:magpietts_2503_tensorboard_wandb_compat?expand=1#diff-cacb4d3cf39593d980bcfcd9c7ed4b9a342b9565c5447511758daf1f02e0f60fR1220); but probably better to just leave the version at `0`. 

This is a draft pull request because it's only minimally tested, but I wanted to first run it by you.

@blisc @paarthneekhara: what do you think?

### DIFF
--- a/nemo/utils/exp_manager.py
+++ b/nemo/utils/exp_manager.py
@@ -219,6 +219,7 @@ class ExpManagerConfig:
     resume_from_checkpoint: Optional[str] = None
     # Logging parameters
     create_tensorboard_logger: Optional[bool] = True
+    tensorboard_resume: Optional[bool] = True
     summary_writer_kwargs: Optional[Dict[Any, Any]] = None
     create_wandb_logger: Optional[bool] = False
     wandb_logger_kwargs: Optional[Dict[Any, Any]] = None
@@ -661,6 +662,7 @@ def exp_manager(trainer: 'lightning.pytorch.Trainer', cfg: Optional[Union[DictCo
             cfg.version,
             cfg.checkpoint_callback_params,
             cfg.create_tensorboard_logger,
+            cfg.tensorboard_resume,
             cfg.summary_writer_kwargs,
             cfg.create_wandb_logger,
             cfg.wandb_logger_kwargs,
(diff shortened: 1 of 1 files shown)


rfejgin:
> I'm pretty sure if you set `explicit_log_dir` and leave exp_dir, name, and version alone, both should work as intended. Just tested locally and it seems to work

Thanks for testing that. @blisc What do you set `explicit_log_dir` to? the same value as what you set `exp_manager.exp_dir` to?

I was worried that `explicit_log_dir` would affect W&B but it doesn't appear to, from examining the code. We'd need to add this line to our configs but probably okay. I'll give it a try - thank you.

rfejgin:
I tested the `explicit_log_dir` solution and it seems to work for both TB and W&B simultaneously, and resuming works too (without multiple colors in TB). I now have these two lines in my training script:
`exp_manager.exp_dir="${EXPERIMENT}" \
++exp_manager.explicit_log_dir="${EXPERIMENT}" \
`

@paarthneekhara I think I can live with this added line since it's simple enough and doesn't need to change from experiment to experiment (when parameterized with $EXPERIMENT as above). Does it work for you? If so I'll close this PR.

rfejgin:
Discussed this with @paarthneekhara who also agrees that `++exp_manager.explicit_log_dir="${EXPERIMENT}"` is a good solution. Closing.

PR #13621, opened 2025-05-16
Enabled C2C-PCie bridge through NCCL

sanandaraj5597:
This PR enables environment variables for NCCL version 2.25 and 2.26. This feature becomes default after NCCL >= 2.27.

When performing NCCL Infini-Band communication over IB switches, we shouldn't use the CPU host memory involvement when moving data from GPU to PCIe. This increases the latency of NCCL communications and interferes with DRAM BW on CPU. Setting this environment variables avoid this.


scripts/performance/utils.py
```diff
@@ -76,6 +76,12 @@ def slurm_executor(
         "NEMO_LOG_MEMORY_USAGE": "1",  # Print memory allocation
         "NEMORUN_HOME": log_dir,
     }
+
+    # Assuming 4 GPU's per node as GB200, might need to revisit in future!
+    if num_gpus_per_node == 4:
+        env_vars["NCCL_NET_GDR_LEVEL"] = "PHB"  # For NCCL 2.25
```
guyueh1, line 84: Can we add a comment here to explain why we need these two variables (and can be removed when NCCL >= 2.27) for future reference

guyueh1 (approved):

scripts/performance/utils.py
```diff
@@ -78,17 +78,28 @@ def slurm_executor(
         "NEMORUN_HOME": log_dir,
         "PYTORCH_CUDA_ALLOC_CONF": "expandable_segments:True",
     }
+
+    # Assuming 4 GPU's per node as GB200, might need to revisit in future!
+    if num_gpus_per_node == 4:
+        env_vars["NCCL_NET_GDR_LEVEL"] = "PHB"  # For NCCL 2.25
+        env_vars["NCCL_NET_GDR_C2C"] = 1  # For NCCL 2.26
```
gdengk, line 85: This needs to be env_vars["NCCL_NET_GDR_C2C"] = "1"

sanandaraj5597: Oh, is this breaking your runs? I'll quickly patch it in that case.

PR #13727, opened 2025-05-25
[magpietts] decoder CE model type

paarthneekhara:


nemo/collections/tts/models/magpietts.py
```diff
@@ -212,7 +213,7 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
                 model_name='titanet_large'
             )
             self._speaker_verification_model.freeze()  #Lightning does requires_grad = False and self.eval()
-            self.speaker_projection_layer = nn.Linear(cfg.speaker_emb_dim, cfg.embedding_dim)
+            self.speaker_projection_layer = nn.Linear(cfg.speaker_emb_dim, cfg.decoder.xa_d_memory)
```
rfejgin: For the `decoder_ce` model type is this the right dimension to project to? Or to `d_model` ?

nemo/collections/tts/models/magpietts.py
```diff
@@ -232,6 +233,14 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             self.transcript_decoder_layers = [
                 idx for idx in range(self.decoder.n_layers)
             ]  # All layers are used for text
+        elif self.model_type == 'decoder_ce':
+            # Similar to decoder_context_tts, but we use context encoder
```
rfejgin, line 242: Suggest to add a text like this along with the other model type descriptions in the `MagpieTTSModel` class docstring (at the top of the class).

rfejgin (changes requested):
Looks good! As mentioned elsewhere, it's cool you were able to make an arch change with these few lines of code changed. It speaks to good structure of the code.

See some minor comments above.

nemo/collections/tts/models/magpietts.py
paarthneekhara: Reversed this change. Had added these because I was handling num x-attn heads incorrectly.

rfejgin:
@paarthneekhara would be nice to add a `decoder_ce` entry here
https://github.com/NVIDIA/NeMo/blob/6b4f22ad6f5c3d8ffe379bca866446dba9055f5b/nemo/collections/tts/models/magpietts.py#L79

... but I heard you're planning a rework of model types anyway, so up to you.

rfejgin (approved):

PR #13519, opened 2025-05-09
Enable changing quant_cfg via command line in the ptq.py script

janekl:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

**Collection**: [Note which collection this PR will affect]

# Changelog

- Add specific line by line info of high level changes in this PR.

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information

- Related to # (issue)

PR #13756, opened 2025-05-28
Use modelopt export and disable dataset calibration for weight only PTQ

jenchen13:
> [!IMPORTANT]  

# What does this PR do ?

Use modelopt to export to HuggingFace, and disable calibration for weight-only PTQ

**Collection**: llm

# Changelog

- modify `llm.export_ckpt` API to use modelopt export for HF
- for weight only PTQ, disable dataset calibration

# Testing

- tested BF16 and NVFP4 `export_ckpt`
- tested int4 PTQ (weight-only) with `scripts/llm/ptq.py`

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/lightning/pytorch/callbacks/model_checkpoint.py
```diff
@@ -678,6 +678,8 @@ def _remove_checkpoint(self, trainer: "lightning.pytorch.Trainer", filepath: str
         """
         if self.async_save and not override_async:
             # Register checkpoint removal in the last (active) checkpoint removal list
+            if len(self.deferred_ckpts_to_remove) == 0:
+                self.deferred_ckpts_to_remove.append([])
```
jenchen13, line 682: sometimes I run into a bug where it tries to remove a ckpt but crashes because `self.deferred_ckpts_to_remove` is an empty list

nemo/collections/llm/api.py
```diff
@@ -988,14 +992,19 @@ def export_ckpt(
...
     if output_path and not isinstance(output_path, Path):
         output_path = Path(output_path)
         if output_path.exists() and not overwrite:
             raise FileExistsError(f"Output path {output_path} exists. Use overwrite=True to force overwrite.")
 
-    output = io.export_ckpt(path, target, output_path, overwrite, load_connector, **kwargs)
+    if target == "hf":
+        output = export_hf_checkpoint(path, output_path, **modelopt_export_kwargs)
+    else:
+        output = io.export_ckpt(path, target, output_path, overwrite, load_connector, **kwargs)
```
Laplasjan107, line 1007: Shouldn't the `if` clause check if the checkpoint is quantized via modelopt?

Now, non-quantized NeMo checkpoints would also end up being exported to `"hf"` via modelopt, wouldn't they?

nemo/lightning/io/api.py
```diff
@@ -247,7 +249,16 @@
         nemo_ckpt_path = Path("/path/to/model.ckpt")
         export_path = export_ckpt(nemo_ckpt_path, "hf")
     """
-    exporter: ModelConnector = load_connector(path, target)
+    from nemo.collections.llm.modelopt.quantization.quantizer import export_hf_checkpoint
```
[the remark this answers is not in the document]
jenchen13: i tested export and this does not raise an error

tests/collections/llm/modelopt/test_quantizer.py
```diff
@@ -0,0 +1,49 @@
...
+        quant_cfg = QuantizationConfig(algorithm="int8_sq")
+        assert not quant_cfg.is_weight_only()
+
+    def test_is_weight_only_block_fp8(self):
+        quant_cfg = QuantizationConfig(algorithm="block_fp8")
+        assert quant_cfg.is_weight_only()
+
+    def test_is_weight_only_int4_awq(self):
+        quant_cfg = QuantizationConfig(algorithm="int4_awq")
+        assert quant_cfg.is_weight_only()
```
janekl: I'm not sure if this is actually correct. INT4 AWQ means activation-aware, so needs some sample data to calibrate. Could you double check this? We might need a more refined `is_weight_only()` detection method.

jenchen13: In the modelopt [config](https://gitlab-master.nvidia.com/omniml/modelopt/-/blob/main/modelopt/torch/quantization/config.py?ref_type=heads#L240) it says `input_quantizer` is disabled. I'll double check with the team

jenchen13: I checked with Asma and this is correct because INT4 AWQ is is INT4 weights + FP16 activation (no activation quantization)

janekl: In that sense yes, it is a weight-only method. But we are using `is_weight_only()` to disable forward pass with calibration data here. However, it seems that INT4 AWQ needs some data to perform activation-based scaling of the weights, so we should still pass it though some data in my view, see also https://arxiv.org/abs/2306.00978:

Zrzut ekranu 2025-06-03 092009

jenchen13: thanks for the explanation! should I modify `is_weight_only()` to check if "awq" is not in the algorithm name? there's no good way to tell that it's not weight only quantization from the config itself

nemo/collections/llm/modelopt/quantization/quantizer.py
```diff
@@ -459,6 +453,41 @@ def export(self, model, model_dir: str, trainer: Optional["Trainer"] = None) ->
             logging.info(f"Export succeeded, model has been exported to {export_dir}.")
 
 
+def export_hf_checkpoint(
```
AAnoosheh, line 456: Should we move this to another file?

nemo/lightning/pytorch/callbacks/model_checkpoint.py
AAnoosheh: What is your `save_top_k` value?

nemo/collections/llm/modelopt/quantization/quantizer.py
jenchen13: I tried moving it to `modelopt/model_utils.py` and then it created a circular import so I kept it here for now

jenchen13:
@AAnoosheh my `save_top_k` value is 1. should I change it to 2 to avoid the error?

tests/collections/llm/modelopt/test_quantizer.py
jenchen13: i have fixed the function & the test case, please take a look!

nemo/collections/llm/api.py
```diff
@@ -556,6 +556,8 @@ def ptq(
             trainer_kwargs={},
             model_config_overrides={"sequence_parallel": False},
         )
+    if forward_loop is None and not quantization_config.is_weight_only():
```
janekl: Would you be fine moving it to the original place i.e. `quantize(...)` method? So that possibly "direct" users of `Quantizer` can also benefit from `is_weight_only` check (now it's only available in api.py)

nemo/collections/llm/api.py
```diff
@@ -937,6 +939,7 @@ def export_ckpt(
     output_path: Optional[AnyPath] = None,
     overwrite: bool = False,
     load_connector: Callable[[Path, str], io.ModelConnector] = load_connector_from_trainer_ckpt,
+    modelopt_export_kwargs: dict[str, Any] = {},
```
janekl: Let's default to `None` and set `kwargs` to `{}` by hand. I don't quite expect this is really risky here but anyway it's a good practice to avoid mutable defaults see, for example, https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

nemo/collections/llm/api.py
```diff
@@ -995,7 +999,7 @@ def export_ckpt(
         if output_path.exists() and not overwrite:
             raise FileExistsError(f"Output path {output_path} exists. Use overwrite=True to force overwrite.")
 
-    output = io.export_ckpt(path, target, output_path, overwrite, load_connector, **kwargs)
+    output = io.export_ckpt(path, target, output_path, overwrite, load_connector, modelopt_export_kwargs, **kwargs)
```
janekl, line 1028: FYI @cuichenx @marcromeyn @suiyoubi a change to `export_ckpt` is being pursued here (a useful one)

janekl:
Could we please add a functional test for `export_ckpt` for a modelopt-modified checkpoint as input?

nemo/collections/llm/modelopt/quantization/quantizer.py
```diff
@@ -79,6 +84,14 @@ class QuantizationConfig:
     calibration_batch_size: int = 64
     calibration_seq_len: int = 128
 
+    def is_weight_only(self):
+        """Check if the quantization is weight only."""
+        quant_cfg = QUANT_CFG_CHOICES[self.algorithm]['quant_cfg']
```
janekl: I'm afraid this will fail when using a custom config loaded from disk that I added in https://github.com/NVIDIA/NeMo/pull/13519. There is a test for such a use-case in our internal CI system. The question is whether we really need such a feature, but currently it's there and we use it for a DeepSeek V2 Lite test to reduce quantization block sizes for a config loaded from JSON.

This is problematic as we can't use the standard name of the config to check it's weight-only like we do here with "awq".

Could you please add two extra lines when entering `is_weight_only` to check `if self.algorithm not in QUANT_CFG_CHOICES` then we just assume and return `False`?

janekl:
Left a final comment, LGTM besides, thanks. There is also a git conflict to resolve, and then we can merge.

nemo/collections/llm/api.py
janekl: Also FYI @oyilmaz-nvidia

nemo/collections/llm/modelopt/quantization/quantizer.py
jenchen13: makes sense, added above!

janekl (approved):

AAnoosheh (approved):

AAnoosheh (approved):

janekl (approved):

nemo/collections/llm/modelopt/quantization/quantizer.py
```diff
@@ -288,7 +293,8 @@ def quantize(self, model: "MegatronParallel", forward_loop=None):
 
         If forward_loop is not provided, a forward loop will be created using the calibration dataset.
         """
-        if forward_loop is None:
+        modelopt_cfg = QUANT_CFG_CHOICES[self.quantization_config.algorithm]
+        if forward_loop is None and mtq.config.need_calibration(modelopt_cfg):
```
janekl, line 297: That's by far the best solution. Now details remain:
1. Using `algorithm = None` a.k.a. `"no_quant"`
2. Loading a config from a JSON file.

(Both cases are tested in the internal CI system, but not here on Github.)

To address these corner cases it's enough to move the check after L310 as:
```
        if forward_loop is None and mtq.config.need_calibration(quant_cfg):
            forward_loop = self._get_forward_loop(model)
```
see also git diff attached with the propsal:

Zrzut ekranu 2025-06-13 103701

janekl: But since the Github CI is currently halfway through, let's change that in a subsequent MR, no need to edit this one anymore. And again thanks for coming with the `mtq.config.need_calibration` method here.

PR #13548, opened 2025-05-12
Magpietts 2503 updateshars

paarthneekhara:
Should run in on multiple GPUs/nodes

I want to know if the work flow looks correct.

I tested training locally on shars generated by this script. Seems to be working ok.

PR #12617, opened 2025-03-14
SpeechLM2 collection

pzelasko:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

Add a one line overview of what this PR aims to accomplish.

**Collection**: [Note which collection this PR will affect]

# Changelog

- Add specific line by line info of high level changes in this PR.

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information

- Related to # (issue)

PR #13821, opened 2025-06-04
[magpietts][lhotse_v2] Adding scripts of converting NeMo manifests to Lhotse Shars, and speedup improvements for codec model inference.  

XuesongYang:
### Speed Up Codec Model Inference
Made slightly changes for the dependency util func `mask_sequence_tensor`. We observed at least 3x speedup. This change is cherry-picked from the part of PR https://github.com/NVIDIA/NeMo/pull/12617

### Lhotse Shar v2
Three scripts were added to preprocess NeMo manifests and convert them into Lhotse Shars (cuts/audio/codes). This new version reduced significantly computation overheads by using rank balanced workloading and writing independently. This process involves four main steps.

1.  **Extend Manifests with Context Audio (on GPU nodes):** Enhance the NeMo manifests by adding context audio information. 
    * The old recipe saved individual speaker embedding files and load again to compute speaker similarity. It is not computation scaling-up friendly. 
    * This new recipe runs speaker embedding extraction on the fly and applies `torch.matmul` to compute similarity matrix. It recursively find the next best context audio if the 1-best candidate does not survive, resulting in preserving more data records.
    * Scaling up computation friendly without access IPC cost. It pre-allocates a distinct subset of speaker records to each GPU rank in consideration of rank workload balance by using greedy bin-packing strategy. Tried round-robin but not ideal for this task.  
    * Buffered writing manifest only when the buffer is filled. 

2.  **Create Lhotse Shards (on CPU nodes):** Convert the extended NeMo manifests into Lhotse shards.
    * Processes a chunk of manifest entries, loads audio, and writes corresponding single shard files for cuts, target audio, and context audio.
    * Designed to be run in a parallel worker process. 
    * Loads and writes audio iteratively to save memory.

3.  **Extend Shards with Audio Codes (on GPU nodes):** Process the Lhotse shards to extract and include audio codes (audio codec extraction).
    * pre-allocate Lhotse shards to each rank, and each rank process and write independently.
    * Pre-allocate padded audio tensors instead of run for..loop of torch.func.pad
    * avoided running zero pads 4x times that was observed in old recipe.
    
**Note:** (**for internal users**) each python script is wrapped as a slurm job sub file that can process multiple datasets at once. Refer [this link](https://gitlab-master.nvidia.com/xueyang/nemo-tts-artifacts-registry/-/tree/model_release_2505/model_release_2505/data_prep?ref_type=heads) for details.


XuesongYang:
part of the code has been reviewed by @pzelasko offline. Piotr, could you pls review the scripts again if you have time?

blisc:
Can we close #13548?

paarthneekhara:
> Can we close #13548?

Yes, we can close that. It is no longer needed.

chtruong814:
@XuesongYang please merge main into this branch again to resolve some issues with getting the CI to run

subhankar-ghosh (commented):
LGTM. Descriptive and clear comments.

paarthneekhara:
Looks good to me. We can add the above PR description to a README.

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
```diff
@@ -0,0 +1,873 @@
...
+            f"Flushing {len(speakers_to_process_now)} completed speakers. "
+            f"Current total accumulated items: {self.total_accumulated_items}"
+        )
+
+        for speaker_id in speakers_to_process_now:
+            speaker_items = self.speaker_data_accumulator.pop(speaker_id)
+            self.total_accumulated_items -= len(speaker_items)
+            self.processed_speakers_set.add(speaker_id)
+
+            # NOTE: Potential OOM (Out Of Memory) risk if a single speaker has an extremely large
```
rfejgin, line 298: @XuesongYang One potential way around this is to sub-sample the utterances. I recall @shehzeen mentioning that's how they looked for similar context in the past. By that I mean, let's say the speaker has 1000 utterances. It would be nice to find the most similar of those 1000 utterance to our target utterance. But finding the one that is most similar among a random subset of 100 utterances would probably work almost as well.

rfejgin:
This is a heavy lift, nice work, @XuesongYang !

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
```diff
@@ -0,0 +1,873 @@
...
+            # number of segments (e.g., tens of thousands). The N x N similarity matrix calculated below
+            # (where N = len(speaker_items)) can consume significant CPU RAM.
```
rfejgin, line 300: nitpick but should this say **G**PU?

scripts/magpietts/extend_lhotse_shards_with_audio_codes.py
```diff
@@ -0,0 +1,912 @@
...
+        context_audios: torch.Tensor,
+        context_audio_lens: torch.Tensor,
+    ) -> Optional[Dict[str, torch.Tensor]]:
+        try:
+            target_audios = target_audios.to(self.device)
+            target_audio_lens = target_audio_lens.to(self.device)
+            context_audios = context_audios.to(self.device)
+            context_audio_lens = context_audio_lens.to(self.device)
+            # NOTE: we avoided directly calling `self.codec_model.encode()` because it pads audios again.
+            with torch.no_grad():
```
rfejgin: Consider using `with torch.inference_mode():` it disables more than than `torch.no_grad()` and is recommended when no training is involved. But the performance benefit is probably small, so not very important.

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
XuesongYang: nope. it is about cpu mem control. the embeddings was detached and converted to cpu, so the similarity matrix is computed using cpu. It can bear with much larger mem than gpu. It is easy to shift to GPU computation if we discard `.cpu()` for the embeddings, but current setup did not hurdle the gpu utils from my observations. We can iterate in the future if we want to process much larger scale dataset.
```
embedding_cpu_fp32 = speaker_embeddings_gpu[i].cpu().type(torch.float32)
```

XuesongYang:
> @XuesongYang please merge main into this branch again to resolve some issues with getting the CI to run

sorry not catch what you suggested. Our code branch `magpietts_2503` is far behind the latest main branch. Rebasing to the latest main will be risky for our experiments for now.

XuesongYang:
> Looks good to me. We can add the above PR description to a README.

yeah, will do.

scripts/magpietts/extend_lhotse_shards_with_audio_codes.py
XuesongYang: that's good recommendation. We should replace it with `torch.inference_mode()`.

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
XuesongYang: agreed. a random subset of 100 makes sense if dealing large utterances for a speaker. So far it doesn't have problems for existing datasets. Let's iterate in the future once we got larger datasets.

XuesongYang:
> Looks good to me. We can add the above PR description to a README.

@paarthneekhara I just added the README file. Please have a look.

PR #13848, opened 2025-06-06
[magpietts][wandb] moved `val_loss` to `val` panel from `Chart` in wandb UI.

XuesongYang:
fixed unexpected panel displaying the `val_loss` in wandb UI. Previously `val_loss` was displayed in the pannel 'Charts'. Now it is displayed in the panel 'val', the same group as other val metrics. Also fixed typos.

**Previously**
Screenshot 2025-06-05 at 5 38 01 PM

**After fix**
Screenshot 2025-06-05 at 5 38 29 PM


blisc (changes requested):
Need to root cause checkpoint issues

blisc (commented):
Unblock, to review

blisc (approved):

XuesongYang:
> Need to root cause checkpoint issues

confirmed that this PR doesn't block the resume of training.

PR #13849, opened 2025-06-06
[magpietts] added logging of consumed samples at each training step.

XuesongYang:
The Lhotse dataloader applies dynamic batching based on duration buckets, causing the number of consumed samples at each training step to vary. Tracking the number of consumed samples enables fairer model comparisons than relying solely on training step counts. The Megatron models also tracked consumed samples.

This PR adds sample consumption logs to the wandb UI and appends this information to checkpoint filenames.

**ckpt filename changes**
`MagpieTTS-EN-Lhotse--val_loss=11.7273-step=1350-epoch=26.ckpt -->`
`MagpieTTS-EN-Lhotse--val_loss=11.7273-step=1350-epoch=26-consumed_samples=96466.ckpt`

**wandb UI changes**
Screenshot 2025-06-05 at 11 52 56 PM

### DIFF
--- a/examples/tts/conf/magpietts/magpietts_dc_en.yaml
+++ b/examples/tts/conf/magpietts/magpietts_dc_en.yaml
@@ -165,5 +165,6 @@ exp_manager:
     save_top_k: 5
     save_best_model: true
     always_save_nemo: true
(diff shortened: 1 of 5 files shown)


XuesongYang:
Lhotse applied dynamic batching for a duration bucket across ranks so that each rank will proceed with varying batch sizes. This PR would then makes training progress hanging right after the validation is complete. Need to find alternative ways of reducing with sum batch sizes across ranks.

convert to draft for further investigation.

PR #13856, opened 2025-06-06
[tranformer_core][magpietts_config] added support to override xattn head dim by xa_d_head

XuesongYang:
## Changes
1. add `xa_d_head` overrides to support customized cross attention head dimension;
2. removed `d_memory` check in `CrossAttention` class since it's already validated by the `Transformer` class before `CrossAttention`.

### DIFF
--- a/.github/workflows/_build_container.yml
+++ b/.github/workflows/_build_container.yml
@@ -1,0 +1,89 @@
+name: ~Build container template
+on:
+  workflow_call:
+    inputs:
+      image-name:
+        required: true
+        type: string
+        description: "The name of the image to build"
+      dockerfile:
+        required: true
+        type: string
+      runner:
+        required: false
+        default: self-hosted-azure-builder
+        type: string
+        description: "The runner to use for the build"
+
+jobs:
+  pre-flight:
+    runs-on: ubuntu-latest
+    outputs:
+      build_args: ${{ steps.manifest.outputs.BUILD_ARGS }}
+      cache-from: ${{ steps.cache_from.outputs.LAST_PRS }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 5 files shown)


nemo/collections/tts/modules/transformer_2501.py
```diff
@@ -362,15 +363,15 @@ def __init__(
...
         """
         super().__init__(
             n_heads=n_heads,
             d_model=d_model,
             p_dropout=p_dropout,
             is_causal=False,
+            d_head=d_head,
         )
-        if d_memory is None:
-            raise ValueError("d_memory must be provided for cross-attention")
```
blisc, line 373: Why remove this check?

examples/tts/conf/magpietts/magpietts_dc_en.yaml
```diff
@@ -119,6 +119,7 @@ model:
     has_xattn: true
     xa_d_memory: 768
     xa_n_heads: 1
+    xa_d_head: null  # use caution when specifying `xa_d_head` as it will override the default value of d_model // xa_n_heads in Attention class.
```
blisc: Should we add this to the yaml config? I would let to add less params to the yaml

nemo/collections/tts/modules/transformer_2501.py
XuesongYang: removed d_memory check in CrossAttention class since it's already validated by the Transformer class before CrossAttention.

examples/tts/conf/magpietts/magpietts_dc_en.yaml
XuesongYang: i don't have any preference about this. It is not common to override `xa_d_head` though. Let's remove them in the yaml.

PR #13861, opened 2025-06-06
bug fix in checkpoint loading

paarthneekhara:


XuesongYang:
could you pls elaborate why to make such changes? i played training and resume training without such changes, but did not have any prominent negative observations. Thanks.

PR #13865, opened 2025-06-08
[WIP] STT with frozen streaming ASR

ankitapasad:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/speechlm2/data/duplex_t2t_dataset.py
```diff
@@ -0,0 +1,303 @@
...
+            # Add random pad tokens within text_ids
+            # Calculate maximum number of pads we can add without exceeding turn_length
+            max_pads_possible = turn_length - len(text_ids)
+            if max_pads_possible <= 0:
+                # If text_ids already fills or exceeds turn_length, truncate and don't add pads
+                expanded_text_ids = text_ids[:turn_length]
+            else:
+                # Determine how many pad tokens to add (10-30% of text length, but capped by available space)
+                desired_pads = random.randint(len(text_ids) // 10, len(text_ids) // 3)
+                num_pads_to_add = min(desired_pads, max_pads_possible)
```
zhehuaichen: do you think this 10-30% is realistic? why can't we make num_pads_to_add = max_pads_possible?  
by looking at some examples , in each turn, i feel our source tokens has a lot of trailing paddings in between source tokens finish and target tokens start

ankitapasad: Hi @zhehuaichen 

1. The 10-30% is for within the source token sequences, the remainder of frames are all populated with trailing pads. 
2. I've dropped this setup as randomly interleaving is not the correct way to design. (we spoke about this but happy to chat more offline) The new setup uses word level timestamps to interleave pad -- i.e. add pad tokens for any non-token frames between start of two words. https://github.com/ankitapasad/NeMo/blob/s2s_freeze_asr/nemo/collections/speechlm2/data/duplex_t2t_dataset.py#L287

zhehuaichen: that's right but it relies on alignment data where i am worried not all of our data has as of now . Or if some data does not have, we can fall back to this setup

r.e. 1, is there any reason to have railing pads in your design? my original thinking was there should be minimum trailing pad/blank in a RNNT decoding hypothsis. if that's the case, we should be consistent with that

ankitapasad: The trailing pads are to fill the gap when agent is speaking between two consecutive user turns. Here's an example frame-level output from CTC, the interleave pad (10-30%) was to mimic the `1024`s within the user token sequence and the trailing pad tokens were to match the long sequences of `1024`s that correspond to the silence between two user turns when the agent is speaking.

```
[1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024,   29, 1024,   20, 1024, 1024,  175,
   8,    8,    1, 1024,   15, 1024,  176, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,   77, 1024,   16,
1024, 1024,  852, 1024, 1024, 1024,    3,  314,   17, 1024,   17,   17,
  21,  222, 1024,   20, 1024,   22, 1024,   80, 1024, 1024, 1024, 1024,
   6, 1024,    3, 1024,  753, 1024, 1024,    9, 1024, 1024,  608, 1024,
   1, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024,   24, 1024, 1024,    2, 1024,  634,   78,
  78,   45,    9, 1024,    2, 1024, 1024,  606, 1024, 1024,  282, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024, 1024,
1024]
```

zhehuaichen: in my understanding,
for supervision in cut.supervisions: max_pads_possible is corresponding to the duration of this supervision. so if this is an user turn, i think the pad should be all over the turn, instead of 10-30% in the front with lots of trailings towards the end 
and then for agent turn, the source_tokens should be as you said, all pads

ankitapasad: Okay yes I get what you mean. Thanks for explaining! 

Yes, spreading it across the whole duration is a way to "mimick" time-aligned but it will still be just an approximation (possibly not a close one). Since we will be adding random distance between tokens (or words) the tokens won't be exactly time aligned to the corresponding spoken parts. And thus, just like the previous interleaving approach, with this randomness, there are not concrete patterns for the model to learn from except the monotonicity.

zhehuaichen: agree. let's fix it later when you use it as a fallback if no alignment data in some cases

examples/speechlm2/s2s_duplex_frozen_asr_train.py
```diff
@@ -38,13 +42,55 @@ def train(cfg):
 
     with trainer.init_module():
         model = DuplexT2TModel(OmegaConf.to_container(cfg.model, resolve=True))
+    
+    if cfg.model.train_retokenizer:
+        # Load FP32 streaming ASR model
+        asr_model = load_pretrained_nemo(ASRModel, cfg.model.pretrained_asr).eval()
```
zhehuaichen, line 49: do you freeze the asr encoder and decoder? where did you do that in the code?

nemo/collections/speechlm2/models/duplex_t2t_model.py
```diff
@@ -337,11 +404,44 @@ def prepare_t2t_inputs(self, batch: dict):
...
+                padding_value=self.text_pad_id
+                )
+            source_token_lens = torch.tensor([len(hyp.y_sequence) for hyp in asr_hyps])
+        else:
+            source_tokens = batch["source_tokens"]  # (B, T)
         target_tokens = batch["target_tokens"]  # (B, T)
-        
+
+        # Append padding to the shorter sequence
+        diff = target_tokens.shape[1] - source_tokens.shape[1]
```
zhehuaichen, line 447: can you remind me why there will be difference between source and target?

nemo/collections/speechlm2/models/duplex_t2t_model.py
```diff
@@ -822,4 +943,31 @@ def tokens_to_turn_str(tokens: torch.Tensor, lengths: torch.Tensor, tokenizer: A
...
+class Retokenizer(nn.Module):
+    def __init__(self, input_dim, output_dim, use_transformer=False, num_layers=1, num_heads=4, dropout=0.1):
+        super().__init__()
+        self.use_transformer = use_transformer
+
+        # First projection to match transformer input size
+        self.linear = nn.Linear(input_dim, output_dim)
+
+        if use_transformer:
+            encoder_layer = nn.TransformerEncoderLayer(
```
zhehuaichen: is this a causal transformer? iiuc we need a causal one right?

nemo/collections/speechlm2/models/duplex_t2t_model.py
```diff
@@ -822,4 +943,31 @@ def tokens_to_turn_str(tokens: torch.Tensor, lengths: torch.Tensor, tokenizer: A
...
+                d_model=output_dim,
+                nhead=num_heads,
+                dim_feedforward=output_dim * 4,
```
zhehuaichen: can you also remind me where did you implement the stt inference code for this learnt retokenization method?

examples/speechlm2/s2s_duplex_frozen_asr_train.py
ankitapasad: No I don't explicitly freeze the asr model. It's loaded in eval model and only used as self._asr.transcribe which shouldn't allow passing of any gradients. Do you think we need to freeze it explicitly?

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: I notice that there can be difference of 1 frame, possibly due to some rounding effect for the edge frames.

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: good point! Yes we do. When I tried a transformer retokenizer the model did not converge, maybe that explains it.

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: I don't have one. I use the train script with 

```
trainer.val_check_interval=1 \
trainer.limit_val_batches=null
```

zhehuaichen: no i mean where is the code on how do the model do greedy decoding? i am curious about how do you do step by step learnable retokenization

nemo/collections/speechlm2/models/duplex_t2t_model.py
zhehuaichen: remind me where is the code for your latest converged and best architecture for learnable retokenization

nemo/collections/speechlm2/models/duplex_t2t_model.py
zhehuaichen: can we change this to an assert of something like diff is less than 2

examples/speechlm2/s2s_duplex_frozen_asr_train.py
zhehuaichen: i am not 100% sure. how about we protect the transcribe call with something like with torch.no_grad()

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: good point, will do

examples/speechlm2/s2s_duplex_frozen_asr_train.py
ankitapasad: Yeah it already is -- https://github.com/ankitapasad/NeMo/blob/s2s_freeze_asr/nemo/collections/speechlm2/models/duplex_t2t_model.py#L309

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: are you looking for this part?

https://github.com/ankitapasad/NeMo/blob/s2s_freeze_asr/nemo/collections/speechlm2/models/duplex_t2t_model.py#L429-L437

nemo/collections/speechlm2/models/duplex_t2t_model.py
ankitapasad: Reference: https://github.com/ankitapasad/NeMo/blob/s2s_freeze_asr/examples/speechlm2/s2s_duplex_frozen_asr_train.py#L72-L102

So far, the following seem to work well (final evaluation pending): 
1. `linear_retokenizer`
```
retokenize_type="no_llm_embed"
num_transformer_layers=0
retokenize_first=True # retokenize user embed before combining with agent embeddings
``` 

2. `soft_token_map`
```
retokenize_type="soft_token_map"
retokenize_first=True # retokenize user embed before combining with agent embeddings
```

PR #13879, opened 2025-06-10
[magpietts][lhotse_v2] make model training recipe adapt to the latest v2 datasets.

XuesongYang:
- change the codes names in cuts into target_codes and context_codes.
- remove `codec_model_name` in yaml config because we've added the codec name in `input_cfg.yaml` file.
- fixed bugs that `spec_len` has extended with bos and eos resulting in unexpected 2 more tokens. Prioritize `normalized_text` over `text` if available.
- added cuts sampling filters in terms of max CER and min context speaker similarity.
- add `sample_rate` overrides the same as codec model's. Without overrides, the audio will be resampled into 16k by default.
- bugfix: ensure start and duration params equal to `None` in `TemporalArray.load(start, duration)` since the audio codes were extracted based on segmented audio.
- downlevel to `logging.warning` since they were verbosed for every step.
- explicitly call `wandb.finish()` after `trainer.fit` or `trainer.test` complete to avoid hangs when debugging with num_worker=0.
- extract numerical value from tensors.

### DIFF
--- a/examples/tts/conf/magpietts/magpietts_lhotse_dc_en.yaml
+++ b/examples/tts/conf/magpietts/magpietts_lhotse_dc_en.yaml
@@ -8,7 +8,6 @@ model:
   use_text_conditioning_encoder: false # If true, distilbert will be used to encode context_text if provided.
   context_duration_min: 5.0
   context_duration_max: 5.0
-  codec_model_name: "21fpsCausalDecoder"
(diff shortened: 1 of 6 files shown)


rfejgin:
Looks good to me.

rfejgin (approved):
Looks good to me.

PR #13893, opened 2025-06-11
Fix for Squad Dataset Download

rhmukundan:
#### Current Issue

When running with multiple GPUs per node and/or multiple nodes, downloading the SQuAD dataset can fail due to a race condition - multiple processes attempt to download the dataset simultaneously. This makes the dataset download to fail. 

#### Fix

This fix ensures that the SQuAD dataset is downloaded in a **separate SLURM job** using only **1 node** and **1 GPU per node**. This prevents concurrent downloads and eliminates the race condition.
```
with run.Experiment(exp_name) as exp:
        if not SKIP_IMPORT:
            assert args.hf_token is not None, "HF token is required for importing checkpoint from HuggingFace"
            exp.add(*import_ckpt_experiment(executor, model(), source=f"hf://{HF_MODEL_URI}"))
            # Prepare SQuAD dataset before finetuning
            exp.add(*prepare_squad_dataset_experiment(executor, HF_MODEL_URI, seq_length=4096))
```


guyueh1:
I have a few questions
* where are these added functions used?
* Is the purpose to avoid downloading squad from the network and use cached files? 
* if yes, I think existing squad dataset should be able to use cached files if you properly set HF_HOME or HF_HUB envvar, but I haven't tried it recently, i can confirm later.

rhmukundan:
@guyueh1 Thanks for your comment. 
If you noticed, it is using the same function in nemo to download the dataset. Currently it is part of the overall job which means (num_nodes > 1 and num_gpus_per_node > 1) which causes the race condition which prevents the download and throws an error. 
In order to fix this, I have created `prepare_squad_dataset` and `prepare_squad_dataset_experiment` functions which create a separate slurm job (1 Node and 1 GPU Per node) to avoid the race condition and download the dataset successfully. The functions used are pretty much what is part of the current nemo implementation (https://github.com/NVIDIA/NeMo/blob/577b7fb77f6767e23dec2926ba9b9d2a3722713b/scripts/performance/utils.py#L523) to get the squad dataset. What has changed is the way how it is downloaded (1 node and 1 GPU-per-node) via a separate slurm job similar to the checkpoint download. 

This implementation will be used in finetune recipes like the following:
```
with run.Experiment(exp_name) as exp:
        if not SKIP_IMPORT:
            assert args.hf_token is not None, "HF token is required for importing checkpoint from HuggingFace"
            exp.add(*import_ckpt_experiment(executor, model(), source=f"hf://{HF_MODEL_URI}"))
            # Prepare SQuAD dataset before finetuning
            exp.add(*prepare_squad_dataset_experiment(executor, HF_MODEL_URI, seq_length=4096))
 ```

**Comment by guyueh1** [comment]
@rhmukundan I see. Can you provide an example of how to use it in run script?

**Comment by rhmukundan** [comment]
@guyueh1: I have pushed the Finetuning LLAMA4 Maverick recipe (that has the fix included) and also added the fix to LLAMA3 Finetuning 70b file. 
cc: @ashbhandare

File: scripts/performance/llm/finetune_llama3_70b.py [hunk 1]
```diff
@@ -230,6 +231,7 @@ def override_recipe_configs(
         if not SKIP_IMPORT:
             assert args.hf_token is not None, "HF token is required for importing checkpoint from HuggingFace"
             exp.add(*import_ckpt_experiment(executor, model(), source=f"hf://{HF_MODEL_URI}"))
+            exp.add(*prepare_squad_dataset_experiment(executor, HF_MODEL_URI, seq_length=4096))
```
guyueh1: it's best to add a command-line argument, for instance `standalone_dataset_preparation`, for users to indicate they want this or the default way; that argument should default to false

guyueh1:
@malay-nagda can you review the added utility functions in `util.py`?

scripts/performance/utils.py
```diff
@@ -495,6 +495,70 @@ def import_ckpt_experiment(executor: run.SlurmExecutor, model: run.Config[GPTMod
...
+        model_name (str): The name of the model
+        seq_length (int): The sequence length to use for packing. Defaults to 2048.
+    """
+    from pathlib import Path
+
+    from nemo.collections.common.tokenizers.huggingface.auto_tokenizer import AutoTokenizer
+    from nemo.collections.llm.gpt.data.packed_sequence import PackedSequenceSpecs
+    from nemo.collections.llm.gpt.data.squad import SquadDataModule
+
+    hf_home = Path(get_hf_home())
```
malay-nagda: Can we use `NEMO_HOME`? 

Reason being 
- user can set from CLI arg and we make sure NEMO_HOME is mounted inside container
- we don't have to maintain another extra path/env_var with HF_HOME. 
- I think main exp still checks for dataset here and tries downloading here if its missing (and we face the same issue this PR is trying to solve, right?)

scripts/performance/utils.py
```diff
@@ -495,6 +495,70 @@ def import_ckpt_experiment(executor: run.SlurmExecutor, model: run.Config[GPTMod
...
+    dataset_root = hf_home / "datasets" / "squad"
+    dataset_root.mkdir(parents=True, exist_ok=True)
+
+    tokenizer = AutoTokenizer(pretrained_model_name=model_name)
```
malay-nagda, line 115: Can we use `NullTokenizer` (and optionally use `AutoTokenizer`)? 

We have CLI arg `use_hf_tokenizer` which you can add as an arg to this method and use like- https://github.com/NVIDIA/NeMo/blob/main/scripts/performance/llm/pretrain_llama3_8b.py#L73-L78

scripts/performance/utils.py
```diff
@@ -495,6 +495,70 @@ def import_ckpt_experiment(executor: run.SlurmExecutor, model: run.Config[GPTMod
...
+    )
+
+    # This will generate both JSONL and packed .bin files
+    datamodule.prepare_data()
+
+    # Verify the output
+    packed_dir = dataset_root / "packed" / model_name.replace("/", "--") / str(seq_length)
+    print(f"Packed files should be in: {packed_dir}")
+    if packed_dir.exists():
+        print("Files found:", list(packed_dir.glob("*.bin")))
```
malay-nagda: Should we raise an exception (crash the job) if dataset is not present?

scripts/performance/llm/finetune_llama4_e128.py
```diff
@@ -0,0 +1,212 @@
...
+        hf_token=args.hf_token,
+        nemo_home=args.nemo_home,
+        wandb_key=args.wandb_key,
+    )
+
+    with run.Experiment(exp_name) as exp:
+        if not SKIP_IMPORT:
+            assert args.hf_token is not None, "HF token is required for importing checkpoint from HuggingFace"
+            exp.add(*import_ckpt_experiment(executor, model(), source=f"hf://{HF_MODEL_URI}"))
+            exp.add(*prepare_squad_dataset_experiment(executor, HF_MODEL_URI, seq_length=4096))
```
malay-nagda: I think we should
- avoid adding an extra Slurm job (as this increases overall exp runtime) if dataset is already present at expected location or
- decouple this from import checkpoint exp (add `SKIP_DATASET_DOWNLOAD`- default set to False)

scripts/performance/llm/finetune_llama3_70b.py
```diff
@@ -120,9 +121,9 @@ def override_recipe_configs(
             get_nmt_tokenizer, library="null", model_name="NullTokenizer", vocab_size=128256
         )
         recipe.model.tokenizer = recipe.data.tokenizer
-    if recipe.data.__fn_or_cls__ == SquadDataModule and not isfile_train_pack_metadata(HF_MODEL_URI, recipe.data):
-        # flag is valid only for SquadDataModule
-        recipe.data.force_redownload = True
+    # if recipe.data.__fn_or_cls__ == SquadDataModule and not isfile_train_pack_metadata(HF_MODEL_URI, recipe.data):
+    #     # flag is valid only for SquadDataModule
+    #     recipe.data.force_redownload = True
```
malay-nagda:
```suggestion
```

scripts/performance/llm/finetune_llama4_e128.py
rhmukundan: Done

scripts/performance/utils.py
rhmukundan: Done

scripts/performance/utils.py
rhmukundan: Done

scripts/performance/utils.py
rhmukundan: Done

scripts/performance/llm/finetune_llama4_e128.py
malay-nagda: Thanks! Can you do the same for 70B too?

scripts/performance/utils.py
malay-nagda: I think you need to add `NullTokenizer` (`vocab_size` will need to be added to args to this method) along with existing tokenizer here like you added to llama4 fine-tuning script (since this will be part of different Slurm exp?).

scripts/performance/utils.py
```diff
@@ -495,6 +495,73 @@ def import_ckpt_experiment(executor: run.SlurmExecutor, model: run.Config[GPTMod
...
+        model_name (str): The name of the model
+        seq_length (int): The sequence length to use for packing. Defaults to 2048.
+    """
+    from pathlib import Path
+
+    from nemo.collections.common.tokenizers.huggingface.auto_tokenizer import AutoTokenizer
+    from nemo.collections.llm.gpt.data.packed_sequence import PackedSequenceSpecs
+    from nemo.collections.llm.gpt.data.squad import SquadDataModule
+
+    nemo_home = Path(get_nemo_home())
```
malay-nagda: I don't think `get_nemo_home` will work as this will be resolved on login node. We don't strictly check on login node for this env var. 

Can you use `args.nemo_home`?

rhmukundan: Done

scripts/performance/llm/finetune_llama4_e128.py
rhmukundan: Done

scripts/performance/utils.py
rhmukundan: Done

scripts/performance/utils.py
```diff
@@ -495,6 +495,102 @@ def import_ckpt_experiment(executor: run.SlurmExecutor, model: run.Config[GPTMod
...
+    from nemo.collections.llm.gpt.data.squad import SquadDataModule
+
+    nemo_home_path = Path(get_nemo_home(nemo_home))
+    dataset_root = nemo_home_path / "datasets" / "squad"
+    dataset_root.mkdir(parents=True, exist_ok=True)
+
+    if use_hf_tokenizer:
+        tokenizer = AutoTokenizer(pretrained_model_name=model_name)
+    else:
+        tokenizer = NullTokenizer(vocab_size=202048)
```
malay-nagda: This should not be hard-coded as we use different vocab sizes for different models.

rhmukundan: Done

scripts/performance/llm/finetune_llama31_405b.py
```diff
@@ -47,6 +45,10 @@
 # downloaded from HuggingFace
 SKIP_IMPORT = False
 
+# Set this to True if dataset is already downloaded. If set to False,
```
guyueh1, line 45: This is not very clear; previously without this PR , are we not downloading the dataset from huggingface? My impression is it's still done somewhere in the dataset building process, just not explicitly, right? I think the difference is here you are separating it out as a new nemo-run experiment. If comment like this, users would think setting it to True without a local file will error out, but in reality it won't?

Could you further explain what different things are happening between False and True here;

scripts/performance/llm/finetune_llama3_70b.py
```diff
@@ -112,9 +124,9 @@
...
```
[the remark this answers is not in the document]
guyueh1: please remove the commented lines

guyueh1:
@rhmukundan please fix merge conflict; to fix the DCO, you need to sign every commit and force push (overwrite the current commits) again (refer to  https://github.com/NVIDIA/NeMo/pull/13893/checks?check_run_id=44185194908 "rebase the branch")

guyueh1:
@rhmukundan is this ready for merge?

guyueh1 (approved):
LGTM

PR #13906, opened 2025-06-12
FCD Metric bugfix: handle empty codes update

rfejgin:
There are instances when Magpie-TTS generates zero-length outputs. If these were then fed into the FCD metric it would crash, breaking our evaluation scripts. Added handling of this case and a unit test.


nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -23,10 +23,7 @@
 be useful to explore).
 """
 
-import warnings
-
```
rfejgin, line 27: cleaned up some unused imports

paarthneekhara (approved):

PR #13907, opened 2025-06-12
[magpietts][eval][bugfix] fixed infer and eval scripts and supported loading wandb hparam config.

XuesongYang:


scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -110,12 +110,10 @@ def evaluate(manifest_path, audio_dir, generated_audio_dir, language="en", sv_mo
...
 
     if language == "en":
-        if asr_model_name == "stt_en_conformer_transducer_large":
-            asr_model = nemo_asr.models.EncDecRNNTBPEModel.from_pretrained(model_name="stt_en_conformer_transducer_large")
-        elif asr_model_name == "nvidia/parakeet-ctc-0.6b":
-            asr_model = nemo_asr.models.EncDecRNNTBPEModel.from_pretrained(model_name="nvidia/parakeet-ctc-0.6b")
-
-        # asr_model = nemo_asr.models.EncDecCTCModelBPE.from_pretrained(model_name="nvidia/parakeet-tdt-1.1b")
+        if asr_model_name in ["nvidia/parakeet-tdt-1.1b", "nvidia/parakeet-ctc-0.6b", "stt_en_conformer_transducer_large"]:
+            asr_model = nemo_asr.models.ASRModel.from_pretrained(model_name=asr_model_name)
```
paarthneekhara, line 114: IIRC, `stt_en_conformer_transducer_large` and `parakeet` return slightly different things when transcribing. For one of them we need to add [0] to the output because it returns a list. Please check if the code works with all ASRs. Maybe this is no longer a problem, but I am quite sure at one point they were different.

shehzeen: +1 and should we also remove "nvidia/parakeet-ctc-0.6b" for now since it is not giving normalized predictions?

XuesongYang: > stt_en_conformer_transducer_large and parakeet return slightly different things when transcribing. For one of them we need to add [0] to the output because it returns a list

@paarthneekhara, standard ASR model inference takes as input a list of audio files so that the output is a list of transcripts. The output of any model should have to choose [0]th item if the input list only contains a single audio. Please check the README files of below models. Maybe ASR folks have fixed the issue as you mentioned 😆
* `stt_en_conformer_transducer_large`: https://huggingface.co/nvidia/stt_en_conformer_transducer_large
* `parakeet-tdt-1.1b`: https://huggingface.co/nvidia/parakeet-tdt-1.1b

But in any case, this PR change is to load pre-trained ASR models using a single API instead of separate classes. I tried to keep the ASR models in the original script.

XuesongYang: @shehzeen `parakeet-ctc-0.6b` transcibed normalized text, but `parakeet-tdt-0.6b-v2` does not. See my tries below,

```python
In [1]: import nemo.collections.asr as nemo_asr

In [2]: parakeet_tdt = nemo_asr.models.ASRModel.from_pretrained(model_name="nvidia/parakeet-tdt-1.1b")
In [5]: print(parakeet_tdt.transcribe(["predicted_audio_87.wav"])[0].text)
sd desdi pass zero zero fail zero fail zero cancelled four hundred and sixteen seventy six

In [6]: stt_en_conformer_transducer_large = nemo_asr.models.ASRModel.from_pretrained(model_name="nvidia/stt_en_conformer_transducer_large")
In [7]: print(stt_en_conformer_transducer_large.transcribe(["predicted_audio_87.wav"])[0].text)
s d desti pass zero zero fail zero fail zero cancelled four hundred sixteen seventy six

In [9]: parakeet_tdt_06b_v2 = nemo_asr.models.ASRModel.from_pretrained(model_name="nvidia/parakeet-tdt-0.6b-v2")
In [10]: print(parakeet_tdt_06b_v2.transcribe(["predicted_audio_87.wav"])[0].text)
SD Des D pass 0 0 fail 0 fail 0 cancelled 416 76

In [11]: parakeet_ctc_06b = nemo_asr.models.ASRModel.from_pretrained(model_name="nvidia/parakeet-ctc-0.6b")
In [12]: print(parakeet_ctc_06b.transcribe(["predicted_audio_87.wav"])[0].text)
sd di pass zero zero fail zero fail zero canceled four hundred and sixteen seventy six
```

XuesongYang: @paarthneekhara the above examples are also good to verify your previous concern disappears.

shehzeen: @XuesongYang got it! Thanks for clarifying. Looks good to me.

shehzeen (approved):

PR #13926, opened 2025-06-15
Onboard LLAMA4 Maverick Finetuning(SFT) with SQUAD Dataset Download Fix

rhmukundan:


scripts/performance/argument_parser.py
```diff
@@ -181,9 +181,8 @@ def parse_cli_args():
     parser.add_argument(
         "-f",
         "--finetuning",
-        choices=["sft", "lora"],
-        help="Finetuning scheme to use. Defaults to 'lora'",
-        default='lora',
+        help="Finetuning scheme. Only SFT (Supervised Fine-Tuning) is supported.",
```
ashbhandare: no need to change the default in arg parser, the default assertion should be handled specifically for llama4.

scripts/performance/argument_parser.py
```diff
@@ -442,4 +441,25 @@ def list_of_strings(arg):
...
+
+    parser.add_argument(
+        "--skip_dataset_download",
+        help="Skips dataset download, finetuning job and only downloads the checkpoint.",
+        action="store_true",
+        required=False,
+    )
+
+    parser.add_argument(
+        "--skip_finetuning",
```
ashbhandare, line 460: you can maybe change it to download_only or something

scripts/performance/llm/finetune_llama4_e128.py
```diff
@@ -0,0 +1,264 @@
...
+    import_ckpt_experiment,
+    isfile_train_pack_metadata,
+    prepare_squad_dataset_experiment,
+    set_exp_logging_configs,
+    set_primary_perf_configs,
+    slurm_executor,
+)
+
+HF_MODEL_URI = "meta-llama/Llama-4-Maverick-17B-128E-Instruct"
+
```
ashbhandare, line 40: comparing with the version in main, this is missing:
```
# Set this to True if checkpoint is available at 'NEMO_HOME'. If set to False,

# extra Slurm job will be scheduled. In this case, if checkpoint is available

# at 'NEMO_HOME', fine-tuning job will use this checkpoint, else, it will be

# downloaded from HuggingFace

SKIP_IMPORT = False



# Set this to True if dataset is already downloaded. If set to False,

# dataset will be downloaded from HuggingFace

SKIP_DATASET_DOWNLOAD = False
```
cant we make these options work as it is

scripts/performance/argument_parser.py
```diff
@@ -442,4 +441,25 @@ def list_of_strings(arg):
         default=0,
     )
 
+    parser.add_argument(
```
ashbhandare, line 445: are these options necessary? Looks like we can handle it with the existing vars SKIP_IMPORT and SKIP_DATASET_DOWNLOAD ? We should try not to diverge too much from main

scripts/performance/llm/finetune_llama4_e128.py
```diff
@@ -0,0 +1,264 @@
...
+        num_layers,
+        hidden_size,
+        etp_size,
+        enable_cuda_graphs,
+        use_mcore_fsdp,
+        recompute_layers,
+        activation_offload_layers,
+    ) = kwargs[0:15]
+
+    # Only configure recipe if finetuning will actually run
```
ashbhandare: what happens if you configure it anyway but just keep the  condition you have added:
`        if args.skip_finetuning is not True:
            exp.add(
                recipe,
                executor=executor,
                name=exp_name,
                plugins=plugins,
            )`

scripts/performance/recommended_model_configs/model_configs_b200.csv
```diff
@@ -25,11 +25,15 @@ pre_train,nemotron4,15b,b200,fp8,64,4096,1,1,1,1,1,2,256,,1,0,0,0,
...
 pre_train,deepseek,v3,b200,bf16,128,4096,2,16,1,8,1,1,1024,1,0,0,0,0,core_attn
-sft,llama3,8b,b200,fp8,8,16384,1,1,1,1,1,1,8,,1,0,0,0,
-sft,llama3,70b,b200,bf16,32,4096,2,4,1,1,5,1,32,,1,0,0,0,
-sft,llama3,70b,b200,fp8,32,4096,2,4,1,1,5,1,32,,1,0,0,0,
-sft,llama3,8b,b200,bf16,8,16384,1,1,1,1,1,1,8,,1,0,0,0,
-pre_train,llama4,e16,b200,bf16,64.0,8192.0,1.0,2.0,1.0,8.0,24.0,1.0,1024.0,1.0,0,0,0,0,
-pre_train,llama4,e128,b200,bf16,128.0,8192.0,1.0,2.0,1.0,64.0,12.0,1.0,1024.0,1.0,0,0,0,0,
-pre_train,vlm_llama4,e16,b200,bf16,64.0,8192.0,1.0,1.0,1.0,16.0,1.0,1.0,1024.0,1.0,0,0,0,0,
-pre_train,vlm_llama4,e128,b200,bf16,128.0,8192.0,1.0,1.0,2.0,128.0,1.0,1.0,1024.0,1.0,0,0,0,0,
+sft,llama3,8b,b200,fp8,8,16384,1,1,1,1,1,1,8,,0,0,0,0
```
ashbhandare: what is changing for llama3 config? Ideally we should only see a change adding the llama4 sft configs

scripts/performance/argument_parser.py
rhmukundan: We would need these args (--skip_finetuning, --skip_dataset_download, --skip_import_checkpoint) in order to control the 3 SLURM jobs from our launch and setup scripts. It is in effect the same behaviour but provides more flexibility. The --skip_finetuning flag is needed when we only want to the checkpoint and dataset downloaded without running the finetuning job (part of the setup phase)

scripts/performance/llm/finetune_llama4_e128.py
rhmukundan: Addressed this in the other thread. We would need the args (`--skip_finetuning`, `--skip_import_checkpoint` and `skip_dataset_download` )  in order to control them from our launch and setup scripts.

scripts/performance/argument_parser.py
rhmukundan: Makes sense, will revert this to reflect what is in `main`  and handle the `sft` check in the llama4 e128 finetuning script.

scripts/performance/llm/finetune_llama4_e128.py
rhmukundan: This comment is a bit misleading. have edited it in the latest commit. It gets skipped based if the `args.skip_finetuning` is not set. Basically, we avoid setting plugins etc for the individual checkpoint and dataset download jobs and only have it set for the finetuning job.

scripts/performance/recommended_model_configs/model_configs_b200.csv
rhmukundan: The order might have changed where I inserted the llama4 sft config somewhere in between. But I have reverted it and added it at the end, so this diff must go away.

PR #13952, opened 2025-06-17
[Draft] Frame-stacking

rfejgin:


nemo/collections/tts/models/magpietts.py
```diff
@@ -388,6 +396,20 @@ def codes_to_audio(self, codes, codes_len):
...
+                embedding = self.audio_embeddings[c + i * C](tokens)
+                if audio_embedding is None:
+                    audio_embedding = embedding
+                else:
+                    audio_embedding += embedding
+        audio_embedding = audio_embedding / (C * self.downsampling_factor)
+        return audio_embedding
+
     def embed_audio_tokens(self, audio_tokens):
         # audio_tokens: (B, C, T')
```
Edresson: It is a duplicate method; we probably need to remove embed_audio_tokens and rename  embed_audio_tokens_with_downsampling to embed_audio_tokens.

nemo/collections/tts/models/magpietts.py
```diff
@@ -1123,14 +1291,22 @@ def process_batch(self, batch, mode="train"):
...
-        audio_codes_target = audio_codes[:, :, 1:]
-        audio_codes_lens_input = audio_codes_lens_target = audio_codes_lens - 1
-        audio_codes_embedded_all = self.embed_audio_tokens(audio_codes) # (B, T, E) # Computing this to be use in the alignment encoder
-        audio_codes_embedded = audio_codes_embedded_all[:, :-1, :] # (B, T', E) Input to the decoder
-
-        audio_codes_mask = get_mask_from_lengths(audio_codes_lens_input)
+        # TODO: @rfejgin: this assert might be slow due to GPU/CPU sync
+        assert (audio_codes[:,:,0] == self.audio_bos_id).all(), "Audio codes do not start with BOS token"
+        if self.downsampling_factor > 1:
+            # repeat the BOS token to downsampling_factor times
```
Edresson: I'm wondering if there is any special reason to duplicate the bos token.

Edresson: Oh, never mind, got it. It is because the bos token is used in inference to the first step.

nemo/collections/tts/models/magpietts.py
rfejgin: Yes, agreed. This PR isn't ready-for-merge, I will clean up such things once before merging.

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -240,23 +246,23 @@ def run_inference(
...
-        apply_attention_prior,
-        attention_prior_epsilon,
-        attention_prior_lookahead_window,
-        start_prior_after_n_audio_steps,
-        "".join([str(l) for l in estimate_alignment_from_layers]) if estimate_alignment_from_layers is not None else "None",
-        "".join([str(l) for l in apply_prior_to_layers]) if apply_prior_to_layers is not None else "None",
-        use_local_transformer,
-        maskgit_n_steps,
-        sv_model
+    checkpoint_name = (
```
rfejgin, line 251: Convert checpoint name construction to use f-strings, which is less error-prone than trying to keep format string and arguments in sync.

Also: make prior inference parameters part of the checkpoint name only if actually using the inference prior. The reason being that the checkpoint names are getting very long (to the point that the full name is not visible in some shell/explorer tools).

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -497,6 +506,9 @@ def main():
     parser.add_argument('--use_cfg', action='store_true')
     parser.add_argument('--use_local_transformer', action='store_true', help="Enables use of local transformer for inference; applies to both Autoregressive and MaskGit sampling.")
     parser.add_argument('--maskgit_n_steps', type=int, default=3)
+    parser.add_argument('--maskgit_noise_scale', type=float, default=0.0)
```
rfejgin, line 509: adding some optional MaskGit-specific command line arguments that are useful for tuning.

rfejgin:
Closing this PR and will open a new one where all commits are squashed signed (hard to fix the signing post-hoc)

PR #14001, opened 2025-06-24
Disable checkpointing for Nemotron-H

scsudhakaran:


nemo/collections/llm/recipes/nemotronh_56b.py
```diff
@@ -149,6 +150,15 @@ def trainer(
 
     callbacks = [
         run.Config(TimingCallback),
+        run.Config(
```
ashbhandare: But if the callback was not added at all why is it still enabled without this fix? Does it explicitly need to be added and disabled?

scsudhakaran: `enable_checkpointing` is set to `None` [here](https://github.com/NVIDIA/NeMo/blob/llmb-nemo-r2.3.0/scripts/performance/utils.py#L449 ) which causes `ModelCheckpoint` to be enabled with default values [ref](https://github.com/Lightning-AI/pytorch-lightning/blob/6b9ddf00dd0a6c41276491641f927147bfedda58/src/lightning/pytorch/trainer/trainer.py#L377)

Setting `enable_checkpointing` to False helps to resolve the issue.

nv-mollys (approved):

PR #14035, opened 2025-06-27
FCD Metric: error handling when provided with codes of unexpected shape

rfejgin:
We have observed very rare cases that the codes provided to the FCD metric by the inference script have an unexpected shape, e.g. 6 codebooks instead of 8. We need to debug the upstream cause of this, but the FCD shouldn't crash either.

We now log a warning if the above situation is detected. Also added a test to check FCD doesn't crash in this case.

### DIFF
--- a/nemo/collections/tts/modules/fcd_metric.py
+++ b/nemo/collections/tts/modules/fcd_metric.py
@@ -172,6 +172,10 @@ class FrechetCodecDistance(Metric):
             logging.warning(f"\nFCD metric received an empty batch of codes - skipping update\n")
             return
 
+        if codes.shape[1] != self.model.codec.num_codebooks:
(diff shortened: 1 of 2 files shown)


nemo/collections/tts/modules/fcd_metric.py
```diff
@@ -172,6 +172,10 @@ def update(self, codes: Tensor, codes_len: Tensor, is_real: bool):
             logging.warning(f"\nFCD metric received an empty batch of codes - skipping update\n")
             return
 
+        if codes.shape[1] != self.model.codec.num_codebooks:
```
paarthneekhara, line 175: What happens in this case? Do we set this metric to 0 in our evals?

rfejgin: If it only happened for one or a few `update()` calls, those updates get ignored.

If all updates had this issues, the `compute()` function will issue a warning about not having seen any frames and FCD will get set to `0.0`.

PR #14038, opened 2025-06-27
EOS detection: check all codebooks

rfejgin:
At inference, we previously only looked at the first codebook to check for EOS being generated. This carried the risk of EOS tokens being returned to the user if they occurred only in codebooks other than codebook zero and would therefore not be detected.

I tested on unseen speakers (subset of `libri_dev_clean`) and found no statistically significant performance changes after this change (as expected).

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -1718,9 +1718,9 @@ class MagpieTTSModel(ModelPT):
 
                 for item_idx in range(all_codes_next_argmax.size(0)):
                     if item_idx not in end_indices:
-                        pred_token = all_codes_next_argmax[item_idx][0].item()
-                        pred_token_multinomial = audio_codes_next[item_idx][0].item()
-                        if (pred_token == self.audio_eos_id) or (pred_token_multinomial == self.audio_eos_id):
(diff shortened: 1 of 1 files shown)


nemo/collections/tts/models/magpietts.py
```diff
@@ -1718,9 +1718,9 @@ def infer_batch(
 
                 for item_idx in range(all_codes_next_argmax.size(0)):
                     if item_idx not in end_indices:
-                        pred_token = all_codes_next_argmax[item_idx][0].item()
-                        pred_token_multinomial = audio_codes_next[item_idx][0].item()
-                        if (pred_token == self.audio_eos_id) or (pred_token_multinomial == self.audio_eos_id):
+                        eos_in_pred_tokens_argmax = (all_codes_next_argmax[item_idx] == self.audio_eos_id).any().item()
```
subhankar-ghosh, line 1721: This is an interesting observation! So, we were checking for eos in only the first codebook? I wonder why?

rfejgin: I think historically folks found that when EOS is generated by the model, it tends to happen for all codebooks at once. But nowhere is that enforced and without that there was a risk that if the model generated an EOS somewhere but not in the first codebook, we would output to the user a frame that has EOS token in it since that frame wouldn't be detected as an end-of-utterance and trimmed. Given a few oddities we've been seeing at inference lately it seemed worthwhile to harden this.

PR #14039, opened 2025-06-27
Interence fix: clean up old files

rfejgin:
Delete leftover generated files (audio and codes) from previous evaluation runs if any exist.

Leaving them there is risky since the same output directory gets reused when we evaluate the same checkpoint (with the same configuration) again. It's especially problematic if the previous evaluation was interrupted and left corrupted files.


paarthneekhara (approved):
LGTM

PR #14040, opened 2025-06-27
inference: add option to include the experiment name in the output folder name

rfejgin:
Adds an option to automatically include the experiment name in the output directory created by the inference script. It helps when analyzing results from checkpoints from multiple training runs. Previously it was hard to tell which training run a checkpoint came from.

The experiment name assumed to be the name of the directory two levels above the checkpoint; this works for the directory structure our training scripts produce.

This is disabled by default.


paarthneekhara (approved):
LGTM

PR #14042, opened 2025-06-27
Set attention backend to "auto" for Nemotron-H

scsudhakaran:


ashbhandare (approved):
LGTM once perf is verified on all platforms

PR #14051, opened 2025-06-27
Inference bugfix: wrong dimension trimmed when saving generated codes for FCD calculation

rfejgin:
At inference we save the generated codes for subsequent analysis in the FCD metric. There was a bug where the wrong dimension was being trimmed to the valid code lengths. It should be the time dimension but was the codebook dimension.

This is likely the root cause of the rare FCD errors we saw, with an unexpected codes tensor shape; it would occur when the valid frame length was less than 8, resulting in trimming in the codebook dimension.


paarthneekhara (approved):
LGTM!

XuesongYang (approved):
Thanks for the fix!

PR #14055, opened 2025-06-28
Inference: pad very short signals before embedding them

rfejgin:
The speaker embedding model crashes on very short signals. So we zero-pad the end of the signal if it's less than 0.5 seconds long before running it through the speaker embedding model.

### DIFF
--- a/scripts/magpietts/evaluate_generated_audio.py
+++ b/scripts/magpietts/evaluate_generated_audio.py
@@ -17,6 +17,7 @@ import os
 import pprint
 import string
 
+import numpy as np
 import torch
 
 import nemo.collections.asr as nemo_asr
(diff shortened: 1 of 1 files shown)


scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -85,9 +86,27 @@ def transcribe_with_whisper(whisper_model, whisper_processor, audio_path, langua
     result = transcription[0]
     return result
 
+def pad_audio_to_min_length(audio_np: np.ndarray, sampling_rate: int, min_seconds: float) -> np.ndarray:
+    """
+    Pad audio to make it at least `min_seconds` long by adding silence at the end if needed.
```
subhankar-ghosh, line 91: Some additional comments on why this might be helpful or use case will be helpful.

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -85,9 +86,27 @@ def transcribe_with_whisper(whisper_model, whisper_processor, audio_path, langua
...
+        padding_needed = min_samples - n_samples
+        audio_np = np.pad(audio_np, (0, padding_needed), mode='constant', constant_values=0)
+    return audio_np
+
+
 def extract_embedding(model, extractor, audio_path, device, sv_model_type):
     speech_array, sampling_rate = librosa.load(audio_path, sr=16000)
-
+    # pad to 0.5 seconds as the extractor may not be able to handle very short signals
+    speech_array = pad_audio_to_min_length(speech_array, int(sampling_rate), min_seconds=0.5)
```
subhankar-ghosh, line 109: Has this been tested, does this affect the final evaluation metrics in any way?

subhankar-ghosh (approved):
LGTM. Left some minor comments.

rfejgin:
> LGTM. Left some minor comments.

Thanks @subhankar-ghosh. It had been set to auto-merge so already got merged as soon as you approved, but I'll still look at your comments.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: I'll add more comments and include in a future PR.

rfejgin: I took another look: there's a comment on the use [here](https://github.com/nvidia/NeMo/pulls?q=is%3Apr+is%3Aopen+base%3Amagpietts_2503). I think it makes more sense to comment on the reason for ouse where the used than in the generic function itself.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: I tested that the padding works correctly. Did not collect pre/post evaluation stats. This should only kick in very rarely, when the generated speech is shorter than 0.5 sec.

paarthneekhara (commented):
Looks good to me.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Following up on this: I ran `libri_unseen_test` with and without the padding fix and found no statistically significant differences in WER and SSIM.

PR #14072, opened 2025-06-30
Load master weights from checkpoint

kunlunl:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/tests/lightning/pytorch/strategies/test_megatron_strategy.py
+++ b/tests/lightning/pytorch/strategies/test_megatron_strategy.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pytest
 from unittest.mock import MagicMock, patch
+
+import pytest
 
 from nemo.lightning.pytorch.strategies import MegatronStrategy
 


nemo/lightning/pytorch/strategies/megatron_strategy.py
```diff
@@ -252,6 +253,7 @@ def __init__(
         find_unused_parameters: bool = False,
         ckpt_load_optimizer: bool = True,
         ckpt_save_optimizer: bool = True,
+        ckpt_load_main_params: bool = False,
```
ananthsub, line 261: when should users set this to True?

ananthsub: from https://github.com/NVIDIA/Megatron-LM/commit/1e42279a994e88ee9329d32143cb8f6cc60e9f34 shouldn't this only be allowed to be set when `ckpt_load_optimizer=False`?

kunlunl: > When should users set this to True?

When checkpoints don't contain master weights, and the precision of checkpoints is higher than model params, for example, checkpoints are bf16, but model prams are fp8, in this case, use fp8 model weights to initialize master weights is less accurate, so you can set this flag to True, to initialize master weights directly from the bf16 checkpoints.

kunlunl: > from https://github.com/NVIDIA/Megatron-LM/commit/1e42279a994e88ee9329d32143cb8f6cc60e9f34 shouldn't this only be allowed to be set when ckpt_load_optimizer=False?

Yeah, conceptually, if the checkpoints already contain master weights, then there is no need to initialize master weights from model weights or checkpoints. But I am not sure whether setting `ckpt_load_optimizer` to False here has the same effect as setting `--no-load-optim` in megatron. Anyway, I added an assertion to prevent setting them both to True, I think it needs a nemo expert to see whether it is reasonable.

PR #14080, opened 2025-07-01
make feature_dir default to None to avoid adding feature_dir in

XuesongYang:
avoid `feature_dir` in the eval config.

### DIFF
--- a/nemo/collections/tts/data/text_to_speech_dataset.py
+++ b/nemo/collections/tts/data/text_to_speech_dataset.py
@@ -44,7 +44,7 @@ from nemo.utils.decorators import experimental
 class DatasetMeta:
     manifest_path: Path
     audio_dir: Path
-    feature_dir: Path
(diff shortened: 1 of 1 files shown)


nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -195,8 +195,8 @@ def _preprocess_manifest(
             sample = DatasetSample(
                 dataset_name=dataset_name,
                 manifest_entry=entry,
-                audio_dir=Path(dataset.audio_dir),
-                feature_dir=Path(dataset.feature_dir) if dataset.feature_dir is not None else None,
+                audio_dir=dataset.audio_dir,
```
XuesongYang, line 198: per `dataclass` definition, `audio_dir` and `feature_dir` are already the type of `Path`, so no need to enforce the conversion again.

PR #14081, opened 2025-07-01
[eval] make compute_fcd optional for now.

XuesongYang:
I frequently came across failures of FCD computation when evaluating several datasets, especially challenging datasets. @rfejgin helped a lot to root-cause the issue and fix. 

I would propose to disable FCD computation by default to save the total time of inference and evaluation. It helps to scale up with multiple datasets with multiple repeats (ususally 10 repeats).

### DIFF
--- a/scripts/magpietts/infer_and_evaluate.py
+++ b/scripts/magpietts/infer_and_evaluate.py
@@ -136,7 +136,8 @@ def run_inference(
         legacy_codebooks=False,
         clean_up_disk=False,
         hparams_file_from_wandb=False,
-        log_exp_name=False
+        log_exp_name=False,
+        compute_fcd=False,
     ):
     # Load model
     if hparams_file is not None and checkpoint_file is not None:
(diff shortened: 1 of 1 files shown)


rfejgin:
@XuesongYang : are you still seeing stability issues after [this fix](https://github.com/NVIDIA/NeMo/pull/14051)? I haven't seen FCD-related issues since then and I've been running quite a few evals (admittedly not with challenging text, though).

Or is speed the concern? Is there a substantial speed difference with FCD on/off?

I was kind of am hoping to get some experience with the metric (e.g. correlations between it and other findings) from the team using it. Maybe default to on an allow disabling from the command line?

XuesongYang:
> I was kind of am hoping to get some experience with the metric (e.g. correlations between it and other findings) from the team using it. Maybe default to on an allow disabling from the command line?

good point! FCD metric is still valuable. Let's keep it by default but with an option to disable it. 

I haven't rebased onto the fix yet. Let me try it later.

XuesongYang:
> > I was kind of am hoping to get some experience with the metric (e.g. correlations between it and other findings) from the team using it. Maybe default to on an allow disabling from the command line?
> 
> good point! FCD metric is still valuable. Let's keep it by default but with an option to disable it.
> 
> I haven't rebased onto the fix yet. Let me try it later.

@rfejgin flipped the default about fcd computation.

rfejgin (approved):
Looks good to me.

Issue #13619, opened 2025-05-16
Low GPU utilization with NEST and OOMptimizer


MahmoudAshraf97:
Hi,
I'm using the default NEST config with a dataset that is around 9k hours, I use tarred dataset and OOMptimizer with a single H100, but the utilization is lower than expected, This is my training dataset config
```yaml
train_ds:
    manifest_filepath: /home/mahmo/development/data/tarred_everything/tarred_audio_manifest.json
    noise_manifest: null
    sample_rate: 16000
    batch_size: 512
    batch_duration: 1100
    shuffle: true
    num_workers: 8
    pin_memory: true
    use_start_end_token: false
    trim_silence: false
    max_duration: 60.0
    min_duration: 1.0
    drop_last: true
    use_lhotse: true
    is_tarred: true
    tarred_audio_filepaths: /home/mahmo/development/data/tarred_everything/audio_0.tar
    use_bucketing: true
    bucket_duration_bins: [10.08325,13.92325,17.076,19.448,19.744,19.9,19.968,20.0,20.064,20.224,20.48,24.069,29.91019,30.0]
    bucket_batch_size: [238,167,135,115,113,112,112,111,111,110,108,89,69,68]
    batch_augmentor:
      _target_: nemo.collections.asr.modules.ssl_modules.MultiSpeakerNoiseAugmentation
      prob: 0.2
      noise_ratio: 0.5
      min_r_speech: -5.0
      max_r_speech: 5.0
      min_r_noise: -5.0
      max_r_noise: 20.0
      min_mix_rate: 0.5
      max_mix_rate: 0.5
      min_num_segments: 1
      max_num_segments: 1
      min_num_speakers: 1
      max_num_speakers: 1
```
These are the metrics on wandb:

![Image](https://github.com/user-attachments/assets/83f3edae-e2c4-41b4-908a-322843460b6e)

![Image](https://github.com/user-attachments/assets/76c2a63a-2717-4153-a69e-f3438fa8a638)

Note that with regular Fast conformer with CTC loss I'm able to get consistent ~95% utilization with a batch size of 32 only along with lhotse and tarred dataset
cc: @pzelasko @stevehuang52 
Thanks in advance

MahmoudAshraf97:
This is another experiment with 8x H100, I kept everything the same as above except for using a regular sharded dataset instead of a tarred dataset, and divided the batch sizes by 8 to keep the global batch size the same

[...]

PR #14089, opened 2025-07-01
Improve NEST GPU Utilization 2/N

MahmoudAshraf97:
This PR is among a series aimed to improve the training speed and GPU utilization of NEST models #13619 

after profiling, the masking module takes a good chunk of the total training loop time while involving no computation, thus reducing the utilization, this PR reduces python for-loops as much as possible, and converts the masking operation to a singe function call

As shown in the following graph, This PR along with #14086 reduces the training step time by almost 17%
{4751020E-438F-4DA7-AE1F-E6F14653A76A}

The green is with the PRs applied

cc @stevehuang52

### DIFF
--- a/nemo/collections/asr/modules/ssl_modules/masking.py
+++ b/nemo/collections/asr/modules/ssl_modules/masking.py
@@ -15,7 +15,6 @@
 
 from typing import Optional, Union
 
-import numpy as np
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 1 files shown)


nemo/collections/asr/modules/ssl_modules/masking.py
```diff
@@ -99,29 +98,42 @@ def forward_without_overlap(self, input_feats: torch.Tensor, input_lengths: torc
             masks (Tensor): the generated masks, shape=(batch, features, time)
         """
         batch_size = input_feats.size(0)
-        mask_value = self.mask_embedding.unsqueeze(-1)
         masks = torch.zeros_like(input_feats)
-        maksed_feats = input_feats.clone()
+        masked_feats = input_feats
```
stevehuang52, line 102: Will this also change `input_feats` as well when `masked_feats` is updated? Ideally we should return the `masked_feats` while keeping `input_feats` unchanged.

MahmoudAshraf97: I did confirm that by:
```python
module = RandomBlockMasking(80, allow_overlap=False, mask_prob=0.01, block_size=40).cuda()
for i in range(100):
    input = torch.randn(1, 80, 1000).cuda()
    input_len = torch.tensor([1000]).cuda()
    masked_feats, masks = module(input, input_len)
    assert not torch.allclose(masked_feats, input)
```

stevehuang52: Awesome, thanks

stevehuang52:
@MahmoudAshraf97 could you please fix the DCO error?

MahmoudAshraf97:
@stevehuang52 Fixed

MahmoudAshraf97:
@stevehuang52 Can you rerun CICD?

stevehuang52 (approved):
Looks good to me, thanks for the improvement!

PR #14143, opened 2025-07-06
[eval][bugfix] avoid overidding cross attention maps for multiple repeats

XuesongYang:
predicted audio files are saved in the dir `parent/audio/repeat_0/predicted_audio_0.wav`, but cross-attention map was overridden by the repeats, and saved in `parent/audio/cross_attn_map_0.png`. Now it is saved in the same dir as predicted audio, `parent/audio/repeat_0/cross_attn_map_0.png`, without overridden.

### DIFF
--- a/scripts/magpietts/infer_and_evaluate.py
+++ b/scripts/magpietts/infer_and_evaluate.py
@@ -316,7 +316,7 @@ def run_inference(
                 print(f"Time taken for inference: {et-st}", predicted_audio.size())
                 for idx in range(predicted_audio.size(0)):
                     cross_attn_map_image = Image.fromarray(cross_attention_maps[idx])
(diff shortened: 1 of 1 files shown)


rfejgin (changes requested):
Now that the `.png` files are in the audio/repeat directory, could you add deletion of them in the file that cleans that directory up?
https://github.com/NVIDIA/NeMo/blob/5e36aa126356389717ec10b0e0aa614667027250/scripts/magpietts/infer_and_evaluate.py#L108

paarthneekhara (approved):
Looks good to me.

XuesongYang:
> Now that the `.png` files are in the audio/repeat directory, could you add deletion of them in the function that cleans that directory up?
> 
> https://github.com/NVIDIA/NeMo/blob/5e36aa126356389717ec10b0e0aa614667027250/scripts/magpietts/infer_and_evaluate.py#L108

addressed in my updates.

PR #14166, opened 2025-07-09
[magpietts] added an argument 'binarize_atten_prior' to trigger whether apply prior binarization.

XuesongYang:
* added an arugment `binarize_atten_prior` to trigger whether apply prior binarization or not during training or evaluating. Previously, the training always applied binary prior. Now it is configurable.
* make prior past and future context window configurable during inference. 
* hardcoded `text_lens - 2` to decide if text sentence is finished or not.
* minor fixed typo..

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -363,7 +363,7 @@ class MagpieTTSModel(ModelPT):
         self._codec_model.eval()
         with torch.no_grad(), torch.autocast(device_type=audio.device.type, dtype=torch.float32):
             codes, codes_len = self._codec_model.encode(audio=audio, audio_len=audio_len)
-            # Add a timestep to begining and end of codes tensor
+            # Add a timestep to beginning and end of codes tensor
             bos_tensor = torch.full(
                 (codes.size(0), codes.size(1), 1), audio_bos_id, dtype=codes.dtype, device=codes.device
             )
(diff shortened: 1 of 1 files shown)


nemo/collections/tts/models/magpietts.py
```diff
@@ -1476,18 +1484,37 @@ def construct_inference_prior(self, prior_epsilon, cross_attention_scores,
...
+                    # Current position
+                    _attn_prior[bidx, 0, text_time_step_attended[bidx]] = base_value
+
+                    # Past context
+                    for past_timestep in range(self.inference_prior_past_context):
+                        past_pos = max(1, text_time_step_attended[bidx] - (past_timestep + 1))
+                        if self.binarize_attn_prior:
+                            _attn_prior[bidx, 0, past_pos] = 1.0  # Binary: no decay, always 1.0
+                        else:
+                            decay_factor = self.inference_prior_past_decay ** (past_timestep + 1)
```
blisc, line 1507: Does this change the prior to an exponential decay? When did we decide to change to this?

paarthneekhara: In training we have arguments like "prior_future_decay", "prior_past_decay"

paarthneekhara:
I think the `binarize_attn_prior` might be redundant. If we dont want to use it, we will just set `use_alignment_encoder` to false. If we a model is trained with it and we want to finetune without it, we can set `binarize_prior_after` to a very high value.

blisc:
Closing as we re-factored it differently

PR #14234, opened 2025-07-14
Improve NEST GPU Utilization 3/N

MahmoudAshraf97:
This PR is among a series aimed to improve the training speed and GPU utilization of NEST models https://github.com/NVIDIA/NeMo/issues/13619

This PR mainly focuses on removing redundant operations from the dataloader and increasing its effeciency by removing unused instances of `noisy_audio` tensors, this tensor is basically unused everywhere except the forward pass, but it's recomputed in many places although it will be overwritten later, so we only compute it once at the end when all the noise related sampling is finalized.

The noise tensor is also removed from the final result that gets passed to the training process, since it's an unused argument, this tensor is redundant and the IPC overhead of moving it to the training process is not justifiable (assuming 128 batch size and 20s samples, this tensor is 160MB)

the second change is using the random stdlib instead of numpy, all the operations are faster than their numpy counterparts except for `np.random.multinomial`

# Results
Tests were made to compare main branch with this PR to train two models with L and XL presets, wandb graphs are smoothed with gaussian=10
{05E97233-0C8E-45ED-BEC9-9CB40296D1D9}

The XL model finished the same number of steps less time

{6059702C-9C24-47A2-8AB3-4789F0A746D5}
The large model shows slightly higher GPU utilization but finished the same number of steps in the same amount of time

I'd say that these results are inconclusive and require further testing but i don't have the compute budget to run them,
for future work, profiling showed that most of the training loop time is spent in the masking and tokenization, so these areas for improvement

cc @pzelasko @stevehuang52

### DIFF
--- a/nemo/collections/asr/data/ssl_dataset.py
+++ b/nemo/collections/asr/data/ssl_dataset.py
@@ -43,6 +43,8 @@ class AudioNoiseItem:
     audio_len: Union[Tensor, None] = None
     noise: Union[Tensor, None] = None
     noise_len: Union[Tensor, None] = None
+    noisy_audio: Union[Tensor, None] = None
+    noisy_audio_len: Union[Tensor, None] = None
 
 
 @dataclass
--- a/nemo/collections/asr/data/ssl_dataset.py
+++ b/nemo/collections/asr/data/ssl_dataset.py
@@ -136,7 +136,6 @@ def _audio_noise_collate_fn(batch: List[AudioNoiseItem], batch_augmentor: Any = 
     else:
         output.noisy_audio = output.audio + output.noise
         output.noisy_audio_len = output.audio_len
-        output.noise, output.noise_len = None, None
 
     return output
 
@@ -466,7 +465,6 @@ class LhotseAudioNoiseDataset(torch.utils.data.Dataset):
         else:
             output.noisy_audio = output.audio + output.noise
             output.noisy_audio_len = output.audio_len
-            output.noise, output.noise_len = None, None
 
         return output
 
--- a/nemo/collections/asr/data/ssl_dataset.py
+++ b/nemo/collections/asr/data/ssl_dataset.py
@@ -43,8 +43,6 @@ class AudioNoiseItem:
     audio_len: Union[Tensor, None] = None
     noise: Union[Tensor, None] = None
     noise_len: Union[Tensor, None] = None
-    noisy_audio: Union[Tensor, None] = None
-    noisy_audio_len: Union[Tensor, None] = None
 
 
 @dataclass
@@ -103,12 +101,8 @@ def _audio_noise_collate_fn(batch: List[AudioNoiseItem], batch_augmentor: Any = 
     noises = [x.noise for x in batch]
     noise_lengths = [x.noise_len for x in batch]
 
-    noisy_audios = [x.noisy_audio for x in batch]
-    noisy_audio_lengths = [x.noisy_audio_len for x in batch]
-
     audio_signal_list = []
     noise_signal_list = []
-    noisy_audio_signal_list = []
     for i, audio in enumerate(audios):
         audio_len = audio.size(0)
         if audio_len < max_audio_len:
@@ -123,31 +117,26 @@ def _audio_noise_collate_fn(batch: List[AudioNoiseItem], batch_augmentor: Any = 
             noise = torch.nn.functional.pad(noise, pad)
         noise_signal_list.append(noise[:max_audio_len])
 
-        noisy_audio = noisy_audios[i]
-        noisy_audio_len = noisy_audio.size(0)
-        if noisy_audio_len < max_audio_len:
-            pad = (0, max_audio_len - noisy_audio_len)
-            noisy_audio = torch.nn.functional.pad(noisy_audio, pad)
-        noisy_audio_signal_list.append(noisy_audio[:max_audio_len])
-
     audio_signal = torch.stack(audio_signal_list).float()
     audio_lengths = torch.stack(audio_lengths).long()
     noise_signal = torch.stack(noise_signal_list).float()
     noise_lengths = torch.stack(noise_lengths).long()
-    noisy_audio_signal = torch.stack(noisy_audio_signal_list).float()
-    noisy_audio_lengths = torch.stack(noisy_audio_lengths).long()
 
     output = AudioNoiseBatch(
         audio=audio_signal,
         audio_len=audio_lengths,
         noise=noise_signal,
         noise_len=noise_lengths,
-        noisy_audio=noisy_audio_signal,
-        noisy_audio_len=noisy_audio_lengths,
+        noisy_audio=None,
+        noisy_audio_len=None,
     )
 
     if batch_augmentor is not None:
         output = batch_augmentor(output)
+    else:
+        output.noisy_audio = output.audio + output.noise
+        output.noisy_audio_len = output.audio_len
+        output.noise, output.noise_len = None, None
 
     return output
 
@@ -344,8 +333,6 @@ class AudioNoiseDataset(audio_to_text.AudioToCharDataset):
             audio_len=audio_len,
             noise=noise,
             noise_len=noise_len,
-            noisy_audio=audio + noise,
-            noisy_audio_len=audio_len,
         )
         return item
 
@@ -421,8 +408,6 @@ class TarredAudioNoiseDataset(audio_to_text.TarredAudioToCharDataset):
             audio_len=audio_len,
             noise=noise,
             noise_len=noise_len,
-            noisy_audio=audio + noise,
-            noisy_audio_len=audio_len,
         )
         return item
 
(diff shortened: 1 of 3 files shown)


nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -43,8 +43,6 @@ class AudioNoiseItem:
     audio_len: Union[Tensor, None] = None
     noise: Union[Tensor, None] = None
     noise_len: Union[Tensor, None] = None
-    noisy_audio: Union[Tensor, None] = None
```
stevehuang52: we want to keep them here for completeness of the class, and they're needed in case the audio and noise are not directly added together

nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -43,8 +43,6 @@ class AudioNoiseItem:
     audio_len: Union[Tensor, None] = None
     noise: Union[Tensor, None] = None
     noise_len: Union[Tensor, None] = None
```
MahmoudAshraf97:
```suggestion
    noise_len: Union[Tensor, None] = None
    noisy_audio: Union[Tensor, None] = None
    noisy_audio_len: Union[Tensor, None] = None
```

nemo/collections/asr/data/ssl_dataset.py
MahmoudAshraf97: readded

nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -123,31 +119,26 @@ def _audio_noise_collate_fn(batch: List[AudioNoiseItem], batch_augmentor: Any =
...
 
     output = AudioNoiseBatch(
         audio=audio_signal,
         audio_len=audio_lengths,
         noise=noise_signal,
         noise_len=noise_lengths,
-        noisy_audio=noisy_audio_signal,
-        noisy_audio_len=noisy_audio_lengths,
+        noisy_audio=None,
+        noisy_audio_len=None,
```
MahmoudAshraf97:
```suggestion
```

nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -460,23 +447,30 @@ def __init__(self, noise_manifest: str | None = None, batch_augmentor_cfg: DictC
...
 
         output = AudioNoiseBatch(
             audio=audios,
             audio_len=audio_lens,
             noise=sampled_noises,
             noise_len=sampled_noises_lens,
-            noisy_audio=audios + sampled_noises,
-            noisy_audio_len=audio_lens,
+            noisy_audio=None,
+            noisy_audio_len=None,
```
MahmoudAshraf97:
```suggestion
```

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -129,20 +128,17 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
         return AudioNoiseBatch(
             sample_id=batch.sample_id,
             audio=batch.audio,
             audio_len=batch.audio_len,
-            noise=noise,
-            noise_len=noise_len,
-            noisy_audio=noisy_audio,
-            noisy_audio_len=noisy_audio_len,
+            noise=None,
+            noise_len=None,
```
MahmoudAshraf97:
```suggestion
```

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -232,21 +226,17 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
```
MahmoudAshraf97:
```suggestion
```

MahmoudAshraf97 (commented):
Since the default is already none, maybe it's not needed to explicitly specify None as the value

MahmoudAshraf97:
@stevehuang52 PR edited and added the results

MahmoudAshraf97:
not stale

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -129,20 +128,15 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
-            noisy_audio_len[i] = audio_lengths[i]
             noise[i] = noise_signal
             noise_len[i] = audio_lengths[i]
 
         return AudioNoiseBatch(
             sample_id=batch.sample_id,
             audio=batch.audio,
             audio_len=batch.audio_len,
-            noise=noise,
-            noise_len=noise_len,
```
stevehuang52, line 143: We want to keep `noise` and `noise_len` so that the caller can have full awareness of what noise is applied to the original audio, in case the noise is not simply added to the original audio.

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -232,21 +224,15 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
-            noisy_audio[i] = audio_signal[i] + noise_signal
-            noisy_audio_len[i] = audio_lengths[i]
             noise[i] = noise_signal
             noise_len[i] = audio_lengths[i]
 
         return AudioNoiseBatch(
             sample_id=batch.sample_id,
             audio=batch.audio,
             audio_len=batch.audio_len,
-            noise=noise,
```
stevehuang52, line 246: We want to keep noise and noise_len so that the caller can have full awareness of what noise is applied to the original audio, in case the noise is not simply added to the original audio.

nemo/collections/asr/models/ssl_models.py
```diff
@@ -1018,8 +1018,6 @@ def training_step(self, batch: ssl_dataset.AudioNoiseBatch, batch_idx: int):
         log_probs, encoded_len, masks, tokens = self.forward(
             input_signal=batch.audio,
             input_signal_length=batch.audio_len,
-            noise_signal=batch.noise,
-            noise_signal_length=batch.noise_len,
```
stevehuang52: Please revert the change

stevehuang52:
> The noise tensor is also removed from the final result that gets passed to the training process, since it's an unused argument, this tensor is redundant and the IPC overhead of moving it to the training process is not justifiable (assuming 128 batch size and 20s samples, this tensor is 160MB)

@MahmoudAshraf97 could you please revert this change? I understand that it looks redundant, but if there's a future extension to the augmentation that's no simply adding noise to the original audio, or we want to add another loss in the model to estimate the noise applied to the audio, we still need to pass the original noise to the model.

nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -123,31 +119,24 @@ def _audio_noise_collate_fn(batch: List[AudioNoiseItem], batch_augmentor: Any =
...
-        noisy_audio=noisy_audio_signal,
-        noisy_audio_len=noisy_audio_lengths,
     )
 
     if batch_augmentor is not None:
         output = batch_augmentor(output)
+    else:
+        output.noisy_audio = output.audio + output.noise
+        output.noisy_audio_len = output.audio_len
+        output.noise, output.noise_len = None, None
```
MahmoudAshraf97:
```suggestion
```

nemo/collections/asr/data/ssl_dataset.py
```diff
@@ -460,23 +445,28 @@ def __init__(self, noise_manifest: str | None = None, batch_augmentor_cfg: DictC
...
-            noisy_audio=audios + sampled_noises,
-            noisy_audio_len=audio_lens,
         )
 
         if self.batch_augmentor is not None:
             output = self.batch_augmentor(output)
+        else:
+            output.noisy_audio = output.audio + output.noise
+            output.noisy_audio_len = output.audio_len
+            output.noise, output.noise_len = None, None
```
MahmoudAshraf97:
```suggestion
```

nemo/collections/asr/models/ssl_models.py
```diff
@@ -1018,8 +1018,6 @@ def training_step(self, batch: ssl_dataset.AudioNoiseBatch, batch_idx: int):
         log_probs, encoded_len, masks, tokens = self.forward(
             input_signal=batch.audio,
             input_signal_length=batch.audio_len,
```
MahmoudAshraf97:
```suggestion
            input_signal_length=batch.audio_len,
            noise_signal=batch.noise,
            noise_signal_length=batch.noise_len,
```

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -129,20 +128,15 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
-            noise_len=noise_len,
-            noisy_audio=noisy_audio,
-            noisy_audio_len=noisy_audio_len,
+            noisy_audio=batch.audio + noise,
+            noisy_audio_len=noise_len,
```
MahmoudAshraf97, line 141:
```suggestion
            noise=noise,
            noise_len=noise_len,
            noisy_audio=batch.audio + noise,
            noisy_audio_len=noise_len,
```

nemo/collections/asr/modules/ssl_modules/augmentation.py
```diff
@@ -232,21 +224,15 @@ def __call__(self, batch: AudioNoiseBatch) -> AudioNoiseBatch:
...
 
         return AudioNoiseBatch(
             sample_id=batch.sample_id,
             audio=batch.audio,
             audio_len=batch.audio_len,
-            noise=noise,
-            noise_len=noise_len,
-            noisy_audio=noisy_audio,
-            noisy_audio_len=noisy_audio_len,
+            noisy_audio=batch.audio + noise,
```
MahmoudAshraf97, line 238:
```suggestion
            noise=noise,
            noise_len=noise_len,
            noisy_audio=batch.audio + noise,
```

MahmoudAshraf97:
@stevehuang52 changes applied

stevehuang52 (approved):
LGTM, thanks~

PR #14251, opened 2025-07-16
Fix in Nemotron-H script

scsudhakaran:


scripts/performance/llm/pretrain_nemotronh_56b.py
```diff
@@ -150,7 +143,7 @@ def override_recipe_configs(
         )
     ]
 
-    custom_env_vars = {"NVTE_FUSED_ATTN": "0", "TRANSFORMERS_OFFLINE": "0"}
+    custom_env_vars = {}
```
ashbhandare, line 146: do we need the TRANSFORMERS_OFFLINE flag still? I am assuming the NVTE_FUSED_ATTN is not required because we want to use attn backend auto

scripts/performance/llm/pretrain_nemotronh_56b.py
```diff
@@ -138,10 +135,6 @@ def override_recipe_configs(
     exp_config = f"gpus{args.num_gpus}_tp{tp_size}_pp{pp_size}_cp{cp_size}_vp{vp_size}_mbs{mbs}_gbs{gbs}"
     exp_name = f"{splitext(basename(__file__))[0]}_{args.compute_dtype}_{exp_config}"
 
-
-    if args.gpu.lower() == 'gb200':
```
ashbhandare, line 142: Is this flag not needed for the nemo 25.07 container? If not, it needs to be removed from all workloads

scripts/performance/llm/pretrain_nemotronh_56b.py
scsudhakaran: We don't need `TRANSFORMERS_OFFLINE` flag.

Yes. `NVTE_FUSED_ATTN` removed because attention_backed is set to "auto"

scripts/performance/llm/pretrain_nemotronh_56b.py
scsudhakaran: This was a bug in the llmb-nemo-2.4.0 branch. The check is still there in code: https://github.com/NVIDIA/NeMo/pull/14251/files#diff-641dc57550ec17d449f11c4f5c03fc98e39d881173b316b4f023422498f63db6L155

PR #14256, opened 2025-07-17
Bugfix for Hyena to the get_t function which comes up when doing longer context inference

jstjohn:
When doing longer context inference with different lengths of `t` that grow beyond previous queries (and L_cache) then you can get into a situation where your t is too short.


JRD971000 (approved):
Good catch! Thanks @jstjohn

PR #14333, opened 2025-07-25
[Flux] small fixes

CarlosGomes98:
> [!IMPORTANT]  

# What does this PR do ?

A few fixes to flux, most importantly ensuring each dp rank has a different seed.

Diffusion models

# Changelog

- Ensure each dp rank has a different seed. This is crucial so that different noise is added by each dp rank.
- Make loss be computed inside the autocast block.
- Disable per token loss and gradient accumulation fusion (the latter causes a huge memory blowup)

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/diffusion/models/flux/model.py
```diff
@@ -463,6 +464,10 @@ def __init__(
         self.text_precached = self.t5_params is None or self.clip_params is None
         self.image_precached = self.vae_config is None
```
mmarcinkiewicz, line 470: where does `self.seed=seed` happen?

scripts/flux/flux_controlnet_training.py
```diff
@@ -76,6 +76,7 @@ def flux_controlnet_training() -> run.Partial:
             MegatronFluxControlNetModel,
             flux_params=run.Config(FluxModelParams),
             flux_controlnet_config=run.Config(FluxControlNetConfig),
+            seed=42,
```
mmarcinkiewicz, line 79: I guess it's overridden somewhere in the config?

nemo/collections/diffusion/models/flux/model.py
CarlosGomes98: True, thanks

scripts/flux/flux_controlnet_training.py
```diff
@@ -76,6 +76,7 @@ def flux_controlnet_training() -> run.Partial:
             MegatronFluxControlNetModel,
             flux_params=run.Config(FluxModelParams),
             flux_controlnet_config=run.Config(FluxControlNetConfig),
+            seed=42
```
mmarcinkiewicz, line 79: this gets overridden, right?

scripts/flux/flux_controlnet_training.py
CarlosGomes98: Im not sure where the best place for the seed is. Its not really a parameter of flux or control net, so i dont think it makes sense there. Its more of a training parameter, so I would put it in this file. Arguably, this could be randomly generated, with the user passing it explicitly in the cli if they want to

scripts/flux/flux_controlnet_training.py
```diff
@@ -191,7 +192,7 @@ def fp8_test(custom_fsdp=True) -> run.Partial:
     )
     recipe.model.flux_params.device = 'cuda'
     recipe.model.flux_params.flux_config = run.Config(
-        FluxConfig, ckpt_path='/ckpts/nemo_flux_transformer.safetensors', guidance_embed=False
+        FluxConfig, ckpt_path='/ckpts/nemo_flux_transformer.safetensors', guidance_embed=False, calculate_per_token_loss=False, gradient_accumulation_fusion=False
     )
     recipe.trainer.devices = 2
     recipe.data = flux_datamodule('/mingyuanm/dataset/fill50k/fill50k_tarfiles/')
```
mmarcinkiewicz, line 204: @tomlifu @Victor49152 did you see the hardcoded path^

nemo/collections/diffusion/models/flux/model.py
```diff
@@ -462,6 +463,11 @@ def __init__(
         self.model_type = ModelType.encoder_or_decoder
         self.text_precached = self.t5_params is None or self.clip_params is None
         self.image_precached = self.vae_config is None
+        self.seed = seed
+
+    def setup(self, stage: str):
+        super().setup(stage)
+        torch.manual_seed(self.seed + 100 * parallel_state.get_data_parallel_rank())
```
alpha0422, line 473: IIUC, I think we should use global rank instead of data parallel rank, because if we use other parallelisms then there's still the chance two ranks have the same seed.

scripts/flux/flux_controlnet_training.py
alpha0422: IIRC, the default value should be a random seed in other models.

nemo/collections/diffusion/models/flux/model.py
mmarcinkiewicz: But we do want for example ranks in the same TP group to have the same seed, no?

alpha0422: Hmm, I was thinking the noise map part for diffusion model, but I just realized that noise map should be deemed as data, so the ones with the same data rank should have the same seed.

scripts/flux/flux_controlnet_training.py
CarlosGomes98: I am unsure of the proper way to implement this for nemo. If I call random() here, I believe different ranks will use different seeds, which is not the desired behaviour at this point. Is there a way to obtain a random seed here which will be shared across all ranks? 

I suppose I can have rank 0 generate it and broadcast to other ranks, is that the proper way in NeMo? @tomlifu

tomlifu: If you'd like to use the same seed for every rank, random() would not work as it will create different seeds on each rank. I think setting it to a specific number, 42 in this case, will make sure all ranks are using 42.

yaoyu-33:
plz fix format issue

CarlosGomes98:
@yaoyu-33 I dont think this is an issue with my files, the error happens before black or isort are run https://github.com/NVIDIA-NeMo/NeMo/actions/runs/16523626000/job/47522819552?pr=14333

Locally black and isort pass.

gautham-kollu:
Rebased the branch does not help either.  

```
Error: Error: Pushing to https://github.com/CarlosGomes98/NeMo
remote: Permission to CarlosGomes98/NeMo.git denied to artbataev.
fatal: unable to access 'https://github.com/CarlosGomes98/NeMo/': The requested URL returned error: 403

```

Suspecting as it is in a fork and in a private branch, possibly does not have access

gautham-kollu (approved):

PR #14393, opened 2025-08-03
Magpietts 2503 po july2025

shehzeen:
Preference optimization updates. 
1. Added thresholding for CERs to ignore groups with very high CERs
2. Local transformer GRPO enabled

### DIFF
--- a/nemo/collections/tts/models/magpietts_preference_optimization.py
+++ b/nemo/collections/tts/models/magpietts_preference_optimization.py
@@ -441,6 +441,11 @@ class MagpieTTSModelOnlinePO(MagpieTTSModel):
             raise ValueError(f"Received loss_type of {self.loss_type}, but the model only accepts one of ['grpo', 'dr_grpo']")
         self.scale_rewards = self.cfg.get('scale_rewards', True)
         self.max_decoder_steps = self.cfg.get('max_decoder_steps', 430)
+        # If the best record in the group is above this threshold, we will not use that group for training
+        self.best_cer_threshold = self.cfg.get('best_cer_threshold', 1.0)
+        # If the worst record in the group exceeds this threshold, we will not use that group for training
+        self.worst_cer_threshold = self.cfg.get('worst_cer_threshold', 1.0)
+        
 
     def state_dict(self, destination=None, prefix='', keep_vars=False):
         state_dict = super().state_dict(destination, prefix, keep_vars)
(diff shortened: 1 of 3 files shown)


nemo/collections/tts/models/magpietts_preference_optimization.py
```diff
@@ -441,6 +441,11 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             raise ValueError(f"Received loss_type of {self.loss_type}, but the model only accepts one of ['grpo', 'dr_grpo']")
         self.scale_rewards = self.cfg.get('scale_rewards', True)
         self.max_decoder_steps = self.cfg.get('max_decoder_steps', 430)
+        # If the best record in the group is above this threshold, we will not use that group for training
+        self.best_cer_threshold = self.cfg.get('best_cer_threshold', 1.0)
+        # If the worst record in the group exceeds this threshold, we will not use that group for training
+        self.worst_cer_threshold = self.cfg.get('worst_cer_threshold', 1.0)
```
blisc: Why are these both defaulted to 1?

nemo/collections/tts/models/magpietts_preference_optimization.py
```diff
@@ -636,6 +660,8 @@ def generate_and_reward(self, batch, num_generations_per_item, mode='train'):
         advantages = [x['advantage'] for x in batch_metrics]
         advantages = torch.tensor(advantages, device=self.device)
         print("Mean reward: ", all_groups_mean_reward)
+        # import ipdb; ipdb.set_trace()
```
blisc: Remove debugging comment

blisc (changes requested):

nemo/collections/tts/models/magpietts_preference_optimization.py
shehzeen: If it is 1.0, we don't filter out anything and include all groups in loss calculation (Since max CER is capped to 1.0)

blisc: Can you add that to your comment? That we cap at 1.0 and the default will not do any filtering

blisc (approved):

nemo/collections/tts/models/magpietts_preference_optimization.py
blisc: I don't think we cap cer at 100% either. I've seen >1.0 on the short utterances before

shehzeen: Added comments.

I only do the clamping during preference optimization and not in the evaluate script. It is done at around line 556 in nemo/collections/tts/models/magpietts_preference_optimization.py

`cer_gt = min(max(cer_gt, 0.0), 1.0)  # Ensure CER is in [0, 1]
`

PR #14395, opened 2025-08-04
Add pyxis flag for container writable.

sudostock:
Certain benchmarks require a writable container for caches and other files.

This fixes errors we've seen when enroot config doesn't match test environment.
**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/scripts/performance/executors.py
+++ b/scripts/performance/executors.py
@@ -71,7 +71,7 @@ def slurm_executor(
     custom_bash_cmds = [] if custom_bash_cmds is None else custom_bash_cmds
     err_msgs = []
     mounts = []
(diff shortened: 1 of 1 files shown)


guyueh1:
Sharing offline chat with @sudostock 
> This flag allows the container environment to be writable. ie basic thing like creating cache files will fail if disabled. Which leads to weird file not found errors with our benchmarks. It does not change the sqsh file on disk, and changes only exist for the lifetime of the job. Usually a cluster would set it to true in enroot config so jobs don't need to set it. This PR is because some CSPs see issues as their out of box default for enroot is writable = false.

LGTM!

PR #14408, opened 2025-08-05
[magpietts] custom tokenizer for text context

paarthneekhara:


nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -33,7 +33,7 @@
 from nemo.utils import logging
 
 
-def setup_tokenizers(all_tokenizers_config, use_text_conditioning_tokenizer, mode='train'):
+def setup_tokenizers(all_tokenizers_config, use_text_conditioning_tokenizer, text_conditioning_tokenizer_name, mode='train'):
```
blisc: Can you update the default of text_conditioning_tokenizer_name to None? And if None is passed, we use the aggregated_tokenizer instead of loading one from AutoTokenizer? I presume in this case, we would also need to default one of the tokenizer in the aggregated one; maybe the first one in the list.

blisc (changes requested):

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
paarthneekhara: I have updated it. However, we'll need to handle backward compatibility for older checkpoints by adding model.text_conditioning_tokenizer_name="google-t5/t5-small" in the yaml.

blisc: Can you handle it inside of infer_and_evaluate instead in the `update_config` using something like the `legacy_codebooks`?
We should aim to better handle all future checkpoints rather than having backwards compatibility for unreleased checkpoints.

blisc: Discussed internally. Two points to be done

- We should add this text conditioning tokenizer into the Aggregated tokenizer
- Enable an option to merge the text tokenizer and the text context tokenizer, does not have to be default

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1098,26 +1098,35 @@ def __init__(self, tokenizers: List[Union[BaseTokenizer, PreTrainedTokenizerBase
...
         for idx, tokenizer in enumerate(tokenizers):
-            self.tokenizers[tokenizer_names[idx]] = tokenizer
-            toknizer_offsets[tokenizer_names[idx]] = tokenizer_offset
+            tokenizer_name = tokenizer_names[idx]
+            self.tokenizers[tokenizer_name] = tokenizer
+            tokenizer_offsets[tokenizer_name] = tokenizer_offset
             if isinstance(tokenizer, BaseTokenizer):
                 tokens.extend(tokenizer.tokens)
                 num_tokens = len(tokenizer.tokens)
+                tokenizer_pad_ids[tokenizer_name] = tokenizer.pad
```
blisc: We should add the offset to this as well

scripts/magpietts/evalset_config.py
```diff
@@ -53,4 +53,24 @@
         'audio_dir' : '/',
         'feature_dir' : None,
     },
+    'j_riva_digits': {
```
blisc: Please undo this change

nemo/collections/tts/models/magpietts.py
```diff
@@ -888,6 +910,15 @@ def compute_alignment_loss(self, attention_scores, text_lens, audio_lens, dec_co
         )
         return alignment_loss
 
+    def embed_context_text(self, context_text_tokens):
+        if self.legacy_text_conditioning:
+            context_text_tokens = context_text_tokens - self.tokenizer.tokenizer_offsets[self.text_conditioning_tokenizer_name]
+            context_text_embedded = self.context_text_embedding(context_text_tokens)  # (B, L, E)
```
blisc, line 920: It looks like we're having issues with the legacy path specifically since the aggregated tokenizer adds offsets to the tokens, but we kept separate embeddings in the legacy path so it shouldn't have these offsets.
It'll be best to subtract the offset here from the tokens instead of trying to work it in the Aggregated tokenization class or in the dataset

scripts/magpietts/evalset_config.py
paarthneekhara: Done.

blisc (approved):

nemo/collections/tts/models/magpietts.py
paarthneekhara: Yes. I believe that's how it is now being done --- Aggregated tokenizer and dataset class work the same way, we just subtract the offset right here.

PR #14411, opened 2025-08-06
Add option for LoRA with Transformer Engine op fuser

timmoon10:
# What does this PR do ?

This PR adds a variant of the `LoRALinear` module using Transformer Engine's operation fuser. It also adds logic in the existing `LoRALinear` module so that it can call the fused variant. This is an experimental feature.

Collection: NLP

# Changelog

- Add `TEFusedLoRALinear` module that implements LoRA with TE op fuser.
- Add logic in `LoRALinear` module to internally call `TEFusedLoRALinear`.

# Usage

Set `use_transformer_engine_op_fuser=True` in the LoRA model config (see https://github.com/NVIDIA/NeMo/pull/13776).

```python
from nemo.collections.llm import Llama2Config70B
from nemo.collections.llm.api import _setup
from nemo.collections.llm.gpt.model import LlamaModel
from nemo.collections.llm.peft.lora import LoRA

config = llm.Llama2Config70B(..., use_transformer_engine_op_fuser=True)
model = LlamaModel(config, ...)
peft = LoRA(...)
_setup(model=model, model_transform=peft, ...)
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Depends on Transformer Engine bugfixes: https://github.com/NVIDIA/TransformerEngine/pull/2027, https://github.com/NVIDIA/TransformerEngine/pull/2063
- Benefits from Transformer Engine optimizations: https://github.com/NVIDIA/TransformerEngine/pull/2042, https://github.com/NVIDIA/TransformerEngine/pull/2014


nemo/collections/llm/peft/lora.py
```diff
@@ -465,6 +838,7 @@ def transform(self, m: nn.Module, name=None, prefix=None):
             input_is_parallel, in_features, out_features, disable_sp_comm, base_linear_is_parallel = (
                 get_adapter_attributes_from_linear(m)
             )
+            enable_op_fuser = hasattr(m, "config") and getattr(m.config, "use_transformer_engine_op_fuser", False)
```
cuichenx: Is this fuser specific to some models? I feel like `use_transformer_engine_op_fuser` should be an argument in the LoRA class, not the model class.

nemo/collections/llm/peft/lora.py
```diff
@@ -47,13 +51,381 @@ class LoRALinear(AdapterWrapper):
...
+        # Fused implementation
+        if self._op_fuser_enabled:
+            return self._fused_forward(x)
+
         linear_output, bias, layernorm_output = self.base_linear_forward(x, *args, **kwargs)
         adapter_output = self.adapter(layernorm_output.contiguous())
         adapter_output = adapter_output.reshape(linear_output.shape)
         return linear_output + adapter_output, bias
 
+    def _fused_forward(self, x: torch.Tensor) -> tuple[torch.Tensor, Optional[torch.Tensor]]:
```
cuichenx: nit, but this functionality seems better as a subclass FusedLoRALinear(LoRALinear) rather than an input arg to init

nemo/collections/llm/peft/lora.py
timmoon10: Currently, this option can be used to enable the fuser in the MLP:
- NeMo: https://github.com/NVIDIA/NeMo/blob/454fabc2e436994b028b748dc7d0d351758db84d/nemo/collections/llm/gpt/model/base.py#L164
- Megatron-LM: https://github.com/NVIDIA/Megatron-LM/blob/2f1027d222e23b1ac5d917802bdddd01dfd10598/megatron/core/models/gpt/gpt_layer_specs.py#L375-L376

My thinking is that this option will enable the op fuser everywhere, similar to how `model.transformer_engine=True` enables TE everywhere. We could split it up and have more granular options, which would make debugging easier but also make it harder to use.

cuichenx: I see, thanks for the context, in that case I agree this is fine

timmoon10:
Currently I don't see a significant difference between the fused and unfused impl (174 ms/step with a small Llama model on B200). We skip one add (fused into dgrad GEMM), but the dropout implementation is quite unoptimized. We should see speedup as more fusions are supported in TE.

cuichenx (dismissed):
LGTM

ko3n1g (approved):

PR #14437, opened 2025-08-08
Fix sequence packing loss calculation

rayandasoriya:
# What does this PR do ?

Fix sequence packing loss calculation

# Changelog

- Fix sequence packing loss calculation

The current loss calculation with sequence packing for one example with a smaller example is significantly high.

On the dataset side, I created a dataset with a smaller example to reproduce this issue.
```
{"messages": [{"role": "user", "content": "What is 1+2?"}, {"role": "assistant", "content": "The answer is 3."}]}
```
I turned off the data parallelism, used a single batch size, one epoch and max_seq_len of 32. When I ran the fine tuning with the same setup, here are the results:
```
# Without sequence packing
reduced_train_loss=3.660, global_step=0.000, consumed_samples=1.000, train_step_timing in s=1.900, val_loss=3.540
# With sequence packing
reduced_train_loss=5.290, global_step=0.000, consumed_samples=1.000, train_step_timing in s=4.360, val_loss=3.650
```

Even with a smaller example and minimum configuration, the loss difference persists. The below code changes fixes the issue.

1. nemo/collections/nlp/data/language_modeling/megatron/gpt_sft_dataset.py and nemo/collections/llm/gpt/data/core.py: The comment mentions that `The second eos_id index marks the length of the original unpadded sequence`. However, in the code, we are using the first index of eos_id which actually represents the end of turn and not the end of sequence.

2. nemo/utils/sequence_packing_utils.py: We are [not considering](https://github.com/NVIDIA/NeMo/blob/d0cb20dace8aa0212416c2b43fc3d53026e5c45f/nemo/collections/llm/gpt/data/core.py#L592) the pad_id in loss_mask calculation without sequence packing. So, we should not consider pad_id with sequence packing as well to be consistent.

Overall results:
```
# Without sequence packing
reduced_train_loss=3.660, global_step=0.000, consumed_samples=1.000, train_step_timing in s=1.900, val_loss=3.540
# With sequence packing (Before fix)
reduced_train_loss=5.290, global_step=0.000, consumed_samples=1.000, train_step_timing in s=4.360, val_loss=3.650
# With sequence packing (After fix)
reduced_train_loss=3.660, global_step=0.000, consumed_samples=1.000, train_step_timing in s=2.120, val_loss=3.540
```


dimapihtar:
@rayandasoriya thanks for the PR.

could you please fix code style by adding `# pylint: skip-file` & `# flake8: noqa` lines at the beginning of `nlp/data/language_modeling/megatron/gpt_sft_dataset.py`?

cuichenx:
Hi @rayandasoriya , thanks for your PR. We recently fixed packed sequence loss problem for chat datasets in this PR: https://github.com/NVIDIA/NeMo/pull/14369
Could you test the changes here?

rayandasoriya:
@cuichenx thanks for pointng out to the PR. I ran with that and the results still don't look good.  The reason is that my packed dataset does not have `loss_mask` as the key. So, it falls back to creating loss mask using `answer_start_idx`.

rayandasoriya:
@dimapihtar I have added those lines to the file.

dimapihtar (approved):
LGTM. Thank you!

PR #14455, opened 2025-08-11
Frame Stacking

rfejgin:


scripts/magpietts/infer_and_evaluate.py
```diff
@@ -240,23 +246,23 @@ def run_inference(
...
-        apply_attention_prior,
-        attention_prior_epsilon,
-        attention_prior_lookahead_window,
-        start_prior_after_n_audio_steps,
-        "".join([str(l) for l in estimate_alignment_from_layers]) if estimate_alignment_from_layers is not None else "None",
-        "".join([str(l) for l in apply_prior_to_layers]) if apply_prior_to_layers is not None else "None",
-        use_local_transformer,
-        maskgit_n_steps,
-        sv_model
+    checkpoint_name = (
```
rfejgin, line 336: Convert checkpoint name construction to use f-strings. The format string was getting long and it was getting difficult to keep the order of the `{}` (placeholders) and arguments in sync.

Also: inference prior parameters are now only part of the checkpoint if we are actually using the inference prior. The reason being that the checkpoint names are getting very long (to the point that the full name was not visible in some shell/explorer tools).

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -497,6 +506,9 @@ def main():
     parser.add_argument('--use_cfg', action='store_true')
     parser.add_argument('--use_local_transformer', action='store_true', help="Enables use of local transformer for inference; applies to both Autoregressive and MaskGit sampling.")
     parser.add_argument('--maskgit_n_steps', type=int, default=3)
+    parser.add_argument('--maskgit_noise_scale', type=float, default=0.0)
```
rfejgin, line 603: adding some optional MaskGit-specific command line arguments that are useful for tuning.

nemo/collections/tts/models/magpietts.py
```diff
@@ -567,79 +588,184 @@ def forward(self, dec_input_embedded, dec_input_mask, cond, cond_mask, attn_prio
...
+                print()
+
+    def get_forbidden_tokens(self) -> list[int]:
+        """
+        Returns a list of tokens that should not be sampled.
+        """
+        # Note that audio EOS is not included because we *do* need to sample it at the end of the utterance.
+        return [self.mask_token_id, self.audio_bos_id, self.context_audio_bos_id, self.context_audio_eos_id]
+
+    def local_transformer_sample_maskgit(self, dec_output, temperature=0.7, topk=80, unfinished_items={}, finished_items={}, use_cfg=False, cfg_scale=1.0, n_steps=3, noise_scale=0.0, fixed_schedule=None, dynamic_cfg_scale=False, sampling_type=None):
```
rfejgin, line 706: MaskGit sampling has a lot of options and parameters that can be tuned, so this function has quite a few knobs and experimental features. We'll reduce them as we settle on a sampling algorithm.

nemo/collections/tts/models/magpietts.py
```diff
@@ -567,79 +587,156 @@ def forward(self, dec_input_embedded, dec_input_mask, cond, cond_mask, attn_prio
...
 
             # apply CFG
             if use_cfg:
                 actual_batch_size = logits.size(0) // 2
                 conditional_logits = logits[:actual_batch_size]
                 unconditional_logits = logits[actual_batch_size:]
-                cfg_logits = cfg_scale * conditional_logits +  (1.0 - cfg_scale) * unconditional_logits
+                if not dynamic_cfg_scale:
+                    current_cfg_scale = cfg_scale
+                else:
```
rfejgin, line 785: experimental: dynamically vary CFG scale through sampling

nemo/collections/tts/models/magpietts.py
```diff
@@ -649,29 +746,54 @@ def local_transformer_sample_maskgit(self, dec_output, temperature=0.7, topk=80,
                 logits[item_idx, :, :] = float('-inf')
                 logits[item_idx, :, self.audio_eos_id] = 0.0
 
+            # Disallow generation of MASK (and other special) tokens
```
rfejgin: Sampling of special tokens (MASK specifically) was observed; explicitly disallowing them here.

nemo/collections/tts/models/magpietts.py
```diff
@@ -649,29 +746,54 @@ def local_transformer_sample_maskgit(self, dec_output, temperature=0.7, topk=80,
...
                 sampled_codes[actual_batch_size:] = sampled_codes[:actual_batch_size]
                 probs[actual_batch_size:] = probs[:actual_batch_size]
-            confidences  = torch.gather(probs, dim=2, index=sampled_codes.unsqueeze(-1)).squeeze(-1)
-
-            # set confidence to max for unmasked codebooks so that they will remain unmasked
-            confidences.scatter_(index=topk_indices, dim=1, src=max_confidence*torch.ones_like(topk_indices, dtype=torch.float))
-
+            if sampling_type != "purity_causal" and sampling_type != "purity_default":
+                confidences = torch.gather(probs, dim=2, index=sampled_codes.unsqueeze(-1)).squeeze(-1)
+            else:
```
rfejgin, line 817: "purity sampling", from [this paper](https://arxiv.org/abs/2304.01515): to measure confidence, rather than logit value, use the maximum logit value in the associated codebook. This seems to work a bit better than standard confidence.

nemo/collections/tts/models/magpietts.py
```diff
@@ -649,29 +746,54 @@ def local_transformer_sample_maskgit(self, dec_output, temperature=0.7, topk=80,
...
+                # use the max probability across all tokens for each codebook as the confidence for each codebook; known as "purity sampling"
+                confidences = probs.max(dim=2)[0]
             # replace entries in sampled_codes with previously unmasked codebooks
             sampled_codes.scatter_(dim=1, index=topk_indices, src=unmasked_codes)
-            # optionally: add noise to confidences here (as in token-critic paper) (not implemented)
-
+            #  add noise to confidences (as in token-critic paper, https://arxiv.org/abs/2209.04439)
+            if noise_scale > 0.0:
```
rfejgin, line 823: Experimental: adding noise to logits to increase diversity (not mentioned in the original MaskGit paper but a similar mechanism was included in the official MaskGit implementation).

nemo/collections/tts/models/magpietts.py
```diff
@@ -711,36 +833,40 @@ def local_transformer_sample_autoregressive(self, dec_output, temperature=0.7, t
...
-            )  # (B, num_tokens_per_codebook)
-            codebook_logits_rescored = codebook_logits.clone()
-            codebook_logits_rescored[indices_to_remove] = float('-inf')
-
-            codebook_probs = torch.softmax(codebook_logits_rescored / temperature, dim=-1)  # (B, num_tokens_per_codebook)
-            codebook_preds = torch.multinomial(codebook_probs, 1)  # (B, 1)
-            all_preds.append(codebook_preds)
-        all_preds = torch.cat(all_preds, dim=1).long()  # (B, num_codebooks)
+        all_preds = [[] for _ in range(self.frame_stacking_factor)]
+        for fs_index in range(self.frame_stacking_factor):
```
rfejgin, line 891: Iterate over frames in the stack

nemo/collections/tts/models/magpietts.py
```diff
@@ -419,31 +428,35 @@ def compute_local_transformer_logits(self, dec_out, audio_codes_target, targets_
...
+        | seq. index |    0    |    1    |    2    |    3    |    4    |    5    |    6    |    7    |    8    |
         +------------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
-
+        
         dec_out: (B, T', E)
         audio_codes_target: (B, C, T')
         targets_offset_by_one: bool, if False, the target for index 0 is codebook 0, for index 1 is codebook 1, etc. (autoregressive)
                                      if True,  the target for index 1 is codebook 0, for index 2 is codebook 1, etc. (MaskGit)
         """
+        C = audio_codes_target.size(1)
```
blisc: Slight nitpick: consider using self.num_all_tokens_per_codebook instead

nemo/collections/tts/models/magpietts.py
```diff
@@ -567,79 +589,156 @@ def forward(self, dec_input_embedded, dec_input_mask, cond, cond_mask, attn_prio
...
+                print("  |timestep| ", end="")
+            print(c, end="")
+        print()
+
+    def get_forbidden_tokens(self) -> list[int]:
+        """
+        Returns a list of tokens that should not be sampled.
+        """
+        # Note that audio EOS is not included because we *do* need to sample it at the end of the utterance.
+        return [self.mask_token_id, self.audio_bos_id, self.context_audio_bos_id, self.context_audio_eos_id]
```
blisc: What about the reserved tokens? Alternatively, if EOS is the only one we care about. Maybe we should have an allowlist instead

blisc: Alternatively, we could define this in the SpecialAudioToken enum itself

scripts/magpietts/evalset_config.py
```diff
@@ -54,6 +55,13 @@
         'manifest_path' : '/datap/misc/speechllm_codecdatasets/manifests/t5_exp/dev_clean_withContextAudioPaths_withTargetCodes_evalset_tiny.json',
         'audio_dir' : '/datap/misc/Datasets/LibriTTS',
         'feature_dir' : '/datap/misc/Datasets/LibriTTS',
+        'load_cached_codes_if_available': False
```
blisc: We are no longer using this file, please undo changes here

blisc (commented):
Overall LGTM. Just some minor comments

scripts/magpietts/evalset_config.py
rfejgin: do mean the entire `evalset_config.py` is not being used? Are we using something else instead? in the latest `infer_and_evaluate.py` it still seems to be used

rfejgin: Oh I see that we standardized the file in Subhankar's recent PR I will remove my changes.

rfejgin: Removed in latest commit

nemo/collections/tts/models/magpietts.py
rfejgin: I think it has to be `self.num_audio_codebooks`. Will change.

nemo/collections/tts/models/magpietts.py
rfejgin: Good catch regarding the reserved tokens. Addressed in the latest commit. Also refactored that part a bit to be easier to maintain.

nemo/collections/tts/models/magpietts.py
rfejgin: Fixed in latest commit

rfejgin:
PR comments have been addressed. Main additional changes since last review:
* EOS handling fixes for both training and inference. 
  * Already seeing far fewer end-of-utterance cutoff issues than before (even with existing checkpoints and just inference fixes).
  * Models with training-side fixes are in training
* Added detection of reserved tokens in forbidden tokens check. Also refactored that part to make it easier to maintain. 
* Detect of configurations not compatible with frame stacking and raising an error
* Comments, docs, cleanup

nemo/collections/tts/models/magpietts.py
```diff
@@ -388,17 +426,18 @@ def codes_to_audio(self, codes, codes_len):
...
-        audio_embedding = audio_embedding / audio_tokens.size(1)
+        for i in range(frame_stacking_factor):
+            for c in range(C):
+                tokens = audio_tokens[:,c , i::self.frame_stacking_factor]
+                embedding = self.audio_embeddings[c + i * C](tokens)
+                if audio_embedding is None:
+                    audio_embedding = embedding
+                else:
+                    audio_embedding += embedding
+        audio_embedding = audio_embedding / (C * self.frame_stacking_factor)
```
blisc: Instead of passing frame_stacking_factor to the function, let's use self.frame_stacking_factor

nemo/collections/tts/models/magpietts.py
```diff
@@ -419,31 +458,35 @@ def compute_local_transformer_logits(self, dec_out, audio_codes_target, targets_
...
-            codes = audio_codes_target[:, codebook_num] # (B, T')
-            codes = codes.reshape(-1) # (B*T',)
-            codebook_embedding = self.audio_embeddings[codebook_num](codes) # (B*T', E)
-            local_transformer_input.append(codebook_embedding)
+        for fs_index in range(self.frame_stacking_factor):
+            for codebook_num in range(C):
+                codes = audio_codes_target[:, codebook_num, fs_index::self.frame_stacking_factor] # (B, T')
+                codes = codes.reshape(-1) # (B*T',)
+                codebook_embedding = self.audio_embeddings[codebook_num + fs_index * C](codes) # (B*T', E)
+                local_transformer_input.append(codebook_embedding)
```
blisc: Can you explain what we're doing here? What is audio_codes_target and why are we adding this to the input?

nemo/collections/tts/models/magpietts.py
```diff
@@ -567,81 +619,164 @@ def forward(self, dec_input_embedded, dec_input_mask, cond, cond_mask, attn_prio
...
+        B, C = codes.shape
+        if B > 1:
+            print("Warning: visualizing only first batch element")
+        codes = codes.clone().detach().cpu().numpy()[0]
+        codes = [code_to_str(c) for c in codes]
+        for i, c in enumerate(codes):
+            if (i) % (C/frame_stacking_rate) == 0:
+                print("  |timestep| ", end="")
+            print(c, end="")
+        print()
```
blisc: If you want to keep this function, change the prints to logging.debug

blisc: Same comment applies to all `if debug_print: print(...)` statements

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -35,7 +37,7 @@ class SpecialAudioToken(Enum):
     """
     Enum for the special tokens to use in the MagpieTTS model.
     The special tokens are appended at the end of the codebook after the actual audio codec tokens.
-    The actual codeco index is this value below plus the number of codec tokens - do not use the Enum directly.
+    The actual codec index is the value below plus the number of codec tokens - do not use the Enum directly.
```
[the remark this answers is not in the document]
rfejgin: fixed

nemo/collections/tts/models/magpietts.py
```diff
@@ -96,14 +103,21 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+        self.mask_token_id = cfg.get('forced_mask_token_id', get_token_index(SpecialAudioToken.MASK_TOKEN))
+        self.num_all_tokens_per_codebook = cfg.get('forced_num_all_tokens_per_codebook', num_audio_tokens + len(SpecialAudioToken))
         self.use_bpe_char_tokenizer = cfg.get('use_bpe_char_tokenizer', False)
 
+        # The frame stacking factor controls how many consecutive frames are processed together by the base decoder
+        # (and then refined into individual frames by the local transformer). A frame stacking factor of 1 means no
+        # frame stacking. We have a separate embedding table for each of the stacked frames, e.g. for frame stacking
+        # factor of 3, the entries of codebook 0 appear 3 times in the embedding table.
+        self.frame_stacking_factor = cfg.get('frame_stacking_factor', 1)
+        assert not 'downsample_factor' in cfg, '`downsample_factor` is deprecated, use `frame_stacking_factor` instead'
```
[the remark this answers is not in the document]
rfejgin: fixed

nemo/collections/tts/models/magpietts.py
```diff
@@ -1199,7 +1392,7 @@ def process_batch(self, batch, mode="train"):
                 )
 
                 aligner_encoder_loss = self.alignment_encoder_loss(
-                    attn_logprob=aligner_attn_logprobs, in_lens=context_tensors['text_lens'], out_lens=audio_codes_lens_input
+                    attn_logprob=aligner_attn_logprobs, in_lens=context_tensors['text_lens'], out_lens=audio_codes_lens_input # is this right?
```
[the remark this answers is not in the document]
rfejgin: cleaned up (not relevant as we don't currently support the alignment encoder with frame stacking)

nemo/collections/tts/models/magpietts.py
```diff
@@ -655,23 +790,45 @@ def local_transformer_sample_maskgit(self, dec_output, temperature=0.7, topk=80,
...
+                # the conditional and unconditional get different noise and must be fixed to be the same again
+                confidences[actual_batch_size:] = confidences[:actual_batch_size]                
+            confidence_eps = 0.1
+            assert confidences.max() + confidence_eps < max_confidence, f"Predicted confidence is approaching max_confidence: {confidences.max()}"
+            # for unmasked codebooks, set confidence to max so that they will remain unmasked
+            confidences.scatter_(index=topk_indices, dim=1, src=max_confidence*torch.ones_like(topk_indices, dtype=torch.float))
+        rand_sampling = False
+        if rand_sampling:
+            print("Using random sampling")
+            confidences = torch.rand_like(confidences)
```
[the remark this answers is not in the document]
rfejgin: Fixed

nemo/collections/tts/models/magpietts.py
rfejgin: This function is for teacher-forced training and validation. The `audio_codes_target` contains ground truth codes. For the LT transformer, you can think of the codebook dimension like what we usually have as a time dimension. In the AR case (causal LT decoder), each codebook can depend on previous (in codebook index) codebooks. We are building an input to the local transformer using the ground truth "history" (in codebook index). Also, since the LT operates independently for each time step (note: not codebook), we fold time dimension into the batch dimension and process all timesteps in parallel. **This is all unchanged from the existing local transformer code.**

The only change in this PR is that we add an external loop across _frame stack indices_ since we handle codebooks differently depending on their position in the frame stack -- different positions in the stack have different embeddings. The indexing `audio_codes_target[:, codebook_num, fs_index::self.frame_stacking_factor]` selects all ground truth tokens of a particular codebook at a particular frame index.

Should I add some comments in the code?

nemo/collections/tts/models/magpietts.py
rfejgin: Sure - now switched to `logging.debug`

nemo/collections/tts/models/magpietts.py
rfejgin: done

nemo/collections/tts/models/magpietts.py
```diff
@@ -1231,32 +1424,41 @@ def process_batch(self, batch, mode="train"):
...
+                ## Maskgit ##
                 # randomly replace some positions with MASK_TOKEN
-                audio_codes_masked, mask_tokens_mask = self.maskgit_apply_random_mask(audio_codes_target)
+                audio_codes_masked, mask_tokens_mask = self.maskgit_apply_random_mask(audio_codes_target_unstacked)
+                # TODO @rfejgin: the very last position might be padding but the local transformer might look at it as part of
+                #                of a pair where the first position is valid. Is this an issue?
                 local_transformer_logits = self.compute_local_transformer_logits(dec_out[:,dec_context_size:,:], audio_codes_masked, targets_offset_by_one=True)
-                #audio_codes_masked = audio_codes_masked[:, 1:, :]
-                local_transformer_loss, _ = self.compute_loss(local_transformer_logits, audio_codes_target, audio_codes_lens_target, mask_tokens_mask)
+                local_transformer_loss, _ = self.compute_loss(local_transformer_logits, audio_codes_target_unstacked, audio_codes_lens_target_unstacked, mask_tokens_mask, frame_stacking_factor=self.frame_stacking_factor)
```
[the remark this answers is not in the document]
rfejgin: We will look into this corner cases later. Only affects MaskGit with frame stacking, both of which are disabled.

nemo/collections/tts/models/magpietts.py
```diff
@@ -1123,12 +1303,25 @@ def process_batch(self, batch, mode="train"):
...
-        audio_codes_input = audio_codes[:, :, :-1]  # B, C, T'
-        audio_codes_target = audio_codes[:, :, 1:]
-        audio_codes_lens_input = audio_codes_lens_target = audio_codes_lens - 1
-        audio_codes_embedded_all = self.embed_audio_tokens(audio_codes) # (B, T, E) # Computing this to be use in the alignment encoder
-        audio_codes_embedded = audio_codes_embedded_all[:, :-1, :] # (B, T', E) Input to the decoder
+        if self.frame_stacking_factor > 1:
+            # repeat the BOS token to frame_stacking_factor times. This is necessary since at inference
+            # we need to start autoregressive generation from a full stack indicating BOS.
+            # TODO: @rfejgin: this assert might be slow due to GPU/CPU sync
+            assert (audio_codes[:,:,0] == self.audio_bos_id).all(), "Audio codes do not start with BOS token"
```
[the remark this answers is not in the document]
rfejgin: Not in production yet; prefer to keep this for safety until this code stabilizes.

rfejgin:
I think this is ready for merge now, from my end.

nemo/collections/tts/models/magpietts.py
rfejgin: Added comments in the code

nemo/collections/tts/models/magpietts.py
blisc: Can you replace all new print statements with logging equivalents?

Can you remove this debug_print parameter? Use NEMO_TESTING=1 env var instead

rfejgin: Updated. We now use the NEMO_TESTING environment variable to control whether logging is on. I still kept a local variable (`debug_maskgit`) guard around the logging statements since they are very detailed and I didn't want to affect runtime by calling logging.nemo() so frequently. `debug_maskgit` is controlled by `NEMO_TESTING`.

rfejgin: Also I believe there are no remaining new print statements (they are all `nemo.logging` calls now).

PR #14533, opened 2025-08-20
Inference: Fix for speaker embedding of short files

rfejgin:
When Magpie generates very short / zero length outputs we pad them before extracting speaker embeddings so as to not crash the embedding model. 

The fix here: the padded signal wasn't actually being given to the embedding model (hence those rare crashes persisted). Needed to write it to a temporary file first and give the file path to the embedding model (which doesn't accept NumPy arrays).

(I was attempting to include this in the previous PR related to speaker embeddings but it got merged already, sorry for the churn)

### DIFF
--- a/scripts/magpietts/evaluate_generated_audio.py
+++ b/scripts/magpietts/evaluate_generated_audio.py
@@ -83,7 +83,7 @@ def transcribe_with_whisper(whisper_model, whisper_processor, audio_path, langua
     inputs = whisper_processor(speech_array, sampling_rate=sampling_rate, return_tensors="pt").input_features
     inputs = inputs.to(device)
     # Generate transcription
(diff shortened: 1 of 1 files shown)


scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -132,8 +134,11 @@ def extract_embedding(model, extractor, audio_path, device, sv_model_type):
         with torch.no_grad():
```
blisc: Double check if we can switch to inference_mode instead of no_grad

rfejgin: I have replaced it (for all instances of no_grad() in the file).

blisc (approved):

PR #14540, opened 2025-08-20
Harden _is_target_allowed by adding runtime class validation on top of prefix checks to prevent unsafe target resolution

KunalDhawan:
> [!IMPORTANT]  

# What does this PR do ?

Previously, `_is_target_allowed` only checked whether the `_target_` string in a config started with an allowed prefix (e.g., `torch.nn`). This was insufficient, since Python allows nested/indirect module resolution (e.g., `torch.nn.utils.rnn.torch.os.system`), which could be abused to execute arbitrary code at model load time.

In the updated implementation, we:
* Resolve the target at runtime using `hydra.utils.get_class`
* Validate the resolved object to ensure it is a class (not a function)
* Require the class to inherit from known safe bases (torch.nn.Module or nemo.core.ModelPT)
* Reject any target that fails these checks

**Collection**: core

# Changelog

- `_is_target_allowed` function in nemo/core/classes/common.py has been updated

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -82,6 +82,7 @@ ALLOWED_NEMO_SUBMODULE_PREFIXES = [
     "nemo.collections.asr.parts",
     "nemo.collections.audio.parts",
     "nemo.collections.speechlm",
+    "nemo.collections.llm",
     "nemo.lightning",
     "megatron.core",
     "tests.collections.llm.common",


nithinraok:
Some tests are failing. @chtruong814 for review as well

nemo/core/classes/common.py
```diff
@@ -76,10 +76,31 @@
...
+
+    # Resolve the target at runtime
+    try:
+        obj = hydra.utils.get_class(target)
+    except Exception:
+        return False
+
+    # Must be a class, not a function!
+    if not isinstance(obj, type):
+        return False
```
erickgalinkin: Something for usability we want to be sure of: this necessarily excludes all of `torch.nn.functional`, so we'll want to be sure there's no regression.

nemo/core/classes/common.py
```diff
@@ -76,10 +76,31 @@
...
+
+    # Import ModelPT lazily to avoid circular import issues
+    from nemo.core.classes.modelPT import ModelPT
+
+    SAFE_BASES = (torch.nn.Module, ModelPT)
+
+    # Must inherit from a known safe base!
+    if not issubclass(obj, SAFE_BASES):
+        return False
```
erickgalinkin: This seems to fix the underlying bug, but we should be very cautious with future changes since 

```python
>>> isinstance(torch.nn.parameter.torch.multiprocessing.Process, type)
True
```

Will pass the previous check but not this one. 

If it comes down to it, may simply want to explicitly exclude things from `r"torch\.nn\.\w+\.torch\.\w+"`

erickgalinkin (commented):
Appears to be failing tests but seems to alleviate the bug. Will review again when tests pass.

KunalDhawan:
Thank you @erickgalinkin  for the review. The earlier tests were failing due to pylint errors in parts of the code that weren’t modified in this PR. @chtruong814 addressed this by adding the `skip-linting` tag. At the moment, the `ubuntu-22.04-amd-py3.10-nemo-install` test is failing on GitHub login, which is also unrelated to this PR. Charlie is looking into the issue, and I’ll update this thread once we have a resolution.

erickgalinkin:
> Thank you @erickgalinkin for the review. The earlier tests were failing due to pylint errors in parts of the code that weren’t modified in this PR. @chtruong814 addressed this by adding the `skip-linting` tag. At the moment, the `ubuntu-22.04-amd-py3.10-nemo-install` test is failing on GitHub login, which is also unrelated to this PR. Charlie is looking into the issue, and I’ll update this thread once we have a resolution.

No sweat! Just want to make sure that if any of these files change, that I'm tagged for review. 

Are we confident that `torch.nn.functional` is not meant/going to be used?

nemo/core/classes/common.py
KunalDhawan: Good point, thanks for catching that!

You're right that relying solely on `issubclass(obj, SAFE_BASES)` could, in theory, still allow unexpected nested references such as `torch.nn.parameter.torch.multiprocessing.Process`. That case would pass the old startswith("torch.nn.") check but isn’t truly safe to instantiate.

The current fix intentionally tightens the logic around `ModelPT` and `torch.nn.Module` inheritance, but I agree it’s worth explicitly excluding pathological nested module patterns as a safeguard.

I’ll plan to follow up with a small hardening patch (e.g., rejecting targets matching r"torch\.nn\..*\.torch\.") or a stricter check on the module path to prevent recursive submodule traversal under unrelated namespaces.

For now, this PR ensures the safe baseline and resolves the remote code execution vector, but I fully agree we should add the explicit exclusion in a future iteration.

KunalDhawan: Good point, thanks for catching that!

You’re right that relying solely on `issubclass(obj, SAFE_BASES)` could, in theory, still allow unexpected nested references such as `torch.nn.parameter.torch.multiprocessing.Process`. That case would have passed the old `startswith("torch.nn.")` check but isn’t truly safe to instantiate.

The current fix intentionally tightens the logic to only allow subclasses of `ModelPT` or `torch.nn.Module`, which covers all valid NeMo and PyTorch model targets and blocks arbitrary imports. This ensures a safe baseline and resolves the immediate security concern.

That said, I agree it’s worth adding an explicit exclusion for pathological nested module patterns (e.g., rejecting targets matching `r"torch\.nn\..*\.torch\."`) or adding a stricter module path validation in a follow-up patch. I’ll plan to follow up with that hardening once we validate that it doesn’t block any legitimate NeMo use cases.

For now, this PR ensures the safe baseline and resolves the remote code execution vector, but I fully agree we should add the explicit exclusion in a future iteration.

chtruong814:
@nithinraok it looks good to me.  Any concerns from you with merging?

chtruong814:
@erickgalinkin could you do a final review of this before we merge it?  Tests were passing previously but got retriggered when a label got changed.

chtruong814:
Previously passing CI pipeline
https://github.com/NVIDIA-NeMo/NeMo/actions/runs/18263670632

PR #14555, opened 2025-08-21
Forbid generation of special tokens (excpet EOS)

rfejgin:
We have observed the model sometimes generating special tokens (e.g. a reserved token, audio_context_bos, etc.), which should never happen and could cause artifacts. This was observed in particular when CFG is on.

Here, we explicitly ensure that such tokens are never sampled by forcing the corresponding logits (post-CFG, pre top-k) to `-inf`. `AUDIO_EOS` is permitted since we do need sample it (and generally trim it away before returning the codes to user/codec).

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -693,14 +693,14 @@ class MagpieTTSModel(ModelPT):
             output_str += c
         logging.debug(output_str)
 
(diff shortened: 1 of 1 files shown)


rfejgin:
I am running some evals to validate the changes are safe but please feel free to review. I'll change to non-draft status when ready.

rfejgin:
Evals look good, I think we can merge.

PR #14569, opened 2025-08-25
End detection updates and Text context remapping during training

paarthneekhara:
This PR introduces two main updates.

1. EOS and Finished/unfinished sentences:
* Added an option to disable finished/unfinished sentence tracking when prior is applied. Sometimes I notice forcing or disallowing EOS prediction introduces artifacts towards the end, unless we really finetune the unifinished and finished sentences constants. I am seeing we don't really need to handhold whether the sentence is finshed or not and can trust the model's decision, even when we apply the prior. So I have added an argument to not force or disallow EOS prediction by setting ignore_finished_sentence_tracking=True.

* EOS detection - Right now, the logic was that if any codebook in multinomial or argmax sampling has an EOS token predicted, we predict the end. VERY RARELY, I notice this leads to predicting EOS abruptly. At some point, our logic was that if argmax of codebook 0 is EOS, we predict the end. I have added a few options - to predict EOS is any, all or zeroth codebook is EOS. And whether to only look at argmax sampling, or either of argmax or multinomial sampling. **Keeping this customizable, because I suspect that parallel prediction EOS logic might be different from MaskGIT EOS prediction logic.**

2. Added an option to map text contexts to a common text context during training (with a probability). We can give a json containing a dictionary mapping stating which text contexts should be mapped to some other text context. For example Lindy_calm, Lindy_angry mapped to Lindy_all. Then during training, we will do this mapping with a supplied probability. Main objective here is, for a speaker, we can learn all emotions and expect the model to infer the emotion from the transcript during inference.

### DIFF
--- a/examples/tts/conf/magpietts/magpietts_inference_multilingual_v1.yaml
+++ b/examples/tts/conf/magpietts/magpietts_inference_multilingual_v1.yaml
@@ -124,9 +124,6 @@ model:
         tone_prefix: "#"
         ascii_letter_prefix: ""
         ascii_letter_case: "upper"
-    multilingual_sentencepiece:
-      _target_: AutoTokenizer
-      pretrained_model: "bert-base-multilingual-uncased"
 
   test_ds:
     dataset:
--- a/examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
+++ b/examples/tts/conf/magpietts/magpietts_multilingual_v1.yaml
@@ -117,9 +117,6 @@ model:
         tone_prefix: "#"
         ascii_letter_prefix: ""
         ascii_letter_case: "upper"
-    multilingual_sentencepiece:
-      _target_: AutoTokenizer
-      pretrained_model: "bert-base-multilingual-uncased"
 
   train_ds:
     dataset:
--- a/nemo/collections/common/data/lhotse/sampling.py
+++ b/nemo/collections/common/data/lhotse/sampling.py
@@ -18,7 +18,7 @@ from dataclasses import dataclass
 from typing import Any, Sequence
 
 import numpy as np
-from lhotse.cut import Cut
+from lhotse.cut import Cut, MonoCut
 from lhotse.dataset import SamplingConstraint, TokenConstraint
 from lhotse.dataset.sampling.dynamic_bucketing import FixedBucketBatchSizeConstraint
 from lhotse.utils import ifnone
@@ -279,7 +279,7 @@ class ValidationStatusFilter:
 
     def __call__(self, example) -> bool:
         if (
-            isinstance(example, Cut)
+            isinstance(example, MonoCut)
             and example.has_custom("validation_status")
             and example.validation_status != self.keep
         ):
@@ -298,7 +298,11 @@ class CERFilter:
         self.max_cer = ifnone(max_cer, float("inf"))
 
     def __call__(self, example) -> bool:
-        if isinstance(example, Cut) and len(example.supervisions) > 0 and example.supervisions[0].has_custom("cer"):
+        if (
+            isinstance(example, MonoCut)
+            and len(example.supervisions) > 0
+            and example.supervisions[0].has_custom("cer")
+        ):
             return example.supervisions[0].cer <= self.max_cer
         else:
             return True
@@ -315,7 +319,7 @@ class ContextSpeakerSimilarityFilter:
 
     def __call__(self, example) -> bool:
         if (
-            isinstance(example, Cut)
+            isinstance(example, MonoCut)
             and len(example.supervisions) > 0
             and example.supervisions[0].has_custom("context_speaker_similarity")
         ):
--- a/nemo/collections/tts/models/magpietts_preference_optimization.py
+++ b/nemo/collections/tts/models/magpietts_preference_optimization.py
@@ -16,6 +16,7 @@ import json
 import os
 import random
 import string
+from typing import Optional
 
 import librosa
 import numpy as np
@@ -36,6 +37,8 @@ try:
     HAVE_TORCHAUDIO = True
 except ImportError:
     HAVE_TORCHAUDIO = False
+
+from nemo_text_processing.text_normalization.normalize import Normalizer
 
 from nemo.collections.tts.models import MagpieTTSModel
 
@@ -65,6 +68,20 @@ class MagpieTTSModelOfflinePODataGen(MagpieTTSModel):
             self.whisper_processor = WhisperProcessor.from_pretrained("openai/whisper-large-v3")
             self.whisper_model = WhisperForConditionalGeneration.from_pretrained("openai/whisper-large-v3")
             self.whisper_model.eval()
+            self._normalize_whisper_transcript = cfg.get('normalize_whisper_transcript', True)
+            if self._normalize_whisper_transcript:
+                self._normalizer_cache = {}
+                # Pre-create normalizer for the configured language
+                lang = cfg.get('pref_set_language', 'en')
+                self._get_cached_normalizer(lang)
+
+    def _get_cached_normalizer(self, lang_key):
+        """Get or create a cached normalizer for the given language."""
+        lang_key = lang_key if lang_key else "en"
+        if lang_key not in self._normalizer_cache:
+            logging.info(f"Creating normalizer for language: {lang_key}")
+            self._normalizer_cache[lang_key] = Normalizer(input_case="cased", lang=lang_key)
+        return self._normalizer_cache[lang_key]
 
     def test_step(self, batch, batch_idx):
         with torch.no_grad():
@@ -118,12 +135,18 @@ class MagpieTTSModelOfflinePODataGen(MagpieTTSModel):
                             else:
                                 pred_transcripts = []
                                 for audio_path in predicted_audio_paths:
+                                    normalizer = (
+                                        self._get_cached_normalizer(self.cfg.pref_set_language)
+                                        if self._normalize_whisper_transcript
+                                        else None
+                                    )
                                     transcript = transcribe_with_whisper(
                                         audio_path,
                                         self.cfg.pref_set_language,
                                         self.whisper_processor,
                                         self.whisper_model,
                                         self.device,
+                                        normalizer,
                                     )
                                     pred_transcripts.append(transcript)
 
@@ -136,7 +159,7 @@ class MagpieTTSModelOfflinePODataGen(MagpieTTSModel):
                             ).any(), f"Expected short audio file to be the only cause of ASR errors, but got error with lengths {predicted_audio_lens}"
                             logging.warning(f"Exception during ASR transcription: {e}")
                             logging.warning(
-                                f"Skipping processing of the batch; generating metrics indicating a WER of 100% and Speaker Similarity of 0.0"
+                                "Skipping processing of the batch; generating metrics indicating a WER of 100% and Speaker Similarity of 0.0"
                             )
                             batch_invalid = True
                             continue  # don't break since we want to continue building audio durations list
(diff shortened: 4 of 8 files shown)


rfejgin:
Regarding EOS detection: that's an interesting that detecting from all codebooks can lead to early termination -- good find!

About the option detecting from codebook 0 only: there is the issue of what happens if an EOS appears only in a codebook other than 0 and we ignore it (for EOS detection purposes). Then we'd end up [replacing the token](https://github.com/NVIDIA/NeMo/blob/magpietts_2508/nemo/collections/tts/models/magpietts.py#L463) with token ID 0 before decoding with the codec, but we know 0 doesn't necessarily correspond to codec silence.

But maybe we need to consider these two mechanisms separately:

1. **EOS detection:** use whatever method works best (e.g. from the set you implemented here)
2. **"cleaning" of tokens before returning to user/codec**: replace EOS tokens with "silent" tokens. We'd need to know, per codec release, which tokens it maps to silence. The complication is that we may have a frame of mixed silence and non-silence tokens; but it's probably still better to use a "quiet" token for the codebook that contained the EOS than arbitrarily choose 0. Anyway, for this mechanism to work we need to query the codec at init for silence tokens (e.g. by encoding a bit of digital silence), or keep track of that per codec release. I prefer the latter so that we don't need to include a codec encoder in our releases. Maybe we should require a `get_silence_tokens()` to be added to the the codec API. This would only have to be computed once per codec release and should be pretty easy for the codec maintainer to do.

If (2) works better that'll give (1) more flexibility.

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -36,6 +36,48 @@ class LocalTransformerType(PrettyStrEnum):
...
+            return "argmax"
+        elif detection_method in [
+            EOSDetectionMethod.ARGMAX_OR_MULTINOMIAL_ANY,
+            EOSDetectionMethod.ARGMAX_OR_MULTINOMIAL_ALL,
+            EOSDetectionMethod.ARGMAX_OR_MULTINOMIAL_ZERO_CB,
+        ]:
+            return "argmax_or_multinomial"
+        else:
+            raise ValueError(f"Invalid EOS detection method: {detection_method}")
+
```
rfejgin, line 79: Up to you but you could consider separating the config of EOS detection type (from the set: all codebooks, 0, any) from the sampling type (multinomial or argmax) and then we wouldn't need to define all combinations of the two and also extract the parameters from the combination enum.

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -578,9 +582,12 @@ def __getitem__(self, index):
...
         if self.use_text_conditioning_tokenizer:
             if 'context_text' in data.manifest_entry:
-                context_tokens = self.text_tokenizer.encode(
-                    data.manifest_entry['context_text'], self.text_conditioning_tokenizer_name
-                )
+                context_text = data.manifest_entry['context_text']
+                if self.text_context_remapping is not None and context_text in self.text_context_remapping:
+                    if self.dataset_type == 'train' and random.random() < self.text_context_remapping_prob:
+                        # Only remap during training. Give the exact text context during inference.
+                        context_text = self.text_context_remapping[context_text]
```
rfejgin, line 591: it looks like the remapping dictionary must cover all possibly text contexts in the dataset, or the lookup in line 589 will error out. Alternatively we could check if a mapping exists for the given text context and if not, default to no remapping. But I guess then the risk is that someone might have a typo in the remapping dictionary and never know about it. Not sure if there's a use case where within one dataset we want to not remap some contexts. If not then this is good as is.

rfejgin:
Cool addition of the text remapping logic, by the way!

nemo/collections/tts/data/text_to_speech_dataset.py
paarthneekhara: It is guarded under the first if statement (line 581) "if self.text_context_remapping is not None and **context_text in self.text_context_remapping** " so it should be fine if text_context_remapping doesn't cover all text contexts. You are right, we don't want to include all text contexts in the remapping dict cause we may miss out some when we add new datasets.

nemo/collections/tts/modules/magpietts_modules.py
paarthneekhara: Yes, I thought about that. That would require two EOS related arguments which I was trying to avoid.

paarthneekhara:
> Regarding EOS detection: that's an interesting that detecting from all codebooks can lead to early termination -- good find!
> 
> About the option detecting from codebook 0 only: there is the issue of what happens if an EOS appears only in a codebook other than 0 and we ignore it (for EOS detection purposes). Then we'd end up [replacing the token](https://github.com/NVIDIA/NeMo/blob/magpietts_2508/nemo/collections/tts/models/magpietts.py#L463) with token ID 0 before decoding with the codec, but we know 0 doesn't necessarily correspond to codec silence.
> 
> But maybe we need to consider these two mechanisms separately:
> 
> 1. **EOS detection:** use whatever method works best (e.g. from the set you implemented here)
> 2. **"cleaning" of tokens before returning to user/codec**: replace EOS tokens with "silent" tokens. We'd need to know, per codec release, which tokens it maps to silence. The complication is that we may have a frame of mixed silence and non-silence tokens; but it's probably still better to use a "quiet" token for the codebook that contained the EOS than arbitrarily choose 0. Anyway, for this mechanism to work we need to query the codec at init for silence tokens (e.g. by encoding a bit of digital silence), or keep track of that per codec release. I prefer the latter so that we don't need to include a codec encoder in our releases. Maybe we should require a `get_silence_tokens()` to be added to the the codec API. This would only have to be computed once per codec release and should be pretty easy for the codec maintainer to do.
> 
> If (2) works better that'll give (1) more flexibility.

That's a good point. We should figure out a better way to clean up the codes. Are you suggesting if we want to clamp a partucular codebook to be within the range, we should make the whole frame silent or just that codebook.

nemo/collections/tts/modules/magpietts_modules.py
rfejgin: sounds good

nemo/collections/tts/data/text_to_speech_dataset.py
rfejgin: Oh you are right, I missed that.

rfejgin:
> 
> That's a good point. We should figure out a better way to clean up the codes. Are you suggesting if we want to clamp a particular codebook to be within the range, we should make the whole frame silent or just that codebook.

I think it probably requires experimentation, e.g. by listening to find out what sounds better of the options we're considering:
1. Replacing the particular codebook with token 0 (what we have now)
2. Replace all codebooks with their corresponding silence tokens
3. Replace only the particular codebook with its silence token

One way to choose would be to load some codes of a real speech signal and randomly choose a few positions (frame+codebooks combinations) that we will corrupt with the 3 methods above; listen to all 3 and choose the one that sounds best.

rfejgin:
> > That's a good point. We should figure out a better way to clean up the codes. Are you suggesting if we want to clamp a particular codebook to be within the range, we should make the whole frame silent or just that codebook.
> 
> I think it probably requires experimentation, e.g. by listening to find out what sounds better of the options we're considering:
> 
> 1. Replacing the particular codebook with token 0 (what we have now)
> 2. Replace all codebooks with their corresponding silence tokens
> 3. Replace only the particular codebook with its silence tokens
> 
> One way to choose would be to load some codes of a real speech signal and randomly choose a few positions (frame+codebooks combinations) that we will corrupt with the 3 methods above; listen to all 3 and choose the one that sounds best.

An alternative that came to mind: if we detected an EOS which we want to ignore, e.g. because it's not in codebook 0, and we need to replace with something: **resample**. It wouldn't require another forward pass and should give something plausible (possibly better than silence). But we'd have to make sure not to get stuck in an infinite resampling loop (could limit how many times we reasample, then accept EOS if it insists...).

Edit: On second thought, the resampling option only works for parallel prediction. If we have a local transformer we'd have to do something else, like either run the LT again (which is higher complexity than resampling), replace with silence, or maybe just resample the particular codebook from the *parallel* head.

nemo/collections/tts/models/magpietts.py
```diff
@@ -1997,7 +2013,7 @@ def get_inference_attention_plots(
 
         return cross_attention_maps, headwise_cross_attention_maps
 
-    def find_eos_frame_index(self, codes) -> Optional[int]:
+    def find_eos_frame_index(self, codes, eos_detection_method) -> Optional[int]:
```
blisc: Since you changed the default return of this function, can you update the typehint and docstring to more accurately match the updated function?

nemo/collections/tts/models/magpietts.py
```diff
@@ -2007,12 +2023,37 @@ def find_eos_frame_index(self, codes) -> Optional[int]:
...
+        else:
+            raise ValueError(f"Invalid EOS detection method: {eos_detection_method}")
         # find first frame with EOS
         if eos_per_frame.any():
             # return index of the first frame with EOS
             return eos_per_frame.nonzero()[0].item()
-        return None
+        return float('inf')
+
+    def detect_eos(self, audio_codes_multinomial, audio_codes_argmax, eos_detection_method):
```
blisc: Can you add typehints and a docstring to this function?

blisc (changes requested):
Some minor style changes

nemo/collections/tts/models/magpietts.py
```diff
@@ -2037,7 +2078,11 @@ def infer_batch(
         maskgit_fixed_schedule=None,
         maskgit_dynamic_cfg_scale=False,
         maskgit_sampling_type=None,
+        ignore_finished_sentence_tracking=False,
+        eos_detection_method="argmax_or_multinomial_any",
     ):
+
```
shehzeen: We should add a docstring to infer_batch since there any many arguments in this function now.

shehzeen:
We should update the infer_batch calls in magpietts_preference_optimization.py to use the right arguments for end detection logic during GRPO/DPO data generation.

XuesongYang:
> Added an option to map text contexts to a common text context during training (with a probability). We can give a json containing a dictionary mapping stating which text contexts should be mapped to some other text context. For example Lindy_calm, Lindy_angry mapped to Lindy_all. Then during training, we will do this mapping with a supplied probability. Main objective here is, for a speaker, we can learn all emotions and expect the model to infer the emotion from the transcript during inference.

I agree with the needs of making the model learn the mapping `{speaker}_{emotions}` -> `{speaker}` for text context records. But there seems a simpler solution to achieve this, i.e. just reusing the `speaker` key/val pair without explicitly loading mapping json file.

FYI, our existing lhotse manifest has below key/val pairs. There exist overlaps between `speaker` and `custom.context_text` fields already. IIRC, `speaker` field is not used for modeling so far, but I was still enforcing `speaker` in this format on purposes with the hope to replace `context_text` in the future. 

The question is why don't we add the emotion tags to `speaker` directly, such as `speaker='| Language:en Dataset:rivaEmmaMeganSeanTom Speaker:Emma Emotion:Calm_Medium_Uncorrelated |'`,  and consider `speaker` as `context_text` instead of duplicating `context_text` with extra mapping rules? 

The `context_text` field again is still useful and can be preserved for natural language description in the future.
```
speaker='| Language:en Dataset:rivaEmmaMeganSeanTom Speaker:Emma |',
...
custom={
    'emotion': 'Calm_Medium_Uncorrelated',
    'context_text': 'Speaker and Emotion: | Language:en Dataset:rivaEmmaMeganSeanTom 
Speaker:Emma_Calm_Medium_Uncorrelated |'
},
...
```

nemo/collections/tts/models/magpietts.py
```diff
@@ -176,6 +178,20 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+        # we can create an override dict {'Emma_neutral' : 'Emma', 'Emma_conversational' : 'Emma'}
+        # This dict is saved in a json file given by cfg.model.text_context_remapping_json
+        # If we want to preserve both behaviours i.e (Emma_neutral, Emma_conversational) and just (Emma)
+        # we can do this mapping with a probability during training, as specified by text_context_remapping_prob
+        self.text_context_remapping = None
+        text_context_remapping_json = cfg.get('text_context_remapping_json', None)
+        self.text_context_remapping_prob = cfg.get('text_context_remapping_prob', 0.0)
+        if text_context_remapping_json is not None:
+            with open(text_context_remapping_json, 'r') as f:
+                self.text_context_remapping = json.load(f)
```
blisc, line 193: Why do we want to pass it in as a json file rather than saving the dictionary inside the yaml itself? This seems like an anti-design pattern.

paarthneekhara: We might want nested yamls in that case. Inside the same yaml (or in the training command), it can become very huge, Eg. I had 60 entries.

paarthneekhara:
@XuesongYang regarding option to map all contexts of a speaker to the speaker name --- Sometimes we want to map things a little differently. Eg. mapping Emma_calm_correlated and Emma_neutral to Emma (but not Emma_calm_uncorrelated), which is why I think it should be more flexible.

nemo/collections/tts/models/magpietts.py
blisc: This complexity has to exist somewhere. Creating another abstraction layer is not the best solution for this. 
In terms of reducing complexity: maybe we should allow for wildcard remapping rather than fixed string to string mapping

PR #14765, opened 2025-09-19
Streaming decoding policies (Wait-K and AlignAtt) for Canary model

andrusenkoau:
> [!IMPORTANT]  

# What does this PR do ?

Add streaming decoding policies support (Wait-K and AlignAtt) for Canary model.

**Wait-K** policy predicts only one token per each new speech chunk that increases the overall latency - [original paper](https://arxiv.org/abs/1810.08398). In this case, it is unclear at what point you can forget part of the left context when recognizing with a limited window. It is recommended to set the left context to the maximum possible value (infinite left context) for the waitk policy.

**AlignAtt** policy predicts tokens according to the cross-attention condition - [original paper](https://arxiv.org/pdf/2305.11408). If the condition is met, then the audio size does not need to be increased, and the prediction of the next token continues. Otherwise, the audio buffer size needs to be increased. This policy shows lower latency in comparison with waitk. This policy is also suitable for window recognition (left context is fixed and not infinite), but you can lose accuracy in this case.

**Collection**: [ASR]

# Changelog

# Usage

Decoding policy is controlled by AEDStreamingDecodingConfig. You can choose waitk or alignatt policy. The remaining parameters (such as alignatt_thr or waitk_lagging) need to be selected depending on the data and the task (for example, for AST, you can increase waitk_lagging, which works for both policies).

```python
python3 ${NEMO_DIR}/examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py \
    pretrained_name=${ASR_MODEL_NAME} \
    dataset_manifest=${DECODING_MANIFEST} \
    output_filename=${OUTPUT_MANIFEST} \
    left_context_secs=10 \
    chunk_secs=1 \
    right_context_secs=0.5 \
    batch_size=32\
    decoding.streaming_policy=waitk \ # [waitk or alignatt]
    decoding.alignatt_thr=8 \
    decoding.waitk_lagging=2 \
    decoding.exclude_sink_frames=8 \
    decoding.xatt_scores_layer=-2 \
    decoding.hallucinations_detector=True
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,501 @@
...
+
+Recommended settings for audio buffer:
+- streaming with 1.5s latency: 10-1-0.5 (10s left, 1s chunk, 0.5s right)
+
+Example usage:
+
+```shell
+# TODO: update this example
+python speech_to_text_aed_streaming_infer.py \
+    pretrained_name=nvidia/canary-180m-flash.nemo \
```
nithinraok: update to latest v2 model. 

Any becnhmarks numbers to share?

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,501 @@
...
+
+    # Set `cuda` to int to define CUDA device. If 'None', will look for CUDA
+    # device anyway, and do inference on CPU only if CUDA device is not found.
+    # If `cuda` is a negative number, inference will be on CPU only.
+    cuda: Optional[int] = None
+    allow_mps: bool = True  # allow to select MPS device (Apple Silicon M-series GPU)
+    compute_dtype: Optional[str] = (
+        None  # "float32", "bfloat16" or "float16"; if None (default): bfloat16 if available, else float32
+    )
+    matmul_precision: str = "high"  # Literal["highest", "high", "medium"]
```
nithinraok, line 129: isn;t medium better for higher accuracy?

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,501 @@
...
+
+    # Recompute model transcription, even if the output folder exists with scores.
+    overwrite_transcripts: bool = True
+
+    # Decoding strategy for RNNT models
+    decoding: AEDStreamingDecodingConfig = field(default_factory=AEDStreamingDecodingConfig)
+
+    # Config for word / character error rate calculation
+    calculate_wer: bool = True  # for ASR task
+    calculate_bleu: bool = True  # for AST task
```
nithinraok: lets keep bleu calculation `False` by default. 

Are we only calculating bleu score for samples in manifest based on prompt selection?

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+from nemo.collections.asr.models.aed_multitask_models import EncDecMultiTaskModel, lens_to_mask, parse_multitask_prompt
+from nemo.collections.asr.parts.submodules.multitask_decoding import AEDStreamingDecodingConfig
+from nemo.collections.asr.parts.utils.streaming_utils import ContextSize
+from nemo.utils import logging
+
+
+@dataclass
+class AEDStreamingState:
+    decoder_input_ids: torch.Tensor = None  # tokens ids of initial AED model prompt
+    tgt: torch.Tensor = None  # buffer with deocoded tokens ids
```
nithinraok: better naming for `tgt` pls.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+    decoding_step: int = -1  # current decoding step
+    decoder_mems_list: list = None  # decoder caches, helps to reduce the memory usage
+    is_last_chunk_batch: torch.Tensor = False  # whether the current chunk is the last speech chunk in the audio
+    max_generation_length: int = (
```
nithinraok, line 33: is this max length considered for each chunk while decoding or for full audio? Looks like this is for full audio and we observed for v2, 128 would be better by default.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+            return self.state
+
+        # wait-k streaming decoding policy
+        elif self.decoding_cfg.streaming_policy == "waitk":
+            self.run_waitk_decoding_step(encoded_speech, encoder_input_mask)
+        # alignatt streaming decoding policy
+        elif self.decoding_cfg.streaming_policy == "alignatt":
+            self.run_alignatt_decoding_step(encoded_speech, encoder_input_mask)
+        else:
+            raise ValueError("Canary streaming decoding supports only alignatt or waitk decodong policy")
```
nithinraok: why to decide the policy at each call and not during initialization of the class? Isn;t that better?

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+    ) -> AEDStreamingState:
+
+        self.state = prev_batched_state
+        self.state.encoder_output_len = encoder_output_len
+
+        # prepare encoder embeddings for the decoding
+        # enc_states = encoder_output.permute(0, 2, 1)
+        encoded_speech = self.asr_model.encoder_decoder_proj(encoder_output)
+
+        encoder_input_mask = lens_to_mask(encoder_output_len, encoded_speech.shape[1]).to(encoded_speech.dtype)
```
nithinraok: Couple of things:
1. lets move this out into `examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py` where rest of the encoder is handled to keep all encoder processing at same place
2. why not use asr_model.forward() with `use_transf_encoder` and `transcript` as none. or maybe modify `aed_multitask_models.py` code accordingly to introduce `forward_encoder` and clean all preprocessing steps in `speech_to_text_aed_streaming_infer.py`

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,501 @@
...
+            for i in range(batch_size):
+                transcription_idx = model_state.tgt[
+                    i, model_state.decoder_input_ids.size(-1) : model_state.current_context_lengths[i]
+                ]
+                transcription = asr_model.tokenizer.ids_to_text(transcription_idx.tolist()).strip()
+                logging.info(f"i: {transcription}")
+                all_hyps.append(transcription)
+                tokens_frame_alignment.append(model_state.tokens_frame_alignment[i])
+                predicted_token_ids.append(
+                    model_state.tgt[i, model_state.decoder_input_ids.size(-1) : model_state.current_context_lengths[i]]
```
nithinraok: looks like we do no special handling for PnC across chunk boundaries. Any effect you noticed for canary-v2 or previous models with PnC?

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+            if self.state.max_tokens_per_alignatt_step is not None:
+                self.state.steps_per_inner_loop += self.state.active_samples_inner_loop
+                disable_samples_mask = self.state.steps_per_inner_loop >= self.state.max_tokens_per_alignatt_step
+                disable_samples_mask *= torch.logical_not(self.state.is_last_chunk_batch)
+                self.state.active_samples_inner_loop *= torch.logical_not(disable_samples_mask)
+
+            if not torch.any(self.state.active_samples_inner_loop):
+                break
+
+    def detect_hallucinations(self, tgt, batch_idxs, current_context_lengths):
```
nithinraok: what is the length of the prompt tokens for v1 and v2? Looks like this function is expecting atleast 8 tokens for `current_context_lengths` else this might break.

tests/collections/asr/decoding/test_multi_task_streaming_decoding.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    with torch.no_grad(), torch.inference_mode():
+        for i in range(0, len(manifest), batch_size):
+            encoder_output, encoder_output_len = get_batch_encoder_outputs_from_records(
+                manifest[i : i + batch_size], model=model, device=device
+            )
+            local_batch_size = encoder_output_len.shape[0]
+            decoder_input_ids = return_decoder_input_ids(streaming_decoding_cfg, model)
+            model_state = initialize_aed_model_state(
```
nithinraok: State is initialized here and also in init of `decoding_computer` we can remove it in init of `GreedyBatchedStreamingAEDComputer`

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+    def __init__(
+        self,
+        asr_model: EncDecMultiTaskModel,
+        frame_chunk_size: int,
+        decoding_cfg: AEDStreamingDecodingConfig,
+    ):
+        """
+        Init method.
+        Args:
+            asr_model: isntace of ASR model (Canary)
```
nithinraok:
```suggestion
            asr_model: instance of ASR model (Canary)
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+        self.state = prev_batched_state
+        self.state.encoder_output_len = encoder_output_len
+
+        # prepare encoder embeddings for the decoding
+        # enc_states = encoder_output.permute(0, 2, 1)
+        encoded_speech = self.asr_model.encoder_decoder_proj(encoder_output)
+
+        encoder_input_mask = lens_to_mask(encoder_output_len, encoded_speech.shape[1]).to(encoded_speech.dtype)
+
+        # initiall waitk lagging. Applicable for Wait-k and AlignAtt decoding policies. Control the start of the decoding process.
```
nithinraok:
```suggestion
        # initial waitk lagging. Applicable for Wait-k and AlignAtt decoding policies. Control the start of the decoding process.
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
nithinraok:
```suggestion
            raise ValueError("Canary streaming decoding supports only alignatt or waitk decoding policy")
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+        elif self.decoding_cfg.streaming_policy == "alignatt":
+            self.run_alignatt_decoding_step(encoded_speech, encoder_input_mask)
+        else:
+            raise ValueError("Canary streaming decoding supports only alignatt or waitk decodong policy")
+
+        return self.state
+
+    def run_waitk_decoding_step(self, encoded_speech, encoder_input_mask):
+        """
+        Run a decoding step for waith streaming policy.
```
nithinraok:
```suggestion
        Run a decoding step for waitk streaming policy.
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+
+        # define start and max generation lengths
+        start_from = self.state.decoding_step + 1
+        if torch.any(torch.logical_not(self.state.is_last_chunk_batch)):
+            # predict only one token per speech chunk if not the last one
+            max_generation_length = start_from + 1
+        else:
+            max_generation_length = self.decoding_cfg.max_generation_length
+
+        # inner deocding loop (with same speech chunk)
```
nithinraok:
```suggestion
        # inner decoding loop (with same speech chunk)
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+                disable_samples_mask *= torch.logical_not(self.state.is_last_chunk_batch)
+                self.state.active_samples_inner_loop *= torch.logical_not(disable_samples_mask)
+
+            if not torch.any(self.state.active_samples_inner_loop):
+                break
+
+    def detect_hallucinations(self, tgt, batch_idxs, current_context_lengths):
+
+        ccl = current_context_lengths
+        # pattern 1: four consequtive tokens are the same: "a a a a"
```
nithinraok:
```suggestion
        # pattern 1: four consecutive tokens are the same: "a a a a"
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+                self.state.active_samples *= torch.logical_not(samples_with_max_context_length)
+                self.state.active_samples_inner_loop *= torch.logical_not(samples_with_max_context_length)
+
+            # zero out decoder_mems_list for non active samples
+            if torch.any(torch.logical_not(self.state.active_samples_inner_loop)):
+                for j in range(len(decoder_mems_list)):
+                    decoder_mems_list[j][:, -1] *= self.state.active_samples_inner_loop.unsqueeze(-1)
+            self.state.decoder_mems_list = decoder_mems_list
+
+    def run_alignatt_decoding_step(self, encoded_speech, encoder_input_mask):
```
nithinraok, line 196: lot of duplicate code in both policies, can we merge them?

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,553 @@
...
+    )
+    model_state.tgt[:, : model_state.decoder_input_ids.size(-1)] = model_state.decoder_input_ids
+    model_state.tokens_frame_alignment = torch.zeros_like(model_state.tgt)
+    model_state.active_samples = torch.ones(batch_size, dtype=torch.bool, device=asr_model.device)
+    model_state.active_samples_inner_loop = torch.ones(batch_size, dtype=torch.bool, device=asr_model.device)
+    model_state.right_context = context_encoder_frames.right
+    model_state.eos_tokens = torch.full(
+        [batch_size], asr_model.tokenizer.eos, dtype=torch.long, device=asr_model.device
+    )
+    model_state.avgpool2d = torch.nn.AvgPool2d(5, stride=1, padding=2, count_include_pad=False)
```
nithinraok: why these parameters, like kernel_size of 5?

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,501 @@
...
+                context_samples=context_samples,
+                dtype=audio_batch.dtype,
+                device=device,
+            )
+            rest_audio_lengths = audio_batch_lengths.clone()
+            is_last_window_batch = torch.zeros_like(rest_audio_lengths)
+
+            step_idx = 0
+            # iterate over audio samples
+            while left_sample < audio_batch.shape[1]:
```
nithinraok, line 347: If audio_batch.shape[1] == 0, loop never executes, but we report results nevertheless .

nithinraok (changes requested):
Add documentation and also docstrings missing for few classes

nithinraok:
@artbataev I remember you are working on gradio like demo for rnnt,ctc streaming. Can we add this aed streaming to that example as well.

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: I have not noticed any difference between "high" and "medium" precision in terms of accuracy and decoding time in my case.

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: Ok, I turned the blue calculation off by default. 

What do you mean by "calculating bleu based on prompt selection"?
I do it in the following way:

```
if cfg.calculate_bleu:
    output_manifest_w_bleu, total_res, _ = cal_write_text_metric(
        pred_manifest=output_filename,
        pred_text_attr_name="pred_text",
        gt_text_attr_name="answer",
        metric="bleu",
    )
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: replaced with `pred_tokens_ids` name

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: max_generation_length applies to the whole audio here. I checked 128 value, this led to accuracy degradation for test_other data. 256 looks like a tradeoff here.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: Good point. I moved it to initialization:

```
        ......
        super().__init__()

        self.asr_model = asr_model
        self.frame_chunk_size = frame_chunk_size
        self.decoding_cfg = decoding_cfg
        self.state = AEDStreamingState()

        # define the decoding method once during initialization
        if decoding_cfg.streaming_policy == "waitk":
            self._run_decoding_step = self.run_waitk_decoding_step
        elif decoding_cfg.streaming_policy == "alignatt":
            self._run_decoding_step = self.run_alignatt_decoding_step
        else:
            raise ValueError("Canary streaming decoding supports only alignatt or waitk decoding policy")

```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: I moved all the encoder part to `speech_to_text_aed_streaming_infer.py` with using `model.forward()`.

```
# get encoder output using full buffer [left-chunk-right]
_, encoded_len, enc_states, _ = asr_model(input_signal=buffer.samples, input_signal_length=buffer.context_size_batch.total())
# remove right context from encoder output length (only for non-last chunks)
encoder_context_batch = buffer.context_size_batch.subsample(factor=encoder_frame2audio_samples)
encoded_len_no_rc = encoder_context_batch.left + encoder_context_batch.chunk
encoded_length_corrected = torch.where(is_last_chunk_batch, encoded_len, encoded_len_no_rc)
encoder_input_mask = lens_to_mask(encoded_length_corrected, enc_states.shape[1]).to(enc_states.dtype)
```

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: Could you clarify the question?
This part of the code converts predicted tokens from ids to text with no difference between PnC and non-PnC results.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: I added an additional check to avoid this problem:

```
def detect_hallucinations(self, pred_tokens_ids, batch_idxs, current_context_lengths):

    # we need to have at least 8 tokens to run hallucinations detector
    if self.state.decoding_step < 8:
        return torch.zeros(pred_tokens_ids.shape[0], dtype=torch.bool, device=pred_tokens_ids.device)
```

tests/collections/asr/decoding/test_multi_task_streaming_decoding.py
andrusenkoau: Could you clarify this point? 
`initialize_aed_model_state` is a separate function in `nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py` script. It is not a part of `GreedyBatchedStreamingAEDComputer` class.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: I would leave them separate. The code for these functions is duplicated only at the beginning (about 30-40 lines). The rest is different. Merging them could make the code much more difficult to understand and maintain. Furthermore, waitk may be removed in the future, as it shows excessively high latency.

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: `use_avgpool_for_alignatt` is an experimental feature and is disabled by default. I left it as is for the following experiments, in case we decide to proceed.

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: What would you recommend to do in the case of empty audio batch? Just return empty transcripts?

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: I changed the model name to v2. However, the v2 model showed a bigger accuracy degradation in the case of streaming decoding. V1 model works much better for small audio chunks.

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
nithinraok: yes, skip calculating metrics

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,499 @@
...
+        dataset=audio_dataset,
+        batch_size=cfg.batch_size,
+        shuffle=False,
+        num_workers=cfg.num_workers,
+        collate_fn=AudioBatch.collate_fn,
+        drop_last=False,
+        in_order=True,
+    )
+
+    decoding_computer = GreedyBatchedStreamingAEDComputer(
```
nithinraok, line 299: Re: https://github.com/NVIDIA-NeMo/NeMo/pull/14765/files#r2402080512 

decoding_computer is initialized here that has self.state()

however in lines 342-349
```
            # initialize aed model state
            model_state = initialize_aed_model_state(
                asr_model=asr_model,
                decoder_input_ids=decoder_input_ids,
                batch_size=batch_size,
                context_encoder_frames=context_encoder_frames,
                chunk_secs=cfg.chunk_secs,
                right_context_secs=cfg.right_context_secs,
            )
```

this state is overridden. So my question was was the initial self.state() is required at all?

nithinraok:
@andrusenkoau  pls Add a CI-CD run for this script with default as alignattention as waitk latency is much higher.

tests/functional_tests/L2_Speech_Transcription_Canary_Streaming_Full_Manifest.sh
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+
+# Test with waitk policy
+coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py \
```
artbataev, line 18: To run this test on CI, you need to add it to `.github/workflows/cicd-main-speech.yml`

nemo/collections/asr/modules/transformer/transformer_generators.py
```diff
@@ -157,12 +157,13 @@ def _one_step_forward(
...
         else:
-            decoder_mems_list = self.decoder.forward(
+            decoder_mems_list, _ = self.decoder.forward(
                 decoder_hidden_states, decoder_input_mask, decoder_mems_list, return_mems=True
             )
         with self.classifier.with_log_softmax_enabled(return_scores) as clf:
             logits = clf.forward(hidden_states=decoder_mems_list[-1][:, -1:])
-        return logits, decoder_mems_list
+
+        return logits, decoder_mems_list, xatt_scores_list
```
artbataev, line 167: if `encoder_hidden_states` is None, `xatt_scores_list` is undefined (here will be an error)

nemo/collections/asr/modules/transformer/transformer_modules.py
```diff
@@ -109,10 +108,24 @@ def forward(self, input_ids, token_type_ids=None, start_pos=0):
...
         position_ids = torch.arange(
             start=start_pos, end=start_pos + seq_length, dtype=torch.long, device=input_ids.device
         )
         position_ids = position_ids.unsqueeze(0).repeat(input_ids.size(0), 1)
+        if torch.is_tensor(shift_pos):
+            # shift_pos is a tensor, so we need to add it to the position_ids
+            # and make sure that the resulting position_ids are within the
+            # range of the positional embedding
+            position_ids = position_ids + shift_pos
+            # position_ids = torch.clamp(position_ids, 0, self.max_sequence_length - 1)
```
artbataev: Please, remove commented out code

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,499 @@
...
+Otherwise, the audio buffer size needs to be increased. This policy shows lower latency in comparison with Wait-k.
+This policy is also suitable for window recognition (left context is fixed and not infinite), but you can lose some accuracy in this case.
+
+Recommended settings for audio buffer:
+- streaming with 1.5s latency: 10-1-0.5 (10s left, 1s chunk, 0.5s right)
+
+Example usage:
+
+```shell
+# TODO: update this example
```
artbataev: What should be updated here?

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
```diff
@@ -0,0 +1,499 @@
...
+        raise ValueError("Both cfg.audio_dir and cfg.dataset_manifest cannot be None!")
+
+    filepaths = None
+    manifest = cfg.dataset_manifest
+    if cfg.audio_dir is not None:
+        filepaths = list(glob.glob(os.path.join(cfg.audio_dir, f"**/*.{cfg.audio_type}"), recursive=True))
+        manifest = None  # ignore dataset_manifest if audio_dir and dataset_manifest both presents
+
+    # setup GPU
+    if cfg.cuda is None:
```
artbataev: There are recently introduced functions for device/dtype selection, please use them

```
from nemo.collections.asr.parts.utils.transcribe_utils import get_inference_device, get_inference_dtype

map_location = get_inference_device(cuda=cfg.cuda, allow_mps=cfg.allow_mps)
compute_dtype = get_inference_dtype(cfg.compute_dtype, device=map_location)
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+    )
+    use_avgpool_for_alignatt: bool = False  # use avgpooling for alignatt decoding policy
+    tokens_frame_alignment: torch.Tensor = (
+        None  # frame alignment of the predicted tokens (used for LAAL calculation in alignatt)
+    )
+    prev_encoder_shift: int = 0  # previous encoder shift (used for LAAL calculation in alignatt)
+    device: torch.device = None
+
+
+class GreedyBatchedStreamingAEDComputer(ABC):
```
artbataev: Is inheritance from `ABC` needed? 
It looks like there are no abstract methods in this class -> ABC is unnecessary

artbataev: This class is also instantiated directly in the script

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+        asr_model.prompt.encode_dialog(turns=turns)["context_ids"]
+        .unsqueeze(0)
+        .repeat(decoding_config.batch_size, 1)
+        .to(asr_model.device)
+    )
+
+    return decoder_input_ids
+
+
+def initialize_aed_model_state(
```
artbataev: Suggestion: maybe it would be a bit cleaner to make it static/classmethod of `GreedyBatchedStreamingAEDComputer`

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+
+from nemo.collections.asr.models.aed_multitask_models import EncDecMultiTaskModel, parse_multitask_prompt
+from nemo.collections.asr.parts.submodules.multitask_decoding import AEDStreamingDecodingConfig
+from nemo.collections.asr.parts.utils.streaming_utils import ContextSize
+from nemo.utils import logging
+
+
+@dataclass
+class AEDStreamingState:
+    decoder_input_ids: torch.Tensor = None  # tokens ids of initial AED model prompt
```
artbataev:
```suggestion
    decoder_input_ids: torch.Tensor | None = None  # tokens ids of initial AED model prompt
    pred_tokens_ids: torch.Tensor | None = None  # buffer with predicted tokens ids
```

artbataev: Implicit optional type like
```python
x: int = None
```
is considered incorrect in Python. It was supported by official type checker `mypy` for some time, but later it was removed by default.
PEP 484 was updated to explicitly recommend avoiding such annotations, see https://peps.python.org/pep-0484/#union-types

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+                disable_samples_mask *= torch.logical_not(self.state.is_last_chunk_batch)
+                self.state.active_samples_inner_loop *= torch.logical_not(disable_samples_mask)
+
+            if not torch.any(self.state.active_samples_inner_loop):
+                break
+
+    def detect_hallucinations(self, pred_tokens_ids, batch_idxs, current_context_lengths):
+
+        # we need to have at least 8 tokens to run hallucinations detector
+        if self.state.decoding_step < 8:
```
artbataev: Do we have any guarantee that on decoding step 8 we have enough tokens for all samples in the batch?
What if for the first utterance we have 8 tokens, but the last utterance in the batch contains 0 tokens (just silence)?

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+        # pattern 3: "a b c a b c a b c"
+        hallucination_mask_3 = (
+            (pred_tokens_ids[batch_idxs, ccl] == pred_tokens_ids[batch_idxs, ccl - 3])
+            * (pred_tokens_ids[batch_idxs, ccl - 1] == pred_tokens_ids[batch_idxs, ccl - 4])
+            * (pred_tokens_ids[batch_idxs, ccl - 2] == pred_tokens_ids[batch_idxs, ccl - 5])
+            * (pred_tokens_ids[batch_idxs, ccl] == pred_tokens_ids[batch_idxs, ccl - 6])
+            * (pred_tokens_ids[batch_idxs, ccl - 1] == pred_tokens_ids[batch_idxs, ccl - 7])
+            * (pred_tokens_ids[batch_idxs, ccl - 2] == pred_tokens_ids[batch_idxs, ccl - 8])
+        )
+        if torch.any(hallucination_mask_3):
```
artbataev, line 377: Each check is a sync operation. Maybe faster (one sync op instead of 3):
```python
# avoid checking each value before finally checking hallucination_mask
if torch.any(hallucination_mask): ... # output hallucination type
```

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
```diff
@@ -0,0 +1,562 @@
...
+            self.state.pred_tokens_ids[self.state.batch_idxs, self.state.current_context_lengths] = next_tokens
+
+            self.state.decoding_step += input_ids.size(-1)
+
+            # check for hallucinations
+            if self.decoding_cfg.hallucinations_detector:
+                hallucination_mask = self.detect_hallucinations(
+                    self.state.pred_tokens_ids, self.state.batch_idxs, self.state.current_context_lengths
+                )
+                if torch.any(hallucination_mask):
```
artbataev: Is this check necessary? Directly multiplying may be faster due to this blocking operation

artbataev (commented):
LGTM! I checked decoding with `nvidia/canary-180m-flash`, works well!
Minor comments/suggestions.

tests/functional_tests/L2_Speech_Transcription_Canary_Streaming_Full_Manifest.sh
andrusenkoau: done, thanks!

nemo/collections/asr/modules/transformer/transformer_generators.py
andrusenkoau: I added `xatt_scores_list = None` for this case.

nemo/collections/asr/modules/transformer/transformer_modules.py
andrusenkoau: done

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: This TODO needs to be removed. Done

examples/asr/asr_chunked_inference/aed/speech_to_text_aed_streaming_infer.py
andrusenkoau: done, thanks

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: done

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: good idea, done

nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py
andrusenkoau: done

nithinraok:
Thanks @andrusenkoau LGTM. 

Excellent work!

PR #14794, opened 2025-09-23
Add new spectral codec definition

rlangman:
This PR add the definition for the low-latency spectral decoder, as well as logic needed to train Magpie with it.

The codec runs at 25 FPS, with each frame having a 60 dimensional embedding, and each embedding value having 5 possible FSQ values.

By default the codec treats this as 10 codebooks each having size 5^6 = 15,625, which can fit in a 16 bit integer for storing on disk. For training I recommend treating it as 20 codebooks each having size 5^3 = 125 and sampling with topk=10. I find that using the smaller codebook size improves TTS quality, and reduces the time for each training step by about 30-40% (when doing on-the-fly codec extraction).

To regroup the codecs during training, add these lines to your training command:
```
+model.vector_quantizer._target_="nemo.collections.tts.modules.audio_codec_modules.GroupFiniteScalarQuantizer" \\
+model.vector_quantizer.num_groups=20 \\
+model.vector_quantizer.num_levels_per_group="[5, 5, 5]" \\
```

### DIFF
--- a/nemo/collections/common/parts/utils.py
+++ b/nemo/collections/common/parts/utils.py
@@ -159,12 +159,16 @@ def mask_sequence_tensor(tensor: torch.Tensor, lengths: torch.Tensor):
 
 class ClampActivation(nn.Module):
 
-    def __init__(self, min_value: float = -1.0, max_value: float = 1.0):
+    def __init__(self, min_value: float = -1.0, max_value: float = 1.0, clamp_training: bool = True):
         super().__init__()
         self.min_value = min_value
         self.max_value = max_value
+        self.clamp_training = clamp_training
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if self.training and not self.clamp_training:
+            return input
+
         return torch.clamp(input, min=self.min_value, max=self.max_value)
 
 
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -36,6 +36,7 @@ from nemo.collections.tts.losses.aligner_loss import ForwardSumLoss
 from nemo.collections.tts.models import AudioCodecModel
 from nemo.collections.tts.modules import transformer_2501
 from nemo.collections.tts.modules.aligner import AlignmentEncoder
+from nemo.collections.tts.modules.audio_codec_modules import VectorQuantizerIndexConverter
 from nemo.collections.tts.modules.magpietts_modules import (
     CharAwareSubwordEncoder,
     EOSDetectionMethod,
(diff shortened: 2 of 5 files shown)


nemo/collections/common/parts/utils.py
```diff
@@ -159,12 +159,16 @@ def mask_sequence_tensor(tensor: torch.Tensor, lengths: torch.Tensor):
 
 class ClampActivation(nn.Module):
```
blisc, line 160: So I understand why we might want to use this as an alternative to tanh, but why are we adding this clamp_training parameter?
We seem to only use it in the new ResNetDecoder class which defaults to clamp_training being False, which in turn, makes this only clamp during inference. Why do we want to do this?

nemo/collections/tts/models/magpietts.py
```diff
@@ -95,17 +96,30 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
 
         # load codec
         codec_model = AudioCodecModel.restore_from(cfg.get('codecmodel_path'), strict=False)
+
         self.sample_rate = codec_model.sample_rate
+        self.codec_model_samples_per_frame = codec_model.samples_per_frame
         # del codec discriminator to free memory
         del codec_model.discriminator
 
+        vector_quantizer = cfg.get('vector_quantizer')
```
blisc, line 108: Can you docstring this new config?

nemo/collections/tts/models/magpietts.py
```diff
@@ -2302,6 +2331,7 @@ def infer_batch(
                 all_codes_next_argmax = self.sample_codes_from_logits(
                     all_code_logits_t,
                     temperature=0.01,
+                    topk=topk,
```
blisc: This is meant to be an argmax since we hard code the temperature to 0.01, so adding the topk param here should not make a difference. In fact, I would revert this so it's clear that we did this by choice. Or alternatively, we should put topk=80 (since 80 is the default in the sample_codes_from_logits() function) and add a comment.

nemo/collections/tts/modules/audio_codec_modules.py
```diff
@@ -2714,6 +2764,9 @@ class MultiResolutionSTFTEncoder(NeuralModule):
             The total down sample rate of the encoder will be 2**(len(resolutions)) * product(down_sample_rate_list)
         kernel_size: Kernel size to use in all convolutions.
         activation: Name of activation function.
+        resample_rates: Optional tuple of two integers. If provided, input audio will be resampled from sampling rate
+            resample_rates[0] to sampling rate resample_rates[1].
+        pad_mode: Type of padding to use.
```
blisc: Can you list the supported options?

blisc (changes requested):

nemo/collections/common/parts/utils.py
rlangman: It is used in the `HiFiGANDecoder`, `CausalHiFiGANDecoder `classes with `clamp_training=True`. However with `clamp_training=True`, audio sample predicted outside the range [-1, 1] produce no gradient, which I found results in some model configurations not converging.

nemo/collections/tts/models/magpietts.py
rlangman: But then the function throws an error if your codebook size is less than 80 (e.g. I tried defining codebook size 5 and 25). If the temperature and topk values are irrelevant, then we should have a different function for argmax which does not require these parameters.

rlangman: Would it clearer to hardcode `topk=1` in the functional call?

blisc (approved):

PR #14797, opened 2025-09-24
Update Black version to 25

rfejgin:
The existing version 24 is no longer supported by VsCode and Cursor black plugins which makes it harder to format the files before submitting PRs.

Tested the updated version on a couple already-formatted files and formatting did not change, so this shouldn't cause major diffs.


artbataev:
Please note that you should also update `black` version:
- in `requirements.txt`
- in `.github/workflows/code-formatting.yml`

Since black reformatting is still incremental, the update should not lead to a huge amount of changes.

cc: @ko3n1g @pablo-garay

blisc:
We shouldn't change this on a dev branch. If we want to upgrade this, we should do this to main.

PR #14813, opened 2025-09-25
Checkout magpie branch instead of main

blisc:


.github/workflows/cicd-main.yml
```diff
@@ -227,11 +226,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
```
rfejgin, line 228: looks like some places use checkout@v2, others @v4 or @v5, not sure if there's a reason for it (or how much it matters)

.github/actions/test-template/action.yml
```diff
@@ -52,11 +52,11 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Noop
-      shell: bash
-      run: |
-        chmod -R u+rwX ${{ github.run_id }}
-        echo "noop"
+    # - name: Noop
```
rfejgin, line 55: are we keeping the commented code around to have a reference of what changed from main? otherwise I guess we can delete it.

PR #14821, opened 2025-09-26
Address problems where sometimes in 1m dataset there are very large m…

jstjohn:
## Description
Add a MAX_TAG_LEN field to handle the case where some unknown pattern is encountered in the data that leads to huge swaths of DNA being masked. Prefer to leave a potential tag unmasked than mask a huge amount of DNA. Non-ACGT characters are still individually masked as before. This addresses divide by zero encountered in num_tokens in megatron context parallelism when training at 1m context length and CP=2.

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/llm/gpt/data/megatron/hyena/evo2_dataset.py
+++ b/nemo/collections/llm/gpt/data/megatron/hyena/evo2_dataset.py
@@ -33,6 +33,7 @@ class Evo2Dataset(GPTDataset):
     RESET_PAD_EOD_MASK: bool = True  # If set, unset the mask for [pad] and [eod] tokens (matches Evo2 paper).
     # Valid DNA tokens: A, C, G, T, U, W, S, M, K, R, Y, B, D, H, V, N, -,  (both uppercase and lowercase and
     #   degenerate bases and RNA)
+    MAX_TAG_LEN = 2048
 
     VALID_DNA_AND_DEGENERATE: ClassVar[set[int]] = {
         45,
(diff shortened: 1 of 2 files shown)


jwilber (approved):
Approved with one comment on potential double-flip.

nemo/collections/llm/gpt/data/megatron/hyena/evo2_dataset.py
```diff
@@ -217,35 +219,52 @@ def process_segment(seg_seq: torch.Tensor) -> torch.Tensor:
...
             for end in pipe_pos:
-                if is_tag:
+                seg_len = end - start
+                if is_tag and seg_len < max_tag_len:  # Prefer faulty tag mask logic over masking entire seqs
                     seg_mask[start:end] = 0
+                elif is_tag and seg_len >= max_tag_len:
+                    is_tag = False  # Flip the state machine because the is_tag=True was for a very long region
                 else:
                     pass
                 is_tag = not is_tag  # Flip the state machine.
```
jwilber: So, if I understand this correctly. on line 259 we say "If we have a tag but it's too long, we'll just not label it as a tag anymore":

```
elif is_tag and seg_len >= max_tag_len: is_tag = False
```

But then, here on line 263, after the conditionals run, we flip things back? `is_tag - not_is_tag`?

jstjohn: Correct, the flip is there for the next block. Like if this block is X then the next block is "not X". So I'm saying above "I was wrong about _this_ block, so change this block" then the state is 'correct' going into the regular logic about flipping for the next block.

jstjohn: Addressed with comments inline.

PR #14833, opened 2025-09-28
[magpie][context audio] add speaker items limit to compute similarity matrix to avoid OOM.

XuesongYang:


scripts/magpietts/extend_nemo_manifest_with_context_audio.py
```diff
@@ -295,46 +308,80 @@ def _process_and_flush_speakers_local(self):
...
-                    original_candidate_idx = sorted_indices_tensor[i, candidate_rank].item()
+                    context_pool_idx = sorted_indices_tensor[i, candidate_rank].item()
 
-                    # Skip if candidate is the item itself (safeguard)
-                    if original_candidate_idx == i:
-                        continue
+                    # if ANY candidate has similarity ≤ -2.0, all subsequent ones will be ≤ -2.0
+                    # since similarities are sorted in descending order, we can break early
+                    if candidate_ssim <= -2.0:
+                        break
```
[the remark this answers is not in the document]
XuesongYang: good point. addressed.

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
```diff
@@ -35,6 +36,10 @@
 
 logger = logging.getLogger(__name__)
 
+# Constant for masking identical items in similarity matrix
+# Set below valid cosine similarity range [-1, 1] to ensure masked items are never selected
+MASKED_SIMILARITY_VALUE = -2.0
```
blisc, line 41: Can you elaborate a bit more on why we now have identical items whereas we didn't in the past?

blisc (approved):

scripts/magpietts/extend_nemo_manifest_with_context_audio.py
XuesongYang: > Can you elaborate a bit more on why we now have identical items whereas we didn't in the past?

The identical items masking logic hasn't changed. we've always needed to prevent items from being selected as their own context.

**Previously (N×N matrix)**: We used `similarity_matrix.fill_diagonal_(-2.0)` because in a full N×N similarity matrix, identical items always appeared on the diagonal (item `i` compared to itself at position `[i,i]`).

**Now (N×M matrix)**: We compute similarities between N items and a smaller context pool of M items (where M ≤ `max_speaker_items`). Since identical items are no longer guaranteed to be on the diagonal, we need to explicitly track and mask them. We create a mapping (`original_index_to_context_position`) to identify when an item from the N set appears in the M context pool, then apply the same masking value (`MASKED_SIMILARITY_VALUE = -2.0`) to prevent self-selection.

PR #14847, opened 2025-09-30
Docstrings and comments on EOS handling.

rfejgin:
This is a follow-up to:
https://github.com/NVIDIA-NeMo/NeMo/pull/14761

Added docstrings and more comments.

Also, changed the default value in of the the `forbid_audio_eos` argument to`get_forbidden_tokens()` to be `False` since permitting EOS is the more common use case. The default value is not was not being used so this doesn't change behavior.

### DIFF
--- a/nemo/collections/tts/modules/magpietts_modules.py
+++ b/nemo/collections/tts/modules/magpietts_modules.py
@@ -103,15 +103,12 @@ class SpecialAudioToken(Enum):
         return base_codebook_size + token.value
 
     @staticmethod
-    def get_forbidden_tokens(base_codebook_size: int, forbid_audio_eos: bool = True) -> list[int]:
+    def get_forbidden_tokens(base_codebook_size: int, forbid_audio_eos: bool = False) -> list[int]:
         """
         Returns a list of token indices that should not be sampled or returned to user.
         Args:
             base_codebook_size (int): The size of the codec codebook (which is the first part of the embedding table).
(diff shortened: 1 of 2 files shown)


nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -103,15 +103,12 @@ def get_index(token: SpecialAudioToken, base_codebook_size: int):
...
+    def get_forbidden_tokens(base_codebook_size: int, forbid_audio_eos: bool = False) -> list[int]:
         """
         Returns a list of token indices that should not be sampled or returned to user.
         Args:
             base_codebook_size (int): The size of the codec codebook (which is the first part of the embedding table).
-            forbid_audio_eos (bool): Whether to forbid the AUDIO_EOS token to be sampled.
-                                    * Set to `False` when internally generating tokens in MagpieTTS sampling
-                                    * Set to `True` when checking validity of tokens to be returned to user
-                                      or given to the codec for decoding
+            forbid_audio_eos (bool): Whether AUDIO_EOS should be forbidden. Default: False (i.e. allowed).
```
blisc, line 111: Let's undo this part. It's helpful to be more verbose.

rfejgin:
Will add more docstrings, moving back to draft for now

rfejgin:
I added detailed docstrings for all three sampling methods.

I am leaving `infer_batch()` for later as it would be a larger change and probably require input from the team since it's the model's external API.

PR #14894, opened 2025-10-07
New ML yaml + changes to allow for Spectral Codec training with text context

blisc:


nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -550,7 +550,9 @@ def __getitem__(self, index):
                 example['context_audio_codes_len'] = context_audio_codes_len
             else:
                 # @shehzeenh: Added this condition so that a batch does not have a mix of context_audio and context_audio_codes
-                context_audio = torch.zeros(self.codec_model_samples_per_frame, dtype=torch.float32)
+                # @blisc: Added a +1. If we send in exactly 882 samples, then a conv layer complains about padding.
+                #         Adding 883 works. This occurs when we use text context during inference.
+                context_audio = torch.zeros(self.codec_model_samples_per_frame + 1, dtype=torch.float32)
```
rlangman, line 555: I am not sure if there is a clean way to handle reflect/replicate padding when input is short. We could have each codec architecture define a minimum length it can handle, and then have `pad_audio` zero pad to at least that length: https://github.com/blisc/NeMo/blob/magpietts_2503/nemo/collections/tts/models/audio_codec.py#L453

rlangman (approved):
Looks good to me.

nemo/collections/tts/data/text_to_speech_dataset.py
blisc: I agree, we should do that.

PR #14923, opened 2025-10-13
Inference metrics improvements

rfejgin:
Added these two metrics to `infer_and_evaluate.py`:
* Add MOS estimation using UTMOSv2
* Add a metric to tracking total generated audio duration per dataset. We can use this as a lightweight indicator of speaking rate.
* Rename a command line option for consistency: `disable_fcd` --> `--no_fcd`

### DIFF
--- a/scripts/magpietts/evaluate_generated_audio.py
+++ b/scripts/magpietts/evaluate_generated_audio.py
@@ -168,10 +168,10 @@ def extract_embedding(model, extractor, audio_path, device, sv_model_type):
     return embeddings.squeeze()
 
 
-def compute_utmosv2_scores(audio_dir):
+def compute_utmosv2_scores(audio_dir, device):
     print(f"\nComputing UTMOSv2 scores for files in {audio_dir}...")
     start_time = time.time()
-    utmosv2_calculator = UTMOSv2Calculator()
+    utmosv2_calculator = UTMOSv2Calculator(device=device)
     utmosv2_scores = utmosv2_calculator.process_directory(audio_dir)
     # convert to to a dictionary indexed by file path
     utmosv2_scores_dict = {os.path.normpath(item['file_path']): item['predicted_mos'] for item in utmosv2_scores}
@@ -245,7 +245,7 @@ def evaluate(
         fcd_metric = None
 
     if with_utmosv2:
-        utmosv2_scores = compute_utmosv2_scores(generated_audio_dir)
+        utmosv2_scores = compute_utmosv2_scores(generated_audio_dir, device)
     filewise_metrics = []
     pred_texts = []
     gt_texts = []
--- a/scripts/magpietts/infer_and_evaluate.py
+++ b/scripts/magpietts/infer_and_evaluate.py
@@ -376,14 +376,6 @@ def run_inference(
         print(f"Evaluating dataset {dataset}")
         metrics_n_repeated = []
         manifest_records = read_manifest(dataset_meta_info[dataset]['manifest_path'])
-        utmos_max_records = 200
-        if with_utmosv2 and len(manifest_records) > utmos_max_records:
-            use_utmosv2_for_dataset = False
-            print(
-                f"Skipping UTMOSv2 calculation for dataset {dataset} because it has more than {utmos_max_records} records"
-            )
-        else:
-            use_utmosv2_for_dataset = with_utmosv2
         language = dataset_meta_info[dataset].get('whisper_language', 'en')
         dataset_meta_for_dl = copy.deepcopy(dataset_meta_info[dataset])
         for key in ["whisper_language", "load_cached_codes_if_available"]:
@@ -546,7 +538,7 @@ def run_inference(
                 sv_model_type=sv_model,
                 asr_model_name=asr_model_name,
                 codecmodel_path=codecmodel_path if compute_fcd else None,
-                with_utmosv2=use_utmosv2_for_dataset,
+                with_utmosv2=with_utmosv2,
             )
             metrics_n_repeated.append(metrics)
             dataset_filewise_metrics_all_repeats.extend(
@@ -572,11 +564,7 @@ def run_inference(
                 print(f"Wrote metrics for {checkpoint_name} and {dataset} to {all_experiment_csv}")
 
             output_png_file = Path(eval_dir) / f"{dataset}_violin_{repeat_idx}.png"
-            violin_plot_metrics_current_dataset = violin_plot_metrics.copy()
-            if 'utmosv2' in violin_plot_metrics_current_dataset and not use_utmosv2_for_dataset:
-                # Don't include utmosv2 in the violin plot if it's not used for this dataset
-                violin_plot_metrics_current_dataset.remove('utmosv2')
-            create_violin_plots(filewise_metrics, violin_plot_metrics_current_dataset, output_png_file)
+            create_violin_plots(filewise_metrics, violin_plot_metrics, output_png_file)
 
             # Clean up temporary codec files
             for codes_file in codec_file_paths:


scripts/magpietts/infer_and_evaluate.py
```diff
@@ -295,6 +295,7 @@ def run_inference(
     violin_plot_metrics=['cer', 'pred_context_ssim'],
```
rfejgin, line 295: I did not add UTMOS to the set of violin plots, should we though?  Would be interested in comments.

Pro: this way it'll actually be looked at more frequently
Con: the combined plot would get crowded with 3 violin plots per dataset. (Or maybe UTMOS can be added only to the single-dataset plots?)

.github/workflows/cicd-main-speech.yml
```diff
@@ -194,6 +194,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      # Temporarily install this manually until we add it to the docker image.
+      - name: Install UTMOSv2 # Needed by evaluate_generated_audio.py.
+        run: pip install git+https://github.com/sarulab-speech/UTMOSv2.git
```
shehzeen: should we use a specific commit or are we comfortable with using their latest main branch?

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -408,7 +431,8 @@ def evaluate(
         hypotheses=gt_audio_texts, references=gt_texts, use_cer=False
     )[0]
     avg_metrics["frechet_codec_distance"] = fcd
-
+    avg_metrics["utmosv2_avg"] = sum([m['utmosv2'] for m in filewise_metrics]) / len(filewise_metrics)
+    avg_metrics["total_gen_audio_seconds"] = total_generated_audio_seconds
```
shehzeen, line 450: How about instead of this, we use phonemes/second? If it's challenging to get number of phonemes, we can use characters or words (split by spaces or special chars) instead.

scripts/magpietts/infer_and_evaluate.py
shehzeen: I think we should add it to the violin plots! It's a valuable metric to look at when comparing models.

.github/workflows/cicd-main-speech.yml
rfejgin: Yeah, I was wondering the same. But I looked at their recent commits and it's basically infrequent maintenance updates (e.g. compatibility with new versions of torch). Probably less than 1 commit a month on average. So I tend to think we benefit more from using latest on main branch since it should have bugfixes etc. but minimal churn from updates.

scripts/magpietts/infer_and_evaluate.py
rfejgin: Okay good — added! `UTMOSv2` is now included in the violin plots by default. The user can still choose a different set a metrics if desired via the `--violin_plot_metrics` option.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Yeah, we started discussing a bit in the meeting and Jason was of the opinion we should start simple, with just total duration. @blisc - any further thoughts?

.github/workflows/cicd-main-speech.yml
rfejgin: Anyways: thought about it some more and decided to update pip-requirements to pin UTMOSv2 to its latest release, `v1.2.1`.

scripts/magpietts/evaluate_generated_audio.py
rfejgin: Discussed this a bit with Jason - speech rate would be useful but we'll leave it for a future PR. This PR was initially meant to focus on UTMOS and we don't want to complicate the PR too much.

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -682,17 +686,18 @@ def main():
...
     parser.add_argument('--cer_target', type=float, default=None)
     parser.add_argument('--ssim_target', type=float, default=None)
+    parser.add_argument('--no-utmosv2', action='store_true', help="Disable UTMOSv2 computation")
     parser.add_argument(
         '--log_exp_name',
         action='store_true',
         help="Include the experiment name (derived from the checkpoint path) in the output folder name.",
     )
-    parser.add_argument('--disable_fcd', action='store_true', help="Disable Frechet Codec Distance computation")
+    parser.add_argument('--no_fcd', action='store_true', help="Disable Frechet Codec Distance computation")
```
blisc: Stylistically, let's keep the older format of `disable_{metric}`

scripts/magpietts/infer_and_evaluate.py
```diff
@@ -295,6 +295,7 @@ def run_inference(
     violin_plot_metrics=['cer', 'pred_context_ssim'],
     eos_detection_method=None,
     ignore_finished_sentence_tracking=False,
+    with_utmosv2=True,
```
blisc, line 298: Since we know that this is a slow inference call, can we default this to be disabled? And enable it only for some smaller eval datasets?

nemo/collections/tts/modules/utmosv2.py
```diff
@@ -0,0 +1,58 @@
...
+            # without actually speeding up prediction. Limit to 4 threads.
+            with threadpool_limits(limits=4):
+                mos_score = self.model.predict(input_path=file_path, num_repetitions=1, num_workers=0)
+        return mos_score
+
+
+def get_available_device():
+    """
+    Get the best available device (prefer GPU, fallback to CPU).
+    """
```
blisc, line 77: Let's refactor this. Can you include a device in UTMOSv2Calculator's init that defaults to None? In our script, we should pass the a device that makes sense to our scripts. If it is None, we can then call this function, but we shouldn't rely on this to decide where to place the eval model.

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -334,6 +351,11 @@ def evaluate(
                 gt_context_ssim_alternate = torch.nn.functional.cosine_similarity(
                     gt_speaker_embedding_alternate, context_speaker_embedding_alternate, dim=0
                 ).item()
+            if with_utmosv2:
+                utmosv2_score = utmosv2_model(pred_audio_filepath)
```
blisc: Is there a reason that we want to run this at batch 1? This might require a larger refactor of the code though. so if we cannot do it in this PR, can we skip this comment

blisc (changes requested):

scripts/magpietts/infer_and_evaluate.py
rfejgin: Done

scripts/magpietts/infer_and_evaluate.py
rfejgin: I've changed it so that if the dataset has more than 200 entries we skip UTMOS calculation.

nemo/collections/tts/modules/utmosv2.py
rfejgin: Done

scripts/magpietts/evaluate_generated_audio.py
rfejgin: I had played around with that earlier on and the issue was the speed was mostly bottlenecked CPU-side processing (feature extraction). And it was tricky to tune the number of workers and OpenMP threads to get an actual speedup from batching.

But I played around it a bit more now and I think I found a faster configuration (with batching and a particular thread configuration). Still need to integrate the directory-based processing with the main evaluation script.

scripts/magpietts/evaluate_generated_audio.py
```diff
@@ -153,6 +168,18 @@ def extract_embedding(model, extractor, audio_path, device, sv_model_type):
     return embeddings.squeeze()
 
 
+def compute_utmosv2_scores(audio_dir):
+    print(f"\nComputing UTMOSv2 scores for files in {audio_dir}...")
+    start_time = time.time()
+    utmosv2_calculator = UTMOSv2Calculator()
```
blisc: Can we use the same "device" as on line 200: `device = "cuda"`

rfejgin: Good catch! Fixed.

rfejgin:
I removed the restriction limiting UTMOSv2 to datasets with under 200 utterances. That's because it's a lot faster to run now that it is batched and with inference workers and threads tuned. For a 2000-utterance dataset, UTMOSv2 only increased test time from ~25 min to 28 min which seems worth it.

Will merge once CI passes (excluding unrelated known CI issue with coverage calculation).

PR #14927, opened 2025-10-14
Update segment.py - add opus to supported formats + librosa to load files more efficient

SAMortazavi:
### PR Description

In the current implementation, `.opus` files are not included in the list of supported audio formats, even though `soundfile (sf)` can successfully load them. Additionally, using `pydub` for segment extraction is inefficient for long audio files, as it first loads the entire file into memory before clipping segments.

This PR addresses the issue by adding `.opus` to the supported formats and improving loading efficiency by introducing `librosa` for audio loading. If `librosa` and `soundfile` fail to load the audio, the fallback mechanism uses `pydub` as a backup loader.

### Relevant Code

 [View on GitHub](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/preprocessing/segment.py#L62)

### Summary of Changes

* Added `.opus` to supported audio formats.
* Added `librosa` for efficient audio loading to improve performance on long audio files.
* Implemented a fallback mechanism: if `librosa` and `soundfile` fail, the loader falls back to `pydub`.
* Updated segment extraction logic to handle `librosa`-loaded data efficiently.


SAMortazavi (commented):
done

PR #14974, opened 2025-10-23
Update checkpoint saving logic: Use step instead of epoch

blisc:
By defaulting to step as opposed to epoch, we now save a checkpoint at end of training.

In the older logic, at the end of training, the checkpoint attempts to save a final checkpoint by formatting the checkpoint into the `f'{name}--{{{params.monitor}:.4f}}-{{epoch}}'` structure. But since this file already exists, as we saved it after the last validation run, the final checkpoint will not save. By replacing epoch with step, we should now always save a unique checkpoint on every run. In addition for our new datasets since we sample data, an "epoch" is not defined so step or tokens consumed is a more natural choice.

### DIFF
--- a/examples/tts/conf/magpietts/magpietts_multilingual_v2_lhotse.yaml
+++ b/examples/tts/conf/magpietts/magpietts_multilingual_v2_lhotse.yaml
@@ -179,6 +179,20 @@ model:
     max_length_causal_mask: 2048
     use_learnable_pos_emb: true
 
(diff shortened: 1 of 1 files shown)


XuesongYang (approved):
LGTM. It would be also great if suffix with consumed samples in the future.

blisc:
> LGTM. It would be also great if suffix with consumed samples in the future.

Does Lhotse give us an option to track this?

PR #15054, opened 2025-11-10
[TTS] Refactor BOS and EOS handling to fix codec conversion

rlangman:
This PR fixes a bug in MagpieTTS training in which `self._codec_converter.convert_original_to_new()` is being called on audio codec tokens after the Lhotse data loader has already pre-populated them with BOS and EOS tokens.

The addition of BOS and EOS tokens is now always done inside the model class, after convert_original_to_new() is called. These tokens are also removed before `convert_new_to_original()` is called, so that they are not mistakenly passed into the codec model.

There are also a few other minor changes, including:
- Fix data loading field mismatch when doing on-the-fly codec extraction (we now use _target_audio_ instead of _recording_ and _context_audio_ instead of _context_recording_)
- Fix bug with inconsistent padding in spectral codec encoder
- Replace some manual padding done with `torch.cat` with simpler call to `torch.nn.functional.pad`


blisc:
Closing, please make a new PR into main

PR #15066, opened 2025-11-13
Implement Nemotron-VoiceChat Speech Decoder

Edresson:
# What does this PR do ?

It implements Nemotron-VoiceChat Speech Decoder.

### DIFF
--- a/tests/collections/speechlm2/test_duplex_eartts.py
+++ b/tests/collections/speechlm2/test_duplex_eartts.py
@@ -219,8 +219,8 @@ def dataset(model):
 
 @pytest.fixture(scope="session")
 def training_cutset_batch():
-    cut = dummy_cut(0, recording=dummy_recording(0, with_data=True))
-    cut.target_audio = dummy_recording(1, with_data=True)
+    cut = dummy_cut(0, recording=dummy_recording(0, with_data=True, duration=1.0, sampling_rate=22050))
+    cut.target_audio = dummy_recording(1, with_data=True, duration=1.0, sampling_rate=22050)
     cut.supervisions = [
         SupervisionSegment(
             id=cut.id,
@@ -301,9 +301,6 @@ def test_eartts_dataset(dataset, training_cutset_batch):
 
     for key in tensor_keys:
         assert torch.is_tensor(batch[key]), f"{key} must be a tensor"
-
-    assert batch["source_audio"].shape == (1, 89082)
-    assert batch["target_audio"].shape == (1, 89082)
 
     # Check target text consistency
     assert batch["target_texts"] == ["hello okay"]


nemo/collections/speechlm2/parts/metrics/wer.py
```diff
@@ -1,65 +0,0 @@
-# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
zhehuaichen, line 1: is it necessary to remove this and add a Intelligence file instead?

nemo/collections/speechlm2/parts/metrics/secs.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
zhehuaichen, line 3: do you have this in TTS collection? can we import it from there?

tests/collections/speechlm2/test_duplex_eartts.py
```diff
@@ -0,0 +1,486 @@
...
+    assert results["loss"] > 0
+
+
+def test_eartts_validation_step(model, dataset, training_cutset_batch):
+    model.eval()
+    model.on_validation_epoch_start()
+    batch = dataset[training_cutset_batch]
+    batch = move_data_to_device(batch, device=model.device)
+    results = model.validation_step({"dummy_val_set": batch}, batch_idx=0)
+    assert results is None  # no return value
```
zhehuaichen, line 537: why no return value?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,573 @@
...
+class DuplexEARTTSDataset(torch.utils.data.Dataset):
+    """
+    A dataset for duplex speech-to-speech models that handles bidirectional conversations.
+
+    This dataset processes Lhotse CutSet objects containing recordings with supervision segments
+    from different speakers (roles). It creates aligned representations of audio and text for
+    both source (input) and target (output) channels, preserving temporal alignment between
+    audio frames and text tokens.
+
+    Args:
```
pzelasko, line 40: Missing parameters in documentation

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,573 @@
...
+        self.num_delay_speech_tokens = num_delay_speech_tokens
+
+        assert tokenizer.bos is not None, "BOS support in the tokenizer is required for S2S models."
+        assert tokenizer.eos is not None, "EOS support in the tokenizer is required for S2S models."
+
+    def generate_prompt_description(self, device):
+        messages = []
+        if random.random() > self.p_drop_description:
+            # ToDo: add extra system prompts
+            system_prompt = (
```
pzelasko: Why is it hardcoded? It looks like something that should come from data itself, with a fallback to reading the default from the config.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,573 @@
...
+                "However, this description should influence only the delivery of your response, not its content. "
+                "Your response should remain independent of any stylistic instructions."
+            )
+            messages.append({"role": "system", "content": system_prompt})
+        else:
+            messages.append({"role": "system", "content": ""})
+
+        # given that descriptions are currently not supported, only added the user prompt
+        # ToDo: add extra user prompts or completly remove it as it is not used in NanoV2
+        user_prompt = "Can you tell me something interesting?"
```
pzelasko: Same comment as above

nemo/collections/speechlm2/modules/ear_tts_commons.py
```diff
@@ -0,0 +1,390 @@
...
+# ==============================================================================
+PYTHON_CONFIG_GETTER_NAME = "get_config"
+CHECKPOINT_FORMAT = "checkpoint_{}/ema.safetensors"
+CONFIG_NAME = "config.json"
+GIT_HASH_NAME = "githash"
+SCRIPT_PLACEHOLDER = "[[[<<<SCRIPT_PLACEHOLDER>>>]]]"
+
+
+# ==============================================================================
+# Configuration Class and Utilities
```
pzelasko: What are these for?

nemo/collections/speechlm2/modules/ear_tts_commons.py
```diff
@@ -0,0 +1,390 @@
...
+    cfg.workdir_path = workdir_path
+    return cfg
+
+
+# ==============================================================================
+# Base Model Classes
+# ==============================================================================
+
+
+class PreTrainedModel(nn.Module):
```
pzelasko: This looks like it duplicates the functionality of HFHubMixin from `speechlm2/parts/pretrained.py` - why is it needed?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,573 @@
...
+
+        input_ids = []
+        for i, non_script in enumerate(non_script_list):
+            desc_ids = self.tokenizer.text_to_ids(non_script)
+            input_ids.extend(desc_ids)
+
+        input_ids = torch.tensor(input_ids, dtype=torch.long, device=device).view(1, -1)
+        return input_ids
+
+    def __getitem__(self, cuts: CutSet) -> dict:
```
pzelasko, line 156: Can we refactor this method into smaller functions, doing one task, with clear descriptive names? This is too complex to parse.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,573 @@
...
+        tokenizer,
+        frame_length: Seconds,
+        source_sample_rate: int,
+        target_sample_rate: int,
+        input_roles: list[str] = None,
+        output_roles: list[str] = None,
+        p_drop_description: float = 0.0,
+        add_text_bos_and_eos_in_each_turn: bool = False,
+        add_audio_prompt_after_description: bool = False,
+        audio_prompt_duration: float = 3.0,
```
pzelasko, line 134: Shouldn't the duration be dynamic and data-dependent? Why is it fixed?

examples/speechlm2/duplex_eartts_eval.py
```diff
@@ -0,0 +1,63 @@
...
+        input_roles=cfg.data.input_roles,
+        output_roles=cfg.data.output_roles,
+        add_text_bos_and_eos_in_each_turn=cfg.data.get("add_text_bos_and_eos_in_each_turn", True),
+        add_audio_prompt_after_description=cfg.data.add_audio_prompt_after_description,
+        audio_prompt_duration=cfg.data.audio_prompt_duration,
+        num_delay_speech_tokens=cfg.model.get("num_delay_speech_tokens", 2),
+    )
+    datamodule = DataModule(cfg.data, tokenizer=model.tokenizer, dataset=dataset)
+
+    trainer.validate(model, datamodule)
```
pzelasko, line 97: Why are we using `.validate` and not `.predict` for inference?

nemo/collections/speechlm2/parts/metrics/bleu.py
```diff
@@ -58,7 +56,8 @@ def compute(self) -> dict[str, torch.Tensor]:
             metric = torch.tensor(sacrebleu.corpus_bleu(self._hyps[name], [self._refs[name]]).score)
             corpus_metric[f"txt_bleu_{name}"] = metric
         corpus_metric["txt_bleu"] = torch.stack(list(corpus_metric.values())).mean()
-        self.reset()
+        self._refs.clear()
+        self._hyps.clear()
```
pzelasko: The changes in this file look like a bug to me.

nemo/collections/speechlm2/parts/metrics/wer.py
pzelasko, line 1: Why is this file removed? What about other models and recipes that use this?

tests/collections/speechlm2/test_duplex_eartts.py
```diff
@@ -0,0 +1,486 @@
...
+    # --- Shape/value checks similar to the original test ---
+
+    # Audio shapes (you can adjust if needed)
+    assert batch["source_audio"].shape == (1, 89082)
+    assert batch["target_audio"].shape == (1, 89082)
+
+    # Target text consistency
+    assert batch["target_texts"] == ["hello okay"]
+
+    # Token checks (same content as your old test)
```
pzelasko: Clean up AI comments :)

tests/collections/speechlm2/test_metrics.py
```diff
@@ -34,19 +34,25 @@ def test_bleu():
     assert ans["txt_bleu"] == 50.0  # average across datasets
 
 
-def test_wer():
```
pzelasko, line 37: We should restore this test and create a new one for "intelligibility"

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+                pred_audio_sr=self.target_sample_rate,
+                user_audio=dataset_batch["source_audio"].float(),
+                user_audio_sr=self.source_sample_rate,
+                eou_pred=eou_labels,
+                fps=self.target_fps,
+                results=results if self.cfg.get("dump_tokens_text", False) else None,
+                tokenizer=self.tokenizer,
+            )
+
+    def validation_step(self, batch: dict, batch_idx: int):
```
pzelasko, line 771: I don't like how this function is structured.
Inside, you are loading data from audio paths that come in batch.
This should be coming from the dataloader instead.

I would prefer if we maintain a clear separation:
* model inference methods accept tensors
* we can build higher level inference methods beyond that, which use model's methods as building blocks

Ideally, the inference script wouldn't be calling `trainer.validate()` on the model.
It would look more like `examples/speechlm2/salm_generate.py`:
It has an explicit dataloading stage, and an explicit loop over batches that runs the inference using methods defined by the model.

It is also OK to have a high-level, user friendly inference method like in `SALM` i.e. `SALM.generate()`.
In the case of half-duplex model it accepts a list of conversation turns with roles, text content, or audio paths.
But in that case make it very explicit that it is a high-level inference method, with public facing API - validation_step is not it (it could be something like `offline_inference` or `simulated_online_inference` etc).

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+                        pred_audio_sr=self.target_sample_rate,
+                        user_audio=None,
+                        user_audio_sr=None,
+                        eou_pred=None,
+                        fps=self.target_fps,
+                        results=None,
+                        tokenizer=self.tokenizer,
+                    )
+
+        else:
```
pzelasko: Avoid such large if/else - refactor this into separate methods that serve different use-cases. 

I am concerned about such major differences between the two conditional branches here.
What is the reason for this? Can this be more streamlined?

Maybe you need to implement one of those branches as `predict_step` (or ideally follow my suggestions from the comment above).

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+            system_prompt = (
+                "You engage in conversation with the user. When delivering your response as speech, "
+                "if the user provides a description such as emotions, scene details, "
+                "or speaker style, you adjust your speaking style accordingly when delivering the response. "
+                "However, this description should influence only the delivery of your response, not its content. "
+                "Your response should remain independent of any stylistic instructions."
+            )
+        messages.append({"role": "system", "content": system_prompt})
+
+        # ToDo: implement dataloading support for descriptions
```
pzelasko: marking TODO comment

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+            )
+            audio_pred, audio_len = self.audio_codec.decode(gen_audio_codes_history, gen_audio_codes_lens)
+
+        # return only the current/lastest audio chunk
+        audio_pred_cur_step = audio_pred.squeeze(1)[:, -self.audio_codec.config.wav_to_token_ratio :]
+        audio_len[:] = self.audio_codec.config.wav_to_token_ratio
+        return audio_pred_cur_step, audio_len
+
+    @torch.no_grad()
+    def offline_inference(
```
pzelasko, line 1086: Why isn't this function used in the inference script rather than `trainer.validate`?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+
+    def backward(self, *args, **kwargs):
+        with loss_parallel():
+            super().backward(*args, **kwargs)
+
+    def configure_optimizers(self):
+        return configure_optimizers(self)
+
+    @property
+    def oomptimizer_schema(self) -> dict:
```
pzelasko, line 1246: Is this correct? If not, raise NotImplementedError

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+                )
+            else:
+                self.embed_text_tokens = fully_shard(self.embed_text_tokens, **fsdp_config)
+                # self.tts_model = fully_shard(self.tts_model, **fsdp_config)
+                self.tts_model.mog_head = fully_shard(self.tts_model.mog_head, **fsdp_config)
+                self.tts_model.embed_subword = fully_shard(self.tts_model.embed_subword, **fsdp_config)
+                self.tts_model.embed_context = fully_shard(self.tts_model.embed_context, **fsdp_config)
+                self.tts_model.embed_code = fully_shard(self.tts_model.embed_code, **fsdp_config)
+                self.tts_model.null_emb = fully_shard(self.tts_model.null_emb, **fsdp_config)
+                self.tts_model.bos_emb = fully_shard(self.tts_model.bos_emb, **fsdp_config)
```
pzelasko: I don't know the size of those null/bos emb, but if they are very small, please don't shard them, it will only generate communication overhead.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+        Loads model weights a pretrained checkpoint file, supporting partial loading from .nemo and PyTorch formats.
+
+        Args:
+            checkpoint_path (str): Path to checkpoint file.
+
+        Returns:
+            None. The model is updated in-place.
+        """
+        if checkpoint_path is not None:
+            if '.nemo' in checkpoint_path:
```
pzelasko: Why would `.nemo` be in checkpoint_path? speechlm2 doesn't save .nemo checkpoints, it saves in .safetensors

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+            # You can check your tokenizer's vocabulary for the best candidate.
+            logging.warning("Tokenizer does not have a `bos_token`. Setting it to '<|im_start|>'.")
+            self.tokenizer.bos_token = '<|im_start|>'
+            self.tokenizer.eos_token = '<|im_end|>'
+
+        elif 'Nemotron' in self.cfg.pretrained_lm_name:
+            # ====== NEMOTRON-SPECIFIC HANDLING ======
+            self.tokenizer.bos_token = '<s>'
+            self.tokenizer.eos_token = '</s>'
+            self.tokenizer.pad_token = '<SPECIAL_12>'
```
pzelasko: Ouch, let's not hardcode specific model names. Can we find a more general way that if qwen / if nemotron (I don't know this recipe well enough to suggest something specific atm)

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+        aligned_position_ids = batch["aligned_position_ids"]
+
+        # extract target audio codes
+        target_audio, target_audio_lens = self.pad_audio_to_factor(
+            target_audio, target_audio_lens, self.target_samples_per_frame, 1
+        )
+        with ensures_target_precision(self.audio_codec_run_dtype), torch.no_grad():
+            target_codes, target_codes_lens = self.audio_codec.encode(target_audio.unsqueeze(1), target_audio_lens)
+
+        # ToDo: consider use the source audio
```
pzelasko: Marking TODO comment

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+        """Get default generation config for EAR-TTS."""
+        return {
+            "num_iter": 8,
+            "guidance_scale": self.cfg.get("inference_guidance_scale", 0.5) if guidance_enabled else None,
+            "top_p_or_k": self.cfg.get("inference_top_p_or_k", 0.8),
+            "noise_scale": self.cfg.get("inference_noise_scale", 0.8),
+            "eos_threshold": -3.0,
+        }
+
+    def offline_inference_with_custom_sentences(
```
pzelasko: This should go into a dedicated inference script, and not into the public API of the model.
From a user's perspective - what would "custom sentences" even mean?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+            fp32_layers (list): Names of layers to keep FP32.
+            mixed_dtype (torch.dtype): Target dtype for mixed precision layers.
+        """
+        logging.info(f"Converting TTS model to mixed precision. FP32 layers: {fp32_layers}")
+        for name, module in self.tts_model.named_modules():
+            if isinstance(module, (nn.Linear, nn.LayerNorm, nn.Embedding)):
+                force_fp32 = name in fp32_layers
+                wrap_module_precision(module, force_fp32, mixed_dtype)
+
+    def run_evaluation_one_batch(self, name, dataset_batch, use_dataloader_init=False):
```
pzelasko, line 611: This should also go to a dedicated inference script I think.

nemo/collections/speechlm2/modules/ear_tts_model.py
```diff
@@ -0,0 +1,1713 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
zhehuaichen, line 1: call this ear_tts_model.py to be consistent

nemo/collections/speechlm2/modules/ear_tts_vae_codec.py
```diff
@@ -0,0 +1,1067 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
zhehuaichen, line 1: call this ear_tts_vae.py to be consistent
meanwhile how about moving everything under vae to ear_tts_model.py ?

pzelasko (commented):
This is a huge PR! 
This is my high-level list of concerns:
* The introduction of PreTrainedModel - we already had `pretrained.py` for HF pretrained models, and I don't understand the purpose of these changes. If we merge as-is, the speechlm2 collection will become confusing. We need to unify this somehow.
* Some methods and functions are extremely long and I have no idea what they do in total. They can be improved in two ways: 1) split into smaller, descriptively named functions (or classes) with single responsibilities. 2) consider if some of the code can be re-written more tersely; I don't have a specific suggestion after quick review, but my gut's feeling is that it is possible to achieve.
* The inference logic is too scattered over the model code. I suggest moving it to a dedicated inference script, and not using `trainer.validate` to run it, look at `salm_generate.py` to get an idea.
* The data format really needs some documentation. I know it wasn't documented very well for earlier models either, but at this point even I have no idea how to construct data for this model.
* Talking about documentation, it should be updated with this PR's contributions (there is speechlm2 doc section in the repo). After it's written, it should be reviewed by some at least 1 other project contributor and 1 person that didn't work on this.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+                k: v.to(mixed_dtype) if isinstance(v, torch.Tensor) and v.is_floating_point() else v
+                for k, v in kwargs.items()
+            }
+            with ensures_target_precision(mixed_dtype):
+                return module._original_forward(*new_args, **new_kwargs)
+
+    module.forward = new_forward
+
+
+def find_sensitive_layers(
```
zhehuaichen: can we remove this in the code to be merged since we won't release a mixed precision model

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+        )
+
+    # Remove hooks
+    for h in hooks:
+        h.remove()
+
+    return stats
+
+
+def classify_precision_layers(model: nn.Module, stats: dict, safe_min: float, safe_max: float) -> list:
```
zhehuaichen: can we remove this in the code to be merged since we won't release a mixed precision model

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1991 @@
...
+    default_dtype = torch.get_default_dtype()
+    torch.set_default_dtype(target_dtype)
+    try:
+        with torch.amp.autocast(device_type="cuda" if torch.cuda.is_available() else "cpu", dtype=target_dtype):
+            yield
+    finally:
+        torch.set_default_dtype(default_dtype)
+
+
+def collect_activation_stats(model: nn.Module, inputs: dict) -> dict:
```
zhehuaichen: can we remove this in the code to be merged since we won't release a mixed precision model

nemo/collections/speechlm2/modules/ear_tts_model.py
```diff
@@ -0,0 +1,1713 @@
...
+    import triton.language as tl
+
+    TRITON_IMPORTED = True
+except ImportError:
+    TRITON_IMPORTED = False
+
+USE_TRITON = TRITON_IMPORTED and torch.cuda.is_available()
+logging.info("Triton available & CUDA detected. Using Triton kernel for batch_matmul.")
+
+if USE_TRITON:
```
zhehuaichen, line 110: do we use this TRITON code in the inference time? if not can we remove

nemo/collections/speechlm2/modules/ear_tts_vae_codec.py
```diff
@@ -0,0 +1,1067 @@
...
+        fmin=f_min,
+        fmax=f_max,
+        norm=norm,
+        htk=(mel_scale == "htk"),
+    )  # [n_mels, n_freqs]
+    fb = torch.from_numpy(fb).float()
+    return fb
+
+
+def mel_spectrogram(
```
zhehuaichen, line 349: can we reuse any of these from nemo codec code?

nemo/collections/speechlm2/modules/ear_tts_commons.py
Edresson: This is required because we maintain two codebases—the ADLR one and the NeMo one—and we need to be able to load the original ADLR checkpoint in NeMo (pretrained checkpoint). It enables us to load both the ADLR model and its configuration. It also lets us load the model’s codec; without this, the codec cannot be loaded.

nemo/collections/speechlm2/modules/ear_tts_commons.py
Edresson: The HFHubMixin class doesn’t support loading checkpoints from the ADLR codebase, which is why I kept this logic here. However, I agree that the current setup can be confusing. One option would be to move this functionality into the class itself and, instead of overloading from_pretrained, introduce a separate method for ADLR checkpoints—perhaps something like `from_pretrained_path`.

Right now, our training pipeline loads both a pretrained codec and a pretrained TTS model, and both rely on this class, so we need a clean way to support those ADLR checkpoints.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: It was to be compatible with the older codebase. Perhaps we can remove it now, given that everyone moved to the new speechlm2-based codebase.

examples/speechlm2/duplex_eartts_eval.py
Edresson: > Why are we using `.validate` and not `.predict` for inference?

That makes sense, I will change it.

nemo/collections/speechlm2/parts/metrics/bleu.py
Edresson: Good catch, fixed.

nemo/collections/speechlm2/parts/metrics/wer.py
Edresson: Yes, I think so. We weren’t using WER anywhere, and the intelligibility metric already includes both CER and WER.

Edresson: Is this being used by any other model in the NeMo main branch? I wasn’t able to find any. If it’s used internally, we should replace it with the Intelligibility metric, which computes both CER and WER and can also transcribe when needed.

tests/collections/speechlm2/test_duplex_eartts.py
Edresson: Done

tests/collections/speechlm2/test_duplex_eartts.py
Edresson: It is the standard in both test_duplex_s2s and test_duplex_s2s_speech_decoder. I tried to just follow it.

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: By default, we are using it for inference. Not sure if they will skip it or not in Niva.

nemo/collections/speechlm2/modules/ear_tts_vae_codec.py
Edresson: rvq_ear_tts_vae.py implements the EARTTS codec, so I don’t think merging it into the same file is a good idea. However, we can definitely rename it. How about ear_tts_codec.py or ear_tts_vae_codec.py?

nemo/collections/speechlm2/parts/metrics/secs.py
Edresson: Yes, we have something similar, but it was designed for a single-file workflow rather than the batched setup we have here, so I don’t think we can reuse it. For reference, this is what the TTS collection uses:
https://github.com/NVIDIA-NeMo/NeMo/blob/6389a71c10eb78db5224745ff1786954a9baea6a/scripts/magpietts/evaluate_generated_audio.py#L154

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: Done

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: This is fixed because the prompt is used only for model initialization. It simplifies model warm-up, and for speaker identity we likely don’t need more than about 3 seconds of audio. Using 20 seconds or more would only waste compute without providing additional benefit.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Fixed.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: We originally used it to fully leverage the ADLR pretrained checkpoint. However, due to limitations in the NanoV2 tokenizer, we are no longer using the system prompt. Therefore, I’ve removed it entirely from the NeMo implementation.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I completely removed descriptions, given that It was not used in your latest model.  So this ToDo was already removed.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Done I removed it.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Makes sense, good catch. I replaced it by:

```
        self.tokenizer.bos_token = self.cfg.get("bos_token", '<s>')
        self.tokenizer.eos_token = self.cfg.get("eos_token", '</s>')
        self.tokenizer.pad_token = self.cfg.get("pad_token", '<SPECIAL_12>')
```

defaulting it to NanoV2, which is our current release candidate. Please let me know if you have any better suggestion.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I removed this ToDo.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I removed it at this time. "custom sentences" are sentences that the user wants to generate as regular TTS task. It is usefull for debug, but not essential.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I think we should keep run_evaluation_one_batch as it is, since it's used for evaluating during training. Moving it to an inference script would prevent us from monitoring training progress (listening to the audios), which is essential for TTS training.

nemo/collections/speechlm2/modules/ear_tts_vae_codec.py
Edresson: Done, I adopted ear_tts_vae_codec.py as the file name. Let me know your thoughts on that.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Done removed.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Done

examples/speechlm2/duplex_eartts_eval.py
Edresson: I checked, and to use` .predict `we would also need to implement the predict_dataloader hook. I don’t think that’s necessary for our use case.

What we actually want in the` duplex_eartts_infer.py` script is to run evaluation on custom datasets and generate speech outputs. For that purpose, using validate is more appropriate.

To better reflect its purpose, I also renamed the file from `duplex_eartts_infer.py` to `duplex_eartts_eval.py`. Let me know if you’re okay with this change.

examples/speechlm2/duplex_eartts_eval.py
```diff
@@ -0,0 +1,63 @@
...
+from omegaconf import OmegaConf
+
+from nemo.collections.speechlm2 import DataModule, DuplexEARTTSDataset
+
+from nemo.collections.speechlm2.models.duplex_ear_tts import DuplexEARTTS
+from nemo.core.config import hydra_runner
+from nemo.utils.exp_manager import exp_manager
+from nemo.utils.trainer_utils import resolve_trainer_cfg
+
+torch.cuda.set_device(int(os.environ["LOCAL_RANK"]))
```
blisc, line 61: Is this necessary?  Same comment for train.py

examples/speechlm2/duplex_eartts_eval.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@hydra_runner(config_path="conf", config_name="s2s_duplex_speech_decoder")
+def inference(cfg):
+    OmegaConf.resolve(cfg)
+    torch.distributed.init_process_group(backend="nccl")
+    torch.set_float32_matmul_precision("medium")
+    torch.backends.cudnn.allow_tf32 = True
```
blisc, line 69: Can you comment why this part is needed? Same comment for train.py

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/parts/metrics/wer.py
Edresson: Although I did not find any model actually using it, I rewrite this metric back. We can remove it in a future PR.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I removed null_emb  and bos_emb, each is very small.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Good catch, done.

nemo/collections/speechlm2/parts/metrics/results_logger.py
```diff
@@ -0,0 +1,198 @@
...
+
+from nemo.collections.audio.parts.utils.resampling import resample
+from nemo.utils import logging
+
+
+def safe_remove_path(path):
+    shutil.rmtree(path, ignore_errors=True)
+
+
+class ResultsLogger:
```
blisc, line 54: Why is this part of the metrics folder?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,810 @@
...
+            segment = prompt_audio[b, start:length]
+
+            repeat_times = (n_sample + (length - start) - 1) // (length - start)
+            repeated = segment.repeat(repeat_times)[:n_sample]
+            out[b] = repeated
+
+    return out
+
+
+def get_mask_from_lengths(
```
blisc: Why not reuse nemo.collections.tts.parts.utils.helpers.get_mask_from_lengths?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,810 @@
...
+        audio_prompt, audio_prompt_lens = collate_random_turn_audio(
+            cuts.resample(target_sample_rate),
+            roles=roles,
+            recording_field=recording_field,
+        )
+
+    return audio_prompt, audio_prompt_lens
+
+
+def collate_random_turn_audio(
```
blisc, line 608: Can you docstring this function?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,810 @@
...
+            offset=max(0, selected_supervision.start), duration=selected_supervision.duration
+        ).load_custom(recording_field)
+
+        selected_turn_audios.append(truncated_audio.squeeze(0))
+        selected_turn_audios_lens.append(truncated_audio.shape[-1])
+
+    return collate_vectors(selected_turn_audios, padding_value=0), torch.tensor(selected_turn_audios_lens)
+
+
+def collate_token_channel(
```
blisc, line 659: Can you docstring this function?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,810 @@
...
+            add_text_bos_and_eos_in_each_turn=add_text_bos_and_eos_in_each_turn,
+        )
+        for c in cuts
+    ]
+    token_lens = torch.tensor([len(tt) for tt in tokens])
+    tokens = collate_vectors(tokens, padding_value=pad_id)
+    return tokens, token_lens
+
+
+def build_token_channel(
```
blisc, line 715: Can you docstring this function?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1522 @@
...
+
+    if torch.isin(speech_codes[:, :1], control_codes).any():
+        return torch.where(
+            torch.isin(speech_codes, control_codes), torch.zeros_like(speech_codes[:, :1]), speech_codes
+        )
+    else:
+        return torch.where(torch.isin(speech_codes, control_codes), speech_codes[:, :1], speech_codes)
+
+
+def get_mask_from_lengths(
```
blisc: Why redefine this get_mask_from_lengths function?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1522 @@
...
+    setup_rvq_audio_codec(self)
+    assert callable(self.tts_model.set_rvq_embs)
+    self.tts_model.set_rvq_embs(torch.stack([x.detach() for x in self.audio_codec.prvq.mus_list], 0))
+    self.tts_model.rvq_embs = self.tts_model.rvq_embs.to(next(self.tts_model.parameters()).dtype)
+    # compute target fps
+    self.target_fps = self.target_sample_rate / self.audio_codec.config.wav_to_token_ratio
+    self.target_samples_per_frame = self.audio_codec.config.wav_to_token_ratio
+
+
+def rescale_state_dict(state_dict, target_std=0.02, first_n_layers=None, layer_prefix="tts_model.backbone.layers."):
```
blisc: Do we still use this? If so, can we replace prints with logging. If not, let's remove

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1522 @@
...
+class DuplexEARTTS(LightningModule, HFHubMixin):
+    def __init__(self, cfg: dict) -> None:
+        assert isinstance(cfg, dict), (
+            "You must pass the config to DuplexEARTTS as a Python dict to support hyperparameter serialization "
+            f"in PTL checkpoints (we got: '{type(cfg)=}')."
+        )
+        super().__init__()
+        self.save_hyperparameters()
+        # convert dict to config
+        cfg = DictConfig(cfg)
```
blisc, line 67: Why do we pass a dict and forcibly convert it to a DictConfig rather than passing a DictConfig?

examples/speechlm2/duplex_eartts_eval.py
Edresson: Yes, it is, it is also the standard used in other train files on the Speechlm2 collection.

examples/speechlm2/duplex_eartts_eval.py
Edresson: Yes, it appears to be the case. This is also the standard used across other training files in the SpeechLM2 collection.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Good catch, done.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Good catch, this is no longer used in this file. I removed its definition.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Yes, we still rely on it to rescale the pretrained checkpoint weights to better FP16 training stability. I replaced the print with logging.info

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: This is the standard used in the SpeechLM2 collection, and I follow it to ensure consistency.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/parts/metrics/results_logger.py
Edresson: Do you have any suggestions on where to add it? It logs the metrics to files, saving audio and the textual output, so I though that keep it on metrics would be approprieted.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/parts/metrics/results_logger.py
blisc: Let's add a top-level comment that describes this file and how it differs from other files in this folder

Edresson: Done

nemo/collections/speechlm2/modules/ear_tts_commons.py
Edresson: @pzelasko I discussed this with @blisc, and we found a simpler approach. I removed this file and updated the DuplexEARTTS class to instantiate the model directly from the config, then manually load and restore the state dict when needed. With this change, the pretrained wrapper class and ear_tts_commons.py are no longer necessary, while we can still leverage the pretrained weights internally.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I restructured it. Please let me know your thoughts on it now.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: This function is used during validation and can also be called from an external script. However, for now, we are only planning to add the validation recipe, since the full inference pipeline (including integration with S2T) will be implemented in NIVA.

nemo/collections/speechlm2/modules/ear_tts_vae_codec.py
Edresson: I don't think so.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1492 @@
...
+            self.language_model = self._load_language_model(self.cfg)
+            self.embed_tokens = self._load_embed_tokens(self.cfg)
+            # delete llm because we use it only to get the  embbeding tokens
+            del self.language_model
+
+        # get codec run precision
+        self.audio_codec_run_dtype = getattr(torch, self.cfg.get("audio_codec_run_dtype", "float32"), torch.float32)
+
+        # instanciate eartts model and codec
+        self._load_tts_model(self.cfg)
```
blisc: - What is the difference between what happens inside of `_load_tts_model()` and `restore_from_pretrained_checkpoint()`? 
- Why do we want to do both of these instead of the model init function?

It seems cleaner if we separate the creation and the loading of a checkpoint. During init() ,we should simply instantiate the classes. After model init, but before training or inference, we should make a call to do checkpoint weight loading.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1492 @@
...
+        if model.cfg.get("pretrained_codec_model", None):
+            checkpoint_state = load_checkpoint(model.cfg.pretrained_codec_model)
+            checkpoint_state = set_model_dict_for_partial_init(checkpoint_state, model.audio_codec.state_dict())
+            model.audio_codec.load_state_dict(checkpoint_state, strict=True)
+
+    for p in model.audio_codec.parameters():
+        p.requires_grad = False
+
+
+def setup_audio_codec(self):
```
blisc: This function does too much. It serves 3 roles

- Model init
- Model checkpoint loading
- Ensuring that the dtype is correct

It would cleaner to separate these out. Right now, you call this function inside of `_load_tts_model`, and before you do training, and before you do eval. But each of these calls use only 1-2 of the 3 roles that `setup_audio_codec` does.
I would suggest that you refactor model init and checkpoint loading as in my other comment.
For the calls to this function before training and eval, it only needs to check that the codec is the correct dtype. So split out this function and only do the dtype check when necessary. Perhaps you might want to do further and do the dtype conversion elsewhere instead of inside of `on_{}_epoch_start()`

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Currently, `_load_tts_model()` is responsible for instantiating the TTS model from the config and, optionally, loading a TTS-only pretrained checkpoint if pretrained_tts_model is provided.
This checkpoint contains only the TTS weights and is not a `DuplexEARTTS `checkpoint. It comes from the ADLR codebase and is used strictly for pretraining initialization.

On the other hand, `restore_from_pretrained_checkpoint()` is used to partially or fully restore a `DuplexEARTTS `training checkpoint.  This also allows us to do the weights rescaling to be able to achieve better results on half precision.

So in short:

- _load_tts_model() → initializes the TTS module and loads a TTS-only pretrained model (for warm-starting).
- restore_from_pretrained_checkpoint() → restores a full training checkpoint from a previous experiment.

Regarding when this happens: both must be handled during model initialization.

If we separated checkpoint loading and did it later (e.g., after init()), we would risk overwriting resumed training weights in multi-run workflows (e.g., resumed Slurm jobs every 4 hours). In that setup, the restored checkpoint could be accidentally replaced by the pretrained weights, which is not what we want. This is why both mechanisms live in the initialization path, but serve distinct and non-overlapping purposes.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Yes, setup_audio_codec currently performs three roles. During initialization, we only need two of them: model instantiation and checkpoint loading. However, during training (e.g., in on_{train,val,test}_epoch_start()), we also need the third role: precision enforcement. If the precision changes, the codec must be reloaded in the target audio_codec_run_dtype.

@blisc and I had a short discussion and decided to split the logic into two functions:

`setup_audio_codec`, responsible only for instantiating the codec, loading the checkpoint, and registering embeddings.

`ensures_codec_target_dtype`, responsible for ensuring that the codec is running in the correct precision and reloading it if needed.

This separation keeps initialization clean while allowing safe precision correction during training without duplicating logic.

nemo/collections/speechlm2/modules/ear_tts_model.py
zhehuaichen: why do we need TRITON in training?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1509 @@
...
+def maybe_to(x, dtype):
+    if x is None:
+        return None
+    if isinstance(x, torch.Tensor) and torch.is_floating_point(x):
+        return x.to(dtype)
+    return x
+
+
+@contextmanager
+def ensures_target_precision(target_dtype):
```
zhehuaichen: do we still need this?

Edresson: Yes, we still need to ensure the codec is using the correct precision; otherwise, we won’t be able to train the model properly.

blisc (dismissed):

zhehuaichen (dismissed):

examples/speechlm2/duplex_eartts_eval.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
pzelasko, line 1: Let's add some documentation how to run the evaluation with this script. It should have:
* a general description of what is being done 
* a description of the main options to specify
* expected input data format
* output description
* an example of script invocation

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,873 @@
...
+                max_start = max(1, length - n_sample + 1)
+                start = torch.randint(0, max_start, (1,), device=device).item()
+            else:
+                # Deterministic: take from start
+                start = 0
+            out[b] = prompt_audio[b, start : start + n_sample]
+
+        else:
+            # Audio shorter than target → repeat
+            if sample:
```
pzelasko: why sample when example's length is shorter than requested? shouldn't we always use the entire prompt in this case (+ partially repeated)?

btw if the prompt audio is of zero length, what happens then? is this scenario actually supported or should this sample be discarded?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,873 @@
...
+            roles=self.input_roles,
+            add_text_bos_and_eos_in_each_turn=self.add_text_bos_and_eos_in_each_turn,
+        )
+
+        audio_prompt, audio_prompt_lens = get_audio_prompt(
+            cuts, self.target_sample_rate, roles=self.output_roles, recording_field="target_audio"
+        )
+
+        # ensures that input_text_tokens is not longer than its duration
+        input_text_tokens = input_text_tokens[:, : target_token_lens.max()]
```
pzelasko: Why is this truncation needed? Because of audio prompt? Shouldn't we increase target_token_lens instead to reflect the full length of the examples? We should avoid data loss at all costs and instead add support to bucketing length estimator to properly support this data format.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,873 @@
...
+        assert tokenizer.bos is not None, "BOS support in the tokenizer is required for S2S models."
+        assert tokenizer.eos is not None, "EOS support in the tokenizer is required for S2S models."
+
+    def __getitem__(self, cuts: CutSet) -> dict:
+        cuts = cuts.transform_text(_strip_timestamps)
+        source_audio, source_audio_lens = collate_audio(cuts.resample(self.source_sample_rate))
+        target_audio, target_audio_lens = collate_audio(
+            cuts.resample(self.target_sample_rate), recording_field="target_audio"
+        )
+        input_text_tokens, target_token_lens = collate_token_channel(
```
pzelasko: Should this be named `target_text_tokens` instead? It's confusing since one field is "input" and the other is "source"; let's decide on either source+target or input+output for consistency

nemo/collections/speechlm2/parts/metrics/asr_cer_wer.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
pzelasko, line 1: consider renaming this file to `intelligibility.py`

nemo/collections/speechlm2/modules/ear_tts_model.py
```diff
@@ -0,0 +1,1725 @@
...
+except ImportError:
+    TRITON_IMPORTED = False
+
+USE_TRITON = TRITON_IMPORTED and torch.cuda.is_available()
+
+if USE_TRITON:
+    logging.info("Triton+CUDA detected. Using Triton kernel for batch_matmul.")
+
+    @triton.jit
+    def batch_matmul_kernel(
```
pzelasko, line 114: This code looks duplicated (pasted twice)

nemo/collections/speechlm2/modules/ear_tts_model.py
```diff
@@ -0,0 +1,1725 @@
...
+        idx_sel = (
+            embs[i].pow(2).sum(-1)  # [g?, v]
+            - 2
+            * (r.unsqueeze(-2) @ embs[i].transpose(-1, -2)).squeeze(-2)  # [b, ?, g?, h] , [g?, h, v] -> [b, ?, g?, v]
+        ).argmin(-1)
+
+        emb_i = F.embedding(idx_sel, embs[i])
+        r = r - emb_i
+
+        # FIX: assign correctly without shape mismatch
```
pzelasko: What is the FIX about? Is it fixed or needs fixing?

nemo/collections/speechlm2/modules/ear_tts_model.py
```diff
@@ -0,0 +1,1725 @@
...
+        self.d_model = d_model
+
+        # Custom pad token for OOVs
+        self.pad_id = self.vocab_size
+        self.register_buffer("pad_tensor", torch.tensor(self.pad_id, dtype=torch.long))
+
+        # Identify BOS and EOS tokens (may be None)
+        tokens = [self.tokenizer.convert_ids_to_tokens(i) for i in range(self.vocab_size)]
+
+        if 'Qwen2.5' in model_name:
```
pzelasko: Can we avoid hardcoding Qwen2.5 name here? Also we should emit a warning about missing bos_token for other models too. 

As a proper fix we can either propagate the information about default bos/eos token from somewhere (upper level model, config, etc), or we can move this "if Qwen2.5" condition into AutoTokenizer definition (this way it will solve this missing bos/eos issue for all dependent models in one place). I think it is OK to hardcode such things in AutoTokenizer as an exception to the rule. WDYT?

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1570 @@
...
+from nemo.collections.speechlm2.parts.precision import fp32_precision
+from nemo.collections.speechlm2.parts.pretrained import (
+    load_checkpoint,
+    load_pretrained_hf,
+    set_model_dict_for_partial_init,
+)
+from nemo.utils import logging
+
+
+def load_audio_librosa(path, sr=None):
```
pzelasko: Please move all of the utility functions to the bottom of the file. The first thing we should see when opening a model file is the model class definition.

nemo/collections/speechlm2/models/duplex_ear_tts.py
pzelasko: PyTorch Lightning `save_hyperparameters` won't work with `DictConfig`; it must be `dict`. Unfortunate hack.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1570 @@
...
+        # get codec silence tokens
+        codec_silence_tokens = self.get_codec_silence_frame()
+        self.register_buffer("codec_silence_tokens", codec_silence_tokens)
+
+        # Load tokenizer
+        self.tokenizer = AutoTokenizer(
+            self.cfg.pretrained_lm_name, use_fast=True, trust_remote_code=True
+        )  # Note that we are using fast tokenizer
+
+        # set tokenizer special tokens
```
pzelasko: Ditto my comment above about moving bos/eos token setting for specific models that don't have then into AutoTokenizer itself. We could get rid of this code then.

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1570 @@
...
+
+        total_g_sq = 0.0
+        total_g_params = 0
+
+        for p in self.parameters():
+            if not p.requires_grad:
+                continue
+
+            # ----- weights -----
+            w = p.detach().cpu().float()  # ✅ safe offline copy
```
pzelasko: cam we remove emoji from comments? paranoia mode

nemo/collections/speechlm2/models/duplex_ear_tts.py
```diff
@@ -0,0 +1,1570 @@
...
+                        use_local_output=False,
+                    ),
+                )
+
+        if (dp_mesh := device_mesh["data_parallel"]).size() > 1:
+            assert dp_mesh.ndim == 1
+            self._use_fsdp = True
+
+            fsdp_config = {"mesh": dp_mesh}
+
```
pzelasko, line 1332: at the end of sharding layers/blocks of individual submodules, you may want to add
`self.llm = fully_shard(self.llm, **fsdp_config)`
`self.tts_model = fully_shard(self.tts_model, **fsdp_config)`
it will shard every parameter that was not included in the layer/blocks sharding further saving memory

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
```diff
@@ -0,0 +1,873 @@
...
+from lhotse.dataset.collation import collate_audio, collate_vectors
+from lhotse.utils import ifnone
+
+from nemo.collections.common.tokenizers import TokenizerSpec
+from nemo.collections.speechlm2.data.utils import get_pad_id
+from nemo.collections.tts.parts.utils.helpers import get_mask_from_lengths
+from nemo.utils import logging
+
+
+def sample_audio_segments_repeat(
```
pzelasko: Please move the helper functions at the bottom of this file, we should see the Dataset definition as the first thing when we open this. 

Also, since you have this nicely structured, it is now easy to test each individual operation in data prep. Can we add unit tests for them?

pzelasko (changes requested):
Thanks, this is a great improvement! I was able to understand more of the PR now and left some additional comments.

examples/speechlm2/duplex_eartts_eval.py
Edresson: @pzelasko  do you have any suggestion on where should we added this documentation?

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Good catch 🙂 — I’ve fixed it now.

Regarding the case where the prompt audio has zero length: in that situation, the prompt becomes an all-zero tensor with a length of n_sample. This should never happen in practice, since we always either sample a random turn from the target speaker or use context_audio as defined in the dataset itself.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Done

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Fixed

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Ok make sense, done.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Done

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I’m not sure I fully understand what you mean by “moving the BOS/EOS token setting to specific models.” Could you clarify?

Do you mean checking whether BOS and EOS tokens are already defined and, if not, avoiding setting them in the tokenizer and instead handling them inside the model class? If that’s the case, we would likely need to either add two new parameters to the dataset or pass the model itself as a dataset parameter.

I’m not sure this is the best approach, especially because it would break the existing convention in the SpeechLM2 collection, where we always pass only the tokenizer to the dataset class.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: Done

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: Good catch — this was an old comment that I forgot to remove. This part differs slightly from the ADLR EARTTS implementation because the previous version was raising errors during batched inference. I fixed the issue and left the comment for Jaehyeon to review the change. He has already done so, and I simply forgot to remove the comment afterward.

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: Good catch, I fixed it.

nemo/collections/speechlm2/parts/metrics/asr_cer_wer.py
Edresson: Originally, the file was named `intelligibility.py`, but @zhehuaichen suggested renaming it to` asr_cer_wer.py` to keep it consistent with the existing metrics file naming.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: Ok, for TTS task  `input_text_tokens ` make more sense than `target_text_tokens` it is why I was using it, but I agree that  source_text_tokens make things confusing.  I renamed it back to target_*.

nemo/collections/speechlm2/data/duplex_ear_tts_dataset.py
Edresson: It is actually not needed. I think I added it due some issue on the dataset but that was later fixed. I removed it.

nemo/collections/speechlm2/modules/ear_tts_commons.py
Edresson: It was removed in the latest version.

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: We actually dont use it on training. But we are using on inference time.

nemo/collections/speechlm2/modules/ear_tts_model.py
Edresson: Ok, so I have reused the NeMo AutoTokenizer instance tokenizer from the main model definition avoiding the need of it here.

nemo/collections/speechlm2/models/duplex_ear_tts.py
Edresson: I have moved this special tokens handling to the AutoTokenizer itself. It already has custom eos/bos/pad support. That way it is more clear.

examples/speechlm2/duplex_eartts_eval.py
Edresson: I added the documentation on the top of this file.

pzelasko (dismissed):

zhehuaichen (dismissed):

PR #15125, opened 2025-11-26
Per-Stream Phrase Boosting in ASR Decoding (Transducers)

artbataev:
> [!IMPORTANT]  

# What does this PR do ?

- Adds **per-stream (per-utterance) phrase boosting**, currently only for greedy label-looping decoding with transducers (RNN-T, TDT)
- Adds **multi-model for boosting** (slow reference `GPUBiasingMultiModelReference` and efficient fast `GPUBiasingMultiModel`), which can be further used with other decoders
- Enhances `BoostingTreeModelConfig` with `key_phrase_items_list` field to specify key phrases with per-phrase options (currently - allows to specify per-phrase lang to use with aggregate tokenizers). NB: it is impossible to mix tuples with strings in `key_phrases` due to `omegaconf` restrictions on using `Union` types (boosting config is used with omegaconf/hydra). 

Known issues: when creating many boosting models, OneLogger can cause `OSError: Too many open files`. There is currently no other workaround rather than increasing this limit.

## Performance

Per-stream biasing testing on internal dataset.
- model: nvidia/stt_en_fastconformer_transducer_large
- inference: NeMo inference (default options) CUDA graphs without while-loops
- batch_size: 128
- greedy decoding with boost weight = 1.0 (when applicable)

Scenarios:
- no biasing
- default shared biasing (previously implemented)
- per-stream biasing: use same 200 phrases, but separately construct model for each utterance
- per-stream biasing: boosting ground truth (sanity check: boosting ground truth should result in low WER)

| Per-stream biasing | Cache   | Biasing                   | WER↓   | RTFx↑      | Comment                                                           |
|--------------------|---------|---------------------------|-------|-----------|-------------------------------------------------------------------|
| -                  | -       | -                         | 14.72 | **2243**  | baseline                                                          |
| enabled            | -       | -                         | 14.72 | 2207      | decoding implementation overhead, <2%                             |
| no                 | -       | 200 phrases               | 13.87 | **2185**  | baseline 2: default shared biasing                                |
| enabled            | no      | 200 phrases per utterance | 13.87 | 1136      | same WER as baseline 2                                            |
| enabled            | disk    | 200 phrases per utterance | 13.87 | 1395	     | loading from disk is faster than building                         | 
| enabled            | memory  | 200 phrases per utterance | 13.87 | 2070      | ~5% overhead compared to shared boosting tree                     |
| enabled            | decoder | 200 phrases per utterance | 13.87 | **2169**	 | RTFx similar to baseline 2: nearly zero pure multi-model overhead |
| enabled            | no      | ground truth              | 5.07  | 1501      | sanity check: boosting ground truth should result in low WER      |
| enabled            | disk    | ground truth              | 5.07  | 1359      | loading from disk is slower than building (tiny boosting models)  |
| enabled            | memory  | ground truth              | 5.07  | 2067      | ~5% overhead compared to shared boosting tree                     |
| enabled            | decoder | ground truth              | 5.07  | **2104**  | RTFx similar to baseline 2, nearly zero pure multi-model overhead |

## Caching

Per-utterance biasing works in the following way:
- biasing model is compiled when the stream starts
- biasing model added to the multi-model owned by the decoder (after compiling)
- biasing model is removed after the stream ended

Since the conversion from list of phrases to the graph (neural module) imposes overhead, there are caching mechanisms available:
- Disk: if the `BoostingTreeModelConfig.model_path` is set, model will be loaded from this path. If the path does not exist, model will be compiled and then saved to this path (next time it will be loaded from the disk)
- Memory: if `BiasingRequestItemConfig.cache_key` is a non-empty string, the compiled model will be loaded from cache in memory. If the cache entry does not exist, the model will be stored by this key
- Decoder: if `BiasingRequestItemConfig.auto_manage_multi_model` is set to `False` (`True` by default), the model can be manually stored in the multi-model owned by the decoder (and also should be removed manually). It is not recommended, but still possible.

NB: Cache keys (`model_path` and `cache_key`), if used, should be unique for `(tokenizer, phrases)` (or `(asr_model, phrases)`). 

**Collection**: [ASR]

# Changelog

# Usage

Users can use per-utterance phrase boosting:
- with the following scripts by specifying `"biasing_request"` field in the manifest
  - `examples/asr/asr_chunked_inference/rnnt/speech_to_text_streaming_infer_rnnt.py` with `use_per_stream_biasing=true` option
  - `examples/asr/asr_streaming_inference/asr_streaming_infer.py`
- in code
  - with `asr_model.transcribe`
  - with a streaming inference pipeline by specifying `ASRRequestOptions`

Manifest example:
```jsonl
{"audio_filepath": "/data/file1.wav", "text": "some ground truth text", "biasing_request": {"boosting_model_cfg": {"key_phrases_list": ["one phrase"]}}}
{"audio_filepath": "/data/file2.wav", "text": "another ground truth text", "biasing_request": {"boosting_model_cfg": {"key_phrases_list": ["other phrases", "and this one"]}}}
```

Code examples (`asr_model.transcribe` + inference pipeline):
```python
from omegaconf import OmegaConf, open_dict

from nemo.collections.asr.inference.factory.pipeline_builder import PipelineBuilder
from nemo.collections.asr.inference.streaming.framing.request_options import ASRRequestOptions
from nemo.collections.asr.inference.utils.progressbar import TQDMProgressBar
from nemo.collections.asr.models import EncDecRNNTBPEModel
from nemo.collections.asr.parts.context_biasing.biasing_multi_model import BiasingRequestItemConfig
from nemo.collections.asr.parts.context_biasing.boosting_graph_batched import BoostingTreeModelConfig
from nemo.collections.asr.parts.utils.rnnt_utils import Hypothesis
from nemo.collections.asr.parts.utils.transcribe_utils import get_auto_inference_device

USE_TRANSCRIBE = False  # toggle inference pipelines

if __name__ == "__main__":
    model_name = "nvidia/parakeet-rnnt-1.1b"
    device = get_auto_inference_device()

    audio_files = [
        "file1.wav",
        "file2.wav",
        "file3.wav",
    ]
    biasing_requests = [
        BiasingRequestItemConfig(
            boosting_model_cfg=BoostingTreeModelConfig(
                key_phrases_list=["one phrase"],
            ),
            boosting_model_alpha=2.0,
        ),
        None, # no biasing
        BiasingRequestItemConfig(
            boosting_model_cfg=BoostingTreeModelConfig(
                key_phrases_list=["other phrases", "and this one"],
            ),
            boosting_model_alpha=1.0,
        ),
    ]

    if USE_TRANSCRIBE:
        # transcribe example
        asr_model: EncDecRNNTBPEModel = EncDecRNNTBPEModel.from_pretrained(model_name=model_name)
        asr_model.to(device)

        # Change Decoding Config: ensure greedy_batch + label-looping enabled
        with open_dict(asr_model.cfg.decoding):
            asr_model.cfg.decoding.strategy = "greedy_batch"
            asr_model.cfg.decoding.greedy.loop_labels = True
            asr_model.cfg.decoding.greedy.enable_per_stream_biasing = True
        asr_model.change_decoding_strategy(asr_model.cfg.decoding)

        # pass partial hypotheses initialized with boosting requests to transcribe
        results = asr_model.transcribe(
            audio=audio_files,
            partial_hypothesis=[
                (
                    Hypothesis.empty_with_biasing_cfg(
                        biasing_cfg=request,
                    )
                    if request is not None
                    else None
                )
                for request in biasing_requests
            ],
            return_hypotheses=True,
        )
        print([result.text for result in results])
    else:
        # new streaming inference example
        cfg = OmegaConf.load("examples/asr/conf/asr_streaming_inference/buffered_rnnt.yaml")
        cfg.asr.model_name = model_name
        cfg.asr.device = device.type
        cfg.asr.decoding.greedy.enable_per_stream_biasing = True
        pipeline = PipelineBuilder.build_pipeline(cfg)
        progress_bar = TQDMProgressBar()
        options = [
            (
                ASRRequestOptions(
                    biasing_cfg=request,
                )
                if request is not None
                else ASRRequestOptions()
            )
            for request in biasing_requests
        ]
        output = pipeline.run(audio_files, options=options, progress_bar=progress_bar)
        for i in range(len(output)):
            print(f"{i}: {output[i]['text']}")
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/asr/asr_streaming_inference/asr_streaming_infer.py
+++ b/examples/asr/asr_streaming_inference/asr_streaming_infer.py
@@ -50,7 +50,6 @@ from nemo.collections.asr.inference.streaming.framing.request_options import ASR
 from nemo.collections.asr.inference.utils.manifest_io import calculate_duration, dump_output, get_audio_filepaths
 from nemo.collections.asr.inference.utils.pipeline_eval import calculate_pipeline_laal, evaluate_pipeline
 from nemo.collections.asr.inference.utils.progressbar import TQDMProgressBar
-from nemo.collections.asr.metrics.wer import word_error_rate
 from nemo.collections.asr.parts.context_biasing.biasing_multi_model import BiasingRequestItemConfig
 from nemo.utils import logging
 from nemo.utils.timers import SimpleTimer


nemo/collections/asr/inference/streaming/state/state.py
```diff
@@ -34,6 +34,8 @@ class StreamingState:
     Generic state for the streaming ASR pipeline
     """
 
+    options: RequestOptions | None
```
naymaraq, line 38: Why is `options` declared here as a class member? At line 109, it was declared as an instance member.

nemo/collections/asr/inference/streaming/framing/request_options.py
```diff
@@ -76,7 +78,11 @@ def augment_with_defaults(
             asr_output_granularity=(
                 default_asr_output_granularity if self.asr_output_granularity is None else self.asr_output_granularity
             ),
+            biasing_cfg=self.biasing_cfg,
```
naymaraq: Just a suggestion:
Sometimes we would like to use the same biasing list for all streams. Usually, in such cases, I am using default values, unless they are overridden by request-specific options

nemo/collections/asr/inference/pipelines/buffered_rnnt_pipeline.py
```diff
@@ -463,29 +465,58 @@ def stateful_transcribe_step(
...
             with torch.inference_mode(), torch.no_grad():
                 best_hyp_chunk, alignments, batched_state = self.decoding_computer(
-                    encs.transpose(1, 2), enc_lens_chunk, batched_rnnt_states
+                    encs.transpose(1, 2),
+                    enc_lens_chunk,
+                    batched_rnnt_states,
+                    multi_biasing_ids=multi_biasing_ids,
                 )
-
+        # TODO: remove double-decoding
```
naymaraq: We perform double decoding: one of them is used to update the state properly, and the other is used to extract the tokens through the end of the buffer.

If you have ideas on how we could do this in a single decoding step, that would be awesome.

nemo/collections/asr/inference/streaming/state/state.py
artbataev: This is only a type annotation (not a declaration), which is fully valid in Python (and actually a recommended way to do annotations for classes).
See, for example, https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#classes (`AuditedBankAccount`, `Car`) in examples.

nemo/collections/asr/inference/streaming/framing/request_options.py
artbataev: Currently, when we have a boosting request, we need to construct a tree, add it to a multi-boosting-model in the decoder (both operations require significant amount of resources). After the request, we remove the boosting model from the multi-model.

There are several options to handle the case, in which some of the boosting configs are shared.
1) We instantiate boosting model using *the decoding config* (boosting model will be applied to all streams regardless requested options, boosting tree in the options can be applied in addition to the default tree). It is currently implemented (by using decoding options).
2) Manually manage all boosting requests: construct trees, add to the multi-model, remove from the multi-model. It is possible (even with the current implementation), but in this case the client app should manage these requests.
3) Manually manage only one the default model constructed from default options (see below my suggestion).

Is it a good idea to add an additional section to the config like
```yaml
asr:
  default_boosting_options:
      boosting_tree:
        model_path: null  # The path to built '.nemo' boosting tree model
        key_phrases_file: null  # The path to the context-biasing list file (one phrase per line)
        key_phrases_list: null  # The list of context-biasing phrases ['word1', 'word2', 'word3', ...]
        source_lang: "en"  # The source language of the context-biasing phrases (for aggregate tokenizer)
      boosting_tree_alpha: 0.0
```
and by making `biasing_cfg` optional (can be None), assign default boosting options to the current stream (if None)?

nemo/collections/asr/inference/pipelines/buffered_rnnt_pipeline.py
artbataev: Yep, this TODO is for myself. Unsure if I will implement it in the current PR, but I will try :)

nemo/collections/asr/inference/streaming/state/state.py
naymaraq: I see. We should do the same for the other members as well, but that can be done later during the refactoring stage.

nemo/collections/asr/inference/streaming/framing/request_options.py
naymaraq: Creating a new default section is probably too heavy, so we should leave it as is. Besides, we can use the `asr.decoding.greedy.boosting_tree` section if we want to apply the same boosting tree to all streams.

examples/asr/conf/asr_streaming_inference/buffered_rnnt.yaml
```diff
@@ -22,7 +23,11 @@ asr:
         model_path: null  # The path to built '.nemo' boosting tree model
         key_phrases_file: null  # The path to the context-biasing list file (one phrase per line)
         key_phrases_list: null  # The list of context-biasing phrases ['word1', 'word2', 'word3', ...]
-        source_lang: "en"  # The source language of the context-biasing phrases (for aggregate tokenizer)
+        key_phrase_items_list: null  # The list of context-biasing phrases with custom fields
+        # in CLI: [{phrase:"word1",lang:en},{phrase:"frase dos",lang:es}]
+        # in code: [PhraseItem(phrase="word1, lang="en"), PhraseItem(phrase2="frase dos", lang="es")]
+        source_lang: "en"  # The source language of the context-biasing phrases (for aggregate tokenizer),
```
lilithgrigoryan: Should we maintain a list of source languages to handle cases where context-biasing lists are in different languages?

nemo/collections/asr/parts/context_biasing/biasing_multi_model.py
```diff
@@ -0,0 +1,709 @@
...
+        self.model2states_offset = nn.Buffer(torch.zeros([self.num_models_reserved], dtype=torch.int64))
+        self.model2arcs_offset = nn.Buffer(torch.zeros([self.num_models_reserved], dtype=torch.int64))
+
+        self.num_states_total = 0
+        self.num_arcs_extended_total = 0  # + extra padding
+        self.num_states_reserved = self.INIT_NUM_STATES
+        self.num_arcs_extended_reserved = self.INIT_NUM_ARCS  # + extra padding
+
+        # arcs-related data
+        self.all_arcs_weights = nn.Parameter(torch.zeros([self.num_arcs_extended_reserved]))
```
andrusenkoau, line 308: Do you have a plan to fine-tune the arc weights for a boosting tree in the future?

nemo/collections/asr/parts/context_biasing/biasing_multi_model.py
```diff
@@ -0,0 +1,715 @@
...
+        accumulated_backoff = torch.zeros(batch_size, device=device)
+        # loop condition
+        start_state_not_processed = model_ids != -1
+
+        states_offsets = self.model2states_offset[model_ids]
+        arcs_offsets = self.model2arcs_offset[model_ids]
+
+        num_iterations = 0
+        while start_state_not_processed.any():
+            # assert num_iterations <= self.max_order, "Infinite loop in LM advance"
```
lilithgrigoryan: Do we need this assertion here?

nemo/collections/asr/parts/context_biasing/biasing_multi_model.py
```diff
@@ -0,0 +1,715 @@
...
+    def remove_from_multi_model(self, biasing_multi_model: "GPUBiasingMultiModelBase"):
+        """Remove biasing model from multi-model"""
+        if self.multi_model_id is None:
+            # nothing to remove
+            return
+        biasing_multi_model.remove_model(self.multi_model_id)
+        self.multi_model_id = None
+
+
+class GPUBiasingMultiModelBase(abc.ABC, nn.Module):
```
lilithgrigoryan, line 95: Please consider adding docstring to this class

artbataev: Done

nemo/collections/asr/parts/context_biasing/biasing_multi_model.py
artbataev: `.max_order` property is not maintained for multi-model. No need for this check here in runtime (this was useful for debugging original models). I removed it.

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
```diff
@@ -62,7 +76,12 @@ class BoostingTreeModelConfig:
 
     @staticmethod
     def is_empty(cfg: "BoostingTreeModelConfig") -> bool:
-        return cfg.model_path is None and cfg.key_phrases_file is None and cfg.key_phrases_list is None
+        return (
+            cfg.model_path is None
```
andrusenkoau, line 80: Is it possible to pass an empty string for `cfg.model_path` and `cfg.key_phrases_file`? Maybe we can use the same `not ...` condition here?

nemo/collections/asr/parts/context_biasing/biasing_multi_model.py
artbataev: If we need it, minor adjustments needed. Keeping here `nn.Parameter` for consistency with original NGPU-LM modules.

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
artbataev: Let's do this in the follow-up PR ;)

PR #15139, opened 2025-12-03
Support inference without context audio or ground truth audio

rfejgin:
In some cases we need to run inference on manifests that do not include context audio and/or ground truth audio:
* Text context manifests may not have a context audio included (and it wouldn't be very relevant anyway)
* When generating from arbitrary text the ground truth audio cannot be assumed to be available.

Note that even without the context and ground truth there are useful metrics that can be calculated like WER, UTMOS, and inference speed.

This PR adapts `infer_and_evaluate.py` and dependent scripts to allow the absence of context and/or ground truth audio.


blisc:
Closing, please make a new PR into main

PR #15158, opened 2025-12-08
[TTS] MagpieTTS Inference Changes

subhankar-ghosh:
> [!IMPORTANT]  

# What does this PR do ?

TTS

# Changelog

1) Refactored Inference script
a) Modularized code
b) Separated inference and evaluation logic
c) Separated different configs
d) Lays the groundwork for longform inference

2) infer_batch returns dataclass object now

3) Moved inference code from scripts to examples

4) Added fixed context embeddings instead of getting it through context encoder, in case checkpoint does not have context encoder.

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/tts/models/magpietts.py
```diff
@@ -322,6 +353,10 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             self.transcript_decoder_layers = [
                 idx for idx in range(cfg.decoder.n_layers)
             ]  # All layers are used for text
+            # Register buffers for baked context embedding (initially None/empty)
+            # These will be populated when loading a checkpoint with baked embedding
+            self.register_buffer('baked_context_embedding', None)
+            self.register_buffer('baked_context_embedding_len', None)
```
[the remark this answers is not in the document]
subhankar-ghosh: @copilot check now.

rfejgin:
Just started reviewing, but I am liking the modularization!

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,152 @@
...
+        with_utmosv2: Whether to compute UTMOSv2 (Mean Opinion Score) metrics.
+    """
+
+    sv_model: str = "titanet"
+    asr_model_name: str = "nvidia/parakeet-tdt-1.1b"
+    language: str = "en"
+    with_utmosv2: bool = True
+
+
+def evaluate_generated_audio_batch(
```
rfejgin: @subhankar-ghosh Not sure if "audio_batch" is the right term here as it could be confused with a GPU batch, but really we mean a whole directory.

Maybe `evaluate_generated_audio_dir` ?

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+        Tuple of (loaded model, checkpoint name for output labeling).
+
+    Raises:
+        ValueError: If configuration is invalid or sample rates don't match.
+    """
+    config.validate()
+
+    if config.hparams_file is not None and config.checkpoint_file is not None:
+        # Mode 1: Load from hparams + checkpoint
+        model_cfg = OmegaConf.load(config.hparams_file)
```
rfejgin, line 204: Might make sense to add a logging statement here too, to be consistent with line 233 below:
https://github.com/NVIDIA-NeMo/NeMo/pull/15158/files#diff-bfc74531da7bd8b29d57622d8b001580302fd277eae5f2ad64e955ef850dcfbfR233

nemo/collections/tts/models/magpietts.py
```diff
@@ -1460,14 +1540,29 @@ def prepare_context_tensors(self, batch):
...
                     context_embeddings = context_input_embedded
                 elif self.model_type == 'decoder_ce':
-                    context_embeddings = self.context_encoder(
-                        context_input_embedded, context_mask, cond=None, cond_mask=None
-                    )['output']
+                    # Check for baked context embedding first
+                    if self.has_baked_context_embedding:
+                        # self.baked_context_embedding is a fixed context embedding that is baked into the model.
+                        # This is used when we do not want users to generate speech with context audio or context text.
+                        # This is done to disable zero-shot inference. Users can only generate speech in 1 voice chosen
```
rfejgin: Not sure if we want to ever release more than 1 voice, if so we'd need a way to choose between baked embeddings (possibly at runtime?).

nemo/collections/tts/modules/magpietts_inference/evalset_config.py
```diff
@@ -11,9 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""
-Used as a datafile for infer_and_evaluate.py
-"""
 dataset_meta_info = {
```
rlangman: Should we replace `evalset_config.py` with a JSON file? It does not really make sense to open source code and config containing several datasets that only we have access to.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,152 @@
...
+        std_err = stats.sem(measurements)
+
+        # t-distribution critical value for confidence interval
+        t_critical = stats.t.ppf((1 + confidence) / 2, len(measurements) - 1)
+        ci = std_err * t_critical
+
+        logger.info(f"{key}: {mean:.4f} ± {ci:.4f}")
+        results[key] = f"{mean:.4f} ± {ci:.4f}"
+
+    return results
```
rlangman: A function like this should probably return (mean, ci) as a tuple. Logging the value and formatting it as a string should be done by the higher-level script.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,152 @@
...
+
+def compute_mean_with_confidence_interval(
+    metrics_list: List[dict],
+    metric_keys: List[str],
+    confidence: float = 0.95,
+) -> Dict[str, str]:
+    """Compute mean and confidence interval for specified metrics.
+
+    Uses Student's t-distribution for confidence interval calculation,
+    which is appropriate for small sample sizes.
```
rlangman: I would remove a questionable statement like "which is appropriate for small sample sizes."

T-test is appropriate when looking at independent, continuous, metrics that roughly follow a normal distribution. It cannot be used at all for non-continuous metrics like WER, CER. At some point we should implement bootstrap confidence intervals for these.

examples/tts/magpietts/evaluate_generated_audio.py
```diff
@@ -25,9 +25,9 @@
 from contextlib import contextmanager
 from functools import partial
 
+import examples.tts.magpietts.evalset_config as evalset_config
 import librosa
 import numpy as np
-import scripts.magpietts.evalset_config as evalset_config
```
rlangman: We might want to avoid importing things that are in examples/scripts directory, as it will not work unless your current working directory is the NeMo codebase. Reusable code should usually be placed somewhere in collections so it gets installed with the NeMo codebase.

nemo/collections/tts/models/magpietts.py
```diff
@@ -53,6 +54,36 @@
 from nemo.utils import logging
 
 
+@dataclass
+class InferBatchOutput:
```
rlangman, line 58: Should we also make an `InferBatchInput` dataclass?

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+        nemo_file: Path to the .nemo archive (alternative to hparams + checkpoint).
+        codecmodel_path: Path to the audio codec model.
+        legacy_codebooks: Use legacy codebook indices for old checkpoints.
+        legacy_text_conditioning: Use legacy text conditioning for old checkpoints.
+        hparams_from_wandb: Whether hparams file is from wandb export.
+    """
+
+    hparams_file: Optional[str] = None
+    checkpoint_file: Optional[str] = None
+    nemo_file: Optional[str] = None
```
rlangman, line 53: Is there a need to support both` hprams + ckpt` and `.nemo` file loading? Can we standardize on 1 method?

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+    Returns:
+        Tuple of (updated config, sample_rate from config if present).
+    """
+    model_cfg.codecmodel_path = codecmodel_path
+
+    # Update text tokenizer paths for backward compatibility
+    if hasattr(model_cfg, 'text_tokenizer'):
+        model_cfg.text_tokenizer.g2p.phoneme_dict = "scripts/tts_dataset_files/ipa_cmudict-0.7b_nv23.01.txt"
+        model_cfg.text_tokenizer.g2p.heteronyms = "scripts/tts_dataset_files/heteronyms-052722"
+        model_cfg.text_tokenizer.g2p.phoneme_probability = 1.0
```
rlangman, line 103: Does this work with multilingual checkpoints? I had to update it locally to handle each language tokenizer separately.

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,603 @@
...
+
+    Returns:
+        Tuple of (mean CER across datasets, mean SSIM across datasets).
+    """
+    if violin_plot_metrics is None:
+        violin_plot_metrics = list(DEFAULT_VIOLIN_METRICS)
+
+    # Remove UTMOSv2 from plots if disabled
+    if not eval_config.with_utmosv2 and 'utmosv2' in violin_plot_metrics:
+        violin_plot_metrics.remove('utmosv2')
```
rlangman, line 168: One would expect that violin_plot_metrics being None would skip the violin plot creation. It might be clearer if `violin_plot_metrics` is a required argument and the caller of this method is expected to provide the values/default.

nemo/collections/tts/models/magpietts.py
rlangman: We should be able to support multiple voices. I propose the following refactor:
(1) `infer_batch` is modified to take context embedding as input argument
(2) Context embeddings can be obtained normally through a separate inference method (which throws an error if we want it disabled/"baked").
(3) Another method is added which allows user to obtain context embeddings for a specific voice (e.g. `model.get_voice_embeddings(['speaker1', 'speaker2']`). The weights for the voices can be hardcoded into the checkpoint or provided in a file.

nemo/collections/tts/modules/magpietts_inference/evalset_config.py
subhankar-ghosh: We can but we need this for CICD as well. I think it is ok to keep only these few datasets. They are public data anyways.

examples/tts/magpietts/evaluate_generated_audio.py
subhankar-ghosh: @rlangman Are you suggesting we move the entire examples/tts/magpietts folder to nemo/collections/tts ?

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: I do not think we should give the users the methods to generate context embeddings and pass it to infer_batch, because we are doing this baked_embedding thing to disable zeroshot and even generate any voice other than the baked in voice. I think we should always bake in these embeddings in the model weights. I know we have removed context encoder, but still I would not want to make it easy for the users.

subhankar-ghosh: I think we can add support for multiple baked in embeddings instead of just one. Happy to discuss.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
subhankar-ghosh: Done

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,603 @@
...
+        apply_prior_to_layers=parse_layer_list(args.apply_prior_to_layers),
+        start_prior_after_n_audio_steps=args.start_prior_after_n_audio_steps,
+        use_local_transformer=args.use_local_transformer,
+        maskgit_n_steps=args.maskgit_n_steps,
+        maskgit_noise_scale=args.maskgit_noise_scale,
+        maskgit_fixed_schedule=args.maskgit_fixed_schedule,
+        maskgit_sampling_type=args.maskgit_sampling_type,
+        eos_detection_method=args.eos_detection_method,
+        ignore_finished_sentence_tracking=args.ignore_finished_sentence_tracking,
+    )
```
XuesongYang, line 534: We can simplify the `dataclass` initialization by defining a new class function as `from_dict(cls, config)` and do `inference_config = InferenceConfig.from_dict(args)`

**Demo Code**
```python
from dataclasses import dataclass, fields

@dataclass
class InferenceConfig:
    ...

    @classmethod
    def from_dict(cls, config):
        field_names = {f.name for f in fields(cls)}
        config_new = {}
        for k, v in config.items():
            if k in field_names:
                if k in ["estimate_alignment_from_layers", "apply_prior_to_layers"]:
                    v = parse_layer_list(v) 
                config_new[k] = v
        return cls(**config_new)

inference_config = InferenceConfig.from_dict(vars(args))
```

nemo/collections/tts/modules/magpietts_inference/evalset_config.py
XuesongYang: it seems our CICD only needs `an4_val_ci`, so I bet it is safe to remove all dataset alias with prefixes `riva_*`.

Alternatively, why bother `riva`? instead, we can keep all, but remove the prefixes "riva_" for all datasets.

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,603 @@
...
+# Import the modular components
+from examples.tts.magpietts.evaluation import (
+    DEFAULT_VIOLIN_METRICS,
+    STANDARD_METRIC_KEYS,
+    EvaluationConfig,
+    compute_mean_with_confidence_interval,
+    evaluate_generated_audio_batch,
+)
+from examples.tts.magpietts.inference import InferenceConfig, MagpieInferenceRunner
+from examples.tts.magpietts.utils import ModelLoadConfig, get_experiment_name_from_checkpoint_path, load_magpie_model
```
XuesongYang: better not to nest `utils` inside `examples.tts.magpeitts`. I would suggest move to `nemo/collections/tts/parts/utils/eval_utils.py` similar to what ASR did as in `nemo/collections/asr/parts/utils/eval_utils.py`

XuesongYang: Similar comment: move `eamples.tts.magpietts.visualization` to `nemo/collections/tts/parts/utils/visualization.py`, or `nemo/collections/tts/parts/utils/helpers.py`

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,603 @@
...
+from examples.tts.magpietts.visualization import create_combined_box_plot, create_violin_plot
+
+from nemo.collections.asr.parts.utils.manifest_utils import read_manifest
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+)
+logger = logging.getLogger(__name__)
```
XuesongYang, line 75: NeMo has a logging wrapper. Please check `nemo/utils/nemo_logging.py` for reusing the codes.

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,603 @@
...
+                dataset=test_dataset,
+                output_dir=repeat_audio_dir,
+                manifest_records=manifest_records,
+                audio_base_dir=meta['audio_dir'],
+                save_cross_attention_maps=True,
+                save_context_audio=(repeat_idx == 0),  # Only save context audio once
+            )
+
+            # Compute mean RTF metrics
+            mean_rtf = runner.compute_mean_rtf_metrics(rtf_metrics_list)
```
XuesongYang, line 251: for official report, we'd better to compute P50/P90/P99 latency percentiles. Only reporting mean value is not sufficient for model comparison.

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+                "Both checkpoint mode and nemo_file provided. Using checkpoint mode (hparams_file + checkpoint_file)."
+            )
+
+
+def update_config_for_inference(
+    model_cfg: DictConfig,
+    codecmodel_path: Optional[str],
+    legacy_codebooks: bool = False,
+    legacy_text_conditioning: bool = False,
+) -> Tuple[DictConfig, Optional[int]]:
```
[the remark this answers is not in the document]
subhankar-ghosh: Added docstring to show it updates as well as returns.

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+    """Update model configuration for inference, handling backward compatibility.
+
+    This function transforms legacy configuration options to their modern equivalents
+    and disables training-specific settings. The function updates the model configuration in place and also returns.
```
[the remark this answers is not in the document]
subhankar-ghosh: The docstring looks good to me. Won't fix.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,149 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Evaluation wrapper for MagpieTTS generated audio.
+
+This module provides a clean interface to the evaluation functionality,
+wrapping the existing `examples.tts.magpietts.evaluate_generated_audio` module.
```
XuesongYang, line 18: module path typo: `nemo.collections.ttx.modules.magpietts_inference.evaluate_generated_audio`.

rfejgin (approved):
Looks good to me, under the assumption that we are okay with just one voice being supported by the pre-baked embedding mechanism.

nemo/collections/tts/models/magpietts.py
rlangman: Randomizing the weights in the context encoder should be enough to "disable zeroshot". We should not be unnecessarily complicating the code. There should also be inference code so the model is functional if someone trains it from scratch.

The code as is too unmaintainable for me to approve. If you don't want to use the approach I suggested, you could also e.g. initialize the baked embeddings to zeros (before loading from checkpoint), have the released context encoder weights randomized, and then have inference code paths which allow a user to utilize both paths seamlessly (while modifying it somehow to support multiple speakers). If that is also not okay, then I think we need a design discussion.

nemo/collections/tts/modules/magpietts_inference/evalset_config.py
rlangman: Yes, I was referring to all of the config here and in the EVALUATION_DATASETS with "riva" in it which are not public. This should not be in the open source code.

examples/tts/magpietts/evaluate_generated_audio.py
rlangman: The top-level scripts for doing inference should be in `examples/` or `scripts/`. But ideally scripts should be self-contained, meaning they only depend on the code inside the script itself and modules/utilities that can be imported from nemo.

nemo/collections/tts/models/magpietts_preference_optimization.py
artem-gorodetskii, line 1: Suggestion: Instead of setting the private attribute _no_state_dict directly in MagpieTTSModelOnlinePO, it’s better to define a property (e.g., no_state_dict) in MagpieTTSModel

nemo/collections/tts/models/magpietts.py
artem-gorodetskii, line 1: Suggestion: remove unused debug prints from the `infer_batch` method.

```python
# print(f"step {idx}")
# print(f"use_cfg {use_cfg}")
# print(f"shape {cfg_audio_codes_embedded.shape}")
# print(f"use kv cahce? {self.use_kv_cache_for_inference}")
```

nemo/collections/tts/modules/magpietts_inference/evalset_config.py
blisc: I agree with Ryan's comment. Can we refactor the current code to remove this file? The suggestion of a .json makes sense, we should store this json in the same location that we store our evaluation datasets and try to version it via git.
As for the an4 ci data, we can keep this in the repo as an example of what such a .json file would look like.

nemo/collections/tts/models/magpietts.py
shehzeen: For me both solutions are fine as long as we support multiple context embeddings. If we pass context_embedding into the infer_batch function as an argument, we'll have to load and manage these embeddings outside this function, which would also add complexity. Right now, my understanding is that the checkpoint file would not have context_encoder weights, but instead have a few baked-embeddings with speaker key. In this case, we can just pass the speaker name into the infer_batch function to choose the correct baked-embedding.

rlangman: To be clear, we can think of it as 2 requests: one for supporting multiple voices, one for making code cleaner. Supporting multiple voices now is good to ensure the implementation does not create issues later (also, I would use this multi-voice caching feature right now in my Magpie inference server, if it was available).

Second is the implementation. As a concrete example, reorganizing the code like:

```
def get_context_emb(context_audio):
    if self.context_encoder is None:
        raise IllegalStateException()
    return ...

def get_voice_emb(voice_id):
    if self.voice_embeddings is None:
        raise IllegalStateException()
        
    return ...

def infer_batch(context_audio, ...):
    context_emb = self.get_context_emb(context_audio)
    return self._infer_batch(context_emb, ...)

def infer_batch_with_voice_id(voice_id, ...):
    context_emb = self.get_voice_emb(voice_id)
    return self._infer_batch(context_emb, ...)
```

In general is easier to read, use, and maintain, compared to overloading `infer_batch` with multiple implicit argument combinations that internally create several if branches.

PR #15172, opened 2025-12-10
[DRAFT][TTS] Magpietts Simple API and loading audiocodec from Huggingface

subhankar-ghosh:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

- Refactored prepare_context_tensor method
- Magpietts Simple API do_tts(transcript, language, apply_TN)
- loading audiocodec from Huggingface
- Enable text_tokenizer files to be registered as artifacts, so that they can be saved in the .nemo checkpoint file instead of loading from external files.
- Move eval_config.json from nemo/collections/tts/modules/magpietts_inference to examples/tts
- eval_config.json is now just an example - which contains the dataset for CICD.
- Changed input to inference -> Now instead of taking in dataset names, the input will be a json file which contains the dataset metadata for the datasets you want to run inference on. It will run inference (and eval) on all datasets in the .json file.

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -3298,6 +3298,7 @@ class MagpieTTSModel(ModelPT):
         max_decoder_steps: int = 500,
         use_cfg: bool = True,
         cfg_scale: float = 2.5,
+        speaker_index: Optional[int] = None,
     ) -> tuple:
         """
         Generate speech from raw text transcript.
@@ -3319,6 +3320,9 @@ class MagpieTTSModel(ModelPT):
             max_decoder_steps: Maximum number of decoder steps.
             use_cfg: Whether to use classifier-free guidance.
             cfg_scale: Scale factor for classifier-free guidance.
+            speaker_index: Speaker index for multi-speaker baked embeddings.
+                Valid range: [0, num_baked_speakers - 1]. If None, uses speaker 0.
+                Only applicable for models with baked context embeddings.
 
         Returns:
             Tuple of (audio, audio_len) where:
@@ -3327,6 +3331,7 @@ class MagpieTTSModel(ModelPT):
 
         Raises:
             ValueError: If model does not have a baked context embedding.
+            ValueError: If speaker_index is out of valid range.
             ImportError: If apply_TN=True but nemo_text_processing is not installed.
 
         Example:
@@ -3337,6 +3342,11 @@ class MagpieTTSModel(ModelPT):
             >>> audio, audio_len = model.do_tts(
             ...     "Hello, how are you today?",
             ...     apply_TN=True,
+            ... )
+            >>>
+            >>> # Use a specific speaker (for multi-speaker models)
+            >>> audio, audio_len = model.do_tts(
+            ...     "Hello!", speaker_index=2
             ... )
         """
         if not self.has_baked_context_embedding:
@@ -3390,6 +3400,7 @@ class MagpieTTSModel(ModelPT):
         batch = {
             'text': text_tensor,
             'text_lens': text_lens,
+            'speaker_indices': speaker_index,
         }
 
         # Run inference


nemo/collections/tts/models/magpietts.py
```diff
@@ -2831,6 +2835,166 @@ def setup_validation_data(self, dataset_cfg):
...
+        assert (
+            self.has_baked_context_embedding
+        ), "Model does not have a baked context embedding. Please use a checkpoint with a baked context embedding."
+        # Apply text normalization if requested
+        normalized_text = transcript
+        if apply_TN:
+            try:
+                from nemo_text_processing.text_normalization.normalize import Normalizer
+
+                normalizer = Normalizer(input_case='cased', lang=language)
```
blisc: What happens if we don't have TN for a specific language?

nemo/collections/tts/models/magpietts.py
```diff
@@ -2831,6 +2835,166 @@ def setup_validation_data(self, dataset_cfg):
     def setup_test_data(self, dataset_cfg):
         self._test_dl = self._setup_test_dataloader(dataset_cfg)
 
+    def setup_dummy_text_context_in_batch(
```
blisc: How much work would it be to re-factor prepare_context_tensors() to not require these dummy tensors?

examples/tts/magpietts_inference.py
```diff
@@ -377,10 +367,10 @@ def create_argument_parser() -> argparse.ArgumentParser:
     # Dataset and output arguments
     data_group = parser.add_argument_group('Dataset and Output')
     data_group.add_argument(
-        '--datasets',
+        '--datasets_json_path',
         type=str,
         default=None,
-        help=f'Comma-separated dataset names (default: {EVALUATION_DATASETS})',
+        help='Path to dataset configuration JSON file (will process all datasets in the file)',
     )
```
blisc, line 374: This should no longer have a default be a required arg

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -32,23 +32,25 @@
 
 import nemo.collections.asr as nemo_asr
 from nemo.collections.asr.metrics.wer import word_error_rate_detail
+from nemo.collections.tts.modules.utmosv2 import UTMOSv2Calculator
 from nemo.utils import logging
```
blisc, line 34: Can you do a find and replace inside of this file of all print statements and switch them to logging?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -32,23 +32,25 @@
 
 import nemo.collections.asr as nemo_asr
 from nemo.collections.asr.metrics.wer import word_error_rate_detail
+from nemo.collections.tts.modules.utmosv2 import UTMOSv2Calculator
```
blisc: We also need to make UTMOS an optional import

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -32,23 +32,25 @@
...
 from nemo.utils import logging
 
 # Path to evalset config JSON
-EVALSET_CONFIG_PATH = Path(__file__).parent / 'evalset_config.json'
+# Get project root: from nemo/collections/tts/modules/magpietts_inference/ go up to NeMo/
+_project_root = Path(__file__).resolve().parents[5]
+EVALSET_CONFIG_PATH = _project_root / 'examples' / 'tts' / 'evalset_config.json'
```
blisc: We should refactor this to work the same way as examples/tts/magpietts_inference.py

nemo/collections/tts/models/magpietts.py
```diff
@@ -2943,6 +3026,166 @@ def setup_validation_data(self, dataset_cfg):
...
+        """Setup dummy audio context tensors in the batch dictionary."""
+        # Model has baked context - create minimal dummy context tensors
+        # These will be ignored in prepare_context_tensors when baked embedding is used
+        dummy_context_codes = torch.zeros(1, self.num_audio_codebooks, 2, device=self.device, dtype=torch.long)
+        dummy_context_codes[:, :, 0] = self.context_audio_bos_id
+        dummy_context_codes[:, :, 1] = self.context_audio_eos_id
+        batch['context_audio_codes'] = dummy_context_codes
+        batch['context_audio_codes_lens'] = torch.tensor([2], device=self.device, dtype=torch.long)
+
+    def do_tts(
```
subhankar-ghosh, line 3291: @blisc @rlangman I wrote this api do_tts currently just for the open source checkpoint (no audio context support, no text context support). Do we want to make this more general to handle audio context and text context support as well? I can write a logic to ignore them when baked context embeddings are there. Thoughts?

examples/tts/magpietts_inference.py
subhankar-ghosh: The default is None in which case the CICD json file (also serves as an example) is loaded - the path to which is hardcoded (default value) in nemo/collectioins/tts/modules/magpietts_inference/evaluate_generated_audio.py. Do you want to remove the hardcoded path and just make it a required arg?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
subhankar-ghosh: There are no print statements in this file. I checked. Only one `pprint.pprint(avg_metrics)` to print the average metrics. Every other print statement has been replaced by logging.info in the previous PR.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Added normalizer cache and checks to skip normalization in case language does not exist in the Normalizer.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
subhankar-ghosh: Done

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
subhankar-ghosh: This is the default/example json file - the CICD file. I can remove this, so no default value. If a valid .json path is not provided we can throw an error, making it a required argument.

examples/tts/magpietts_inference.py
```diff
@@ -487,11 +477,10 @@ def main():
...
     args = parser.parse_args()
 
-    # Set default datasets if not provided
-    if args.datasets is None:
-        args.datasets = EVALUATION_DATASETS
+    dataset_meta_info = load_evalset_config(args.datasets_json_path)
+    datasets = list(dataset_meta_info.keys())
 
-    datasets = args.datasets.split(",")
+    logging.info(f"Loaded {len(datasets)} datasets from {args.datasets_json_path}: {', '.join(datasets)}")
```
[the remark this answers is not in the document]
subhankar-ghosh: Removed. Added logging about filename where it is being loaded

nemo/collections/tts/models/magpietts.py
```diff
@@ -120,13 +120,20 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+
         # load codec, disable loading of loss modules not needed during inference
-        codec_model_cfg = AudioCodecModel.restore_from(cfg.get('codecmodel_path'), return_config=True)
-        if "use_scl_loss" in codec_model_cfg:
-            codec_model_cfg.use_scl_loss = False
-        codec_model = AudioCodecModel.restore_from(
-            cfg.get('codecmodel_path'), strict=False, override_config_path=codec_model_cfg
-        )
+        codec_model_path = cfg.get('codecmodel_path')
+        if codec_model_path.startswith('nvidia/'):
```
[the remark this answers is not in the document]
subhankar-ghosh: This is NeMo's standard way of checking huggingface model name.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -34,21 +34,34 @@
...
+try:
+    from nemo.collections.tts.modules.utmosv2 import UTMOSv2Calculator
+
+    UTMOSV2_AVAILABLE = True
+except (ImportError, ModuleNotFoundError) as e:
+    UTMOSV2_AVAILABLE = False
+    logging.warning(
+        f"UTMOSv2Calculator not available: {e}. "
+        "UTMOSv2 metrics will be disabled. Install required dependencies to enable."
+    )
```
blisc, line 47: Can you remove the UTMOS from the requirements? And update this comment on how to install it

nemo/collections/tts/models/magpietts.py
blisc: Eventually yes, but let's move it out of this PR

examples/tts/magpietts_inference.py
blisc: >Do you want to remove the hardcoded path and just make it a required arg?

Yes. Is there a reason we don't want this behaviour?

nemo/collections/tts/models/magpietts.py
```diff
@@ -84,6 +84,52 @@ class InferBatchOutput:
     headwise_cross_attention_maps: Optional[List[Any]] = None
 
 
+@dataclass
+class ContextTensorsOutput:
```
blisc, line 88: The changes in the refactor are great. I love how we split prepare_context_tensors into appropriate helper functions now.
One comment on the use of dataclasses. As we refactor and make the code more legible, we should aim to simplify the inputs and outputs of our functions to minimize the use of dictionaries and thus dataclasses moving forward.
The dictionary return was really bad and the dataclasses are a huge step above that. I opine that the most optimal solution is to have smaller functions that return a reasonable number of outputs rather than dataclasses

rlangman: It is definitely an improvement. Though it also highlights that the `prepare_context` method is too large right now. There is almost no situation where a helper function should have 16 output parameters, with many of them being `None` depending on the input. It looks like it could be split into 3 separate methods which prepare (1) audio context (2) text context (3) attention prior. Not a requirement for this PR, though.

nemo/collections/tts/models/magpietts.py
```diff
@@ -395,6 +448,85 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+                    try:
+                        # register_artifact handles both:
+                        # - Local paths: registers for .nemo packaging, returns absolute path
+                        # - nemo: paths: resolves to extracted file location
+                        artifact_path = self.register_artifact(
+                            f'text_tokenizers.{tokenizer_name}.g2p.phoneme_dict',
+                            phoneme_dict_path,
+                            verify_src_exists=True,
+                        )
+                        if artifact_path:
```
rlangman, line 499: When would `register_artifact()` return None?

nemo/collections/tts/models/magpietts.py
```diff
@@ -395,6 +448,85 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+                            with open_dict(cfg):
+                                cfg.text_tokenizers[tokenizer_name].g2p.phoneme_dict = artifact_path
+                    except FileNotFoundError:
+                        logging.warning(
+                            f"phoneme_dict file not found for tokenizer '{tokenizer_name}': "
+                            f"{phoneme_dict_path}. Artifact will not be packaged in .nemo file."
+                        )
```
rlangman, line 506: I feel like phoneme dictionary not existing should be an error. Granted, for backwards compatibility any checkpoint we use would need to be updated to have the registered artifacts.

nemo/collections/tts/models/magpietts.py
```diff
@@ -395,6 +448,85 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+
+                # Register heteronyms (or resolve nemo: path if restoring from .nemo)
+                heteronyms_path = (
+                    g2p_cfg.get('heteronyms', None)
+                    if hasattr(g2p_cfg, 'get')
+                    else getattr(g2p_cfg, 'heteronyms', None)
+                )
```
rlangman, line 513: What are the two scenarios leading to this if/else?

nemo/collections/tts/models/magpietts.py
```diff
@@ -1570,163 +1702,341 @@ def embed_context_text(self, context_text_tokens):
...
+            context_audio_embedded, context_text_embedded
+        )
+
+        # Interpolate between audio and text embeddings based on has_text_context flag
+        has_text_context = batch['has_text_context'].unsqueeze(-1).unsqueeze(-1).float()  # (B, 1, 1)
+        context_embedded = has_text_context * context_text_embedded + (1 - has_text_context) * context_audio_embedded
+        context_lens = (
+            batch['has_text_context'].float() * context_text_lens
+            + (1 - batch['has_text_context'].float()) * context_audio_codes_lens
+        )  # (B,)
```
rlangman: It is not relevant to this PR, but why are the tensors interpolated in this way (multiplying by 0 or 1) instead of using something like `torch.where(has_text_context, text_context, audio_context)`

subhankar-ghosh: Changed it to `torch.where`

subhankar-ghosh:
fixed the UTMOS requirements and the json in magpietts_inference. Changed to a better way of using has_text_context in magpietts.py.

examples/tts/evalset_config.json
rfejgin, line 1: This may be a good time to remove `feature_dir`, I believe we don't need that anymore (@paarthneekhara could you confirm?) and it would simplify both the JSON file and the code. But we'd have to remove it in a few places in the code.

blisc (approved):

blisc (approved):

examples/tts/evalset_config.json
subhankar-ghosh: Do you mean from the evalset_config.json metadata file? I would not do that in this PR, because I think this might need testing of different scenarios. But you raise a good point. I did a quick search of `feature_dir` and I did not see it being used in a lot of places. So, may be with my next PR on long-form, I would remove it.

PR #15178, opened 2025-12-11
[TTS] MagpieTTS Inference Refactoring

subhankar-ghosh:
> [!IMPORTANT]  

# What does this PR do ?

Copy of [[TTS] MagpieTTS Inference Changes](https://github.com/NVIDIA-NeMo/NeMo/pull/15158) #15158
Because after rebasing with main git showed the PR had 231 files changed, which was wrong. So I created a fresh PR. This shows the correct number of files changed.

# Changelog

1. Refactored Inference script
a) Modularized code
b) Separated inference and evaluation logic
c) Separated different configs
d) Lays the groundwork for longform inference

2. infer_batch returns dataclass object now

3. Moved inference code from scripts to examples

4. Added fixed context embeddings instead of getting it through context encoder, in case checkpoint does not have context encoder

5. Removed evalset_config.py and replaced with evalset.json

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -547,7 +547,6 @@ class MagpieTTSModel(ModelPT):
         indices = self._normalize_speaker_indices(speaker_indices, batch_size, device)
 
         # Lookup flattened embeddings via nn.Embedding: (B,) -> (B, T*D)
-        print(f"indices: {indices}")
         flat_embeddings = self.baked_context_embedding(indices)
 
         # Reshape to 3D: (B, T*D) -> (B, T, D)
@@ -2542,10 +2541,6 @@ class MagpieTTSModel(ModelPT):
                             dummy_addition_dec_mask
                         )
 
-                    # print(f"step {idx}")
-                    # print(f"use_cfg {use_cfg}")
-                    # print(f"shape {cfg_audio_codes_embedded.shape}")
-                    # print(f"use kv cahce? {self.use_kv_cache_for_inference}")
                     combined_logits, attn_probs, dec_out = self.forward(
                         dec_input_embedded=cfg_audio_codes_embedded,
                         dec_input_mask=cfg_audio_codes_mask,


nemo/collections/tts/models/magpietts.py
```diff
@@ -397,6 +438,110 @@ def check_frame_stacking_config_validity(self):
...
+        if not self.has_baked_context_embedding:
+            return 0
+        if self.baked_context_embedding.dim() == 2:
+            return 1  # Single speaker: (T, D)
+        elif self.baked_context_embedding.dim() == 3:
+            return self.baked_context_embedding.size(0)  # Multi-speaker: (N, T, D)
+        else:
+            raise ValueError(f"Unexpected baked_context_embedding shape: {self.baked_context_embedding.shape}")
+
+    def get_baked_context_embeddings_batch(
```
blisc, line 521: This function is overly complicated. We should split it into two:

- Handles 1 speaker_index, None speaker_index, and a list of speaker_indices and this should create a tensor of size [B] that we can use in a simple torch embedding layer
- Baked_embedding should just be a simple torch embedding layer

subhankar-ghosh: Made changes to handle this.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -27,13 +27,25 @@
...
-import scripts.magpietts.evalset_config as evalset_config
 import soundfile as sf
 import torch
 from transformers import Wav2Vec2FeatureExtractor, WavLMForXVector, WhisperForConditionalGeneration, WhisperProcessor
 
 import nemo.collections.asr as nemo_asr
 from nemo.collections.asr.metrics.wer import word_error_rate_detail
+
+# Path to evalset config JSON
+EVALSET_CONFIG_PATH = os.path.join(os.path.dirname(__file__), 'evalset_config.json')
```
blisc: Nitpick: would recommend switch to pathlib instead of os.path
```suggestion
EVALSET_CONFIG_PATH = Path(__file__).parent / 'evalset_config.json'
```

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -0,0 +1,271 @@
...
+import os
+from dataclasses import dataclass
+from typing import Optional, Tuple
+
+import torch
+from omegaconf import DictConfig, OmegaConf, open_dict
+
+from nemo.collections.tts.models import MagpieTTSModel
+
+logger = logging.getLogger(__name__)
```
blisc: Use nemo's logging instead: from nemo.utils import logging

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -0,0 +1,404 @@
...
+
+import soundfile as sf
+import torch
+from PIL import Image
+
+from nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers import AggregatedTTSTokenizer, IPATokenizer
+from nemo.collections.tts.data.text_to_speech_dataset import MagpieTTSDataset
+from nemo.collections.tts.models import MagpieTTSModel
+
+logger = logging.getLogger(__name__)
```
blisc: Use NeMo's logging instead

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,149 @@
...
+from dataclasses import dataclass
+from typing import Dict, List, Tuple
+
+import numpy as np
+import scipy.stats as stats
+
+# Import the existing evaluation module
+import nemo.collections.tts.modules.magpietts_inference.evaluate_generated_audio as evaluate_generated_audio
+
+logger = logging.getLogger(__name__)
```
blisc: Use Nemo's logging

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,618 @@
...
+    load_magpie_model,
+)
+from nemo.collections.tts.modules.magpietts_inference.visualization import create_combined_box_plot, create_violin_plot
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+)
+logger = logging.getLogger(__name__)
```
blisc: Use Nemo's logging

nemo/collections/tts/models/magpietts.py
```diff
@@ -397,6 +438,117 @@ def check_frame_stacking_config_validity(self):
...
+        if not self.has_baked_context_embedding:
+            raise ValueError("No baked context embedding available")
+
+        device = self.baked_context_embedding.device
+        indices = self._normalize_speaker_indices(speaker_indices, batch_size, device)
+
+        # Index into embeddings: (N, T, D) -> (B, T, D)
+        embeddings = self.baked_context_embedding[indices]  # (B, T, D)
+        lengths = self.baked_context_embedding_len[indices]  # (B,)
+        return embeddings, lengths
```
blisc: This section looks really strange to me. What is self.baked_context_embedding? Shouldn't it be a torch.nn.Embedding? In which case, we just forward on it and do not index into it. I don't understand how this works for batch_size > 1.

subhankar-ghosh: It is the output of context_encoder:
`context_embedding = model.context_encoder(
            context_audio_embedded, context_mask, cond=None, cond_mask=None
        )['output']  # Shape: (N, T, E)`

We save this output as weights/tensors.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
subhankar-ghosh: Done

nemo/collections/tts/modules/magpietts_inference/utils.py
subhankar-ghosh: Done

nemo/collections/tts/modules/magpietts_inference/inference.py
subhankar-ghosh: Done

nemo/collections/tts/modules/magpietts_inference/evaluation.py
subhankar-ghosh: Done

examples/tts/magpietts_inference.py
subhankar-ghosh: Done

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: The output of context_encoder is 3D (B, T, E), not really embeddings. We pass the audio through context_encoder and save the output as baked_context_embedding

nemo/collections/tts/modules/magpietts_inference/utils.py
blisc: Can we switch to a consistent format of `from nemo.utils import logging` not `from nemo.utils import logging as logger`? Like how it's done inside of the nemo/collections/tts/models/magpietts.py

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -16,24 +16,36 @@
 """
```
blisc, line 16: Can you do a find and replace inside of this file of all print statements and switch them to logging?

nemo/collections/tts/models/magpietts.py
blisc: This still seems incorrect to me: Our context embeddings are always of a fixed time dimension. Why don't we switch to saving it into an embedding table of size: [N, T*E]? And then reshape it afterwards.
In the cases where we use text context and T is not always equivalent to 5 seconds of audio, there should still be padding up to T and we can save the lengths in another tensor.

examples/tts/magpietts_inference.py
```diff
@@ -0,0 +1,612 @@
...
+        "ssim_gt_context_avg_alternate,cer_gt_audio_cumulative,wer_gt_audio_cumulative,"
+        "utmosv2_avg,total_gen_audio_seconds"
+    )
+
+    for dataset in datasets:
+        logger.info(f"Processing dataset: {dataset}")
+
+        if dataset not in dataset_meta_info:
+            logger.warning(f"Dataset '{dataset}' not found in evalset_config.json, skipping.")
+            continue
```
rlangman, line 198: It would be safer if this raised an error

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -16,24 +16,36 @@
...
 import soundfile as sf
 import torch
 from transformers import Wav2Vec2FeatureExtractor, WavLMForXVector, WhisperForConditionalGeneration, WhisperProcessor
 
 import nemo.collections.asr as nemo_asr
 from nemo.collections.asr.metrics.wer import word_error_rate_detail
+from nemo.utils import logging as logger
+
+# Path to evalset config JSON
+EVALSET_CONFIG_PATH = Path(__file__).parent / 'evalset_config.json'
```
rlangman, line 38: Json input should be passed as an argument to the function/script. I do not see why a default value would be needed (CI tests should be updated to provide config as an input argument, either from the `/examples` directory or from S3). But if it is really needed, `load_evalset_config` could return the hardcoded pre-parsed dictionary `{"an4_val_ci": ...}` instead having a static json file in the code.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -0,0 +1,147 @@
...
+) -> Dict[str, str]:
+    """Compute mean and confidence interval for specified metrics.
+
+    Args:
+        metrics_list: List of metric dictionaries (one per repeat/run).
+        metric_keys: List of metric names to compute statistics for.
+        confidence: Confidence level (default: 0.95 for 95% CI).
+
+    Returns:
+        Dictionary mapping metric names to formatted "mean ± CI" strings.
```
rlangman: Should now say the map contains tuples

nemo/collections/tts/modules/magpietts_inference/utils.py
subhankar-ghosh: Fixed

nemo/collections/tts/models/magpietts.py
```diff
@@ -397,6 +438,117 @@ def check_frame_stacking_config_validity(self):
...
+        """Check if the model has a baked context embedding.
+
+        Returns:
+            True if baked_context_embedding buffer is set, not None, and has elements.
+        """
+        return (
+            self.model_type == 'decoder_ce'
+            and hasattr(self, 'baked_context_embedding')
+            and self.baked_context_embedding is not None
+            and self.baked_context_embedding.numel() > 0
```
rlangman: The last condition should probably throw an error, as it implies someone tried to load an embedding but did it incorrectly.

nemo/collections/tts/models/magpietts.py
```diff
@@ -397,6 +438,117 @@ def check_frame_stacking_config_validity(self):
...
 
+    @property
+    def has_baked_context_embedding(self) -> bool:
+        """Check if the model has a baked context embedding.
+
+        Returns:
+            True if baked_context_embedding buffer is set, not None, and has elements.
+        """
+        return (
+            self.model_type == 'decoder_ce'
```
rlangman, line 451: Why do we want to enforce that baked embedding can only be used with `decoder_ce`? Its not a big deal since I imagine we will be removing most of the other `model_type` code paths later. Just as first glance, it looks like the implementation would be easier/simpler if it was not gated by model type.

nemo/collections/tts/models/magpietts.py
rlangman: A benefit of using an embedding layer is you could validate the shape of the embedding during model initialization. The value of N, `num_baked_speakers`, would need to be provided as a buffer from the checkpoint, replacing the current definition of it as an `@property`

nemo/collections/tts/models/magpietts.py
```diff
@@ -1460,14 +1644,27 @@ def prepare_context_tensors(self, batch):
...
-                dec_context_size = context_mask.size(1)
                 context_embeddings = None  # Address CodeQL
                 if self.model_type == 'decoder_context_tts':
                     context_embeddings = context_input_embedded
                 elif self.model_type == 'decoder_ce':
-                    context_embeddings = self.context_encoder(
-                        context_input_embedded, context_mask, cond=None, cond_mask=None
-                    )['output']
+                    # Check for baked context embedding first
+                    if self.has_baked_context_embedding:
```
rlangman, line 1667: I would have this branch decided by whether speaker_indices is None or not. If speaker indices is provided, then use this path, else use the context encoder. An error should be thrown if someone provides neither or both context_audio and speaker_indices.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
blisc: > Json input should be passed as an argument to the function/script. 

I agree, let's change the current `args.evalset` parameter that requires the name of the dataset to take in a json instead.

> But if it is really needed, load_evalset_config could return the hardcoded pre-parsed dictionary {"an4_val_ci": ...} instead having a static json file in the code.

I would prefer it to point to a json file in the code to serve as an example of what the json structure should look like for this function in addition to docstringing it.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: > The value of N, num_baked_speakers, would need to be provided as a buffer from the checkpoint

Why do we need to have number of baked speakers saved as a buffer? We need to save T and D sure, but I am not sure why we need to save N as a buffer. We can get the num embeddings by `embedding_layer.weight.shape[0]`
Nonetheless, we are changing the implementation to torch.nn.embedding

subhankar-ghosh: > I would have this branch decided by whether speaker_indices is None or not. If speaker indices is provided, then use this path, else use the context encoder. An error should be thrown if someone provides neither or both context_audio and speaker_indices.

The idea is if baked_embeddings are there the user should only use that and not context encoder. If speaker_indices are not provided and baked_embeddings are present, the default behavior is to use the first speaker embedding. I thought this is what we had decided - baked_embeddings so we do not let users use any other voice.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
subhankar-ghosh: done

examples/tts/magpietts_inference.py
subhankar-ghosh: Ideally yes. But if someone is inferring on 10 datasets and the 9th doesn't exist, in that case the user needs to start from the beginning. What do you think?

rlangman: That can be avoided by having this check validate that all datasets exist before entering the current `for dataset in datasets:` loop

nemo/collections/tts/models/magpietts.py
rlangman: > > The value of N, num_baked_speakers, would need to be provided as a buffer from the checkpoint
> 
> Why do we need to have number of baked speakers saved as a buffer? We need to save T and D sure, but I am not sure why we need to save N as a buffer. We can get the num embeddings by `embedding_layer.weight.shape[0]`

The request from Jason was to make the parameters a `nn.Embedding` layer. Creating this layer requires knowing the value of N ahead of time. Like:

```
num_baked_embeddings = self.register_buffer('num_baked_embeddings ', 0)
self.embedding_layer = torch.nn.Embedding(num_embeddings=num_baked_embeddings, embedding_dim=max_context_len * context_dim, _freeze=True)
```

This is a clean approach that guarantees the embedding is the correct dimensions when the model is first loaded. Otherwise if the `embedding_layer` happens to be wrong in the model checkpoint, then you would only know by hitting errors at inference time. Granted this is a pretty minor concern.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: has_baked_context_embedding property is used as a check in many places. In cases where the condition should absolutely be true it does throw an error. Hence I think this method just returns if baked_context_embedding is present and properly loaded.

nemo/collections/tts/models/magpietts.py
blisc: > > I would have this branch decided by whether speaker_indices is None or not. If speaker indices is provided, then use this path, else use the context encoder. An error should be thrown if someone provides neither or both context_audio and speaker_indices.
> 
> The idea is if baked_embeddings are there the user should only use that and not context encoder. If speaker_indices are not provided and baked_embeddings are present, the default behavior is to use the first speaker embedding. I thought this is what we had decided - baked_embeddings so we do not let users use any other voice.

I think we assume, and we should add this as a check to ensure that our assumption is correct, that a model can either have backed context embeddings OR a context encoder. Not both. So the path where speaker_indices is None is ok with me

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
subhankar-ghosh: I agree, but this might affect CICD tests so I would like to take care of this in the next PR - which would contain simple do_tts API.

blisc (approved):

nemo/collections/tts/models/magpietts.py
rlangman: Unrelated to this release, there should be some inference path that allows a user to pre-compute and reuse context embeddings. So that when hosting the model, it is not necessary to recompute the same context for every request. It would be nice if that use case and the baked context embeddings could be consolidated, but we can revisit that in the later PR which defines the actual inference functions/APIs.

examples/tts/magpietts_inference.py
blisc: This looks to be fixed in #15172, resolving

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
blisc: This looks to be fixed in https://github.com/NVIDIA-NeMo/NeMo/pull/15172, resolving

PR #15191, opened 2025-12-15
[TTS] Refactor Magpie to support codec conversion and bandwidth extension

rlangman:
This PR fixes a bug in MagpieTTS training in which self._codec_converter.convert_original_to_new() is being called on audio codec tokens after the Lhotse data loader has already pre-populated them with BOS and EOS tokens.

The addition of BOS and EOS tokens is now always done inside the model class, after convert_original_to_new() is called. These tokens are also removed before convert_new_to_original() is called, so that they are not mistakenly passed into the codec model.

Additionally, this includes changes to the magpie code and codec interfaces to support doing bandwidth extension within the codec, where the sample rate of the input and output audio is different.

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -899,7 +899,9 @@ class MagpieTTSModel(ModelPT):
         # codes_len: (B,)
         codes = torch.nn.functional.pad(input=codes, pad=(num_bos_tokens, 0), value=bos_id)
         codes_len = codes_len + num_bos_tokens
-        codes, codes_len = self.add_eos_token(codes=codes, codes_len=codes_len, eos_id=eos_id, num_eos_tokens=num_eos_tokens)
+        codes, codes_len = self.add_eos_token(
+            codes=codes, codes_len=codes_len, eos_id=eos_id, num_eos_tokens=num_eos_tokens
+        )
         return codes, codes_len
 
     def remove_bos_token(self, codes, codes_len, num_tokens=1):
--- a/nemo/collections/tts/models/magpietts_preference_optimization.py
+++ b/nemo/collections/tts/models/magpietts_preference_optimization.py
@@ -893,7 +893,9 @@ class MagpieTTSModelOnlinePO(MagpieTTSModel):
             with torch.no_grad():
                 reference_model_output = self._reference_model.process_batch(batch_repeated)
 
-        codebook_targets, _ = self.add_eos_token(codes=predicted_codes, codes_len=predicted_codes_lens, eos_id=self.audio_eos_id)
+        codebook_targets, _ = self.add_eos_token(
+            codes=predicted_codes, codes_len=predicted_codes_lens, eos_id=self.audio_eos_id
+        )
 
         total_loss = None
         total_kl = None


nemo/collections/tts/models/magpietts.py
```diff
@@ -645,62 +646,77 @@ def load_state_dict(self, state_dict, strict=True):
...
-            raise ValueError(f"Received audio_type of {audio_type}. Must be `target` or `context`")
+    def add_special_tokens(self, codes, codes_len, bos_id, eos_id, num_bos_tokens=1):
+        # codes: (B, C, T')
+        # codes_len: (B,)
+        codes = torch.nn.functional.pad(input=codes, pad=(num_bos_tokens, 0), value=bos_id)
+        codes = torch.nn.functional.pad(input=codes, pad=(0, 1), value=0)
+        codes_len = codes_len + num_bos_tokens + 1
+        # Insert EOS token at new final token entry
+        for idx in range(codes.size(0)):
+            codes[idx, :, codes_len[idx] - 1] = eos_id
```
blisc: Why is the logic for bos and eos different?

nemo/collections/tts/models/magpietts.py
```diff
@@ -1828,55 +1855,51 @@ def prepare_dummy_cond_for_cfg(self, cond, cond_mask, additional_decoder_input,
...
+        )
+        audio_codes_target_unstacked = self.pad_audio_codes(audio_codes_target_unstacked)
+        # Remove BOS token for aligner targets
+        audio_codes_embedded_target, audio_codes_lens_target = self.remove_embedded_bos_token(
+            embedded=audio_codes_embedded_all, embedded_len=audio_codes_lens_all
+        )
+        # Remove EOS token for decoder inputs
+        audio_codes_embedded_input, audio_codes_lens_input = self.remove_embedded_eos_token(
+            embedded=audio_codes_embedded_all, embedded_len=audio_codes_lens_all
+        )
```
blisc: What happens if the frame_stacking is enabled, and the EOS is part of the frame-stacked sequence in the middle of the sequence?
IE if frame_stacking is 4, and the audio codes becomes:

1. BOS, BOS, BOS, BOS
2. Audio_1, Audio_2, Audio_3, Audio_4
3. Audio_5, EOS, ???, ???

nemo/collections/tts/models/magpietts.py
rlangman: Because BOS position is independent of the audio lengths (always insert/remove at index 0), and EOS position is dependent on the audio lengths.

nemo/collections/tts/models/magpietts.py
rlangman: For BOS, this works because it is effectively a standalone token. The token sequence with EOS removed is only used as input to the decoder. It removes token Audio_5 as well, but that is the correct input to the decoder in frame stacking mode.

I did test that the training code runs successfully when `frame_stacking_factor=2`. Though I have not verified that it produces a good model after a full training run.

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -448,9 +436,6 @@ def __getitem__(self, index):
             audio_codes_path = data.manifest_entry['target_audio_codes_path']
             audio_codes = torch.load(audio_codes_path).long()  # (C, T)
             spec_len = audio_codes.shape[1] + 1  # +1 for EOS
```
rfejgin, line 439: Suggest to move this `+ 1` to the model class as well since we don't want the addition of the EOS+BOS and the manipulation of lengths to be in unrelated parts of the code.

Or is it needed for getting a correct beta_binomial prior shape further below?

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -536,14 +514,8 @@ def __getitem__(self, index):
...
             if self.load_cached_codes_if_available:
-                context_bos_tensor = torch.full(
-                    (self.num_audio_codebooks, 1), self.context_audio_bos_id, dtype=torch.int32
-                )
-                context_eos_tensor = torch.full(
-                    (self.num_audio_codebooks, 1), self.context_audio_eos_id, dtype=torch.int32
-                )
-                context_audio_codes = torch.cat([context_bos_tensor, context_eos_tensor], dim=1)
-                context_audio_codes_len = context_audio_codes.shape[1]
+                context_audio_codes = torch.zeros([0, self.num_audio_codebooks], dtype=torch.int32)
```
rfejgin: To match the deleted code, should the order of dimensions be reversed, i.e. 
```
torch.zeros([self.num_audio_codebooks, 0], dtype=torch.int32)
```

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -240,17 +224,15 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List]]:
...
-                spec_len = audio_codes.shape[1] + 1  # +1 for EOS
-                audio_bos_tensor = torch.full((audio_codes.shape[0], 1), self.audio_bos_id, dtype=audio_codes.dtype)
-                audio_eos_tensor = torch.full((audio_codes.shape[0], 1), self.audio_eos_id, dtype=audio_codes.dtype)
-                audio_codes = torch.cat([audio_bos_tensor, audio_codes, audio_eos_tensor], dim=1)
-                audio_codes_len = audio_codes.shape[1]
-                audio_codes_list.append(audio_codes.T)  # transpose to (T, C) to use collate_matrices to process batch.
+                audio_codes = torch.from_numpy(cut.target_codes.load())  # (C, T)
+                audio_codes = audio_codes.T.long()  # transpose to (T, C) to use collate_matrices to process batch.
+                audio_codes_len = audio_codes.shape[0]
+                spec_len = audio_codes_len + 1  # +1 for EOS
```
rfejgin, line 228: Same comment about moving this to model

nemo/collections/tts/models/audio_codec.py
rfejgin, line 300: need to add sample rate argument to docstring.

nemo/collections/tts/models/magpietts.py
```diff
@@ -645,62 +646,77 @@ def load_state_dict(self, state_dict, strict=True):
...
-            audio_eos_id = self.context_audio_eos_id
-            audio_bos_id = self.context_audio_bos_id
-        else:
-            raise ValueError(f"Received audio_type of {audio_type}. Must be `target` or `context`")
+    def add_special_tokens(self, codes, codes_len, bos_id, eos_id, num_bos_tokens=1):
+        # codes: (B, C, T')
+        # codes_len: (B,)
+        codes = torch.nn.functional.pad(input=codes, pad=(num_bos_tokens, 0), value=bos_id)
+        codes = torch.nn.functional.pad(input=codes, pad=(0, 1), value=0)
+        codes_len = codes_len + num_bos_tokens + 1
```
rfejgin: could you add a comment saying the the +1 is for EOS?

nemo/collections/tts/models/magpietts.py
```diff
@@ -645,62 +646,77 @@ def load_state_dict(self, state_dict, strict=True):
...
+        # codes_len: (B,)
+        embedded = embedded[:, 1:, :]
+        embedded_len = embedded_len - 1
+        return embedded, embedded_len
+
+    def remove_eos_token(self, codes, codes_len):
+        # codes: (B, C, T')
+        # codes_len: (B,)
+        codes_len = codes_len - 1
+        codes = codes[:, :, :-1]
```
rfejgin, line 925: If we have a debug flag it would be nice to add an assert or error check that verifies indeed it's an EOS token in the position we are removing, as expected.

nemo/collections/tts/models/magpietts.py
```diff
@@ -645,62 +646,77 @@ def load_state_dict(self, state_dict, strict=True):
...
+        for idx in range(codes.size(0)):
+            codes[idx, :, codes_len[idx] - 1] = eos_id
+
+        return codes, codes_len
+
+    def remove_bos_token(self, codes, codes_len, num_tokens=1):
+        # codes: (B, C, T')
+        # codes_len: (B,)
+        codes = codes[:, :, num_tokens:]
+        codes_len = codes_len - num_tokens
```
rfejgin, line 911: It would be good to add a debug check confirming that post this operation no BOS codes remain in the tensor. To catch cases like having `frame_stacking_factor`'s worth of BOS tokens and forgetting to remove them all.

It has to be done under some kind of debug flag though since this would cause a host-gpu synchronization (this only matters if this is something that happens in a "hot" path like every training step or every inference step).

nemo/collections/tts/models/magpietts.py
```diff
@@ -1828,55 +1855,51 @@ def prepare_dummy_cond_for_cfg(self, cond, cond_mask, additional_decoder_input,
...
-        audio_codes_target_unstacked = audio_codes[:, :, self.frame_stacking_factor :]
-        audio_codes_lens_input_unstacked = audio_codes_lens - 1  # don't count EOS for input
-        audio_codes_lens_target_unstacked = audio_codes_lens - self.frame_stacking_factor  # don't count BOS for target
-        audio_codes_lens_input = torch.floor(audio_codes_lens_input_unstacked / self.frame_stacking_factor).long()
-        audio_codes_embedded_all = self.embed_audio_tokens(
-            audio_codes
-        )  # (B, T, E) # Computing this to be use in the alignment encoder
-        audio_codes_embedded = audio_codes_embedded_all[
-            :, :-1, :
-        ]  # (B, T', E) Input to the decoder; this is already in the frame-stacked domain, hence the -1 (not `frame_stacking_factor`)
```
rfejgin, line 2290: This code (pre change) is bit delicate so I'm nervous about changing it; took a while to get this right as there are a few corner cases.

I'll review in greater detail but I suggest to train with 4x-frame-stacking (not just 2x) as I've observed 4x to uncover some corner cases that don't come up in 2x-frame-stacking. Also it would be a good idea to collect some inference metrics; when I got some details here wrong it showed up as slightly elevated WERs (by ~1 percentage point) in the 4x-stacked case. The baseline would be the issue, might have to train with the pre-refactored code too.

That said, the refactored code does look cleaner!

nemo/collections/tts/models/magpietts.py
rlangman: Given that we are supposed to avoid of if/else in our training and inference code due to overhead, debug flags does not sound feasible. Unit tests would be more appropriate.

nemo/collections/tts/data/text_to_speech_dataset.py
rlangman: This is needed here because of the beta binomial prior. The prior could also be computed inside the model class (having the model take it as input is extremely awkward) instead of the data loader with only a very small slowdown to training, but I would leave testing that to a different PR.

nemo/collections/tts/models/magpietts.py
rlangman: So for testing I need to validate that the code runs when frame stacking is set to 4.

It sounds like we also want to train 2 models with 2x frame stacking before and after this change and validate that they have the same evaluation performance. Would we block the merging of this PR on that multi-week test?

nemo/collections/tts/data/text_to_speech_dataset.py
rfejgin: Okay, make sense to me

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
rfejgin: Rationale explained in the other comment

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -240,17 +224,15 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List]]:
...
                 # and duration are None to the load function.
-                audio_codes = torch.from_numpy(cut.target_codes.load()).long()  # (C, T)
-                spec_len = audio_codes.shape[1] + 1  # +1 for EOS
-                audio_bos_tensor = torch.full((audio_codes.shape[0], 1), self.audio_bos_id, dtype=audio_codes.dtype)
-                audio_eos_tensor = torch.full((audio_codes.shape[0], 1), self.audio_eos_id, dtype=audio_codes.dtype)
-                audio_codes = torch.cat([audio_bos_tensor, audio_codes, audio_eos_tensor], dim=1)
-                audio_codes_len = audio_codes.shape[1]
-                audio_codes_list.append(audio_codes.T)  # transpose to (T, C) to use collate_matrices to process batch.
+                audio_codes = torch.from_numpy(cut.target_codes.load())  # (C, T)
+                audio_codes = audio_codes.T.long()  # transpose to (T, C) to use collate_matrices to process batch.
```
XuesongYang: do we still require `long()` conversion?

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -240,17 +224,15 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List]]:
...
+                audio_codes_len = audio_codes.shape[0]
+                spec_len = audio_codes_len + 1  # +1 for EOS
+                audio_codes_list.append(audio_codes)
                 audio_codes_len_list.append(audio_codes_len)
             else:
                 # Only load audio if codes are not available
-                audio_array = cut.recording.resample(self.sample_rate).load_audio().squeeze(0)
+                audio_array = cut.target_audio.resample(self.sample_rate).load_audio().squeeze(0)
```
XuesongYang, line 233: not sure why making such changes. by default, `recording` field in the `MonoCut`is target audio.

XuesongYang: discussed offline.

nemo/collections/tts/models/magpietts.py
rlangman: I have training runs with 4x frame sacking ongoing, using code before and after this commit. Training and validation losses are identical. I can run evaluation when they are finished.

nemo/collections/tts/data/text_to_speech_dataset.py
rlangman: Changed. It is pretty unintuitive that MagpieTTSDataset and MagpieTTSLhotseDataset have many duplicated code blocks, except that each code block in MagpieTTSLhotseDataset returns the transpose.

nemo/collections/tts/models/audio_codec.py
rlangman: Added

nemo/collections/tts/models/magpietts.py
rlangman: Replaced 1 with a variable.

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
rlangman: So far as I can tell, the only reason to have audio codes represented as `long` type is for indexing into embedding time, and computing softmax loss.

Currently the code does this by converting audio codes to long at every instance in the code where an audio code tensor is created. I have updated it so that long conversion only happens at the specific parts of the code that require it.

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
rlangman: For reference, it looks like reading `cut.recording` only works if you have the `flac/wav` file on the file system.

nemo/collections/tts/models/magpietts.py
rfejgin: Thanks @rlangman for going through the extra validation. Please run evaluation once training is done or long enough (say 300k steps). If metrics are similar I'm good with merging. For reference: with a 4x-stacked model with 5-second context, trained on English, these are the metrics I get (though you could also use your pre-change model as baseline):

**Unseen**
SSIM=0.70
WER=1.2%
UTMOSv2=3.55
FCD=0.681

**Seen**
SSIM=0.77
WER=1.05%
UTMOSv2=3.53
FCD=0.594

(if you used a longer context of 10 seconds the metrics will be better)

nemo/collections/tts/models/magpietts.py
```diff
@@ -777,62 +778,77 @@ def load_state_dict(self, state_dict, strict=True):
...
+    def remove_eos_token(self, codes, codes_len):
+        # codes: (B, C, T')
+        # codes_len: (B,)
+        codes_len = codes_len - 1
+        codes = codes[:, :, :-1]
+        mask = get_mask_from_lengths(lengths=codes_len)
+        codes = codes * mask.unsqueeze(1)
+        return codes, codes_len
+
+    def remove_embedded_eos_token(self, embedded, embedded_len):
```
rfejgin, line 930: regarding `remove_embedded_eos_token()`: It might make sense to add a comment saying that when using frame stacking, this embedding may be formed out of a mix of EOS and non-EOS tokens (but that is still correct, as discussed). Maybe a better name would be just "remove_final_embedding"

nemo/collections/tts/data/text_to_speech_dataset.py
rfejgin: yes, agreed I feel we have some technical debt here and should find a way to de-duplicate and unify as much of that code as possible

rfejgin: I've opened an internal JIRA ticket to keep track of the need to improve this.

nemo/collections/tts/models/magpietts.py
rfejgin: @rlangman One additional thing worth checking with regards to frame stacking:
Listen to a few output samples from the 4x-stacked model. Look for cutoff in the last word of the utterance - there shouldn't be any (or at least not more than a non-stacked model). Cutoff mid-last-word is one issue we had when  EOS related corner cases were not handled correctly (earlier on), and the refactored code touched those code areas. Maybe also look for first word cutoff in case something similar happens with BOS.

rlangman: This is the evaluation on LibriTTS unseen speakers.

Before code change:
```
SSIM: 0.702
CER: 1.25%
WER: 2.16%
UTMOSv2: 3.23
```
After code change:
```
SSIM: 0.696
CER: 0.96%
WER: 1.84%
UTMOSv2: 3.24
```
Compared to equivalent non-frame stacked versions, there is a drop in speaker similarity (0.8 -> 0.7). There is also higher CER (0.3% -> 1%) due to the frame-stacked models being prone to earlier termination. However the results between the two frame-stacked models are similar.

rfejgin: Okay, great! Thanks for doing the extra work to verify this.

The lower SSIM for the 4x-frame-stacked model compared to non-frame-stacked is not surprising, especially since this is a small local transformer and we have not done the "repeat context" trick. The fact that the two frame-stacked models have similar stats is what we were looking for and found here.

As discussed offline, I am also training a pair of 4x-stacked, large-LT, long-context models which we can use to do extra verification later on in en even more more controlled setting (without termination issues). But no need to wait for that, these results are already quite convincing.

Good to merge from my end.

blisc (approved):

PR #15196, opened 2025-12-16
Switch Nemotron 15/340b to always use Nemotron Tokenizer.

sudostock:
Align Nemotron to always use Nemotron tokenizer, in offline mode or online mode.


bdubauski (approved):
LGTM

PR #15200, opened 2025-12-17
[lhotse] Added support for re-weighting datasets with temperature on the fly.

XuesongYang:
## Summary

Add on-the-fly temperature-based dataset reweighting to the Lhotse dataloader, allowing users to control sampling distributions across nested `input_cfg` groups without manually recalculating weights each time datasets are added or removed. 

The feature applies the formula `w_i^τ / Σ w_j^τ` at each nesting level, where `τ` is configurable per level. Validation is strict: `reweight_temperature` must be either a scalar (broadcast to all levels) or a list whose length exactly matches the `input_cfg` nesting depth.

ŵᵢ = wᵢ^τ / Σⱼ wⱼ^τ.

- `τ = 1.0`: Preserves original weight ratios (neutral) 
- `τ = 0.0`: Equalizes all datasets regardless of original weights
- `0 < τ < 1.0`: Over-samples smaller datasets relative to larger ones

### New functionality (`cutset.py`)
- `temperature_reweighting(weights, temperature)`: applies temperature scaling and normalizes weights.
- `count_input_cfg_levels(config)`: computes maximum `input_cfg` nesting depth.
- Validation in `read_dataset_config()`: scalar is broadcast with a logged warning; list length mismatch raises `ValueError`.
- `parse_and_combine_datasets()` pops the current-level temperature and passes remaining values to child groups via `propagate_attrs`.

### Config (`dataloader.py`)
- Added `reweight_temperature` field to `LhotseDataLoadingConfig` (default `None`)

### Tests
- Unit tests for `temperature_reweighting()` and `count_input_cfg_levels()`
- Integration tests for dataloader with various temperature configurations
- Validation tests for scalar/list normalization behavior

### Documentation (`docs/source/audio/configs.rst`)
- New "Dataset Reweighting with Temperature" section with formula, configuration examples (scalar and list), nesting depth explanation, and a multi-task-group walkthrough.

### Example configs
- `magpietts_lhotse.yaml` / `magpietts_lhotse_moe.yaml`: added `reweight_temperature: 1.0` (neutral default) with comment showing list alternative
- Example of preserving top-level ratios but equalize within sub-groups:
```yaml
reweight_temperature: [1.0, 0.0]  # Level 1: preserve, Level 2: equalize
```

### DIFF
--- a/docs/source/broken_links_false_positives.json
+++ b/docs/source/broken_links_false_positives.json
@@ -73,3 +73,5 @@
 { "filename": "audio/datasets.rst", "lineno": 27, "status": "broken", "code": 0, "uri": "./api.html#datasets", "info": "" }
 { "filename": "audio/configs.rst", "lineno": 17, "status": "broken", "code": 0, "uri": "./api.html#datasets", "info": "" }
 { "filename": "tts/configs.rst", "lineno": 19, "status": "broken", "code": 0, "uri": "./api.html#datasets", "info": "" }
+{ "filename": "tts/datasets.rst", "lineno": 1, "status": "broken", "code": 0, "uri": "https://zenodo.org/record/7265581/files/ThorstenVoice-Dataset_2022.10.zip?download=1", "info": "403 Client Error: Forbidden for url: https://zenodo.org/record/7265581/files/ThorstenVoice-Dataset_2022.10.zip?download=1" }
+{ "filename": "tts/datasets.rst", "lineno": 1, "status": "broken", "code": 0, "uri": "https://zenodo.org/record/5525342/files/thorsten-neutral_v03.tgz?download=1", "info": "403 Client Error: Forbidden for url: https://zenodo.org/record/5525342/files/thorsten-neutral_v03.tgz?download=1" }


examples/tts/conf/magpietts/magpietts_lhotse.yaml
```diff
@@ -87,6 +87,7 @@ model:
       shuffle: true
       num_workers: 6
       pin_memory: true
+      reweight_temperature: [1.0, 0.0]  #  # Temperature for re-weighting datasets. 1 is a neutral value. Lower temperature over-samples smaller datasets, and vice versa. Can be specified multiple times to apply a different temperature to each group level in the YAML config.
```
XuesongYang:
```suggestion
      reweight_temperature: [1.0, 0.0]  # Temperature for re-weighting datasets. 1 is a neutral value. Lower temperature over-samples smaller datasets, and vice versa.
```

tests/collections/common/test_lhotse_temperature_reweighting.py
```diff
@@ -0,0 +1,277 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Unit and integration tests for temperature-based data reweighting in Lhotse dataloader.
+Integration tests mock data loading and verify that weights passed to mux() are correct.
+"""
+from unittest.mock import MagicMock, patch
+
+import numpy as np
+import pytest
```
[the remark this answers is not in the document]
XuesongYang:
```suggestion

```

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -20,6 +20,7 @@
 from pathlib import Path
 from typing import KeysView, Mapping, Sequence, Tuple, Union
```
XuesongYang:
```suggestion
from typing import KeysView, List, Mapping, Sequence, Tuple, Union
```

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -45,6 +46,34 @@
 from nemo.collections.common.parts.preprocessing.manifest import get_full_path
 
 
+def temperature_reweighting(weights: list[float | int], temperature: float = 1.0):
```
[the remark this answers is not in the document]
XuesongYang:
```suggestion
def temperature_reweighting(weights: List[Union[float, int]], temperature: float = 1.0) -> List[float]:
```

tests/collections/common/test_lhotse_temperature_reweighting.py
```diff
@@ -0,0 +1,282 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Unit and integration tests for temperature-based data reweighting in Lhotse dataloader.
+Integration tests mock data loading and verify that weights passed to mux() are correct.
+"""
+from unittest.mock import MagicMock, patch
+
+import numpy as np
```
XuesongYang, line 28:
```suggestion
import numpy as np
import pytest
```

examples/tts/conf/magpietts/magpietts_lhotse.yaml
```diff
@@ -87,6 +87,7 @@ model:
       shuffle: true
       num_workers: 6
       pin_memory: true
+      reweight_temperature: [1.0, 0.0]  # Temperature for re-weighting datasets. 1 is a neutral value. Lower temperature over-samples smaller datasets, and vice versa.
```
blisc: Where is the documentation for what this parameter does? And can we put it in the same place as the other arguments?

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -346,6 +382,14 @@ def parse_and_combine_datasets(
     weights = []
     tarred_status = []
 
+    # Extract the temperature for re-weighting datasets.
+    if not propagate_attrs["reweight_temperature"]:
+        temperature = 1.0
+        next_temperatures = None
+    else:
+        temperature, *next_temperatures = propagate_attrs["reweight_temperature"]
+    propagate_attrs["reweight_temperature"] = next_temperatures
```
blisc, line 528: What happens if the length of the reweight_temperature list is not equal to the number of datasets in the config?
What happens if reweight_temperature is just a float?

examples/tts/conf/magpietts/magpietts_lhotse.yaml
pzelasko: I think you also need to add this to `dataloader.py` in `LhotseDataLoadingConfig` otherwise it won't get propagated to `cutset.py` functions. We should also have a test that uses this option through `get_lhotse_dataloader_from_config` to check it works.

pzelasko: + we should support this value being a float (same temperature for all levels) and also define what happens when this list is shorter than the number of nested groups in `input_cfg` (e.g., the last temperature is re-used for each next level).

XuesongYang: good point. indeed, my test using `get_lhotse_dataloader_from_config` failed to pass this new param to `propagate_attrs`. 

Fixed now.

XuesongYang: > Where is the documentation for what this parameter does? And can we put it in the same place as the other arguments?

good point. The only place where this new param fits in seems here: https://github.com/NVIDIA-NeMo/NeMo/blob/main/docs/source/audio/configs.rst#lhotse-cutset. i will add description there.

XuesongYang: > * we should support this value being a float (same temperature for all levels) and also define what happens when this list is shorter than the number of nested groups in `input_cfg` (e.g., the last temperature is re-used for each next level).

i intentionally designed this with a strict `List` requirement to prioritize explicit configuration over flexibility. This ensures users are fully aware of their nested data structures and define temperatures intentionally.

but I agree that we need to validate that the number of nested groups matches the list length. I'll look into how to enforce that check.

nemo/collections/common/data/lhotse/cutset.py
XuesongYang: > What happens if the length of the reweight_temperature list is not equal to the number of datasets in the config?

The `reweight_temperature` list length only needs to match the number of nested groups.

XuesongYang: > What happens if reweight_temperature is just a float?

similar question as @pzelasko, pls check my answer above.

If both of you voted for supporting a float value, meaning apply the same temperature to all levels of nested group. i can make changes accordingly.

examples/tts/conf/magpietts/magpietts_lhotse.yaml
XuesongYang: addressed all above comments.

nemo/collections/common/data/lhotse/cutset.py
XuesongYang: added support of scalar.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -16,7 +16,7 @@
 import warnings
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Optional, Sequence, Union
+from typing import Any, List, Optional, Sequence, Union
```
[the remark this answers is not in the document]
XuesongYang:
```suggestion
from typing import Any, Optional, Sequence, Union
```

examples/tts/conf/magpietts/magpietts_lhotse.yaml
subhankar-ghosh: This is my understanding, correct me if I am wrong: Level 1 (Language): preserves the weight, Level 2 (datasets within each language): equalizes.
Based on this understanding, don't we want to normalize the weights based on number of shar files/size of the datasets at level 2?

XuesongYang: you're right. But this is an example demonstrating how to use this new param. In practice, we should override this one based on your needs.

docs/source/audio/configs.rst
```diff
@@ -128,6 +128,159 @@ An example train dataset in Lhotse shar format can be configured as follows:
...
+
+.. code-block:: yaml
+
+    train_ds:
+      use_lhotse: true
+      reweight_temperature: [1.0, 0.5, 0.0]  # Trimmed to [1.0] for 1 level
+      input_cfg:
+        - type: lhotse_shar
+          shar_path: /path/to/dataset1
+          weight: 100
```
blisc: We should support 1 and 2. 3 and 4 should throw an error

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -216,12 +249,34 @@ def read_dataset_config(config) -> tuple[CutSet, bool]:
         "force_map_dataset": config.get("force_map_dataset", False),
         "force_iterable_dataset": config.get("force_iterable_dataset", False),
         "slice_length": config.get("slice_length", None),
+        # Temperature for re-weighting datasets. 1 is a neutral value. Lower temperature over-samples smaller datasets, and vice versa.
+        "reweight_temperature": config.get("reweight_temperature", None),
     }
-    input_cfg = config.input_cfg
-    if isinstance(input_cfg, (str, Path)):
```
pzelasko, line 221: Why is this branch removed completely?

nemo/collections/common/data/lhotse/cutset.py
```diff
@@ -216,12 +249,34 @@ def read_dataset_config(config) -> tuple[CutSet, bool]:
...
-        # Resolve /path/to/input_cfg.yaml into config contents if needed.
-        input_cfg = OmegaConf.load(input_cfg)
-    cuts, is_tarred = parse_and_combine_datasets(input_cfg, propagate_attrs=propagate_attrs)
+
+    # Validate and standardize reweight_temperature to a list matching the nesting depth.
```
pzelasko: This entire check should be extracted to a separate function `validate_and_propagate_reweight_temperature(config, propagate_attrs)` to read more cleanly

tests/collections/common/test_lhotse_temperature_reweighting.py
```diff
@@ -0,0 +1,507 @@
...
+                "reweight_temperature": [1.0],  # Only 1 temperature for 2 levels
+                "shuffle": True,
+                "shard_seed": 42,
+            }
+        )
+
+        with pytest.raises(ValueError, match="does not match"):
+            read_cutset_from_config(config)
+
+    def test_correct_temperature_list_length_no_warning(self):
```
pzelasko, line 478: If this test mocks logging, how does it test that no warning is emitted? I'm confused.

tests/collections/common/test_lhotse_dataloading.py
```diff
@@ -2876,3 +2876,350 @@ def test_dataloader_nemo_tarred_slice_length_multi_epoch_different_sample(
...
+    # Sample multiple batches and count occurrences
+    dataset_counts = Counter()
+    for batch in islice(dl, 100):
+        for cut in batch:
+            dataset_counts[cut.dataset_name] += 1
+
+    total = sum(dataset_counts.values())
+
+    # Level 1: Leaf gets 30%, Nested group gets 70% (temperature=1.0 preserves weights)
+    assert dataset_counts["Leaf"] / total == pytest.approx(0.3, abs=0.1)
```
pzelasko, line 3232: are these tolerances wide enough for the test not to be flaky?

nemo/collections/common/data/lhotse/cutset.py
XuesongYang: it is redundant because `parse_and_combine_datasets()` already handles string/Path resolution as in Line 516 - 519:
```python
    if isinstance(config_list, (str, Path)):
        # Resolve local filepath /path/to/input_cfg.yaml or
        # remote url s3://bucket/path/to/input_cfg.yaml into config contents if needed.
        config_list = OmegaConf.create(load_yaml(config_list))
```

nemo/collections/common/data/lhotse/cutset.py
XuesongYang: agreed. It looks distracted a little. Let me wrap into a separate fucntion.

tests/collections/common/test_lhotse_temperature_reweighting.py
XuesongYang: Good question! The test patches logging with a `MagicMock`, then uses `mock_logging.warning.assert_not_called()` to verify that no warning was emitted. It confirms the "happy path" when the list length exactly matches the nesting depth, so `logging.warning` should never fire.

this is just a "runs without error" check. Happy to remove it if you think it is not necessary.

tests/collections/common/test_lhotse_dataloading.py
XuesongYang: fair concern, but statistically, `abs=0.1` with 100 batches of 4 (so 400 samples total) is about a 4-sigma tolerance window, which is very wide.

you can model this as 400 independent bernoulli trials, where each sample is "leaf" with prob `p=0.3`. Let `X` be the number of "leaf" outcomes. Then,
- `X ~ Binomial(n=400, p=0.3)`
- `E[X] = np = 120`
- `SD[X] = sqrt(np(1-p)) = 9.165`

the test, however, checks the observed propotion `p_hat = X / 400`, not the raw count. so for sample propotion, 
- `E[p_hat] = p = 0.3`
- `SD[p_hat] = sqrt(p(1-p)/n) = 0.023`

with tolerance `abs=0.1`, the test passes when `p_hat` falls in `[0.2, 0.4]`, converting bounds to z-scores,
- lower bounds: (0.2 - 0.3) / 0.023 = -4.348
- upper bounds: (0.4 - 0.3) / 0.023 = +4.348

so the test only fails if the observed `p_hat` deviates by more than 4.348 \sigma from expected value (4-sigma rules). This is an extreme unlikely event though.

pzelasko: thanks for a detailed explanation. let's fix the seed in this test just to be 100% sure (statistical property tests can be very irritating). I have a pytest fixture in lhotse for these cases https://github.com/lhotse-speech/lhotse/blob/master/lhotse/testing/random.py#L9

nemo/collections/common/data/lhotse/cutset.py
pzelasko: good catch, thanks

nemo/collections/common/data/lhotse/cutset.py
XuesongYang: done!

tests/collections/common/test_lhotse_dataloading.py
XuesongYang: done!

PR #15210, opened 2025-12-19
[TTS][MagpieTTS] Longform TTS using MagpieTTS

subhankar-ghosh:
> [!IMPORTANT]  

# What does this PR do ?

 Add Longform TTS support in MagpieTTS.

TTS

# Changelog

- LongFormTTSInferenceDataset - A new dataset class subclass of MagpieTTSDataset, that splits input text into chunks based on sentence delimiters. This minimal class handles inference-only data: context_audio_codes/audio, speaker_indices, and chunked text.
- --longform_mode option - Added to examples/tts/magpietts_inference.py to enable longform inference with auto-detection based on text length. Nothing changes for user running inference script.
- Longform methods in MagpieTTSModel - Core logic in `generate_long_form_speech` and `construct_longform_inference_prior`, plus helper methods for managing state across chunks.
- Dataclass-based state management - Uses LongformConfig for immutable constants and LongformChunkState for mutable state that persists across chunk iterations (per @blisc's suggestion).
- Unified inference runner - MagpieInferenceRunner now handles both standard and longform inference with auto-detection, eliminating the need for a separate LongFormInferenceRunner class.

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/.github/workflows/cicd-main-speech.yml
+++ b/.github/workflows/cicd-main-speech.yml
@@ -193,6 +193,8 @@ jobs:
             script: L2_TTS_InferEvaluate_Magpietts_ZeroShot
           - runner: self-hosted-azure
             script: L2_TTS_InferEvaluate_Magpietts_SeenSpeakers
+          - runner: self-hosted-azure
+            script: L2_TTS_InferEvaluatelongform_Magpietts_ZeroShot
     needs: [unit-tests]
     runs-on: ${{ matrix.runner }}
     name: ${{ matrix.is-optional && 'PLEASEFIXME_' || '' }}${{ matrix.script }}
--- a/tests/functional_tests/L2_TTS_InferEvaluatelongform_Magpietts_ZeroShot.sh
+++ b/tests/functional_tests/L2_TTS_InferEvaluatelongform_Magpietts_ZeroShot.sh
@@ -1,0 +1,33 @@
+# Copyright (c) 2020-2025, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/magpietts_inference.py \
+    --codecmodel_path /home/TestData/tts/AudioCodec_21Hz_no_eliz_without_wavlm_disc.nemo \
+    --datasets_json_path examples/tts/evalset_config.json \
+    --out_dir ./mplf_zs_0 \
+    --batch_size 4 \
+    --longform_mode always \
+    --longform_max_decoder_steps 50000 \
+    --use_cfg \
+    --cfg_scale 2.5 \
+    --num_repeats 1 \
+    --temperature 0.6 \
+    --hparams_files /home/TestData/tts/2506_ZeroShot/lrhm_short_yt_prioralways_alignement_0.002_priorscale_0.1.yaml \
+    --checkpoint_files /home/TestData/tts/2506_ZeroShot/dpo-T5TTS--val_loss=0.4513-epoch=3.ckpt \
+    --legacy_codebooks \
+    --legacy_text_conditioning \
+    --apply_attention_prior \
+    --run_evaluation \
+    --clean_up_disk \
+    --cer_target 0.25 \
+    --ssim_target 0.7


nemo/collections/tts/models/magpietts.py
```diff
@@ -2547,22 +2580,53 @@ def get_most_attended_text_timestep(
...
             if attended_timestep_counter[bidx].get(last_attended_timestep, 0) >= 8:
                 # This is probably an attention sink! Move to the next timestep
                 last_attended_timestep += 1
+            last_attended_timestep_in_this_window = last_attended_timestep - left_offset[bidx]
             window_size = lookahead_window_size
-            window_end = min(last_attended_timestep + window_size, text_lens[bidx] - 3)  # Ignore the last 3 timesteps
-            item_attention_scores = alignment_attention_scores[bidx, last_attended_timestep:window_end]
+            window_end = min(
+                last_attended_timestep_in_this_window + window_size, text_lens[bidx] - 3
+            )  # Ignore the last 3 timesteps
```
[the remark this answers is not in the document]
subhankar-ghosh: Already existing, not part of longform PR.

examples/tts/magpietts_inference.py
```diff
@@ -166,8 +172,12 @@ def run_inference_and_evaluation(
...
     full_checkpoint_name = f"{checkpoint_name}_{inference_config.build_identifier()}_SV_{eval_config.sv_model}"
 
-    # Create inference runner
-    runner = MagpieInferenceRunner(model, inference_config)
+    # Create appropriate inference runner based on longform flag
+    if longform:
+        logging.info("Using longform inference mode (sentence-by-sentence processing)")
+        runner = LongFormInferenceRunner(model, inference_config)
+    else:
+        runner = MagpieInferenceRunner(model, inference_config)
```
blisc: Do we need this? Can we natively switch to longform once we go over the 20s of decoder generation?

nemo/collections/tts/models/magpietts.py
```diff
@@ -3369,7 +3433,7 @@ def do_tts(
             "en": ["english_phoneme", "english"],
             "de": ["german_phoneme", "german"],
             "es": ["spanish_phoneme", "spanish"],
-            "fr": ["french_phoneme", "french"],
+            "fr": ["french_chartokenizer", "french"],
```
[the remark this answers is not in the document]
blisc: Can you merge with main?

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -798,3 +799,300 @@ def collate_fn(self, batch: List[dict]):
         chosen_collated = super().collate_fn(chosen_batch)
         rejected_collated = super().collate_fn(rejected_batch)
         return {"chosen": chosen_collated, "rejected": rejected_collated}
+
+
+class LongFormTTSInferenceDataset(torch.utils.data.Dataset):
```
blisc: What is the difference between this dataset and the magpiettsdataset? From the docstring, it looks like the only difference is that it is model agnostic. In which case, why not just update magpiettsdataset and make it more generic: eg. EncodedTTSDataset or something to that effect

nemo/collections/tts/models/magpietts.py
```diff
@@ -3419,3 +3484,807 @@ def do_tts(
     @classmethod
     def list_available_models(cls) -> List[PretrainedModelInfo]:
         return []
+
+    def set_longform_inference_variables(self, batch_size=1):
```
blisc: Does this need to be part of the magpie class? IE do you have to save these parameter to self? It looks like a good use case for another dataclass

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -400,3 +403,379 @@ def compute_mean_rtf_metrics(rtf_metrics_list: List[dict]) -> Dict[str, float]:
             mean_metrics[key] = float(sum(values) / len(values)) if values else 0.0
 
         return mean_metrics
+
+
+class LongFormInferenceRunner:
```
blisc: This class shares a few functions with MagpieInferenceRunner. Why not subclass MagpieInferenceRunner and avoid duplicating code?

nemo/collections/tts/data/text_to_speech_dataset.py
subhankar-ghosh: LongFormTTSInferenceDataset returns chunked texts and text lens (https://github.com/NVIDIA-NeMo/NeMo/blob/magpietts_os_longform/nemo/collections/tts/data/text_to_speech_dataset.py#L905-L906)
Which is very different from how we return text from magpiettsdataset. Other than that mostly they should be similar.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Looking into this. That is a good suggestion.

nemo/collections/tts/modules/magpietts_inference/inference.py
subhankar-ghosh: The `_configure_tokenizer` and `__init__` are the same. Other methods - `create_dataset` and `run_inference_on_dataset`, only the signature is the same, the logic is very different. I can subclass MagpieInferenceRunner
`class LongFormInferenceRunner(MagpieInferenceRunner)`

examples/tts/magpietts_inference.py
subhankar-ghosh: We have to predetermine LongForm or standard generation because for LongForm we need to save and update history variables, which are necessary for the window mechanism. However, to make the user experience seemless I can try to write a logic that would determine LongForm or standard generation given the number of words in the input text (~40-50 words in 20sec). What do you think?

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: We can use a dataclass but the dataclass object has to be a class variable. This is because these state variables need to be maintained/updated as `generate_long_form_speech` is called for different chunks of text. These state variables cannot be local to `generate_long_form_speech` method.

```
for chunk_id, chunk in enumerate(chunk_of_text_list):
    model.generate_long_form_speech(chunk, ...)
    # State variables (current implementation or through a dataclass) needs to be maintained.
```

examples/tts/magpietts_inference.py
blisc: Is it not possible to initialize these parameters on the fly?

subhankar-ghosh: This does batch processing, so determining if each datapoint is longform or short and running the corresponding Runner would be complicated. `MagpieInferenceRunner` cannot do longform, `MagpieTTSDataset` cannot be used for longform. But it might be possible to do standard and longform with `LongFormInferenceRunner`.

subhankar-ghosh: Let me try: 
1) The auto-detect logic -> based on the input manifest, if any of the entries is longform (`len(text) > 50 words`) we use longform else standard inference.
2) Merge the two inference runners into one single runner.

nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -798,3 +799,300 @@ def collate_fn(self, batch: List[dict]):
...
+            if self.use_text_conditioning_encoder:
+                text_context_tokens = self.text_tokenizer.encode(
+                    "[NO TEXT CONTEXT]",
+                    "english_phoneme" if self.tokenizer_name is None else self.tokenizer_name,
+                )
+                if self.pad_context_text_to_max_duration:
+                    required_len = (
+                        int(self.context_duration_max * self.sample_rate / self.codec_model_samples_per_frame) + 2
+                    )
+                    pad_id = self.text_tokenizer.tokenizer_pad_ids.get("english_phoneme", 0)
```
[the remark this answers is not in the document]
subhankar-ghosh: This is the same.

examples/tts/magpietts_inference.py
subhankar-ghosh: @blisc Please check implementation now. It is much cleaner and reusing most of the existing code. User experience is also seamless as they do not need to decide between longform path or standard path.

nemo/collections/tts/data/text_to_speech_dataset.py
subhankar-ghosh: Subclassed

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Now we represent all these variables through a dataclass and maintain this state.

nemo/collections/tts/modules/magpietts_inference/inference.py
subhankar-ghosh: Now we have a unified runner class.

nemo/collections/tts/models/magpietts.py
```diff
@@ -3419,3 +3556,820 @@ def do_tts(
...
+            finished_texts_counter: Dict to update in-place.
+        """
+        is_near_end = attended_pos >= text_len - self.longform_config.near_end_threshold
+
+        # Text is unfinished if not near end AND not already marked finished
+        unfinished_texts[batch_idx] = not is_near_end and not is_finished
+
+        # Start counting when near end or already finished
+        if is_near_end or is_finished:
+            finished_texts_counter.setdefault(batch_idx, 0)
```
blisc, line 3689: These two functions look very similar to the non-longform counterparts. While not a blocker for this PR, we should attempt to merge both codepaths together when possible.

subhankar-ghosh:
> Please add back https://github.com/NVIDIA-NeMo/NeMo/blob/magpietts_2508/tests/functional_tests/L2_TTS_InferEvaluateStreaming_Magpietts_ZeroShot.sh and then we can merge

Added.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: I will try to incorporate this in the next PR.

PR #15213, opened 2025-12-20
[TTS] Allow inference without reference audio

rfejgin:
In some cases we need to run inference on manifests that do not include context audio and/or ground truth audio:

* Text context manifests may not have a context audio included (and it wouldn't be very relevant anyway)
* When generating arbitrary text, the ground truth audio cannot be assumed to be available.

Note that even without the context and ground truth there are useful metrics that can be calculated, like WER, UTMOS, and inference speed.

This PR adapts the inference scripts and the data loader used by them to allow the absence of context and/or ground truth audio. When these is missing, the the metrics that depend on them are set to `0.0`.

Note: this PR replaces an [earlier PR]( https://github.com/NVIDIA-NeMo/NeMo/pull/15139) that was based off `magpietts_2508`. We are now merging directly into `main` instead.

### DIFF
--- a/nemo/collections/tts/data/text_to_speech_dataset.py
+++ b/nemo/collections/tts/data/text_to_speech_dataset.py
@@ -776,9 +776,9 @@ class MagpieTTSDataset(TextToSpeechDataset):
         if len(speaker_indices_list) > 0:
             batch_dict['speaker_indices'] = torch.tensor(speaker_indices_list, dtype=torch.int64)
 
-        # Assert no more than one of audio or audio_filepath in the batch
+        # Assert no more than one of audio or audio_codes in the batch
         if 'audio' in batch_dict:
-            assert 'audio_filepath' not in batch_dict
+            assert 'audio_codes' not in batch_dict
 
         # Assert no more than one of context_audio or context_audio_codes in the batch
         if 'context_audio' in batch_dict:


nemo/collections/tts/data/text_to_speech_dataset.py
```diff
@@ -762,14 +764,13 @@ def collate_fn(self, batch: List[dict]):
...
+        # Assert no more than one of audio or audio_filepath in the batch
+        if 'audio' in batch_dict:
+            assert 'audio_filepath' not in batch_dict
 
-        # Assert only ONE of context_audio or context_audio_codes in the batch
+        # Assert no more than one of context_audio or context_audio_codes in the batch
         if 'context_audio' in batch_dict:
             assert 'context_audio_codes' not in batch_dict
-        if 'context_audio_codes' in batch_dict:
-            assert 'context_audio' not in batch_dict
```
blisc: The comments no longer match the logic, but the real question is why do we care that the dataset only returns one of audio vs audio_codes? Should we just remove these checks?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -318,19 +326,29 @@ def evaluate(
...
-                gt_speaker_embedding, pred_speaker_embedding, dim=0
-            ).item()
+            # Initialize SSIMs with a default since the contest or ground truth audio
+            # may be unavailable.
+            pred_context_ssim = 0.0
+            gt_context_ssim = 0.0
+            pred_context_ssim_alternate = 0.0
+            gt_context_ssim_alternate = 0.0
+            pred_gt_ssim = 0.0
+            pred_gt_ssim_alternate = 0.0
```
blisc: While this seems ok, wouldn't it be better to default to None or NaN instead? Or will that break some averaging logic down the road?
Might be a better idea to stick to None or NaN and check that the dataset of interest does not contain context audios.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -265,17 +265,25 @@ def evaluate(
...
                     pred_text = process_text(pred_text)
-                    gt_audio_text = asr_model.transcribe([gt_audio_filepath], batch_size=1, use_lhotse=False)[0].text
-                    gt_audio_text = process_text(gt_audio_text)
+                    if gt_audio_filepath is not None:
+                        gt_audio_text = asr_model.transcribe([gt_audio_filepath], batch_size=1, use_lhotse=False)[
+                            0
+                        ].text
+                        gt_audio_text = process_text(gt_audio_text)
+                    else:
+                        gt_audio_text = ""
```
blisc: Same question as SSIM, why not default to None? I'm guessing this still goes through the WER calculations, and empty string is a valid result from ASR (ie in the case of empty audio or silent audio)

blisc (changes requested):

nemo/collections/tts/data/text_to_speech_dataset.py
rfejgin: I suspect that these checks were put there since if both the audio waveform and the audio codes are provided, it's ambiguous which of the two should be used by the model; we wouldn't one part of the code choosing to use the audio and another to use the codes. So it simplifies things to know that only (at most) one is present and any time.

About the comments not matching the logic: which part?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: `None` would probably mess up subsequent logic (averaging and logging) but `NaN` should work, I will change it.
(it would be tricky to check if a dataset contains ground truth and context audio as this can vary per manifest entry)

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: I can change it to None and later replace CER+WERs with `nan`. It does mean that if even one entry in a manifest is missing the ground truth audio, the final WER (as well as WER averages across datasets) will be NaN. But maybe that is the correct thing to do.

nemo/collections/tts/data/text_to_speech_dataset.py
blisc: > About the comments not matching the logic: which part?

I think I was mistaken; the comments do match the logic.

Now that I read this again shouldn't `# Assert no more than one of audio or audio_filepath in the batch` be audio or **audio_codes** instead?

rfejgin: You are right, good catch!

Fixed.

blisc (approved):

PR #15223, opened 2025-12-23
[TTS] MagpieTTS: Implement Frechet Codec Distance metric + some minor inference bugfixes

rfejgin:
# What does this PR do ?

Adds the Frechet Codec Distance metric and integrates it in MagpieTTS inference scripts. Also fixes some minor MagpieTTS inference bugs.

**Collection**: TTS

# Changelog

The Frechet Distance (FD) is commonly used to evaluate generative models (e.g. Frechet Inception Distance, Frechet Audio Distance). In this PR we implements  FD in the embedding space of a neural codec. This is a metric that measures how closely the distributions of real and generated codec frames match, at the single frame level.

Changes:
* `frechet_codec_distance.py`: An implementation of FD in codec embedding space. Builds on TorchMetrics' FID implementation. We provide the audio codec as a custom feature extractor.
* `test_frechet_coec_distance.py`: Unit test
* Integration of the FCD in MagpieTTS inference scripts. If desired, FCD calculation can be disabled using the `--disable_fcd` command line argument to `magpietts_inference.py`
* Inference bugfixes
  * fix a logging statement that was reporting errors due to incorrect formatting syntax
  * disable logging of thousands of messages during loading of the `titanet_small` speaker representation model. This was present in earlier versions of the inference scripts and appears to have been accidentally lost in recent refactorings 
  * Fix an issue where filewise metrics were not being filtered to a spcified subset as intended

**PR Type**:

- [x] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
+++ b/nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
@@ -194,12 +194,15 @@ def evaluate(
     sv_model_type="titanet",
     asr_model_name="stt_en_conformer_transducer_large",
     with_utmosv2=True,
+    with_fcd=True,
     codec_model_path=None,
 ):
     audio_file_lists = find_generated_audio_files(generated_audio_dir)
     records = read_manifest(manifest_path)
     assert len(audio_file_lists) == len(records)
-    if codec_model_path is not None:
+    if with_fcd:
+        if codec_model_path is None:
+            raise ValueError("codec_model_path is required when with_fcd is True")
         codes_file_lists = find_generated_codec_files(generated_audio_dir)
         assert len(codes_file_lists) == len(records)
 
@@ -240,10 +243,9 @@ def evaluate(
     speaker_verification_model_alternate = speaker_verification_model_alternate.to(device)
     speaker_verification_model_alternate.eval()
 
-    if codec_model_path is not None:
+    if with_fcd:
         fcd_metric = FrechetCodecDistance(codec_name=codec_model_path).to(device)
     else:
-        logging.info("No codec model provided, skipping FCD metric")
         fcd_metric = None
 
     if with_utmosv2:
@@ -405,7 +407,7 @@ def evaluate(
         fcd = fcd_metric.compute().cpu().item()
         fcd_metric.reset()
     else:
-        fcd = 0.0
+        fcd = float('nan')
 
     filewise_metrics_keys_to_save = [
         'cer',
--- a/examples/tts/magpietts_inference.py
+++ b/examples/tts/magpietts_inference.py
@@ -252,6 +252,7 @@ def run_inference_and_evaluation(
                 audio_base_dir=meta['audio_dir'],
                 save_cross_attention_maps=True,
                 save_context_audio=(repeat_idx == 0),  # Only save context audio once
+                save_predicted_codes=eval_config.with_fcd,  # Code files are only needed for FCD computation
             )
 
             # Compute mean RTF metrics
@@ -269,6 +270,7 @@ def run_inference_and_evaluation(
                 asr_model_name=eval_config.asr_model_name,
                 language=language,
                 with_utmosv2=eval_config.with_utmosv2,
+                with_fcd=eval_config.with_fcd,
                 codec_model_path=eval_config.codec_model_path,
             )
 
@@ -587,6 +589,7 @@ def main():
         sv_model=args.sv_model,
         asr_model_name=args.asr_model_name,
         with_utmosv2=not args.disable_utmosv2,
+        with_fcd=not args.disable_fcd,
         codec_model_path=args.codecmodel_path if not args.disable_fcd else None,
     )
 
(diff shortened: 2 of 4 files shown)


nemo/collections/tts/metrics/frechet_codec_distance.py
```diff
@@ -0,0 +1,193 @@
...
+        # We treat all frames as one large batch element, since the codec requires (B, C, T) input and
+        # we don't have the per-batch-element lengths at this point due to FID API limitations
+
+        # Consturct a length tensor: one batch element, all frames.
+        x_len = torch.tensor(x.shape[0], device=x.device, dtype=torch.long).unsqueeze(0)  # (1, 1)
+        tokens = x.permute(1, 0).unsqueeze(0)  # 1, C, B*T
+        embeddings = self.codec.dequantize(tokens=tokens, tokens_len=x_len)  # (B, D, T)
+        # we treat each time step as a separate example
+        embeddings = rearrange(embeddings, 'B D T -> (B T) D')
+        return embeddings
```
blisc, line 59: Don't we need some sort of masking here? If we are reducing B and T into one dimension, how do we ensure that no padding gets passed to the model?

blisc: If this only works on batch 1, we should add a check that x has a batch dimension of 1 or output an error or a warning

blisc (changes requested):

nemo/collections/tts/metrics/frechet_codec_distance.py
rfejgin: It works for all batch sizes. 

Removal of padding is taken care of in [FrechetCodecDistance.update()](https://github.com/NVIDIA-NeMo/NeMo/pull/15223/changes#diff-7d4075bbfa3672447ab46a1aa0d746f9f4a70b2429175e255341e53db9848128R146)

The way it works is:
`fcd.update()` constructs a large batch of shape `(B*T, C)` which eventually gets routed to the `CodecEmbedder`. All frames in this batch are valid; padding is explicitly removed. The `CodecEmbedder` then internally reshapes the tensor to `1,C, B*T` before feeding into the codec for dequantization.

The reason we need to do it this way is that `FrechetInceptionDistance`'s APIs don't allow passing in a second tensor indicating which frames are valid and not (as we would usually do using `tokens_len`); so to work around that we pass in a flattened tensor all of whose entries are valid.

rfejgin:
@subhankar-ghosh Could you please review just the latest commit in this PR? That part touches EvaluationConfig and longform inference. I mean the commit titled `Integrate FCD in longform inference and rework --disable_fcd`. Thanks!

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -240,10 +243,9 @@ def evaluate(
     speaker_verification_model_alternate = speaker_verification_model_alternate.to(device)
     speaker_verification_model_alternate.eval()
 
-    if codec_model_path is not None:
+    if with_fcd:
         fcd_metric = FrechetCodecDistance(codec_name=codec_model_path).to(device)
```
subhankar-ghosh, line 247: I am not sure if `from torchmetrics.image.fid import FrechetInceptionDistance` is installed in NeMo container by default. If it is not then we might want to check for it's availability and based on that log a warning message if it is not installed.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -405,7 +407,7 @@ def evaluate(
         fcd = fcd_metric.compute().cpu().item()
         fcd_metric.reset()
     else:
-        fcd = 0.0
+        fcd = float('nan')
```
subhankar-ghosh, line 410: Be careful about setting the None value for metrics, check how it is formatted in `def compute_mean_with_confidence_interval`. We should have a uniform default value for the condition `if <any_metric> is None:` case.

subhankar-ghosh (approved):

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
blisc: @rfejgin, can you add a test wrapper around this in another PR?

rfejgin: In [another PR](https://github.com/NVIDIA-NeMo/NeMo/pull/15213) @blisc asked to use `NaN` as a default value for metrics that don't get computed in the specific inference run, so I'm aligning with that. 

Confirmed that `compute_mean_with_confidence_interval` doesn't crash with this NaN value. We end up printing `nan`, which seems reasonable (or if there are repeats `nan +- nan`).

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: Thanks. I double checked and the import proceeds correctly on recent NeMo containers (`nemo:dev` and `nemo:25.11`).

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: > @rfejgin, can you add a test wrapper around this in another PR?
@blisc 

Sure. Do you mean to add a test that checks that when setting `--disable_fcd` the inference script doesn't crash?

PR #15225, opened 2025-12-23
Correct CONTRIBUTING.md spelling

yujis2000:


blisc:
Thanks for the PR, but it looks like we changed this file recently and removed this already.

Issue #14961, opened 2025-10-21
Encoder-decoder attention extraction in ASR transcribe


mgaido91:
**Is your feature request related to a problem? Please describe.**

There are many studies showing that the encoder-decoder can be used for auxiliary tasks (e.g. with DTW to get word-level timestamps, or for simultanoeus/streaming translation -- such as with the StreamAtt method). However, there is currently no way to get the encoder-decoder attention when running the beam search. Thsi kind of feature is instead available in other repositories (e.g. HF transformers).

**Describe the solution you'd like**

It would be great if the beam search generation would return also the attention mask and this would be included in the returned Hypothesis.

**Describe alternatives you've considered**

An alternative might be an additional pass on the decoder at the end of the generation only to compute the attention, but this would add unnecessary computational costs.

**Additional context**

nithinraok:
Hi Marco, 
thanks for the feature request. Will consider adding to upcoming releases.

nithinraok:
You should be able to access cross attention values here https://github.com/NVIDIA-NeMo/NeMo/blob/0b1be8d1165f49ee2ef1e74f72f2ff07350f6798/nemo/collections/asr/modules/transformer/transformer_decoders.py#L288 

Is this what you looking for?

andrusenkoau:
Hi @mgaido91,

You can find an example of this functionality in the streaming Canary decoding PR: [link](https://github.com/NVIDIA-NeMo/NeMo/pull/14765). We are using cross-attention outputs to guide the streaming decoding process for AlignAtt policy. You can easily extract the cross-attention weights as it is done here: [link1](https://github.com/andrusenkoau/NeMo/blob/9693311613e341b3d3e39a67b386f2b9a3d017e8/nemo/collections/asr/parts/submodules/aed_decoding/aed_batched_streaming.py#L229), [link2](https://github.com/andrusenkoau/NeMo/blob/9693311613e341b3d3e39a67b386f2b9a3d017e8/nemo/collections/asr/modules/transformer/transformer_generators.py#L151).

mgaido91:
[...]

PR #15229, opened 2025-12-24
Add cross-attention to output hypotheses

mgaido91:
> [!IMPORTANT]  

# What does this PR do ?

The PR adds the encoder-decoder cross-attention to the output hypotheses returned by ASR models.

**Collection**: ASR

# Changelog

- Returns the cross-attention scores in the output of the greedy generator
- Returns the cross-attention scores in the output of the beam search generator

# Usage

```python
from nemo.collections.asr.models import ASRModel
from nemo.collections.asr.models.aed_multitask_models import MultiTaskTranscriptionConfig
from nemo.collections.asr.parts.submodules.multitask_decoding import MultiTaskDecodingConfig

model = ASRModel.from_pretrained(model_name="nvidia/canary-1b-v2")
decoding_config = MultiTaskDecodingConfig()
decoding_config.return_xattn_scores = True
model.change_decoding_strategy(decoding_config)
config = MultiTaskTranscriptionConfig(
    batch_size=4,
    return_hypotheses=True,
    num_workers=0,
    verbose=False,
    prompt={'source_lang': 'en', 'target_lang': 'en'},
    enable_chunking=False
)
output = model.transcribe("/Users/mgaido/Downloads/vp-test/aa.wav", override_config=config)
assert output[0].xatt_scores is not None
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

@nithinraok @andrusenkoau 

# Additional Information

- Related to #14961 .


nemo/collections/asr/parts/utils/rnnt_utils.py
```diff
@@ -87,6 +87,9 @@ class Hypothesis:
     last_token (Optional): A token or batch of tokens which was predicted in the last step.
 
     last_frame (Optional): Index of the last decoding step hypothesis was updated including blank token prediction.
+
+    xatt_scores (Optional): List of cross-attention scores for each decoder layer. Each element of the list
```
nithinraok: Shouldn;t shape is List[`BxHxT1xT2`] . Also best to add: this is used with AED models

nemo/collections/asr/modules/transformer/transformer_generators.py
```diff
@@ -234,14 +235,19 @@ def _forward(
...
                 input_ids,
                 encoder_hidden_states,
                 encoder_input_mask,
                 decoder_mems_list,
                 i,
                 return_scores=return_beam_scores,
             )
+            if xatt_scores_list is not None:
+                for layer in range(len(xatt_scores_list)):
+                    xatt_scores_list[layer] = torch.cat((xatt_scores_list[layer], new_xatt_scores_list[layer]), dim=2)
```
nithinraok: what about condition when `new_xattn_scores_list` is None? cat would fail

nemo/collections/asr/modules/transformer/transformer_generators.py
nithinraok, line 374: could you update here as well and also include in returns tuple

nemo/collections/asr/modules/transformer/transformer_generators.py
```diff
@@ -478,6 +488,21 @@ def _forward(
             prefixes_ids = indices_i.unsqueeze(2).repeat(1, 1, p_len)
             prefixes = prefixes.gather(1, prefixes_ids).view(-1, p_len)
 
+            # select xatt scores corresponding to chosen hypotheses
+            if next_xatt_scores_list is not None:
+                num_heads = xatt_scores_list[0].shape[1]
```
nithinraok, line 502: check for `xatt_scores_list` if None

nemo/collections/asr/modules/transformer/transformer_generators.py
```diff
@@ -501,13 +526,26 @@ def _forward(
...
+                    .view(-1, self.beam_size, num_heads, tgt_len, src_len)
+                    .gather(1, xatt_best_guesses)
+                    .squeeze(1)
+                )
 
         if return_beam_scores:
-            return prefixes, scores * len_penalties, tgt
+            return prefixes, scores * len_penalties, tgt, xatt_scores_list
         else:
             return tgt
```
nithinraok, line 559: we might also return `xatt_scores_list` here as `return_beam_scores` is independent of return xattn_scores.

nemo/collections/asr/modules/transformer/transformer_generators.py
mgaido91: if `new_xattn_scores_list` is `None`, `xatt_scores_list` will stay `None`, so we never enter in this if

nemo/collections/asr/modules/transformer/transformer_generators.py
mgaido91: if `next_xatt_scores_list` is not None, then `xatt_scores_list` cannot be None. `_one_step_forward` either always returns None or always returns a not-None value

nemo/collections/asr/parts/utils/rnnt_utils.py
mgaido91: this is for a single hypothesis, so there is no `B`... So this would be `List[HxT1xT2]`. If you prefer I can rename `HxUxT` to `HxT1xT2`. I will also add the indication about AED models.

nemo/collections/asr/modules/transformer/transformer_generators.py
mgaido91: maybe I am missing something, but I see this being used with `return_beam_scores=True`, while `return_beam_scores` is set to False only in the validation step in `EncDecTransfModelBPE`, where there is no point in returning the attention scores. So I thought it would be better to avoid an API change for no reason. I can implement that, but just wanted to double check whether we really want this.

mgaido91:
@nithinraok thanks for your review. I do have a question, though. You said to add a param in `MultiTaskDecodingConfig`, however IIUC we should rather focus on `TranscribeConfig`, which is what is set through the `transcribe` API. Changing the `MultiTaskDecodingConfig ` might be tricky for a user. Also, there are some config tricks required in the `TrascribeConfig` (i.e. `enable_chunking=False`) to enable getting the attention. So if we add a param in `MultiTaskDecodingConfig`, it might be quite complex for a user to get the attention, as it would require changing two different configs. WDYT?

nemo/collections/asr/modules/transformer/transformer_generators.py
nithinraok: I meant to ask for each step, but probably its fine.

nithinraok:
> @nithinraok thanks for your review. I do have a question, though. You said to add a param in `MultiTaskDecodingConfig`, however IIUC we should rather focus on `TranscribeConfig`, which is what is set through the `transcribe` API. Changing the `MultiTaskDecodingConfig ` might be tricky for a user. Also, there are some config tricks required in the `TrascribeConfig` (i.e. `enable_chunking=False`) to enable getting the attention. So if we add a param in `MultiTaskDecodingConfig`, it might be quite complex for a user to get the attention, as it would require changing two different configs. WDYT?

Very good point. 

Well, first, the option need to be passed to `MultiTaskDecodingConfig` for sure. To make it independent of the behavior of transcribe inference or not. But however, I get your point. 
Two options:
* If the option needs to be through `TranscribeConfig` then we need to run `model.change_decoding_strategy(updated_cfg)` to change the default decoding behavior (which is return xattn_scores=None) inside .transcribe() each time. Which I don;t think is a good idea. 
* 2nd options is we create decoding cfg. like:
```python
multitask_decoding = MultiTaskDecodingConfig()
multitask_decoding.strategy = "greedy"
multitask_decoding.return_xattn_scores = True
```
and call
```python
asr_model.change_decoding_strategy(multitask_decoding)
```
before performing .transcribe()

Lets go with later option for now. I will keep thinking about this as this needs to be changed across. Are there any other options like these you are interested on to be changed through .transcribe()?

nemo/collections/asr/modules/transformer/transformer_generators.py
mgaido91: I got it but if `xatt_scores_list` is not None, then `new_xattn_scores_list` must be not None, otherwise there is something weird happening (ie. attention scores are returned for some tokens and for some others are None in the same generation). I'd rather add an assert on `new_xattn_scores_list`, WDYT?

mgaido91:
> Lets go with later option for now. I will keep thinking about this as this needs to be changed across. Are there any other options like these you are interested on to be changed through .transcribe()?

Ok, I will work on this in the next days. Maybe it would be worth adding some checks and logs to guide the user though. I will try to come up with a proposal for that while working on this. Thanks.

tests/collections/asr/decoding/test_multi_task_decoding.py
```diff
@@ -103,11 +111,11 @@ def test_greedy_decoding(inputs, nnet, deterministic_rng, with_confidence):
 
 
 def test_temperature_sampling_decoding(inputs, nnet):
-    gen = GreedySequenceGenerator(*nnet, temperature=10.0, n_samples=2)
+    gen = GreedySequenceGenerator(*nnet, return_xattn_scores=True, temperature=10.0, n_samples=2)
```
andrusenkoau: Could you add the check for both `return_xattn_scores` options (as above) here?

andrusenkoau (commented):
Hi @mgaido91, thank you for the great work! I have almost no questions. Let's wait your final changes with decoding config.

mgaido91:
Thanks @andrusenkoau !

> Let's wait your final changes with decoding config.

I already made them. You can find the `return_xattn_scores` in the decoding config. I also added a warning message to instruct the users. Is there something missing that you have in mind?


PS I do not see why `[CICD NeMo / Nemo_CICD_Test (pull_request)](https://github.com/NVIDIA-NeMo/NeMo/actions/runs/20991982507/job/60354816538?pr=15229)` is failing, I cannot see meaningful logs, maybe I just do not know where to look at as I am not that familiar with NeMo CI. It is also weird as it passed before I updated the branch with the master. May you point me to the logs?

tests/collections/asr/decoding/test_multi_task_decoding.py
mgaido91: yes, sure, I did not do it to minimize the CI cost, I am updating it, thanks!

mgaido91:
@nithinraok I think I addressed your comments, may you please take another look at this?

nemo/collections/asr/modules/transformer/transformer_generators.py
nithinraok: @mgaido91 could you update this one as well

nithinraok (approved):
Thanks Marco. Minor comment which you might have missed earlier. LGTM otherwise. Thanks for the PR, great work!

mgaido91:
@nithinraok @andrusenkoau thank you for your guidance and reviews!

PR #15254, opened 2026-01-05
Update MagpieTTS' Inference Parameter Configuration

blisc:
# What does this PR do ?

This PR separates model specific and run specific inference parameters. Model specific parameters are not stored inside a ModelInferenceParameters dataclass within the model itself. This parameters should be stored in the model yaml, currently this needs to be done manually. In particular, the latest checkpoint: https://huggingface.co/nvidia/magpie_tts_multilingual_357m has been updated to include a inference_parameters field in the top level yaml.
Run specific inference parameters, such as applying local transformer or using CFG, remain as arguments to the inference functions.
Updates inference defaults, and merges the nemo/collections/tts/modules/magpietts_inference/inference.py configuration with this new setup.

**Collection**: tts

# Changelog

- Update MagpieTTS Inference Configurations and enables LT and prior in do_tts()

**PR Type**:

- [x] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -341,6 +341,7 @@ class MagpieTTSModel(ModelPT):
                 codec_model_path, strict=False, override_config_path=codec_model_cfg
             )
         self.sample_rate = codec_model.sample_rate
+        self.output_sample_rate = codec_model.output_sample_rate
         self.codec_model_samples_per_frame = codec_model.samples_per_frame
         # del codec discriminator to free memory
         del codec_model.discriminator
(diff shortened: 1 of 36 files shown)


nemo/collections/tts/models/magpietts.py
```diff
@@ -3544,11 +3579,8 @@ def do_tts(
         with torch.no_grad():
             output = self.infer_batch(
                 batch,
-                max_decoder_steps=max_decoder_steps,
-                temperature=temperature,
-                topk=topk,
                 use_cfg=use_cfg,
-                cfg_scale=cfg_scale,
+                use_local_transformer_for_inference=True,
```
rfejgin, line 3622: Should `use_local_transformer_for_inference` be in `InferenceParameters`?

nemo/collections/tts/models/magpietts.py
```diff
@@ -232,6 +232,45 @@ def __post_init__(self):
...
+        min_generated_frames (int): Setting this greater than 0 prevents rare cases of first-frame termination. Any
+            number greater between 1 and 4 should work, but 4 lines up with the codec's minimum frame requirement.
+    """
+
+    max_decoder_steps: int = 500
+    temperature: float = 0.7
+    topk: int = 80
+    cfg_scale: float = 2.5
+    apply_attention_prior: bool = True
+    prior_epsilon: float = 1e-5
```
rfejgin: These inference prior defaults looks different from the ones I last heard from @paarthneekhara  (about a month ago, so it may have changed?).
`prior_epsilon`, `lookahead_window_size`, `estimate_alignment_from_layers`, `apply_prior_to_layers`, `start_prior_after_n_audio_steps` are all different. Or will we be setting other values via the command line?

paarthneekhara: Yes, the defaults do look very different from the ones we had for the latest checkpoint. I think these should be universally set as follows for all checkpoints we have: `start_prior_after_n_audio_steps=0`, `lookahead_window_size=5`, `ignore_finished_sentence_tracking=True`,

blisc: These were the defaults from infer_batch, but now would be a good time to update them. Let me do that

nemo/collections/tts/models/magpietts.py
blisc: No, the intension of `InferenceParameters` is to save model specific hyperparameters so we can save it as part of the yaml config and subsequently, part of the nemo file. Since some parameters like `estimate_alignment_from_layers` are model-specific, I want to keep this new class to contain only those parameters. Options like use_cfg or use_LT are run-specific and should lie elsewhere.

nemo/collections/tts/models/magpietts.py
blisc: Updated the defaults

nemo/collections/tts/models/magpietts.py
```diff
@@ -232,6 +232,63 @@ def __post_init__(self):
...
+    topk: int = 80
+    cfg_scale: float = 2.5
+    apply_attention_prior: bool = True
+    attention_prior_epsilon: float = 1e-5
+    attention_prior_lookahead_window: int = 5
+    estimate_alignment_from_layers: Optional[List[int]] = None
+    apply_prior_to_layers: Optional[List[int]] = None
+    start_prior_after_n_audio_steps: int = 0
+    use_LT_kv_cache: bool = True
+    ignore_finished_sentence_tracking: bool = True
```
subhankar-ghosh: Should `ignore_finished_sentence_tracking` be fixed into the model? I think we might want to experiment with it during inference runs.
Also, the default value was False I think. During longform `ignore_finished_sentence_tracking=False` works better.

subhankar-ghosh:
Left one minor comment, otherwise LGTM given standard and long-form inference has been tested and nothing breaks.

XuesongYang (approved):

rfejgin (approved):
Looks good to me (some internal discussions ongoing but not needed for this PR)

paarthneekhara:
LGTM.

PR #15272, opened 2026-01-08
[TTS] Add tests of the MagpieTTS inference CLI

rfejgin:
**Changelog:**

Add tests focusing on exercising MagpieTTS inference CLI options. PyTest-based.

* New file: `tests/collections/tts/test_magpietts_inference_cli.py`
This is a PyTest file that we can use to test the inference CLI options. Doing these tests in PyTest (as opposed to shell scripts) should work better because the tests will be easier to write, maintain, parameterize, etc.

* A new manifest was added in the CI storage area on Azure. It is a small version of the `an4` validation set, containing only 4 records. We refer to it from `evalset_config.json` by the name `an4_val_tiny_ci`. This manifest is meant for use in tests that should run quickly and where we do not need metrics resulting from testing many items, e.g. for command line argument tests.

Two simple tests are included in this PR:
- A test that verifies that the `--disable_fcd` option doesn't crash and produces `NaN` for the `frechet_codec_distance` metric
- A test of the same for the `--disable_utmosv2` option and the `utmosv2` metric.

The two tests are combined into one test function by utilizing PyTest's `@parametrize` functionality.

**Collection**: TTS

**PR Type**:

- [ ] New Feature
- [x] Tests 
- [ ] Bugfix
- [ ] Documentation


tests/collections/tts/modules/magpietts_inference/test_magpietts_inference.py
```diff
@@ -0,0 +1,88 @@
...
+    @pytest.mark.run_only_on('GPU')
+    @pytest.mark.parametrize(
+        "disable_flag,num_repeats,metric_key",
+        [
+            # Test both the --disable_fcd and --disable_utmosv2 flags
+            ("--disable_fcd", "2", "frechet_codec_distance"),
+            # Multiple repeats to test that NaNs don't crash the confidence interval calculation
+            ("--disable_utmosv2", "1", "utmosv2_avg"),
+        ],
+        ids=["disable_fcd", "disable_utmosv2"],
```
subhankar-ghosh, line 39: If one fails, would we know which one fails?

tests/collections/tts/test_magpietts_inference_cli.py
```diff
@@ -0,0 +1,88 @@
...
+    HPARAMS_FILE = "/home/TestData/tts/2506_ZeroShot/lrhm_short_yt_prioralways_alignement_0.002_priorscale_0.1.yaml"
+    CHECKPOINT_FILE = "/home/TestData/tts/2506_ZeroShot/dpo-T5TTS--val_loss=0.4513-epoch=3.ckpt"
+    EVALSET_CONFIG = "examples/tts/evalset_config.json"
+
+    @pytest.mark.run_only_on('GPU')
+    @pytest.mark.parametrize(
+        "disable_flag,num_repeats,metric_key",
+        [
+            # Test both the --disable_fcd and --disable_utmosv2 flags
+            ("--disable_fcd", "2", "frechet_codec_distance"),
```
subhankar-ghosh: I think num_repeats can be 1 for this test.

tests/collections/tts/test_magpietts_inference_cli.py
```diff
@@ -0,0 +1,88 @@
...
+import pytest
+
+from examples.tts.magpietts_inference import main as magpietts_inference_main
+
+
+class TestMagpieTTSInferenceCLI:
+    """Tests for MagpieTTS inference command-line interface options."""
+
+    # Test data paths - these should match CI environment
+    CODEC_MODEL_PATH = "/home/TestData/tts/AudioCodec_21Hz_no_eliz_without_wavlm_disc.nemo"
```
subhankar-ghosh: If we add more such tests that need hardcoding the model, hparams and codec filepaths we need to add some kind of config file to read these from one single source. For this test it is fine.

subhankar-ghosh (changes requested):
Overall looks good. Requested one change.

tests/collections/tts/test_magpietts_inference_cli.py
rfejgin: hmm, the reason I had one test with multiple repeats is to make sure the NaN doesn't crash the confidence interval calculation, which only happens when num_repeats > 1. But maybe that's overkill for a tests that runs the whole system. I'll change back to 1.

tests/collections/tts/modules/magpietts_inference/test_magpietts_inference.py
rfejgin: Yes, PyTest reports the error using the test name specified in `ids=[...]`. I just added a comment saying that those are test names.

tests/collections/tts/test_magpietts_inference_cli.py
rfejgin: Fixed - changed to 1.

tests/collections/tts/test_magpietts_inference_cli.py
rfejgin: These particular tests don't really care about the model and the codec, I just had to choose something. If we do need test-specific config later on we can such config into the test (or a set of global configs to choose from).

subhankar-ghosh (approved):
LGTM.

examples/tts/evalset_config.json
```diff
@@ -3,6 +3,11 @@
         "manifest_path": "/home/TestData/an4_dataset/an4_val_context_v1.json",
         "audio_dir": "/",
         "feature_dir": null
+    },
+    "an4_val_tiny_ci": {
+        "manifest_path": "/home/TestData/an4_dataset/an4_val_context_v1_tiny.json",
+        "audio_dir": "/",
+        "feature_dir": null
     }
```
blisc, line 11: You need to update the CI to only run on the an4_val_ci dataset. Right now, since you don't pass in --datasets, it runs on both datasets. It either breaks the CER calculations or times out when running on both datasets.
https://github.com/NVIDIA-NeMo/NeMo/blob/main/tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_SeenSpeakers.sh
https://github.com/NVIDIA-NeMo/NeMo/blob/main/tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_ZeroShot.sh
https://github.com/NVIDIA-NeMo/NeMo/blob/main/tests/functional_tests/L2_TTS_InferEvaluatelongform_Magpietts_ZeroShot.sh
Not sure why longform is finishing correctly but the other two are erroring out

rfejgin: Ah indeed, thank you for that. I was wondering why CI was suddenly timing out...
I've pushed a fix.

blisc (approved):

tests/collections/tts/test_magpietts_inference_cli.py.bak
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
blisc: Can you delete this file?

blisc (changes requested):

tests/collections/tts/test_magpietts_inference_cli.py.bak
rfejgin: Done

PR #15277, opened 2026-01-08
MagpieTTS decoder model on top of NeMo main branch

paarthneekhara:


examples/tts/conf/magpietts/easy_magpietts.yaml
```diff
@@ -0,0 +1,173 @@
+name: Magpie-TTS-DecoderOnly-EN
```
blisc, line 1: Have we tested the non-Lhotse path?

nemo/collections/tts/modules/nemotron_h_decoder.py
```diff
@@ -0,0 +1,1464 @@
+# Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
blisc, line 1: I've been told there's an updated nemotronh model file at https://github.com/NVIDIA-NeMo/Automodel/pull/1091/changes#diff-047ae4149c298197dac7920766cbfabb8a82ef350e6b00dafbfc7361534fd85a
we should move to this.

examples/tts/magpietts_inference.py
```diff
@@ -182,7 +182,11 @@ def run_inference_and_evaluation(
         violin_plot_metrics.remove('utmosv2')
```
blisc, line 214: Let's split this file in 3:

1. This should be renamed to tts_infer.py and contain the common elements: dataset loading
2. Create one helper function for magpietts and one helper function for em-tts with different commandline arguments.

examples/tts/magpietts_streaming_inference.py
```diff
@@ -0,0 +1,1030 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+MagpieTTS Streaming Inference Test Script.
+
+This script tests the streaming TTS inference functionality, supporting both
+single sample (batch_size=1) and batched inference (batch_size>1).
+
+For batched inference, each item in the batch can have different context lengths
+and be in different processing phases (context, prompt, phoneme-only, audio).
```
blisc: Can you add to this as to how this differs from magpietts_inference.py?

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1097,6 +1097,39 @@ def encode_from_g2p(self, g2p_text: List[str], raw_text: Optional[str] = None):
         return [self._token2id[p] for p in ps]
 
 
+class IPABPETokenizer:
```
blisc: Should we subclass Tokenizer instead of instantiation within the class?

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1097,6 +1097,39 @@ def encode_from_g2p(self, g2p_text: List[str], raw_text: Optional[str] = None):
...
+    """Simple IPA BPE tokenizer wrapper around HuggingFace tokenizers.
+
+    Args:
+        tokenizer_path: Path to the tokenizer.json file (or directory containing it).
+    """
+
+    def __init__(self, tokenizer_path: str):
+        import os
+
+        from tokenizers import Tokenizer
```
blisc: Move import statements to the top of the file.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1127,7 +1160,13 @@ def __init__(self, tokenizers: List[Union[BaseTokenizer, PreTrainedTokenizerBase
...
                 tokens.extend(_tokens)
                 num_tokens = len(_tokens)
-                tokenizer_pad_ids[tokenizer_name] = tokenizer.pad_token_id + tokenizer_offset
+                pad_token_id = tokenizer.pad_token_id if tokenizer.pad_token_id is not None else tokenizer.unk_token_id
+                if pad_token_id is None:
+                    raise ValueError(
+                        f"Tokenizer '{tokenizer_name}' has no pad_token_id or unk_token_id. "
+                        "Please set one before using with AggregatedTTSTokenizer."
+                    )
+                tokenizer_pad_ids[tokenizer_name] = pad_token_id + tokenizer_offset
```
blisc, line 1288: Does this affect existing MagpieTTS checkpoints?

scripts/tts_dataset_files/bpe_ipa_tokenizer_2048_en_de_es_fr_hi_it_vi_zh.json
```diff
@@ -0,0 +1,9954 @@
+{
```
blisc: Let's move this file internally. When and we release a checkpoint that uses this, we should bundle this json into the .nemo archive.

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -940,3 +943,81 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
     # for backward compatibility
     g2p_target_new = g2p_target.replace("nemo_text_processing.g2p", "nemo.collections.tts.g2p")
     return g2p_target_new
+
+
+def process_text_for_cer(input_text):
+    """
+    Normalizes text for CER/WER calculation.
+    """
```
blisc: FYI: @rlangman @rfejgin since we were talking about this. Let's lift this from the decoder PR and move it to main early

nemo/collections/tts/modules/__init__.py
```diff
@@ -13,6 +13,12 @@
 # limitations under the License.
 
 import nemo.collections.tts.modules.adapters
+from nemo.collections.tts.modules.nemotron_h_decoder import (
+    HybridMambaAttentionDynamicCache,
+    NemotronHConfig,
+    NemotronHForCausalLM,
+    NemotronHModel,
+)
```
blisc: This isn't necessary. Let's remove

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -244,10 +255,19 @@ def create_dataset(
...
         else:
+            if self.config.is_decoder_only_model:
+                _load_16khz_audio = False
+                _use_text_conditioning_encoder = True
+                _pad_context_text_to_max_duration = False
+            else:
+                _load_16khz_audio = self.model.model_type == 'single_encoder_sv_tts'
+                _use_text_conditioning_encoder = self.model.use_text_conditioning_encoder
+                _pad_context_text_to_max_duration = self.model.pad_context_text_to_max_duration
+
```
blisc: Instead of having if else's throughout the code, let's split this into 2 classes: 1 for encoder-decoder Magpie and 1 for EasyMagpie

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -188,6 +201,12 @@ def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List]]:
             self.eos_id = self.bos_id + 1
             self.pad_id = self.text_tokenizer.pad
 
+        if self.phoneme_tokenizer is None and self.phoneme_tokenizer_config is not None:
+            worker_info = torch.utils.data.get_worker_info()
+            worker_id = worker_info.id if worker_info is not None else 0
+            logging.info(f"Worker {worker_id} initializing phoneme tokenizer...")
+            self.phoneme_tokenizer = instantiate_phoneme_tokenizer(self.phoneme_tokenizer_config)
```
blisc: Can you add a comment explaining why we have to do this?

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -60,6 +60,16 @@ def setup_tokenizers(all_tokenizers_config, mode='train'):
...
 
 
+def instantiate_phoneme_tokenizer(phoneme_tokenizer_config):
+    phoneme_tokenizer = instantiate(phoneme_tokenizer_config)
+    phoneme_vocab_size = len(phoneme_tokenizer.tokens)
+    phoneme_tokenizer.bos_token_id = phoneme_vocab_size
+    phoneme_tokenizer.eos_token_id = phoneme_vocab_size + 1
+    phoneme_tokenizer.unk_token_id = phoneme_vocab_size + 2
+    phoneme_tokenizer.vocab_size = phoneme_vocab_size + 3
+    return phoneme_tokenizer
```
blisc: I'm not sure when you call this function, but this should be part of the tokenizer class not a util function in the dataset.

blisc: However, this only exists in the Lhotse file but not the non-Lhotse file?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,3560 @@
...
+
+    Attributes:
+        name: Unique identifier for this mode (e.g., "full", "streaming_4_8")
+        text_input_mode: Either "full" or "streaming"
+        streaming_phonemes_delay: Delay for phoneme stream (only used in streaming mode)
+        streaming_speech_delay: Delay for speech stream (only used in streaming mode)
+        mode_idx: Index of this mode in the list of modes (used for task embedding lookup)
+    """
+
+    name: str
```
blisc: Should we define the name automatically from the parameters rather than relying on the user to specify it?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,3560 @@
...
+    logging.info(f"Worker {worker_id} initializing...")
+    worker_info = get_worker_info()
+    dataset = worker_info.dataset  # Get the dataset instance in this worker
+    tokenizer = setup_tokenizers(dataset.tokenizer_config, mode=dataset.dataset_type)
+    dataset.text_tokenizer = tokenizer
+    if hasattr(dataset, 'phoneme_tokenizer_config'):
+        dataset.phoneme_tokenizer = instantiate_phoneme_tokenizer(dataset.phoneme_tokenizer_config)
+
+
+class EasyMagpieTTSModel(ModelPT):
```
blisc: This is a really large file. Can we split it up? Some suggestions

- Anything that's common with Encoder-Decoder Magpie, let's move to a separate base class: 
  - The code manipulaiton functions
  - The local transformer functions
  - etc
- Let's move the dataclasses to another file, although we can debate this
- Let's move worker_init_fn too since it should be common to both models
- Could consider splitting training and inference into two classes as well

nemo/collections/tts/models/easy_magpietts_preference_optimization.py
```diff
@@ -0,0 +1,1178 @@
...
+)
+from nemo.utils import logging
+
+try:
+    import torchaudio
+    from torchaudio.pipelines import SQUIM_OBJECTIVE
+
+    HAVE_TORCHAUDIO = True
+except ImportError:
+    HAVE_TORCHAUDIO = False
```
blisc: Let's remove SQUIM, I don't think we use it anymore

nemo/collections/tts/models/easy_magpietts_preference_optimization.py
```diff
@@ -0,0 +1,1178 @@
...
+            cfg_scale = self.cfg.get('inference_cfg_scale', 1.0)
+
+        phoneme_input_type = 'pred'
+        gt_phoneme_input_prob = self.cfg.get('gt_phoneme_input_prob', 0.0)
+        can_use_gt_phonemes = ('phoneme_tokens' in batch_repeated) and ('phoneme_tokens_lens' in batch_repeated)
+        if can_use_gt_phonemes and gt_phoneme_input_prob > 0.0 and mode == 'train':
+            phoneme_input_type = 'gt' if random.random() < gt_phoneme_input_prob else 'pred'
+
+        generation_start_time = time.perf_counter()
+        print("Inference started")
```
blisc: Switch print statments to logging

nemo/collections/tts/models/easy_magpietts_preference_optimization.py
```diff
@@ -0,0 +1,1178 @@
...
+    @torch.no_grad()
+    def _snapshot_trainable_weights(self) -> Dict[int, torch.Tensor]:
+        """Take a snapshot of all trainable parameter values (by param id)."""
+        snapshot = {}
+        for params in self._get_trainable_module_groups().values():
+            for p in params:
+                snapshot[id(p)] = p.data.clone()
+        return snapshot
+
+    def _print_grad_weight_summary(self, metrics: Dict[str, float], step: int) -> None:
```
blisc: This function does not depend on `self`. Consider moving all helper print functions into a separate file and call them within the model instead of defining additional class functions

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
paarthneekhara: Subclassing Tokenizer may not be the best idea because we will have to reassign some internal methods/params to self for it to work correctly. If we call from_file() from the __init__ function, we'll have to reassign normalizer, pre_tokenizer, post_processor and decoder of the loaded model to self. I am moving the imports to the top. It would look something like this if we were to subclass Tokenizer.
```
class IPABPETokenizer(Tokenizer):
    """Simple IPA BPE tokenizer subclassing HuggingFace tokenizers.Tokenizer.

    Args:
        tokenizer_path: Path to the tokenizer.json file (or directory containing it).
    """

    def __init__(self, tokenizer_path: str):
        if os.path.isdir(tokenizer_path):
            tokenizer_file = os.path.join(tokenizer_path, "tokenizer.json")
        else:
            tokenizer_file = tokenizer_path

        if not os.path.exists(tokenizer_file):
            raise ValueError(f"Tokenizer file not found: {tokenizer_file}")

        loaded = Tokenizer.from_file(tokenizer_file)
        super().__init__(loaded.model)
        self.normalizer = loaded.normalizer
        self.pre_tokenizer = loaded.pre_tokenizer
        self.post_processor = loaded.post_processor
        self.decoder = loaded.decoder

        self.tokens = self.get_vocab()
        self.pad = self.tokens.get("<pad>", None)

    def encode(self, text: str) -> List[int]:
        """Encode IPA text to token IDs."""
        return super().encode(text).ids

    def decode(self, tokens: List[int]) -> str:
        """Decode token IDs back to IPA text."""
        return super().decode(tokens)
```

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
paarthneekhara: @shehzeen Can you check this.

nemo/collections/tts/modules/nemotron_h_decoder.py
paarthneekhara: Keeping this version for this merge. Will consider moving to the new file later (once we have an inference path setup for the automodel implementation).

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
shehzeen: MagpieTTS should work the same way. Nemotron tokenizer has pad_token_id as None, so we are using the unk_token_id as the pad_token_id in EasyMagpie. In Magpie tokenizers, tokenizer.pad_token_id is not None (if it's ever None, the tokenizer setup would error out in the old code). So the code functionality should stay the same for MagpieTTS since the tokenizers have the pad_token_id.

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
shehzeen: removed some debugging statements and left a comment.

examples/tts/conf/magpietts/easy_magpietts.yaml
paarthneekhara: Yes

examples/tts/conf/magpietts/easy_magpietts.yaml
```diff
@@ -0,0 +1,187 @@
...
+# Adjust batch size based on GPU memory
+batch_size: 2
+# When doing weighted sampling with multiple manifests, this defines how many training steps are in an epoch.
+# If null, then weighted sampling is disabled.
+weighted_sampling_steps_per_epoch: null
+
+# Dataset metadata for each manifest
+# https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/data/vocoder_dataset.py#L39-L41
+train_ds_meta: ???
+val_ds_meta: ???
```
blisc: This comment no longer points to anything that makes sense, can we update it?

examples/tts/conf/magpietts/easy_magpietts.yaml
```diff
@@ -0,0 +1,187 @@
...
+  
+  # Multi-mode training configuration
+  # The model will randomly select one of the modes for each batch during training.
+  # Each mode has its own task embedding that is prepended to the context.
+  # During inference, you can specify which mode to use via the derived
+  # 'inference_mode' string: "{text_input_mode}_{streaming_phonemes_delay}_{streaming_speech_delay}".
+  training_modes:
+    - text_input_mode: "streaming" # Options: "full", "streaming"
+      streaming_phonemes_delay: 0
+      streaming_speech_delay: 1
```
blisc, line 92: Reading this is confusing, maybe we should point to a comment in the model file instead?

examples/tts/conf/magpietts/easy_magpietts_lhotse.yaml
```diff
@@ -0,0 +1,213 @@
+name: Magpie-TTS-DecoderOnly-EN
+
+quadratic_duration: 20
+
+# Adjust batch size based on GPU memory
+# When doing weighted sampling with multiple manifests, this defines how many training steps are in an epoch.
+# If null, then weighted sampling is disabled.
+
+# Dataset metadata for each manifest
+# https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/tts/data/vocoder_dataset.py#L39-L41
```
blisc: Please remove extra comments

examples/tts/conf/magpietts/easy_magpietts_lhotse.yaml
```diff
@@ -0,0 +1,213 @@
...
+  
+  # Multi-mode training configuration
+  # The model will randomly select one of the modes for each batch during training.
+  # Each mode has its own task embedding that is prepended to the context.
+  # During inference, you can specify which mode to use via the derived
+  # 'inference_mode' string: "{text_input_mode}_{streaming_phonemes_delay}_{streaming_speech_delay}".
+  training_modes:
+    - text_input_mode: "streaming" # Options: "full", "streaming"
+      streaming_phonemes_delay: 0
+      streaming_speech_delay: 1
```
blisc, line 86: Reading this is confusing, maybe we should point to a comment in the model file instead?

examples/tts/magpietts_inference.py
```diff
@@ -185,36 +204,17 @@ def run_inference_and_evaluation(
...
-
-    # Add experiment name prefix if requested
-    if log_exp_name and model_config.checkpoint_file:
-        exp_name = get_experiment_name_from_checkpoint_path(model_config.checkpoint_file)
-        checkpoint_name = f"{exp_name}__{checkpoint_name}"
-
-    # Build full checkpoint identifier (include MoE info if present)
     full_checkpoint_name = (
         f"{checkpoint_name}_{moe_info}{inference_config.build_identifier()}_SV_{eval_config.sv_model}"
     )
```
blisc, line 219: This looks like a bad merge. Can you add back the comments? It looks like full_checkpoint_name now attempts to use variables that are not defined

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -1172,6 +1174,43 @@ def encode_from_g2p(self, g2p_text: List[str], raw_text: Optional[str] = None):
         return [self._token2id[p] for p in ps]
 
 
+class IPABPETokenizer:
```
blisc: @pzelasko @XuesongYang : Do we want Tokenizer classes to subclass nemo/collections/common/tokenizers/tokenizer_spec.py::TokenizerSpec or should we no longer enforce that?

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +808,189 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+    return speaker_embeddings
+
+
+def compute_utmos_scores_from_filepaths(
+    audio_filepaths: Sequence[str],
+    utmos_calculator: Any,
+    batch_size: int = 8,
+    num_workers: int = 0,
+    rank_tag: str = "0",
+) -> List[float]:
```
blisc: @rfejgin Can you take a look at the UTMOS changes (here and in the utmos file)? Do we need this function as I understand we already support batched inference?
I presume this used in post-training.

nemo/collections/tts/models/easy_magpietts_preference_optimization.py
blisc: There are still some print statements that we need to switch

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1560 @@
...
+                    "train/audio_samples_pad_ratio_percent_in_batch": 100
+                    * (1 - audio_samples_total_num / (batch_size * audio_samples_max_len)),
+                }
+            )
+
+        self.log_dict(batch_info_dict, on_step=True)
+
+        return loss
+
+    def validation_step(self, batch, batch_idx):
```
blisc, line 1019: @XuesongYang Can you take a look at the dataset setup and validation logging especially since you recently made changes to these inside of the magpie class

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2108 @@
...
+class TrainingMode:
+    """
+    Configuration for a training mode in multi-mode training.
+
+    Attributes:
+        text_input_mode: Either "full" or "streaming"
+        streaming_phonemes_delay: Delay for phoneme stream (only used in streaming mode)
+        streaming_speech_delay: Delay for speech stream (only used in streaming mode)
+        mode_idx: Index of this mode in the list of modes (used for task embedding lookup)
+    """
```
blisc, line 59: This needs more documentation on what the modes mean

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2108 @@
...
+    mode_idx: int
+
+    @property
+    def name(self) -> str:
+        """Derived identifier used for inference selection and logging."""
+        return f"{self.text_input_mode}_{self.streaming_phonemes_delay}_{self.streaming_speech_delay}"
+
+
+@dataclass
+class StreamingState:
```
blisc: Let's go over this streaming code later.

nemo/collections/tts/models/easy_magpietts_preference_optimization.py
```diff
@@ -0,0 +1,1291 @@
...
+
+    def state_dict(self, destination=None, prefix='', keep_vars=False):
+        state_dict = super().state_dict(destination=destination, prefix=prefix, keep_vars=keep_vars)
+        keys_substrings_to_exclude = ['_reference_model', '_utmos_calculator']
+        for key in list(state_dict.keys()):
+            if any(substring in key for substring in keys_substrings_to_exclude):
+                del state_dict[key]
+        return state_dict
+
+    def _get_cached_normalizer(self, lang_key: Optional[str]):
```
blisc, line 274: @shehzeen : Can you do a pass over all the functions and add a docstring to them?

examples/tts/magpietts_inference.py
paarthneekhara: I brought back comments, but full_checkpoint_name assignment looks correct. The variables being used are being passed into the funciton.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
pzelasko: If it's not too much hassle, let's subclass and implement its API.

nemo/collections/tts/parts/utils/helpers.py
shehzeen: This calls into the batched implementation of UTMOS by @rfejgin . Since that function expects a directory (instead of a list of filepaths), this function creates a `tempfile.TemporaryDirectory` from the filepaths, before calling the batched utmos function. Lines 981 to 994 can be simplified since they are overly cautious right now.

shehzeen: I have simplified it now.

rfejgin: Sorry folks, missed the earlier notification on this discussion.

@shehzeen : there may be a simpler way to do this. 
The current `utmosv2.py` [wrapper](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/tts/modules/utmosv2.py) I wrote indeed [only takes a directory name](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/tts/modules/utmosv2.py#L65) and processes all files in it; but doesn't allow an explicit list of files.

That wrapper calls utmosv2 library's`predict()` API. But that `predict()` API does *also* [support sending in an explicit list of files](https://github.com/sarulab-speech/UTMOSv2/blob/e53a6762948b908105d48d6cfd453f1b58156ed0/utmosv2/_core/model/_common.py#L110) via its `val_list` argument. I haven't tested that mode, and was worried about some assert triggering from peeking at the code a few months ago, but it's worth a try.

Essentially we could extend the wrapper's API to optionally accept a list of files. Then it wouldn't require the additional compute_utmos_scores_from_filepaths() and temporary directory etc.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+        self.local_transformer_loss_scale = cfg.get('local_transformer_loss_scale', 1.0)
+
+        self.cross_entropy_loss = nn.CrossEntropyLoss(reduction='none')
+
+        # Validation inference with metrics (optional)
+        self.run_val_inference = cfg.get('run_val_inference', False)
+        self.use_multilingual_asr = cfg.get('use_multilingual_asr', False)
+        if self.run_val_inference:
+            logging.info("Loading eval models for validation inference (ASR and speaker verification)...")
+            if self.use_multilingual_asr:
```
subhankar-ghosh, line 130: Why do we have these imports in the constructor? These imports might cause issues during model loading while doing inference. IMO we should not include evaluation metrics in the class. Keep it separate like we do in MagpieTTS.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+        return super()._get_state_dict_keys_to_exclude() + [
+            '_speaker_verification_model',
+            '_eval_asr_model',
+            '_eval_speaker_verification_model',
+            'whisper_model',
+            'whisper_processor',
+            '_utmos_calculator',
+        ]
+
+    def compute_loss(self, logits, audio_codes, audio_codes_lens):
```
subhankar-ghosh, line 170: Is it possible to reuse the compute_loss method from MagpieTTS? I think this is calculating the same loss if we pass frame_stacking_factor=1.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+            phoneme_loss = phoneme_loss * loss_mask
+            phoneme_loss = phoneme_loss.sum() / loss_mask.sum()
+            if total_phoneme_loss is None:
+                total_phoneme_loss = phoneme_loss
+            else:
+                total_phoneme_loss = total_phoneme_loss + phoneme_loss
+        total_phoneme_loss = total_phoneme_loss / self.phoneme_stacking_factor
+        return total_phoneme_loss, loss_mask
+
+    def log_val_audio_example(
```
subhankar-ghosh, line 219: This change can be part of a later PR. Can we move log_audio_* and log_plot_* to helper files and reuse them across models? I think we can reuse a lot of these plot and log methods from MagpieTTS.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+            lengths=[delay, audio_codes_lens_target],
+        )
+
+        return audio_channel_embedding, audio_channel_lens, audio_codes_target, audio_codes_lens_target
+
+    def slice_pred_embeddings(self, transformer_out, context_lens, target_lens):
+        """
+        Slices the transformer output to get the predicted embeddings for the target sequence.
+        Args:
+            transformer_out: (B, T, E)
```
subhankar-ghosh: rename transformer_out here, it can be confusing because this is actually transformer_out.last_hidden_state and transformer_out is used to represent the whole object returned by the forward function.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+                    )
+                    ctx_path = os.path.join(audio_dir, f'rank{self.global_rank}_batch{batch_idx}_idx{idx}_context.wav')
+                    sf.write(ctx_path, ctx_audio_np, self.output_sample_rate)
+                    context_audio_paths.append(ctx_path)
+
+            # Compute metrics if we have audio paths
+            if predicted_audio_paths and context_audio_paths:
+                with torch.no_grad():
+                    # ASR transcription for CER/WER
+                    if self.use_multilingual_asr:
```
subhankar-ghosh, line 1171: do we really need cer/wer and ssim during validation step? It might make sense for experimentation but for open sourcing it does not make sense especially after we have a working recipe. Makes val step heavier.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1554 @@
...
+
+        if self.run_val_inference:
+            # Collect metrics only from outputs that have them
+            def collect_if_exists(key):
+                values = [x[key] for x in self.validation_step_outputs if key in x]
+                if values:
+                    return torch.stack(values).mean()
+                return None
+
+            val_cer = collect_if_exists("val_cer")
```
subhankar-ghosh: use a list and iterate
```
val_metrics = ["val_cer", "val_wer", "val_ssim", "val_utmos"]
for val_metric in val_metrics:
    metric_value = collect_if_exists(val_metric)
    if metric_value is not None:
        self.log(val_metric, metric_value, prog_bar=True, sync_dist=True)
```

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+    This dataclass maintains the dynamically changing state for autoregressive
+    streaming generation. Static configuration lives in StreamingConfig (accessed
+    via the `config` field).
+
+    The streaming operates in four phases (per batch item):
+    1. Context phase (context_position < full_context_lens): Processing remaining context
+    2. Prompt phase (text_tokens_seen < phoneme_delay): Only text, no predictions
+    3. Phoneme-only phase (phoneme_delay <= text_tokens_seen < speech_delay): Phoneme predictions only
+    4. Audio phase (text_tokens_seen >= speech_delay): Both phoneme and audio predictions
+    """
```
subhankar-ghosh, line 118: needs docstring for each of the parameters in here.

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+    last_phoneme_tokens: Optional[torch.Tensor]
+    last_audio_codes: Optional[torch.Tensor]
+    audio_prediction_start_idx: torch.Tensor
+    audio_prediction_end_idx: torch.Tensor
+    phoneme_prediction_start_idx: torch.Tensor
+    phoneme_prediction_end_idx: torch.Tensor
+    gt_phoneme_embeddings: Optional[torch.Tensor] = None  # (B, T', E) pre-computed GT embeddings
+    gt_phoneme_lens: Optional[torch.Tensor] = None  # (B,) lengths after stacking
+    gt_audio_embeddings: Optional[torch.Tensor] = None  # (B, T', E) pre-computed GT audio embeddings
+    gt_audio_lens: Optional[torch.Tensor] = None  # (B,) lengths after stacking
```
subhankar-ghosh, line 148: Is there a possibility of reducing the number of fields in this dataclass? Looks like there are too many.

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+        # In the case where there are no context codes saved, and there is no context audio (in the text context path),
+        # We create a dummy context code tensor that is only [context_BOS, context_EOS] that is repeated for
+        # data_num_audio_codebooks
+        self.data_num_audio_codebooks = data_num_audio_codebooks
+        self.num_audio_codebooks = num_audio_codebooks
+        self.codebook_size = codebook_size
+
+        self.codec_model_samples_per_frame = codec_model.samples_per_frame
+        # Our codebooks start with actual audio codec tokens, followed by special tokens.
+        # The `forced_*` options are for backward compatibility for models trained with older code.
```
subhankar-ghosh, line 252: do we used forced_* anywhere in EasyMagpie? if not then remove this.

examples/tts/magpietts_inference.py
```diff
@@ -134,50 +154,49 @@ def create_formatted_metrics_mean_ci(metrics_mean_ci: dict) -> dict:
...
-        # Dataset filtering not specified, return all datasets
         return list(dataset_meta_info.keys())
     else:
         datasets = datasets.split(",")
-        # Check if datasets are valid
         for dataset in datasets:
             if dataset not in dataset_meta_info:
                 raise ValueError(f"Dataset {dataset} not found in dataset meta info")
-        # Return all requsted datasets
         return datasets
```
blisc, line 166: Let's add these back

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+        """Unstack frame-stacked codes and pad short sequences before decoding."""
+        if self.frame_stacking_factor > 1 and codes.size(1) == self.num_audio_codebooks * self.frame_stacking_factor:
+            codes, codes_len = self.unstack_codes(codes, codes_len, self.frame_stacking_factor)
+        if min_len > 0 and codes_len.min() < min_len:
+            codes = torch.nn.functional.pad(input=codes, pad=(0, min_len - codes_len.min()), value=0)
+            codes_len = torch.where(codes_len < min_len, torch.ones_like(codes_len) * min_len, codes_len)
+            codes = codes[:, :, : codes_len.max()]
+        return codes, codes_len
+
+    def embed_audio_tokens(self, audio_tokens):
```
subhankar-ghosh, line 573: Can we reuse from magpietts or extract this as a common method so both the model classes can use it?

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+        else:
+            backend_out = self.decoder(
+                inputs_embeds=inputs_embeds,
+                attention_mask=attention_mask,
+                use_cache=use_cache,
+                past_key_values=past_key_values,
+            )
+        return backend_out
+
+    def logits_to_audio_codes(self, all_code_logits, audio_codes_lens):
```
subhankar-ghosh, line 619: same. Can we reuse magpietts?

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -0,0 +1,2098 @@
...
+            codebook_preds = torch.argmax(codebook_probs, dim=-1)  # (B, T')
+            all_preds.append(codebook_preds)
+
+        all_preds = torch.stack(all_preds, dim=1)  # (B, C, T')
+        audio_mask = get_mask_from_lengths(audio_codes_lens)
+        all_preds = all_preds * audio_mask.unsqueeze(1)
+
+        return all_preds
+
+    def sample_codes_from_logits(
```
subhankar-ghosh, line 638: I think we also do something similar in Magpietts, if yes then can we reuse?

nemo/collections/tts/modules/utmosv2.py
```diff
@@ -62,22 +62,39 @@ def __call__(self, file_path):
...
+            filenames: If provided, only score these basenames (e.g. ``["000000.wav", "000001.wav"]``)
+                via the library's ``val_list`` parameter instead of globbing the whole directory.
+                If None, all ``*.wav`` files in ``input_dir`` are scored.
         Returns:
             A list of dictionaries, each containing the file path and the UTMOSv2 score.
         """
+        if num_workers is None:
+            num_workers = batch_size
+
+        predict_kwargs = dict(input_dir=input_dir, num_repetitions=1, num_workers=num_workers, batch_size=batch_size)
```
blisc: Can we remove the use of a kwargs dict?

nemo/collections/tts/modules/utmosv2.py
```diff
@@ -62,22 +62,39 @@ def __call__(self, file_path):
...
+        if filenames is not None:
+            predict_kwargs["val_list"] = list(filenames)
```
blisc: Let's use the same parameters as predict so, use val_list not filenames. And add a check if this is already a list.

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -12,75 +12,84 @@
...
     maskgit_sampling_type: Optional[str] = None
 
     def build_identifier(self) -> str:
-        """Build a unique identifier string for this configuration.
-
-        Used for naming output directories and files.
-
-        Returns:
-            String identifier incorporating key config values.
-        """
```
blisc, line 83: Can we add back all the deleted comments in this file?

nemo/collections/tts/models/easy_magpietts.py
paarthneekhara: I am not wanting to reuse it because it uses several self variables. @rlangman suggested to dedupe code only if the function is reusable across other models as well. I dont think it is very reusable and it's just additive cross-entropy loss. Also the function is fairly small (~20 lines) so I am not too inclined towards a shared implementation.

nemo/collections/tts/models/easy_magpietts.py
paarthneekhara: That makes sense to me. I'll make another PR once this is merged to unify the audio/image logging.

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -12,62 +12,74 @@
...
-    # Core sampling parameters
     batch_size: int = 32
     use_cfg: bool = False
-    apply_attention_prior: bool = False
+    use_local_transformer: bool = False
+
+    @abc.abstractmethod
+    def build_identifier(self) -> str:
+        """Build a unique identifier string for naming output directories."""
+        ...
```
blisc:
```suggestion
        pass
```
Switch your `...` into pass statements

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -111,148 +123,154 @@ def build_identifier(self) -> str:
...
+            model: Loaded TTS model (should be on GPU and in eval mode).
             config: Inference configuration.
         """
         self.model = model
         self.config = config
-
-        # Set phoneme probability to 1 for inference
         self._configure_tokenizer()
-
-        # Cached state from create_dataset (set when create_dataset is called)
```
blisc, line 146: During your refactor, you deleted most of the comments from the Encoder-Decoder path, can you add them back?

nemo/collections/tts/models/easy_magpietts.py
paarthneekhara: I find it quite helpful to track CER/SSIM during training to know if the model is working and also comparing different experiments. Val loss comparison often doesnt translate to which experiment is doing well. The overhead of this step is not much since it is spread across 4 nodes, so less than 10 iterations of the infer_batch cover the entire val set.

nemo/collections/tts/models/easy_magpietts_inference.py
paarthneekhara: Done.

nemo/collections/tts/models/easy_magpietts.py
paarthneekhara: Moved them above.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -0,0 +1,1546 @@
...
+    phoneme_tokens_target: Optional[torch.Tensor]
+    phoneme_tokens_lens_target: Optional[torch.Tensor]
+    audio_codes_target: torch.Tensor
+    audio_codes_lens_target: torch.Tensor
+    context_audio_codes: torch.Tensor
+    context_audio_codes_lens: torch.Tensor
+    selected_training_mode: Optional[str]
+
+
+class EasyMagpieTTSModel(EasyMagpieTTSInferenceModel):
```
XuesongYang, line 101: any reason why subclassing inference class for pretraining? aren't we expecting inference subclassing pretraining class?

examples/tts/conf/magpietts/easy_magpietts_lhotse.yaml
```diff
@@ -0,0 +1,206 @@
...
+    volume_norm: true
+
+    dataset:
+      min_duration: 0.2
+      min_context_speaker_similarity: 0.6
+      max_cer: 0.03
+      batch_duration: ???   # recommend to use smaller batch_duration for validation dataset than training dataset.
+      quadratic_duration: ${quadratic_duration}
+      use_bucketing: false
+      force_finite: true
```
XuesongYang, line 146: pls add the two latest changes: 
1. https://github.com/NVIDIA-NeMo/NeMo/pull/15387
2. https://github.com/NVIDIA-NeMo/NeMo/pull/15427

nemo/collections/tts/models/easy_magpietts.py
XuesongYang: I checked the model class and the Lhotse YAML config, and I can confirm that this model doesn’t currently support multiple validation dataloaders. 

Since this PR is already quite large and aggregates a significant number of changes, let’s avoid adding multiple dataloader support here. Instead, we can plan to implement that in a separate PR (similar to https://github.com/NVIDIA-NeMo/NeMo/pull/15348).

examples/tts/conf/magpietts/easy_magpietts_lhotse.yaml
blisc: @paarthneekhara Let's make these changes.

paarthneekhara: Thanks for pointing these. I have made these changes.

PR #15311, opened 2026-01-23
Fix alignment logits when using fusion models in TDT greedy decoder

Garvys:
# What does this PR do ?

Fixes an inconsistency in alignment logits when using preserve_alignments=True with TDT models and fusion models (e.g., LM, boosting tree).

## Problem
When fusion models are enabled, the label selection logic correctly applies:
- If blank is the most likely token before fusion → emit blank
- Otherwise → emit the non-blank token with the highest logit after fusion

However, the logits stored in alignments were always the pre-fusion logits. This caused a mismatch where the stored label (correctly chosen with the above logic) did not correspond to the argmax of the stored logits.

## Solution
Update the logits sent to alignments to be consistent with the label selection:

- If blank was selected (best before fusion) → store pre-fusion logits (unchanged)
- If a non-blank token was selected → store post-fusion logits with blank set to -inf

This ensures `argmax(logits) == label` for every alignment entry, making the alignments reliable for downstream analysis.

## Scope
Implementation in `torch_impl` method only. CUDA graphs implementation is not affected.

Collection: ASR

# Changelog

- `tdt_label_looping.py`: Updated `torch_impl` to store alignment-consistent logits when fusion models are used (both outer loop and inner loop)
- `test_rnnt_decoding.py`: Added `test_tdt_greedy_decoding_preserve_alignments_with_fusion` to verify label/logits consistency with fusion models

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

None

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,0 @@
-.github/ @pablo-garay @ko3n1g @thomasdhc @chtruong814
-docker/Dockerfile.ci @pablo-garay @ko3n1g @thomasdhc @chtruong814
-.pylintrc.* @pablo-garay @ko3n1g @thomasdhc @chtruong814
-.flake8.* @pablo-garay @ko3n1g @thomasdhc @chtruong814
--- a/.github/workflows/cicd-main.yml
+++ b/.github/workflows/cicd-main.yml
@@ -46,6 +46,8 @@ jobs:
       TESTS_TO_RUN: ${{ inputs.test_to_run }}
       EVENT_NAME: ${{ github.event_name }}
       HAS_LABEL: ${{ github.event.label.name == 'Run CICD' }}
+      HAS_FORCE_SPEECH_CI_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'force-run-speech-ci') }}
+      HAS_FORCE_NEMO2_CI_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'force-run-nemo2-ci') }}
     steps:
       - name: Checkout branch
         uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


Garvys:
If you have time for a review @nithinraok. Thanks a lot!!

artbataev (changes requested):
First, thanks for this contribution.
Unfortunately, I'm working on a concurrent PR with modifications to the decoding algorithm, so asking to wait before it is merged.
But I would be happy to fix bugs and inconsistencies in this implementation.

Also, it would be great if you could clarify your use-case of `alignments`: `alignments` data structure (which stores all logits) requires an enormous amount of memory and makes decoding slower. 

Regarding the PR itself:
- I agree about using consistent logits with LM scores in alignments (currently inconsistent, but for both TDT and RNN-T)
- Also, we need to maintain consistency of `tdt_label_looping` and `rnnt_label_looping`
- I disagree with using `-inf` for blank, since it will break some of our use cases of alignments.

If you clarify the core use-case, we could discuss how to implement it.

artbataev:
Concurrent PR with decoding changes: https://github.com/NVIDIA-NeMo/NeMo/pull/15315

Garvys:
Hello @artbataev !

Thanks a lot for taking the time to review my PR!

No worries, I can wait. 

Regarding my use-case for the `alignments` data structure: I use it to have access to the top tokens at each decoding step in order to troubleshoot why the model is not behaving as expected. I attached a screenshot of the tool.

Screenshot 2026-01-21 at 16 20 05

It is very interesting to see, at each prediction step, which tokens are the most likely according to the model and to compare that with how the Fusion model is trained. This really allow to have a deep understanding of the behaviour of the models and to see which changes could improve the results (fine tuning, changing how the fusion model is trained ...)

What i noticed is that the token that was predicted at each prediction step was not necessarily the one with the highest score in the logits of the relevant prediction step. My proposed change allows to fix that but indeed I agree that setting -inf for the blank logit is not very satisfying. 

When the blank is not the most likely, keeping the blank logit to its value and updating the logits with the fusion scores of all the other tokens won't guarantee that the blank token is not the most likely token. 
It does seem to me that the good way of implementing this would be to store the blank logits in another tensor when shallow fusion is applied.

WDYT ? How would you fix this issue with another implementation ?

Also, thanks a lot for linking the other PR. I am having a look ! 

Thanks a lot for all the PRs you merged related to TDT, very impressive work 💪

artbataev:
@Garvys, are the changes still relevant?

If so, please look at the new decoding code.

Note that there is no longer a separate alignments data structure; everything is stored in `BatchedHyps`
Currently, if we require alignments at each step (see `preserve_logits` and `record_all_steps` flags; >2x slower full model, decoding itself is many times slower and involves huge memory usage), unbiased logits are stored.
We can also store logits with biasing (as introduced in this PR), but for greedy decoding, such logits are not normalized and are summed with the biasing model output (no log-softmax; this makes decoding faster but only works with greedy decoding, since the normalization coefficient is the same for all outputs for the current step only).

This looks fine if the logic is explicitly toggled with a flag, and is also consistent across implementations (at least label-looping TDT and RNN-T, both torch/cuda graphs implementations).

PR #15348, opened 2026-02-02
[magpietts] added multiple validation dataloaders and log metrics per val data

XuesongYang:
## Summary
- Cherry-picked [#15189](https://github.com/NVIDIA-NeMo/NeMo/pull/15189) (multi-validation dataloaders) to `main`.
- Refactored media artifact logging.
- Added MoE expert usage monitoring (per-expert scalars + layer-wise heatmaps).
- Unified `validation_step_outputs` to always use `list[list]` structure, eliminating conditional branching for single vs. multi-dataloader paths.
- Unified config key from `dataset` (singular) to `datasets` across non-lhotse YAML configs.
- Fixed PO subclasses (`OfflinePO`, `OnlinePO`) to work with the new `list[list]` validation outputs.

### Multi-Validation Dataloaders

Added support for running validation on multiple datasets simultaneously. Each dataset gets its own dataloader with per-dataset metric logging (`Loss:<dataset_name>/<metric>`), plus an averaged `Loss:val_avg/<metric>`.

**Training command:**

```bash
python examples/tts/magpietts.py \
    ...
    model.train_ds.input_cfg="/data/manifests/train_input_cfg.yaml" \
    model.validation_ds.datasets="/data/val_datasets.yaml" \
    ...
```

**Validation datasets YAML** (generalizes to multiple languages/splits):

```yaml
- name: "LibriTTS_dev_clean"
  input_cfg: "/data/manifests/val_input_cfg_en.yaml"
- name: "LibriTTS_test_clean"
  input_cfg: "/data/manifests/val_input_cfg_en_testClean.yaml"
```

Shared settings (e.g., `batch_duration`, `volume_norm`, `min_duration`) live at the `validation_ds` level; per-dataset entries inherit and can override them.

### Config Changes

- **`train_ds`**: Fields previously nested under `train_ds.dataset` are now directly under `train_ds` for lhotse configs; under `train_ds.datasets` for non-lhotse configs.
- **`validation_ds`**: Removed the `dataset` nesting level. Now requires a `datasets` key, a list for lhotse (one dataloader per entry) or a dict for non-lhotse (single dataloader, multiplicity via `dataset_meta`).

### MoE Expert Usage Monitoring

- Per-expert usage scalars under `MoE:train/` and `MoE:<dataset>/` panels.
- Layer-wise expert usage heatmaps (deviation from ideal usage `1/num_experts`) at validation intervals.
- Aggregate `moe_expert_usage_variance` for training as a load-balance health indicator.
- MoE auxiliary loss logging skipped when coefficient is 0 to avoid constant-zero noise.

### Example WandB Plots

Screenshot 2026-03-04 at 11 14 08 AM
Screenshot 2026-03-04 at 11 14 16 AM
Screenshot 2026-02-02 at 4 26 33 PM
Screenshot 2026-02-02 at 4 27 07 PM
Screenshot 2026-02-02 at 4 27 30 PM
Screenshot 2026-02-02 at 4 27 52 PM

Full WandB run: https://wandb.ai/aiapps/debug_magpieTTS_EN_2509_multiValSet/runs/5za0abz7

### DIFF
--- a/docs/source/tts/magpietts-longform.rst
+++ b/docs/source/tts/magpietts-longform.rst
@@ -68,7 +68,7 @@ The input text is split into individual sentences using punctuation markers (``.
 Step 2: State Initialization
 ----------------------------
 
-A ``LongformChunkState`` object is created to track information across sentence chunks:
+A ``ChunkState`` object is created to track information across sentence chunks:
 
 - **History text tokens**: Text from previous chunks for context
 - **History encoder context**: Encoder outputs that provide continuity
@@ -112,7 +112,7 @@ Key Components
 
 1. **Sentence Splitting** (``split_by_sentence``): Intelligently splits text on sentence boundaries while handling abbreviations (e.g., "Dr.", "Mr.").
 
-2. **Chunk State** (``LongformChunkState``): Maintains context across chunks:
+2. **Chunk State** (``ChunkState``): Maintains context across chunks:
 
    - ``history_text``: Text tokens from previous chunks
    - ``history_context_tensor``: Encoder outputs for continuity
@@ -211,24 +211,24 @@ Configuration Dataclasses
 #########################
 
 
-``LongformConfig``
-------------------
+``ChunkedInferenceConfig``
+--------------------------
 
 Immutable tuning parameters (set in model):
 
 .. literalinclude:: ../../../nemo/collections/tts/models/magpietts.py
    :language: python
-   :pyobject: LongformConfig
+   :pyobject: ChunkedInferenceConfig
 
 
-``LongformChunkState``
-----------------------
+``ChunkState``
+--------------
 
 Mutable state passed between chunk iterations:
 
 .. literalinclude:: ../../../nemo/collections/tts/models/magpietts.py
    :language: python
-   :pyobject: LongformChunkState
+   :pyobject: ChunkState
 
 
 Best Practices


nemo/collections/tts/models/magpietts.py
```diff
@@ -3321,32 +3440,204 @@ def test_step(self, batch, batch_idx):
...
+            self.log(
+                "val_loss",
+                avg_val_loss,
+                prog_bar=False,
+                sync_dist=True,
+                on_step=False,
+                on_epoch=True,
+                logger=False,  # Don't log to wandb (already logged as "Loss:val_avg/loss" above)
+                enable_graph=False,
+            )
```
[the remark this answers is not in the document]
XuesongYang: good suggestion, but for now we would consider averaged val loss as a criteria to select model. maybe in future, we can make it more flexible.

nemo/collections/tts/models/magpietts.py
```diff
@@ -3373,6 +3664,115 @@ def get_dataset(self, dataset_cfg, dataset_type):
...
+
+        # Check if datasets is a path to an external YAML file
+        if 'datasets' in val_data_config and isinstance(val_data_config.datasets, (str, Path)):
+            # Load datasets from external YAML file (supports local paths and remote URLs like s3://)
+            datasets_file_path = val_data_config.datasets
+            logging.info(f"Loading validation datasets from external file: {datasets_file_path}")
+            datasets_list = OmegaConf.create(load_yaml(datasets_file_path))
+            # Replace the string path with the loaded list in config
+            with open_dict(val_data_config):
+                val_data_config.datasets = datasets_list
```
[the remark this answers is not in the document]
XuesongYang: not exactly. `self._update_dataset_config` already takes care of convertion to DictConfig from Dict.

nemo/collections/tts/models/magpietts.py
```diff
@@ -3321,32 +3440,204 @@ def test_step(self, batch, batch_idx):
...
-            val_local_transformer_loss = collect("val_local_transformer_loss")
-            self.log("val/local_transformer_loss", val_local_transformer_loss, prog_bar=True, sync_dist=True)
-        self.validation_step_outputs.clear()  # free memory
+        """
+        Computes and logs averaged metrics across all validation dataloaders.
+
+        This method:
+        1. Computes per-dataloader metrics from validation outputs.
+        2. Computes averages across all dataloaders.
+        3. Logs both per-dataloader and averaged metrics (e.g., "val_set_0/loss", "val/loss").
```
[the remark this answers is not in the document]
XuesongYang: fixed in the new commit.

nemo/collections/tts/models/magpietts.py
```diff
@@ -1711,127 +1713,217 @@ def sample_codes_from_logits(
...
+        This method creates wandb.Audio/wandb.Image objects and logs them.
+        It batches all WandB logs into a single call for efficiency.
+
+        Args:
+            media_data: Dict containing:
+                - 'dataset_prefix': str, prefix for log keys (e.g., 'val', 'val_set_0')
+                - 'audio_data': Dict with 'pred_audios', 'target_audios', 'context_audios' lists
+                - 'attention_data': Dict mapping attention names to lists of numpy images
+            global_step: Current training step for logging.
+        """
```
blisc: Please refactor this function to not use a dictionary as the input. You are hiding arguments inside a dictionary without a good reason to do so.

- dataset_prefix should just be a string argument as that's all you use it for
- audio_data should probably be split into three arguments that are lists
- attention_data might be best kept as a dictionary; please do not make it nested. should just be key: data

nemo/collections/tts/models/magpietts.py
```diff
@@ -3373,6 +3667,108 @@ def get_dataset(self, dataset_cfg, dataset_type):
         )  # This will be used in worker_init_fn for instantiating tokenizer
         return dataset
 
+    def setup_multiple_validation_data(self, val_data_config: Union[DictConfig, Dict]):
```
blisc, line 3927: Can you add a test that tests that function does as you want with the new style of configurations as well as the old style of configurations?

nemo/collections/tts/models/magpietts.py
```diff
@@ -3373,6 +3667,108 @@ def get_dataset(self, dataset_cfg, dataset_type):
...
+                dataloaders.append(dataloader)
+                logging.info(f"  - Validation dataset {idx}: '{dataset_name}'")
+
+            self._validation_dl = dataloaders
+            self._validation_names = dataset_names
+            logging.info(f"Successfully setup {len(dataloaders)} validation dataloaders")
+
+        else:
+            # Single validation dataset - use the standard setup
+            logging.info("Setting up single validation dataset")
```
blisc: Can you add a warning that this style is deprecated and to move to the new style or do you plan on supporting both for the future?

nemo/collections/tts/models/magpietts.py
XuesongYang: good point! i would wrap up individual arguments with a dataclass for better logically grouped.

nemo/collections/tts/models/magpietts.py
XuesongYang: my plan is to backward support legacy validation config format for a while, and then deprecate it in future major release of NeMo. 

Let me add a deprecation warning here for clarity.

nemo/collections/tts/models/magpietts.py
XuesongYang: adding unit tests now for legacy and new configs. in the future, we would deprecate legacy ones.

nemo/collections/tts/models/magpietts.py
```diff
@@ -3321,32 +3440,207 @@ def test_step(self, batch, batch_idx):
...
+        if val_alignment_loss is not None:
+            log_dict['alignment_loss'] = val_alignment_loss
+        if val_aligner_encoder_loss is not None:
+            log_dict['aligner_encoder_loss'] = val_aligner_encoder_loss
+        if val_local_transformer_loss is not None:
+            log_dict['local_transformer_loss'] = val_local_transformer_loss
+
+        return log_dict
+
     def on_validation_epoch_end(self):
```
subhankar-ghosh, line 3766: This method is too long. I would suggest creating two separate methods like:
```
if is_multi_dataloader:
    self._process_multi_dataloader_validation()
else:
    self._process_single_dataloader_validation()
```

Instead of maintaining 4 lists 
all_losses, all_codebook_losses, all_alignment_losses, all_aligner_encoder_losses, all_local_transformer_losses switch to a dict. Keeps it clean, something like:

```
for metric_name, metric_value in dataloader_logs.items():
                self.log(f"Loss:{dataloader_prefix}/{metric_name}", metric_value, prog_bar=False, sync_dist=True)
                aggregated_metrics.setdefault(metric_name, []).append(metric_value)
```

nemo/collections/tts/models/magpietts.py
```diff
@@ -3321,32 +3456,207 @@ def test_step(self, batch, batch_idx):
...
+            return torch.stack(values).mean()
+
+        if len(outputs) == 0:
+            raise ValueError(
+                f"No validation outputs for dataloader {dataloader_idx}. "
+                f"This indicates an issue with the validation dataloader or validation step."
+            )
+
+        # Compute required metrics
+        val_loss = collect_required_metric(outputs, 'val_loss')
```
subhankar-ghosh, line 3717: Just a suggestion. I see a the metrics being hard coded everywhere and loosely used. Instead we should define them in a constant Enum.

nemo/collections/tts/models/magpietts.py
```diff
@@ -3321,32 +3456,207 @@ def test_step(self, batch, batch_idx):
...
+        val_codebook_loss = collect_required_metric(outputs, 'val_codebook_loss')
+
+        # Compute optional metrics
+        val_alignment_loss = collect_optional_metric(outputs, 'val_alignment_loss')
```
subhankar-ghosh: Can something like this be done?
```
for metric in VAL_METRIC_LIST:
   log_dict[metric] = collect_optional_metric(outputs, metric)
```
We need to check for Nones.

subhankar-ghosh (changes requested):

nemo/collections/tts/models/magpietts.py
XuesongYang: I may not follow what are the "hard coded" metrics in this context. Could you pls elaborate?

nemo/collections/tts/models/magpietts.py
XuesongYang: Nones are already managed in `validation_step`. No need to check again here. Besides, `collect_optional_metric` would return None if any `metric` does not exist in `outputs`.

XuesongYang: > Can something like this be done?
> 
> ```
> for metric in VAL_METRIC_LIST:
>    log_dict[metric] = collect_optional_metric(outputs, metric)
> ```
> 
> We need to check for Nones.

we only have two required losses, so i believe no need to wrap up as a for loop. Similar to optional losses, we only have 3. So I would leave it if we want to access many more losses in the future.

nemo/collections/tts/models/magpietts.py
XuesongYang: 1. Using a dict instead of 4 separate lists: This is a good suggestion! will make changes.

2. Splitting into two methods: Reasonable for readability, though the method is currently ~115 lines which is manageable. The split would reduce cognitive load. After refactoring by item 1, 115 lines are reduced to 50 lines.

nemo/collections/tts/models/magpietts.py
XuesongYang: a second thought on your suggestion, maybe we can get something like below,
```python
from enum import Enum

class ValidationMetric(Enum):
    """Validation metric keys used in validation_step outputs and logging."""
    LOSS = 'loss'
    CODEBOOK_LOSS = 'codebook_loss'
    ALIGNMENT_LOSS = 'alignment_loss'
    ALIGNER_ENCODER_LOSS = 'aligner_encoder_loss'
    LOCAL_TRANSFORMER_LOSS = 'local_transformer_loss'

# Usage:
val_loss = collect_required_metric(outputs, f'val_{ValidationMetric.LOSS.value}')
log_dict = {
    ValidationMetric.LOSS.value: val_loss,
    ValidationMetric.CODEBOOK_LOSS.value: val_codebook_loss,
}
```

If my understanding aligns with yours, I would leave it for future cleanup because we are not adding new metrics frequently. It is a good suggestion, and we can track as a follow-up imrpovement.

XuesongYang:
@subhankar-ghosh @blisc I addressed all your suggestions. Please have a look.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Sure, let's take it up for future cleanup.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: I meant do something like the following. So if we in the future need to add any additional loss we do not need to touch this code, also makes it much cleaner.

```
VAL_METRIC_LIST = ['val_loss', 'val_codebook_loss', 'val_alignment_loss', ...]
# This is where a global list/Enum of metrics would help, so we can just iterate over them.
# But for now a local list would be fine.

for metric in VAL_METRIC_LIST:
   metric_value = collect_optional_metric(outputs, metric)
   if metric_value is not None:
       log_dict[metric] = metric_value
```

nemo/collections/tts/models/magpietts.py
paarthneekhara: Overall looks good to me. I would recommend just dropping the Legacy validaition path and updating our training sub files on gitlab.

nemo/collections/tts/models/magpietts.py
```diff
@@ -1711,127 +1737,206 @@ def sample_codes_from_logits(
...
+                else:
+                    context_audios.append(None)
+
+            return {
+                'pred_audios': pred_audios,
+                'target_audios': target_audios,
+                'context_audios': context_audios,
+            }
+
+    def _log_media_to_wandb_and_tb(self, media_data: ValidationMediaData, global_step: int) -> None:
```
blisc: I am not a fan of passing dataclasses to functions especially if you do not re-use them. It hides the underlying datatypes of the class and it's not intuitive from the docstring what you are supposed to pass to this function. Is there an argument for why you want to do it this way?

blisc: Let's remove this dataclass for now

nemo/collections/tts/models/magpietts.py
XuesongYang: refactored according to your suggestion since we have more losses from MoE.

nemo/collections/tts/models/magpietts.py
XuesongYang: refactored by removing dataclass.

XuesongYang:
add wandb log example for our MoE training progress: https://wandb.ai/aiapps/debug_magpieTTS_EN_2509_multiValSet/runs/5za0abz7?nw=nwuserxuesong_yang

nemo/collections/tts/models/magpietts.py
XuesongYang: refactored to deprecate legacy validation setup. Now single validation dataset is just a special case where only one item in the multiple dataset list.

XuesongYang:
working on support for non-lhoste data yaml config along with adding new tests covering non-lhoste cases.

nemo/collections/tts/models/magpietts.py
XuesongYang: @subhankar-ghosh could you pls have a look based on my latest changes? I unified `validation_step_outputs` to always use `list[list]` structure, eliminating conditional branching for single vs. multi-dataloader paths.

XuesongYang:
@blisc addressed merging conflicts and fixed bugs of building longform doc.

PR #15381, opened 2026-02-10
[TTS] Fix bugs in MagpieTTS evaluation

rlangman:
# What does this PR do ?

Fix three bugs in MagpieTTS evaluation and validation logic

**Collection**: TTS

# Changelog

- Save `utmosv2` values during evaluation so that they show up in the final evaluation plots
- Fix `_get_context_embeddings` so that after it pads audio and text context tensors to be the same length, it removes extra padding if the output only contains the shorter context type.
- Avoid logging context audio when text context is used. The current logic avoids this if all items in a batch are text context, but does not work when the batch contains a mix and text and audio context examples (which happens during our multi-lingual training). Logging empty audio throws a runtime error if tensorboard logger is enabled.

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/models/magpietts.py
+++ b/nemo/collections/tts/models/magpietts.py
@@ -2024,7 +2024,7 @@ class MagpieTTSModel(ModelPT):
 
         # For 1D tensor - direct use
         context_lens = torch.where(batch['has_text_context'], context_text_lens, context_audio_codes_lens)
-        context_embedded = context_embedded[:, :context_lens.max(), :]
+        context_embedded = context_embedded[:, : context_lens.max(), :]
 
         return context_embedded, context_lens
 


nemo/collections/tts/models/magpietts.py
```diff
@@ -1803,7 +1804,7 @@ def log_val_audio_example(
 
                 if is_wandb:
                     wandb_audio_log[f"Audio/Example_{idx}"] = list()
-                    if context_audio_np is not None:
+                    if context_audio_np is not None and context_audio_np.shape[0] > 0:
```
XuesongYang, line 1807: logging only happens to the first batch at global rank 0. Let's say 1 of 3 examples only has text context, then `context_audio_np` should be `None`, then `context_audio_np` would not have shape. The current test condition seems already covering the scenario.

Am i missing something?

XuesongYang: again, we can simply the changes as below to cover both tests for `None` and/or empty numpy array. 
```suggestion
                    if not context_audio_np:
```

rlangman: Lets say a batch has 3 examples, the first having text conditioning and the other having audio conditioning. `context_audio_np` is not None, and `context_audio_lens[0] = 0`. `context_audio_np = context_audio_np[: context_audio_lens[0]]` will be an empty array, which is when the runtime error happens.

Changing it to  `if not context_audio_np:` also works, but numpy logs a warning saying you are not supposed to do that:

`DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use array.size > 0 to check that an array is not empty.`

PR #15405, opened 2026-02-17
Fix Nanocodec's FSQ codes function: tensor creation misplacement

Ferdydh:
Bug: torch.arange always creates a tensor on cpu, but the FSQ might be moved to another device, crashing codes function when called (because of self.decode function: indices//self.dim_base_index...)

Fix: just move to device when creating tensor.

> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/tts/modules/audio_codec_modules.py
+++ b/nemo/collections/tts/modules/audio_codec_modules.py
@@ -1256,7 +1256,7 @@ class FiniteScalarQuantizer(VectorQuantizerBase):
 
         Note that the codebook entries are implicitly defined by the number of levels.
         """
-        indices = torch.arange(self.codebook_size)
+        indices = torch.arange(self.codebook_size, device=self.dim_base_index.device)
         # [D, B, T]
         indices = rearrange(indices, 'B -> 1 B 1')
         # [B, D, T]


pzelasko:
CC @rlangman @Edresson

blisc:
Most of the properties of this class return native python types not tensors. @rlangman, what is self.codes supposed to return and should it be a torch.tensor?

rlangman:
> Most of the properties of this class return native python types not tensors. @rlangman, what is self.codes supposed to return and should it be a torch.tensor?

`self.codes` and `self.codebook` return a map of the implicit FSQ codebook. For example if the FSQ has levels [5, 5, 5] with codebooks size 5^3=125, then self.codes returns a tensor of size `[125, 3]` with `codes[0] = [0, 0, 0]`, `codes[1] =[0, 0, 0.5]`, etc.

From what I remember these functions are just there to help people visualize what the codebook structure looks like. They would never be used during training or inference.

The reason this can throw an error is because the `decode()` method uses `self.dim_base_index` and `self.num_levels` which are saved as buffers. Even if we returned the value as a native type, it would still encounter this problem when calculating it. In retrospect I think these fields should have been constants instead of buffers/model parameters. It would also work if `self.codes` were computed once at initialization and saved as a normal class field instead of as an `@property`.

The way it is implemented right now, I think the change here is appropriate.

blisc:
Seems like the only failure is https://github.com/NVIDIA-NeMo/NeMo/actions/runs/22234205557/job/64335308428 which is a bit flakey. Merging manually. Thanks @Ferdydh for the fix.

PR #15417, opened 2026-02-20
[TTS] Infrastructure for parallelization of evaluation (scoring)

rfejgin:
This PR makes MagpieTTS's evaluation a little faster and much more parallelizable.

**Changes:**
* Made the ASR step batched, rather than batch-size-1. We first collect all audios that need ASR and run ASR on them in a batched manner before the main per-sample evaluation loop. This substantially speeds up the ASR part of evaluation.
* Infrastructure towards multi-GPU evaluation (scoring). That is something we will do (and have prototyped) with NeMo Skills later on. To enable that, evaluation was broken down into two steps: a first step where evaluation of each utterance is independent of other utternaces, and a second step that focuses on parts that require with global state.
* During refactoring I also removed the constant `STANDARD_METRIC_KEYS` since the set of metrics on which to compute CIs can be inferred from the metrics themselves, which should be easier to maintain.

So `evaluate()` has been broken into:
 1. `evaluate_dir()`: computes metrics for all audios in a given directory+manifest and outputs per-file metrics.
 2. `compute_global_metrics()`: takes per-file metrics collected (1) and computes global metrics from those. This mostly amounts to computing averages. But it also includes computing FCD computation, since that is not something that can easily be broken down into directory-wise chunks due to the statefulness of the FCD metric.
3. `evaluate()`: wrapper that chains (1) and (2) for easy use in NeMo. NeMo skills would call (1) and (2) directly.

Running on a single GPU (local machine), these changes yield ~20% speedup of evaluation (more for larger sets, less for small ones, due to overhead of loading models). The benefit is much larger when evaluating on multiple GPUs in parallel, which we have prototyped in NeMo Skills (and will merge later on).

@vmendelev : adding you just as FYI. I think your existing Skills integration should work as-is after this PR, since the `evaluate()` API hasn't changed. Later, we can break down how Skills does Magpie scoring into the parallelizable part (`evaluate_dir()`) and the aggregation step (`compute_global_metrics()`) – I experimented with that and it seemed to work well.


nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -95,31 +95,26 @@ def evaluate_generated_audio_dir(
...
         return {key: f"{metrics_list[0].get(key, 0):.4f} (single sample)" for key in metric_keys}
 
     results = {}
     for key in metric_keys:
         measurements = [m[key] for m in metrics_list if key in m]
-        if not measurements:
-            logging.warning(f"Metric '{key}' not found in any measurements")
-            results[key] = "N/A"
-            continue
-
```
rfejgin, line 122: This check no longer makes sense since the metric names are now inferred from the metrics themselves.

nemo/collections/tts/modules/magpietts_inference/evaluation.py
```diff
@@ -132,24 +127,4 @@ def compute_mean_with_confidence_interval(
     return results
 
 
-# Define the standard metric keys used in evaluation
-STANDARD_METRIC_KEYS = [
```
rfejgin, line 136: Removed for better maintainability - we can infer these names from the metrics themselves.

nemo/collections/tts/modules/utmosv2.py
```diff
@@ -21,6 +23,14 @@
 import torch
 from threadpoolctl import threadpool_limits
 
+# If UTMOSv2 cache is not set but HF_HOME is, use an area under HF_HOME for the cache location
```
rfejgin, line 26: As part of making evaluation more efficient, we want to ensure UTMOS models don't get re-downloaded each time.

examples/tts/magpietts_inference.py
```diff
@@ -308,8 +307,9 @@ def run_inference_and_evaluation(
             with open(os.path.join(eval_dir, f"{dataset}_metrics_{repeat_idx}.json"), "w") as f:
                 json.dump(metrics, f, indent=4)
 
+            sorted_filewise = sorted(filewise_metrics, key=lambda x: x.get('cer', 0), reverse=True)
```
blisc, line 311: Weren't they already sorted elsewhere? I recall this file's outputs were always sorted

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -186,128 +230,179 @@ def compute_utmosv2_scores(audio_dir, device):
...
+    audio_file_lists = find_generated_audio_files(generated_audio_dir)
+    records = read_manifest(manifest_path)
+    assert len(audio_file_lists) == len(records)
+    codes_file_lists = find_generated_codec_files(generated_audio_dir)
+    has_codes = len(codes_file_lists) == len(records)
+    # Resolve ground-truth and context audio paths for all records
+    gt_audio_paths = [_resolve_path(audio_dir, r.get('audio_filepath')) for r in records]
+    context_audio_paths = [_resolve_path(audio_dir, r.get('context_audio_filepath')) for r in records]
 
+    device = "cuda"
```
blisc: Hmm, guess we always hard-coded this, but should we remove this hardcode?

examples/tts/magpietts_inference.py
blisc: I see, you moved it from nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: Done in latest commit. It's part of EvaluationConfig now, still defaulting to "cuda".

PR #15424, opened 2026-02-24
Add a functional test of frame stacking

rfejgin:
Adding a functional test that verifies that frame stacked models are meeting performance expectations.

In the test, we use a model with a `frame_stacking_factor` of `4` since that tends to uncover more corner cases than lower frame stacking factors. Local transformer inference is enabled, as required for frame-stacked models.

A corresponding checkpoint has been uploaded to CI storage on Azure.

Also: loosened the SSIM threshold for the MoE Longform test since it was sporadically failing with an SSIM slightly below the target.

### DIFF
--- a/tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
+++ b/tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
@@ -31,8 +31,4 @@ TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.cover
     --cer_target 0.11 \
     --ssim_target 0.6 \
     --use_local_transformer \
-    --longform_mode never \
-    --apply_attention_prior \
-    --attention_prior_epsilon 0.1 \
-    --attention_prior_lookahead_window 10 \
-    --estimate_alignment_from_layers 4
+    --longform_mode never 


tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
```diff
@@ -0,0 +1,35 @@
...
+    --use_cfg \
+    --cfg_scale 2.5 \
+    --num_repeats 1 \
+    --temperature 0.6 \
+    --hparams_files /home/TestData/tts/2602_FrameStacking4x/hparams.yaml \
+    --checkpoint_files /home/TestData/tts/2602_FrameStacking4x/frame-stacking-4x-english-nanocodec.ckpt \
+    --apply_attention_prior \
+    --run_evaluation \
+    --clean_up_disk \
+    --cer_target 0.2 \
```
blisc: 20% CER is too high for the latest checkpoints; can you make it tighter?

tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests a 4x-stacked model with local transformer inference.
+
+TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/magpietts_inference.py \
```
blisc, line 17: Please make sure that CI runs your test before you merge

tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
rfejgin: Yes. Tightened it to 10% now. Typical results I get with this checkpoint are between 4% - 7.5%. Setting it to 10% to leave some tolerance for variability.

The checkpoint could probably do better, but I have not tuned inference prior parameters for it as this model is not a release candidate. The main purpose of this test is to catch it it if frame stacking gets broken in the inference code.

tests/functional_tests/L2_TTS_InferEvaluate_Magpietts_FrameStacking.sh
rfejgin: good catch, actaully: just added a call to the test from `.github/workflows/cicd-main-speech.yml`. Will verify it runs before merging.

rfejgin:
Attention prior parameters need to be tuned for it to work well with 4x frame stacking. Disabling it for now for this test.

tests/functional_tests/L2_TTS_InferEvaluatelongform_Magpietts_MoE_ZeroShot.sh
```diff
@@ -24,4 +24,4 @@ TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.cover
     --run_evaluation \
     --clean_up_disk \
     --cer_target 0.08 \
-    --ssim_target 0.50
+    --ssim_target 0.40
```
XuesongYang, line 27: LGTM. This would make L2 pass with high confidence. Thanks!

XuesongYang (dismissed):

PR #15438, opened 2026-02-25
Relax `kaldialign` pin in ASR requirements to unblock Python 3.14 compatibility

basnijholt:
## Summary
- relax `kaldialign` in `requirements/requirements_asr.txt` from `<=0.9.1` to `kaldialign`

## Context
In PR #14775, commit `ce4ad4b` changed this requirement from `kaldialign>=0.9.2` to `kaldialign<=0.9.1` with the message "decrease kaldialign version".

I could not find additional rationale in the PR discussion/review thread, so the reason for the cap is currently not documented for downstream maintainers.

## Why This Change
`kaldialign` now publishes Python 3.14 wheels (e.g. `0.9.3`, including `cp314`).

The `<=0.9.1` cap blocks those newer releases, which prevents consumers from resolving newer Python-compatible builds.

Relaxing the cap allows dependency resolution to use current releases and improves Python 3.14 compatibility.

## Request For Context
@andrusenkoau if there was a specific compatibility reason for the `<=0.9.1` cap, could you share it here? If needed, I can revise this PR to a softer bound (for example `<1`) instead of fully unbounded.


basnijholt:
@andrusenkoau Could you share the original reason for pinning `kaldialign<=0.9.1` in #14775? I could not find context in the thread, and newer `kaldialign` releases are needed for clean Python 3.14 compatibility.

andrusenkoau:
> @andrusenkoau Could you share the original reason for pinning `kaldialign<=0.9.1` in #14775? I could not find context in the thread, and newer `kaldialign` releases are needed for clean Python 3.14 compatibility.

Hi @basnijholt, 
I had a CI error for arm platform with fresh kaldialign versions. I had no other reason to keep kaldialign<=0.9.1. We can fix this limitation once the tests pass.

pzelasko:
Thanks!

pzelasko (approved):

basnijholt:
@pzelasko, thanks again for merging this.

Small follow-up: I noticed `v2.7.1`, `v2.7.2`, and `v2.7.3` were released after this PR, but `v2.7.3` still contains `kaldialign<=0.9.1` in `requirements/requirements_asr.txt`, and the PyPI metadata for `nemo-toolkit==2.7.3` still exposes that cap.

It looks like the 2.7.x releases are cut from the `r2.7.0` branch, while this PR only landed on `main`. As a result, downstream installs on Python 3.14 are still blocked for released NeMo packages, even though current `main` resolves cleanly with newer `kaldialign` wheels.

Would you consider backporting this change to the active 2.7 release branch, or should downstream projects wait for 2.8.0? If a bounded requirement is preferred, e.g. `kaldialign>=0.9.3,<1`, I’m happy to open a small backport PR.

xref: https://github.com/basnijholt/agent-cli/pull/436

PR #15457, opened 2026-03-02
fix(tdt): preserve disable_cuda_graphs state across change_decoding_strategy() calls

### LINKED ISSUE #15423: TDT decoder CUDA graph capture corrupts `torch.load()` — process-wide, irreversible after first `model.transcribe()`
**Describe the bug**

After calling `model.transcribe()` on a Parakeet-TDT model with CUDA graphs enabled (the default), `torch.load()` becomes permanently broken for the remainder of the process. Any attempt to deserialize a payload containing 4 or more tensors fails with:

- `TypeError: 'str' object is not callable` (CUDA tensors)
- `TypeError: 'tuple' object is not callable` (CPU tensors)

The corruption is **process-wide and irreversible** within the same Python process. Once a single `model.transcribe()` call runs with CUDA graphs enabled, `torch.load()` is broken until the process is restarted. This affects any workflow that calls `torch.load()` after inference — including model reloading, checkpoint loading, weight swapping, or any library that internally uses `torch.load()`.

Key diagnostic findings:

1. **Corruption threshold is exactly 4 tensors.** 3 or fewer always works; 4 or more always fails. Independent of tensor size — 4 tensors of shape `(1,)` fail; 3 tensors of shape `(4096, 4096)` succeed.
2. **Raw pickle is unaffected.** Standard `pickle.loads()` / `pickle.dumps()` work normally. The corruption is in PyTorch's custom unpickler or `persistent_load` machinery.
3. **Irreversible.** No combination of `gc.collect()`, `torch.cuda.empty_cache()`, `importlib.reload()`, or model deletion fixes it.
4. **Disabling CUDA graphs prevents it entirely.** `model.decoding.decoding.decoding_computer.disable_cuda_graphs()` before inference prevents the issue.
5. **Disabling `torch._dynamo` has no effect.**
6. **No performance penalty.** Disabling CUDA graphs actually resulted in ~15% faster wall-clock time for single-file transcription.

```python
import io, wave
from collections import OrderedDict
import numpy as np
import torch

# 1. Create minimal test audio (5 seconds of silence)
samples = np.zeros(16000 * 5, dtype=np.int16)
with wave.open('/tmp/test.wav', 'wb') as wf:
    wf.setnchannels(1)
    wf.setsampwidth(2)
    wf.setframerate(16000)
    wf.writeframes(samples.tobytes())

# 2. Load model (CUDA graphs enabled by default in TDT decoder)
model = nemo_asr.models.ASRModel.from_pretrained('nvidia/parakeet-tdt-0.6b-v3')
model = model.cuda().eval()

# 3. Verify torch.load works BEFORE inference
sd = OrderedDict((f'layer.{i}', torch.randn(32, 32, device='cuda')) for i in range(10))
buf = io.BytesIO()
torch.save(sd, buf)
buf.seek(0)
loaded = torch.load(buf, weights_only=False)
print(f"Before inference: torch.load OK, loaded {len(loaded)} tensors")

# 4. Run inference (triggers CUDA graph capture in the TDT decoder)
model.transcribe(['/tmp/test.wav'])

# 5. torch.load is now BROKEN for 4+ tensors
sd = OrderedDict((f'layer.{i}', torch.randn(32, 32, device='cuda')) for i in range(10))
buf = io.BytesIO()
torch.save(sd, buf)
buf.seek(0)
try:
    torch.load(buf, weights_only=False)
    print("After inference: torch.load OK")
except TypeError as e:
    print(f"After inference: torch.load FAILED — {e}")
    # TypeError: 'str' object is not callable
```

Workaround — disable CUDA graphs before inference:

```python
model = nemo_asr.models.ASRModel.from_pretrained('nvidia/parakeet-tdt-0.6b-v3')
model = model.cuda().eval()
model.decoding.decoding.decoding_computer.disable_cuda_graphs()
model.transcribe(['/tmp/test.wav'])  # torch.load works fine after this
```

**Expected behavior**

`torch.load()` should work normally after `model.transcribe()`. CUDA graph capture/replay in the TDT decoder should not have side effects on unrelated PyTorch subsystems like serialization.

- Environment location: Docker (NVIDIA DGX Spark, Grace Blackwell aarch64)
- Method of NeMo install: Docker
- Docker image: `nvcr.io/nvidia/nemo:25.11.nemotron_3_nano`
- `docker pull nvcr.io/nvidia/nemo:25.11.nemotron_3_nano`

**Environment details**

NVIDIA docker image is used. Additional details for reference:

- NeMo: 2.6.0rc0
- PyTorch: 2.9.0a0+50eac811a6.nv25.09
- CUDA Toolkit: 13.0
- NVIDIA Driver: 580.126.09
- Python: 3.12.3
- Platform: Linux aarch64

**Additional context**

- GPU: NVIDIA GB10 (DGX Spark, Grace Blackwell, unified memory, compute capability 12.1)
- The corruption is triggered specifically by the CUDA graph capture/replay in `GreedyBatchedTDTLabelLoopingComputer`. The code path is: `model.transcribe()` → `GreedyBatchedTDTInfer.forward()` → `GreedyBatchedTDTLabelLoopingComputer.__call__()` → `cuda_graphs_impl()`.
- The CUDA graph code itself never touches pickle or `torch.load` — the corruption is an indirect side effect of graph capture on PyTorch's internal tensor reconstruction dispatch.
- A corresponding issue will be filed with PyTorch, as the root cause may be in PyTorch's serialization internals being corrupted by CUDA graph state changes.
- This may be related to PyTorch PR [#165474](https://github.com/pytorch/pytorch/pull/165474) (merged in 2.10) which fixes AOTAutogradCache serialization — untested on 2.10+ yet.

CodersAcademy006:
## Problem
When `disable_cuda_graphs()` is called on the decoding computer, the disabled
state is lost whenever `change_decoding_strategy()` is invoked — because it
creates a brand new `decoding_computer` object with CUDA graphs re-enabled
by default (`cuda_graphs_mode='full_graph'`).

This silently re-enables CUDA graphs even when the user explicitly disabled
them, re-introducing the `torch.load()` corruption bug reported in #15423.

Concretely, calling `model.transcribe(timestamps=True)` triggers
`change_decoding_strategy()` internally, which replaces the entire
`decoding_computer` object — resetting CUDA graph state in the process.

## Root Cause
- `disable_cuda_graphs()` state was only stored inside `_decoding_computer`
- `change_decoding_strategy()` creates a completely new `_decoding_computer`
- The new object always has CUDA graphs enabled by default
- No state was preserved or restored across this replacement

## Fix
Two minimal changes across two files:

1. `tdt_decoding.py` — `BeamBatchedTDTInfer.disable_cuda_graphs()`:
   Track the disabled state via a `_cuda_graphs_disabled = True` flag on the
   outer object so it survives `_decoding_computer` replacement.

2. `rnnt_models.py` — `EncDecRNNTModel.change_decoding_strategy()`:
   - Read `_cuda_graphs_disabled` from the current `decoding_computer` BEFORE
     replacing `self.decoding`
   - After the new `self.decoding` is created, restore the disabled state on
     the new `decoding_computer`
   - Both read and write are wrapped in `try/except AttributeError` to remain
     safe across all model types that call this method

## Testing
The reproducer from #15423 no longer corrupts `torch.load()` after
`transcribe(timestamps=True)` with this fix applied:
```python
model = nemo_asr.models.ASRModel.from_pretrained('nvidia/parakeet-tdt-0.6b-v3')
model = model.cuda().eval()
model.decoding.decoding.decoding_computer.disable_cuda_graphs()
model.transcribe(['/tmp/test.wav'], timestamps=True, return_hypotheses=True)

# torch.load now works correctly after this — previously failed with:
# TypeError: 'str' object is not callable
sd = OrderedDict((f'layer.{i}', torch.randn(32, 32, device='cuda')) for i in range(10))
buf = io.BytesIO()
torch.save(sd, buf)
buf.seek(0)
torch.load(buf, weights_only=False)  # ✅ Works
```

Fixes #15423

## Notes
- Zero breaking changes — all existing behaviour is preserved
- The `try/except AttributeError` guards make this safe for all RNNT model
  variants, not just TDT
- The deeper PyTorch-side corruption of `torch.load()` dispatch table entries
  during CUDA graph capture is a separate upstream issue and should be tracked
  in pytorch/pytorch


CodersAcademy006:
@rsclafani Please look into this as well, Thank You. And please provide me what else i can change in this?

artbataev:
Thanks for the contribution.
Unfortunately, I need to close this PR: the correct way to disable CUDA graphs is to change the values in the config.

```python
from omegaconf import open_dict

# greedy decoding
with open_dict(model.cfg.decoding.greedy):
    model.cfg.decoding.greedy.use_cuda_graph_decoder = False

# beam decoding
with open_dict(model.cfg.decoding.beam):
    model.cfg.decoding.beam.allow_cuda_graphs = False

model.change_decoding_strategy(model.cfg.decoding)
```

We should not introduce such stateful changes - `change_decoding_strategy` expected to fully reinstantiate the decoder using the provided decoding parameters.

CodersAcademy006:
@artbataev I sincerely apologize if I moved forward without formal confirmation. Please guide me on the necessary steps to rectify this and ensure the issue is completely resolved.

PR #15462, opened 2026-03-04
End-of-Utterance metric

rfejgin:
This PR adds a metric that measures end-of-utterance performance.

Each utterance is classified as one of:
* "good": natural ending
* "cutoff": early cutoff
* "silence": a long silence at the end
* "noise": noise after the last word

New metrics reported by the evaluation script are `eou_cutoff_rate`, `eou_silence_rate`, `eou_noise_rate` and `eou_error_rate`, where `eou_error_rate` accumulates all non-good cases (cutoff OR silence OR noise).

## End-of-Utterance (EoU) Classifier Algorithm

This classifier detects whether a speech sample ends naturally or has an artifact (cutoff, trailing silence, or trailing noise). It works in two stages:

### 1. CTC Forced Alignment

We run the generated audio through a pretrained Wav2Vec2 CTC model (`facebook/wav2vec2-base-960h`) and use NeMo Forced Aligner's `viterbi_decoding` to force-align the audio frames to the target transcript. This produces per-character token segments with timestamps and confidence scores. The end of the last aligned token gives us and estimate of the **speech boundary** — the point where intelligible speech ends and any trailing audio begins.

### 2. Trailing-Region Analysis & Classification

Using the speech boundary, we split the audio into a speech region and a trailing region (with a small padding of the speech region of 100-150ms to account for Wav2Vec2's end-of-segment inaccuracy). We then extract features from both regions:

- **Trailing duration** — how much audio remains after aligned speech ends
- **Trail RMS ratio** — RMS energy of the tail region relative to the full utterance
- **Last token confidence** — alignment confidence of the final speech token (falls back to average of last two tokens if near zero, as a zero-confidence was found to be an unreliable estimate of actual confidence).
- **Last token gap** — length in seconds of the blank-frame gap between the last two aligned tokens — tends to be large when there is junk audio at the end.

These features feed a simple decision tree:

| Condition | Label |
|---|---|
| Short tail (< 0.1s) + low confidence + no large gap | **cutoff** |
| Long noisy tail (> 0.15s, RMS ratio > 0.4) OR large gap + low confidence | **noise** |
| Long tail (> 1.4s) without high energy | **silence** |
| Everything else | **good** |

### 3. Additional Notes
* Utterances whose ending includes _both_ noise and silence can be classified as either class, but tend to be classified as silence, especially if the trailing silence is long.
* Utterances with looping at the end are not considered a separate category, but I tested a few and their ending was classified as "noise", as desired.

## Accuracy
This metric is still somewhat experimental and we don't have enough labeled data to measure its accuracy. But it was tested on both ground truth test sets and noisy generated speech sets and worked quite well. It was found to be,
* Good at detecting utterances with cutoff and silence tails, in the sense that when it detects them, it's usually correct (low false positives). It was able to find cutoff issues in a ground-truth test set that we were previously unaware included such issues.
* Occasionally under or over-detect noise tails. That said, on a test set of 122 utterances nearly all of which likely contain a noise tail, it detected 97% utterances as noisy. For LibriTTS ground truth data, it only flagged 4/2275 as noisy (and one of those was indeed noisy). For studio-quality data (Riva), it classified all 100 utterances tested as good. 

## Limitations
Support English only, for now. The metric is set to `nan` for other languages.

## Speed
It's pretty fast – the metric processed 2300 utterances in about 36 seconds on my machine. The speed is aided by batching.

## Testing
Included is a unit test that verifies some utterances whose ending types is known are classified correctly.

### DIFF
--- a/docs/source/speechlm2/intro.rst
+++ b/docs/source/speechlm2/intro.rst
@@ -4,16 +4,21 @@ SpeechLM2
 .. note::
    The SpeechLM2 collection is still in active development and the code is likely to keep changing.
 
+
+
 SpeechLM2 refers to a collection that augments pre-trained Large Language Models (LLMs) with speech understanding and generation capabilities. 
 
 This collection is designed to be compact, efficient, and to support easy swapping of different LLMs backed by HuggingFace AutoModel. 
 It has a first-class support for using dynamic batch sizes via Lhotse and various model parallelism techniques (e.g., FSDP2, Tensor Parallel, Sequence Parallel) via PyTorch DTensor API.
 
-We currently support four main model types:
-* SALM (Speech-Augmented Language Model) - a simple but effective approach to augmenting pre-trained LLMs with speech understanding capabilities.
-* DuplexS2SModel - a full-duplex speech-to-speech model with an ASR encoder, directly predicting discrete audio codes.
-* DuplexS2SSpeechDecoderModel - a variant of DuplexS2SModel with a separate transformer decoder for speech generation.
-* DuplexSTTModel - a decoder model to generate agent text in duplex, in response to both user speech and text inputs.
+We currently support six main model types:
+
+* **SALM** (Speech-Augmented Language Model) - a simple but effective approach to augmenting pre-trained LLMs with speech understanding capabilities.
+* **DuplexS2SModel** - a full-duplex speech-to-speech model with an ASR encoder, directly predicting discrete audio codes.
+* **DuplexS2SSpeechDecoderModel** - a variant of DuplexS2SModel with a separate transformer decoder for speech generation.
+* **DuplexEARTTS** - a ready-to-use duplex text-to-speech model that supports user interruption via a special text interruption token.
+* **DuplexSTTModel** - a decoder model to generate agent text in duplex, in response to both user speech and text inputs.
+* **NemotronVoiceChat** - an *inference-only* pipeline that seamlessly merges `DuplexSTTModel` and `DuplexEARTTS` to deliver an end-to-end, full-duplex conversational agent with high-fidelity speech generation.
 
 Using Pretrained Models
 -----------------------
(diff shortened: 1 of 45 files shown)


nemo/collections/tts/metrics/eou_classifier.py
```diff
@@ -0,0 +1,481 @@
...
+from typing import Union
+
+import librosa
+import numpy as np
+import torch
+from transformers import Wav2Vec2ForCTC, Wav2Vec2Processor
+
+from nemo.collections.asr.parts.utils.aligner_utils import viterbi_decoding
+
+SR = 16000
```
blisc: If you only use this magic number once, add it to the function definition itself rather than make a global variable

nemo/collections/tts/metrics/eou_classifier.py
```diff
@@ -0,0 +1,481 @@
...
+    print(f"trailing_duration:  {result.trailing_duration:.3f}s")
+    print(f"trail_rms_ratio:    {result.trail_rms_ratio:.4f}")
+    print(f"last_token_dur:     {result.last_token_duration:.3f}s")
+    print(f"last_token_conf:    {result.last_token_confidence:.3f}")
+    print(f"last_token_gap:     {result.last_token_gap:.3f}s")
+    print(f"last_2_ph_avg_conf: {result.last_two_phoneme_avg_confidence:.3f}")
+    print(f"last_token:         {result.last_token!r}")
+    print(f"\nToken segments ({len(result.token_segments)}):")
+    for seg in result.token_segments:
+        print(f"  {seg.token!r:<6} {seg.start:.3f}-{seg.end:.3f}s  dur={seg.duration:.3f}s  conf={seg.confidence:.3f}")
```
blisc: Do you want this API in this file? I would recommend removal

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -662,6 +721,21 @@ def compute_global_metrics(
     avg_metrics['utmosv2_avg'] = sum(m['utmosv2'] for m in filewise_metrics) / n
     avg_metrics['total_gen_audio_seconds'] = sum(m['total_gen_audio_seconds'] for m in filewise_metrics)
 
+    # EoU classification rates
+    eou_types = [m.get('eou_type') for m in filewise_metrics]
+    if eou_types[0] is not None:
+        from collections import Counter
```
blisc: Move import statements to the top of the file

nemo/collections/tts/metrics/eou_classifier.py
```diff
@@ -0,0 +1,481 @@
...
+        """Extract end-of-utterance features from per-frame log_probs and a Viterbi alignment path.
+
+        Args:
+            log_probs: (T, V) log-probability tensor for a single sample.
+            token_ids_with_blanks: Interleaved-blank token sequence.
+            alignment_path: Viterbi path indices into token_ids_with_blanks.
+
+        Returns:
+            Dict with information about when the speech ended, what the last token was,
+            what its confidence was, and detailed per-segment information.
```
blisc, line 162: This dictionary needs to be flattened out or moved to a dataclass. It is very dense

nemo/collections/tts/metrics/eou_classifier.py
rfejgin: Yeah, makes sense - removed

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
rfejgin: Moved

nemo/collections/tts/metrics/eou_classifier.py
rfejgin: Sure - I've now wrapped this as a dataclass called `AlignmentFeatures`. It is also reused by in the EoUClassification where it's now one of the fields, which reduces duplication.

nemo/collections/tts/metrics/eou_classifier.py
rfejgin: Done

PR #15470, opened 2026-03-06
[core] Unify validation_step_outputs to always return list-of-lists

XuesongYang:
# What does this PR do ?

Unify `ModelPT.validation_step_outputs` (and `test_step_outputs`) to always return a list of lists, so a single dataloader is simply the N=1 case and subclasses no longer need to branch on the output shape. Normalize both `_validation_dl` and `_test_dl` to `Optional[List[DataLoader]]` via their respective resolvers.

**Collection**: Core, ASR, TTS, Audio

# Changelog

- `modelPT.py`: `validation_step_outputs` / `test_step_outputs` properties always return `[[] for _ in range(num_dl)]`; `on_validation_epoch_end` / `on_test_epoch_end` use `len() == 1` instead of `isinstance(..., dict)` for single-vs-multi dispatch; empty-output guard updated to `all(len(o) == 0 for o in ...)` since `[[]]` is truthy; empty dataloader buckets skipped in multi-DL loop
- `model_utils.py`: `resolve_validation_dataloaders` and `resolve_test_dataloaders` wrap bare `DataLoader` into `[DataLoader]` at both single-value paths, normalizing `_validation_dl` and `_test_dl` to `Optional[List[DataLoader]]`
- `modelPT.py` (`setup_multiple_validation_data`): type annotation updated; `isinstance` guard simplified to truthiness check after normalization
- 15 model files (ASR, TTS G2P, Audio): remove `if/else` branching in `validation_step` / `test_step`; always use `self.validation_step_outputs[dataloader_idx].append(...)`
- `transformer_bpe_models.py`: remove `isinstance(outputs[0], dict)` normalization loop in `multi_validation_epoch_end` — base class now iterates dataloaders and calls it once per DL
- `audio_to_audio.py`: simplify `_get_num_dataloaders` (both val and test) and logging callback setup after normalization
- `fastpitch.py`, `magpietts.py`, `magpietts_preference_optimization.py`: add `RuntimeError` guard for `len(validation_step_outputs) != 1`; add early-return on empty outputs; use `self.validation_step_outputs[0]` consistently
- `ssl_models.py`: fix `EncDecMaskedTokenPredModel.test_step` — was appending to `validation_step_outputs` instead of `test_step_outputs`
- Test models (`test_ema.py`, `check_for_ranks.py`, `test_ptl_stateless_timer.py`): override `multi_validation_epoch_end` instead of `on_validation_epoch_end`; base class handles iteration, clearing, and per-DL prefix

# Usage

No API changes for single-dataloader models — `dataloader_idx=0` is the default. Subclasses should use the `[dataloader_idx]` indexing pattern:

```python
# In validation_step:
def validation_step(self, batch, batch_idx, dataloader_idx=0):
    metrics = self.compute_metrics(batch)
    self.validation_step_outputs[dataloader_idx].append(metrics)
    return metrics

# In multi_validation_epoch_end (preferred over on_validation_epoch_end):
def multi_validation_epoch_end(self, outputs, dataloader_idx=0):
    avg_loss = torch.stack([x['val_loss'] for x in outputs]).mean()
    self.log("val_loss", avg_loss)
```

### DIFF
--- a/nemo/collections/audio/models/audio_to_audio.py
+++ b/nemo/collections/audio/models/audio_to_audio.py
@@ -62,7 +62,7 @@ class AudioToAudioModel(ModelPT, ABC):
 
     def _get_num_dataloaders(self, tag: str = 'val'):
         if tag == 'val':
-            num_dataloaders = len(self._validation_dl) if isinstance(self._validation_dl, List) else 1
+            num_dataloaders = len(self._validation_dl) if self._validation_dl else 1
         elif tag == 'test':
             num_dataloaders = len(self._test_dl) if isinstance(self._test_dl, List) else 1
         else:
@@ -144,18 +144,12 @@ class AudioToAudioModel(ModelPT, ABC):
 
     def validation_step(self, batch, batch_idx, dataloader_idx: int = 0):
         output_dict = self.evaluation_step(batch, batch_idx, dataloader_idx, 'val')
-        if isinstance(self.trainer.val_dataloaders, (list, tuple)) and len(self.trainer.val_dataloaders) > 1:
-            self.validation_step_outputs[dataloader_idx].append(output_dict)
-        else:
-            self.validation_step_outputs.append(output_dict)
+        self.validation_step_outputs[dataloader_idx].append(output_dict)
         return output_dict
 
     def test_step(self, batch, batch_idx, dataloader_idx=0):
         output_dict = self.evaluation_step(batch, batch_idx, dataloader_idx, 'test')
-        if isinstance(self.trainer.test_dataloaders, (list, tuple)) and len(self.trainer.test_dataloaders) > 1:
-            self.test_step_outputs[dataloader_idx].append(output_dict)
-        else:
-            self.test_step_outputs.append(output_dict)
+        self.test_step_outputs[dataloader_idx].append(output_dict)
         return output_dict
 
     def multi_evaluation_epoch_end(self, outputs, dataloader_idx: int = 0, tag: str = 'val'):
@@ -514,10 +508,7 @@ class AudioToAudioModel(ModelPT, ABC):
         log_callbacks = []
         from nemo.collections.audio.parts.utils.callbacks import SpeechEnhancementLoggingCallback
 
-        if isinstance(self._validation_dl, List):
-            data_loaders = self._validation_dl
-        else:
-            data_loaders = [self._validation_dl]
+        data_loaders = self._validation_dl if self._validation_dl else []
 
         for data_loader_idx, data_loader in enumerate(data_loaders):
             log_callbacks.append(
--- a/nemo/collections/audio/models/audio_to_audio.py
+++ b/nemo/collections/audio/models/audio_to_audio.py
@@ -64,7 +64,7 @@ class AudioToAudioModel(ModelPT, ABC):
         if tag == 'val':
             num_dataloaders = len(self._validation_dl) if self._validation_dl else 1
         elif tag == 'test':
-            num_dataloaders = len(self._test_dl) if isinstance(self._test_dl, List) else 1
+            num_dataloaders = len(self._test_dl) if self._test_dl else 1
         else:
             raise ValueError(f'Unexpected tag {tag}.')
 
--- a/nemo/core/classes/modelPT.py
+++ b/nemo/core/classes/modelPT.py
@@ -585,7 +585,7 @@ class ModelPT(LightningModule, Model):
         # Set some placeholder overriden by helper method
         self._val_dl_idx: int = 0
         self._validation_names: Optional[List[str]] = None
-        self._validation_dl: Optional[torch.utils.data.DataLoader] = None
+        self._validation_dl: Optional[List[torch.utils.data.DataLoader]] = None
 
         # preserve config
         self._update_dataset_config(dataset_name='validation', config=val_data_config)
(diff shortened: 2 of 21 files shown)


nemo/collections/audio/models/audio_to_audio.py
```diff
@@ -62,9 +62,9 @@ def _setup_loss(self):
 
     def _get_num_dataloaders(self, tag: str = 'val'):
         if tag == 'val':
-            num_dataloaders = len(self._validation_dl) if isinstance(self._validation_dl, List) else 1
+            num_dataloaders = len(self._validation_dl) if self._validation_dl else 1
```
pzelasko, line 65: technically I think you want 

```suggestion
            num_dataloaders = len(self._validation_dl) if self._validation_dl is not None else 1
```

and similar everywhere else - make it clear this is a None check

pzelasko (commented):
Looks good in general, but before I approve - which models did you test it on, and did you test with both single and multi validation dataset?

nithinraok (commented):
LGTM in general. How are you validating the changes? CI only runs on single gpu have you tested on multiple GPUs with multiple nodes?

nemo/core/classes/modelPT.py
```diff
@@ -1692,26 +1698,17 @@ def hparams(self):
...
-        # list: [<Dataloader obj>] when ds_item in config has 1 item passed in a list
-        if (
-            self._validation_dl is not None
-            and isinstance(self._validation_dl, (list, tuple))
-            and len(self._validation_dl) > 1
-        ):
-            for _ in range(len(self._validation_dl)):
-                self._validation_step_outputs.append([])
+        num_dl = len(self._validation_dl) if self._validation_dl else 1
+        self._validation_step_outputs = [[] for _ in range(num_dl)]
```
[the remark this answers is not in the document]
XuesongYang: fixed.

nemo/core/classes/modelPT.py
```diff
@@ -1722,22 +1719,17 @@ def validation_step_outputs(self, value):
...
 
-        # Initialize new output list
-        self._test_step_outputs = []
-        # Check len(self._test_dl) > 1 as sometimes single dataloader can be in a list: [<Dataloader obj>]
-        # when ds_item in config has 1 item passed in a list
-        if self._test_dl is not None and isinstance(self._test_dl, (list, tuple)) and len(self._test_dl) > 1:
-            for _ in range(len(self._test_dl)):
-                self._test_step_outputs.append([])
+        num_dl = len(self._test_dl) if self._test_dl else 1
+        self._test_step_outputs = [[] for _ in range(num_dl)]
```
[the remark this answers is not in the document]
XuesongYang: fixed.

nemo/collections/audio/models/audio_to_audio.py
XuesongYang: The truthiness check (`if self._validation_dl`) and `if self._validation_dl is not None` are equivalent.

because after resolver normalization, `self._validation_dl` is always either `None` or a `non-empty List[DataLoader]`. The `None` case is converted to `[]` by `val_dataloader()` (PTL 2.0+ doesn't accept None), but PTL then skips validation entirely for an empty list, so `on_validation_start` or _`get_num_dataloaders` never executes with `[]`.

pzelasko: Why are we setting `num_dataloaders = 1` when the list is empty then?

XuesongYang:
@pzelasko @nithinraok all questions regarding test by modeling are valid.

In my experience, 
* I only tested by training MagpieTTS on multi-node setup, and it worked as expected.
* CI covers single-GPU across the changed ASR, TTS, and Audio model files

If we want to proceed with the move, we may need to test EncDecCTCModel, EncDecRNNTModel, FastPitch, AudioToAudio, ..., on multi-val dataloaders on multi-gpus integration tests. Do we have such resources/tests on CI?

pzelasko:
> @pzelasko @nithinraok all questions regarding test by modeling are valid.
> 
> In my experience,
> 
> * I only tested by training MagpieTTS on multi-node setup, and it worked as expected.
> * CI covers single-GPU across the changed ASR, TTS, and Audio model files
> 
> If we want to proceed with the move, we may need to test EncDecCTCModel, EncDecRNNTModel, FastPitch, AudioToAudio, ..., on multi-val dataloaders on multi-gpus integration tests. Do we have such resources/tests on CI?

I'm afraid we don't have CI tests for these conditions. I think we can use up to 2 GPUs in CI currently (cc @chtruong814 to verify). It might be a good idea to add at least one test multi-GPU multi-validation training of some representative model for ~10 training steps with 1 validation and parse the CLI output to verify it's OK (Magpie is probably OK, or maybe Parakeet as it's most popular?).

nemo/core/classes/modelPT.py
```diff
@@ -1692,26 +1698,20 @@ def hparams(self):
...
-            self._validation_dl is not None
-            and isinstance(self._validation_dl, (list, tuple))
-            and len(self._validation_dl) > 1
-        ):
-            for _ in range(len(self._validation_dl)):
-                self._validation_step_outputs.append([])
+        if isinstance(self._validation_dl, (list, tuple)) and len(self._validation_dl) > 0:
+            num_dl = len(self._validation_dl)
+        else:
+            num_dl = 1
```
stevehuang52, line 1713: Why do we set num_dl = 1 when there's no dataloader? What's the reason that we don't set it to zero when no dataloader provided?

stevehuang52: In `setup_multiple_test_data` you have:
```
# Case where we dont provide data loaders, or all dataloaders produced no batches.
if not self.validation_step_outputs or all(len(outputs) == 0 for outputs in self.validation_step_outputs):
```
which implies that `self.validation_step_outputs` can be either `None` or its length can be zero, then why set do we need it to be of length 1 when no dataloaders?

PR #15477, opened 2026-03-09
[MagpieTTS] Magpietts longform unify

subhankar-ghosh:
> [!IMPORTANT]  

# What does this PR do ?

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/tts/models/magpietts.py
```diff
@@ -4782,13 +4789,12 @@ def generate_speech(
 
             predicted_codes = torch.cat(state.all_predictions, dim=-1)  # (B, C, F*T_steps)
             num_steps = len(state.all_predictions)
+            default_frame_len = num_steps * self.frame_stacking_factor
```
rfejgin, line 5179: Could you add back the comment that was here originally, I think it got lost:
`            # Concatenate the list of predictions along the time dimension. Note that when frame stacking is on, this also undoes the stacking.`

nemo/collections/tts/models/magpietts.py
```diff
@@ -4570,8 +4574,10 @@ def generate_speech(
                 finished_texts_counter={},
                 attn_prior=initial_attn_prior,
             )
+            chunk_end_frame_lens: Dict[int, int] = {}
```
rfejgin, line 4962: Could you add a comment saying what this tracks and if it persists across chunked calls to generate_speech()? Because it appears that this one keeps state locally, unlike chunk_state which is persistent between calls but not super clear from the naming.

Side note, maybe we could find a better name than `chunk_state` since that structure appears not to be associated with a particular chunk but rather tracks overall inference state (I think). E.g. could call it inference_state or chunk**ed**_inference_state (the latter is admittedly kind of verbose).

nemo/collections/tts/models/magpietts.py
```diff
@@ -4137,9 +4137,10 @@ def _check_eos_and_update_state(
 
         Args:
             chunk_state: Mutable state object tracking history across chunks.
-            audio_codes_next: Sampled audio codes. Shape: (B, num_codebooks).
+            audio_codes_next: Sampled audio codes. Shape: (B, num_codebooks) or (B, num_codebooks, frame_stacking_factor).
```
rfejgin: Isn't it always 3-dimensional, with `frame_stacking_factor` being 1 if there's no frame stacking?

rfejgin (commented):
Overall looks good but please see comments.
Also, I earlier confirmed that CERs for the frame stacking model look good after this change (same as pre-unification), so it looks like this fixes the issue we observed.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Added. It does not maintain state across generate_speech calls. It is local only.

nemo/collections/tts/models/magpietts.py
subhankar-ghosh: Added

PR #15502, opened 2026-03-16
remove protobuf from reqs

nithinraok:
# What does this PR do ?

Remove protobuf from requirements

**Collection**: All

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/core/classes/mixins/hf_io_mixin.py
+++ b/nemo/core/classes/mixins/hf_io_mixin.py
@@ -116,9 +116,7 @@ class HuggingFaceFileIO(ABC):
         # Setup extra arguments for model filtering
         all_results = []  # type: List[ModelInfo]
 
-        results = api.list_models(
-            token=hf_token, sort="lastModified", **model_filter
-        )  # type: Iterable[ModelInfo]
+        results = api.list_models(token=hf_token, sort="lastModified", **model_filter)  # type: Iterable[ModelInfo]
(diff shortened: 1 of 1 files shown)


pzelasko:
Do we need to directly depend on protobuf? We are not using it ourselves - let's let our transitive dependencies pull it instead?

nithinraok:
For that, we should add wandb into our dependencies to avoid protobuf conflicts later.

nithinraok:
It is currently installed via TensorBoard. Even if we do not explicitly include it in our dependencies, it will still be installed transitively without version pinning.

We can add wandb and tensorboard to our dependencies since most of our training workflows rely on wandb. Then we can remove protobuf as an explicit dependency from the requirements.

pzelasko:
wandb is already inside requirements_lightning.txt which is always installed
tensorboard(x) is automatically pulled by pytorch

there is no dependency we don't have already :)

nithinraok:
Interesting, thanks Piotr. I am confused now why wandb is not part of latest containers. For this PR, I am removing explicit installation of protobuf.

pzelasko:
Makes sense. I don't know about containers. Let's make sure the next one has everything we need.

PR #15503, opened 2026-03-16
zero shot disabling magpie

paarthneekhara:


nemo/collections/tts/models/magpietts.py
```diff
@@ -601,6 +601,7 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
             self.register_buffer('_baked_embedding_T', None)  # Time dimension
             self.register_buffer('_baked_embedding_D', None)  # Embedding dimension
             self.register_buffer('baked_context_embedding_len', None)  # Per-speaker lengths (N,)
+            self.zero_shot_disable_prob = cfg.get('zero_shot_disable_prob', 0.0)
```
blisc: Can you add the comment about what this does here as well?
Let's add a check that the batch size is greater than 1.
Can we rename this to match the functionality of this argument. Maybe train_shuffle_context_embedding_prob. It's becoming really long, but not sure what else we can name it as.

nemo/collections/tts/models/magpietts.py
```diff
@@ -2299,9 +2300,16 @@ def _prepare_decoder_context(
                 context_input_lens = context_input_lens.to(text.device)
                 context_mask = get_mask_from_lengths(context_input_lens)
             else:
-                context_embeddings = self.context_encoder(
-                    context_input_embedded, context_mask, cond=None, cond_mask=None
-                )['output']
+                # Zero-shot disable: with some probability, bypass the context encoder and feed
+                # batch-shuffled raw embeddings so the model learns to not clone from untransformed input.
+                if self.training and self.zero_shot_disable_prob > 0 and random.random() < self.zero_shot_disable_prob:
+                    shuffle_idx = torch.randperm(context_input_embedded.size(0), device=context_input_embedded.device)
```
blisc: Instead of a randperm, can we rotate by some number. It could be a random number or just fixed

nemo/collections/tts/models/magpietts.py
paarthneekhara: Batch size can be dynamic in lhotse. I added a check to not go into the random permutation branch of the code, if batch size is 1.

PR #15504, opened 2026-03-16
MagpieTTS refactor 

paarthneekhara:
This change is mainly done because EasyMagpie will be reusing some shared functionalities with Magpie. So to avoid code duplication, we are moving common things together. 

After this, i will raise a separate PR for EasyMagpie changes.


nemo/collections/tts/models/magpietts.py
```diff
@@ -759,30 +778,41 @@ def _setup_inference_parameters(self, cfg: DictConfig) -> None:
...
+    def _get_state_dict_keys_to_exclude(self):
+        keys = ['_speaker_verification_model', '_codec_model']
+        if self.has_baked_context_embedding:
+            keys.append('context_encoder')
+        return keys
 
-        If the model has a baked context embedding, the context_encoder weights are also excluded
-        since they are no longer needed for inference.
-        """
+    def state_dict(self, destination=None, prefix='', keep_vars=False):
```
blisc, line 794: Can you add back the docstring?

nemo/collections/tts/models/magpietts.py
```diff
@@ -759,30 +778,41 @@ def _setup_inference_parameters(self, cfg: DictConfig) -> None:
...
         self.inference_parameters = ModelInferenceParameters.from_dict(cfg.get("inference_parameters", {}))
 
-    def state_dict(self, destination=None, prefix='', keep_vars=False):
-        """
-        Only used for saving checkpoints. On save, we remove _speaker_verification_model and _codec_model
-        from the checkpoint. The codec model is saved in a separate checkpoint.
-
-        _speaker_verification_model is only included in older checkpoints with the older single_encoder_sv_tts
-        model_type that is no longer supported and can likely be removed in a future version.
+    def _get_state_dict_keys_to_exclude(self):
```
blisc, line 780: Can you add a docstring to this?

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -251,3 +256,536 @@ def forward(self, subword_ids: Tensor, subword_mask: Tensor | None = None) -> Te
...
+def remove_bos_token(codes, codes_len, num_tokens=1):
+    codes = codes[:, :, num_tokens:]
+    codes_len = codes_len - num_tokens
+    return codes, codes_len
+
+
+def remove_embedded_bos_token(embedded, embedded_len):
+    embedded = embedded[:, 1:, :]
+    embedded_len = embedded_len - 1
+    return embedded, embedded_len
```
blisc, line 315: These two functions look identical. Do we need both?

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -251,3 +256,536 @@ def forward(self, subword_ids: Tensor, subword_mask: Tensor | None = None) -> Te
...
+    """Remove the last token from embedded sequences.
+
+    Args:
+        embedded: (B, T', D)
+    """
+    embedded_len = embedded_len - 1
+    embedded = embedded[:, :-1, :]
+    mask = get_mask_from_lengths(lengths=embedded_len)
+    embedded = embedded * mask.unsqueeze(2)
+    return embedded, embedded_len
```
blisc, line 336: These two functions look identical. Do we need both?

nemo/collections/tts/modules/magpietts_modules.py
paarthneekhara: Yeah, both of them were there earlier and being used. One does the removal in the embedded tensor and one does the removal in code tensor (before embedding). They are also different in their implementation.

nemo/collections/tts/modules/magpietts_modules.py
paarthneekhara: Same as the above.

nemo/collections/tts/models/magpietts.py
paarthneekhara: added the docstrings.

nemo/collections/tts/models/magpietts.py
paarthneekhara: added the docstrings.

nemo/collections/tts/models/magpietts.py
blisc: It looks like they got switched around, I see a docstring for `_get_state_dict_keys_to_exclude` but not for `state_dict`. Can you make sure both are correct?

nemo/collections/tts/modules/magpietts_modules.py
blisc: I'm looking at these two functions and they look identical, apart from one being removed in the 2nd dimension and one being removed in the last dimension. Do we really need both of these?

paarthneekhara: I believe @rlangman made these utilities. To me it makes sense - It avoids slicing the wrong dimension accidentally (if we were to not use any utility functions) and makes things explicit for embeddings vs codes (Since T is in a different dimension for both). If we use the same function, we'll end up with if-else statements  or some other conditional handling for different dimensions, which doesn't seem like a good idea either.

nemo/collections/tts/models/magpietts.py
paarthneekhara: Added docstrings in state_dict also. This change was done to exclude those modules from both state dict and optimizer param groups thats why there is a new _get_state_dict_keys_to_exclude. In EasyMagpie, we realized that excluding these modules from optimizer param groups saves memory.

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +805,133 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+    # Convert text to lowercase
+    lower_case_text = input_text.lower()
+
+    # Remove commas from text
+    no_comma_text = lower_case_text.replace(",", "")
+    # Replace "-" with spaces
+    no_dash_text = no_comma_text.replace("-", " ")
+    no_dash_text = no_dash_text.replace("'", "")
+    no_dash_text = no_dash_text.replace(";", "")
+    no_dash_text = no_dash_text.replace(".", "")
```
rfejgin: `no_dash_text` becomes a bit of a misnomer at this point. Maybe we can just reuse one name for the variable that keeps the normalized text, and keep updating it. Might be less error prone too.

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +805,133 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+
+    # Replace double spaces with single space
+    single_space_text = " ".join(no_dash_text.split())
+
```
rfejgin: Could you add a comment on what the next line does?

nemo/collections/tts/models/magpietts.py
```diff
@@ -489,12 +487,18 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
         self._codec_model.freeze()  # Lightning does requires_grad = False and self.eval()
         self._codec_converter = codec_converter
+        self._codec_helper = CodecHelper(self._codec_model, self._codec_converter)
 
         audio_embeddings = []
         for _ in range(self.num_audio_codebooks * self.frame_stacking_factor):
             audio_embeddings.append(nn.Embedding(self.num_all_tokens_per_codebook, cfg.embedding_dim))
         self.audio_embeddings = nn.ModuleList(audio_embeddings)
 
+        # Identity projections required by LocalTransformerHelper methods.
```
rfejgin, line 497: Why are these two identity projections needed? Are they non-identity in the decoder-only model?

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +805,133 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+    if isinstance(language, str) or language is None:
+        languages = [language] * len(audio_filepaths)
+    else:
+        if len(language) != len(audio_filepaths):
+            raise ValueError(
+                f"Expected len(language) == len(audio_filepaths), but got {len(language)} and {len(audio_filepaths)}."
+            )
+        languages = list(language)
+
+    grouped_indices = defaultdict(list)
```
rfejgin, line 877: Does the grouping end up reordering the order of the transcripts compared to the ordering coming in? Might it matter to calling functions?

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +805,133 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+            batch_transcripts = whisper_processor.batch_decode(predicted_ids, skip_special_tokens=True)
+            if normalizer is not None:
+                batch_transcripts = [normalizer.normalize(text) for text in batch_transcripts]
+            for idx, text in zip(batch_indices, batch_transcripts):
+                transcripts[idx] = text
+
+    return transcripts
+
+
+def get_speaker_embeddings_from_filepaths(filepaths, speaker_verification_model, device):
```
rfejgin, line 904: In `nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py` we have a function `extract_embedding()` - consider combining

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -251,3 +256,536 @@ def forward(self, subword_ids: Tensor, subword_mask: Tensor | None = None) -> Te
...
+
+# ---------------------------------------------------------------------------
+# LocalTransformerHelper
+# ---------------------------------------------------------------------------
+
+
+class LocalTransformerHelper:
+    """Orchestrates local-transformer forward passes and sampling.
+
+    This is a plain Python class (not ``nn.Module``) that holds *references*
```
rfejgin, line 414: when saving a MagpieTTSModel checkpoint, does the local transformer get saved, even though this class is not a module? Not sure what happens, just something to confirm

paarthneekhara: Yes, that should work the same way since local_transformer is still a module within MagpieTTS. The module gets passed into this helper.

nemo/collections/tts/parts/utils/helpers.py
paarthneekhara: It looks fine to me, we save the original batch indices in the grouped_indices for each language; and then appropriately assign the transcripts list .

nemo/collections/tts/parts/utils/helpers.py
paarthneekhara: Hmm. Keeping it separate for now - this one is batched, and has resampling going on. The other one is supporting both WavLM and titanet.

nemo/collections/tts/models/magpietts.py
paarthneekhara: Yes, they can be non-identity in the decoder-only model, if we use a different dimension.

nemo/collections/tts/parts/utils/helpers.py
rfejgin: They both do resampling I think (in the other function it's done through the target SR passed to `librosa.load())`. There's also the padding short signals to avoid crashing the speaker embedding model which we handle in extract_embedding which might be useful here too. But then the interaction with the embedding model itself looks different, this function uses the sv model's `forward()` whereas `extract_audio` uses `get_embedding` (not sure what the differences are). So the common part is the load-resample-pad. Up to you if you want to combine parts or not.

nemo/collections/tts/modules/magpietts_modules.py
rfejgin: sounds good

nemo/collections/tts/parts/utils/helpers.py
rfejgin: Fixed in latest commit

nemo/collections/tts/parts/utils/helpers.py
rfejgin: fixed in latest commit

rfejgin (dismissed):
Comments have been addressed, see remaining optional comment (on speaker embedding function).

nemo/collections/tts/modules/magpietts_modules.py
rlangman: The two differences are (1) the BOS being on the 2nd vs 3rd dimension and (2) the token version has frame stacking with multiple BOS tokens while the embedding has no frame stacking.

You can theoretically replace `remove_embedded_bos_token` with `remove_bos_token` if you reshape the input and output dimensions in the method that calls it. Both approaches would be fine, I just chose this one because it is a little more succinct and makes it easier to see the differences in how we represent the token data versus the equivalent embedded representation.

blisc: Ok, let's keep it as is

nemo/collections/tts/modules/magpietts_modules.py
```diff
@@ -251,3 +256,526 @@ def forward(self, subword_ids: Tensor, subword_mask: Tensor | None = None) -> Te
...
+            return audio, audio_len, codes
+
+
+class LocalTransformerHelper:
+    """Orchestrates local-transformer forward passes and sampling.
+
+    This is a plain Python class (not ``nn.Module``) that holds *references*
+    to nn.Module sub-modules owned by the parent model.  Keeping it non-Module
+    preserves checkpoint key compatibility.
+
```
rlangman, line 417: I am not opposed to trying this approach of deduping code since it is easily reversible, but its not too clear what the benefit is.

Deduping is mainly helpful if you want modifications to this function to automatically back-propagate to the current Magpie model class. If our intention is to put the previous iteration of Magpie into maintenance mode, then it could be better for the new version to use a different implementation of LT in order to make backwards compatibility easier. This is irrelevant if we have no intention of ever cleaning up this logic.

Usually if we refactor something to be reusable, it should be a generic implementation that can be reused in any system. For local transformer, this would mean creating something like a `nn.Module` with a generic interface that does not rely on knowing implementation details that only make sense for this specific model implementation (e.g. the concept of audio_embeddings table, frame stacking, special EOS/BOS ID tokens, have nothing to do with local transformer). So if we wanted to reuse local transformer in something else like in our codec or S2S models, this would not help us.

nemo/collections/tts/parts/utils/helpers.py
```diff
@@ -802,3 +805,122 @@ def g2p_backward_compatible_support(g2p_target: str) -> str:
...
+    text = input_text.lower()
+    for char in [",", "'", ";", "."]:
+        text = text.replace(char, "")
+    text = text.replace("-", " ")
+    text = " ".join(text.split())
+    # Strip any remaining punctuation characters
+    text = text.translate(str.maketrans('', '', string.punctuation))
+    # Fix common ASR transcript artifacts
+    text = text.replace("h t t p", "http")
+    text = text.replace("w w w", "www")
```
rlangman, line 821: I was considering doing this for common normalization problems like "mr" and "mrs" (without periods) in English, but I thought it would make more sense to require these text fixes outside of the CER function. Would we want to add that here?

blisc: If we want to correct for ASR transcription errors, it would be useful to make this helper function. Not all scripts should use this helper function.

nemo/collections/tts/modules/magpietts_modules.py
blisc: > For local transformer, this would mean creating something like a nn.Module

It is a fair point that we should consider making this a nn.Module. Let's make a point of trying to update this in another PR

> (e.g. the concept of audio_embeddings table, frame stacking, special EOS/BOS ID tokens, have nothing to do with local transformer).

On this point, I would disagree. The main point of this module is to transform continuous embedding spaces into a stack of audio tokens. I'm not sure if it's worth trying to make it more generic than that. At that point, you might as well just instantiate a transformer module.

nemo/collections/tts/models/magpietts.py
```diff
@@ -2303,24 +1773,9 @@ def _prepare_decoder_context(
...
-                batch_size = context_input_embedded.size(0)
-                if (
-                    self.training
-                    and batch_size > 1
-                    and self.train_shuffle_context_embedding_prob > 0
-                    and random.random() < self.train_shuffle_context_embedding_prob
-                ):
-                    shift = random.randint(1, batch_size - 1)
-                    context_embeddings = context_input_embedded.roll(shift, dims=0)
-                    context_mask = context_mask.roll(shift, dims=0)
```
XuesongYang, line 2319: @paarthneekhara not sure if this is removed on purpose. Did you try alternative solution to bypass context encoder instead?

PR #15524, opened 2026-03-19
[TTS] Add code for training semantic codec

rlangman:
# What does this PR do ?

Add code needed to train a single codebook semantic token and embed it inside a multi-codebook codec.

**Collection**: [TTS]

# Changelog

- Add semantic distillation using w2v-bert
- Add inference logic to embed semantic token as first codebook in multi-codebook codec
- Add option in data loader to resample audio
- Remove dead commented out code related to ASR loss. The `PhonemeASR` module it references is not defined in NeMo.

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/tests/functional_tests/L2_TTS_Fast_dev_runs_AcousticCodec.sh
+++ b/tests/functional_tests/L2_TTS_Fast_dev_runs_AcousticCodec.sh
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/audio_codec.py \
+HF_HUB_OFFLINE=1 TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/audio_codec.py \
     --config-name acoustic_codec_16000.yaml \
     semantic_codec_path="/home/TestData/tts/TestSemanticCodec.nemo" \
     +train_ds_meta.an4.manifest_path="/home/TestData/an4_dataset/an4_train_context_v1.json" \
--- a/tests/functional_tests/L2_TTS_Fast_dev_runs_AudioCodec.sh
+++ b/tests/functional_tests/L2_TTS_Fast_dev_runs_AudioCodec.sh
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/audio_codec.py \
+HF_HUB_OFFLINE=1 TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1 coverage run -a --data-file=/workspace/.coverage --source=/workspace/nemo examples/tts/audio_codec.py \
     --config-name audio_codec_22050.yaml \
     +train_ds_meta.an4.manifest_path="/home/TestData/an4_dataset/an4_train_context_v1.json" \
     +train_ds_meta.an4.audio_dir="/" \


nemo/collections/tts/modules/audio_codec_modules.py
```diff
@@ -177,6 +177,157 @@ def forward(self, audio_real, audio_gen):
...
+    """Encoder wrapping a speech language model (SLM) which produces semantic embeddings for use in semantic distillation.
+
+    Args:
+        slm_model_name: Name of Hugging Face model.
+        slm_sr: Sample rate SLM model requires for input.
+        input_sr: Sampling rate of audio that will be input to this encoder.
+        hidden_layer: Index of hidden layer to extract embeddings from.
+            Defaults to 16, which for research suggests is effective for w2v-bert and TTS.
+        padding: Number of audio samples to pad before encoding to ensure output has a frame rate compatible with the audio codec.
+        scaling_factor: Constant factor to scale output embedding by.
```
rfejgin: It looks like in practice we divide by this factor, not multiply. Maybe either change the contract to provide 1/scaling and then multiply, or change the comment to say we divide by it

rfejgin: Also, why is this scaling needed? Okay to keep it, just curious.

nemo/collections/tts/modules/audio_codec_modules.py
```diff
@@ -177,6 +177,157 @@ def forward(self, audio_real, audio_gen):
...
+        with torch.no_grad():
+            out = self.semantic_model(feats)
+            slm_emb = out.hidden_states[self.hidden_layer] / self.scaling_factor
+
+        slm_emb = rearrange(slm_emb, 'B T D -> B D T')
+
+        return slm_emb
+
+
+class SLMDecoder(NeuralModule):
```
rfejgin: I'm not sure if SLMDecoder is the best name for what this class does; on the face of it, it would look like it's decoding the outputs of the SLM encoder, but that's not what really does. Maybe SLM Predictor? Or any name that you think makes sense.

That said, I hope changing this won't cause too much trouble in invalidating existing checkpoints etc...

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -135,6 +144,7 @@ def __init__(
         self,
         dataset_meta: Dict,
         sample_rate: int,
+        resample_rate: Optional[int] = None,
```
rfejgin, line 148: could you update the docstring to add this argument?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -237,16 +266,21 @@ def state_dict(self, destination=None, prefix='', keep_vars=False):
...
-            if "discriminator" in key and ".slm_model.ssl_model." in key:
+            if "discriminator" in key and ".slm_model.slm_model." in key:
+                del state_dict[key]
+            if key.startswith("slm_encoder."):
                 del state_dict[key]
+
         return state_dict
 
     def load_state_dict(self, state_dict, strict=True):
         # Override to load all the keys except .speaker_encoder. and WavLM model
```
rfejgin: Could you update the comment to say why we are skipping some keys?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -110,7 +111,43 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
+        else:
+            self.discriminator = None
+
+        # If 'semantic_codec_path' is provided, the semantic codec will be initialized from the provided path.
+        # It will then be registered as a submodule and automatically loaded from the 'semantic_codec' field
+        if cfg.get("semantic_codec"):
+            semantic_codec_cfg = cfg.get("semantic_codec")
+            semantic_codec = AudioCodecModel(cfg=semantic_codec_cfg)
+        elif cfg.get("semantic_codec_path"):
+            semantic_codec_path = cfg.get("semantic_codec_path")
```
rfejgin, line 125: are `semantic_codec` and `semantic_codec_path` mutually excludive? If so, maybe we can add an error check.

nemo/collections/tts/models/audio_codec.py
```diff
@@ -831,20 +871,26 @@ def configure_optimizers(self):
...
-            sch1, sch2 = schedulers
-            sch1.step()
-            sch2.step()
+        if schedulers is None or self.lr_schedule_interval != interval:
+            return
+
+        if self.discriminator is None:
+            schedulers.step()
+        else:
+            schedulers[0].step()
```
rfejgin: maybe it would be cleaner to iterate on all schedulers in the list

nemo/collections/tts/models/audio_codec.py
```diff
@@ -284,8 +318,19 @@ def encode_audio(
...
+        if not sample_rate:
+            sample_rate = self.sample_rate
+
+        audio_preprocessed, audio_preprocessed_len = self.preprocess_audio(
+            audio=audio, audio_len=audio_len, sample_rate=sample_rate
+        )
+        encoded, encoded_len = self.audio_encoder(audio=audio_preprocessed, audio_len=audio_preprocessed_len)
+
+        if self.semantic_codec is not None:
+            semantic, _ = self.semantic_codec.encode_audio(audio=audio, audio_len=audio_len, sample_rate=sample_rate)
```
rfejgin: should we add `with torch.no_grad()`?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -831,20 +871,26 @@ def configure_optimizers(self):
...
     def update_lr(self, interval="step"):
         schedulers = self.lr_schedulers()
-        if schedulers is not None and self.lr_schedule_interval == interval:
-            sch1, sch2 = schedulers
-            sch1.step()
-            sch2.step()
+        if schedulers is None or self.lr_schedule_interval != interval:
+            return
+
+        if self.discriminator is None:
```
rfejgin: maybe instead just check if it's a list or single item, then this function doesn't need to know about discriminators

nemo/collections/tts/models/audio_codec.py
rfejgin: Also, a question: how does training of the semantic codec itself happen?

nemo/collections/tts/modules/audio_codec_modules.py
rlangman: All loss functions in the codec training are are implemented to produce approximately the same scale by default (around 0.1 - 0.3). The w2v embedding scale is random based on watever scale the layer norm in the model learned. It ends up producing an embedding where the max value is around 5, so I scale the embedding down to have a max value of about 1, which also reduces the scale of the SLM loss to about 0.2 to be comparable to the other losses.

The alternative would be to have the SLM loss scale in the AudioCodecModel class default to something like 0.1 or 0.2.

nemo/collections/tts/modules/audio_codec_modules.py
rlangman: I think the terms "predict" and "decode" are usually interchangeable. It is however a bit confusing in this instance because the target being generated by the decoder happens to be the latent space of a different encoder's output.

This can be safely changed, as I deleted the SLM encoder and decoders from the checkpoints I am using for inference anyways.

nemo/collections/tts/data/vocoder_dataset.py
rlangman: Added docstring. The functionality might be a bit confusing, because the feature that is actually being added is the option to resample using batched NeMo code instead of librosa.

nemo/collections/tts/models/audio_codec.py
rlangman: I tried to describe it succinctly in the comment. But what happens is that the first time you train this model there is no "semantic_codec" config, and it reads it from the "semantic_codec_path" you provide in your yaml file. It then loads the checkpoint, and stores it in a new config called "semantic_codec". On all future training runs, or during inference, both config values are present, but it prioritizes using the submodule instead of reading the checkpoint again. The "semantic_codec" is only auto-generated in this way, never defined by a user.

This was the only way I could find to get `register_nemo_submodule` to work in this way, and it feels like an awkward interface. If anyone knows a cleaner way to implement this, I would be happy to hear it.

rlangman: > Also, a question: how does training of the semantic codec itself happen?

You run the recipe using a config file that has 1 codebook, no discriminator, and the SLM loss enabled.

nemo/collections/tts/data/vocoder_dataset.py
```diff
@@ -117,7 +126,8 @@ class VocoderDataset(Dataset):
     Args:
         dataset_meta: Dict of dataset names (string) to dataset metadata.
         sample_rate: Sample rate to load audio as. If the audio is stored at a different sample rate, then it will
-            be resampled.
+            be resampled using librosa.
```
blisc, line 129: Why add the with librosa? I don't think you do.

nemo/collections/tts/models/audio_codec.py
blisc: What is this `register_nemo_submodule` function?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -110,7 +111,43 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
...
         # Discriminator setup
-        self.discriminator = instantiate(cfg.discriminator)
+        if cfg.get("discriminator"):
+            self.discriminator = instantiate(cfg.discriminator)
+        else:
+            self.discriminator = None
+
+        # If 'semantic_codec_path' is provided, the semantic codec will be initialized from the provided path.
+        # It will then be registered as a submodule and automatically loaded from the 'semantic_codec' field
+        if cfg.get("semantic_codec"):
```
blisc, line 121: Can you add a test path that includes the use of a semantic codec? I'm concerned that this is no L0 nor L2 test coverage of this pattern

nemo/collections/tts/data/vocoder_dataset.py
rlangman: This dataset class reads audio using `sample_audio` or `load_audio`, both of which eventually resample at this line of code with librosa https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/preprocessing/segment.py#L200.

The reason I added `resample_rate` was to create a dataloading path using our torch resampling (formerly torchaudio), since our most recent codecs do not work properly on audio downsampled with librosa.

nemo/collections/tts/models/audio_codec.py
rlangman: In Magpie, we have a `codecmodel_path` which is a local path to a `.nemo` checkpoint. Magpie checkpoints cannot be loaded until this path is manually overwritten by the user. `register_nemo_submodule` is how this can be avoided, so that the sub-codec is loaded automatically without needing the original `.nemo` path.

https://github.com/NVIDIA-NeMo/NeMo/blame/main/nemo/core/classes/modelPT.py#L315
https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/core/classes/modelPT.py#L275-L279

nemo/collections/tts/models/audio_codec.py
rlangman: We currently have no L0 or L2 tests for codec models in general, which makes it difficult to add one for just the semantic codec. Creating these tests looks fairly involved. Should I create a separate PR to add automated tests for codec model training and inference?

blisc: I'm not as concerned about the other codecs as the inference paths are covered during the magpie tests. For this specific training setup where we load a semantic codec and train a complimentary acoustic codec, we should add a fastdevrun using an interim checkpoint that you have.

rlangman: Added L0 and L2 tests. Will wait and verify that they run successfully in the CI pipeline.

blisc (approved):

PR #15558, opened 2026-03-27
VoiceChat EA STT training reproducible features

ankitapasad:
# What does this PR do ?

Adds following features to the dataset class to support VoiceChat EA STT training and fine-tuning
1. Correct agent EOS placement
2. Clean implementation of token IDs and update user BOS ID to match EA
3. MCQ system prompt, disabled by default
4. Filler responses for ASR training data, disabled by default 
5. Number normalization, enabled by default
6. Corresponding tests

Internal-access [document](https://docs.google.com/document/d/1CAmH8wcZAVFApuq31YOof4nPuFtawvsltvWfpRFhCHk/edit?usp=sharing) with training, inference recipes, and notes on parity.

**Collection**: speechlm2

# Usage

```python
```

**PR Type**:

- New Feature
- Bugfix

### DIFF
--- a/nemo/collections/speechlm2/data/datamodule.py
+++ b/nemo/collections/speechlm2/data/datamodule.py
@@ -53,12 +53,23 @@ class DataModule(LightningDataModule):
     Args:
         cfg: a DictConfig instance, typically corresponding to `data` namespace in YAML configs.
         tokenizer: a tokenizer instance, typically NeMo's AutoTokenizer wrapping HF's AutoTokenizer.
-        dataset: a torch.utils.data.Dataset instance, expected to define __getitem__ that accepts
+        train_dataset: a torch.utils.data.Dataset instance, expected to define __getitem__ that accepts
             a lhotse.CutSet. It converts metadata + raw data to a batch of PyTorch tensors.
             The data sampling is controlled by Lhotse samplers rather than the dataset.
+            Used for train_dataloader (is_training=True).
+        val_dataset: same contract as train_dataset, but for validation/test (is_training=False).
+            If None, falls back to train_dataset.
+        dataset: deprecated alias for train_dataset. Kept for backward compatibility.
     """
 
-    def __init__(self, cfg, tokenizer: TokenizerSpec, dataset: torch.utils.data.Dataset) -> None:
+    def __init__(
+        self,
+        cfg,
+        tokenizer: TokenizerSpec,
+        train_dataset: torch.utils.data.Dataset = None,
+        val_dataset: torch.utils.data.Dataset = None,
+        dataset: torch.utils.data.Dataset = None,
+    ) -> None:
         super().__init__()
         self.cfg = cfg
         with open_dict(self.cfg):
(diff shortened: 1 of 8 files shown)


nemo/collections/speechlm2/data/duplex_stt_dataset.py
```diff
@@ -111,12 +126,81 @@ def __init__(
 
         assert tokenizer.bos is not None, "BOS support in the tokenizer is required."
         assert tokenizer.eos is not None, "EOS support in the tokenizer is required."
+        
+        user_bos_token = '^'
```
kevinhu-nv, line 130: I use the same bos and eos for user and agent channels. I feel that is cleaner and I verified that does not impact model performance. I see you want to make exactly match EA, let's make these as configurable and one can set ^ and $? When we release the EA ckpt, we will release a config anyway to make it use ^ and $.

nemo/collections/speechlm2/data/duplex_stt_dataset.py
```diff
@@ -525,23 +634,37 @@ def collate_token_channel(
...
+    add_mcq_prompt: bool = False,
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    """Collate system prompts from cuts stored in cut.custom['system_prompt']."""
-    pad_id = get_pad_id(tokenizer)
+    """Collate system prompts from cuts.
+
+    Prompt selection priority:
+      1. Per-cut custom prompt (cut.custom['system_prompt'])
+      2. MCQ training cut -> THINK prompt for think-cuts, NOTHINK prompt for others
+      3. MCQ validation cut (when add_mcq_prompt=True) -> THINK prompt
```
kevinhu-nv, line 647: Can you also add a support for custom prompt? We can then easily evaluate different demo setups we have used.

kevinhu-nv:
A high-level question: Can you also share a training script/wandb to make sure metrics look roughly good? I think additional efforts may be needed to catch the EA ckpt but it is better to check at intermediate steps as well.

nemo/collections/speechlm2/data/datamodule.py
```diff
@@ -53,12 +53,23 @@ class DataModule(LightningDataModule):
...
     """
 
-    def __init__(self, cfg, tokenizer: TokenizerSpec, dataset: torch.utils.data.Dataset) -> None:
+    def __init__(
+        self,
+        cfg,
+        tokenizer: TokenizerSpec,
+        train_dataset: torch.utils.data.Dataset = None,
+        val_dataset: torch.utils.data.Dataset = None,
+        dataset: torch.utils.data.Dataset = None,
```
pzelasko, line 71: That's too many datasets. It's OK to remove `dataset` parameter and property, and update the code across collection to use `train_dataset` instead.

nemo/collections/speechlm2/data/utils.py
```diff
@@ -11,9 +11,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
pzelasko, line 11: Move the changes in this file to a new file `nemo/collections/speechlm2/data/mcq.py` and document what is the purpose and the top-level entry-point, with expected usage. Let's make this re-usable across models/projects.

nemo/collections/speechlm2/data/duplex_stt_dataset.py
```diff
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import json
```
pzelasko, line 14: I skipped the review of this file - I don't understand the logic being added here from a quick skim. 

Something for your consideration: this file has 1k+ lines of complex data preparation logic that isn't well documented. Users trying to train or finetune VoiceChat may have a hard time understanding how to prepare data and which options to use. Try to build this documentation, with examples showing what's the expected input and output for each of these steps (and the entire pipeline).

PR #15561, opened 2026-03-29
fix: clamp shift perturbation for short audio instead of silently skipping

stanley1208:
# What does this PR do?

Fix a bug in `ShiftPerturbation.perturb()` where audio augmentation was silently skipped when the randomly sampled shift exceeded the audio duration. This caused short audio clips to receive no shift augmentation at all, with no warning to the user.

**Before**: If `abs(shift_ms) / 1000 > data.duration`, the method returned immediately — a silent no-op. Users with short utterances or large `max_shift_ms` had their augmentation silently disabled.

**After**: The shift is clamped to the audio duration so some shift is always applied. This resolves the TODO at line 442: `"do something smarter than just ignore this condition"`.

**Collection**: ASR

# Changelog

- Fix `ShiftPerturbation.perturb()` to clamp shift amount instead of silently skipping augmentation for short audio
- Remove resolved TODO comment
- Add 7 unit tests for `ShiftPerturbation` covering normal operation, short audio clamping, length preservation, and zero-shift edge case

# Usage

```python
from nemo.collections.asr.parts.preprocessing.perturb import ShiftPerturbation

# Previously: short audio (5ms) with max_shift=20ms would silently skip augmentation
# Now: shift is clamped to audio duration and applied
perturb = ShiftPerturbation(min_shift_ms=10.0, max_shift_ms=20.0)
perturb.perturb(short_audio_segment)  # now always applies shift


stanley1208:
@nithinraok  @pzelasko  Ready for review — fixes a real bug where ShiftPerturbation silently skipped augmentation for short audio clips. Resolves the TODO at perturb.py line 442. Includes 7 unit tests. Thanks!

pzelasko (dismissed):
Thanks! Looks good to me.

stanley1208:
@pzelasko @nithinraok Friendly ping , this was approved but hasn't been merged yet. Is there anything else needed? Happy to address any feedback. Thanks!

pzelasko:
Can you fix the linters? Otherwise the CI with tests won't trigger

blisc (approved):

blisc:
Thanks for the contribution!

PR #15608, opened 2026-04-15
Update text processing in Magpietts eval

rlangman:
# What does this PR do ?

Adds an interface for doing language-specific text processing for TTS.

**Collection**: [TTS]

# Changelog

- Adds a interface TextProcessor which can be used fornormalization of training/eval data, and for processing text for WER calculation.
- Adds English implementation of TextProcessor to Magpie evaluation code. This reduces false positives in TTS WER caused by ASR models producing text in written (instead of spoken) form.
- Updated punctuation removal to use `isalnum()`. Previous logic using `string.punctuation` only removes a limited set of English punctuation.
- Added `datasets_base_path` to allow doing inference and evaluation using a config file with relative paths.
- Modified Longform CI to use a manifest with a "normalized_text" field. Otherwise, tests had higher WER because it was feeding Magpie raw text with several digits in its input.

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -769,3 +772,93 @@ def resample_batch(audio, audio_len, input_sample_rate, output_sample_rate):
...
+        pass
+
+
+class DefaultTextProcessor(TextProcessor):
+    """Default text processing behavior, if language-specific processing is not yet implemented."""
+
+    def normalize_text(self, text):
+        return text
+
+    def process_text_for_wer(self, text):
```
rfejgin, line 820: Could you add a comment on what the function does? E.g. to clarify that this should only be used for WER comparisons, not for normalizing TTS input text (where e.g. punctuation is useful I think). Or maybe put that comment in the base class.
For this function a high level comment could be useful, like: "removes punctuation and trims whitespace to single space"

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -769,3 +772,93 @@ def resample_batch(audio, audio_len, input_sample_rate, output_sample_rate):
...
+        text = text.lower()
+        # Replace dash with a single space
+        text = text.replace("-", " ")
+        # Replace whitespace with a single space
+        text = re.sub(pattern=r"\s\s+", string=text, repl=" ")
+        # Remove all non-alphanumeric characters.
+        text = "".join([c for c in text if c == " " or c.isalnum()])
```
rfejgin, line 827: Turns out that indeed c.alnum() also works for non-Latin (i.e. it goes beyond just a-z and 0-9) alphabets but might make sense to add it to the comment to avoid confusion

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -769,3 +772,93 @@ def resample_batch(audio, audio_len, input_sample_rate, output_sample_rate):
...
+        text = self.normalize_text(text)
+        text = self.default_processor.process_text_for_wer(text)
+        return text
+
+
+def get_text_processor(language):
+    if language == "en":
+        return EnglishTextProcessor()
+    else:
+        return DefaultTextProcessor()
```
rfejgin, line 864: Consider adding a log message warning that no language-specific processor is available, so using a generic one.

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -769,3 +772,93 @@ def resample_batch(audio, audio_len, input_sample_rate, output_sample_rate):
     audio_len = audio_len.int()
     audio = mask_sequence_tensor(audio, audio_len)
     return audio, audio_len
+
+
+def normalize_text_by_pattern(text, pattern, replacement):
+    """Normalize input text using a regular expression.
+
+    This function will search for and replace any string matching the input 'pattern' surrounded by any punctuation.
```
rfejgin, line 780: Could you add an example?

rfejgin (commented):
Overall looks good to me. Left some minor comments.

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: Added docstring in base class

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: Added logging statement

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: Done

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: I updated the comment, and documented in the PR why this is better than the previous implementation based on `string.punctuation`.

rfejgin (commented):
All looks good. One question: have we tested the updated code at all on non-English?  At least to the extent that we'd expect WERs to not be higher than before the change, or even to be lower. One possible way to test that while removing variability due to the TTS model would be to run evaluation on a pre-populated directory of Magpie outputs (if the code allows doing that easily), once before the changes and once after.

nemo/collections/tts/parts/utils/tts_dataset_utils.py
```diff
@@ -769,3 +780,132 @@ def resample_batch(audio, audio_len, input_sample_rate, output_sample_rate):
...
+        text = normalize_text_by_pattern(text=text, pattern="[eE]tc\.?", replacement="et cetera")
+        return text
+
+    def process_text_for_wer(self, text: str) -> str:
+        text = self.normalize_text(text)
+        text = self.default_processor.process_text_for_wer(text)
+        return text
+
+
+def get_text_processor(language: str) -> TextProcessor:
```
blisc, line 906: This function and https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/tts/parts/utils/helpers.py#L810 needs to be de-duplicated before we merge. The functionalities of process_text_for_cer() should be merged in and GRPO should use this text processor.

blisc (dismissed):

nemo/collections/tts/parts/utils/tts_dataset_utils.py
rlangman: That method is used by 3 different recipes: Magpie GPRO, Easy Magpie, and Magpie GRPO, and the refactoring needed for each of those is complicated both to implement and test. It is too large and dangerous of a change to couple with this PR, which is relatively small and required for us to proceed with updating our evaluation datasets.

I could implement something and rely on CI tests to validate each recipe still runs, but that is not safe because it is changing the GRPO criteria itself.

rlangman:
> All looks good. One question: have we tested the updated code at all on non-English? At least to the extent that we'd expect WERs to not be higher than before the change, or even to be lower. One possible way to test that while removing variability due to the TTS model would be to run evaluation on a pre-populated directory of Magpie outputs (if the code allows doing that easily), once before the changes and once after.

Using the same Magpie outputs I ran evaluation on CML test (french, german, italian, spanish) and all metrics were identical, except CER on french dropped from 0.022 to 0.019 because the updated logic with `isalnum()` removes french quotation marks before CER calculation.

blisc (dismissed):

rfejgin (dismissed):

blisc (dismissed):

.github/workflows/cicd-main-speech.yml
```diff
@@ -55,9 +55,11 @@ jobs:
             timeout: 30
           - script: L0_Unit_Tests_GPU_TTS
```
blisc, line 56: You will collide with https://github.com/NVIDIA-NeMo/NeMo/pull/15651

blisc (approved):

PR #15567, opened 2026-03-31
Add hi-IN , Ko-KR and pt-BR IPA tokenizer support

quapham:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?
Extends the IPAG2p tokenizer to support Hindi (hi-IN) with English code-switching , Korean (ko-KR) and Brazilian Portuguese (pt-BR) locale.

**Collection**: [Note which collection this PR will affect]
tts, common

# Changelog

- Add hi-IN, ko-KR and pt-BR to SUPPORTED_LOCALES in ipa_lexicon.py
- Add INDIC_CHARS_ALL support to tokenizer_utils.py and IpaG2p (enables Indic script dict parsing)
- Extend IpaG2p._parse_phoneme_dict() to accept a list of dicts enabling multi-dict code-switching (e.g. Hindi + English)
- Add pronunciation dict files: hi_prondict-v0.1.dict (hi_IN) and pt_br_prondict-v1.0.dict (pt_BR), ko_prondict-v1.0.dict
- Add unit tests: test_ipa_tokenizer_hi_in (hi-IN/en code-switching), test_ipa_ko_kr and test_ipa_tokenizer_pt_br

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [x ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [x ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[...]

PR #15614, opened 2026-04-16
Add Arabic char tokenizer and Japanese-English katakana support

quapham:
> [!IMPORTANT]  

# What does this PR do ?

Add Arabic character tokenizer support and improve mixed Japanese-English katakana handling in Japanese G2P/tokenizer tests.

**Collection**: [Note which collection this PR will affect] TTS, common

# Changelog
- add Arabic character tokenizer coverage/support
- update Japanese katakana accent handling for mixed English/Japanese text
- add unit tests for tokenizer behavior

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
+++ b/nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
@@ -44,6 +44,15 @@ from nemo.collections.common.tokenizers.text_to_speech.tokenizer_utils import (
 from nemo.collections.common.tokenizers.tokenizer_spec import TokenizerSpec
 from nemo.utils import logging
 
+CASELESS_SCRIPT_TOKENIZER_TARGETS = frozenset(
+    {
+        'nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers.HindiCharsTokenizer',
+        'nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers.ArabicCharsTokenizer',
+    }
+)
+
+DEFAULT_CHARSET_VERSION = 2
+
 
 class BaseTokenizer(ABC):
     """Abstract class for creating an arbitrary tokenizer to convert string to list of int tokens.
@@ -164,9 +173,8 @@ class BaseCharsTokenizer(BaseTokenizer):
                 logging.warning(f"Text: [{text}] contains unknown char: [{c}]. Symbol will be skipped.")
 
         # Remove trailing spaces
-        if cs:
-            while cs[-1] == space:
-                cs.pop()
+        while cs and cs[-1] == space:
+            cs.pop()
 
         if self.pad_with_space:
             cs = [space] + cs + [space]
@@ -382,6 +390,14 @@ class ItalianCharsTokenizer(BaseCharsTokenizer):
 class HindiCharsTokenizer(BaseCharsTokenizer):
     """Hindi grapheme tokenizer (character-based, no phonemes).
     Args:
+        chars: Explicit character set string. When provided, ``charset_version`` is ignored.
+        charset_version: Controls which default character set to use (only when ``chars`` is None).
+            ``2`` (default) — ``case="upper"`` Devanagari + ``ascii_letters``.
+            Hindi/Devanagari has no case distinction, so ``case="upper"`` avoids duplicating
+            every code-point. ``ascii_letters`` covers both upper- and lower-case English for
+            mixed-language text.
+            ``1`` — legacy ``case="mixed"`` Devanagari + ``ascii_lowercase``. Use this value to
+            restore models that were trained before the charset fix.
         punct: Whether to reserve grapheme for basic punctuation or not.
         apostrophe: Whether to use apostrophe or not.
         add_blank_at: Add blank to labels in the specified order ("last") or after tokens (any non None),
@@ -406,13 +422,14 @@ class HindiCharsTokenizer(BaseCharsTokenizer):
 
     _LOCALE = "hi-IN"
     _PUNCT_LIST = get_ipa_punctuation_list(_LOCALE)
+    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="upper") + string.ascii_letters
     _PUNCT_LIST_V1 = sorted(list(DEFAULT_PUNCTUATION))
-    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="mixed")
-    _CHARSET_STR += string.ascii_lowercase
+    _CHARSET_STR_V1 = get_grapheme_character_set(locale=_LOCALE, case="mixed") + string.ascii_lowercase
 
     def __init__(
         self,
-        chars=_CHARSET_STR,
+        chars=None,
+        charset_version=2,
         punct=True,
         apostrophe=True,
         add_blank_at=None,
@@ -421,6 +438,22 @@ class HindiCharsTokenizer(BaseCharsTokenizer):
         punct_version=2,
         text_preprocessing_func=any_locale_text_preprocessing,
     ):
+        if chars is None:
+            if charset_version == 1:
+                warnings.warn(
+                    "HindiCharsTokenizer charset_version=1 (case='mixed' + ascii_lowercase) is deprecated "
+                    "and will be removed in a future release. "
+                    "Migrate to charset_version=2 (case='upper' + ascii_letters) and retrain.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                chars = self._CHARSET_STR_V1
+            elif charset_version == 2:
+                chars = self._CHARSET_STR
+            else:
+                raise ValueError(
+                    f"HindiCharsTokenizer: unsupported charset_version={charset_version!r}. Use 1 (legacy) or 2."
+                )
         if non_default_punct_list is None:
             if punct_version == 1:
                 warnings.warn(
@@ -467,9 +500,100 @@ class HindiCharsTokenizer(BaseCharsTokenizer):
                 logging.warning(f"Text: [{text}] contains unknown char: [{c}]. Symbol will be skipped.")
 
         # Remove trailing spaces
-        if cs:
-            while cs[-1] == space:
-                cs.pop()
+        while cs and cs[-1] == space:
+            cs.pop()
+
+        if self.pad_with_space:
+            cs = [space] + cs + [space]
+
+        return [self._token2id[p] for p in cs]
+
+
+class ArabicCharsTokenizer(BaseCharsTokenizer):
+    """Arabic grapheme tokenizer (character-based, no phonemes).
+    Args:
+        chars: Explicit character set string. When provided, ``charset_version`` is ignored.
+        charset_version: Controls which default character set to use (only when ``chars`` is None).
+            ``2`` (default) — ``case="upper"`` Arabic + ``ascii_letters``.
+            Arabic script has no case distinction, so ``case="upper"`` avoids duplicating
+            every code-point. ``ascii_letters`` covers both upper- and lower-case English for
+            mixed-language text.
+            ``1`` — legacy ``case="mixed"`` Arabic + ``ascii_letters``. Use this value to
+            restore models that were trained before the charset fix.
+        punct: Whether to reserve grapheme for basic punctuation or not.
+        apostrophe: Whether to use apostrophe or not.
+        add_blank_at: Add blank to labels in the specified order ("last") or after tokens (any non None),
+            if None then no blank in labels.
+        pad_with_space: Whether to pad text with spaces at the beginning and at the end or not.
+        non_default_punct_list: List of punctuation marks which will be used instead default.
+        text_preprocessing_func: Text preprocessing function. Keeps Arabic unchanged.
+
+        Each Unicode code point becomes 1 token (letters, diacritics, and Arabic punct from ipa_lexicon).
+        Supports both upper and lower English letters (e.g. mixed-language text).
+
+        Input Text: مرحبا Hello
+        Chars: ['م', 'ر', 'ح', 'ب', 'ا', ' ', 'H', 'e', 'l', 'l', 'o']
+    """
+
+    _LOCALE = "ar-MSA"
+    _PUNCT_LIST = get_ipa_punctuation_list(_LOCALE)
+    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="upper") + string.ascii_letters
+    _CHARSET_STR_V1 = get_grapheme_character_set(locale=_LOCALE, case="mixed") + string.ascii_letters
+
+    def __init__(
+        self,
+        chars=None,
+        charset_version=2,
+        punct=True,
+        apostrophe=True,
+        add_blank_at=None,
+        pad_with_space=False,
+        non_default_punct_list=_PUNCT_LIST,
+        text_preprocessing_func=any_locale_text_preprocessing,
+    ):
+        if chars is None:
+            if charset_version == 1:
+                warnings.warn(
+                    "ArabicCharsTokenizer charset_version=1 (case='mixed' + ascii_letters) is deprecated "
+                    "and will be removed in a future release. "
+                    "Migrate to charset_version=2 (case='upper' + ascii_letters) and retrain.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                chars = self._CHARSET_STR_V1
+            elif charset_version == 2:
+                chars = self._CHARSET_STR
+            else:
+                raise ValueError(
+                    f"ArabicCharsTokenizer: unsupported charset_version={charset_version!r}. Use 1 (legacy) or 2."
+                )
+        super().__init__(
+            chars=chars,
+            punct=punct,
+            apostrophe=apostrophe,
+            add_blank_at=add_blank_at,
+            pad_with_space=pad_with_space,
+            non_default_punct_list=non_default_punct_list,
+            text_preprocessing_func=text_preprocessing_func,
+        )
+
+    def encode(self, text):
+        """Encode Arabic text, handling diacritics and English (upper/lower) correctly."""
+        cs, space, tokens = [], self.tokens[self.space], set(self.tokens)
+
+        text = self.text_preprocessing_func(text)
+        for c in text:
+            if c == space and len(cs) > 0 and cs[-1] != space:
+                cs.append(c)
+            elif c in tokens and c != space:
+                cs.append(c)
+            elif (c in self.PUNCT_LIST) and self.punct:
+                cs.append(c)
+            elif c != space:
+                logging.warning(f"Text: [{text}] contains unknown char: [{c}]. Symbol will be skipped.")
+
+        while cs and cs[-1] == space:
+            cs.pop()
 
         if self.pad_with_space:
             cs = [space] + cs + [space]
--- a/nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
+++ b/nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
@@ -24,7 +24,12 @@ from lhotse.dataset.collation import collate_matrices, collate_vectors
 from omegaconf import DictConfig, open_dict
 from transformers import AutoTokenizer, T5Tokenizer
 
-from nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers import AggregatedTTSTokenizer, IPABPETokenizer
+from nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers import (
+    CASELESS_SCRIPT_TOKENIZER_TARGETS,
+    DEFAULT_CHARSET_VERSION,
+    AggregatedTTSTokenizer,
+    IPABPETokenizer,
+)
 from nemo.collections.tts.parts.utils.tts_dataset_utils import (
     beta_binomial_prior_distribution,
     normalize_volume,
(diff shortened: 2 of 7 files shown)


XuesongYang:
thanks for adding new tokenizers.

Regarding 4 arabic locales, we decided to use separate tokenizers for ar-SY, ar-AE, ar-MSA, and ar-SA. In this PR, I only found `ArabicCharsTokenizer` was defined. Could you pls add tokenizers below?
```
  ar-AE: [arabic_AE_chartokenizer]
  ar-MSA: [arabic_MSA_chartokenizer]
  ar-SA: [arabic_SA_chartokenizer]
  ar-SY: [arabic_SY_chartokenizer]
```

XuesongYang:
> thanks for adding new tokenizers.
> 
> Regarding 4 arabic locales, we decided to use separate tokenizers for ar-SY, ar-AE, ar-MSA, and ar-SA. In this PR, I only found `ArabicCharsTokenizer` was defined. Could you pls add tokenizers below?
> 
> ```
>   ar-AE: [arabic_AE_chartokenizer]
>   ar-MSA: [arabic_MSA_chartokenizer]
>   ar-SA: [arabic_SA_chartokenizer]
>   ar-SY: [arabic_SY_chartokenizer]
> ```

discussed offline. No need to add one tokenizer per locale.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -458,6 +458,74 @@ def encode(self, text):
...
+        Each Unicode code point becomes 1 token (letters, diacritics, and Arabic punct from ipa_lexicon).
+        Supports both upper and lower English letters (e.g. mixed-language text).
+
+        Input Text: مرحبا Hello
+        Chars: ['م', 'ر', 'ح', 'ب', 'ا', ' ', 'H', 'e', 'l', 'l', 'o']
+    """
+
+    _LOCALE = "ar-MSA"
+    _PUNCT_LIST = get_ipa_punctuation_list(_LOCALE)
+    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="mixed")
```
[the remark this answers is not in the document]
XuesongYang: I think Copilot identified correct issues. Arabic script has no case distinction, and the char set of `ar-MSA` doesn't have any alphabet letters, so specifying `case="mixed"` would duplicate char tokens. i would suggest specifying "upper" instead.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
XuesongYang, line 407: similar comment as `ArabicCharsTokenizer`: 
Hindi has no case distinction, and the char set of `hi-IN` doesn't have any alphabet letters, so specifying case="mixed" would duplicate char tokens. i would suggest specifying "upper" instead.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -458,6 +458,74 @@ def encode(self, text):
...
+    _CHARSET_STR += string.ascii_letters
```
XuesongYang, line 482: I compare the changes with Hindi since both languages written systems both have no case distinction.

Hindi uses `string.ascii_lowercase` (line 408), while Arabic uses `string.ascii_letters` (upper + lower). The docstring says this is for "mixed-language text" support, which is a valid use case, which means the Arabic tokenizer is `case-preserving` for English letters, while Hindi is `case-folding`. 

Are these differences intentional? Shall we also make Hindi use `string.ascii_letters` as well?

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
XuesongYang, line 408: Hindi uses `string.ascii_lowercase` (line 408), while Arabic uses `string.ascii_letters` (upper + lower). The docstring says this is for "mixed-language text" support, which is a valid use case, which means the Arabic tokenizer is `case-preserving` for English letters, while Hindi is `case-folding`.

Are these differences intentional? Shall we also make Hindi use string.ascii_letters as well?

tests/collections/common/tokenizers/text_to_speech/test_tts_tokenizers.py
```diff
@@ -287,22 +288,16 @@ def test_japanese_phoneme_tokenizer(self):
 
     @pytest.mark.run_only_on('CPU')
     @pytest.mark.unit
-    def test_japanese_katakana_accent_tokenizer(self):
-        input_chopsticks = "箸"
-        input_bridge = "橋"
```
XuesongYang, line 292: The old test `test_japanese_katakana_accent_tokenizer` validated a critical property: that **homonyms with different pitch accents produce distinct outputs (箸 vs 橋)**. The new test `test_japanese_mix_english_katakana_accent_tokenizer` replaces it entirely and only checks a single input/output pair.

Let's keep the old test alongside the new one. The pitch accent differentiation check is valuable and independent from the English-mixing feature:

```suggestion
    def test_japanese_katakana_accent_tokenizer(self):
        """Homonyms with different pitch accents should produce different outputs."""
        g2p = JapaneseKatakanaAccentG2p()
        tokenizer = JapanesePhonemeTokenizer(g2p=g2p, punct=False)
        chars_chopsticks, _ = self._parse_text(tokenizer, "箸")
        chars_bridge, _ = self._parse_text(tokenizer, "橋")
        assert chars_chopsticks != chars_bridge
```

nemo/collections/tts/g2p/models/ja_jp_ipa.py
```diff
@@ -298,28 +298,41 @@ def __call__(self, text: str) -> List[str]:
...
                     current_chain = []
                 if string.isspace():
                     result.append(' ')
                 elif string in punctuation:
                     result.append(string)
+                else:
+                    logging.warning(
+                        f"Unknown symbol '{string}' (pos={pos}) not in punctuation list, replacing with space. original text: {text}"
+                    )
+                    result.append(' ')
```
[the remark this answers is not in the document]
XuesongYang: I agree with copilot's comment. So there are three places needed to change by adding a guard,

```suggestion
                else:
                    logging.warning(
                        f"Unknown symbol '{string}' (pos={pos}) not in punctuation list, replacing with space. original text: {text}"
                    )
                    if not result or result[-1] != ' ':
                        result.append(' ')
```

nemo/collections/tts/g2p/models/ja_jp_ipa.py
```diff
@@ -298,28 +298,41 @@ def __call__(self, text: str) -> List[str]:
...
 
-            # Handle punctuation
-            if pos in ('記号', '補助記号'):
+            # Handle punctuation (記号), but keep alphabet symbols (アルファベット) as regular words
+            if pos in ('記号', '補助記号') and pos_group1 != 'アルファベット':
                 if current_chain:
                     self._process_chain(current_chain, result)
                     current_chain = []
                 if string.isspace():
                     result.append(' ')
```
XuesongYang, line 320: I agree with below copilot's comment. So here needed to change by adding a guard,

```suggestion
                    if not result or result[-1] != ' ':
                        result.append(' ')
```

nemo/collections/tts/g2p/models/ja_jp_ipa.py
```diff
@@ -298,28 +298,41 @@ def __call__(self, text: str) -> List[str]:
...
+                    )
+                    result.append(' ')
                 continue
 
             if not pron or mora_size == 0:
+                if string and not string.isspace():
+                    logging.warning(
+                        f"Unknown symbol '{string}' (pos={pos}) not in punctuation list, replacing with space. original text: {text}"
+                    )
+                    result.append(' ')
```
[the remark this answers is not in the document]
XuesongYang: I agree with copilot's comment. So there are three places needed to change by adding a guard,

```suggestion
                if string and not string.isspace():
                    logging.warning(
                        f"Unknown symbol '{string}' (pos={pos}) not in punctuation list, replacing with space. original text: {text}"
                    )
                    if not result or result[-1] != ' ':
                        result.append(' ')
```

tests/collections/common/tokenizers/text_to_speech/test_tts_tokenizers.py
```diff
@@ -314,3 +309,14 @@ def test_hindi_chars_tokenizer(self):
...
+    @pytest.mark.run_only_on('CPU')
+    @pytest.mark.unit
+    def test_arabic_chars_tokenizer(self):
+        input_text = "مرحبا بالعالم!"
+        expected_output = "مرحبا بالعالم!"
+
+        tokenizer = ArabicCharsTokenizer()
+        chars, tokens = self._parse_text(tokenizer, input_text)
+
+        assert chars == expected_output
```
XuesongYang, line 322: The test only checks a basic Arabic-only string. Can you add missing coverages below?
- Mixed Arabic-English text (e.g., `"مرحبا Hello"`) — this is a headline feature per the docstring.
- Arabic diacritics (e.g., `"بِسْمِ"` — text with harakat): better to cover 4 dialects that we are using.
- Arabic-specific punctuation (`،`, `؛`, `؟`)
- Unknown character handling (e.g. CJK mixed with Arabic)

XuesongYang (dismissed):
lgtm in genral, but need some changes before merging. @quapham pls have a look and take actions accordingly. 

i am not familiar either languages. But pls let me know if I am wrong. Thanks!

tests/collections/common/tokenizers/text_to_speech/test_tts_tokenizers.py
XuesongYang, line 322: here are comprehensive tests generated by Claude Code. pls have a look.
```suggestion
    @pytest.mark.run_only_on('CPU')
    @pytest.mark.unit
    def test_arabic_chars_tokenizer_mixed_english(self):
        input_text = "مرحبا Hello"
        expected_output = "مرحبا Hello"

        tokenizer = ArabicCharsTokenizer()
        chars, tokens = self._parse_text(tokenizer, input_text)

        assert chars == expected_output

    @pytest.mark.run_only_on('CPU')
    @pytest.mark.unit
    @pytest.mark.parametrize(
        "input_text, expected_output",
        [
            ("بِسْمِ", "بِسْمِ"),  # kasra + sukun
            ("كَتَبَ", "كَتَبَ"),  # fatha
            ("كُتُبٌ", "كُتُبٌ"),  # damma + tanwin
            ("شَدَّة", "شَدَّة"),  # shadda
        ],
        ids=["kasra_sukun", "fatha", "damma_tanwin", "shadda"],
    )
    def test_arabic_chars_tokenizer_diacritics(self, input_text, expected_output):
        tokenizer = ArabicCharsTokenizer()
        chars, tokens = self._parse_text(tokenizer, input_text)

        assert chars == expected_output

    @pytest.mark.run_only_on('CPU')
    @pytest.mark.unit
    def test_arabic_chars_tokenizer_punctuation(self):
        input_text = "مرحبا، كيف حالك؟"
        expected_output = "مرحبا، كيف حالك؟"

        tokenizer = ArabicCharsTokenizer()
        chars, tokens = self._parse_text(tokenizer, input_text)

        assert chars == expected_output

    @pytest.mark.run_only_on('CPU')
    @pytest.mark.unit
    def test_arabic_chars_tokenizer_unknown_token(self):
        input_text = "مرحبا 你好 عالم"
        expected_output = "مرحبا عالم"

        tokenizer = ArabicCharsTokenizer()
        chars, tokens = self._parse_text(tokenizer, input_text)

        assert chars == expected_output
```

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
quapham: Thanks. Hindi does not have case distinction, so case="mixed" is unnecessary here. I’ll update it to case="upper".

quapham: The difference was not intentional. Previously, we did not pay much attention to English code-switching use cases. I’ll change Hindi from string.ascii_lowercase to string.ascii_letters so mixed Hindi-English text can preserve uppercase English letters as well.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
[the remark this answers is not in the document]
quapham: Thanks. For Arabic, I’ll keep the current setup for now because changing it to upper would alter the tokenizer vocabulary/IDs and break checkpoint compatibility. I’ll clean that up in a future release.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -404,8 +404,8 @@ class HindiCharsTokenizer(BaseCharsTokenizer):
 
     _LOCALE = "hi-IN"
     _PUNCT_LIST = get_ipa_punctuation_list(_LOCALE)
-    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="mixed")
-    _CHARSET_STR += string.ascii_lowercase
+    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="upper")
+    _CHARSET_STR += string.ascii_letters
```
blisc: @quapham how do you plan to enable backwards compatibility on existing checkpoints? As you can see in the current tests, this changes breaks the current Feb checkpoint.

blisc (changes requested):

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
XuesongYang: good catch! we should add back-compatibility support now and add a deprecation warning of this for next release.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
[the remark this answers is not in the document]
XuesongYang: gotcha! let me add changes that back support existing checkpoints, and add a deprecation warning for future releases.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
XuesongYang: again, let me add changes that back support existing checkpoints, and add a deprecation warning for future releases.

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
blisc: It looks like we pass in `chars` as an argument to this function. Likely, we just need to hard-code the old list into the Feb checkpoint and it should work (update the model yaml to add in the correct, older char list as an argument to this constructor).

quapham: Thanks for catching this! I've reverted the HindiCharsTokenizer default charset back to case="mixed" + ascii_lowercase to maintain backward compatibility with the Feb checkpoint. I also added a deprecation warning in __init__ so users know the default will change to case="upper" + ascii_letters in the next release.

tests/collections/common/tokenizers/text_to_speech/test_tts_tokenizers.py
quapham: Added more Arabic tokenizer test coverage:
- Mixed Arabic-English text
- Arabic dialects + diacritics/harakat
- Arabic-specific punctuation (،, ؛, ؟) 
- Unknown character handling (CJK mixed with Arabic -- unknown chars are skipped)

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
XuesongYang: i don't think we should revert back to the charset ("mixed" + ascii_lowercase) for hindi, ("mixed" + ascii_letters) for arabic. Instead, we should fix it but add back-compatibility + deprecation warning. something like below,

```
    _CHARSET_STR = get_grapheme_character_set(locale=_LOCALE, case="upper") + string.ascii_letters
    _CHARSET_STR_V1 = get_grapheme_character_set(locale=_LOCALE, case="mixed") + string.ascii_lowercase

    def __init__(
        self,
        chars=None,
        charset_version=2,
        punct=True,
        apostrophe=True,
        add_blank_at=None,
        pad_with_space=False,
        non_default_punct_list=_PUNCT_LIST,
        text_preprocessing_func=any_locale_text_preprocessing,
    ):
        if chars is None:
            if charset_version == 1:
                warnings.warn(
                    "HindiCharsTokenizer charset_version=1 (case='mixed' + ascii_lowercase) is deprecated "
                    "and will be removed in a future release. "
                    "Migrate to charset_version=2 (case='upper' + ascii_letters) and retrain.",
                    DeprecationWarning,
                    stacklevel=2,
                )
                chars = self._CHARSET_STR_V1
            elif charset_version == 2:
                chars = self._CHARSET_STR
            else:
                raise ValueError(
                    f"HindiCharsTokenizer: unsupported charset_version={charset_version!r}. Use 1 (legacy) or 2."
                )
```

XuesongYang: I will add changes accordingly soon.

blisc: I don't agree with this design. We should have not a charset_version inside of the arguments to the tokenizer, the current argument of chars already handles this. If you want to make the changes backwards compatible silently, I would change the chars argument of older checkpoints instead of adding a new version argument and adding the version argument.

> I don't think we should revert back to the charset

+1, we should not revert the charset defaults if you already used them in 2605.

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -52,6 +57,18 @@ def setup_tokenizers(all_tokenizers_config, mode='train'):
...
+            # Persist charset_version so it's saved in .nemo archives and
+            # update_config_for_inference can distinguish old checkpoints
+            # (missing charset_version → v1) from new ones.
+            if (
+                hasattr(tokenizer_config, '_target_')
+                and tokenizer_config._target_ in CASELESS_SCRIPT_TOKENIZER_TARGETS
+                and not hasattr(tokenizer_config, 'charset_version')
+            ):
+                with open_dict(all_tokenizers_config):
+                    tokenizer_config.charset_version = DEFAULT_CHARSET_VERSION
```
blisc, line 92: Instead of persisting charset_version, just persist the chars argument instead.

1. Use `tokenizer_config._target_` to get the class
2. Then add tokenizer_config.chars  = cls._CHARSET_STR

nemo/collections/tts/modules/magpietts_inference/utils.py
```diff
@@ -149,6 +150,24 @@ def validate(self) -> None:
...
 
+def _migrate_charset_version(model_cfg: DictConfig) -> None:
+    """Pin charset_version=1 for Hindi/Arabic tokenizers in old checkpoints.
+
+    New models have ``charset_version`` persisted by ``setup_tokenizers()``.
+    Old checkpoints lack it, so without this migration the new default (v2)
+    would silently change the token-to-ID mapping and break the model.
+
+    Must be called inside ``open_dict(model_cfg)``.
+    """
```
blisc, line 161: Here, check to see if chars have been set by setup_tokenizers. If not, set the older chars set

blisc (dismissed):

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
blisc:
```suggestion
            from importlib import import_module
            mod = import_module("nemo.collections.common.tokenizers.text_to_speech.tts_tokenizers ")
            cls = getattr(mod, tokenizer_config._target_.split(".")[-1])
            tokenizer_config.chars = cls._CHARSET_STR
```

blisc: ^ Something like that, at a high level psuedo-code guide.

XuesongYang: thanks for bring it up. Actually I was considering whether to persist `chars` vs `charset_version`, or saying `an integer` vs `a long string`, when implementing back-compatibility support. And my final decision goes with `charset_version`. Here is why,

it is definitely human-readable if persisting `chars`, but it would add long unicode strings that make configs hard to read/diff. For example, Hindi `charset` string is hundreds of Unicode codepoints, which would bloat every config/log dump, compared against a single integer `charset_version`. We should be good with `charset_version` added in .nemo, and add clear docstring. 

but again, we need to avoid such scenarios to make design complicated in long run.

XuesongYang: a second thought, considering `charset_version` is still much better than `chars`. So we'd better keeping this implementation.

1. **Easy config/CLI override** — charset_version=1 is a single integer in YAML or Hydra CLI. The Hindi v1 charset is 172 Unicode characters including zero-width combining marks (्, ़, ँ); inserting that as a chars string into a config is awkward and error-prone.
2. **Concrete example** — this very PR shows the benefit. Fixing the CFG distillation test (`L2_TTS_Fast_dev_runs_Magpietts_OnlineCFGDistillation.sh`) was just adding one line: `+model.text_tokenizers.hindi_chartokenizer.charset_version=1`. With chars, that line would be a 172-char Unicode blob.
3. **`chars` still works as an escape hatch** — the if chars is None: guard means anyone who needs a truly custom charset can still pass chars directly and charset_version is ignored.

quapham: You’re right, keeping the config cleaner is better if we can. If the change already covers this safely, I’m happy to keep the config clean and rely on that instead.

XuesongYang:
@blisc let's pause merging this one now, and merge https://github.com/NVIDIA-NeMo/NeMo/pull/15567 first. I believe there are some merging conflicts to address.

XuesongYang:
> @blisc let's pause merging this one now, and merge #15567 first. I believe there are some merging conflicts to address.

all CI passed. @blisc pls review.

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
blisc: I'm ok to merge this as this. But in the future, we need to refactor this default version to be part of the class not a top level global in the file.

blisc (approved):

PR #15622, opened 2026-04-17
[audio codec] Add support for Lhotse training format

rfejgin:
This PR adds a Lhotse data loader for audio codec training. It also introduces mechanisms to make the training process more stable and to help debug potential `NaN` issues.

### Data Loading
The functionality is split between built-in Lhotse capabilities and a simple custom dataset class.

#### Lhotse
These operations are handled directly in Lhotse:
1. **Duration filtering:** Only keeps cuts whose duration is at least that of the training chunk size (`n_samples`).
2. **Random segment selection:** Selects a random segment of length `n_samples` from the cut.

Configuring Lhotse happens in `AudioCodecModel._get_lhotse_dataloader()`.

#### `AudioCodecLhotseDataset`
The custom dataset class, `AudioCodecLhotseDataset`, receives a `CutSet` from Lhotse and performs the following:
1. **Loads** and collates the audio.
2. **Resamples** to the target sample rate, which is the codec's `output_sample_rate`.
3. Performs **sanity checks** on the audio (details below).

### Training Robustness
When training models on Lhotse datasets, we observed convergence similar to the previous training recipe. However, in one instance (out of ~4 training runs), the training ran into `NaNs`. To help debug this issue if it ever reoccurs and to stabilize training, the following mechanisms were added:

1. **Sanity checks in the loader:** Errors out if the loaded audio has an unexpected length or contains `NaN` or infinite values. It also warns if suspicious sample values (`abs(sample) > 1.5`) are encountered.
2. **Gradient norm tracking** (operates separately for the discriminator and generator):
   * **(Optional) Skip updates:** If an infinite or `NaN` gradient norm is detected, it triggers a warning and skips the current parameter update.
   * **(Optional) Gradient clipping:** Applies gradient norm clipping.
   * **Logging:** Logs the gradient norm both before and after clipping.

### Additional notes
* I have confirmed that old YAML configs still run without error after these changes
* A companion PR is being prepared in the internal repo that correspondingly updates the training recipe.


nemo/collections/tts/models/audio_codec.py
```diff
@@ -64,12 +66,12 @@ def __init__(self, cfg: DictConfig, trainer: Trainer = None):
         if trainer is not None:
             self.world_size = trainer.num_nodes * trainer.num_devices
 
-        super().__init__(cfg=cfg, trainer=trainer)
-
         # Expected sample rate for input and output audio
         self.sample_rate = cfg.sample_rate
         self.output_sample_rate = cfg.get("output_sample_rate", self.sample_rate)
 
+        super().__init__(cfg=cfg, trainer=trainer)
```
rfejgin, line 76: This had to be moved because data loaders are created during parent class init, and their setup requires access to the sample rate.

nemo/collections/tts/models/audio_codec.py
```diff
@@ -787,36 +826,76 @@ def _setup_train_dataloader(self, cfg):
...
-                )
-            elif self._trainer is None:
-                logging.warning(
-                    "Model Trainer was not set before constructing the dataset, incorrect number of "
-                    "training batches will be used. Please set the trainer and rebuild the dataset."
-                )
+        if cfg.get("use_lhotse", False):
+            self._train_dl = self._get_lhotse_dataloader(cfg)
+        else:
+            # --- Legacy data loading (non-Lhotse): Webdataset and NeMo dataset formats. ---
```
rfejgin: No changes to legacy loading besides indentation

nemo/collections/tts/models/audio_codec.py
```diff
@@ -787,33 +826,76 @@ def _setup_train_dataloader(self, cfg):
...
+
+            self._train_dl = self._setup_train_dataloader(cfg)
```
rlangman: Would look cleaner if the second block were also moved into a function like:

```
if cfg.get("use_lhotse", False):
    self._train_dl = self._get_lhotse_dataloader(cfg)
else:
    self._train_dl = self._get_nemo_or_webdataset_dataloader(cfg)
```

Ideally it would be refactored like this (taking into account that the large IterableDataset if statement only applies to webdataset):

```
if cfg.get("use_lhotse", False):
    self._train_dl = self._get_lhotse_dataloader(cfg)
elif cfg.get("is_sharded", False):
    self._train_dl = self._get_webdataset_dataloader(cfg)
else:
    self._train_dl = self._get_nemo_dataloader(cfg)
```

nemo/collections/tts/models/audio_codec.py
```diff
@@ -787,33 +826,76 @@ def _setup_train_dataloader(self, cfg):
...
+                        "Model Trainer was not set before constructing the dataset, incorrect number of "
+                        "training batches will be used. Please set the trainer and rebuild the dataset."
+                    )
 
     def setup_validation_data(self, cfg):
-        self._validation_dl = self._setup_test_dataloader(cfg)
+        if cfg.get("use_lhotse", False):
+            raise ValueError("Lhotse data loading is not supported yet for validation.")
+        else:
+            # For validation, we still use legacy data loading (Webdataset and NeMo dataset formats).
```
rlangman: Comment is slightly off, since Webdataset can never be used for validation. It would be easier to see from the existing code if `_setup_test_dataloader()` explicitly set `is_sharded=False`.

nemo/collections/tts/data/audio_codec_dataset_lhotse.py
```diff
@@ -0,0 +1,121 @@
...
+            cut.target_audio = cut.target_audio.resample(self.sample_rate)
+
+        # Load and collate the audio, applying any transformations that were
+        # configured in Lhotse in the process.
+        # Note: fault_tolerant=False for now to avoid masking errors until we are more
+        # confident in the new loader.
+        batch_audio, batch_audio_len = collate_audio(cuts, recording_field="target_audio", fault_tolerant=False)
+
+        # Sanity checks on the audio and its length
+        self.sanity_check_audio(batch_audio, batch_audio_len, cuts)
```
rlangman: I don't think we should have long debugging statements as code that always runs.

Reading online it sounds like the most correct way to do this is have the sanity check in an `assert` so that it will run by default but be ignored if someone runs their code with the Python -O (optimize flag).

Alternatively, the check could be skipped if user sets `min_samples_for_sanity` to None

nemo/collections/tts/models/audio_codec.py
```diff
@@ -787,33 +826,76 @@ def _setup_train_dataloader(self, cfg):
...
+            cfg = OmegaConf.create(cfg)
+
+        # Create the dataset
+        dataset = AudioCodecLhotseDataset(
+            sample_rate=self.output_sample_rate, min_samples_for_sanity=cfg.n_samples // 2
+        )
+
+        # --- Update the Lhotse configuration ---
+
+        OmegaConf.set_struct(cfg, False)
```
rlangman: For readability can we cast the config to a `LhotseDataLoadingConfig` before updating all its values?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -787,33 +826,76 @@ def _setup_train_dataloader(self, cfg):
...
+
+        # Set the data loader's sample rate to the codec's output sample rate.
+        # Note that this isn't enough for the Lhotse to automatically resample the audio
+        # since our audio is in a custom field ('target_audio'). We do the resampling
+        # manually in the dataset class.
+        cfg.sample_rate = self.output_sample_rate
+        # Only keep audio files that are at least `min_duration` seconds (and thereby avoid zero-padding)
+        cfg.min_duration = cfg.n_samples / self.output_sample_rate
+        # Randomly select a segment of `n_samples` samples from the audio
+        cfg.truncate_duration = cfg.min_duration
```
rlangman: It would be more intuitive to specify 'truncate_duration' in the config than `n_samples`. `n_samples` is not used for any other purpose than calculating the truncate duration.

rfejgin: Changed to `truncate_duration`

nemo/collections/tts/data/audio_codec_dataset_lhotse.py
rfejgin: There's now a config option `check_audio_sanity`; we only run these checks if it is set to `true`.

nemo/collections/tts/models/audio_codec.py
rfejgin: Comment updated

nemo/collections/tts/models/audio_codec.py
rfejgin: Refactored as suggested. It's also simpler now because I've removed Webdataset support.

nemo/collections/tts/models/audio_codec.py
rfejgin: I've added checking that the config confoms to the`LhotseDataLoadingConfig` schema.

rfejgin:
I've addressed the PR comments. Apologies, but the commit history is lost because I needed to force-push to repair some unsigned commits that were blocking CI from running.

nemo/collections/tts/models/audio_codec.py
blisc: Consider a more descriptive name other than legacy. Maybe "loose_files", "nontarred", or "simple"

blisc (dismissed):

nemo/collections/tts/models/audio_codec.py
rlangman: Our NVIDIA documentation refers to them as "non-tarred" https://docs.nvidia.com/nemo-framework/user-guide/24.09/nemotoolkit/asr/datasets.html

rlangman (dismissed):

nemo/collections/tts/models/audio_codec.py
rfejgin: Thanks - renamed to `non_tarred`.

blisc (approved):

PR #15617, opened 2026-04-16
to_hf.py: add --vllm flag for vLLM-ready checkpoint export

DongjiGao:
## Summary

- Add `vllm=True` flag to `to_hf.py` that produces vLLM-ready checkpoints in one step
- `prepare_for_vllm()` runs after checkpoint consolidation (rank 0 only) and handles architecture detection, chat template generation, tokenizer saving, and generation config
- Depends on #15616 (prompt formatters `to_jinja()`)

### `examples/speechlm2/to_hf.py`
- `HfExportConfig.vllm: bool = False` — opt-in flag
- `_detect_vllm_architecture()` — determines hybrid vs standard from `pretrained_llm`
- `prepare_for_vllm(output_dir, model_cfg)`:
  1. Patches `config.json` with `model_type: "nemo_speechlm"` and detected `architectures`
  2. Calls `PromptFormatter.resolve(prompt_format).to_jinja()` for chat template
  3. Saves tokenizer with `<|audio|>` token and chat template
  4. Writes minimal `generation_config.json` (EOS only — inference params are task-specific)
- Raises `ValueError` for missing `prompt_format`/`pretrained_llm`
- Warns when overwriting existing files

### Architecture mapping
| LLM backbone | Architecture name |
| ------------ | ----------------- |
| Hybrid (NemotronH) | `NeMoSpeechLMHybridForConditionalGeneration` |
| Standard (Qwen3, etc.) | `NeMoSpeechLMForConditionalGeneration` |

## Usage
```bash
torchrun --nproc-per-node=8 --nnodes=4 \
    examples/speechlm2/to_hf.py \
    class_path=nemo.collections.speechlm2.models.salm_automodel.SALMAutomodel \
    ckpt_path=/path/to/checkpoint.ckpt \
    ckpt_config=/path/to/exp_config.yaml \
    output_dir=/path/to/hf_output \
    vllm=True
```

## Test plan
- [x] End-to-end: `to_hf.py vllm=True` on nano-v3-canary1bflash (32 GPUs, 5.5 min)
- [x] Output verified: correct `model_type`, `architectures`, chat template (1149 chars), `generation_config.json`
- [x] Eval: `burst_eval_vllm.py` on produced checkpoint — WER matches baseline (6.31 overall)
- [ ] CI: add "Run CICD" label

Made with [Cursor](https://cursor.com)

PR #15623, opened 2026-04-18
to_hf.py + PromptFormatter: produce vLLM-ready SpeechLM checkpoints with backbone-native chat_template

DongjiGao:
## Background

Reopens [#15617](https://github.com/NVIDIA-NeMo/NeMo/pull/15617) after its base branch (`speechlm2-with-nemo-automodel-merge`) was deleted; now rebased on `main`. Incorporates @pzelasko's review comments from the closed thread.

## Summary

Produces a single checkpoint artifact that serves both HF / NeMo SALM inference and vLLM (SpeechLM plugin) without an opt-in flag, and eliminates the drift between our hand-rolled jinja and the backbone LLM's canonical chat_template.

### Commit 1 - PromptFormatter uses backbone `chat_template` as source of truth

Replaces the hardcoded `TEMPLATE` dict + `to_jinja()` reconstruction in `QwenPromptFormatter` and `NemotronNanoV3PromptFormatter` with a shared base helper that delegates to `tokenizer.apply_chat_template` on the backbone's native jinja. Training and exported-ckpt inference now share one jinja string shipped by the backbone's HF repo - no more train/infer drift from reconstruction mismatches.

- `formatter.py`: add `AUDIO_TOKEN`, `_turn_to_openai_message`, `_is_reasoning_divergence`, `_encode_dialog_from_chat_template`. Drop the abstract `to_jinja` stub.
- `qwen.py`: `QwenPromptFormatter.encode_dialog` delegates; hardcoded `TEMPLATE`, `INFERENCE_PREFIX` and `to_jinja` removed. `Qwen3PromptFormatter` (reasoning, out of refactor scope) unchanged.
- `nemotron_nano_v3.py`: `encode_dialog` delegates and forwards `enable_thinking` to the backbone jinja. Default is `False`; reasoning training raises a clear 'not supported' error because the prefix-based loss mask can't align with the template's fresh `<think>` generation prompt.

### Commit 2 - `to_hf.py` unconditionally produces vLLM-ready checkpoints

- Removes the opt-in `vllm: bool = False` flag; every export is vLLM-ready (addresses @pzelasko's review comment on #15617).
- `_detect_vllm_architecture` now raises `ValueError` on missing `architectures` instead of silently defaulting (addresses the 'broad except' review comment).
- `_try_prepare_for_vllm` wrapper downgrades a `ValueError` to a warning so legacy SALM callers loading the same dir still get a valid HF-only checkpoint.
- Tokenizer saved from the backbone (brings its native `chat_template`) + augmented with `<|audio|>`. `extra_special_tokens` normalized to dict (HF loaders crash on the list form `transformers` writes for `additional_special_tokens`).
- For reasoning backbones (nemotron-nano-v3), the exported `chat_template`'s `enable_thinking` default is flipped to `False` so vLLM's request-time render matches training-time render; otherwise vLLM silently prepends `<think>` and WER regresses from 1.57 to 5.92.
- `hf_hub.py`: `setdefault('model_type', ...)` and `setdefault('architectures', [...])` in `HFHubMixin.save_pretrained` so NeMo-saved checkpoints carry the metadata vLLM / transformers need.

## Architecture mapping

| LLM backbone | Architecture name |
| ------------ | ----------------- |
| Hybrid (NemotronH) | `NeMoSpeechLMHybridForConditionalGeneration` |
| Standard (Qwen3, etc.) | `NeMoSpeechLMForConditionalGeneration` |

## Usage

```bash
torchrun --nproc-per-node=8 --nnodes=4 \
    examples/speechlm2/to_hf.py \
    class_path=nemo.collections.speechlm2.models.salm_automodel.SALMAutomodel \
    ckpt_path=/path/to/checkpoint.ckpt \
    ckpt_config=/path/to/exp_config.yaml \
    output_dir=/path/to/hf_output
```

No `vllm=True` flag any more - output is always vLLM-ready.

## Test plan

- [x] Parity test against backbone jinja for Qwen2.5-7B-Instruct and Nemotron-3-Nano-30B-A3B-BF16; rendered text byte-identical to the previous Python + to_jinja pipeline when `enable_thinking=False`.
- [x] Mask-construction invariants on training turns: `len(context) + len(answer) == len(input)` and `sum(mask) == len(answer)` across Qwen / Nemotron, plain / multi-turn, with / without system, multimodal list content.
- [x] End-to-end vLLM librispeech-pc eval on re-exported nemotron-nano-v3 ckpt: **WER 1.57 (== baseline)**. Without the `enable_thinking` default flip, WER regressed to 5.92 - regression confirmed, fix verified.
- [x] End-to-end NeMo SALM librispeech-pc eval on same ckpt: **WER 1.62, 0 empty hypotheses** (previous pipeline had 75 empties and 5.50 WER-all).
- [x] Backward compat: legacy SALM caller exports fine -- `_try_prepare_for_vllm` downgrades missing-prereq `ValueError` to a warning.
- [ ] CI: add 'Run CICD' label.

Made with Cursor

### DIFF
--- a/examples/speechlm2/to_hf.py
+++ b/examples/speechlm2/to_hf.py
@@ -205,12 +205,6 @@ def prepare_for_vllm(output_dir: str, model_cfg: dict) -> None:
     # unknown to HF; the underlying tokenizer.json is standard, so force
     # the universal base class.
     tok_cfg["tokenizer_class"] = "PreTrainedTokenizerFast"
-    # SpeechLM fine-tuning renders without ``<think>``; flip the template's
-    # default so vLLM rendering matches training (nemotron-nano-v3, etc.).
-    tok_cfg["chat_template"] = tok_cfg.get("chat_template", "").replace(
-        "enable_thinking if enable_thinking is defined else True",
-        "enable_thinking if enable_thinking is defined else False",
-    )
     tok_cfg_path.write_text(json.dumps(tok_cfg, indent=2) + "\n")
 
     # 4. Minimal generation_config.json (EOS only; sampling params belong on
--- a/examples/speechlm2/to_hf.py
+++ b/examples/speechlm2/to_hf.py
@@ -106,6 +106,125 @@ def save_hf_checkpoint(model: torch.nn.Module, state_dict: dict, cfg: HfExportCo
         json.dump(config, f, indent=2)
 
 
+_HYBRID_ARCHITECTURES = {"NemotronHForCausalLM", "NemotronHybridForCausalLM"}
+_AUDIO_TOKEN = "<|audio|>"
+
+
+def _detect_vllm_architecture(model_cfg: dict) -> str:
+    """Determine the vLLM plugin model class from the pretrained LLM backbone.
+
+    Raises:
+        ValueError: if the HF config can't be loaded or has no 'architectures'.
+    """
+    pretrained_llm = model_cfg.get("pretrained_llm", "")
+    try:
+        from transformers import AutoConfig
+
+        llm_cfg = AutoConfig.from_pretrained(pretrained_llm, trust_remote_code=True)
+    except Exception as e:
+        raise ValueError(
+            f"Could not load HF config for pretrained_llm={pretrained_llm!r}: {e}. "
+            f"Fix the 'pretrained_llm' field or ensure HF access during conversion."
+        ) from e
+
+    archs = getattr(llm_cfg, "architectures", [])
+    if not archs:
+        raise ValueError(
+            f"HF config for {pretrained_llm!r} has empty 'architectures'."
+        )
+
+    if set(archs) & _HYBRID_ARCHITECTURES:
+        return "NeMoSpeechLMHybridForConditionalGeneration"
+    return "NeMoSpeechLMForConditionalGeneration"
+
+
+def prepare_for_vllm(output_dir: str, model_cfg: dict) -> None:
+    """Patch a saved checkpoint to be vLLM-ready.
+
+    Adds tokenizer (with audio token and chat template), patches config.json
+    with model_type/architectures, and writes generation_config.json.
+
+    Args:
+        output_dir: Path to the HuggingFace checkpoint directory.
+        model_cfg: Model config dict (from experiment YAML).
+
+    Raises:
+        ValueError: If ``pretrained_llm`` is missing.
+    """
+    from transformers import AutoTokenizer
+
+    from nemo.utils import logging as LOG
+
+    output_dir = Path(output_dir)
+    pretrained_llm = model_cfg.get("pretrained_llm", "")
+    if not pretrained_llm:
+        raise ValueError("model config has no 'pretrained_llm'; cannot load tokenizer for vLLM")
+
+    # 1. Detect architecture and patch config.json
+    arch = _detect_vllm_architecture(model_cfg)
+    config_path = output_dir / "config.json"
+    config = json.loads(config_path.read_text())
+    config["model_type"] = "nemo_speechlm"
+    config["architectures"] = [arch]
+    config_path.write_text(json.dumps(config, indent=2) + "\n")
+
+    # 2. Save tokenizer (backbone chat_template carries over via save_pretrained)
+    existing = [
+        f.name
+        for f in output_dir.iterdir()
+        if f.name in ("tokenizer_config.json", "tokenizer.json", "generation_config.json")
+    ]
+    if existing:
+        LOG.info("Overwriting existing files in %s: %s", output_dir, existing)
+    tok = AutoTokenizer.from_pretrained(pretrained_llm, trust_remote_code=True)
+    if _AUDIO_TOKEN not in tok.get_vocab():
+        tok.add_special_tokens({"additional_special_tokens": [_AUDIO_TOKEN]})
+    tok.save_pretrained(str(output_dir))
+    # A separate chat_template.jinja file, if present, overrides the inline copy
+    # in tokenizer_config.json. Remove it so tokenizer_config.json wins.
+    jinja_file = output_dir / "chat_template.jinja"
+    if jinja_file.exists():
+        jinja_file.unlink()
+    # Normalize extra_special_tokens: transformers writes our added audio token
+    # as a list, but HF/vLLM loaders expect a dict keyed by semantic name.
+    tok_cfg_path = output_dir / "tokenizer_config.json"
+    tok_cfg = json.loads(tok_cfg_path.read_text())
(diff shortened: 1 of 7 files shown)


nemo/collections/common/prompts/formatter.py
```diff
@@ -201,6 +201,120 @@ def __init_subclass__(cls, **kwargs) -> None:
...
+                messages,
+                tokenize=False,
+                add_generation_prompt=True,
+                **template_kwargs,
+            )
+            input_ids = torch.tensor(self._apply_tokenizer(text), dtype=torch.long)
+            return {"input_ids": input_ids, "context_ids": input_ids}
+
+        # Training: render twice (full dialog + context without last turn) so the
+        # tail of the full rendering becomes the loss-masked answer tokens.
```
DongjiGao, line 290: not sure if there's a better way to do this

examples/speechlm2/to_hf.py
```diff
@@ -158,14 +275,10 @@ def main(cfg: HfExportConfig):
 
     strategy_cfg = full_cfg.get("trainer", {}).get("strategy", {})
 
-    _is_torchrun = "RANK" in os.environ
-    if _is_torchrun and dist.is_available() and not dist.is_initialized():
+    if dist.is_available() and not dist.is_initialized():
```
pzelasko, line 295: Why are we removing `_is_torchrun`?

nemo/collections/common/prompts/formatter.py
```diff
@@ -201,6 +201,120 @@ def __init_subclass__(cls, **kwargs) -> None:
                 ), f"{ERR} Slot '{slot}' not found in template '{template}' for role '{role}'"
         super().__init_subclass__(**kwargs)
 
+    AUDIO_TOKEN = "<|audio|>"
```
pzelasko, line 204: Why is this defined globally? What happens when the model defines a different audio placeholder?

nemo/collections/common/prompts/formatter.py
pzelasko: When we have:
```python
messages = [
  {"role": "user", "content": "..."},
  {"role": "assistant", "content": "..."},
  {"role": "user", "content": "..."},
  {"role": "assistant", "content": "..."},
]
```

the mask should be `[0, 0, 0, ..., 1, 1, 1, ..., 0, 0, 0, ..., 1, 1, 1]` to be compatible with current PromptFormatter implementation. This is not possible with `apply_chat_template` approach in the current form.

pzelasko (changes requested):
I like your thinking about using chat template directly but for the reasons below I don't think it is feasible in the current shape. 

Let's approach this from two angles.

1. In this PR don't add chat template. Revert the changes from nemotron and qwen prompt formatter. Then fix `Nemotron...PromptFormatter` so that it's logic is equivalent to that of jinja, but using PromptFormatter. 
2. We can add `ChatTemplatePromptFormatter` in a separate PR which would be auto-selected when `prompt_format: "tokenizer_chat_template"` and make it generic for HF models.

pzelasko:
This PR also needs unit test coverage of the changes.

DongjiGao:
> I like your thinking about using chat template directly but for the reasons below I don't think it is feasible in the current shape.
> 
> Let's approach this from two angles.
> 
> 1. In this PR don't add chat template. Revert the changes from nemotron and qwen prompt formatter. Then fix `Nemotron...PromptFormatter` so that it's logic is equivalent to that of jinja, but using PromptFormatter.
> 2. We can add `ChatTemplatePromptFormatter` in a separate PR which would be auto-selected when `prompt_format: "tokenizer_chat_template"` and make it generic for HF models.

Could you please explain a bit why it is not feasible in current shape? Thanks!

examples/speechlm2/to_hf.py
DongjiGao: Reverted.

nemo/collections/common/prompts/formatter.py
DongjiGao: Fix. Now directly read from training config. Raise error if not exist.

nemo/collections/common/prompts/qwen.py
```diff
@@ -24,7 +24,10 @@
 class QwenPromptFormatter(PromptFormatter):
     NAME = "qwen"
     OUTPUT_ROLE = "assistant"
-    INFERENCE_PREFIX = f"{QWEN_BOT}assistant\n"
+    # Qwen3's chat_template injects this when enable_thinking=False; mirror it
+    # here so the SpeechLM fine-tune matches pre-training "reasoning-off" mode.
+    NO_THINK_PREFIX = "<think>\n\n</think>\n\n"
+    INFERENCE_PREFIX = f"{QWEN_BOT}assistant\n{NO_THINK_PREFIX}"
```
pzelasko: I'm afraid this will be a breaking change for `canary-qwen-2.5b`. Let's roll back and add a separate `Qwen3PromptFormatter` in a follow up PR to reduce the scope here.

DongjiGao: Yes you are right. I have verified that this would degrade canary-qwen-2.5 as it did not see the thinking tag during training. Reverting it.

pzelasko (approved):
LGTM

nemo/collections/common/prompts/formatter.py
DongjiGao: Will solve this in a separate PR.

pzelasko (approved):

examples/speechlm2/to_hf.py
```diff
@@ -106,14 +107,134 @@ def save_hf_checkpoint(model: torch.nn.Module, state_dict: dict, cfg: HfExportCo
...
+    output_dir = Path(output_dir)
+    pretrained_llm = model_cfg.get("pretrained_llm", "")
+    if not pretrained_llm:
+        raise ValueError("model config has no 'pretrained_llm'; cannot load tokenizer for vLLM")
+
+    # ``model.audio_locator_tag`` is the SoT for the audio placeholder;
+    # fail loud rather than default, since a mismatch is silent at inference.
+    audio_token = model_cfg.get("audio_locator_tag")
+    if not audio_token:
+        raise ValueError("model config has no 'audio_locator_tag' (set it in the training YAML).")
```
stevehuang52, line 165: I think we shouldn't make `audio_locator_tag` as a required field for this script, since this script is used for any model in speechlm2 collection, and we shouldn't assume that this param will always be set in the config. If this script is `salm_to_hf.py`, it's fine to have such assumption, but this script is for more geneal speechlm models. @DongjiGao our streaming speechlm doesn't have this param in cfg.

PR #15639, opened 2026-04-22
Easymagpie ZeroShot disabling option using speaker encoder

paarthneekhara:


nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -362,6 +363,33 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+        # This enables keeping the zero-shot conditioning module private at release time.
+
+        self.use_speaker_encoder = cfg.get('use_speaker_encoder', False)
+        self.train_shuffle_context_embedding_prob = 0.0
+        if self.use_speaker_encoder:
+            speaker_encoder_cfg = cfg.get('speaker_encoder', cfg.get('context_encoder', None))
+            if speaker_encoder_cfg is not None:
+                speaker_encoder_cfg = dict(speaker_encoder_cfg)
+                speaker_encoder_cfg.pop('router_load_balancing_loss_coeff', None)
+                speaker_encoder_cfg.pop('router_z_loss_coeff', None)
```
blisc: Why do you expect the context_encoder to have these parameters related to MoE?

blisc: Let's raise an Exception if use_moe is true. And let's log a warning if we detect these parameters intead of popping them from the dictionary. Something like we do not recommend using moe for the context encoder

blisc: Let's raise [NeMoBaseException ](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/utils/exceptions.py#L16)instead of Exception or create another more appropriate one

blisc (commented):
Can you make sure tests run and pass?

blisc:
/ok to test [7c76327](https://github.com/NVIDIA-NeMo/NeMo/pull/15639/commits/7c76327a258b45346ce96a25c139b71c246c0973)

blisc (approved):

PR #15447, opened 2026-02-26
SALM with NeMo Automodel integration for Nemotron Nano V3 LLM backbone

pzelasko:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

Add `SALMAutomodel` — a new SALM variant that uses [NeMo Automodel](https://github.com/NVIDIA-NeMo/Automodel) instead of HuggingFace Transformers for the LLM backbone, enabling efficient training and inference of Speech LLMs with MoE architectures (e.g., Nemotron Nano V3 30B/3B) via native FSDP2, TP, EP, and Grouped GEMM support. CP support will be added later and PP support is not planned.

**Collection**: speechlm2, common

# Changelog

- Add `SALMAutomodel` model class (`nemo/collections/speechlm2/models/salm_automodel.py`) — a `LightningModule` that replaces HF `AutoModelForCausalLM` with NeMo Automodel's distributed LLM loader, supporting deferred initialization via `configure_model()` for memory-efficient per-GPU shard loading.
- Add `AutomodelParallelStrategy` (`nemo/collections/speechlm2/parts/parallel.py`) — a custom Lightning `ModelParallelStrategy` that delegates device mesh creation to NeMo Automodel, supporting FSDP2, TP, PP, CP, EP (MoE), and HSDP.
- Add native Automodel LoRA support (`nemo/collections/speechlm2/parts/automodel_lora.py`) — replaces HuggingFace PEFT with Automodel's built-in `ModuleMatcher`-based LoRA, applied before FSDP2 sharding for correct meta-device handling.
- Add `FlashPrecision` plugin (`nemo/utils/trainer_utils.py`) — a new `bf16-flash`/`fp16-flash` precision mode that sets module dtype without mutating `torch.set_default_dtype`, preventing dtype conflicts with Automodel's internal initialization.
- Add `FlashOptim` compatibility patch (`nemo/core/optim/flash_optim.py`) — fixes `DTensor.from_local()` shape inference for unevenly-sharded FSDP2 parameters during DCP save/load.
- Add `NemotronNanoV3PromptFormatter` (`nemo/collections/common/prompts/nemotron_nano_v3.py`) with `<think>` reasoning support.
- Add dataloader DP rank resolution patch for Automodel's device mesh (handles `dp_replicate`/`dp_shard` naming).
[...]

PR #15659, opened 2026-05-04
[ci, build] fix: unblock testpypi publish — drop direct git URL dep, bump CI template to v0.94.1

ko3n1g:
<details><summary>Claude summary</summary>

The TestPyPI publish job has failed on every push to `main` since 2026-04-23 with `400 Bad Request` from `https://test.pypi.org/legacy/`.

**Cause.** PyPI/TestPyPI rejects any project whose metadata declares a direct URL/VCS dependency (PEP 508 `pkg @ <url>` form). `requirements/requirements_speechlm2.txt` (introduced 2026-04-23 in #15447) contained:

```
nemo_automodel @ git+https://github.com/NVIDIA-NeMo/Automodel.git@9eccbb6102a260efd7cbdffa890fc57b94f94528
```

With `--verbose` enabled, TestPyPI surfaces the actual response:

```
400 Can't have direct dependency: nemo_automodel @ ...fa890fc57b94f94528 ; extra == "speechlm2-only"
```

**Fix.**

1. **Keep `nemo_automodel` as a tracked dependency, just not in the wheel metadata.** Remove the `nemo_automodel @ git+...` line from `requirements/requirements_speechlm2.txt` (which feeds `extras_require['speechlm2-only']` via `setup.py`, and is therefore baked into wheel metadata) and instead pin the commit in `requirements/manifest.json` alongside the other VCS deps.
2. **Override-install the pinned commit in the GitHub CI image.** `docker/common/install_dep.sh extra` (invoked by `docker/Dockerfile.ci`, which is what GitHub CI builds) reads the pinned ref from `manifest.json` and adds `nemo_automodel @ git+<repo>@<ref>` to its `pip install` step. Net effect: the published wheel has clean metadata, but every CI image still gets the exact same `nemo_automodel` commit.
3. **Surface future twine errors directly.** Bump `_build_test_publish_wheel` from `v0.22.3` → `v0.94.1` (the newer reusable workflow already passes `--verbose --skip-existing` to `twine upload`). Add `workflow_dispatch:` so the workflow can be triggered for testing without waiting on a push to `main`.

**Layout after this PR:**

| Where | What |
|------|------|
| `requirements/manifest.json` → `vcs-dependencies.nemo_automodel` | Pinned `repo` + `ref` (single source of truth) |
| `docker/common/install_dep.sh` → `extra()` | Reads manifest, override-installs `nemo_automodel @ git+<repo>@<ref>` |
| `requirements/requirements_speechlm2.txt` | `flashoptim` + a comment pointing to the manifest/install_dep entry |

**Verification.** Re-triggered on this branch via `workflow_dispatch`; all jobs green. `https://test.pypi.org/simple/nemo-toolkit/` now returns `200` (was `404`).

</details>

### DIFF
--- a/requirements/requirements_speechlm2.txt
+++ b/requirements/requirements_speechlm2.txt
@@ -1,1 +1,5 @@
+# nemo_automodel is also required at runtime but is pinned in
+# requirements/manifest.json (vcs-dependencies.nemo_automodel) and installed by
+# docker/common/install_dep.sh `extra` — direct git URL deps are rejected by
+# PyPI/TestPyPI (`400 Can't have direct dependency`) so it cannot live here.
 flashoptim
--- a/requirements/requirements_speechlm2.txt
+++ b/requirements/requirements_speechlm2.txt
@@ -1,2 +1,1 @@
-nemo_automodel @ git+https://github.com/NVIDIA-NeMo/Automodel.git@9eccbb6102a260efd7cbdffa890fc57b94f94528  # version with relaxed pytorch constraints
 flashoptim
--- a/.github/workflows/build-test-publish-wheel.yml
+++ b/.github/workflows/build-test-publish-wheel.yml
@@ -19,6 +19,7 @@ on:
     branches:
       - main
       - "r[0-9]*"
+  workflow_dispatch:
 
 defaults:
   run:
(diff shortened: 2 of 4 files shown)


requirements/requirements_speechlm2.txt
```diff
@@ -1,2 +1 @@
-nemo_automodel @ git+https://github.com/NVIDIA-NeMo/Automodel.git@9eccbb6102a260efd7cbdffa890fc57b94f94528  # version with relaxed pytorch constraints
```
chtruong814, line 1: We should leave automodel here?  or pin to a specific version that included this commit?

chtruong814 (approved):

PR #15661, opened 2026-05-04
Add Transformer Encoder for ASR 

nithinraok:
> [!IMPORTANT]  

# What does this PR do ?

Adds a new ASR transformer encoder with frame stacking, configurable subsampling, FlexAttention-based full attention, and optional QK Norm 

**Collection**: ASR

# Changelog

  - Add TransformerEncoder in `nemo/collections/asr/modules/transformer_encoder.py` as a lightweight pre-norm transformer encoder for ASR.
  - Add `TransformerEncoderConfig` dataclass to capture encoder hyperparameters such as d_model, n_heads, n_layers, qkv_bias, qk_norm, and
    subsampling_factor.
  - Add `FeatureStacking` pre-encoder module to stack consecutive frames, reduce sequence length, and project stacked features into the model
    dimension.
  - Add transformer building blocks: `FeedForward`, `MultiHeadAttention`, and `TransformerBlock`.
  - Implement full-attention masking with PyTorch FlexAttention and padding-aware block masks.
  - Add optional per-head QK normalization before attention score computation.
  - Restrict the initial PR scope to attn_mode="full" and raise an error for unsupported future modes.
  - Export TransformerEncoder from `nemo/collections/asr/modules/__init__.py` for config-based instantiation.
  - Add example config in `examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml` showing how to train an RNNT/TDT model with the new
    encoder.
# Usage

```python
  import torch

  from nemo.collections.asr.modules import TransformerEncoder

  encoder = TransformerEncoder(
      feat_in=128,
      d_model=512,
      n_heads=8,
      n_layers=12,
      drop_rate=0.1,
      qkv_bias=False,
      qk_norm=True,
      subsampling_factor=4,
      attn_mode="full",
  )

  audio_signal = torch.randn(2, 128, 400)   # (B, C, T)
  lengths = torch.tensor([400, 360])        # valid frame lengths

  encoded, encoded_lengths = encoder(audio_signal, lengths)

  print(encoded.shape)         # (B, D, T')
  print(encoded_lengths)       # ceil(length / subsampling_factor)

```

You can also instantiate it from Hydra config with:

```yaml
  encoder:
    _target_: nemo.collections.asr.modules.transformer_encoder.TransformerEncoder
    feat_in: ${model.preprocessor.features}
    d_model: 1280
    n_heads: 16
    n_layers: 32
    drop_rate: 0.1
    qkv_bias: false
    qk_norm: true
    subsampling_factor: 8
    attn_mode: full
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
+++ b/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
@@ -81,6 +81,7 @@ model:
     qkv_bias: false
     qk_norm: true
     ff_expansion: 4
+    pre_block_norm: true
     subsampling_factor: 8
 
   decoder:
--- a/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
+++ b/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
@@ -1,0 +1,184 @@
+name: "Transformer-Stacking-TDT-BPE"
+
+model:
+  sample_rate: 16000
+  compute_eval_loss: false
+  log_prediction: true
+  rnnt_reduction: 'mean_volume'
+  skip_nan_grad: false
+
+  model_defaults:
+    enc_hidden: ${model.encoder.d_model}
+    pred_hidden: 640
+    joint_hidden: 640
+    tdt_durations: [0, 1, 2, 3, 4]
+    num_tdt_durations: 5
+
+  train_ds:
+    manifest_filepath: ???
+    sample_rate: ${model.sample_rate}
+    batch_size: 16
+    shuffle: true
+    num_workers: 8
+    pin_memory: true
+    max_duration: 20
+    min_duration: 0.1
+    is_tarred: false
+    tarred_audio_filepaths: null
+    shuffle_n: 2048
+    bucketing_strategy: "fully_randomized"
+    bucketing_batch_size: null
+
+  validation_ds:
+    manifest_filepath: ???
+    sample_rate: ${model.sample_rate}
+    batch_size: 16
+    shuffle: false
+    use_start_end_token: false
+    num_workers: 8
+    pin_memory: true
+
+  test_ds:
+    manifest_filepath: null
+    sample_rate: ${model.sample_rate}
+    batch_size: 16
+    shuffle: false
+    use_start_end_token: false
+    num_workers: 8
+    pin_memory: true
+
+  tokenizer:
+    dir: ???
+    type: bpe
+
+  preprocessor:
+    _target_: nemo.collections.asr.modules.AudioToMelSpectrogramPreprocessor
+    sample_rate: ${model.sample_rate}
+    normalize: "per_feature"
+    window_size: 0.025
+    window_stride: 0.01
+    window: "hann"
+    features: 128
+    n_fft: 512
+    frame_splicing: 1
+    dither: 0.00001
+    pad_to: 0
+
+  spec_augment:
+    _target_: nemo.collections.asr.modules.SpectrogramAugmentation
+    freq_masks: 2
+    time_masks: 10
+    freq_width: 27
+    time_width: 0.05
+
+  encoder:
+    _target_: nemo.collections.asr.modules.transformer_encoder.TransformerEncoder
+    feat_in: ${model.preprocessor.features}
+    d_model: 1280
+    n_heads: 16
+    n_layers: 32
+    drop_rate: 0.1
+    qkv_bias: false
+    qk_norm: true
+    subsampling_factor: 8
+
+  decoder:
+    _target_: nemo.collections.asr.modules.RNNTDecoder
+    normalization_mode: null
+    random_state_sampling: false
+    blank_as_pad: true
+
+    prednet:
+      pred_hidden: ${model.model_defaults.pred_hidden}
+      pred_rnn_layers: 1
+      t_max: null
+      dropout: 0.2
+
+  joint:
+    _target_: nemo.collections.asr.modules.RNNTJoint
+    log_softmax: null
+    preserve_memory: false
+    fuse_loss_wer: false
+    fused_batch_size: 4
+
+    jointnet:
+      joint_hidden: ${model.model_defaults.joint_hidden}
+      activation: "relu"
+      dropout: 0.2
+
+  decoding:
+    strategy: "greedy_batch"
+    model_type: "tdt"
+    durations: ${model.model_defaults.tdt_durations}
+
+    greedy:
+      max_symbols: 10
+      use_cuda_graph_decoder: true
+
+    beam:
+      beam_size: 2
+      return_best_hypothesis: false
+      score_norm: true
+      tsd_max_sym_exp: 50
+      alsd_max_target_len: 2.0
+
+  loss:
+    loss_name: "tdt"
+    tdt_kwargs:
+      fastemit_lambda: 0.0
+      clamp: -1.0
+      durations: ${model.model_defaults.tdt_durations}
+      sigma: 0.02
+      omega: 0.1
+
+  optim:
+    name: adamw
+    lr: 5e-4
+    betas: [0.9, 0.95]
+    weight_decay: 1e-2
+
+    sched:
+      name: CosineAnnealing
+      warmup_steps: 25000
+      warmup_ratio: null
+      min_lr: 5e-5
+
+trainer:
+  devices: -1
+  num_nodes: 1
+  max_epochs: 500
+  max_steps: -1
+  val_check_interval: 1.0
+  accelerator: auto
+  strategy:
+    _target_: lightning.pytorch.strategies.DDPStrategy
+    gradient_as_bucket_view: true
+  accumulate_grad_batches: 1
+  gradient_clip_val: 1.0
+  precision: bf16
+  log_every_n_steps: 10
+  enable_progress_bar: true
+  num_sanity_val_steps: 0
+  check_val_every_n_epoch: 1
+  sync_batchnorm: true
+  enable_checkpointing: false
+  logger: false
+  benchmark: false
+
+exp_manager:
+  exp_dir: null
+  name: ${name}
+  create_tensorboard_logger: false
+  create_checkpoint_callback: true
+  checkpoint_callback_params:
+    monitor: "val_wer"
+    mode: "min"
+    save_top_k: 5
+    always_save_nemo: true
+  resume_from_checkpoint: null
+  resume_if_exists: false
+  resume_ignore_no_checkpoint: false
+  create_wandb_logger: false
+  wandb_logger_kwargs:
+    name: null
+    project: null
--- a/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
+++ b/examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
@@ -80,6 +80,7 @@ model:
     drop_rate: 0.1
     qkv_bias: false
     qk_norm: true
+    ff_expansion: 4
     subsampling_factor: 8
 
   decoder:
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -31,6 +31,7 @@ class TransformerEncoderConfig:
     qkv_bias: bool = False
     qk_norm: bool = False
     ff_expansion: float = 4.0
+    pre_block_norm: bool = True
     subsampling_factor: int = 4
     # Attention mode — currently only "full" is supported.
     # Future: "causal", "lookahead", "local", "sliding_window"
@@ -168,6 +169,10 @@ class TransformerEncoder(nn.Module):
         qkv_bias: Whether to use bias in Q/K/V projections.
         qk_norm: Whether to apply per-head LayerNorm to Q and K before the dot product.
         ff_expansion: Feed-forward expansion factor (float to support sub-1x for MoE).
+        pre_block_norm: If True (default), apply LayerNorm to embeddings before the first
+            transformer block (BERT/ViT-style). Set False to match pre-norm transformers
+            such as Whisper or GPT-2 — required when loading pretrained weights from those
+            checkpoints.
         subsampling_factor: Frame stacking factor for the pre-encoder.
         attn_mode: Attention pattern — currently only "full" (bidirectional) is supported.
     """
@@ -182,6 +187,7 @@ class TransformerEncoder(nn.Module):
         qkv_bias: bool = False,
         qk_norm: bool = False,
         ff_expansion: float = 4.0,
+        pre_block_norm: bool = True,
         subsampling_factor: int = 4,
         attn_mode: str = "full",
     ):
@@ -200,13 +206,14 @@ class TransformerEncoder(nn.Module):
             qkv_bias=qkv_bias,
             qk_norm=qk_norm,
             ff_expansion=ff_expansion,
+            pre_block_norm=pre_block_norm,
             subsampling_factor=subsampling_factor,
             attn_mode=attn_mode,
         )
         self.d_model = d_model
 
         self.pre_encode = FeatureStacking(subsampling_factor, feat_in, d_model)
-        self.embed_norm = nn.LayerNorm(d_model)
+        self.embed_norm = nn.LayerNorm(d_model) if pre_block_norm else nn.Identity()
         self.layers = nn.ModuleList([TransformerBlock(cfg) for _ in range(n_layers)])
         self.final_norm = nn.LayerNorm(d_model)
 
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -218,7 +218,7 @@ class TransformerEncoder(nn.Module):
             length: (B,) — output lengths after subsampling.
         """
         x, length = self.pre_encode(audio_signal, length)
-        
+
         x = self.embed_norm(x)
 
         B, T, _ = x.shape
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -30,6 +30,7 @@ class TransformerEncoderConfig:
     drop_rate: float = 0.1
     qkv_bias: bool = False
     qk_norm: bool = False
+    ff_expansion: float = 4.0
     subsampling_factor: int = 4
     # Attention mode — currently only "full" is supported.
     # Future: "causal", "lookahead", "local", "sliding_window"
@@ -63,6 +64,9 @@ class FeatureStacking(nn.Module):
         self.subsampling_factor = subsampling_factor
         self.proj = nn.Linear(subsampling_factor * feat_in, feat_out, bias=False)
 
+    def compute_num_out_frames(self, in_frames):
+        return (in_frames + self.subsampling_factor - 1) // self.subsampling_factor
+
     def forward(self, x, lengths):
         """
         Args:
@@ -80,18 +84,19 @@ class FeatureStacking(nn.Module):
         t_new = (t + pad_size) // self.subsampling_factor
         x = x.reshape(b, t_new, c * self.subsampling_factor)
         x = self.proj(x)
-        lengths = torch.div(lengths + self.subsampling_factor - 1, self.subsampling_factor, rounding_mode='floor')
+        lengths = self.compute_num_out_frames(lengths)
         return x, lengths
 
 
 class FeedForward(nn.Module):
     def __init__(self, cfg: TransformerEncoderConfig):
         super().__init__()
+        ff_hidden = int(cfg.ff_expansion * cfg.d_model)
         self.net = nn.Sequential(
-            nn.Linear(cfg.d_model, 4 * cfg.d_model),
+            nn.Linear(cfg.d_model, ff_hidden),
             nn.GELU(),
             nn.Dropout(cfg.drop_rate),
-            nn.Linear(4 * cfg.d_model, cfg.d_model),
+            nn.Linear(ff_hidden, cfg.d_model),
             nn.Dropout(cfg.drop_rate),
         )
 
@@ -106,9 +111,7 @@ class MultiHeadAttention(nn.Module):
         self.head_dim = cfg.d_model // cfg.n_heads
         self.d_model = cfg.d_model
 
-        self.w_q = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
-        self.w_k = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
-        self.w_v = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
+        self.w_qkv = nn.Linear(cfg.d_model, 3 * cfg.d_model, bias=cfg.qkv_bias)
         self.out_proj = nn.Linear(cfg.d_model, cfg.d_model)
 
         self.qk_norm = cfg.qk_norm
@@ -120,9 +123,8 @@ class MultiHeadAttention(nn.Module):
         B, T, _ = x.shape
         H, D = self.n_heads, self.head_dim
 
-        q = self.w_q(x).view(B, T, H, D).transpose(1, 2)
-        k = self.w_k(x).view(B, T, H, D).transpose(1, 2)
-        v = self.w_v(x).view(B, T, H, D).transpose(1, 2)
+        qkv = self.w_qkv(x).view(B, T, 3, H, D).permute(2, 0, 3, 1, 4)
+        q, k, v = qkv.unbind(0)
 
         if self.qk_norm:
             q = self.q_norm(q).to(v.dtype)
@@ -165,6 +167,7 @@ class TransformerEncoder(nn.Module):
         drop_rate: Dropout probability.
         qkv_bias: Whether to use bias in Q/K/V projections.
         qk_norm: Whether to apply per-head LayerNorm to Q and K before the dot product.
+        ff_expansion: Feed-forward expansion factor (float to support sub-1x for MoE).
         subsampling_factor: Frame stacking factor for the pre-encoder.
         attn_mode: Attention pattern — currently only "full" (bidirectional) is supported.
     """
@@ -178,6 +181,7 @@ class TransformerEncoder(nn.Module):
         drop_rate: float = 0.1,
         qkv_bias: bool = False,
         qk_norm: bool = False,
+        ff_expansion: float = 4.0,
         subsampling_factor: int = 4,
         attn_mode: str = "full",
     ):
@@ -193,6 +197,7 @@ class TransformerEncoder(nn.Module):
             drop_rate=drop_rate,
             qkv_bias=qkv_bias,
             qk_norm=qk_norm,
+            ff_expansion=ff_expansion,
             subsampling_factor=subsampling_factor,
             attn_mode=attn_mode,
         )
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -218,7 +218,7 @@ class TransformerEncoder(nn.Module):
             length: (B,) — output lengths after subsampling.
         """
         x, length = self.pre_encode(audio_signal, length)
-        x = x * (self.d_model**0.5)
+        
         x = self.embed_norm(x)
 
         B, T, _ = x.shape
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -80,7 +80,7 @@ class FeatureStacking(nn.Module):
         t_new = (t + pad_size) // self.subsampling_factor
         x = x.reshape(b, t_new, c * self.subsampling_factor)
         x = self.proj(x)
-        lengths = torch.div(lengths + pad_size, self.subsampling_factor, rounding_mode='floor')
+        lengths = torch.div(lengths + self.subsampling_factor - 1, self.subsampling_factor, rounding_mode='floor')
         return x, lengths
 
 
--- a/nemo/collections/asr/modules/transformer_encoder.py
+++ b/nemo/collections/asr/modules/transformer_encoder.py
@@ -1,0 +1,227 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+
+import torch
+import torch.nn as nn
+from torch.nn.attention.flex_attention import create_block_mask, flex_attention
+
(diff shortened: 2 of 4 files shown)


nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+class MultiHeadAttention(nn.Module):
+    def __init__(self, cfg: TransformerEncoderConfig):
+        super().__init__()
+        self.n_heads = cfg.n_heads
+        self.head_dim = cfg.d_model // cfg.n_heads
+        self.d_model = cfg.d_model
+
+        self.w_q = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
```
pzelasko: let's merge these 3 lines into w_qkv to batch 3 ops in one

pzelasko: let's merge these 3 lines into w_qkv to batch 3 ops in one

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+        self.qk_norm = cfg.qk_norm
+        if cfg.qk_norm:
+            self.q_norm = nn.LayerNorm(self.head_dim)
+            self.k_norm = nn.LayerNorm(self.head_dim)
+
+    def forward(self, x, block_mask=None):
+        B, T, _ = x.shape
+        H, D = self.n_heads, self.head_dim
+
+        q = self.w_q(x).view(B, T, H, D).transpose(1, 2)
```
pzelasko: and then batch here as well

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self.w_q = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
+        self.w_k = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
+        self.w_v = nn.Linear(cfg.d_model, cfg.d_model, bias=cfg.qkv_bias)
+        self.out_proj = nn.Linear(cfg.d_model, cfg.d_model)
+
+        self.qk_norm = cfg.qk_norm
+        if cfg.qk_norm:
+            self.q_norm = nn.LayerNorm(self.head_dim)
+            self.k_norm = nn.LayerNorm(self.head_dim)
```
pzelasko, line 121: what about RMSNorm?

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+            lengths: (B,) — updated lengths after subsampling.
+        """
+        x = x.transpose(1, 2)  # (B, C, T) -> (B, T, C)
+        b, t, c = x.size()
+        pad_size = (self.subsampling_factor - (t % self.subsampling_factor)) % self.subsampling_factor
+        if pad_size > 0:
+            x = nn.functional.pad(x, (0, 0, 0, pad_size))
+        t_new = (t + pad_size) // self.subsampling_factor
+        x = x.reshape(b, t_new, c * self.subsampling_factor)
+        x = self.proj(x)
```
pzelasko, line 87: do we want to add some norm before projection to center the activations? rms norm could be reasonable although then we might need to change the zero-padding above to something that doesn't skew the norm too much

pzelasko, line 87: do we want to add some norm before projection to center the activations? rms norm could be reasonable although then we might need to change the zero-padding above to something that doesn't skew the norm too much

nemo/collections/asr/modules/transformer_encoder.py
pzelasko, line 121: what about RMSNorm?

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        Args:
+            audio_signal: (B, C, T) — mel spectrogram from preprocessor.
+            length: (B,) — valid frame counts per sample.
+        Returns:
+            x: (B, D, T') — encoded representation (channels-first).
+            length: (B,) — output lengths after subsampling.
+        """
+        x, length = self.pre_encode(audio_signal, length)
+        x = x * (self.d_model**0.5)
```
pzelasko: why re-scale just before norm?

tests/collections/asr/test_transformer_encoder.py
```diff
@@ -0,0 +1,266 @@
...
+    def test_padding_when_not_divisible(self):
+        B, C, T = 1, 80, 401
+        subsampling_factor = 4
+        stacking = FeatureStacking(subsampling_factor=subsampling_factor, feat_in=C, feat_out=256)
+        x = torch.randn(B, C, T)
+        lengths = torch.tensor([401])
+
+        out, out_lengths = stacking(x, lengths)
+        pad_size = (subsampling_factor - (T % subsampling_factor)) % subsampling_factor
+        expected_t = (T + pad_size) // subsampling_factor
```
pzelasko: can we have a method `FeatureStacking.compute_num_out_frames(in_frames: int) -> int` that is a reliable source of truth for the shape info?

tests/collections/asr/test_transformer_encoder.py
```diff
@@ -0,0 +1,266 @@
...
+
+        model.eval()
+        with torch.no_grad():
+            out, _ = model(x, lengths)
+
+        assert out.shape[1] == 64  # D dimension
+        assert out.shape[2] == 200 // 4  # T dimension
+
+    @pytest.mark.run_only_on('GPU')
+    def test_eval_deterministic(self):
```
pzelasko, line 214: not sure we can actually guarantee that? at the very least let's use torch.allclose

tests/collections/asr/test_transformer_encoder.py
```diff
@@ -0,0 +1,266 @@
...
+        lengths = torch.tensor([200], device='cuda')
+
+        with torch.no_grad():
+            out1, _ = model(x, lengths)
+            out2, _ = model(x, lengths)
+
+        assert torch.equal(out1, out2)
+
+    @pytest.mark.run_only_on('GPU')
+    def test_padding_does_not_affect_valid_output(self):
```
pzelasko, line 229: wouldn't the last frame's result depend on the amount of padding added in frame stacking though?

tests/collections/asr/test_transformer_encoder.py
pzelasko, line 214: not sure we can actually guarantee that? at the very least let's use torch.allclose

pzelasko (commented):
nice work!

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+        x = self.proj(x)
+        lengths = torch.div(lengths + self.subsampling_factor - 1, self.subsampling_factor, rounding_mode='floor')
+        return x, lengths
+
+
+class FeedForward(nn.Module):
+    def __init__(self, cfg: TransformerEncoderConfig):
+        super().__init__()
+        self.net = nn.Sequential(
+            nn.Linear(cfg.d_model, 4 * cfg.d_model),
```
tango4j: This should never be hardcoded since this is very frequently customized variable. 
`cfg.ff_expansion` should be added, replacing the hardcoded `4`.

Also, I think we should add this function to `TransformerEncoderConfig`
```
@property
    def ff_hidden_size(self) -> int:
        return int(self.ff_expansion * self.d_model)
```

and make this line
`nn.Linear(cfg.d_model, cfg.ff_hidden_size)`

tango4j (changes requested):

examples/asr/conf/fastconformer/transformer_stacking_tdt_bpe.yaml
```diff
@@ -0,0 +1,184 @@
...
+    freq_masks: 2
+    time_masks: 10
+    freq_width: 27
+    time_width: 0.05
+
+  encoder:
+    _target_: nemo.collections.asr.modules.transformer_encoder.TransformerEncoder
+    feat_in: ${model.preprocessor.features}
+    d_model: 1280
+    n_heads: 16
```
tango4j, line 78: Need to add "ff_expansion"
and default 4
`ff_expansion: 4`

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+from torch.nn.attention.flex_attention import create_block_mask, flex_attention
+
+flex_attention_compiled = torch.compile(flex_attention, dynamic=True)
+
+
+@dataclass
+class TransformerEncoderConfig:
+    feat_in: int = 80
+    d_model: int = 512
+    n_heads: int = 8
```
tango4j, line 28: ff_expansion: float = 4 
(We sometimes test ff_expansion = 0.5 for MoE models, so we need this to be `float`)

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+    n_layers: int = 17
+    drop_rate: float = 0.1
+    qkv_bias: bool = False
+    qk_norm: bool = False
+    subsampling_factor: int = 4
+    # Attention mode — currently only "full" is supported.
+    # Future: "causal", "lookahead", "local", "sliding_window"
+    attn_mode: str = "full"
+
```
tango4j, line 39: Needs to add 
```
@property
    def ff_hidden_size(self) -> int:
        return int(self.ff_expansion * self.d_model)
```

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+    Architecture: FeatureStacking -> EmbedScale -> LayerNorm -> N x TransformerBlock -> FinalNorm
+
+    Uses PyTorch FlexAttention for attention computation. On CUDA, mask functions
+    are compiled into fused Triton kernels with block-sparse optimization. On CPU,
+    FlexAttention falls back to an unfused implementation automatically.
+
+    Args:
+        feat_in: Input feature dimension (number of mel bins).
+        d_model: Transformer hidden dimension.
+        n_heads: Number of attention heads.
```
tango4j, line 166: ff_expansion: float = 4.0 # Allow float

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+        qk_norm: Whether to apply per-head LayerNorm to Q and K before the dot product.
+        subsampling_factor: Frame stacking factor for the pre-encoder.
+        attn_mode: Attention pattern — currently only "full" (bidirectional) is supported.
+    """
+
+    def __init__(
+        self,
+        feat_in: int = 80,
+        d_model: int = 512,
+        n_heads: int = 8,
```
tango4j, line 184: ff_expansion: float = 4.0,

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,227 @@
...
+        attn_mode: str = "full",
+    ):
+        super().__init__()
+        if attn_mode != "full":
+            raise ValueError(f"attn_mode='{attn_mode}' is not yet supported. Currently only 'full' is available.")
+
+        cfg = TransformerEncoderConfig(
+            feat_in=feat_in,
+            d_model=d_model,
+            n_heads=n_heads,
```
tango4j, line 203: Add
`ff_expansion=ff_expansion,`

nemo/collections/asr/modules/transformer_encoder.py
tango4j: Since the layer norm is after this, this line has no effect.
If this is needed, needs to be after `x = self.embed_norm(x)`..?

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: This was done due to QK-Norm option to enable. But yes, I can unbind again for such ops.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: I haven;t validated runs with RMSNorm yet, so will keep this for future work.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: done

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Added to prevent NaNs during training, similar to the fix we applied in FastConformer during scaling. 

I need to check again to remove this if this has no effect after I added LayerNorm post Encoder for additional stability. Will revist

nithinraok: Added to prevent NaNs during training, similar to the fix we applied in FastConformer during scaling. 

I need to check again to remove this if this has no effect after I added LayerNorm post Encoder for additional stability. Will revist

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: duplicate comment

nithinraok: Thanks for suggestion. I will need to check by running experiments with this effect. So for now will defer to later PR.

tests/collections/asr/test_transformer_encoder.py
nithinraok: added

tests/collections/asr/test_transformer_encoder.py
nithinraok: good point. done

tests/collections/asr/test_transformer_encoder.py
nithinraok: In the general case where T_valid is not divisible, the stacking zero-pads based on the batch tensor length t, but the content beyond per-sample lengths is zeros in both the short and long tensors, so the stacked frames at valid positions are identical either way. 
The `block_mask` then excludes everything beyond lengths from attention.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: I donlt think adding `property` to dataclass is a good idea. however added the `ff_exansion` parameter

tango4j: Don't need to add property function. But wrapping the variable with int() is necessary.
And `ff_expansion` should be float.
Maybe one line of a variable sanity check on the "ffn_hidden_size" would make the code easy to track errors.

nithinraok: Yes added now, pls check

tango4j: Checked the change and `ff_hidden` variable. Looks good.

tests/collections/asr/test_transformer_encoder.py
pzelasko: Right, took me a while to convince myself that's true but I see it now.

nemo/collections/asr/modules/transformer_encoder.py
pzelasko: Where would you get NaN from if you have embed_norm just below? I don't get it.

nithinraok: So couple of things:
1. I was getting NaNs in training loss so applied bunch of techniques to see which could fix like embedding scaling, layer norm after last layer and padding after this model converged well.
2. Then I didn't went back and change each to see exactly which one contributed heavily for smoother training. I think Post LayerNorm but I haven't validated. 

I don't think this might have effect but I didn't check without it

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Please approve if all looks good from your end

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Good point. 

So I disabled it and re-ran for already trained ckpt and here are results:

with:
```
librispeech_test_clean_manifest: WER=0.0177
librispeech_test_other_manifest: WER=0.0359
```
commenting it out:
```
librispeech_test_clean_manifest: WER=0.0177
librispeech_test_other_manifest: WER=0.0359
```
You are correct, looks like pre-norm from attention is removing the added scale.

nithinraok: Good point. 

So I disabled it and re-ran for already trained ckpt and here are results:

with:
```
librispeech_test_clean_manifest: WER=0.0177
librispeech_test_other_manifest: WER=0.0359
```
commenting it out:
```
librispeech_test_clean_manifest: WER=0.0177
librispeech_test_other_manifest: WER=0.0359
```
You are correct, looks like pre-norm from attention is removing the added scale.

nithinraok: so I believe its the post Layer Norm that helped NaNs issue.

pzelasko (commented):
Thanks @nithinraok ! It looks good from my side, let's wait for @tango4j to approve and LGTM

tango4j:
This doesn't need to be happening in this PR but to ensure standardization, but later on, we probably need to verify this implementation is 100% compatible (transferable weights and identical results) with standard baselines like the Hugging Face Transformer.

If there are any custom scaling or structural tweaks, it would be great to make them optional flags. This keeps our baseline universally compatible while still allowing for optimizations when needed.

I think we should merge this after getting approval from @KunalDhawan and @stevehuang52 too.

nemo/collections/asr/modules/transformer_encoder.py
tango4j: Very nice. This was a concern for me for breaking the symmetry with other TF encoder implementations.

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+            nn.Dropout(cfg.drop_rate),
+            nn.Linear(ff_hidden, cfg.d_model),
+            nn.Dropout(cfg.drop_rate),
+        )
+
+    def forward(self, x):
+        return self.net(x)
+
+
+class MultiHeadAttention(nn.Module):
```
stevehuang52, line 108: Can we reuse or unify with the `MultiHeadAttention` in https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/submodules/multi_head_attention.py?

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+
+import torch
+import torch.nn as nn
+from torch.nn.attention.flex_attention import create_block_mask, flex_attention
```
KunalDhawan, line 19: We should add import guard for flex_attention so that people using older PyTorch (<2.5) don't run into unexpected import errors.

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+def _make_padding_mod(lengths):
+    """Mask out padding positions based on per-sample lengths."""
+
+    def pad_mask(b, h, q_idx, kv_idx):
+        return kv_idx < lengths[b]
+
+    return pad_mask
+
+
+class FeatureStacking(nn.Module):
```
KunalDhawan, line 50: `FeatureStacking` class has a lot of overlaps with `StackingSubsampling` in [nemo/collections/asr/parts/submodules/subsampling.py](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/submodules/subsampling.py) (i see a difference with length update where you uses individual sample length, not batch-padded length). Maybe it would be better to extend `StackingSubsampling` with the changes required and re-use the same?

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+        self.norm2 = nn.LayerNorm(cfg.d_model)
+        self.ffn = FeedForward(cfg)
+
+    def forward(self, x, block_mask=None):
+        x = x + self.drop(self.attn(self.norm1(x), block_mask=block_mask))
+        x = x + self.drop(self.ffn(self.norm2(x)))
+        return x
+
+
+class TransformerEncoder(nn.Module):
```
KunalDhawan, line 154: Here shouldn't TransformerEncoder inherit from NeuralModule and Exportable like other ASR encoders? This would enable `@typecheck`, avoid integration issues with `save_to`/`restore_from` and `from_pretrained` for which NeuralModule provides serialization hooks.

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+    """Pre-norm Transformer encoder for ASR.
+
+    Architecture: FeatureStacking -> EmbedScale -> LayerNorm -> N x TransformerBlock -> FinalNorm
```
KunalDhawan: Need to update docstring, EmbedScale was removed in commit `d277c3f`

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+        Returns:
+            x: (B, D, T') — encoded representation (channels-first).
+            length: (B,) — output lengths after subsampling.
+        """
+        x, length = self.pre_encode(audio_signal, length)
+
+        x = self.embed_norm(x)
+
+        B, T, _ = x.shape
+        block_mask = create_block_mask(_make_padding_mod(length), B=B, H=1, Q_LEN=T, KV_LEN=T, device=x.device)
```
KunalDhawan, line 234: Non blocker, but here we are recomputing `block_mask` in every forward pass. This might be wasted compute when length and T are unchanged across micro-batches, and especially for inference. Maybe worth caching this.

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def forward(self, x):
+        return self.net(x)
+
+
+class MultiHeadAttention(nn.Module):
+    def __init__(self, cfg: TransformerEncoderConfig):
+        super().__init__()
+        self.n_heads = cfg.n_heads
+        self.head_dim = cfg.d_model // cfg.n_heads
```
KunalDhawan, line 112: Maybe worth adding a warning here when `d_model % n_heads != 0`. This would lead to silent integer truncation if user picks incompatible values (e.g., d_model=512, n_heads=6 → head_dim=85, but qkv reshape will misalign)

KunalDhawan (changes requested):
Great work, Nithin! See minor comments below, other than that LGTM

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Good question :) 

The existing `MultiHeadAttention` uses a different attention backend (SDPA) with relative positional encoding. But here we are using `flex_attention` with compiled block masks. There are also legacy complexities we don't need like float16 autocast workarounds. 

I intentionally kept this separate to avoid coupling between architectures, which makes both harder to change independently. 

Going forward will be adding KV-Caching and other features to this one as well, so other architectures could use for best performance across.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Minimum version of pytorch for nemo is 2.6 and current stable pytorch version is 2.11, so I don't think we should add a import guard.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Good point. 

I think its better to keep this as base and update in that instead. However I feel the key difference in length computation makes it harder to merge them. 

Also `StackingSubsampling` has get_sampling_frames() and get_streaming_cache_size() for the Conformer streaming pipeline, which doesn;t apply here yet.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Good question, I wanted someone to ask this question :D 

`save_to` and `restore_from` would work based on Model Class (ex: EncDecCTCBPE) . From next release, we're hoping to be moving away from requiring `NeuralModule` for new  modules as we see little benefit for amount of code/work to support.

I agree with your point on Exportable, but here its tightly couple with NeuralModule for input and output types, which we need to come up with a better way for modules like this. CC: @pzelasko

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Updated.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: Good catch. In practice `T` and `length` change every batch during training so a cache wouldn't help. During inference the cost of `create_block_mask` is small relative to the attention layers themselves IMO.

nemo/collections/asr/modules/transformer_encoder.py
nithinraok: gooid point, rather than warning,  I think we should raise error. updated.

nemo/collections/asr/modules/transformer_encoder.py
```diff
@@ -0,0 +1,234 @@
...
+            qkv_bias=qkv_bias,
+            qk_norm=qk_norm,
+            ff_expansion=ff_expansion,
+            subsampling_factor=subsampling_factor,
+            attn_mode=attn_mode,
+        )
+        self.d_model = d_model
+
+        self.pre_encode = FeatureStacking(subsampling_factor, feat_in, d_model)
+        self.embed_norm = nn.LayerNorm(d_model)
```
tango4j: We should be adding this `pre_block_norm: bool = True` parameter

```        
self.embed_norm = nn.LayerNorm(d_model) if pre_block_norm else nn.Identity()
```

This input layer norm did not exist in the early TF implementations (Attention is all you need, Whisper etc) but the more modern implementations (BERT, ViT, etc) they have this input layer norm.

If we turn off by `pre_block_norm: bool = False`, then it achieves RMS error of 1.4 × 10⁻⁶ with the Whisper v3 implementations.

So I think by adding this `pre_block_norm: bool = True` parameter, we can make this TF encoder achieve parity with Whisper encoders or other old style Transformer encoders.

Of course, default should be `pre_block_norm: bool = True`.

This could be added as docstring: 
```
        pre_block_norm: If True (default), apply a LayerNorm to embeddings before the
            first transformer block (BERT/ViT-style). Set False to match standard
            pre-norm transformer encoders such as Whisper or GPT-2 — required when
            loading pretrained weights from those checkpoints.
```

tango4j (changes requested):
Request to add
`pre_block_norm: bool = True`
option for "embed_norm" line.

This can make TF encoder implementation achieve parity with old style (e.g., Whisper) TF encoder implementations, and weights are compatible with those models.

nemo/collections/asr/modules/transformer_encoder.py
tango4j: I have quickly did the test of with/without `embed_norm()`. 

## Headline parity result

NeMo `TransformerEncoder` output vs HuggingFace `WhisperEncoder` output, on
identical input mel, with identical Whisper-large-v3 weights mapped into both,
fp32, CUDA, batch=2, 30 s mel (3000 frames):

| Configuration | `max_abs` | `mean_abs` | **`rms`** | `rel_mean` | Verdict |
|---|---:|---:|---:|---:|---|
| `pre_block_norm=True`  *(PR's original behavior, pre-fix)* | **1.642 × 10⁺¹** | 2.220 × 10⁻¹ | **3.110 × 10⁻¹** | **5.449 × 10⁻¹** | **FAIL** — encoder output ≈ noise relative to reference |
| `pre_block_norm=False` *(new option, post-fix)*           | 3.471 × 10⁻⁴      | 9.966 × 10⁻⁷ | **1.419 × 10⁻⁶** | **2.446 × 10⁻⁶** | **PASS** — fp32 round-off floor |

Reference output magnitude: `mean_abs(ref) = 4.074 × 10⁻¹`.

**RMS-only side-by-side**

| `pre_block_norm` | RMS error vs Whisper reference |
|---|---:|
| `True`  | **3.110 × 10⁻¹** |
| `False` | **1.419 × 10⁻⁶** |

→ **~219,000× reduction in RMS error** from a single-line code change.

nemo/collections/asr/modules/transformer_encoder.py
tango4j: +1 on Nithin's comment. Reusing the MHA implementation makes us lose the point of building this implementation again.

nemo/collections/asr/modules/transformer_encoder.py
pzelasko: +1, this model definition is a testing ground for modern re-write / refactoring of NeMo Speech, it is OK to not re-use similar components to have an easier time developing.

I am thinking maybe we need to sprinkle some `@experimental` tags across the codebase to indicate new modules for which the APIs are likely to change until they are stabilized. This is a good candidate (we can follow up in a separate PR before release code cut off)

nemo/collections/asr/modules/transformer_encoder.py
pzelasko: I remember playing with flex_attention >1 year ago and back then it required `_compile=True` to be efficient. But since Nithin is reporting good throughput numbers with this code, they may have developed and optimized the internals more (especially since flex_attention is compiled with dynamic=True and has no issues which back then was also not true).

tango4j (approved):
All the issues that I had have been resolved.
@nithinraok I appreciate the contribution.

pzelasko (approved):

stevehuang52 (approved):

KunalDhawan (approved):

MahmoudAshraf97:
I feel this is a good place to discuss why is the nemo attention implementation still superior in training vs the torch sdpa? was there any investigation done on your side? I managed to reproduce @pzelasko findings at https://github.com/NVIDIA-NeMo/NeMo/pull/9590#issuecomment-2322364928 consistently, so we still use nemo attn implementation in training, but I think this is worth a revisit after two years
TIA

pzelasko:
@MahmoudAshraf97 I did not do a thorough investigation myself. I suspect it's related to activation norm growing too much in ConformerEncoder. Probably adding a targeted norm op somewhere could amend it. We're concentrating the effort on next-gen architectures, but if you can log the norms and add either re-scaling or extra norm to address, we'd be happy to accept a PR.

MahmoudAshraf97:
> @MahmoudAshraf97 I did not do a thorough investigation myself. I suspect it's related to activation norm growing too much in ConformerEncoder. Probably adding a targeted norm op somewhere could amend it. We're concentrating the effort on next-gen architectures, but if you can log the norms and add either re-scaling or extra norm to address, we'd be happy to accept a PR.

I did some investigation with a minimal fastconformer-ctc model with a single layer, I was able to reproduce the issue only when using RelPos with SDPA using any backend other than MATH.

Since regular MHA works well with all backends, I suspect that the issue is with the gradient flow of the positional bias in the fused kernels.

I suggest leaving RelPos as it, and implementing RoPe in conformer encoder, as per [this paper](https://arxiv.org/html/2501.06051v2), RoPe has the same baseline or better performance than RelPos and trains faster

I'm happy to implement it and open a PR if you are open to the idea @pzelasko

pzelasko:
Thanks @MahmoudAshraf97! That would be a welcome contribution.

PR #15666, opened 2026-05-06
Add unified prompt model support for multilingual ASR and streaming inference

enas-albasiri:
> [!IMPORTANT]  

# What does this PR do ?

Add unified prompt architecture for multilingual ASR, enabling language-ID conditioned for both Hybrid RNNT-CTC and RNNT models as well as streaming inference support.

**Collection**: ASR

# Changelog

- Add `EncDecRNNTBPEModelWithPrompt` — RNNT prompt-conditioned model with training script and 600M streaming config
- Add `LhotseSpeechToTextBpeDatasetWithPromptIndex` — index-based Lhotse dataset with per-dataset prompt mode support (`langID`/`auto`/`unified`) via lhotse `input_cfg` tags
- Add backward-compatible `forward()` in hybrid model accepting both legacy prompt tensors `[B, T, D]` and new prompt indices `[B]`
- Add `set_inference_prompt()` + `conformer_stream_step()` override for streaming inference in both model variants
- Add `target_lang` parameter to `speech_to_text_cache_aware_streaming_infer.py` (defaults to `"auto"` for prompt models)
- Add `strip_lang_tags` config in `RNNTDecodingConfig` to remove `<xx-XX>` locale tags from decoded output

# Usage

```python
# Training with per-dataset prompt modes via lhotse input_cfg:
# In your input_cfg yaml, tag datasets with prompt_mode:
#   - type: nemo_tarred
#     tags: { prompt_mode: langID }    # AST data — always pass lang ID
#   - type: nemo_tarred
#     tags: { prompt_mode: unified }   # ASR data — 50/50 ratio
#   - type: nemo_tarred
#     tags: { prompt_mode: auto }      # Code-switching — always auto

# Streaming inference with target language:
# python speech_to_text_cache_aware_streaming_infer.py \
#     model_path=model.nemo \
#     dataset_manifest=test.json \
#     target_lang=en-US

```

# GitHub Actions CI

**Pre checks**:

- [Y] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
  
**PR Type**:

- [Y] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
```diff
@@ -193,14 +195,191 @@ def initialize_prompt_feature(self):
...
+
+                if return_transcription:
+                    decoded_out = decoding.ctc_decoder_predictions_tensor(
+                        decoder_outputs=greedy_predictions_concat.unsqueeze(0),
+                        decoder_lengths=encoded_len[preds_idx : preds_idx + 1],
+                        return_hypotheses=False,
+                    )
+                    all_hyp_or_transcribed_texts.append(decoded_out[0])
+            best_hyp = None
+        else:
```
KunalDhawan: `log_probs` is not assigned in the RNNT branch, but is referenced in line 359. Calling `conformer_stream_step(..., return_log_probs=True)` while `cur_decoder == "rnnt"` will raise `NameError: name 'log_probs' is not defined`.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
```diff
@@ -1855,6 +1862,10 @@ class RNNTDecodingConfig:
     # config for multiblank decoding.
     big_blank_durations: Optional[List[int]] = field(default_factory=list)
 
+    # Strip language-ID tags (e.g. <en-US>) from decoded output.
+    # Enable for prompt-conditioned models that emit locale tags after punctuation.
+    strip_lang_tags: bool = True
```
KunalDhawan: This is a global default change that touches every existing RNNT model. The dataclass default (True) is also inconsistent with the runtime self.cfg.get('strip_lang_tags', False) (set in line 332 of this file), depending on whether cfg comes from the dataclass vs YAML, user will get different behavior.
Could you set the dataclass default to False and align both, then opt in only via the new YAML config. This is the safer backward-compatible choice.

nemo/collections/asr/models/rnnt_bpe_models_prompt.py
```diff
@@ -0,0 +1,784 @@
...
+        restore_path,
+        override_config_path=None,
+        map_location=None,
+        strict=True,
+        return_config=False,
+        save_restore_connector=None,
+        trainer=None,
+        validate_access_integrity=True,
+    ):
+        """Delegate to base EncDecRNNTBPEModel to avoid subclass substitution.
```
KunalDhawan, line 121: This `restore_from` workaround silently delegates to `EncDecRNNTBPEModel.restore_from`. As highlighted in your comment, restoring a checkpoint that was saved by `EncDecRNNTBPEModelWithPrompt` via this method will return a plain `EncDecRNNTBPEModel`, losing prompt_kernel, set_inference_prompt, and the streaming override. Please fix this.

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
```diff
@@ -0,0 +1,191 @@
...
+        return (
+            audio,  # Audio signal [B, T]
+            audio_lens,  # Audio lengths [B]
+            tokens,  # Text tokens [B, T]
+            token_lens,  # Token lengths [B]
+            prompt_indices,  # Language ID indices [B] - model creates full tensor
+        )
+
+
+class TokenizerWrapper:
```
KunalDhawan: This is similar to TokenizerWrapper in `NeMo/nemo/collections/common/tokenizers/aggregate_tokenizer.py`. Could you please upstream the desired changes to the exisiting TokenizerWrapper and import it here?

KunalDhawan (commented):
Thanks @enas-albasiri! Could you also share some numbers and validate accuracy of training monolingual/multlingual models with the unified-prompt approach?


Adding @pzelasko for review of `LhotseSpeechToTextBpeDatasetWithPromptIndex`

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
kingformatty: function `conformer_stream_step` is rewritten based on [link](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/mixins/mixins.py), where `log_probs` only applies for CTC model. RNNT shouldn't bother.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
kingformatty: Set strip_lang_tags to False for inference script.

nemo/collections/asr/models/rnnt_bpe_models_prompt.py
kingformatty: Losing "prompt_kernel" "set_inference_prompt" "streaming override" will only happen if the model is `EncDecRNNTBPEModel` class but loaded using `EncDecRNNTBPEModelWithPrompt.restore_from()`. And it's doesn't matter in this case. 

As long as the model is `EncDecRNNTBPEModelWithPrompt` class, it can be loaded correctly, either loaded using `EncDecRNNTBPEModel.restore_from()` or `EncDecRNNTBPEModelWithPrompt.restore_from()`. Or there is other cases where those prompt-specific attr will be lost when loading the prompt model?

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
kingformatty: Done imported `TokenizerWrapper` instead.

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
```diff
@@ -0,0 +1,164 @@
...
+from lhotse.dataset import AudioSamples
+from lhotse.dataset.collation import collate_vectors
+
+from nemo.collections.common.tokenizers.aggregate_tokenizer import TokenizerWrapper
+from nemo.collections.common.tokenizers.tokenizer_spec import TokenizerSpec
+from nemo.core.neural_types import AudioSignal, LabelsType, LengthsType, NeuralType
+from nemo.utils import logging
+
+
+class LhotseSpeechToTextBpeDatasetWithPromptIndex(torch.utils.data.Dataset):
```
pzelasko, line 34: Why is a separate class needed for this? Can we replace or add an option to the old class instead?

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
```diff
@@ -193,14 +195,191 @@ def initialize_prompt_feature(self):
...
+            dtype=torch.long,
+            device=encoded.device,
+        )
+        prompt.scatter_(2, idx.view(batch_size, 1, 1).expand(-1, time_steps, -1), 1.0)
+
+        out_dtype = encoded.dtype
+        encoded = self.prompt_kernel(torch.cat([encoded, prompt], dim=-1)).to(out_dtype)
+        return encoded.transpose(1, 2)  # (B, T, D) -> (B, D, T)
+
+    def conformer_stream_step(
```
pzelasko: Should it be `encoder_stream_step`? Why hardcode the architecture?

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
```diff
@@ -193,14 +195,191 @@ def initialize_prompt_feature(self):
         # setting the RNNT decoder as the default one
         self.cur_decoder = "rnnt"
 
+    # Streaming inference with language-ID prompt
```
pzelasko: Looks to me like the methods in this section could just as well be a Mixin class `PromptableRNNTMixin` to avoid duplication in hybrid/non-hybrid model definition

nemo/collections/asr/parts/submodules/rnnt_decoding.py
```diff
@@ -681,6 +685,11 @@ def __init__(self, decoding_cfg, decoder, joint, blank_id: int, supported_punctu
 
         # Update the joint fused batch size or disable it entirely if needed.
         self.update_joint_fused_batch_size()
+    
+    def set_strip_lang_tags(self, strip_lang_tags: bool):
```
pzelasko: this method should also accept the regex pattern, with the current one as default if set to None

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
```diff
@@ -0,0 +1,164 @@
...
+        self.prompt_dict = cfg.get('prompt_dictionary')
+        if not self.prompt_dict:
+            raise ValueError("prompt_dictionary is required in config")
+
+        self.num_prompts = cfg.get('num_prompts', 128)
+
+        # Field to use for prompt key (default to 'target_lang')
+        self.prompt_field = cfg.get('prompt_field', 'target_lang')
+
+        self.training_mode = cfg.get('training_mode', True)
```
pzelasko: Can we remove this and add `default_prompt_mode: "inference"` instead? One less option.

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    bucketing_strategy: "fully_randomized"
+    bucketing_batch_size: null
+    bucket_buffer_size: 10000
+    shuffle_buffer_size: 10000
+    prompt_field: target_lang
+    prompt_dictionary: ${model.model_defaults.prompt_dictionary}
+    num_prompts: ${model.model_defaults.num_prompts}
+    subsampling_factor: ${model.encoder.subsampling_factor}
+    lang_field: target_lang
+    training_mode: true
```
pzelasko: from the perspective of reading the config, `training_mode` is even more cryptical - let's remove as suggested in the other comment

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    'fr-CA': 100,
+    'mt-MT': 102,
+    'auto': 101
+    }
+
+  train_ds:
+    manifest_filepath: ???
+    sample_rate: ${model.sample_rate}
+    use_lhotse: true
+    shard_manifests: true
```
pzelasko: this option doesn't exist in lhotse dataloader

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    quadratic_duration: 15
+    num_buckets: 30
+    shuffle: true
+    num_workers: 8
+    pin_memory: true
+    max_duration: 39.99
+    min_duration: 0.1
+    is_tarred: true
+    tarred_audio_filepaths: null
+    shuffle_n: 2048
```
pzelasko: this option doesn't exist in lhotse dataloader

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    slice_length: 100
+    bucketing_strategy: "fully_randomized"
```
pzelasko: this option doesn't exist in lhotse dataloader

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    sample_rate: ${model.sample_rate}
+    batch_size: 2
+    shuffle: false
+    use_start_end_token: false
+    num_workers: 2
+    pin_memory: true
+    batch_duration: null
+    use_lhotse: true
+    use_bucketing: false
+    max_cuts: 8
```
pzelasko: this option doesn't exist in lhotse dataloader (use batch_size)

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
```diff
@@ -0,0 +1,395 @@
...
+    prompt_mode_field: prompt_mode
+    default_prompt_mode: unified
+    unified_auto_ratio: 0.5
+
+  validation_ds:
+    manifest_filepath: ???
+    sample_rate: ${model.sample_rate}
+    batch_size: 2
+    shuffle: false
+    use_start_end_token: false
```
pzelasko, line 188: this option doesn't exist in lhotse dataloader (unless you pass it onto the dataset class?)

pzelasko (changes requested):
Nice work!

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
kingformatty: There is an older version `nemo/collections/asr/data/audio_to_text_lhotse_prompt.py` creating full prompt_tensor from the dataloader, which is heavy-memory. This updated one only returns prompt indices with size [B] (B = batch_size) and model builds the prompt_tensor. This updated one will be used later on. Will work on to remove the older one.

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
kingformatty: Added a hook function in `nemo/collections/asr/parts/mixins/mixins.py`. Removed `conformer_stream_step` from both classes.

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: removed

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: removed

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: removed

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: removed

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
KunalDhawan: The upstream parallel here is what concerns me, in `ASRModuleMixin.conformer_stream_step` the `if return_log_probs: result.append(log_probs) `block sits inside the CTC branch where log_probs is defined. This PR moves the same block outside that branch, after the `if isinstance(self, EncDecCTCModel) or ...: ... else: ..`. split. With the current layout, anything that goes down the RNNT `else` branch (which is the default, `self.cur_decoder == "rnnt"`) and is called with `return_log_probs=True` will hit `NameError: name 'log_probs' is not defined` at line 359, it doesn't need a user to do anything unusual, just pass that flag on an RNNT path.
Three equivalent fixes, any one is fine imo:
1. Move the `if return_log_probs`: block back inside the CTC branch (matches upstream exactly).
2. Initialize l`og_probs = None` at the top of the function and let the RNNT path return `None` for log_probs.
3. `if return_log_probs and self.cur_decoder == "rnnt": raise NotImplementedError(...)` early in the function.

Same shape exists in the RNNT-only variant at `rnnt_bpe_models_prompt.py:247`, the parameter is in the signature but completely unused; either drop it from the signature or raise `NotImplementedError` so callers get a clear signal rather than a silently dropped flag.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
KunalDhawan: Thanks! `RNNTDecodingConfig.strip_lang_tags: bool = False` (line 1872) plus `self.cfg.get('strip_lang_tags', False) `at line 332 are now consistent, and the streaming-infer script exposes it as an explicit opt-in. Resolving this thread.
One follow-up I'd like to file separately rather than reopening here: the new `set_strip_lang_tags` method at `rnnt_decoding.py:689` only sets `self.lang_tag_pattern `and not `self.strip_lang_tags` (the flag the decode path actually reads at line 966), which is why the inference script has to assign both on lines 378–379. I'll leave that as a new inline comment so it doesn't clutter this thread.

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
kingformatty: The entire `conformer_stream_step` function is removed from both rnnt and hybrid prompt model class. As a replacement, I've added a hook function inside mixin.py which calls model-defined `_apply_prompt_to_encoded` using the existing `conformer_stream_step`.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
kingformatty: You are right. We will come to that point later for sure as it doesn't break the logic for now. There are also other places worth optimization and consolidation, too.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
KunalDhawan: Currently this method only sets `self.lang_tag_pattern` (and only when called with True). It does not touch `self.strip_lang_tags`, which is the flag the decode path actually reads in`decode_tokens_to_str_with_strip_punctuation` at line 966 (if self.strip_lang_tags: ...). That's why the streaming-infer script has to do this two-step at lines 378–379:
```
asr_model.decoding.strip_lang_tags = cfg.strip_lang_tags     # set the flag
asr_model.decoding.set_strip_lang_tags(cfg.strip_lang_tags)  # also compile the pattern
```

Could we either (a) have the method set both the flag and the pattern in one shot:
```
def set_strip_lang_tags(self, strip_lang_tags: bool):
    self.strip_lang_tags = strip_lang_tags
    if strip_lang_tags and not hasattr(self, 'lang_tag_pattern'):
        self.lang_tag_pattern = re.compile(r'\s*<[a-z]{2}-[A-Z]{2}>')
```

so the caller becomes `asr_model.decoding.set_strip_lang_tags(cfg.strip_lang_tags)`. Or (b) drop the method entirely, move the regex compile out of the `if` block in `__init__` (line 333) so it's always available, and let callers just assign the attribute? Either way, the regex literal at line 334 and line 692 is currently duplicated, would be good to lift it to a module-level constant.

kingformatty: Modularize set_strip_lang_tags. Now strip_lang related attribute only can be set from this function.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
kingformatty: Resolved in the other comment.

nemo/collections/asr/models/hybrid_rnnt_ctc_bpe_models_prompt.py
kingformatty: Create a class `PromptStreamingMixin` in mixin.py. Shared functions for prompt models (rnnt, hybrid) will be there to eliminate duplication.

nemo/collections/asr/parts/submodules/rnnt_decoding.py
kingformatty: Set `lang_tag_pattern` to be customizable. Declare at `set_string_lang_tags`. Default to None -> `<xx-XX>`

nemo/collections/asr/data/audio_to_text_lhotse_prompt_index.py
kingformatty: Remove `training_mode` option. Previously for test dataloader, `training_model = False` and it uses langID as prompt mode. Now at inference time, `default_prompt_mode` is set to `langID` to achieve the same setup.

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: removed

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: These are nemo dataloader config. In dataloader setup, if not lhotse, we still use `audio_to_text_dataset` dataloader. So I guess these should be better to preserve.

pzelasko: Do you have a non-Lhotse dataset/dataloader defined that supports training the promptable RNNT model? If not, it's better to remove all non-Lhotse options.

nemo/collections/asr/parts/mixins/mixins.py
```diff
@@ -897,3 +905,68 @@ def diarize(self, paths2audio_files: List[str], batch_size: int = 1) -> List[str
             Speaker labels
         """
         pass
+
+class PromptStreamingMixin:
```
pzelasko, line 910: `initialize_prompt_feature` should also be defined in this class. At the very least it should be a stub the pre-populates `self.concat`, `self.num_prompts`, and `self.prompt_kernel` to `None` so that it's known this is a part of the contract expected by the mixin. Ideally though, pull as much shared logic into it and call `super().initialize_prompt_feature()` in the overrides in child classes.

pzelasko (commented):
One more comment, thanks!

examples/asr/conf/fastconformer/cache_aware_streaming/fastconformer_transducer_bpe_streaming_prompt.yaml
kingformatty: Removed `use_start_end_token`

nemo/collections/asr/parts/mixins/mixins.py
kingformatty: Consolidate `initialize_prompt_feature` into mixin.py: `PromptStreamingMixin`. Defines `self.concat`, `self.num_prompts`, `self.prompt_kernel`.

However, I think definitions mentioned above are not only for streaming purpose, but also for offline model, I might change the mixin structure in the future.

kingformatty:
@pzelasko CICD tests are pushed and the PR passes pre-requested check. Could you please approve the CICD jobs?

KunalDhawan:
/ok to test [6e34ad1](https://github.com/NVIDIA-NeMo/NeMo/pull/15666/commits/6e34ad1d0f4831ab8ec783fa1ff243e033c433bb)

kingformatty:
/ok to test [8fde795](https://github.com/NVIDIA-NeMo/NeMo/pull/15666/commits/8fde795cf3e9eee42f3336719eef1cd6ad880d0b)

kingformatty:
@pzelasko CICD is passed. Could you please approve the merge?

pzelasko (approved):

KunalDhawan (approved):
Great work, thanks @kingformatty @enas-albasiri!

PR #15689, opened 2026-05-12
ASR: Use length recurrence for streaming pre-encode drop count

### LINKED ISSUE #15482: Problem with computing drop_extra_pre_encoded when varying pre_encode_cache_size for SubSampling and VGG frontends
**Problem Description**

In the current streaming encoder implementation, the frontends (ConvSubsampling) use two parameters: pre_encode_cache_size (the size of the left context in input frames) and drop_extra_pre_encoded (the number of tokens to discard after subsampling to remove the influence of this context).

For ConvSubsampling, drop_extra_pre_encoded is computed using the [formula](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/modules/conformer_encoder.py#L1006-L1019):
drop_extra_pre_encoded = 1 + (pre_encode_cache_size - 1) // subsampling_factor

However, this formula does not work for arbitrary choices of pre_encode_cache_size, because the length transformation in causal convolutions with stride=2 and kernel=3 is not a simple linear division. Instead, it follows a recurrent relation:
L1 = floor(L0 / 2) + 1
L2 = floor(L1 / 2) + 1
L3 = floor(L2 / 2) + 1   (for subsampling_factor=8, i.e., three layers)

**Example**
For subsampling_factor=8 (3 layers) and chunk_size=56:
1) If pre_encode_cache_size = 11, the actual drop = f(11) = 3, but the formula gives 1 + (11-1)//8 = 2.
2) If pre_encode_cache_size = 9, the actual drop = f(9) = 2, and the formula gives 2 (coincidental match).
Because of this, when pre_encode_cache_size takes different values, the wrong number of tokens may be discarded, leading to a mismatch between chunked inference and full pass.

**Other**:
-  Why was this particular value chosen for [pre_encode_cache_size](https://github.com/NVIDIA-NeMo/NeMo/blob/main/nemo/collections/asr/parts/submodules/subsampling.py#L382-L383), and is there a clear justification for it? It appears to diverge from the formula in the original paper (log(Dr) ∗ 2 + 1) and the reference NeMo implementation. 
Specifically, when we set an arbitrary value (e.g., pre_encode_cache_size = 11), our compare_vs_offline test (analogous to [test](https://github.com/NVIDIA-NeMo/NeMo/blob/main/examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py#L221)) fails. This indicates a mismatch between chunked and full inference. Looking at the NeMo source code (subsampling.py#L382-L383), we see that _max_cache_len for causal subsampling is set as: self._max_cache_len = subsampling_factor + 1

-  VGG frontend for streaming: After implementing causal VGG blocks (CausalMaskedConv2d, CausalMaskedMaxPool2d), the compare_vs_offline test also fails. Is the VGG frontend inherently unsuitable for streaming due to the presence of pooling layers that break equivalence between chunked and full inference without state caching? Are there any plans to support VGG in streaming mode, or are there known limitations that prevent its use? If it is not recommended, could you clarify why, and perhaps document this limitation?

1fanwang:
Closes #15482.

## What

The streaming encoder's `drop_extra_pre_encoded` count was computed as `1 + (cache_size - 1) // subsampling_factor`. For convolutional subsampling that formula is only accurate at the default `cache_size = subsampling_factor + 1`, because the actual `ConvSubsampling.forward` uses the convolutional length recurrence

```
L_next = floor((L + all_paddings - kernel_size) / stride) + 1
```

(or `ceil` under `_ceil_mode`), composed over `_sampling_num` layers.

For any other `pre_encode_cache_size`, the divisor approximation diverges from what `forward` produces, so streaming inference drops the wrong number of frames and the chunked output disagrees with a full pass, the mismatch reported in #15482.

This PR routes the drop count through a new `get_streaming_drop_size(cache_size)` on each subsampler:

- `ConvSubsampling.get_streaming_drop_size` uses the same `calc_length` helper the encoder already uses for the forward pass, so the streaming drop count stays consistent with the encoder's own length bookkeeping.
- `StackingSubsampling.get_streaming_drop_size` exposes the exact `cache_size // factor` relation.
- `ConformerEncoder.setup_streaming_params` calls the new method when available; for custom `pre_encode` modules that predate it, it falls back to the legacy formula (which coincides with the new one only at the default `cache_size`).

## Tests

`tests/collections/asr/test_asr_subsampling.py::TestStreamingDropExtraPreEncoded`:

- `test_drop_size_matches_forward`, parametrized over 4 subsampler shapes (`striding`/`dw_striding` × `subsampling_factor=4`/`8`) and 7 cache sizes (1, 4, 8, 9, 11, 16, 32). Each case runs the actual `forward` on a `cache_size`-long input and asserts the returned `out_lengths[0]` equals `get_streaming_drop_size(cache_size)`.
- `test_drop_size_legacy_formula_diverges_for_non_default_cache`, documents the bug: `subsampling_factor=8`, `cache_size=11` returns 2 under the old formula but the convolutional recurrence (and the actual forward) returns 3.
- `test_drop_size_zero_for_empty_cache`, `cache_size <= 0` → 0.
- `test_stacking_drop_size`, exact `cache_size // factor` for `StackingSubsampling`.

The new tests fail on main and pass with this PR; the legacy formula case demonstrates the divergence.

### DIFF
--- a/nemo/collections/asr/modules/conformer_encoder.py
+++ b/nemo/collections/asr/modules/conformer_encoder.py
@@ -56,6 +56,7 @@ from nemo.core.neural_types import (
 )
 from nemo.utils import logging
 
+
 __all__ = ['ConformerEncoder', 'ConformerMultiLayerFeatureExtractor']
 
 
@@ -1065,15 +1066,24 @@ class ConformerEncoder(NeuralModule, StreamingEncoder, Exportable, AccessMixin):
         else:
             streaming_cfg.pre_encode_cache_size = 0
 
+        # Number of subsampled output frames produced from the pre-encode left-context
+        # cache. This is what we drop after subsampling so chunked inference matches a
+        # full pass. For convolutional subsampling with stride > 1, this is NOT a simple
+        # floor division — see ``ConvSubsampling.get_streaming_drop_size``.
         if isinstance(streaming_cfg.pre_encode_cache_size, list):
-            if streaming_cfg.pre_encode_cache_size[1] >= 1:
-                streaming_cfg.drop_extra_pre_encoded = (
-                    1 + (streaming_cfg.pre_encode_cache_size[1] - 1) // self.subsampling_factor
-                )
-            else:
-                streaming_cfg.drop_extra_pre_encoded = 0
+            pre_encode_cache = streaming_cfg.pre_encode_cache_size[1]
         else:
-            streaming_cfg.drop_extra_pre_encoded = streaming_cfg.pre_encode_cache_size // self.subsampling_factor
+            pre_encode_cache = streaming_cfg.pre_encode_cache_size
+
+        if pre_encode_cache <= 0:
+            streaming_cfg.drop_extra_pre_encoded = 0
+        elif hasattr(self.pre_encode, "get_streaming_drop_size"):
+            streaming_cfg.drop_extra_pre_encoded = self.pre_encode.get_streaming_drop_size(pre_encode_cache)
+        else:
+            # Legacy fallback for custom pre_encode modules that pre-date
+            # ``get_streaming_drop_size``. Coincides with the convolutional recurrence at
+            # the default ``cache_size = subsampling_factor + 1`` but diverges otherwise.
+            streaming_cfg.drop_extra_pre_encoded = 1 + (pre_encode_cache - 1) // self.subsampling_factor
 
         for m in self.layers.modules():
             if hasattr(m, "_max_cache_len"):
--- a/nemo/collections/asr/parts/submodules/subsampling.py
+++ b/nemo/collections/asr/parts/submodules/subsampling.py
@@ -46,6 +46,18 @@ class StackingSubsampling(torch.nn.Module):
     def get_streaming_cache_size(self):
         return 0
 
+    def get_streaming_drop_size(self, cache_size: int) -> int:
+        """Number of subsampled output frames produced from `cache_size` input frames.
+
+        Used by streaming encoders to know how many leading frames of the encoder output
+        correspond to the pre-encode left-context cache, so they can be dropped after
+        subsampling. For StackingSubsampling the relation is exact: subsampling_factor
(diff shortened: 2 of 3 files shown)


nemo/collections/asr/modules/conformer_encoder.py
```diff
@@ -1065,15 +1066,24 @@ def setup_streaming_params(
...
+        if pre_encode_cache <= 0:
+            streaming_cfg.drop_extra_pre_encoded = 0
+        elif hasattr(self.pre_encode, "get_streaming_drop_size"):
+            streaming_cfg.drop_extra_pre_encoded = self.pre_encode.get_streaming_drop_size(pre_encode_cache)
         else:
-            streaming_cfg.drop_extra_pre_encoded = streaming_cfg.pre_encode_cache_size // self.subsampling_factor
+            # Legacy fallback for custom pre_encode modules that pre-date
+            # ``get_streaming_drop_size``. Coincides with the convolutional recurrence at
+            # the default ``cache_size = subsampling_factor + 1`` but diverges otherwise.
+            streaming_cfg.drop_extra_pre_encoded = 1 + (pre_encode_cache - 1) // self.subsampling_factor
```
nithinraok, line 1086: Could we preserve the previous fallback behavior for custom `pre_encode` modules that do not implement `get_streaming_drop_size()`?

Before this change, a list cache size used `1 + (cache_size - 1) // subsampling_factor`, but a scalar cache size used `cache_size // subsampling_factor`. The new fallback applies the first formula to both forms. For example, a scalar cache size of `11` with factor `8` changes from `1` to `2`, although the comment identifies this as a legacy fallback.

Please retain the original type-dependent formulas, or require custom modules to implement the new method. A unit test with a positive scalar cache size would protect this compatibility behavior.

nithinraok (commented):
I found one backward-compatibility concern in the fallback behavior.

PR #15696, opened 2026-05-13
[TTS] add max_duration = 20.0 to lhotse training yaml configs.

XuesongYang:
pretraining dataset has max duration 56 seconds that would occasionally hit OOM. Enforce 20 seconds.


XuesongYang:
@paarthneekhara do you agree with max duration 20 seconds for easy magpie?

XuesongYang:
/ok to test [1b91f04](https://github.com/NVIDIA-NeMo/NeMo/pull/15696/commits/1b91f04a6ce5d0d5606a5d729a7bc62ca8c1f9f1)

blisc (approved):

PR #15698, opened 2026-05-13
[TTS] Add comprehensive MoE dispatch unit tests with golden-value regression checks

XuesongYang:
Add behavioral and golden-value tests for PositionwiseConvFFMoE.forward():
- Determinism (same seed → identical output across re-init)
- Gradient flow (backward produces non-zero grads for router + experts)
- Edge cases: all-padding, single token, top_k == num_experts
- Bias / no-bias configurations
- Per-sample padding isolation
- Golden-value regression: expected output_sum, logits_sum, first expert indices, and output[0,0,:4] captured from the current sequential implementation under seed=42, verified with atol=1e-5

These tests lock in the numerical behaviour of the forward pass so that any future refactoring (e.g. batched dispatch) can be validated against the same golden values.

### DIFF
--- a/tests/collections/tts/modules/test_transformer_2501.py
+++ b/tests/collections/tts/modules/test_transformer_2501.py
@@ -1051,6 +1051,50 @@ class TestMoERouter:
 class TestPositionwiseConvFFMoE:
     """Test the PositionwiseConvFFMoE class."""
 
+    # Golden expected values captured from the original sequential implementation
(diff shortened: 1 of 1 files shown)


XuesongYang:
/ok to test [b912b96](https://github.com/NVIDIA-NeMo/NeMo/pull/15698/commits/b912b965cc939d5de935ddd89ef6683f89256ff7)

blisc (approved):

PR #15699, opened 2026-05-13
Add verbose flag to UTMOSv2Calculator

rlangman:
# What does this PR do ?

Allow user to disable verbose logging (progress bar) when computing utmosv2.

The utmosv2 model does not have a good interface for batch computing; you either provide it a single file path or a directory path containing .wav files. I have a use case where I cannot use the directory method and instead run the single file inference multiple times, for which it is helpful to have the progress bar disabled.

**Collection**: [TTS]

# Changelog

- Add verbose flag to constructor of UTMOSv2Calculator

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/modules/utmosv2.py
+++ b/nemo/collections/tts/modules/utmosv2.py
@@ -44,12 +44,13 @@ class UTMOSv2Calculator:
             Default is None.
     """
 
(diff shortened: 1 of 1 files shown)


rfejgin (approved):
Looks good to me.

PR #15714, opened 2026-05-20
Add RoPE to ConformerEncoder

MahmoudAshraf97:
> [!IMPORTANT]  

# What does this PR do ?

Adds RoPE as a positional encoding method to ConformerEncoder, it enables the use of torch SDPA while maintaining accuracy
This is a least-friction implementation, there is still path for more improvements (such as flex attention and fused qkv proj) but I preferred to not diverge from other conformer submodules
**Collection**: asr

## Motivation
Since #9590, SDPA caused loss explosion in training when Relative Positional Embeddings were used, investigation showed that SDPA had issues with gradients when attention additive biases were used

as per [this paper](https://arxiv.org/html/2501.06051v2), RoPe has the same baseline or better performance than RelPos and trains faster because of its ability to use SDPA

## Results

I tested two Fastconformer medium models, everything was constant except the positional encoding method (RoPE vs RelPos) and attention implementation (SDPA vs NeMo), both models were randomly initialized, although I forgot to fix the seed.

{276B1633-4755-4637-ADD3-72FEBD5E8F24}

One takeaway is that when the two models were initialized from a pretrained checkpoint except for the positional encoding parameters, RelPos converged faster, implying that RoPE attention geometry is different than RelPos and not necessarily transferable

{05E9B7C5-B8A0-41FA-977F-1AD7553A67B8}

## Improvements
The Rope was around 6% faster E2E, and used less memory which allowed for higher fused batch size for the RNNT loss and better compute utilization. The training time was dominated by RNNT loss calculation, so it diminishes any benefits from SDPA or `torch.compile`

RNNT loss compute time is data-dependent, that's why the step time is much lower at the beginning of the run when the model hasn't learned anything yet, once it starts to generate tokens, the time goes up

# Changelog

- Add RoPE as Positional Encoding Method for ConformerEncoder

# Usage

- pass `self_attention_model: rope` to encoder config

```yaml
model:
  encoder:
    self_attention_model: rope
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

cc @pzelasko 

# Additional Information

- Related to # (issue)


pzelasko:
/ok to test [44348a9](https://github.com/NVIDIA-NeMo/NeMo/pull/15714/commits/44348a9f5f3313ba43f59e92bda6c14a6ae0da6a)

pzelasko (commented):
Thanks for the PR. The overall direction looks reasonable to me, and I'm leaving this as a non-blocking review comment. I don't see the global defaults changing for existing configs (`self_attention_model` remains `rel_pos`, `use_pytorch_sdpa` remains `False`), but I found a few compatibility issues that should be addressed before merge.

1. **RoPE skips the existing pre-encoder scale/dropout contract**

   `nemo/collections/asr/modules/conformer_encoder.py`, around the new RoPE branch in `forward_internal`.

   Existing `rel_pos`/`abs_pos` paths call `self.pos_enc(x=audio_signal, ...)`, which applies `xscale` and `dropout_pre_encoder`. The new RoPE path sets `pos_emb = None` and leaves `audio_signal` untouched, while `RotaryPositionalEncoding` only rotates Q/K later in attention. As a result, `xscaling=True` and `dropout_pre_encoder` silently become no-ops for `self_attention_model='rope'`.

   I would keep the Q/K rotation inside `RoPEMultiHeadAttention`, but preserve the existing encoder input transform semantics by applying the same input scale/dropout once before the layer stack.

2. **The public `change_attention_model()` wrapper can corrupt the saved config**

   `nemo/collections/asr/parts/mixins/mixins.py`, in the `update_config` block.

   The encoder resolves `None` values to the current attention model/context, but the mixin writes the raw arguments back into `self.cfg.encoder`. For example, calling `change_attention_model(rotary_fraction=0.5)` on an existing RoPE model can leave `cfg.encoder.self_attention_model = None` and `cfg.encoder.att_context_size = None`, and it skips copying the updated RoPE params because `self_attention_model != 'rope'`.

   The top-level config should mirror the resolved values from `self.encoder` after the call, e.g. `self.encoder.self_attention_model`, `self.encoder.att_context_size`, and the resolved encoder RoPE fields.

3. **`change_attention_model('rope')` does not preserve `use_bias=False`**

   `nemo/collections/asr/modules/conformer_encoder.py`, in the RoPE `new_attn = RoPEMultiHeadAttention(...)` conversion path.

   The constructor path forwards `use_bias` into `ConformerLayer`/`RoPEMultiHeadAttention`, but the conversion helper creates RoPE attention without `use_bias`, so it defaults to `True`. A bias-free model converted to RoPE silently gains randomly initialized bias parameters and no longer matches its config.

   Please pass through `getattr(self._cfg, 'use_bias', True)` or an equivalent stored value, and add a regression test for converting a `use_bias=False` encoder to RoPE.

MahmoudAshraf97:
@pzelasko I implemented fixes for all three issues mentioned
1. pre-encoder dropout is now applied separately along with xscaling
2. updated the wrapper to use the current model parameters after updating them instead of using the function args
3. updated all attention modules because this bug existed in all of them, not just RoPE

as for changing the defaults, I can do that, do I have to also change `ConformerEncoder` init args or just the configs? although I'd prefer to do that in a followup PR

note, the `time` variable was removed because it causes the CI tests to fail, but it's unrelated to the PR

pzelasko:
> as for changing the defaults, I can do that, do I have to also change ConformerEncoder init args or just the configs? although I'd prefer to do that in a followup PR

Not sure what do you mean by "changing the defaults"? Let's keep the defaults as they were for backward compatibility.

ipmedenn (commented):
@MahmoudAshraf97 Thanks for the great contribution!
I tried your RoPE implementation for speaker diarization training, and I got very similar results to those you reported for ASR.

pzelasko:
/ok to test [ad45a90](https://github.com/NVIDIA-NeMo/NeMo/pull/15714/commits/ad45a9054de868deb5349224af681147e28e77df)

pzelasko (dismissed):

nithinraok (dismissed):
LGTM! Thanks for the great contribution @MahmoudAshraf97

pzelasko:
/ok to test [7fdf842](https://github.com/NVIDIA-NeMo/NeMo/pull/15714/commits/7fdf842a00edd076c7a0863aabc8b4098edf5a01)

pzelasko:
I fixed the linter issues, will monitor the PR until it's merged.

nemo/core/utils/cuda_python_utils.py
```diff
@@ -216,17 +217,28 @@ def with_conditional_node(while_loop_kernel, while_loop_args, while_loop_conditi
...
+        cuda.cuLaunchKernel(
+            while_loop_kernel, 1, 1, 1, 1, 1, 1, 0, body_stream.cuda_stream, while_loop_args.ctypes.data, 0
+        )
 
-    torch.cuda.set_stream(previous_stream)
+        end_capture_out = cudart.cudaStreamEndCapture(body_stream.cuda_stream)
+        body_capture_active = False
+        cu_call(end_capture_out)
+    finally:
+        if body_capture_active:
```
artbataev, line 234: In general looks fine.
If I correctly understand, this could be a bit cleaner without `body_capture_active`
```python
try:
   torch.cuda.set_stream(body_stream)
   yield body_stream, body_graph
   cuda.cuLaunchKernel(...)
finally:
    try:
        end_capture_out = cudart.cudaStreamEndCapture(body_stream.cuda_stream)
        cu_call(end_capture_out)
     except Exception as e:
         raise NeMoCUDAPythonException(...) from e
```

Also, `NeMoCUDAPythonException` can be caught in the caller

MahmoudAshraf97:
Please let me know if there anything I can do to address the CI failures @pzelasko

MahmoudAshraf97:
Hi @pzelasko, we have multiple training jobs that are waiting on this, any chance to merge it soon?

pzelasko:
/ok to test [a67d345](https://github.com/NVIDIA-NeMo/NeMo/pull/15714/commits/a67d3457cd8cd1625868fe400134b9aa84609d81)

MahmoudAshraf97:
Please let me know if there anything I can do to address the CI failures, is there a way to run them locally?

stevehuang52:
/ok to test [d21833d](https://github.com/NVIDIA-NeMo/Speech/pull/15714/commits/d21833dfbba18a1364e7cd18a6afe2aa74cdbcc2)

nithinraok:
Both failing CI tests are related to docs which can be ignored. Lets wait for others to finish and we should be good to merge.

nithinraok:
Merging as failing docs are not related to this PR

PR #15732, opened 2026-05-27
Fix LazyNeMoIterator for non-tarred S3 audio data

melllinia:
## Summary

Fixes two issues that prevent `LazyNeMoIterator` from reading NeMo manifests referencing audio on object stores (s3://, ais://) in non-tarred format.

### Issue 1: `get_full_path` joins S3 URIs with manifest directory

`os.path.isabs("s3://bucket/file.opus")` returns `False`, causing `get_full_path` to treat S3 URIs as relative paths and join with the manifest directory:

```
Input:  s3://audio-riva-originals/youtube/audios/file.opus
Result: s3:/manifest_dir/s3:/audio-riva-originals/youtube/audios/file.opus  (broken)
```

**Fix:** Check `is_datastore_path(audio_file)` at the top of `get_full_path` before any relative-path logic.

### Issue 2: `LazyNeMoIterator` only pops `"sampling_rate"`, ignoring `"sample_rate"`

Many datasets use `"sample_rate"` (e.g., ssl_yt_harvested). When the field is missing, NeMo falls back to `Recording.from_file()` which calls `soundfile.info()` on an S3 URI — which fails.

**Fix:** `data.pop("sampling_rate", None) or data.pop("sample_rate", None)`

## How to reproduce

```python
# Manifest with: {"audio_filepath": "s3://bucket/file.opus", "duration": 100, "sample_rate": 48000}
from lhotse import CutSet
from nemo.collections.common.data.lhotse.nemo_adapters import LazyNeMoIterator

cutset = CutSet(LazyNeMoIterator("manifest.jsonl"))
for cut in cutset:  # Fails without fix
    audio = cut.load_audio()
```

## When does this happen?

- Only affects **non-tarred** data (`type: nemo`) with S3/AIS audio paths
- Tarred data (`type: nemo_tarred`) is unaffected because `LazyNeMoTarredIterator` reads audio bytes directly from tar
- Tested with ssl_yt_harvested dataset (opus files on PDX S3)

## Files changed

- `nemo/collections/common/parts/preprocessing/manifest.py` — early return for datastore URIs
- `nemo/collections/common/data/lhotse/nemo_adapters.py` — sample_rate fallback

## Test plan

- [x] Verified `get_full_path` returns S3 URIs unchanged
- [x] Verified relative/absolute local paths still work correctly
- [x] End-to-end test: loaded 3 opus files from `s3://audio-riva-originals` via `LazyNeMoIterator`


pzelasko (dismissed):

pzelasko:
/ok to test [eec9e20](https://github.com/NVIDIA-NeMo/NeMo/pull/15732/commits/eec9e20c61b0206dd58480258c98cd4680be719b)

nithinraok (approved):
Thanks!

PR #15737, opened 2026-05-29
Build Automodel compiled dependencies in CI image

pzelasko:
## Summary
- Add a compiled-deps wheel stage to docker/Dockerfile.ci for NeMo Automodel packages.
- Build and install TransformerEngine, DeepEP V1, nv-grouped-gemm, causal-conv1d, mamba-ssm, and flash-attn into CI Dockerfile images.
- Add GPU_TARGET with default h100plus and a100 option; A100 applies a DeepEP V1 patch to disable NVSHMEM.
- Keep CI resource settings and Docker runtime resource limits unchanged.

## Validation
- bash -n over Dockerfile heredoc snippets
- git diff --check
- docker buildx build --call=check -f docker/Dockerfile.ci .
- Local constrained builds/import smoke tests were run before the final review cleanup for both GPU_TARGET=h100plus and GPU_TARGET=a100.


pzelasko:
/ok to test [14c7818](https://github.com/NVIDIA-NeMo/NeMo/pull/15737/commits/14c7818e75be957a09d33db5edf12e6a19d799ce)

docker/Dockerfile.ci
```diff
@@ -54,6 +54,206 @@ RUN <<"EOF" bash -ex
...
+    --no-build-isolation \
+    --no-binary=mamba-ssm \
+    --no-deps \
+    --wheel-dir /opt/automodel-wheels \
+    "mamba-ssm==${MAMBA_SSM_VERSION}"
+
+MAX_JOBS="${FLASH_ATTN_MAX_JOBS}" \
+NVCC_THREADS="${FLASH_ATTN_NVCC_THREADS}" \
+FLASH_ATTENTION_FORCE_BUILD=TRUE \
+    python -m pip wheel \
```
chtruong814: Why do we need to build flash-attn?  Is the available wheel not compatible?

chtruong814 (commented):
We can remove a lot of these to the pyproject.toml and have uv build during install.  RL does something similar.
https://github.com/NVIDIA-NeMo/RL/blob/main/pyproject.toml#L111

Would you like my help to move more of this to within the uv and pyproject toml?  Or is there a reason you prefer to keep it separate from there?

blisc (dismissed):
Docker builds on my end. Did a quick sanity check and it runs. Still need to test with inference/training but good to go from my end.

docker/Dockerfile.ci
pzelasko: flash attention doesn't have precompiled wheel for our cuda + python + pytorch combination selected for the container

pzelasko:
/ok to test [c40ca1a](https://github.com/NVIDIA-NeMo/NeMo/pull/15737/commits/c40ca1a75e626c99576f376dee115a12c3bac919)

chtruong814 (dismissed):
Yeah I guess if we have conditional patches for what has to build from source, probably need to do it the way you have.   Can take a look later to see if we can push anymore into the pyproject and uv lock.  But this shoudl be fien as well.

pzelasko:
/ok to build [16710b8](https://github.com/NVIDIA-NeMo/NeMo/pull/15737/commits/16710b87341a8ebbf516d9cdd5f3d879af51f2a3)

pzelasko:
/ok to test [16710b8](https://github.com/NVIDIA-NeMo/NeMo/pull/15737/commits/16710b87341a8ebbf516d9cdd5f3d879af51f2a3)

pzelasko:
/ok to test [87e557e](https://github.com/NVIDIA-NeMo/NeMo/pull/15737/commits/87e557e03f8099256c67f03b86c4fc4a7f6dcfbd)

blisc (approved):

PR #15741, opened 2026-06-01
[Superseded by #15931] EasyMP vLLM-Omni model definition

vklimkov-nvidia:
EasyMP model defnition, where backbone and LT are compiled into a single cuda graph for uniform batches.
Loads real weights, doesn't produce valid acoustic tokens at this point.


blisc:
I'm assuming that this is superseded by #15931.

vklimkov-nvidia:
my bad, it is indeed. was used to share the code for voicechat.

PR #15742, opened 2026-06-01
[Audio Codec] Lhotse data loading updates and fixes

rfejgin:
1. Move the random-segment-selection functionality from Lhotse to our dataset class, `AudioCodecLhotseDataset`. The corresponding built-in Lhotse functionality (`truncate_duration`) operates on the parent recording, which is not what we want.

2. Switch from `batch_duration` to `batch_size` for specifying the training batch size. In our setting, they are equivalent since the item size is fixed for all batch items, and it's clearer this way, now that segment selection is happening in the Dataset class.


nemo/collections/tts/data/audio_codec_dataset_lhotse.py
```diff
@@ -28,33 +29,64 @@ class AudioCodecLhotseDataset(torch.utils.data.Dataset):
...
+        segment_duration: float,
         sanity_check_audio: bool = False,
-        min_samples_for_sanity: Optional[int] = None,
     ):
         """
         Args:
             sample_rate: The sample rate to resample the audio to.
+            segment_duration: Length of each training segment in seconds. A random
+                segment of this length is taken from each cut's `target_audio` field
+                (not from the parent `recording`, which may span a much longer duration).
```
blisc, line 53: Are they shars that have a non-empty recording field?

blisc (commented):
Looks good from my end. Please get final approval from Ryan

nemo/collections/tts/data/audio_codec_dataset_lhotse.py
rfejgin: From examining the cuts, they all seem to have a `recording` field, at least in two datasets I checked. What varies is whether the recording is the target audio itself or rather a longer recording of which the target audio is a subset (which is only true for certain datasets, but it's something we need to support).

rlangman (approved):
LGTM

PR #15763, opened 2026-06-07
True-bfloat16 inference for cache aware pipelines

naymaraq:
> [!IMPORTANT]  

# What does this PR do ?

Fix true `bfloat16` (use_amp=false) inference for cache-aware streaming ASR. Observed significant RTFx improvment and 2x cache compression.

**Collection**: [ASR]

# Changelog

**Problem:** With `use_amp`: false, the cache-aware wrappers cast the model weights to `bfloat16`, but disabled autocast, while the input mel features and the encoder caches stayed `float32`. The encoder then received `float32` inputs/caches against `bfloat16` weights with no autocast to reconcile them, raising a **dtype-mismatch error.**

- Cast input features: In both `CacheAwareRNNTInferenceWrapper.stream_step` and `CacheAwareCTCInferenceWrapper.stream_step`,  cast `processed_signal` to `self.cast_dtype` after the device move.

- Align cache dtype: `CacheAwareASRInferenceWrapper.get_initial_cache_state` now passes `dtype=self.cast_dtype`, so the context manager's persistent cache storage matches the encoder's output caches.

- Config defaults: Set **use_amp: false** in `cache_aware_rnnt.yaml` and `cache_aware_ctc.yaml` so the example configs run true bf16 by default.

# Results

Setup:

- _num_slots=256_ and _batch_size=64_
- _Experiments done on NVIDIA RTX 5000 Ada GPU_

Key Findings:

- **True-bf16** inference reduced cache memory 2x times versus fp32 (936 MB vs. 1872 MB) with essentially no WER degradation across all attention context sizes.
- **True-bf16** inference also significantly improved RTFx across all attention context sizes.

| Method | AMP | Att Context Size | Comp Ratio | LS CLEAN | LS OTHER | TED | VOX | AVG. | RTFX (LS-OTHER) |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| bf16 | TRUE | [70, 13] | x2 | 5.28% | 8.26% | 12.03% | 11.13% | 9.18% | 407 |
| bf16 | FALSE | [70, 13] | x2 | 5.29% | 8.26% | 12.01% | 11.15% | 9.18% | 499 |
| bf16 | TRUE | [70, 6] | x2 | 5.38% | 8.49% | 12.02% | 11.23% | 9.28% | 264 |
| bf16 | FALSE | [70, 6] | x2 | 5.37% | 8.48% | 12.00% | 11.21% | 9.26% | 338 |
| bf16 | TRUE | [70, 1] | x2 | 5.57% | 8.91% | 12.30% | 11.56% | 9.58% | 107 |
| bf16 | FALSE | [70, 1] | x2 | 5.56% | 8.91% | 12.30% | 11.51% | 9.57% | 136 |
| bf16 | TRUE | [70, 0] | x2 | 6.06% | 9.80% | 12.61% | 12.70% | 10.29% | 63 |
| bf16 | FALSE | [70, 0] | x2 | 6.08% | 9.79% | 12.66% | 12.67% | 10.30% | 80 |

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)

### DIFF
--- a/examples/asr/conf/asr_streaming_inference/cache_aware_ctc.yaml
+++ b/examples/asr/conf/asr_streaming_inference/cache_aware_ctc.yaml
@@ -6,7 +6,7 @@ asr:
   device: cuda                                                          # Device for inference: 'cuda' or 'cpu'
   device_id: 0                                                          # GPU device ID
   compute_dtype: bfloat16                                               # Compute precision: 'bfloat16' for Ampere+, 'float16' for older GPUs, or 'float32'
-  use_amp: true                                                         # Enable Automatic Mixed Precision
+  use_amp: false                                                         # Enable Automatic Mixed Precision
 
 
 # ==========================================
--- a/examples/asr/conf/asr_streaming_inference/cache_aware_rnnt.yaml
+++ b/examples/asr/conf/asr_streaming_inference/cache_aware_rnnt.yaml
@@ -6,7 +6,7 @@ asr:
   device: cuda                                                          # Device for inference: 'cuda' or 'cpu'
   device_id: 0                                                          # GPU device ID
   compute_dtype: bfloat16                                               # Compute precision: 'bfloat16' for Ampere+, 'float16' for older GPUs, or 'float32'               
-  use_amp: true                                                         # Enable Automatic Mixed Precision
+  use_amp: false                                                         # Enable Automatic Mixed Precision
   decoding:
     strategy: "greedy_batch"
     preserve_alignments: false
--- a/nemo/collections/asr/inference/model_wrappers/cache_aware_asr_inference_wrapper.py
+++ b/nemo/collections/asr/inference/model_wrappers/cache_aware_asr_inference_wrapper.py
@@ -55,7 +55,7 @@ class CacheAwareASRInferenceWrapper(ASRInferenceWrapper):
         Returns:
             (tuple[Tensor, Tensor, Tensor]) the initial cache state of the encoder.
         """
-        return self.asr_model.encoder.get_initial_cache_state(batch_size=batch_size)
+        return self.asr_model.encoder.get_initial_cache_state(batch_size=batch_size, dtype=self.cast_dtype)
 
     def get_drop_extra_pre_encoded(self) -> int:
         """
--- a/nemo/collections/asr/inference/model_wrappers/cache_aware_ctc_inference_wrapper.py
+++ b/nemo/collections/asr/inference/model_wrappers/cache_aware_ctc_inference_wrapper.py
@@ -57,6 +57,9 @@ class CacheAwareCTCInferenceWrapper(CacheAwareASRInferenceWrapper):
 
         self.drop_extra_pre_encoded = self.get_drop_extra_pre_encoded()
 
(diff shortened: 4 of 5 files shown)


MahmoudAshraf97:
If we are looking for RTF improvements, maybe we should move the caching step after the qkv proj, this is a computation that is being repeated every step on cached inputs, the cost will be double the memory needed for cache but that is already affordable since the cache is capped at a certain size

artbataev (approved):
Thanks a lot! Great improvement!

naymaraq:
> If we are looking for RTF improvements, maybe we should move the caching step after the qkv proj, this is a computation that is being repeated every step on cached inputs, the cost will be double the memory needed for cache but that is already affordable since the cache is capped at a certain size

Good point. We need to understand what kind of RTFx gain it will provide at the cost of doubling the memory usage.

MahmoudAshraf97:
> > If we are looking for RTF improvements, maybe we should move the caching step after the qkv proj, this is a computation that is being repeated every step on cached inputs, the cost will be double the memory needed for cache but that is already affordable since the cache is capped at a certain size
> 
> Good point. We need to understand what kind of RTFx gain it will provide at the cost of doubling the memory usage.

I did a quick implementation to test the gains, the following graph measure a single forward pass duration in ms

image

as you can see, the gains are much larger with smaller chunk sizes as the projection compute is the largest percentage of the total compute

note that for models that use longformer attention, the memory usage will quadruple since they have two projections, but I don't think there are any cache-aware models that use it

A downside of this is that the inference is now memory bandwidth bottlenecked, although `torch.compile` is showing good speedups, but TensorRT showed 10% lower throughput with fp32 cache and 10% higher throughput with fp16 cache, which makes memory bandwidth an important factor

naymaraq:
> > > If we are looking for RTF improvements, maybe we should move the caching step after the qkv proj, this is a computation that is being repeated every step on cached inputs, the cost will be double the memory needed for cache but that is already affordable since the cache is capped at a certain size
> > 
> > 
> > Good point. We need to understand what kind of RTFx gain it will provide at the cost of doubling the memory usage.
> 
> I did a quick implementation to test the gains, the following graph measure a single forward pass duration in ms
> 
> image
> as you can see, the gains are much larger with smaller chunk sizes as the projection compute is the largest percentage of the total compute
> 
> note that for models that use longformer attention, the memory usage will quadruple since they have two projections, but I don't think there are any cache-aware models that use it
> 
> A downside of this is that the inference is now memory bandwidth bottlenecked, although `torch.compile` is showing good speedups, but TensorRT showed 10% lower throughput with fp32 cache and 10% higher throughput with fp16 cache, which makes memory bandwidth an important factor

Great work, did you get the same WER?

MahmoudAshraf97:
> Great work, did you get the same WER?

Yes the outputs are the 1.0 cosine similarity, ans diff is 2×10^-3 since there might be some difference in GEMM accumulation 

I can open a PR if you are willing to review it

PR #15764, opened 2026-06-08
Implement Multiturn support for Nemotron-TTS

Edresson:
# What does this PR do ?

Implements multiturn support for EasyMagpieTTS, including multiturn Lhotse training/evaluation data handling, user-audio-conditioned inference, multiturn metrics/export support, and related tests.

**Collection**: TTS

# Changelog

- Added a new EasyMagpie multiturn Lhotse config: `examples/tts/conf/magpietts/easy_magpietts_lhotse_multiturn.yaml`.
- Added support for restoring EasyMagpie models from a pretrained checkpoint before normal initialization.
- Added multiturn Lhotse TTS dataset support for duplex / multi-turn conversational samples.
- Added EasyMagpie multiturn user-audio inference mode via `--easy_magpie_inference_mode multiturn_user_audio`.
- Added support for user-audio conditioning, user silence handling, user turn end-token handling, and max evaluation turns during multiturn inference.
- Added distributed / multi-GPU inference support for multiturn EasyMagpie evaluation by sharding generation across ranks and merging rank-local outputs for evaluation.
- Added grouped turn-level and sample-level metrics export for multiturn evaluation, including JSON/CSV outputs.
- Added phoneme prediction export fields to generated `.json` / `.csv` evaluation outputs.
- Added optional predicted-phoneme inference controls, including phoneme input type, phoneme sampling method, tokenizer override, and text-input dropout.
- Added additional evaluation options for ASR / EOU batch sizes, emotion cosine similarity, emotion match rate, and text-annotation stripping for metrics.
- Added Lhotse dataloader speaker filtering support.
- Added / updated Nemotron-H decoder support needed for prefill in EasyMagpie multiturn inference.
- Added unit tests for Lhotse TTS filters, MagpieTTS Lhotse dataset behavior, and EasyMagpieTTS model behavior.

# Usage

Example EasyMagpie multiturn inference with a `.nemo` checkpoint:

```python
# Run from the NeMo repository root.
# Replace paths with your local model, codec, evalset config, and output directory.

python examples/tts/magpietts_inference.py \
  --model_type easy_magpie \
  --easy_magpie_inference_mode multiturn_user_audio \
  --nemo_files /path/to/easy_magpie_model.nemo \
  --datasets_json_path /path/to/evalset_config.json \
  --out_dir /path/to/output \
  --codecmodel_path /path/to/codec_model.nemo \
  --batch_size 1 \
  --max_eval_turns 6 \
  --run_evaluation

### CODE REVIEW

File: nemo/collections/tts/modules/nemotron_h_decoder.py
Comment (Edresson) [inline nemo/collections/tts/modules/nemotron_h_decoder.py:1]: FYI, these changes was done to support prefill in one step on nemotron_h class.

File: examples/tts/magpietts_inference.py [hunk 1]
```diff
@@ -658,12 +1120,33 @@ def _build_easy_magpie_config(args) -> EasyMagpieInferenceConfig:
...
     )
+    if cfg_cls is EasyMagpieMultiturnUserAudioInferenceConfig:
+        kwargs.update(
+            max_eval_turns=args.max_eval_turns,
+            save_debug_multiturn_audio=not args.no_save_debug_multiturn_audio,
+        )
+    return cfg_cls(**kwargs)
+
+
+def _select_runner_cls(args):
```
shehzeen: @blisc

examples/tts/magpietts_inference.py
```diff
@@ -141,6 +146,244 @@ def append_metrics_to_csv(csv_path: str, checkpoint_name: str, dataset: str, met
     logging.info(f"Metrics appended to: {csv_path}")
 
 
+def _mean_finite(values: list):
```
blisc: This file is getting quite large. Let's start by moving out all the helper functions into NeMo itself and see how large this file is.

nemo/collections/common/data/lhotse/dataloader.py
```diff
@@ -256,6 +259,21 @@ class LhotseDataLoadingConfig:
     slice_length: Optional[int] = None
 
 
+def resolve_excluded_speaker_ids(excluded_speaker_ids):
```
blisc, line 262: Can you docstring this function?

nemo/collections/tts/data/text_to_speech_dataset_lhotse_multiturn.py
```diff
@@ -0,0 +1,943 @@
...
+from nemo.collections.tts.parts.utils.tts_dataset_utils import (
+    beta_binomial_prior_distribution,
+    normalize_volume,
+    stack_tensors,
+)
+from nemo.core.classes.common import safe_instantiate
+from nemo.utils import logging
+
+
+def setup_tokenizers(all_tokenizers_config, mode='train'):
```
blisc: Is this the same as another function from somewhere else? Can we import instead of copy?

nemo/collections/tts/data/text_to_speech_dataset_lhotse_multiturn.py
```diff
@@ -0,0 +1,943 @@
...
+        self.frame_length = (
+            self.codec_model_samples_per_frame / codec_model_input_sample_rate
+        ) * frame_stacking_factor
+
+    def get_num_audio_samples_to_slice(self, duration, sample_rate):
+        num_codec_frames = int(duration * sample_rate / self.codec_model_samples_per_frame)
+        num_audio_samples = num_codec_frames * self.codec_model_samples_per_frame
+        return num_audio_samples
+
+    def __getitem__(self, cuts: CutSet) -> Dict[str, Union[torch.Tensor, List]]:
```
blisc: Can we split this function into smaller and simpler subfunctions?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -523,7 +595,10 @@ def prepare_audio_channel_embeddings(
         audio_codes: torch.Tensor,
         audio_codes_lens: torch.Tensor,
         delay: torch.Tensor,
-    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        speech_eos_mask: Optional[torch.Tensor] = None,
+        agent_mask: Optional[torch.Tensor] = None,
+        current_streaming_speech_delay: Optional[int] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, Optional[torch.Tensor]]:
```
blisc, line 627: Can you update the docstring for the changes in the arguments?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -1393,34 +1923,66 @@ def get_dataset(self, dataset_cfg, dataset_type):
...
+                add_language_to_context_text=self.add_language_to_context_text,
+                source_sample_rate=self.sample_rate,
+                input_roles=["user", "User"],
+                output_roles=["assistant", "Assistant", "agent", "Agent"],
+                add_text_bos=self.cfg.get("add_text_bos", False),
+                phoneme_turn_dropout_batch_prob=self.cfg.get("phoneme_turn_dropout_batch_prob", 0.0),
+                phoneme_turn_dropout_turn_prob=self.cfg.get("phoneme_turn_dropout_turn_prob", 0.0),
+                phoneme_turn_max_words_to_drop=self.cfg.get("phoneme_turn_max_words_to_drop", 2),
+            )
+            dataset = FallbackDataset(dataset)
```
blisc, line 1986: What's the purpose of this fallbackdataset?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -940,6 +1253,205 @@ def training_step(self, batch, batch_idx):
             audio_lens = batch['audio_lens']
             audio_codes, audio_codes_lens = self._codec_helper.audio_to_codes(audio, audio_lens)
 
+        if (
```
blisc, line 1299: Why are you doing data splitting and trimming inside of training_step. Why is it not part of the dataloader?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -377,14 +460,23 @@ def evaluate_dir(
...
+        emotion_cache_dir=emotion_cache_dir,
+    )
 
     asr_model = models['asr_model']
     whisper_model = models['whisper_model']
     whisper_processor = models['whisper_processor']
     feature_extractor = models['feature_extractor']
     speaker_verification_model = models['sv_model']
     speaker_verification_model_alternate = models['sv_model_alternate']
+    emotion_model = models['emotion_model']
```
blisc: What is the idea between trying to get an emotional error rate? What does it measure? What does the measure tell us about generated speech?

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
```diff
@@ -494,7 +603,7 @@ def evaluate_dir(
                 model=speaker_verification_model_alternate,
                 extractor=feature_extractor,
                 device=device,
-                sv_model_type=sv_model_type,
+                sv_model_type="titanet",  # alternate is always titanet
```
blisc, line 514: why are we hardcoding this?

blisc (changes requested):

nemo/collections/common/data/lhotse/dataloader.py
shehzeen: Added

nemo/collections/tts/data/text_to_speech_dataset_lhotse_multiturn.py
shehzeen: Updated, reused the already defined function from `text_to_speech_dataset_lhotse`

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -587,6 +624,193 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+                sil_tensor_converted = sil_tensor_unconverted.clone()
+
+        if not hasattr(self, "_codec_sil_codes_buffer"):
+            self.register_buffer("_codec_sil_codes_buffer", sil_tensor_converted, persistent=False)
+            self.register_buffer("_codec_sil_codes_buffer_unconverted", sil_tensor_unconverted, persistent=False)
+        else:
+            self._codec_sil_codes_buffer.copy_(sil_tensor_converted)
+            self._codec_sil_codes_buffer_unconverted.copy_(sil_tensor_unconverted)
+
+    def streaming_prefill_profile(
```
paarthneekhara: Why is this function named `streaming_prefill_profile` ? Shouldn't it just be `streaming_prefill` ?

nemo/collections/tts/models/easy_magpietts_inference.py
paarthneekhara, line 1695: Add docstrings for new parameters: user_audio_channel_embedding, prefill_like_step, prefill_like_is_last_step

nemo/collections/tts/models/easy_magpietts.py
Edresson: This is a function that we use in SpeechLM2 collection. It is basically a safer check, if data loader fails to load a batch and returns None this fallback function will return the previous batch.

nemo/collections/tts/models/easy_magpietts.py
Edresson: Done, I have also added extra missing docstring easy_magpietts.py and easy_magpietts_inference.py files.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
Edresson: It is necessary, because [here](https://github.com/NVIDIA-NeMo/Speech/blob/7be3e2bf64ffbcf28529a5f4e8eaec07d75591f7/nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py#L270-L273) we are hardcoding the alternate speech encoder to titanet_small. Without hardcoding here if we select the main evaluation speaker encoder to _wavlm_ it would raise a error here.

nemo/collections/tts/models/easy_magpietts_inference.py
Edresson: Good catch, fixed :). It was streaming_prefill_profile  because previous we had multiples methods to this prefill inference approach and I forgot to rename it.

nemo/collections/tts/models/easy_magpietts_inference.py
Edresson: Done, I also added few others docstrings that was missing in this file.

nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py
Edresson: Emotion cosine similarity measures how closely the emotional characteristics of generated speech match those of a reference utterance. It is typically computed by comparing embeddings from a pretrained emotion encoder.

This metric is useful for evaluating emotional accommodation or mirroring: whether the model adapts its emotional expression to the user or conversational context. A higher similarity score suggests that the generated speech conveys an emotional state closer to the reference, while a lower score indicates weaker emotional alignment.

Similar embedding-based measures are used in emotional mirroring benchmarks, such as the [LIVA Mirroring benchmark](https://bench.theliva.ai/legacy/mirroring.html).

The same emotion encoder could also be used during post-training as a reward or scoring model, encouraging the system to generate speech with stronger and more appropriate emotional alignment.

nemo/collections/tts/models/easy_magpietts.py
Edresson: The dataloader already returns the user audio split by conversational turn as batch["user_audio_turn_splitted"], with each turn represented independently as [num_turns, T].

Codec extraction is performed inside training_step because the codec model cant run on dataloader workers. We therefore encode each user turn separately in training_step, obtain its codec embeddings, and then place those embeddings back into the corresponding positions in the full dialogue sequence, ensuring that each user turn precedes the associated agent turn.

Processing each turn independently also allows us to use a non-causal codec encoder without leaking information across turn boundaries. In particular, past or future dialogue audio cannot affect the codec representation of the current user turn. Although the codec receptive field is typically limited to approximately one or two seconds, explicitly separating turns ensures that the training pipeline matches streaming inference behavior as closely as possible.

Currently, this implementation does not support caching the user audio codes, so the codec extraction must be performed during training_step.

examples/tts/magpietts_inference.py
Edresson: Done I moved all the helpers/utils to nemo.collections.tts.modules.magpietts_inference.utils, now the file has only ~600 lines

nemo/collections/tts/data/text_to_speech_dataset_lhotse_multiturn.py
Edresson: Done, now the __getitem__ is much smaller, having only the follow functions call: `_prepare_cuts`, `_collate_audio_channels`, `_collate_text_channels`, `_collate_phoneme_tokens`, `_collect_cut_features` and `_build_batch_dict`

nemo/collections/tts/data/text_to_speech_dataset_lhotse_multiturn.py
```diff
@@ -0,0 +1,1027 @@
...
+from nemo.collections.tts.parts.utils.tts_dataset_utils import (
+    beta_binomial_prior_distribution,
+    normalize_volume,
+    stack_tensors,
+)
+from nemo.core.classes.common import safe_instantiate
+from nemo.utils import logging
+
+
+def check_speaker_format(item: str):
```
blisc, line 40: We should do another pass of this file after we merge. Specifically in regards to adding docstrings.

nemo/collections/tts/metrics/emotion_encoder.py
```diff
@@ -0,0 +1,1293 @@
...
+        self.classifiers.clear()
+        self.whisper_model.cpu()
+
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+
+
+# =============================================================================
+# CLI utilities
+# =============================================================================
```
blisc: Do we need this part?

nemo/collections/tts/metrics/emotion_encoder.py
```diff
@@ -0,0 +1,1293 @@
...
+AUXILIARY_SIMILARITY_LABELS: list[str] = []
+
+# =============================================================================
+# Model architecture specs
+# =============================================================================
+
+MODEL_SPECS: dict[str, dict[str, Any]] = {
+    "small": {
+        "repo_id": "laion/Empathic-Insight-Voice-Small",
+        "whisper_model_id": "mkrausio/EmoWhisper-AnS-Small-v0.1",
```
blisc: What is this whisper model, and why do we need it if we're using laion's models?

blisc (changes requested):
Blocking on the whisper encoder used in the emotion classifier metric

nemo/collections/tts/metrics/emotion_encoder.py
Edresson: It is part of the same model. I believe it was released separately because otherwise it would be difficult to load the Whisper encoder directly from Hugging Face. In the Empathic-Insight-Voice paper, https://arxiv.org/pdf/2506.09827, the authors describe how they pretrained the Whisper model to serve as the backbone of the Empathic-Insight-Voice model.

nemo/collections/tts/metrics/emotion_encoder.py
shehzeen: Do you mean the whole script or the CI utilities?

nemo/collections/speechlm2/parts/pretrained.py
```diff
@@ -302,46 +302,111 @@ def setup_parallel_expert_encoder(model: torch.nn.Module):
 
 
 def set_model_dict_for_partial_init(
-    pretrained_dict: Dict[str, torch.Tensor], model_dict: Dict[str, torch.Tensor]
+    pretrained_dict: Dict[str, torch.Tensor],
+    model_dict: Dict[str, torch.Tensor],
+    allow_partial_copy: bool = False,
```
shehzeen, line 311: @Edresson Should we still keep this change? Since we are no longer loading from a pretrained TTS-only checkpoint.

nemo/collections/tts/modules/nemotron_h_decoder.py
```diff
@@ -250,7 +294,7 @@ def __init__(self, config: NemotronHConfig, batch_size: int, dtype=torch.float16
         for i in range(config.num_hidden_layers):
             if config.layers_block_type[i] == "mamba":
                 self.conv_states.append(
-                    torch.zeros(batch_size, intermediate_size, conv_kernel_size, device=device, dtype=dtype)
+                    torch.zeros(batch_size, conv_dim, conv_kernel_size, device=device, dtype=dtype)
```
paarthneekhara, line 297: Is this the correct implementation? Seems different from what it used to be.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -940,6 +1296,205 @@ def training_step(self, batch, batch_idx):
...
+                end_frame = int(end_sample.item()) // input_samples_per_frame
+
+                start_frame = max(0, min(start_frame, T))
+                end_frame = max(start_frame, min(end_frame, T))
+
+                seq_len = end_frame - start_frame
+                if seq_len <= 0:
+                    continue
+
+                boundary_trim = self.cfg.get("user_audio_boundary_trim", 0)
```
paarthneekhara: Can you add some comments on what this hyper-parameter does?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -940,6 +1296,205 @@ def training_step(self, batch, batch_idx):
...
+                                high=2,  # {-1, 0, 1}
+                                size=(),
+                                device=user_audio_embedded.device,
+                            ).item()
+                        )
+
+                        trim_amount = max(1, base_trim + trim_delta)
+                        trim_amount = min(trim_amount, max(1, copy_len - 1))
+
+                        aug_choice = random.choices(
```
paarthneekhara, line 1444: I feel like there is a lot of code for this trimming augmentation, Is this enabled during training? If not, we can drop some of this.

nemo/collections/tts/modules/magpietts_inference/inference.py
```diff
@@ -824,3 +1021,1176 @@ def _run_decoder_only_inference(
...
+                        state=state,
+                        text_tokens=current_tokens,
+                        use_inference_mode=True,
+                    )
+
+                    if bool(text_exhausted[0].item()) and bool(state.finished[0].item()):
+                        break
+
+                state.audio_prediction_end_idx.fill_(-1)
+                state.finished.zero_()
```
paarthneekhara, line 1431: Why are such custom overrides needed for state? Please add some comments.

paarthneekhara:
Left a few comments, but most changes made to easy_magpietts model make sense. User audio channel code could probably be simplified.

nemo/collections/speechlm2/parts/pretrained.py
Edresson: I think so, this is a nice addition to the set_model_dict_for_partial_init.

nemo/collections/tts/models/easy_magpietts.py
Edresson: It is currently used on training. Except user_audio_boundary_trim, I removed this one.

nemo/collections/tts/modules/magpietts_inference/inference.py
Edresson: Now we have multiples turns, without those changes we would not produce anything else after the first turn, most of overrides done is this class is to handle this scenario.

blisc (dismissed):

nemo/collections/tts/modules/nemotron_h_decoder.py
shehzeen: I believe the old implementation was incorrect for the cached multi-token prefill behavior. The allocation should be as per the new conv_dim and not intermediate size. The old implementation might work with single-step size but not with prefill. @Edresson can you confirm?

Edresson: Yes, Thanks @shehzeen  that is correct.

blisc (dismissed):

paarthneekhara (dismissed):
LGTM

blisc:
Merged bypassing tests. If ASR_3 continues to fail, we will add a pleasefixme. Looking at the changes, nothing should affect the ASR test

PR #15802, opened 2026-06-15
fix: Replace direct hydra.utils.instantiate with safe_instantiate

chtruong814:
# What does this PR do ?

fix: Replace direct hydra.utils.instantiate with safe_instantiate

# Changelog

# Usage

```python
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Related to # (issue)


pzelasko (approved):

nemo/core/classes/common.py
```diff
@@ -114,45 +142,174 @@ def _is_target_allowed(target: str) -> bool:
...
+                return issubclass(obj, Metric)
+            except (ImportError, TypeError):
+                return False
+
+        if target == "torch.distributed.fsdp.MixedPrecisionPolicy":
+            return is_dataclass(obj)
+
         module_name = getattr(obj, "__module__", "") or ""
-        if any(module_name.startswith(p) for p in ALLOWED_NEMO_SUBMODULE_PREFIXES):
+        if any(module_name.startswith(p) for p in ALLOWED_ADAPTER_STRATEGY_PREFIXES):
```
stevehuang52, line 207: Why change to `ALLOWED_ADAPTER_STRATEGY_PREFIXES`?

stevehuang52:
Getting error due to this PR:
```
   File "/home/heh/github/NeMo-main-latest/nemo/core/classes/common.py", line 761, in from_config_dict
      instance = safe_instantiate(config=config)
    File "/home/heh/github/NeMo-main-latest/nemo/core/classes/common.py", line 326, in safe_instantiate
      _validate_config_targets_recursive(config)
    File "/home/heh/github/NeMo-main-latest/nemo/core/classes/common.py", line 312, in _validate_config_targets_recursive
      raise _unsafe_target_error(target_path, "_target_")
  nemo.core.classes.common.UnsafeTargetError: Instantiation of unsafe target 'nemo.collections.asr.modules.transformer_encoder.TransformerEncoder' is blocked. The '_target_' must point to a class or function within an approved namespace. This restriction is in place to prevent potential arbitrary code execution.
  . don't edit anything, just find the cause
```

How can we check if there are other modules affected by this PR? We should also put a note and inform NeMo-Speech developers on what are the base classes that they can safely use, if a normal base python class is not permitted (e.g., `class ABC(object)` fails with the new safe import mechanism but worked before)

Calude code found the bug:

> TransformerEncoder -- `@experimental` wraps it in a wrapt.FunctionWrapper. isinstance(obj, type) returns True (wrapt proxies type checks), but issubclass(obj, Serialization) raises TypeError: issubclass() arg 1 must be a class. The except TypeError: return False in _is_target_allowed line 174 short-circuits immediately, skipping the nn.Module check that would have allowed it.
> 
> The underlying bug in _is_target_allowed is that it uses except TypeError: return False for the Serialization check, which was meant to guard against edge cases but instead kills the entire evaluation chain when wrapt-decorated classes are involved. The nn.Module safe-base check at line 178 is unreachable for any @experimental-decorated class.
> 


cc @chtruong814 @pzelasko @nithinraok

stevehuang52:
script to reproduce:
```python
"""
Reproduces: UnsafeTargetError for TransformerEncoder and MultiSpeakerNoiseAugmentation.

Root cause:
  Both classes are wrapped by or inherit from objects that cause
  `issubclass(obj, Serialization)` to raise TypeError inside _is_target_allowed.
  The `except TypeError: return False` short-circuits before the nn.Module check runs.

  TransformerEncoder: decorated with @experimental (wrapt.FunctionWrapper).
    isinstance(TransformerEncoder, type) -> True  (wrapt proxies type checks)
    issubclass(TransformerEncoder, Serialization) -> TypeError (FunctionWrapper is not a real class)
    -> return False, skipping the nn.Module safe-base check

  MultiSpeakerNoiseAugmentation: inherits from plain object, not Serialization or nn.Module.
    issubclass passes without TypeError, but no safe-base matches -> returns False.
"""
import sys
sys.path.insert(0, '/home/heh/github/NeMo-main-latest')

from nemo.core.classes.common import _is_target_allowed, Serialization
import torch.nn as nn


def check_target(target):
    print(f"Checking target: {target}")
    import hydra
    try:
        obj = hydra.utils.get_class(target)
        ok = True
    except Exception:
        try:
            obj = hydra.utils.get_object(target)
            ok = True
        except Exception as e:
            print(f"  import failed: {e}")
            ok = False
            obj = None

    if obj is not None:
        print(f"  type(obj)                       : {type(obj)}")
        print(f"  isinstance(obj, type)           : {isinstance(obj, type)}")
        print(f"  callable(obj)                   : {callable(obj)}")
        try:
            r = issubclass(obj, Serialization)
            print(f"  issubclass(obj, Serialization)  : {r}")
        except TypeError as e:
            print(f"  issubclass(obj, Serialization)  : TypeError -> {e}  <-- short-circuits to return False!")
        try:
            r = issubclass(obj, (nn.Module,))
            print(f"  issubclass(obj, nn.Module)      : {r}")
        except TypeError as e:
            print(f"  issubclass(obj, nn.Module)      : TypeError -> {e}")

    result = _is_target_allowed(target)
    print(f"  _is_target_allowed              : {result}  {'OK' if result else '<-- BLOCKED (bug)'}")


print("=== TransformerEncoder ===")
check_target("nemo.collections.asr.modules.transformer_encoder.TransformerEncoder")

print()
print("=== MultiSpeakerNoiseAugmentation ===")
check_target("nemo.collections.asr.modules.ssl_modules.MultiSpeakerNoiseAugmentation")
```

nemo/core/classes/common.py
stevehuang52: @chtruong814  @nithinraok could you please help check if this is intended?

nemo/core/classes/common.py
```diff
@@ -114,45 +142,174 @@ def _is_target_allowed(target: str) -> bool:
...
+
         SAFE_BASES = (torch.nn.Module, ModelPT)
         try:
             if issubclass(obj, SAFE_BASES):
                 return True
         except TypeError:
             return False
 
-    # If it's a callable function: allow only if in approved NeMo submodules
-    if callable(obj):
```
stevehuang52, line 136: What's the reason to remove this? @chtruong814

stevehuang52:
I think we need more explanation on the changes in [nemo/core/classes/common.py](https://github.com/NVIDIA-NeMo/Speech/pull/15802/files/3c83c3b2c648afd14135ff21be53a051efb61586#diff-7d9f7edfbaa742f06d56b151859b15e2c3b85bf13483dc2767da60293ab73346), since a lot of them are not very intuitive, and we need some methods to make sure they don't change the original rules unexpectedly.

PR #15805, opened 2026-06-16
[TTS] Fix: derive do_tts local-transformer flag from model config

XuesongYang:
### What & why
`do_tts` requested local-transformer sampling unconditionally:
- `MagpieTTSModel.do_tts` hardcoded `use_local_transformer_for_inference=True`
- `EasyMagpieTTSInferenceModel.do_tts` defaulted `use_local_transformer=True`

This crashes any model whose sampler can't honor it — e.g. one trained with
`local_transformer_type: none` (the config default) — with
`"Local transformer inference requested but local transformer type is ..."`.
It worked for the deployed 357M only because that model is autoregressive (AR).

### Fix
Derive the flag from each model's actual sampler contract:
- **MagpieTTSModel** → `local_transformer_type != NO_LT` (`generate_speech` handles AR **and** MaskGIT)
- **EasyMagpie** → `local_transformer_type == AR` (`_sample_audio_codes` supports **only** AR; MaskGIT/NO_LT decode via parallel sampling). Default is now `Optional[bool] = None` so it auto-derives, while an explicit value still overrides.

No behavior change for AR models (incl. the deployed 357M): both rules yield `True` for AR.

### Tests
New CPU unit tests that drive the real `do_tts` with a mock self and assert the flag passed to the sampler:
- `test_magpietts_do_tts.py` — 3 cases: `NO_LT→False`, `AR→True`, `MASKGIT→True`
- `test_easy_magpietts_do_tts.py` — 6 cases: default derivation (`None`) + explicit override

The NO_LT case fails on the pre-fix code (verified by reintroducing the bug).


rfejgin (commented):
Might make sense to add @subhankar-ghosh to this review since I believe he designed the inference APIs and would be most familiar with each one's intended use case.

In general this change looks good to me. The one thing I'm unsure about, regarding deriving the LT flag from the checkpoint, is this:

In the past, we had checkpoints where the model was trained with LT, but at inference we used the non-LT path. E.g. I believe the most recent Riva release (early this year) worked that way. If we automatically derive the flag from the checkpoint that would not be possible. That said, the flag was already hard-coded to `True` in the `do_tts` path so I'm assuming this flexibility isn't needed (at least not in the `do_tts` API).

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -2137,7 +2137,7 @@ def do_tts(
         context_audio_duration: float = 5.0,
         use_cfg: bool = True,
         cfg_scale: float = 2.5,
-        use_local_transformer: bool = True,
+        use_local_transformer: Optional[bool] = None,
```
blisc:
```suggestion
        use_local_transformer: Optional[bool] = None,  # If unset, defaults to True if AR LT is present
```

blisc (approved):

blisc:
/ok to test [742c957](https://github.com/NVIDIA-NeMo/NeMo/pull/15805/commits/742c9578362b4adeba2c5a1c7e7c737804cc3e9c)

PR #15842, opened 2026-06-27
Fix manifest reference path conversion

fallintoplace:
## Summary
Fix NeMo-to-Lhotse manifest conversion so reference recordings keep their own manifest path instead of reusing the target path.

## Root cause
`convert_manifest_nemo_to_lhotse()` loaded `item_reference_key`, but the reference branch passed `item_target_key` into `_as_relative()`. That made reference-only manifests crash with an unbound local and could serialize the wrong relative metadata when target and reference paths differed.

## What changed
- route the reference path override through `item_reference_key`
- add a regression test for manifests that have `reference_filepath` without `target_filepath`
- add a regression test for manifests whose target and reference paths are both present and intentionally different

## Impact
Reference-only manifests now convert successfully, and mixed target/reference manifests preserve the correct relative path metadata for both recordings.

## Validation
- `uv run pytest tests/collections/audio/test_audio_datasets.py -k "convert_manifest_nemo_to_lhotse_with_reference_only or convert_manifest_nemo_to_lhotse_with_different_target_and_reference_paths" -v`
- `uv run pytest tests/collections/audio/test_audio_datasets.py -k "audio_to_target_with_reference_dataset" -v`
- `uv run --with black==24.10.0 black --check nemo/collections/audio/data/audio_to_audio_lhotse.py tests/collections/audio/test_audio_datasets.py`
- `uv run isort --check-only nemo/collections/audio/data/audio_to_audio_lhotse.py tests/collections/audio/test_audio_datasets.py`

### DIFF
--- a/nemo/collections/audio/data/audio_to_audio_lhotse.py
+++ b/nemo/collections/audio/data/audio_to_audio_lhotse.py
@@ -190,7 +190,7 @@ def convert_manifest_nemo_to_lhotse(
                     get_full_path(audio_file=item_reference_key, manifest_file=input_manifest)
                 )
 
-                _as_relative(cut.reference_recording, item_target_key, enabled=not force_absolute_paths)
+                _as_relative(cut.reference_recording, item_reference_key, enabled=not force_absolute_paths)
 
                 if (channels := item.pop(REFERENCE_CHANNEL_SELECTOR, None)) is not None:
                     if cut.reference_recording.num_channels == 1:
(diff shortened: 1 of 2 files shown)


pzelasko (approved):
Thank you, great catch!

pzelasko:
@fallintoplace please sign off your commits to resolve the failing DCO check which blocks us from merging.

pzelasko:
/ok to test [0fd107b](https://github.com/NVIDIA-NeMo/Speech/pull/15842/commits/0fd107b09a4297e9aa995a16f52677ff4bd780a6)

pzelasko (approved):

PR #15858, opened 2026-06-30
Fix 32-bit indexing overflow in conv subsampling chunking guard

Jeronymous:
# What does this PR do ?

Fixes an intermittent crash in `ConvSubsampling` (FastConformer / Conformer encoders) where a large conv activation exceeds PyTorch's 32-bit indexing limit despite the built-in auto-chunking that is supposed to prevent exactly that.

**Collection**: ASR, speech lm2

## The bug

Training a speech model with a FastConformer encoder (`dw_striding` subsampling) on larger batches / longer audio fails after a while with:
```
RuntimeError: Expected canUse32BitIndexMath(input) && canUse32BitIndexMath(output) to be true, but got false.
```
Trimmed traceback:
```
.../asr/modules/conformer_encoder.py:654  self.pre_encode(x=audio_signal, lengths=length)
.../asr/parts/submodules/subsampling.py:466  x, lengths = self.conv(x, lengths)   # need_to_split == False
.../asr/parts/submodules/subsampling.py:710  x = layer(x)                         # first Conv2d
.../torch/nn/modules/conv.py:543             return F.conv2d(...)
RuntimeError: Expected canUse32BitIndexMath(...) ... got false.
```

cuDNN and PyTorch's native CUDA kernels index tensor elements with 32-bit integers, so a conv input/output must hold fewer than `INT_MAX = 2**31 - 1` elements (see https://github.com/pytorch/pytorch/issues/80020). The first conv of the subsampling stack (`1 -> conv_channels`, strided in both T and F) is the largest activation and is the one that overflows.

`ConvSubsampling` already has logic (`subsampling_conv_chunking_factor`) meant to split the batch so each conv stays under that limit — but the guard let the offending tensor through.

## Root cause

The previous guard estimated the first-conv output size indirectly and with a
loose boundary:
```python
x_ceil = 2**31 / self._conv_channels * self._stride * self._stride   # float approximation
if torch.numel(x) > x_ceil:                                          # '>' against 2**31
    need_to_split = True
```
Three problems:
1. **Wrong boundary.** canUse32BitIndexMath fails when numel >= 2**31 - 1, but the guard splits only when the estimate is strictly greater than 2**31. A tensor of exactly INT_MAX elements — the value that trips the check — is left unsplit and crashes.
2. **Indirect / approximate size.** It derives the output size algebraically from the input numel and conv_channels / stride² instead of computing the actual conv output, so it does not exactly match the tensor that cuDNN sees.
3. **Fractional chunking factor.** The auto split-factor helpers used math.ceil(math.log(numel / 2**31, 2)), which can yield a fractional cf (e.g. a tensor just under the limit → negative exponent → cf = 0.5) and chunks that land exactly at the limit rather than strictly below it.

**The fix**
- Introduce `_MAX_CONV_NUMEL_32BIT = 2**31 - 1`, the exact threshold at which `canUse32BitIndexMath` fails.
- Add `_first_conv_output_numel()` to compute the **exact** element count of the first conv's output (the largest activation) using `calculate_conv_output_size` (real kernel/stride/padding), instead of the algebraic approximation.
- Split when the exact first-conv output or the conv input is >= the limit (correct >= boundary).
- In both auto-factor helpers, compute cf from `numel // MAX + 1` so every chunk is strictly below the limit and cf is always an integer power of two >= 1 (removes the fractional-factor bug).

# Changelog

- Fix canUse32BitIndexMath overflow in ConvSubsampling: compute the exact first-conv output size and split against the true INT_MAX element limit; remove the fractional auto-chunking-factor case.

# Usage

# GitHub Actions CI

**Pre checks**:

- [x] Did you add or update any necessary documentation? -> code comments
  
**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/asr/parts/submodules/subsampling.py
+++ b/nemo/collections/asr/parts/submodules/subsampling.py
@@ -538,9 +538,12 @@ class ConvSubsampling(torch.nn.Module):
         else:
             # avoiding a bug / feature limiting indexing of tensors to 2**31
             # see https://github.com/pytorch/pytorch/issues/80020
-            # Smallest power-of-two batch split that keeps each chunk's exact first-conv
-            # output strictly below the 32-bit element limit (the +1 forces "strictly").
-            cf = 2 ** math.ceil(math.log(self._first_conv_output_numel(x) // _MAX_CONV_NUMEL_32BIT + 1, 2))
+            # Smallest power-of-two batch split that keeps each chunk strictly below the
+            # 32-bit element limit (the +1 forces "strictly"). Mirror the forward() guard
+            # by sizing against whichever of the conv input or its (larger) first-conv
+            # output reaches the limit.
+            numel = max(self._first_conv_output_numel(x), torch.numel(x))
+            cf = 2 ** math.ceil(math.log(numel // _MAX_CONV_NUMEL_32BIT + 1, 2))
             logging.debug(f'using auto set chunking factor: {cf}')
 
         new_batch_size = b // cf
--- a/nemo/collections/asr/parts/submodules/subsampling.py
+++ b/nemo/collections/asr/parts/submodules/subsampling.py
@@ -101,6 +101,13 @@ class StackingSubsampling(torch.nn.Module):
         return x, lengths
 
 
+# cuDNN and PyTorch's native CUDA kernels index tensor elements with 32-bit integers, so
+# any tensor entering or leaving a conv must hold fewer than this many elements; exceeding
+# it raises "Expected canUse32BitIndexMath(...) to be true, but got false".
+# See https://github.com/pytorch/pytorch/issues/80020
+_MAX_CONV_NUMEL_32BIT = 2**31 - 1
+
+
 class ConvSubsampling(torch.nn.Module):
     """Convolutional subsampling which supports VGGNet and striding approach introduced in:
     VGGNet Subsampling: Transformer-transducer: end-to-end speech recognition with self-attention (https://arxiv.org/pdf/1910.12977.pdf)
@@ -424,6 +431,22 @@ class ConvSubsampling(torch.nn.Module):
     def get_streaming_cache_size(self):
         return [0, self.subsampling_factor + 1]
 
+    def _first_conv_output_numel(self, x):
+        """Elements in the first conv layer's output - the largest activation in the stack.
+
+        The first conv (1 -> conv_channels, strided in both T and F) produces the biggest
+        tensor for the 'striding' and 'dw_striding' variants; every later layer only shrinks
+        T and F. ``x`` is the ``(B, T, F)`` input before the channel dim is added.
+
+        Note: this assumes a strided first conv. The 'vgg' variant starts with stride-1
+        convs, so its largest activation is not bounded by this estimate.
+        """
+        b, t, f = x.size()
+        pad = (self._left_padding, self._right_padding)
+        out_t = calculate_conv_output_size(t, self._kernel_size, self._stride, pad)
+        out_f = calculate_conv_output_size(f, self._kernel_size, self._stride, pad)
+        return b * self._conv_channels * out_t * out_f
+
     def forward(self, x, lengths):
         out_lengths = calc_length(
             lengths,
@@ -444,11 +467,14 @@ class ConvSubsampling(torch.nn.Module):
                 # if subsampling_conv_chunking_factor is 1, we split only if needed
                 # avoiding a bug / feature limiting indexing of tensors to 2**31
                 # see https://github.com/pytorch/pytorch/issues/80020
-                x_ceil = 2**31 / self._conv_channels * self._stride * self._stride
-                if torch.numel(x) > x_ceil:
-                    need_to_split = True
-                else:
-                    need_to_split = False
+                # Compare the exact first-conv output (the largest activation in the
+                # stack) against the hard 32-bit element limit, splitting on '>=': at
+                # equality the tensor already has INT_MAX elements, which is exactly what
+                # trips canUse32BitIndexMath. The conv input is guarded too.
+                need_to_split = (
+                    self._first_conv_output_numel(x) >= _MAX_CONV_NUMEL_32BIT
+                    or torch.numel(x) >= _MAX_CONV_NUMEL_32BIT
+                )
             else:
                 # if subsampling_conv_chunking_factor > 1 we always split
                 need_to_split = True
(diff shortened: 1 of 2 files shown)


nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -512,9 +538,9 @@ def conv_split_by_batch(self, x, lengths):
         else:
             # avoiding a bug / feature limiting indexing of tensors to 2**31
             # see https://github.com/pytorch/pytorch/issues/80020
-            x_ceil = 2**31 / self._conv_channels * self._stride * self._stride
-            p = math.ceil(math.log(torch.numel(x) / x_ceil, 2))
-            cf = 2**p
+            # Smallest power-of-two batch split that keeps each chunk's exact first-conv
+            # output strictly below the 32-bit element limit (the +1 forces "strictly").
+            cf = 2 ** math.ceil(math.log(self._first_conv_output_numel(x) // _MAX_CONV_NUMEL_32BIT + 1, 2))
```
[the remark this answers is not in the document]
Jeronymous: done

Jeronymous:
Thanks for the review.
I've added unit tests. Could a maintainer add the Run CICD label to run the test suite when you get a chance?

Jeronymous:
Can someone please review this PR?
Maybe @nithinraok ?

tests/collections/asr/test_asr_subsampling.py
```diff
@@ -11,10 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import math
```
nithinraok: remove this un=used import

nithinraok (dismissed):
Thanks. LGTM

tests/collections/asr/test_asr_subsampling.py
Jeronymous: Thanks for spotting that. Done

nithinraok (dismissed):
Thanks. LGTM

nithinraok:
could you sign-off yours commits if you haven;t already.

Jeronymous:
@nithinraok is it good to merge now?

Jeronymous:
@nithinraok I solved some conflict. Is there something missing before merging ?

nithinraok (commented):
One question on the auto-chunking fallback.

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -584,9 +610,12 @@ def conv_split_by_batch(self, x, lengths):
...
-            cf = 2**p
+            # Smallest power-of-two batch split that keeps each chunk strictly below the
+            # 32-bit element limit (the +1 forces "strictly"). Mirror the forward() guard
+            # by sizing against whichever of the conv input or its (larger) first-conv
+            # output reaches the limit.
+            numel = max(self._first_conv_output_numel(x), torch.numel(x))
+            cf = 2 ** math.ceil(math.log(numel // _MAX_CONV_NUMEL_32BIT + 1, 2))
             logging.debug(f'using auto set chunking factor: {cf}')
 
         new_batch_size = b // cf
```
nithinraok, line 612: Looking at this PR again, I spotted something: Wouldn't this fail when `cf > b`? With `B=3` and `cf=4`, `new_batch_size` becomes zero. The fallback then treats the batch as convolution channels. Could we use a batch chunk size of one when each sample is below the limit and add a `B=3` regression test? For `B=1`, the fallback still runs the full first convolution before splitting.

Jeronymous: Good catch. You're right, and it actually crashes rather than just being suboptimal. When the factor rounds up past the batch size (your b=3, cf=4 case), b // cf is 0, so we fall back to conv_split_by_channel, which unsqueezes the batch and runs the first conv with the batch read as channels. That raises expected input to have 1 channels, but got 3.

I now split into batches of one in that case, and only fall back to channel splitting when a single sample alone exceeds the limit. Added a b=3 regression test that checks we get three single-sample chunks with no channel fallback and the same output as the unsplit run.

On the B=1 point: when a single sample's first conv output alone is over the limit, conv_split_by_channel still runs the full first conv before splitting the later layers, so it can't help there. That's pre-existing and independent of this change.

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -584,14 +610,24 @@ def conv_split_by_batch(self, x, lengths):
...
-        if new_batch_size == 0:  # input is too big
-            return x, lengths, False
+        if new_batch_size == 0:
+            # Power-of-two rounding can push cf past b (e.g. b=3 -> cf=4), or the batch is
```
nithinraok: Could we reduce the comments introduced in this PR across both `subsampling.py` and the new tests? Several comments repeat the code or the PR description. Please keep only non-obvious constraints. For example, this block could be: `# If cf > b and one sample fits, use single-sample batches.` Please apply the same cleanup to the other new comment blocks.

nemo/collections/asr/parts/submodules/subsampling.py
```diff
@@ -584,14 +610,24 @@ def conv_split_by_batch(self, x, lengths):
...
+            # smaller than a manually set factor. If a single sample still fits under the
+            # limit, split into batches of one rather than giving up; only fall back to
+            # channel splitting when one sample alone exceeds the limit, since
+            # conv_split_by_channel runs the full first conv on the whole input up front.
+            if self._first_conv_output_numel(x[:1]) >= _MAX_CONV_NUMEL_32BIT:
```
nithinraok: With an input-dominant configuration, the input can exceed the limit while the output remains below it. Could we use `max(self._first_conv_output_numel(x[:1]), torch.numel(x[:1]))` here?

nithinraok (commented):
Two small follow-ups on the latest update.

nemo/collections/asr/parts/submodules/subsampling.py
Jeronymous: No problem. Done

nemo/collections/asr/parts/submodules/subsampling.py
Jeronymous: Right. Done

nithinraok (approved):
LGTM

PR #15863, opened 2026-07-02
Cache-aware Streaming ASR: up to 5x faster encoder with CUDA graphs

hamuzhan:
# What does this PR do ?

This is an opt-in, off-by-default speedup for the cache-aware FastConformer streaming encoder
steady-state path. Non-steady-state steps, CUDA autocast, capture failures and unseen shapes all
run eager, the number of retained graphs is bounded, and the tests cover the autocast fallback,
cache-buffer aliasing, the train/val lifecycle and the `max_graphs` limit.

Adds an opt-in CUDA graph path for the cache-aware streaming encoder step
(`StreamingEncoder.cache_aware_stream_step`). This targets any cache-aware FastConformer streaming
encoder that goes through `StreamingEncoder.cache_aware_stream_step` (Parakeet / Nemotron Speech /
`stt_*_fastconformer_*_streaming_*`); it is not a generic path for all ASR encoders. Low-latency
streaming inference of these models spends most of its time waiting on the host, not computing.
Each streaming step launches around 1.5k small kernels, so the GPU is idle for most of the step
while the host enqueues them.

The steady-state streaming step has fully static shapes. The caches are allocated at full size up
front and only the `cache_last_channel_len` values change between steps. That means the whole step
can be captured once into a `torch.cuda.CUDAGraph` and replayed with a single launch. The existing
RNNT/TDT label-looping decoder (`GreedyBatchedLabelLoopingComputerBase` and the `WithOptionalCudaGraphs`
infrastructure) already applies this idea to decoding; this PR reuses the same `WithOptionalCudaGraphs`
enable/disable and fallback design for the encoder step. The encoder is simpler because it has no
data-dependent control flow, so plain CUDA graphs are enough (no conditional nodes, no `cuda-python`
dependency).

You turn it on with `asr_model.encoder.set_streaming_cuda_graphs(True)` or `use_cuda_graphs=true` in
the cache-aware streaming inference script. For the covered non-autocast configurations the graph
path is expected to preserve eager execution semantics, and the unit tests assert `torch.equal`
(not `allclose`) against eager. Steps that are not part of the steady state (the first chunk, the
final `keep_all_outputs=True` chunk, or any shape we have not seen) just run eager.

A few design notes:

- **Stable output buffers.** The captured step copies its outputs into buffers owned by the regular
  caching allocator. Without this, the decoder's own CUDA graph aliased the encoder graph's private
  memory pool and corrupted decoding, so it is needed for correctness.
- **Safe teardown.** Graphs are dropped only after a full device sync. Tearing down a graph pool
  while replays are still in flight is not safe.
- **`WithOptionalCudaGraphs`.** Implemented so the existing `ASRModel` train/val hooks disable and
  re-enable the encoder graphs the same way they already do for the decoder (graphs do not release
  memory during training).
- **Graph key.** A step is captured/replayed only when the input shape, dtype, device, the
  streaming params that change the computation (`att_context_size`, `last_channel_cache_size`,
  `valid_out_len`, `drop_extra_pre_encoded`) and `keep_all_outputs` all match. The
  `cache_last_channel_len` value (which changes during the cache-fill phase) is a graph input, not
  part of the key: its shape is fixed, so it is just copied into the static input buffer each step.
- **CUDA autocast runs eager.** A captured graph fixes the dtype/kernel/workspace choices at capture
  time, so replaying a non-autocast graph under autocast (or the reverse) can silently return
  wrong-precision outputs. We could add the autocast state to the graph key, but a wrong key there
  is a silent-correctness bug, so the graph path is skipped whenever CUDA autocast is active
  (cache-aware streaming models run in float32 anyway). There is a test for this. The autocast check
  handles both the device-specific and the no-argument `torch.is_autocast_enabled` signatures for
  PyTorch compatibility.
- **Eager fallback** if capture ever fails; after a failure the mode is set to no-graphs so the
  same shape is not retried on every step (unless a mode is forced for testing).

Results on `nvidia/nemotron-3.5-asr-streaming-0.6b` (GH200, fp32/TF32), steady-state step p50:

| att_context_size (chunk) | batch | eager | graphed | speedup | real-time streams (est.) |
|---|---|---|---|---|---|
| `[56, 0]` (80 ms) | 1 | 22.5 ms | 4.4 ms | 5.14x | 4 -> 18 |
| `[56, 0]` (80 ms) | 32 | 27.0 ms | 8.8 ms | 3.08x | 95 -> 292 |
| `[56, 3]` (320 ms) | 8 | 28.9 ms | 8.4 ms | 3.45x | 89 -> 305 |
| `[56, 3]` (320 ms) | 32 | 31.0 ms | 13.3 ms | 2.32x | 331 -> 767 |
| `[56, 13]` (1120 ms) | 32 | 39.4 ms | 23.2 ms | 1.69x | 911 -> 1543 |

The win is biggest in the low-latency, low-batch case (the voice-agent use case), where launch
overhead dominates the step. End-to-end on the documented script (`[56,0]`, batch 10): 9.87 s ->
3.08 s (3.2x).

`real-time streams (est.)` is a rough capacity estimate, not a full-pipeline benchmark. It is
`batch * chunk_duration / step_p50` using the per-step encoder+decoder time measured above (chunk
duration = `(right_context + 1) * 80 ms`); it does not model audio I/O, tokenization or host-side
scheduling at high concurrency, so treat it as an upper bound that scales with the step speedup.

Memory overhead (extra peak device memory retained by the graph pool plus static input/output
buffers vs eager, encoder-only, GH200, same model):

| att_context_size | batch | eager peak | graphed peak | overhead |
|---|---|---|---|---|
| `[56, 0]` | 1 | 2651 MB | 2697 MB | +46 MB |
| `[56, 0]` | 32 | 3317 MB | 3757 MB | +440 MB |
| `[56, 3]` | 32 | 3325 MB | 3763 MB | +438 MB |
| `[56, 13]` | 32 | 3363 MB | 3800 MB | +437 MB |

The overhead scales with batch (one step of intermediate activations) and is essentially
independent of the look-ahead. Each captured graph retains its pool until `disable_cuda_graphs()`
or the helper is detached, so at most `max_graphs` (default 8) pools can be live at once.

Correctness: `speech_to_text_cache_aware_streaming_infer.py ... use_cuda_graphs=true
compare_vs_offline=true` reports 0 differences streaming-vs-offline, identical WER, and transcripts
that match the eager run exactly (checked at `[56,13]` and `[56,0]`). Tests cover:

- graphed vs eager encoder outputs, exact (`torch.equal`), including the cache-fill phase and the
  final `keep_all_outputs=True` step;
- the static input cache buffers and the stable output cache buffers are distinct (a replay does
  not overwrite its own inputs);
- a new graph is captured after an `att_context_size` switch (no stale-graph reuse), and the
  `max_graphs` limit is respected (further keys stay eager);
- eager fallback in training mode, under CUDA autocast, and in forced no-graphs mode;
- disable/re-enable lifecycle, the recursive `WithOptionalCudaGraphs` toggle, and that the real
  `ASRModel.disable_cuda_graphs()` / `maybe_enable_cuda_graphs()` also toggle the encoder graphs;
- no-op without CUDA, and argument validation.

**Collection**: ASR

# Changelog

- Add `CudaGraphsStreamingEncoderStep` (`nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py`):
  keyed capture/replay of the streaming encoder step with warmup, stable output buffers, eager
  fallback, and a forced-mode hook for testing; implements `WithOptionalCudaGraphs`.
- Add `StreamingEncoder.set_streaming_cuda_graphs()` and route `cache_aware_stream_step` through the
  optional graph path (`nemo/collections/asr/parts/mixins/streaming.py`).
- Extend `ASRModel.maybe_enable_cuda_graphs()` / `disable_cuda_graphs()` to also toggle the encoder
  streaming graphs, so the training/validation hooks manage them like the decoder graphs
  (`nemo/collections/asr/models/asr_model.py`).
- Add a `use_cuda_graphs` flag to `speech_to_text_cache_aware_streaming_infer.py`.
- Add `CacheAwareASRInferenceWrapper.set_streaming_cuda_graphs()` delegation.
- Document the feature in the cache-aware streaming Conformer section of the ASR docs.
- Add `tests/collections/asr/test_streaming_encoder_cuda_graphs.py` (14 tests): graphed-vs-eager
  correctness for the supported non-autocast CUDA inference path (cache-fill phase, final
  `keep_all_outputs` fallback), `max_graphs` behavior, input/output cache-buffer aliasing,
  model-level train/val hooks, and CUDA autocast fallback.

# Usage

```python

asr_model = nemo_asr.models.ASRModel.from_pretrained("nvidia/nemotron-3.5-asr-streaming-0.6b").eval()
asr_model.encoder.set_streaming_cuda_graphs(True)  # opt-in; CUDA autocast falls back to eager
# ... then run the usual cache-aware streaming loop (conformer_stream_step / the inference wrapper)
```

From the simulation script:

```bash
python examples/asr/asr_cache_aware_streaming/speech_to_text_cache_aware_streaming_infer.py \
    pretrained_name=nvidia/nemotron-3.5-asr-streaming-0.6b \
    dataset_manifest=<manifest.json> \
    att_context_size="[56,0]" target_lang=en-US \
    use_cuda_graphs=true
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

cc @nithinraok (NeMo core / ASR).

# Additional Information

- The feature is opt-in and off by default, so there is no behavior change unless
  `set_streaming_cuda_graphs(True)` or `use_cuda_graphs=true` is set. It no-ops without CUDA.
- This extends the existing decoder CUDA-graph work to the encoder side: it reuses the
  `WithOptionalCudaGraphs` interface and the same enable/disable + graceful-fallback pattern used by
  the label-looping decoders, so the train/val lifecycle behaves identically for both.

Limitations / scope:

- CUDA inference in eval mode only; training and validation disable it via the existing hooks.
- First chunk, final `keep_all_outputs=True` chunk, unseen shapes, active autocast, and capture
  failures fall back to eager.
- Graph pools are retained until `disable_cuda_graphs()` / detach (see the memory table above).
- Wired through the encoder setter and the cache-aware streaming script; the streaming Pipeline API
  config wiring is left for a follow-up.
- Not validated for multi-GPU / model-parallel encoders; the graph is keyed by device index and a
  device change re-captures, but DDP inference has not been exercised here.

Follow-ups (separate PRs, not in here): wiring the flag through the streaming Pipeline API config;
cutting the attention op count (SDPA plus positional-embedding precompute); a cache ring-buffer.


hamuzhan:
@nithinraok

hamuzhan:
@artbataev

artbataev:
@hamuzhan thank you so much for your contribution! The speedup is truly impressive!
I'm a bit swamped this week, but I'll make time to review the PR by the end of the week.

hamuzhan:
@artbataev Thank you for taking the time to review it. Looking forward to your feedback! Have a great week!

hamuzhan:
Hello @artbataev , any updates?

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
```diff
@@ -0,0 +1,326 @@
...
+``cuda-python`` requirement). Outputs are written into stable buffers owned by the regular caching
+allocator so that no consumer (for example the decoder's own CUDA graph) aliases the encoder graph's
+private memory pool.
+
+Non-uniform steps run eager: the first step(s) (different pre-encode cache handling), the final step
+of an utterance (``keep_all_outputs=True``), and any call whose shapes/parameters do not match an
+already-captured graph.
+"""
+
+from typing import Any, Dict, Optional, Tuple
```
artbataev: Please prefer `dict`, `tuple`, `list` instead of using `typing.Dict` etc.

tests/collections/asr/test_streaming_encoder_cuda_graphs.py
```diff
@@ -0,0 +1,327 @@
...
+        assert encoder._stream_step_cuda_graphs is not None
+        encoder.set_streaming_cuda_graphs(enabled=False)
+        assert encoder._stream_step_cuda_graphs is None
+
+    @pytest.mark.unit
+    def test_asr_model_hooks_toggle_encoder_graphs(self, monkeypatch):
+        """`ASRModel.disable_cuda_graphs()` / `maybe_enable_cuda_graphs()` must also toggle the
+        encoder streaming graphs, so the train/val Lightning hooks manage them like the decoder
+        graphs. Uses the real ASRModel methods on a minimal container (no model download)."""
+        from nemo.collections.asr.models.asr_model import ASRModel
```
artbataev: Please use global imports if possible

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
```diff
@@ -0,0 +1,326 @@
...
+            return False
+        self.cuda_graphs_mode = None
+        self.reset_cuda_graphs_state()
+        return True
+
+    def reset_cuda_graphs_state(self):
+        """Drop all captured graphs. Synchronizes the device first: destroying a graph's
+        private memory pool while replays or other graphs are in flight is unsafe."""
+        if self._graphs and torch.cuda.is_available():
+            torch.cuda.synchronize()
```
artbataev: `torch.cuda.synchronize()` without a device synchronizes only the current device
The graph can belong to another device
Please, fix this with passing device handled by graph(s)

nemo/collections/asr/parts/mixins/streaming.py
```diff
@@ -37,6 +45,41 @@ def to_numpy(tensor):
...
+            The attached ``CudaGraphsStreamingEncoderStep`` helper, or None when disabling.
+        """
+        if not enabled:
+            helper = self._stream_step_cuda_graphs
+            if helper is not None:
+                helper.disable_cuda_graphs()
+            self._stream_step_cuda_graphs = None
+            return None
+        # local import: avoid a circular import at module load time
+        from nemo.collections.asr.parts.submodules.streaming_encoder_cuda_graphs import CudaGraphsStreamingEncoderStep
```
artbataev: Why this is needed? If I correctly understand, `streaming_encoder_cuda_graphs` does not depend on this mixin, so local import looks redundant

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
```diff
@@ -0,0 +1,326 @@
...
+    versions it takes no argument and reports the CUDA autocast state. We try the device-specific
+    call first and fall back to the no-argument form.
+    """
+    try:
+        return torch.is_autocast_enabled("cuda")
+    except TypeError:
+        return torch.is_autocast_enabled()
+
+
+class _CapturedStep:
```
artbataev, line 62: Simplify with dataclass

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
```diff
@@ -0,0 +1,326 @@
...
+        # in-place updates to them are only allowed from inside inference mode
+        with torch.inference_mode():
+            static = captured.static_inputs
+            static["signal"].copy_(signal)
+            static["length"].copy_(length)
+            static["cache_last_channel"].copy_(cache_last_channel)
+            static["cache_last_time"].copy_(cache_last_time)
+            static["cache_last_channel_len"].copy_(cache_last_channel_len)
+            captured.graph.replay()
+        return captured.stable_outputs
```
artbataev: Please return a copy, not directly `captured.stable_outputs`.
Current `stable_outputs` is a rewritable storage, and if any code will try operating them, it will have bugs (since the content of `stable_outputs` is overwritten on each step).
My suggestion: return something like `captured.get_stable_outputs_copy()` where `get_stable_outputs_copy` is a function that clones tensors of `stable_outputs` internally and returns new tensors not attached to the graph

artbataev (changes requested):
Thanks a lot for great work! 
The improvement is significant (I checked it on GPU), this is definitely valuable and should be in NeMo!

Please see my comments and suggestions below.

**Critical**: return copy of the outputs.
**Also, please add this feature to official inference API** `examples/asr/asr_streaming_inference/asr_streaming_infer.py`. Add  `asr.use_cuda_graphs` param to `cache_aware_rnnt.yaml`, and support the parameter in inference pipeline.
@naymaraq FYI

hamuzhan:
@artbataev Thank you for the detailed review! Addressed all your requested changes in 9de4ddab6
 - return independent copies of replay outputs.
 - add asr.use_cuda_graphs to the official cache-aware streaming config and wire it through the inference pipeline
 - store and synchronize each captured graph device on teardown
 - apply the import, dataclass, and built-in type annotation cleanups

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
hamuzhan: Done. replaced the typing container aliases with built-in dict and tuple annotations, only Any remains imported from typing

tests/collections/asr/test_streaming_encoder_cuda_graphs.py
hamuzhan: moved the ASRModel import to module scope

nemo/collections/asr/parts/mixins/streaming.py
hamuzhan: Yeah you're right sorry for confusion, there is no circular dependency here. I moved CudaGraphsStreamingEncoderStep to a module-level import and removed the local import

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
hamuzhan: _CapturedStep is now a dataclass

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
hamuzhan: each captured step now retains its capture device, and reset_cuda_graphs_state() synchronizes every unique captured graph device before clearing the graph pools. Added regression coverage for this behavior

nemo/collections/asr/parts/submodules/streaming_encoder_cuda_graphs.py
hamuzhan: replay now returns clones via get_stable_outputs_copy() instead of exposing the rewritable stable graph buffers. Added a regression test that verifies a prior replay result remains unchanged after a subsequent replay

artbataev (approved):
Thanks a lot, great work!
Approved from my side.

nithinraok:
@hamuzhan great PR. Thanks for the contribution. Could you help sign off the commits? We need to pass the DCO check.

hamuzhan:
Hello again @nithinraok sorry for being late, I was away for a while unfortunately. I've signed off on the commits, is there anything else needed?

PR #15876, opened 2026-07-06
Fix ASR transducer nemo checkpoint export

KunalDhawan:
> [!IMPORTANT]  

# What does this PR do ?

Avoid reloading the best checkpoint into the active model when exporting `.nemo` checkpoints.

**Collection**: ASR / Core checkpointing

# Changelog

- Save best `.nemo` exports directly from the current module when a new best checkpoint is detected.
- Remove the temporary checkpoint reload/restore path during `always_save_nemo`.

# Usage

No API changes.

```python
trainer.fit(model)
```

# Additional Information

- Issue: ASR_with_Transducers.ipynb fails during checkpoint save because NeMoModelCheckpoint reloads the best .ckpt into the live model before saving .nemo, and that reload hits pos_bias_u/v shape mismatches.

### DIFF
--- a/nemo/utils/callbacks/nemo_model_checkpoint.py
+++ b/nemo/utils/callbacks/nemo_model_checkpoint.py
@@ -16,7 +16,6 @@ import os
 import re
 import shutil
 import time
-from copy import deepcopy
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


nithinraok (approved):
LGTM

PR #15915, opened 2026-07-15
[TTS] Update multi-language eval report format

rlangman:
# What does this PR do ?

Update format of TTS comparison report

**Collection**: [TTS]

# Changelog

- Add language of each dataset to the UI
- Fix bug where language in eval report folder name was default language instead of dataset specific language
- Write filewise metrics file with raw text instead of ASCII encoded

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Summary by CodeRabbit

- **New Features**
  - TTS comparison reports now display each benchmark’s language alongside its name.
  - Benchmark metadata is presented more consistently across audio and evaluation reports.

- **Bug Fixes**
  - Evaluation output folders now separate results by language and dataset.
  - Metrics files preserve non-ASCII characters when exported as JSON.

### DIFF
--- a/scripts/tts_comparison_report/README.md
+++ b/scripts/tts_comparison_report/README.md
@@ -127,7 +127,7 @@ python scripts/tts_comparison_report/generate_report.py \
   --s3_region us-west-2 \
   --task_id NEMOTTS-2007 \
   --audio_report \
-  --audio_report_benchmarks libritts_test_clean,riva_hard_digits \
+  --audio_report_benchmarks libritts,riva_en \
   --samples_per_benchmark 20
 ```
 
@@ -145,7 +145,7 @@ python scripts/tts_comparison_report/generate_report.py \
   --s3_region us-west-2 \
   --task_id NEMOTTS-2007 \
   --audio_report \
-  --audio_report_benchmarks libritts_test_clean,riva_hard_digits \
+  --audio_report_benchmarks libritts,riva_en \
   --samples_per_benchmark 20 \
   --remote_hostname your_remote_host \
   --remote_username your_user
@@ -160,13 +160,13 @@ python scripts/tts_comparison_report/generate_report.py \
   --baseline_path /workspace/NeMo/exp/buckets/baseline \
   --candidate_name "Model B" \
   --candidate_path /workspace/NeMo/exp/buckets/candidate \
-  --benchmarks libritts_test_clean,riva_hard_digits,riva_hard_letters \
+  --benchmarks libritts,riva_en,riva_en_hard_sentences \
   --s3_endpoint https://your-s3-endpoint \
   --s3_bucket your_bucket_name \
   --s3_region us-west-2 \
   --task_id NEMOTTS-2007 \
   --audio_report \
-  --audio_report_benchmarks libritts_test_clean,riva_hard_digits \
+  --audio_report_benchmarks libritts,riva_en \
   --samples_per_benchmark 20
 ```
 
--- a/scripts/tts_comparison_report/README.md
+++ b/scripts/tts_comparison_report/README.md
@@ -187,8 +187,8 @@ The default expiration time is one year.
 - The expiration time is also included as a suffix in the uploaded artifacts
 directory name, using the format `%Y-%m-%dT%H-%M-%SZ`, so uploaded reports
 can be filtered and deleted later if needed.
-- Both generated reports include a clickable POR link derived from `--task_id`. 
-If no task ID is specified, the link points to the POR project page.
+- Both generated reports include a clickable Jira link derived from `--task_id`. 
+If no task ID is specified, the link points to the Jira project page.
 
 ## Maintenance
 
--- a/scripts/tts_comparison_report/README.md
+++ b/scripts/tts_comparison_report/README.md
@@ -187,8 +187,8 @@ The default expiration time is one year.
 - The expiration time is also included as a suffix in the uploaded artifacts
 directory name, using the format `%Y-%m-%dT%H-%M-%SZ`, so uploaded reports
 can be filtered and deleted later if needed.
-- Both generated reports include a clickable Jira link derived from `--task_id`. 
-If no task ID is specified, the link points to the Jira project page.
+- Both generated reports include a clickable POR link derived from `--task_id`. 
+If no task ID is specified, the link points to the POR project page.
 
 ## Maintenance
 
--- a/scripts/tts_comparison_report/reporting/constants.py
+++ b/scripts/tts_comparison_report/reporting/constants.py
@@ -17,7 +17,7 @@ from pathlib import Path
 _ROOT: Path = Path(__file__).parent.parent
 
 # Benchmark names supported by the comparison report pipeline.
-BENCHMARK_META = {
+BENCHMARK_META: dict[str, str] = {
     'libritts': 'en',
     'riva_en': 'en',
     'riva_en_hard_sentences': 'en',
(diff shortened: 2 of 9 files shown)


scripts/tts_comparison_report/reporting/constants.py
```diff
@@ -17,15 +17,30 @@
...
+    'AI4bharat': 'hi',
+    'cmltts_it': 'it',
+    'jvs_jsut': 'ja',
+    'F5I9N7A1': 'ko',
+    'cmltts_pt': 'pt',
+    'vivos': 'vi',
+    'mscenespeech': 'zh',
+}
+
+SUPPORTED_BENCHMARK_NAMES = BENCHMARK_META.keys()
```
artem-gorodetskii: The type was `list[str]`, but it became `KeysView`; it should be converted to a list

scripts/tts_comparison_report/reporting/constants.py
```diff
@@ -52,4 +67,4 @@
 DUMMY_TASK_ID: str = "NEMOTTS-0000"
```
artem-gorodetskii, line 67: Since we decided to move from Jira to POR, `DUMMY_TASK_ID` is no longer valid. Linking a report to Jira used to be optional, so the task ID was optional as well. However, all development work must now be linked to a POR ticket, which means every comparison report should also be linked to a POR ticket. This suggests that we no longer need `DUMMY_TASK_ID`, and that `--task_id` should become a required argument in the report-generation CLI. This will also require updates to dependent functionality, such as the `TaskInfo` structure.

scripts/tts_comparison_report/reporting/helpers.py
artem-gorodetskii, line 39: Since we decided to move from Jira to POR, `TaskInfo` should be updated accordingly. All Jira-specific references should be removed, including variable names and docstrings. It would make sense to keep only the `task_id` and `task_url` attributes.

scripts/tts_comparison_report/templates/audio_report.jinja
```diff
@@ -281,7 +281,7 @@
       <h2>Links ↗</h2>
       <ul>
         <li class="link-comment">
-          <a href="{{ jira_url }}" target="_blank" rel="noopener noreferrer">Jira {{ jira_id }}</a>
+          <a href="{{ jira_url }}" target="_blank" rel="noopener noreferrer">POR {{ jira_id }}</a>
```
artem-gorodetskii: Please rename `jira_url` to `task_url` and `jira_id` to `task_id` for this template to match the updated terminology

scripts/tts_comparison_report/templates/eval_report.jinja
```diff
@@ -384,7 +384,7 @@
       <h2>Links ↗</h2>
       <ul>
         <li class="link-comment">
-        <a href="{{ jira_url }}" target="_blank" rel="noopener noreferrer">Jira {{ jira_id }}</a>
+        <a href="{{ jira_url }}" target="_blank" rel="noopener noreferrer">POR {{ jira_id }}</a>
```
artem-gorodetskii: Same: please rename `jira_url` to `task_url` and `jira_id` to `task_id` for this template to match the updated terminology

scripts/tts_comparison_report/README.md
```diff
@@ -187,8 +187,8 @@ The default expiration time is one year.
 - The expiration time is also included as a suffix in the uploaded artifacts
 directory name, using the format `%Y-%m-%dT%H-%M-%SZ`, so uploaded reports
 can be filtered and deleted later if needed.
-- Both generated reports include a clickable Jira link derived from `--task_id`. 
-If no task ID is specified, the link points to the Jira project page.
+- Both generated reports include a clickable POR link derived from `--task_id`. 
+If no task ID is specified, the link points to the POR project page.
```
artem-gorodetskii: Suggestion: remove this sentence, since every comparison report should now be linked to a POR ticket, so a missing task ID is no longer a valid case

artem-gorodetskii:
[RESOLVED] Since `SUPPORTED_BENCHMARK_NAMES` now includes multilingual benchmarks, it should not be used directly as the default for `_DEFAULT_BENCHMARK_NAMES` in `generate_report.py`. Please keep the default limited to a small set of key English benchmarks. Also, the default value for `--audio_report_benchmarks` should be updated, since `libritts_test_clean,riva_hard_digits,riva_hard_letters` is no longer a valid default

artem-gorodetskii (changes requested):

artem-gorodetskii:
@coderabbitai review

scripts/tts_comparison_report/reporting/constants.py
blisc: Let's keep it as optional, but let's add a warning or error message if not supplied. And suggest experimenters to add one.
Not all work will have a clear mapping to a POR item, hence why it should be strongly encouraged but not required.

scripts/tts_comparison_report/reporting/helpers.py
rlangman: I have removed all references to jira from the `tts_comparison_report` directory

scripts/tts_comparison_report/reporting/helpers.py
```diff
@@ -37,21 +37,19 @@ def make_expiration_info(expires_in: int) -> ExpirationInfo:
...
 
     Returns:
-        Task information with the original task ID, derived Jira ID, and Jira URL.
+        Task information with the original task ID and task URL.
     """
-    jira_id = task_id if task_id != DUMMY_TASK_ID else task_id.split("-")[0]
-    jira_url = f"{JIRA_TICKET_URL_PREFIX}/{jira_id}"
+    task_url = f"{TICKET_URL_PREFIX}/{task_id}"
 
     return TaskInfo(
```
artem-gorodetskii, line 51: The previous workflow was:

- `task_id` was the internal task identifier. If not specified, it defaulted to `DUMMY_TASK_ID`. It was used in the S3 report name and this naming is important for maintenance: it allows us to remove reports for old tasks and to deprioritize reports without explicit task IDs.
- `jira_id` was derived from `task_id` and used only for display in the report sidebar. If `DUMMY_TASK_ID` was used, `jira_id` became `NEMOTTS`, so the report still showed our Jira space.
- `jira_url` was derived from `JIRA_TICKET_URL_PREFIX` and `jira_id`, so it always pointed to a valid page: either the real Jira ticket or, when no task ID was provided, the Jira project page.

Currently, `task_id` is being used for all three purposes:
- S3 report naming,
- displayed task ID in the report,
- and construction of `task_url`.

As a result, when `task_id` is not specified, the report shows `DUMMY_TASK_ID` and generates a broken POR link.

If we want to keep `task_id` optional, the previous behavior should be preserved. In other words, we should keep separate values for:
- the identifier used in S3 naming (actual POR ID or `NMP-I-000`),
- the identifier displayed in the report (actual POR ID or something like `NMP-I` or `NMP`),
- and the clickable URL (real POR ticket URL or the default `TICKET_URL_PREFIX`).

Sorry for the earlier misunderstanding.

artem-gorodetskii (changes requested):

scripts/tts_comparison_report/reporting/helpers.py
rlangman: Unfortunately I do not fully understand what you are trying to say, and the significant refactoring required is out of scope of this PR (which was intended to add multi-language support for evaluation).

So I have reverted all changes to the URL and task ID logic. Somebody can open a separate PR to fix them, if they want.

artem-gorodetskii: Ok

artem-gorodetskii (approved):

PR #15923, opened 2026-07-19
[ASR] Wordboosting by phrase

AadamHaq:
> [!IMPORTANT]  

# What does this PR do ?

Adds support for per-phrase `context_score` and `boosting_tree_alpha` in GPU Phrase Boosting, so individual phrases can be boosted by different amounts instead of all sharing the global values.

**Collection**: ASR

# Changelog

- `PhraseItem` gains optional `context_score` and `boosting_tree_alpha` fields (`lang` is now also optional, falling back to `source_lang`), usable via the existing `key_phrase_items_list` config - fully backwards compatible.
- Per-phrase `boosting_tree_alpha` is baked into the graph weights at build time as a *relative* multiplier (effective boost = decode-time `boosting_tree_alpha` × per-phrase alpha). No decoder changes: backoff weights are node-score differences and inherit the scaling automatically; `final_eos_score` scales by the ending phrase's alpha. Baked weights survive `.nemo` save/load.
- `ContextGraph.build()` accepts per-phrase `alphas`; the per-phrase `scores` sentinel changed from `0.0` to `None` so an explicit zero score is expressible.
- `GPUBoostingTreeModel.from_config` resolves per-phrase params (per-phrase `context_score` takes precedence over `score_per_phrase`), validates them (negative alpha raises, zero warns), and applies them to all BPE-dropout tokenizations of a phrase.
- New unit tests (`TestPerPhraseBoostingParams`, 9 tests) and a "Per-phrase boosting parameters" section in `word_boosting.rst`.

# Usage

```python
boosting_tree.key_phrase_items_list='[{phrase:"nvlink",boosting_tree_alpha:2.0},{phrase:"omniverse cloud",context_score:1.5},{phrase:"gtc"}]'
```
Omitted fields fall back to the global values. Also works with `scripts/asr_context_biasing/build_gpu_boosting_tree.py` (parameters are baked into the saved tree).

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

## Who can review?

cc @nithinraok


nithinraok:
@andrusenkoau could you have a look at this PR

artbataev (changes requested):
@AadamHaq Thanks for the contribution.
Please wait while ongoing work https://github.com/NVIDIA-NeMo/Speech/pull/15800 is merged into main - I will let you know.
That PR introduces additional function for building the graph, so we need to make sure that both approaches are supported with your enhancements.

AadamHaq:
@artbataev I just saw your PR merged. Are there any other PRs I should wait for or is it good for me to refactor given the changes you've done?

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
```diff
@@ -35,8 +35,11 @@
 @dataclass
 class PhraseItem:
     phrase: str  # phrase itself
-    lang: str  # per-phrase language (for aggregate tokenizer)
-    # custom weight can be further added
+    lang: Optional[str] = None  # per-phrase language (for aggregate tokenizer); None -> cfg.source_lang
+    context_score: Optional[float] = None  # per-phrase score for each arc transition; None -> global behavior
+    # per-phrase boosting weight multiplier applied on top of the decode-time boosting_tree_alpha,
+    # baked into the graph weights at build time; None -> 1.0 (global behavior)
+    boosting_tree_alpha: Optional[float] = None
```
artbataev: Please use `alpha` to avoid confusion with higher-level boosting tree weights, which are set in runtime

artbataev:
@AadamHaq Yep, case-insensitive boosting is merged, let's integrate everything with your customizations.
Please refactor the PR and resolve conflicts with the changed files. 
Ensure that both approaches for building graphs are supported (default and var-bpe/case-insensitive). 
If you need any help with var-bpe graph, please let me know - I will try to help with it.

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
```diff
@@ -35,8 +35,11 @@
 @dataclass
 class PhraseItem:
     phrase: str  # phrase itself
-    lang: str  # per-phrase language (for aggregate tokenizer)
-    # custom weight can be further added
+    lang: Optional[str] = None  # per-phrase language (for aggregate tokenizer); None -> cfg.source_lang
+    context_score: Optional[float] = None  # per-phrase score for each arc transition; None -> global behavior
```
artbataev: In what cases do we need to separate context score and phrase alpha?
Maybe it is ok to have this parameter, but also it may look confusing to the user who will need to tune more hyperparams.

AadamHaq: My team have found that for some particularly tricky words, we need to tune both params as an option. On the whole, global params work fine but in some cases we've noticed we need to tune either or both

artbataev: If I correctly understand, the proposed approach is to score token as `context_score * phrase_alpha` (instead of `context_score`). So, I still do not understand why we need to tune both separately for each phrase (one of them should be enough).

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
AadamHaq: Sure, makes sense. I've renamed it

AadamHaq:
@artbataev all the changes should be done, I've added the logic for the var-bpe/case-insensitive version too.
If there's any more issues just let me know

nemo/collections/asr/parts/context_biasing/boosting_graph_batched.py
artbataev: Please also note that https://github.com/NVIDIA-NeMo/Speech/pull/15800 introduced a substantial redesign for how scores are handled: previously we used token scores on arcs. Now `token_score` is a difference of potentials of nodes, `token_score = score(arc_dst) - score(arc_src)`, and it is computed on-the-fly (see `GPUBoostingTreeModel`). For just trie (which was constructed by `build(...)` method) this is fully equivalent handling of the weight, but for graph based on trie `build_from_var_bpe(...)` - it is important.

artbataev:
> @artbataev all the changes should be done, I've added the logic for the var-bpe/case-insensitive version too. If there's any more issues just let me know

@AadamHaq The PR looks good to me except the issue with separating 2 per-phrase parameters which are used in multiplication`token_score = context_score * phrase_alpha`, as I indicated above.

I see that there is another parameter `final_eos_score` which is adjusted with `phrase_alpha`, but it works only for AED models (final weight in the graph is not used with CTC/Transducer models) and used to prevent hallucinations.
If this is the scenario in which you see the need to modify it, I would still suggest removing per-phrase `context_score` and use per-phrase `final_eos_score` instead, since it would be more explicit.

E.g., instructions like "by default, tune per-phrase alpha, and for AED models additionally adjust `final_eos_score` if you face hallucinations" look much better than "by default, tune any of per-phrase `context_score` or `alpha` for CTC/Transducer models, but be careful with `final_eos_score` with AED models that is multiplied by `alpha` , so in case of problems when modifying alpha, modify only context score".

_Maybe I would also avoid multiplying `final_eos_score` by per-phrase alpha, since this is more like a universal hyperparam for preventing hallucinations for AED rather than important per-phrase parameter.  But this is up to you._

If there is any other scenario when you need to separate `context_score` - please let me know.

artbataev:
@andrusenkoau what do you think about modifying per-phrase `final_eos_score`?

AadamHaq:
Hi @artbataev 

Sorry for the delay. I was discussing/testing with my team. We decided to go ahead with what you said and are just tuning `alpha` and keeping `context_score` global. I updated the tests and docs to reflect this too.

On your point about the scoring redesign (`token_score` now being a difference of node potentials computed on-the-fly): per-phrase `alpha` is baked into the node potentials before they accumulate, so the on-the-fly arc weights come out as `alpha * (score(dst) -score(src))` so uniformly scaled, with no change needed in the `GPUBoostingTreeModel`/GPU layer. This preserves the merged/split equivalence in `build_from_var_bpe` (all tokenisations of a phrase still get the same total, just scaled), and the EOS `final_eos_score` is scaled by that same per-phrase alpha at the end node.

Also added a var-BPE test to confirm all  tokenisations score identically and scale exactly by `alpha`, EOS included

Let me know if there's anything else - the team thinks this should be enough for our WB needs for now but if we do come across another scenario where we need to change other params, I can let you know and do it in another PR

artbataev (approved):
Thanks for this feature and adjustments, @AadamHaq! Approved

PR #15941, opened 2026-07-23
Fix loading of MagpieTTS and add test for all 3 versions

blisc:
# What does this PR do ?

Fixes loading of released MagpieTTS checkpoints, which broke when the tokenizer
character/punctuation sets gained a second version, and adds coverage for all three
published releases.

**Collection**: tts

# Problem

`HindiCharsTokenizer` / `ArabicCharsTokenizer` gained `charset_version` / `punct_version`,
and the pt-BR `IPATokenizer` gained `locale_specific_punct`, in #15567 and #15614 — after
models had already been trained and released. A config that omits those fields is ambiguous:
it is either an archive predating them (vocabulary built with the v1 values) or a fresh
training config that should get today's defaults. Guessing wrong shifts every token ID and
surfaces as a `text_embedding` size mismatch during `restore_from`. That is how v2602 broke.

Inspecting the `model_config.yaml` inside each published `.nemo` shows the three shapes this
has to handle:

| release | `nemo_version` | version-sensitive tokenizers | fields pinned? |
|---|---|---|---|
| v2512 | `2.6.0rc0` | none | n/a — nothing to resolve |
| v2602 | `2.6.0rc0` | `HindiCharsTokenizer` | **no** → must resolve to `charset_version=1`, `punct_version=1` |
| v2607 | `2.8.0rc0` | Arabic ×3, pt-BR IPA | **yes** — `charset_version: 1`, `locale_specific_punct: false` |

# Solution

When a versioned tokenizer field is missing, decide it from the config's own `nemo_version` stamp: a config authored before the release that introduced the field was trained with the old value, anything newer gets the current default. The resolved values are written back into the config, so every checkpoint saved from here on is self-describing.

The rules live in one table (`VERSIONED_TOKENIZER_FIELDS`), one row per field, next to the default constants it reads — so adding a field or flipping a default is a single row, and the migration path cannot drift from the tokenizer signatures. Because the values get pinned at model construction, only the two model `__init__`s need the version; dataloader, worker, and inference paths read an already-unambiguous config, which lets the two hand-written migrations in `magpietts_inference/utils.py` be deleted.

Alongside that:

* Explicitly configured fields always win — that is what keeps v2607 loading correctly.
* A tokenizer/checkpoint mismatch now raises an actionable error naming the fields to pin, instead of an opaque `load_state_dict` size mismatch.
* `charset_version=1` / `punct_version=1` are documented as frozen for checkpoint compatibility rather than pending removal — v2607 pins v1, so removing it would break a shipped release.

# Changelog

- Fix loading of MagpieTTS and add test for all 3 versions
  
**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation


blisc:
/ok to test [7c3d29a](https://github.com/NVIDIA-NeMo/Speech/pull/15941/commits/7c3d29a1aa6dc54ad75fb0621bb9e9c5ec4c4373)

blisc:
/ok to test [6b177e6](https://github.com/NVIDIA-NeMo/Speech/pull/15941/commits/6b177e6f3841f170ec123e1f28ea3b15171de9f0)

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -66,32 +66,26 @@ def setup_tokenizers(all_tokenizers_config, mode='train'):
                 and not hasattr(tokenizer_config, 'locale_specific_punct')
             ):
                 with open_dict(tokenizer_config):
-                    tokenizer_config.locale_specific_punct = True
-            # Persist punct_version=2 for HindiCharsTokenizer so .nemo save/restore
-            # always uses the expanded punctuation set (with dandas).
+                    tokenizer_config.locale_specific_punct = not use_legacy_defaults
```
XuesongYang: `tokenizer_config.locale_specific_punct` was default `True`, but your new change flips to `False`. Is this behavior expected? was wondering if we should choose the default `use_legacy_defaults=False` instead?

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -66,32 +66,26 @@ def setup_tokenizers(all_tokenizers_config, mode='train'):
...
-                    tokenizer_config.punct_version = 2
-            tokenizer = safe_instantiate(tokenizer_config, **text_tokenizer_kwargs)
-            # TODO @xueyang: is it really necessary to set phone probability to 1.0 for test mode?
-            if mode == 'test' and hasattr(tokenizer, "set_phone_prob"):
-                tokenizer.set_phone_prob(1.0)
-
-            # Persist charset_version so it's saved in .nemo archives and
-            # update_config_for_inference can distinguish old checkpoints
-            # (missing charset_version → v1) from new ones.
+                    tokenizer_config.punct_version = 1 if use_legacy_defaults else 2
```
XuesongYang: Same concern as line 69 — new Hindi training now gets `punct_version=1` by default.

blisc: Ok how do you want to handle legacy checkpoint loading? #15567 broke loading the 2602 checkpoint because it doesn't have any of these parameters in the config and the tokenizer shape is now a mixmatch.

XuesongYang: these params should be configured with default values when any old model does nt have any. 

Let me fix this bug.

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -211,6 +211,11 @@ class EasyMagpieTTSInferenceModel(ModelPT):
     """
 
     def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
+        # Must be read before the audio codec is restored below, since that nested restore clears the
+        # global restore flag this relies on. Decides whether tokenizer fields missing from `cfg` fall
+        # back to their pre-versioning values (trained model) or to the current defaults (fresh run).
+        restored_model_config = is_restored_model_config(cfg)
```
blisc: I don't like this fix. Don't all nemo files has nemo_version persisted in it? Codex tried to do this too and it made no sense to me. Why should we have nemo_version as the check for being restored. And why does being restore imply that it is a legacy checkpoint?

nemo/collections/tts/data/text_to_speech_dataset_lhotse.py
```diff
@@ -39,10 +40,108 @@
...
+    otherwise; ``persist_versioned_tokenizer_defaults`` logs a warning naming the fields to set
+    explicitly, so the fallback is never silent.
+
+    Args:
+        model_cfg: The model-level config passed to ``__init__`` (before ``super().__init__`` stamps it).
+
+    Returns:
+        True if absent versioned fields should resolve to their pre-versioning values.
+    """
+    return 'nemo_version' in model_cfg
```
blisc:
```suggestion
def predates_versioned_tokenizer_fields(model_cfg) -> bool:
    """
    Fill this out
    """
    if 'nemo_version' not in model_cfg:
      return False  # use function defaults
    major_version = int(model_cfg["nemo_version"].split("."))
    return major_version < 3
```

nemo/collections/common/tokenizers/text_to_speech/tts_tokenizers.py
```diff
@@ -44,14 +47,128 @@
...
+    current defaults even though the fields landed mid-window -- genuinely undecidable from the stamp,
+    tie-broken toward current because ArabicCharsTokenizer was added by the very commit that added
+    ``charset_version`` (so for Arabic this reading is always right), and because guessing legacy
+    downgrades new training silently while guessing current fails loudly. Released checkpoints from
+    that window pin their fields anyway, so none of them reach here.
+    """
+    if cfg_nemo_version is None:
+        return False
+    try:
+        return Version(str(cfg_nemo_version)).release < Version(changed_in).release
```
XuesongYang, line 134: @blisc this is where we decide whether predating over release version.

blisc:
Approving and merging

PR #15947, opened 2026-07-23
fix: add missing f-string prefix in t5nmt error message

andrewwhitecdw:
Fixes a missing f-string prefix so the error message interpolates the cut object instead of printing literal `{cut}`.

- One-character fix in `nemo/collections/common/prompts/t5nmt.py`
- Added a regression test in `tests/collections/common/test_lhotse_prompt_format_data_types.py`

Tested:
```bash
uv run python -m pytest tests/collections/common/test_lhotse_prompt_format_data_types.py -v
# 11 passed
```


andrewwhitecdw:

PR #15948, opened 2026-07-24
fix: report ASR model type instead of name in unknown type error

andrewwhitecdw:
When `load_evaluation_models` receives an unsupported `asr_model_type`, the error message was printing the model name/path instead of the invalid type, making debugging harder.

- Fixed the interpolation in `nemo/collections/tts/modules/magpietts_inference/evaluate_generated_audio.py`
- Added a unit test verifying the error message contains the invalid type value

Tested:
```bash
uv run python -m pytest tests/collections/tts/modules/magpietts_inference/test_evaluate_generated_audio.py -v
# 1 passed
```

cc @blisc for TTS review


andrewwhitecdw:

PR #15950, opened 2026-07-24
fix: add missing f-string prefix in canary2 assertion

andrewwhitecdw:
Fixes a missing f-string prefix in the canary2 prompt formatter so the assertion message interpolates the EOS token and id instead of printing literal `{CANARY_EOS}` and `{eos}`.

- One-character fix in `nemo/collections/common/prompts/canary2.py`
- Added `tests/collections/common/test_canary2_prompt.py` with a regression test

Tested:
```bash
uv run python -m pytest tests/collections/common/test_canary2_prompt.py -v
# 1 passed
```

cc @nithinraok for review


andrewwhitecdw:

tests/collections/common/test_canary2_prompt.py
nithinraok, line 1: remove test

nithinraok, line 1: this shouldn't be under common

andrewwhitecdw:
Done — removed the test file (also resolves the "shouldn't be under common" note). The PR is now just the one-line f-string fix in `canary2.py`. Still happy to re-add a test in a different location if you'd prefer one.

nithinraok (approved):

nithinraok:
only comment change. CI tests not required

PR #15952, opened 2026-07-24
fix: avoid mutable default arguments in exp_manager timer callbacks

andrewwhitecdw:
Replaces mutable default arguments `{}` with `None` in `TimingCallback` and `DeltaTimingCallback`, then instantiates a fresh empty dict inside each constructor. This prevents shared state across instances if the dict is ever mutated.

- `nemo/utils/exp_manager.py`: changed `timer_kwargs={}` to `timer_kwargs: Optional[dict] = None` for both callbacks
- Added regression tests verifying fresh dicts are passed and the default is `None`

Tested:
```bash
uv run python -m pytest tests/core/test_exp_manager_timer_callbacks.py -v
# 2 passed
```

cc @nithinraok for core review


andrewwhitecdw:

tests/core/test_exp_manager_timer_callbacks.py
nithinraok, line 1: remove test

PR #15953, opened 2026-07-24
fix: avoid mutable default argument in ExactStringPerCategoryMatchMetric

andrewwhitecdw:
Replaces the mutable default argument `categories=[]` with `categories=None` and instantiates a fresh list inside the constructor. This prevents shared state across metric instances.

- `nemo/collections/common/metrics/classification_accuracy.py`: changed default and added None guard
- Added regression tests in `tests/collections/common/test_metrics.py`

Tested:
```bash
uv run python -m pytest tests/collections/common/test_metrics.py::TestCommonMetrics::test_exact_string_per_category_match_metric_default_categories_is_none tests/collections/common/test_metrics.py::TestCommonMetrics::test_exact_string_per_category_match_metric_tracks_categories -v
# 2 passed
```

cc @nithinraok for common/core review

### DIFF
--- a/nemo/collections/common/metrics/classification_accuracy.py
+++ b/nemo/collections/common/metrics/classification_accuracy.py
@@ -161,8 +161,10 @@ def compute_topk_accuracy(correct_counts_k, total_counts_k):
 
 
 class ExactStringPerCategoryMatchMetric(Metric):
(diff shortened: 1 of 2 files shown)


andrewwhitecdw:

nithinraok:
These are all blind AI generated PRs. Why do we need a test case to test None as default?

PR #15954, opened 2026-07-24
fix: correct 'occured' typo in neural type shape mismatch errors

andrewwhitecdw:
Fixes the spelling of "occurred" in four user-facing shape-mismatch error messages in `nemo/core/classes/common.py`.

- 2× "Input shape mismatch occured" → "Input shape mismatch occurred"
- 2× "Output shape mismatch occured" → "Output shape mismatch occurred"

Added regression tests that trigger both input and output shape mismatch paths and verify the corrected spelling.

Tested:
```bash
uv run python -m pytest tests/core/test_neural_type_shape_mismatch.py -v
# 2 passed
```

cc @nithinraok for core review

### DIFF
--- a/nemo/core/classes/common.py
+++ b/nemo/core/classes/common.py
@@ -538,7 +538,7 @@ class Typing(ABC):
 
                     if type_shape is not None and len(value_shape) != len(type_shape):
                         raise TypeError(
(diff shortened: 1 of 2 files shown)


andrewwhitecdw:

nithinraok:
Why do we need a test case for this?

PR #15957, opened 2026-07-24
fix(asr): catch RuntimeError when checking CUDA graphs conditional-node support

### LINKED ISSUE #15905: TDT/label-looping decoder crashes instead of falling back when libcuda is absent (e.g. ROCm builds of torch)
## Environment
- nemo_toolkit[asr] 2.x (pip, 2026-07), torch 2.9.1+rocm6.4 (AMD gfx1100; `torch.cuda` works via HIP)
- `cuda-python` installed (pulled in transitively)

## Repro
```python
m = nemo_asr.models.EncDecRNNTBPEModel.from_pretrained("nvidia/parakeet-tdt-0.6b-v3").to("cuda")
m.transcribe(["any.wav"])   # RuntimeError: Failed to dlopen libcuda.so.1
```

## Analysis
`label_looping_base.py::maybe_enable_cuda_graphs()` guards CUDA-graph support with
`check_cuda_python_cuda_graphs_conditional_nodes_supported()`, catching only
`(ImportError, ModuleNotFoundError, EnvironmentError)`. But on a system where
`cuda-python` is importable yet no NVIDIA driver exists (ROCm torch, CPU-only
boxes with cuda-python installed), `cuda.cuDriverGetVersion()` raises
`RuntimeError("Failed to dlopen libcuda.so.1")` from inside the C extension —
before NeMo's own `CUDA_SUCCESS` check — and nothing catches it.

## Expected
Same graceful degradation the function already provides when `cuda-python` is
missing: log a warning, decode without CUDA graphs (that path works perfectly —
we transcribed ~650 h of audio on ROCm this way).

## Suggested fix
Add `RuntimeError` to the except clause (or wrap the driver-version probe).
Not a ROCm-support request — purely a graceful-degradation gap.

## Workarounds (verified)
- `decoding.greedy.use_cuda_graph_decoder = False` via `change_decoding_strategy`, or
- uninstall `cuda-python`.

---
*Found while batch-transcribing ~650h of audio with parakeet-tdt-0.6b-v3 on an AMD gfx1100 workstation.*

mittalpk:
Fixes #15905

# What does this PR do ?

`maybe_enable_cuda_graphs()` crashes instead of falling back gracefully when `cuda-python` is importable but there's no real NVIDIA driver to talk to (e.g. ROCm/AMD builds of torch).

**Collection**: asr

# Changelog

- `nemo/collections/asr/parts/submodules/transducer_decoding/label_looping_base.py`: added `RuntimeError` to the exception types caught around `check_cuda_python_cuda_graphs_conditional_nodes_supported()` in `maybe_enable_cuda_graphs()`.

# Root cause

On systems where `torch.cuda.is_available()` returns `True` (e.g. ROCm/AMD builds of torch, which shim the CUDA API) but no real NVIDIA driver/`libcuda.so` is present, `check_cuda_python_cuda_graphs_conditional_nodes_supported()`'s call to `cuda.cuDriverGetVersion()` raises a raw `RuntimeError` ("Failed to dlopen libcuda.so.1") from inside the `cuda-python` native extension — before the function's own `EnvironmentError`/`ImportError` checks even run. `maybe_enable_cuda_graphs()` only caught `(ImportError, ModuleNotFoundError, EnvironmentError)`, so this `RuntimeError` propagated uncaught and crashed decoding instead of falling back to `CudaGraphsMode.NO_WHILE_LOOPS` as intended.

# Testing

Added `tests/collections/asr/decoding/test_label_looping_cuda_graphs_fallback.py` (3 tests, no real GPU/ROCm hardware needed — mocks `check_cuda_python_cuda_graphs_conditional_nodes_supported` directly):
- falls back to `NO_WHILE_LOOPS` when the check raises `RuntimeError` (the new case)
- still falls back on the previously-handled error types (`EnvironmentError`)
- still enables `FULL_GRAPH` mode when the check succeeds

Verified the new `RuntimeError` test fails against the pre-fix code and passes against the fix. Ran the full `tests/collections/asr/decoding/` + `tests/collections/common/test_optional_cuda_graphs.py` suites (283 passed). One unrelated pre-existing failure (`test_ctc_decoding.py::test_batched_decoding_labels`) reproduces identically with and without this change — confirmed by running it both in isolation and as part of the full suite on unmodified `main`, so it's not something this PR introduces.

**Pre checks**:
- [ ] Did you add or update any necessary documentation? (not applicable — no user-facing behavior/API changed, only an internal fallback's exception handling)

### DIFF
--- a/nemo/collections/asr/parts/submodules/transducer_decoding/label_looping_base.py
+++ b/nemo/collections/asr/parts/submodules/transducer_decoding/label_looping_base.py
@@ -139,7 +139,7 @@ class GreedyBatchedLabelLoopingComputerBase(WithOptionalCudaGraphs, ABC):
             try:
                 check_cuda_python_cuda_graphs_conditional_nodes_supported()
                 self.cuda_graphs_mode = self.CudaGraphsMode.FULL_GRAPH
-            except (ImportError, ModuleNotFoundError, EnvironmentError) as e:
+            except (ImportError, ModuleNotFoundError, EnvironmentError, RuntimeError) as e:
                 logging.warning(
                     "No conditional node support for Cuda.\n"
                     "Cuda graphs with while loops are disabled, decoding speed will be slower\n"
--- a/tests/collections/asr/decoding/test_label_looping_cuda_graphs_fallback.py
+++ b/tests/collections/asr/decoding/test_label_looping_cuda_graphs_fallback.py
@@ -1,0 +1,95 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Regression test: maybe_enable_cuda_graphs() must fall back gracefully, not
+crash, when cuda-python is importable but there is no real NVIDIA driver to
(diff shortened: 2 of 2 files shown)


mittalpk:
Checking in here, happy to make any changes if that helps move review along.

PR #15966, opened 2026-07-24
[TTS] Add codebook dropout in codec training

rlangman:
# What does this PR do ?

Adds optional codebook dropout during audio codec training.

Training with a dropout of ~10% enables an FSQ based codec to reconstruct audio with any number of codebooks (quality increasing with more codebooks) with only a very small drop in performance when decoding all codebooks. Higher dropout rates produce a significantly larger drop in overall performance.

**Collection**: [TTS]

# Changelog

- Add parameter for randomized codebook dropout during training
- Add parameter to `encode()` making it easy to do inference with fewer codebooks
- Add a typecast in `SLMEncoder` necessary to do semantic codec training in BF16

**Pre checks**:

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/models/audio_codec.py
+++ b/nemo/collections/tts/models/audio_codec.py
@@ -39,11 +39,7 @@ from nemo.collections.tts.losses.audio_codec_loss import (
     SISDRLoss,
     TimeDomainLoss,
 )
-from nemo.collections.tts.modules.audio_codec_modules import (
-    GroupFiniteScalarQuantizer,
-    ResNetSpeakerEncoder,
-    default_precision,
-)
+from nemo.collections.tts.modules.audio_codec_modules import ResNetSpeakerEncoder, default_precision
 from nemo.collections.tts.modules.common import GaussianDropout
 from nemo.collections.tts.parts.utils.callbacks import LoggingCallback
 from nemo.collections.tts.parts.utils.helpers import get_batch_size, get_num_workers
@@ -101,6 +97,7 @@ class AudioCodecModel(ModelPT):
 
         if "vector_quantizer" in cfg:
             self.vector_quantizer = safe_instantiate(cfg.vector_quantizer)
+            self.codebook_dropout_rate = cfg.get("codebook_dropout_rate", 0.0)
 
             vq_output_types = list(self.vector_quantizer.output_types.keys())
 
@@ -110,13 +107,10 @@ class AudioCodecModel(ModelPT):
             else:
                 self.vector_quantizer_has_commit_loss = False
                 logging.info('Vector quantizer does not support commit loss.')
-
-            self.codebook_dropout_rate = cfg.get("codebook_dropout_rate", 0.0)
-            if self.codebook_dropout_rate and not isinstance(self.vector_quantizer, GroupFiniteScalarQuantizer):
-                raise ValueError("Codebook dropout only supported for GroupFiniteScalarQuantizer")
         else:
             logging.warning('Vector quantizer will not be used.')
             self.vector_quantizer = None
+            self.codebook_dropout_rate = 0.0
 
         # Decoder setup
         self.audio_decoder = safe_instantiate(cfg.audio_decoder)
@@ -466,7 +460,9 @@ class AudioCodecModel(ModelPT):
         encoded, encoded_len = self.encode_audio(audio=audio, audio_len=audio_len, sample_rate=sample_rate)
 
         if num_codebooks:
-            encoded = self._dropout_codebooks(encoded=encoded, num_codebooks=num_codebooks)
+            assert self.vector_quantizer is not None
+            num_codebooks_batch = num_codebooks * torch.ones([encoded.shape[0]])
+            encoded = self.vector_quantizer.dropout_codebooks(encoded=encoded, num_codebooks=num_codebooks_batch)
 
         # Apply quantizer to obtain discrete representation per frame
         tokens = self.quantize(encoded=encoded, encoded_len=encoded_len)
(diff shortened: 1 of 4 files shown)


blisc (dismissed):
LGTM, can you run tests?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -578,6 +635,9 @@ def _process_batch(self, batch):
         else:
             commit_loss = 0.0
 
+        if self.training and self.codebook_dropout_rate:
```
vklimkov-nvidia, line 606: will fail with AttributeError for no-quantizer path. maybe unconditionally initialize codebook_dropout_rate in constructor? in `else` we already have vector_quantizer = None

nemo/collections/tts/modules/audio_codec_modules.py
```diff
@@ -243,6 +243,7 @@ def forward(self, audio):
             slm_emb = out.hidden_states[self.hidden_layer] / self.scaling_factor
 
         slm_emb = rearrange(slm_emb, 'B T D -> B D T')
+        slm_emb = slm_emb.to(audio.dtype)
```
vklimkov-nvidia: we always go to float32 here. is it intentional?

nemo/collections/tts/models/audio_codec.py
```diff
@@ -550,6 +568,45 @@ def preprocess_audio(self, audio, audio_len, sample_rate):
         audio, audio_len = self.pad_audio(audio=audio, audio_len=audio_len, samples_per_frame=self.samples_per_frame)
         return audio, audio_len
 
+    def _dropout_codebooks(self, encoded, num_codebooks):
```
vklimkov-nvidia: called in generic encode but makes sense only for fsq

nemo/collections/tts/models/audio_codec.py
```diff
@@ -550,6 +568,45 @@ def preprocess_audio(self, audio, audio_len, sample_rate):
...
+        Args:
+            encoded: encoder output (B, D, T)
+            num_codebooks: number of codebooks to keep. This can be an integer, or a 3-D tensor compatible with the
+                dimensions of the encoder output
+
+        Returns:
+            Encoder output with all codebooks above index 'num_codebooks' masked out
+        """
+        batch_size, embed_dim, num_frames = encoded.shape
+        emb_unmask_dim = self.vector_quantizer.codebook_dim_per_group * num_codebooks
```
vklimkov-nvidia: codebook_dim_per_group might not exist

nemo/collections/tts/models/audio_codec.py
rlangman: Changed to set `codebook_dropout_rate` to 0 when `vector_quantizer` is None

nemo/collections/tts/modules/audio_codec_modules.py
rlangman: So far as I can tell the w2v-bert feature extractor is a CPU based extractor that needs float32. The encoder network can be run in other precisions. Changed the casting to only effect the feature extraction portion.

nemo/collections/tts/models/audio_codec.py
rlangman: Changed codebook dropout to a class method. It is implemented for FSQ, but if someone tries to call it for any other VQ class (by setting `codebook_dropout_rate > 0` or running encode with `num_codebooks` parameter) it will throw an error. This way someone could also implement the dropout for RVQ if they want.

nemo/collections/tts/models/audio_codec.py
```diff
@@ -428,28 +429,41 @@ def dequantize(self, tokens: torch.Tensor, tokens_len: torch.Tensor) -> torch.Te
...
+        num_codebooks: Optional[int] = None,
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         """Convert input time-domain audio signal into a discrete representation (tokens).
 
         Args:
             audio: input time-domain signal, shape `(batch, number of samples)`
             audio_len: valid length for each example in the batch, shape `(batch size,)`
             sample_rate: sample rate of input audio (int)
+            num_codebooks: number of codebooks to use for reconstructing audio.
+                Using fewer codebooks will only be accurate for codecs trained with FSQ codebook dropout.
```
rfejgin: Can drop the word "FSQ" here since the API permits any quantizer to implement the dropout (and we know it can be done for RVQ)

nemo/collections/tts/models/audio_codec.py
```diff
@@ -428,28 +429,41 @@ def dequantize(self, tokens: torch.Tensor, tokens_len: torch.Tensor) -> torch.Te
...
             Tokens for each codebook for each frame, shape `(batch, number of codebooks, number of frames)`,
             and the corresponding valid lengths, shape `(batch,)`
         """
         # Apply encoder to obtain a continuous vector for each frame
         encoded, encoded_len = self.encode_audio(audio=audio, audio_len=audio_len, sample_rate=sample_rate)
+
+        if num_codebooks:
+            assert self.vector_quantizer is not None
+            num_codebooks_batch = num_codebooks * torch.ones([encoded.shape[0]])
+            encoded = self.vector_quantizer.dropout_codebooks(encoded=encoded, num_codebooks=num_codebooks_batch)
```
rfejgin, line 465: If we ever try to do this for RVQ, I'm not sure doing dropout after the fact will work since the codebooks are not separated by dimension; i.e. this would have to be part of `encode_audio` API. Not crucial to change now.

rfejgin (commented):
Overall looks good to me, see some comments inline.

How about adding a test? We could test e.g. that
(1) `self.vector_quantizer.dropout_codebooks()` indeed returns the relevant codebooks zeroed out.
(2) `encode()` with `num_codebooks=None` gives the same result as `encode()` with `num_codebooks=self.num_codebooks`

nemo/collections/tts/models/audio_codec.py
rfejgin: Could even consider adding a warning if the codec's config does not indicate that it was trained with dropout, and we attempt to infer with the num_codebooks parameter set to a less than self.num_codebooks

nemo/collections/tts/models/audio_codec.py
rlangman: That is true.

rfejgin:
> Overall looks good to me, see some comments inline.
> 
> How about adding a test? We could test e.g. that (1) `self.vector_quantizer.dropout_codebooks()` indeed returns the relevant codebooks zeroed out. (2) `encode()` with `num_codebooks=None` gives the same result as `encode()` with `num_codebooks=self.num_codebooks`

@rlangman Have you considered the above? I swear that Codex will do this kind of thing in a heartbeat :)

rfejgin (approved):
Looks good to me!

PR #15976, opened 2026-07-28
docs: qualify SALM Automodel parallelism support

DongjiGao:
## What changed

- Clarify that `AutomodelParallelStrategy` can construct meshes for FSDP2, TP, PP, CP, EP, and HSDP, while actual support remains model-, backend-, and combination-dependent.
- Document the validated native Nemotron-V3 SALMAutomodel envelope: FSDP2, EP, and CP with packed sequences and Transformer Engine attention.
- Direct users to keep `tp_size`, `dp_replicate_size`, and the currently unvalidated `pp_size` at 1 for this configuration.

## Why

The previous blanket support statement drove QA and users toward unsupported Nemotron-V3 configurations. Nemotron-V3 has no TP plan, while the SALMAutomodel HSDP path is unsupported and currently fails with an opaque DeviceMesh error. PP is capability-signaled by Automodel but has no SpeechLM2 training validation.

This is a documentation correction only; it does not change parallelism behavior. The HSDP validation/runtime issue remains separate.

## Validation

- `sphinx-lint docs/source/speechlm2/training_and_scaling.rst` — no problems found
- `git diff --check`

Related: internal bug 6437904.

### DIFF
--- a/docs/source/speechlm2/training_and_scaling.rst
+++ b/docs/source/speechlm2/training_and_scaling.rst
@@ -101,8 +101,10 @@ AutomodelParallelStrategy (SALMAutomodel)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 For ``SALMAutomodel``, the collection provides ``AutomodelParallelStrategy`` which
-delegates device mesh creation and parallelism to NeMo Automodel. This strategy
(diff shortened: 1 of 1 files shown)


blisc (approved):
Makes sense to me

blisc:
Do you want to merge or keep as draft?

DongjiGao:
> Do you want to merge or keep as draft?

I think this one is good to merge.

PR #16001, opened 2026-07-30
Align CUDA 12 and DeepEP dependencies in the Automodel container

pzelasko:
## Summary

- Align the CUDA 12 extra and TorchCodec with the CUDA 12.9 base by using PyTorch 2.11/cu129.
- Pin DeepEP to the exact revision expected by the pinned Automodel revision and regenerate the lockfile, including its NVML dependencies.
- Prefer the locked wheel libraries at runtime so PyTorch and compiled extensions resolve one CUDA dependency set.
- Apply the narrow shared-expert main-stream correctness fix from Automodel commit `c2571088` to the pinned Automodel package during the image build.
- Regenerate the A100 DeepEP patch for the pinned revision and strengthen the compiled-dependency smoke test with a configurable world size and three consecutive forward/backward iterations.

## Root cause

The container selected DeepEP through the `v1.2.1` tag while the pinned Automodel revision declares an exact DeepEP commit. Those references resolve to different dispatcher implementations despite sharing the same public version. Under expert parallelism, that mismatch combines with Automodel's shared-expert side stream to create a race in the dispatch path.

The CUDA 12 image also combined a CUDA 12.9 base/toolkit with cu126 PyTorch and TorchCodec wheels, allowing a different set of CUDA shared libraries to win at runtime than the extensions were built against.

## Validation

- `uv lock --check`
- Ruff, Python compilation, shell syntax, and staged whitespace checks
- Patch applicability checks against the exact Automodel and DeepEP revisions
- Full CUDA 12.9 H100+ Docker build, including the in-image Automodel patch assertion
- In-image GPU ABI probe with PyTorch `2.11.0+cu129`, CUDA 12.9, cuDNN 9.17.1, NCCL 2.28.9, NVSHMEM 3.4.5, and DeepEP `1.2.1+7febc6e`
- Multi-GPU H100 expert-parallel smoke with three consecutive forward/backward iterations using the full launched world size

## Known non-target limitation

The existing FlashAttention 2.8.3 backward path on SM120 currently raises an upstream `dQ_single_wg` initialization error. This is independent of the H100 expert-parallel fix; the image's CUDA ABI/version probe succeeds on SM120.


pzelasko:
/ok to test [90b0543](https://github.com/NVIDIA-NeMo/Speech/pull/16001/commits/90b0543ed6b01483657f844286560bc2293bc976)

pzelasko:
/ok to test [25a52a4](https://github.com/NVIDIA-NeMo/Speech/pull/16001/commits/25a52a432aaba2dfa5fcdd710357f7e07a1c31b1)

chtruong814 (dismissed):

blisc (approved):

PR #16015, opened 2026-08-03
Support multi-system TTS comparison reports

vklimkov-nvidia:
## What changed

- allow `--candidate_name` and `--candidate_path` to be repeated for N-system comparisons
- render metrics, configurations, box plots, and matched audio for every system
- run the existing Mann–Whitney statistical procedure for every unordered system pair
- preserve the original two-system CLI, statistical summary behavior, generated prefix, and audio object names
- add optional S3 prefix override and complete local report output

## Why

The report generator was limited to one baseline and one candidate, which prevented side-by-side codec comparisons involving three or more systems.

## Validation

- `conda run -n nemotts pytest -q tests/collections/tts/test_tts_comparison_report.py` (13 passed)
- Black check
- isort check
- Python bytecode compilation
- `git diff --check`

### DIFF
--- a/scripts/tts_comparison_report/README.md
+++ b/scripts/tts_comparison_report/README.md
@@ -2,7 +2,8 @@
 
 This tool generates HTML comparison reports for TTS evaluation buckets and uploads them to S3.
 
-The `generate_report` script compares two evaluation buckets produced by `magpietts_inference` and generates:
+The `generate_report` script compares a baseline with one or more candidate evaluation buckets produced by
+`magpietts_inference` and generates:
 1. an HTML evaluation report with aggregated and per-benchmark metrics;
 2. an optional HTML audio comparison report with side-by-side audio samples.
 
@@ -48,7 +49,7 @@ experiment_root/
     └── benchmark_3
 ```
 
-In both cases, `--baseline_path` and `--candidate_path` should point to the experiment root. 
+In both cases, `--baseline_path` and each `--candidate_path` should point to an experiment root.
 If evaluation artifacts are stored directly inside the experiment root, set `--results_subdir` 
 to an empty string. If evaluation artifacts are stored under a dedicated subdirectory, 
 set `--results_subdir` to that subdirectory name.
@@ -110,6 +111,10 @@ python scripts/tts_comparison_report/generate_report.py \
 
 If the evaluation artifacts are stored in a non-default results subdirectory, 
 use `--results_subdir`.
+
+For more than two systems, repeat `--candidate_name` and `--candidate_path` for each additional candidate.
+Use `--s3_prefix` to override the generated S3 directory name, or `--local_output_dir` to create the report
+locally without uploading.
 
 To generate both the evaluation report and the audio comparison report, use `--audio_report`.
 You can also use `--audio_report_benchmarks` and `--samples_per_benchmark` 
@@ -178,7 +183,7 @@ only artifacts from repetition `0`.
 the experiment root that contains evaluation outputs such as metrics and generated audio. 
 If evaluation artifacts are stored directly inside the experiment root, `--results_subdir` 
 should be set to an empty string.
-- Both generated reports are HTML reports uploaded to S3-compatible object storage.
+- Reports are uploaded to S3-compatible object storage unless `--local_output_dir` is used.
 - If the audio report is enabled, the evaluation report includes a link to the audio report.
 - Audio files referenced by the audio report are uploaded separately and linked through presigned URLs.
 - Box plot images are also uploaded to S3 and embedded into the evaluation report via presigned URLs.
--- a/scripts/tts_comparison_report/generate_report.py
+++ b/scripts/tts_comparison_report/generate_report.py
@@ -64,14 +64,16 @@ def _create_argparser() -> ArgumentParser:
     parser.add_argument(
         "--candidate_name",
         type=str,
+        action="append",
         required=True,
-        help="Name of the candidate model that will be used in report.",
+        help="Name of a candidate model used in the report. Repeat with --candidate_path to compare more systems.",
     )
     parser.add_argument(
         "--candidate_path",
         type=str,
+        action="append",
         required=True,
-        help="Path to the generated evaluation bucket for the candidate model.",
+        help="Path to a candidate evaluation bucket. Repeat with --candidate_name to compare more systems.",
     )
     parser.add_argument(
         "--benchmarks",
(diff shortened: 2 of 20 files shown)


blisc (commented):
Generally, LGTM. But should we limit the number of items that can be added? I'm concerned that the report and the statistical comparisons might become too bloated. Deferring to @artem-gorodetskii for final approval.

artem-gorodetskii:
@blisc, @vklimkov-nvidia I have a few concerns regarding this PR.

The purpose of this tool is to standardize comparison between a baseline model and a proposed candidate model, making the process closer to how it is done in production teams and moving away from an unstructured workflow.

In the current design, the tool has a clear semantic that is very close to A/B testing:
* baseline = current reference or production-like model,
* candidate = one hypothesis being evaluated,
* report = one decision artifact for that hypothesis, used to justify whether the candidate configuration should be accepted into the production workflow.

Later, this report can serve as evidence that a specific model, configuration or training-related change was valid. It also serves as a record of what was tested and why it was rejected or accepted.


If we move to comparing many candidates in a single report, then the purpose of the tool changes from a decision report into a **research comparison dashboard**. In my view, this is too far from the original A/B-testing-style semantic and leads to several problems:
* It breaks the original meaning of the report. Originally, the report answers a simple question: "Is this candidate better than the baseline, and should we move forward with it?" That is a simple but very strong and useful workflow.
* It weakens the report as a production artifact. A single baseline-vs-candidate report is easy to review, attach to a ticket, discuss and use as a promotion artifact. A multi-candidate  report is much harder to use as a clean decision record because it mixes several hypotheses together, makes conclusions less focused and is no longer tied to one specific decision.
* It makes the interface heavier. Once we have many candidates, the report will contain more columns, more metrics to compare visually and more audio samples. This can quickly become overwhelming. From a UX perspective, I do not think this is a good direction.
* It changes the tool from standardized production evaluation to research exploration. A standardized binary decision workflow is easy to review and track and it always clearly reflects the validity of a particular decision. If we add a multi-candidate mode into the same report, it can easily become a research zoo, which makes the workflow unstructured.

A multi-candidate report may look convenient if you compare a baseline against only two candidates. But even in that case, I would still argue that two separate reports are better. Also, in practice we often still need an additional comparison between the strongest candidates, so we end up running another comparison anyway.

And more importantly: how should this functionality be used when we need to validate a large number of candidates? This is not a theoretical question - it already happened in the difficulty-aware sampling work, where we ran around 100 evaluations. In that scenario, a single multi-candidate report would not be practical.


So I think the better design is to preserve the current production-style A/B-testing semantic, and for multi-candidate evaluation implement a separate CLI, for example `generate_batch_report.py`, which would generate one report for each baseline-candidate pair.

This should require only a relatively small refactoring of the `Orchestrator` class:
* move the baseline loading logic into a separate method, for example `orchestrator.setup()`
* keep the rest of the per-candidate logic in `orchestrator.run()`

This would allow us to load the baseline data once and then generate multiple binary reports efficiently, while preserving the original standardized A/B-testing semantics.


I also have two important concerns regarding `s3_prefix_override` and `local_output_dir`:
* `s3_prefix_override`. I do not think we should allow arbitrary changes to the S3 prefix. It is better to store all generated reports in one standardized location. This makes the workflow easier to manage and keeps the process transparent. All these reports related to one model and its POR process, so they should be stored in one place. Otherwise, we will end up with chaos in S3 because someone decided to place the report in a personal folder. That is not acceptable, because the report is an artifact and decisions based on that artifact affect the entire development process.
* `local_output_dir`. I have the same concern here. This functionality encourages a workflow where developers keep decision artifacts on their personal machines, which then become inaccessible to others. That breaks transparency and reduces reliability.  It is better to store decision artifacts in one shared place so that everyone has the opportunity to verify the validity of previous decisions. This is very important for long-term development.

vklimkov-nvidia:
i feel like concerns above are about usage of the tool not about tool itself.
change is additive and doesn't modify the original A/B workflow. so lets say it doesnt "change" meaning of report but "extends" :) if we agree that for prod decisions we only use 2-systems reports thats ok with me. For day to day work i definitely want to have several systems on the same page. My only concern currently is that pairwise statistical significance currently reported for each pair, that seems excessive.

currently one can just specify "candidate" multiple times. i think in such setup there is no need for separate client. but its your call. i am happy to refactor to introduce a separate client.

- s3_prefix_override - i've seen the opposite already where the POR is copypasted and unrelated reports end up in the same bucket. also i wanted a demo page for effort without POR yet.
- local_output_dir - its to check how report looks, do some side-by-side listening yourself before uploading. consider it a dryrun

artem-gorodetskii:
Thanks, I see your point.

The main value of this tool is not simply that it can generate a report, but that it enforces a standardized decision workflow.

Yes, with multi-candidate support in the same report, the tool would still allow the original binary workflow, but it would no longer strongly guide users toward it. Instead, it would start to encourage  unstandardized workflow. I think this is important, especially for “day-to-day work”: the purpose of this comparison tool is to encourage developers to rely on production-style practices during applied research (use A/B testing). The original version serves this purpose well. It supports a simple and disciplined workflow: accept or reject one proposed configuration, provide supporting evidence and show p-values or other statistical metrics as part of that decision.

I also think this change weakens the design of the tool. With this extension, the tool would start to serve two different workflows:
* creating production-oriented decision artifacts,
* creating an exploratory research dashboard.

These two purposes should not be combined in the same tool. Over time, this would make the reporting tool less focused and push it toward becoming a research script, which would be harder to maintain and extend.

For these reasons, I do not support strict multi-candidate comparison in the current tool.

Regarding report storage:
* Publishing reports under a defined POR or Jira ticket should remain the developer’s responsibility.  If a developer does not want to use a real ticket, the report can still be published under the default task ID. That is acceptable for demo use cases. We can assume that reports without a real ticket do not have long-term value and may be removed if necessary. So it is in the developer’s interest to attach reports to a proper ticket. Even if a POR ticket does not yet exist, the developer can still create a Jira task and use that task ID.
* If a dry run is required, I think that should be treated as a manual testing workflow. In that case, I would suggest adding a separate scripts/tts_comparison_report/debug_scripts directory with helper scripts that save only local audio reports, with restricted metadata such as fixed model names like A and B. This would support dry-run listening while still preserving the main publishing workflow and the role of the report as a shared production artifact.

vklimkov-nvidia:
Thanks for clarifying, Artem. I hadn’t realized that enforcing the standardized A/B decision workflow is an intentional part of the tool’s design, rather than simply its current usage pattern. With that context, I’m happy to leave the tool as-is and close this PR.
For exploratory research comparisons, I’ll keep the multi-system script in my garage repo. If others on the team encounter the same need, we can revisit whether a separate research-oriented tool would be useful.

artem-gorodetskii:
Thanks, I appreciate it. I think a separate research dashboard outside this repo would be the right place for this functionality. It could be implemented either as a hosted service or as a simpler first step, as a static web page published through GitLab Pages. I think that would be a better architectural fit for this use case and would let us extend research-oriented functionality without weakening the original A/B-style semantics of the reporting tool. I would be happy to contribute to it.

PR #16030, opened 2026-08-05
[TTS] EasymagpieTTS Online CFG Distillation

artem-gorodetskii:
## Summary

This PR adds online classifier-free guidance (CFG) distillation for EasyMagpieTTS, including local transformer (LT) distillation, phoneme guidance, rollout rejection, and single-turn/multiturn checkpoint support.

## Summary by CodeRabbit

- **New Features**
  - Added online CFG distillation stage for Easy Magpie TTS.
  - Added configurable frame-major and codebook-major ordering for distillation losses.
  - Fixed KL-divergence normalization across individual codebook/frame slices.

- **Testing**
  - Added an end-to-end functional test covering online CFG distillation training.
  - Included the new scenario in the continuous integration test suite.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,26 @@
+{
+  "name": "NeMoTTS Dev Container",
+  "build": {
+    "dockerfile": "../.devcontainer/devDockerfile"
+  },
+  "runArgs": [
+    "--gpus",
+    "\"device=0,1\"",
+    "--ipc=host",
+    "--name",
+    "${localEnv:USER}_nemo_tts_devcontainer_easymagpietts_online_cfg_distillation"
+  ],
+  "workspaceFolder": "/workspace/NeMo",
+  "mounts": [
+    "source=${localWorkspaceFolder},target=/workspace/NeMo,type=bind"
+  ],
+  "remoteUser": "root",
+  "postCreateCommand": "pip install -e /workspace/NeMo --no-deps && pip install git+https://github.com/sarulab-speech/UTMOSv2.git@v1.2.1",
+  "customizations": {
+    "vscode": {
+        "settings": {
+            "terminal.integrated.shell.linux": "/bin/bash"
+        }
+    }
+  }
+}
--- a/nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
+++ b/nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
@@ -1,0 +1,4270 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Adapted EasyMagpieTTSModel class for classifier-free guidance distillation.
+"""
+import copy
+import json
+import random
+from dataclasses import dataclass, field
+from enum import Enum
+from io import TextIOWrapper
+from pathlib import Path
+from typing import Callable, Optional
+
+import numpy as np
+import soundfile as sf
+import torch
+import wandb
+from lightning.pytorch import Trainer
+from lightning.pytorch.loggers import TensorBoardLogger, WandbLogger
+from lightning.pytorch.utilities import rank_zero_only
+from omegaconf import DictConfig, OmegaConf, open_dict
+from torch import Tensor
+from transformers import WhisperForConditionalGeneration, WhisperProcessor
+
+import nemo.collections.asr as nemo_asr
+from nemo.collections.asr.metrics.wer import word_error_rate
+from nemo.collections.asr.parts.mixins.transcription import TranscribeConfig
+from nemo.collections.tts.losses.magpietts_cfg_distillation import (
+    CodesCrossEntropyLoss,
+    KLDivergenceLoss,
+    NRMSELogitsLoss,
+)
+from nemo.collections.tts.models.easy_magpietts import EasyMagpieTTSModel
+from nemo.collections.tts.models.easy_magpietts_inference import InferBatchOutput, StreamingConfig, TrainingMode
+from nemo.collections.tts.modules.magpietts_modules import (
+    LocalTransformerType,
+    clear_forbidden_logits,
+    remove_special_tokens,
+)
+from nemo.collections.tts.parts.utils.helpers import (
+    get_mask_from_lengths,
+    get_speaker_embeddings_from_filepaths,
+    process_text_for_cer,
+    transcribe_with_whisper,
+    transcribe_with_whisper_from_filepaths,
+)
+from nemo.lightning.callback_group import CallbackGroup
+from nemo.utils import logging
+
+try:
+    from nemo.collections.tts.modules.utmosv2 import UTMOSv2Calculator
+
+    _HAVE_UTMOSV2 = True
+except (ImportError, ModuleNotFoundError):
+    _HAVE_UTMOSV2 = False
+
+__all__ = ["EasyMagpieCFGDistillation"]
+
+
+_STATE_DICT_EXCLUDE_NAMES: list[str] = ["_teacher_model"]
+
+
+@dataclass
+class _DefaultParams:
+    """Default hyperparameters for EasyMagpieTTS CFG distillation."""
+
+    # Maximum number of decoding steps during audio rollout generation.
+    max_decoder_steps: int = 300
+    # Sampling temperature during rollout generation.
+    rollout_temperature: float = 0.7
+    # Top-k sampling limit for token selection.
+    rollout_topk: int = 80
+    # Whether to use key-value cache during rollout inference.
+    use_kv_cache_during_rollout: bool = True
+    # Classifier-free guidance (CFG) scale used during distillation.
+    distillation_cfg_scale: float = 2.5
+    # Temperature used for softening logits during distillation (1.0 means no change).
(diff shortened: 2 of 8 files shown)


artem-gorodetskii:
@coderabbitai review

.devcontainer/devcontainer.json
XuesongYang, line 1: we don't need to merge anything under `.devcontainer` to the main branch, right? so far as I know, it is only used for running Cursor IDE.

If your recipe requires some dependency built in docker container, we may need to add those in https://github.com/NVIDIA-NeMo/Speech/blob/main/docker/Dockerfile

artem-gorodetskii: This is a local development environment, I will delete it when I change the PR status from draft

artem-gorodetskii:
@coderabbitai generate sequence diagram for the training workflow inside nemo/collections/tts/models/easy_magpietts_cfg_distillation.py

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,3288 @@
...
+    dummy_embeds_uncond: Tensor
+    cache_position: Tensor
+    audio_codes: Tensor
+    audio_codes_lens: Tensor
+
+
+@dataclass
+class _GroundTruthPhonemeEmbeds:
+    embeds: Optional[Tensor]
+    lens: Optional[Tensor]
```
blisc: Why do you prefer to have a dataclass with two elements as opposed to returning two tensors from a function?

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,4232 @@
...
+        return self._is_active
+
+    @property
+    def use_multilingual_asr(self) -> bool:
+        """Get whether validation uses the multilingual ASR model."""
+        return self.params.use_multilingual_asr
+
+    def _setup_asr_model(self) -> None:
+        """Initialize the ASR model used for validation transcription."""
+        if self.params.use_multilingual_asr:
```
blisc: Why define a property if you don't plan on using it?

Either not do define the property and use params.use_multilingual_asr, or keep the property and use it.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,4232 @@
...
+            self.whisper_model = WhisperForConditionalGeneration.from_pretrained("openai/whisper-large-v3")
+            self.whisper_model.eval()
+            self.whisper_model.to(self.default_device)
+            self.nemo_asr_model = None
+
+            for p in self.whisper_model.parameters():
+                p.requires_grad = False
+
+        else:
+            self.nemo_asr_model = nemo_asr.models.EncDecRNNTBPEModel.from_pretrained("nvidia/parakeet-ctc-0.6b")
```
blisc: I think you'd be better off passing these through the params rather than hard-coding it. ctc-0.6b is also an odd choice to hard-code. I would have used 1.1b-tdt

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,4232 @@
...
+            torch.cuda.empty_cache()
+
+        self._is_active = True
+
+    def switch_device(self, device: torch.device | str) -> None:
+        """Move loaded validation models to a target device.
+
+        Args:
+            device: Target device for the loaded ASR and speaker-verification models.
+        """
```
blisc, line 2451: I'm surprised that you didn't run into my buffer issues that resulted in this function: https://github.com/NVIDIA-NeMo/Speech/pull/16035/changes#diff-313ec9e7123666a0330dead9717b70459284e851ffbf8ddb1ae886c3c72b35baR168-R173

Is it because you only load models on_fit_start() instead of init()?

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,4232 @@
...
+            embeddings=[zero_delay_tensor, embedded],
+            lengths=[delay, audio_codes_lens],
+        )
+        return embedding, lens
+
+    def _configure_training_mode(
+        self,
+        mode: str,
+        training_mode: Optional[TrainingMode],
+    ) -> TrainingMode:
```
blisc, line 3112: What does this function do?

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
```diff
@@ -0,0 +1,4232 @@
...
+
+        Returns:
+            Mapping containing the total loss, enabled component losses and the
+            rejection-rate metric when applicable.
+        """
+        # The multiturn dataset inserts interruption markers aligned to the ground-truth audio.
+        # This distillation path is intentionally restricted to ordinary single-turn TTS cuts.
+        if self.cfg.get("use_multiturn_dataset", False):
+            text = batch["text"]
+            batch["text"] = text.masked_fill(text == self.interruption_token_id, self.pad_id)
```
blisc, line 3897: According to the comment, we do not support distillation on multi-turn. Should this code path throw an error? Does this path actually work on multi-turn data?

blisc (commented):
Most of the PR looks good to me. Left a few comments, mainly on style choices. One final comment on style: Should we only keep the three main classes (_TeacherInferenceEngine, _ValidationInferenceEngine, EasyMagpieCFGDistillation‎) in the total level easy_magpietts_cfg_distillation.py and move the private dataclasses and functions elsewhere?

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: I used a dataclass because the two values form one logical result and named fields made their roles explicit. It was a readability choice rather than a requirement. I’ve changed it to return a tuple as suggested.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: The property is needed because `EasyMagpieCFGDistillation` accesses validation-engine attributes and state. Internally, direct access through `params` is acceptable, the property is retained for external consumers. Kept only property.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: Agreed. I moved the ASR and speaker-verification model names into `_ValidationInferenceParams` and exposed them through the model configuration.

I kept `nvidia/parakeet-ctc-0.6b` as the default to preserve consistency with the ASR evaluator [already used](https://github.com/NVIDIA-NeMo/Speech/blob/main/nemo/collections/tts/models/easy_magpietts.py#L142) by `EasyMagpieTTSModel`. The model is no longer hard-coded: `nemo_asr_model_name` can be set to `nvidia/parakeet-tdt-1.1b` or another appropriate model in the training configuration.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: Yes, but there is an additional distinction: the validation models are owned by `_ValidationInferenceEngine`, which is intentionally a plain python class rather than a `nn.Module`. Consequently, they are not registered as submodules of `EasyMagpieCFGDistillation` and DDP environment does not see or synchronize their parameters or buffers. They are also instantiated in `on_fit_start()` after DDP environment setup. Because of those two properties, I did not need to explicitly mark their parameters and buffers as DDP-ignored.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: `_configure_training_mode()` resolves the `TrainingMode` used to prepare a batch.

`EasyMagpieTTSInferenceModel` supports multiple training modes and `EasyMagpieTTSModel.process_batch()` randomly selects one during training. We currently use only one mode, but I kept this behavior for compatibility and moved it into a separate helper. This also makes it easier to adopt future changes from the parent classes if we later remove the unused multi-mode functionality.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
artem-gorodetskii: Multi-turn distillation is not supported. I added batch-level task validation that raises a `ValueError` if any sample has a task other than `tts`. We cannot reject `use_multiturn_dataset=True` itself because that representation is also used for ordinary single-turn TTS cuts. For validated TTS batches, the interruption marker is still replaced with padding, [consistent with ](https://github.com/NVIDIA-NeMo/Speech/blob/main/nemo/collections/tts/models/easy_magpietts.py#L972)`EasyMagpieTTSModel.process_batch()`.

artem-gorodetskii:
> One final comment on style: Should we only keep the three main classes (_TeacherInferenceEngine, _ValidationInferenceEngine, EasyMagpieCFGDistillation‎) in the total level easy_magpietts_cfg_distillation.py and move the private dataclasses and functions elsewhere?

Ideally, each engine should be moved together with its private dataclasses and helper functions. For example, `_ValidationInferenceEngine`, its parameters and its metric helpers could live in a module under `nemo.collections.tts.modules.magpietts_validation`. However, I suggest handling this in a separate PR together with the broader `EasyMagpie` refactoring.

nemo/collections/tts/models/easy_magpietts_cfg_distillation.py
blisc: Ok, this is used to select how much delay we use between text, phones, and audio.

blisc (dismissed):
We are good to merge

blisc:
/ok to test [c5b0986](https://github.com/NVIDIA-NeMo/Speech/pull/16030/commits/c5b0986c070fdf701e85b542d74f132ada8983f9)

blisc:
/ok to test https://github.com/NVIDIA-NeMo/Speech/commit/73dc2e157c811e7e146c5915399359abe1bb49e9

blisc (approved):

chtruong814 (approved):

PR #16032, opened 2026-08-05
Core updates for Streaming Sortformer inference

### LINKED ISSUE #16002: `SortformerModules.streaming_update_async`: first-compression `spkcache_preds` seeding branch is unreachable — cache retains stale pop-time predictions, amplified into phantom speakers by `strong_boost`
**Component:** `nemo/collections/asr/modules/sortformer_modules.py` (observed in nemo-toolkit 2.7.3; verified present on current `main` at the time of writing: same `torch.full(..., 0.0, ...)` allocation, same `>= 0` sentinel, no seeding flag in the state class)

## Summary

In streaming diarization with `async_streaming=True`, the branch that is
supposed to seed `spkcache_preds` with the current forward's re-predictions at
the first speaker-cache compression can never execute. As a result the async
updater permanently scores cached frames with **stale pop-time posteriors** —
computed when the speaker cache was empty, i.e. the model's least-informed
predictions of the session — where the synchronous updater
(`streaming_update`) re-scores them with the current forward at seeding time.
Downstream, `_compress_spkcache`'s `strong_boost` reserves cache capacity per
speaker column, so a handful of stale, marginal (>0.5) posteriors on an unused
slot can lock in a **phantom speaker** that the sync updater never produces on
identical input.

## Mechanism

1. `streaming_update_async` reallocates its scratch buffer every call:
   `updated_spkcache_preds = torch.full((B, ..., n_spk), 0.0, ...)`
   and copies the live state (also zero-initialized at
   `init_streaming_state`) into it.
2. The seeding sentinel then tests
   `if updated_spkcache_preds[batch_index, 0, 0] >= 0:` — intended to mean
   "already compressed at least once". Sigmoid posteriors are non-negative
   and disabled frames are written as `0.0`, so this condition is **always
   true**, and the `elif` first-compression branch
   (`updated_spkcache_preds[b, :spkcache_len] = preds[b, :spkcache_len]`)
   is **unreachable**.
3. Consequently the cache keeps the posteriors recorded at FIFO-pop time
   (`pop_out_preds` from the *first* pop, when `spkcache` was empty). The
   sync updater instead seeds the accumulated cache from the **current**
   forward at the first compression
   (`spkcache_preds = cat([preds[:, :spkcache_len], pop_out_preds])`).
4. In `_compress_spkcache`, admission requires a posterior > 0.5
   (`_disable_low_scores`) and `strong_boost` adds ~+1.39 to the top-k
   frames of **each speaker column independently** — so a few stale frames
   marginally above 0.5 on an otherwise-unused slot are boosted into the
   kept set, self-reinforce on subsequent forwards (the model attends over
   cache frames labeled with that slot), and lock in a phantom speaker.

## Reproduction / evidence

- **Synthetic unit regime** (hand-crafted preds exercising `strong_boost`):
  default async vs sync `spkcache_preds` diverge with max abs difference
  ~0.90; forcing the seeding branch to execute makes async == sync exactly
  (0.0).
- **Real audio, batch size 1, fp32, streaming (0.96 s chunks, default
  geometry `spkcache_len=188, fifo_len=188, spkcache_update_period=144`):**
  on a 3-speaker meeting excerpt (public AMI corpus material), sync and
  async first diverge at exactly the first forward after the first
  compression (chunk 28 with this geometry; the arithmetic predicts pops at
  chunks 15 and 27, first compression at 27). Async goes on to commit a
  4th speaker slot (52 chunks of slot-4 activity; ~64 s of one real
  participant relabeled as a new speaker) while sync stays at 3 speakers on
  bit-identical input features.
- **Fix validation:** patching only the seeding semantics (see below) makes
  async's posterior stream **bit-identical to sync at all 187 chunks** of
  the same session in fp32, and eliminates the phantom speaker entirely.
  (Under bf16 autocast, exact equality is not expected — the async layout
  executes a different op sequence — and we observe small residual numeric
  divergence unrelated to this bug.)

## Suggested fix

Replace the sign-sentinel with an explicit per-row `spkcache_seeded: bool`
(or equivalent) in the streaming state, and on the first compression seed
`updated_spkcache_preds[b, :spkcache_len]` from the current `preds` exactly
as `streaming_update` does. A minimal proof-of-concept that restores sync
semantics is to change the scratch-buffer fill value from `0.0` to `-1.0`
(making the sentinel test meaningful); we validated that variant as
described above, but an explicit flag is the robust form.

Related, lower-priority observation while reading the same function: the
per-row pop length uses the batch-wide `max_chunk_len`
(`pop_out_len = max(pop_out_len, max_chunk_len - max_fifo_len + fifo_len)`)
where the sync path uses the row's own chunk length — inert for
uniform-length batches, but a latent cross-row coupling for ragged ones.

Happy to provide the standalone reproduction scripts and the patch as a PR
if useful.

ipmedenn:
> [!IMPORTANT]  

# What does this PR do ?

This PR introduces a large set of mostly inference-side updates for the upcoming 8-speaker Sortformer model release.

The main new feature is high-resolution output. Previously, Sortformer prediction resolution was tied to the encoder resolution, which is 80 ms in the standard model configurations. High-resolution mode decouples the two: the encoder continues operating at its configured resolution, while a trainable subpixel upsampling layer produces predictions at a native 10 ms resolution. These predictions can then be returned at any configured multiple of 10 ms, such as 20, 30, 80, or 160 ms. Previously released models retain their existing output resolution and behavior by default.

To make training practical for higher speaker counts, PIL and ATS targets are now generated using Hungarian assignment instead of exhaustive permutation search.

Asynchronous inference now uses fully vectorized ragged-batch state updates, organized into focused prediction-gathering, FIFO, silence-profile, and speaker-cache update stages. This eliminates per-row Python loops and scalar device-to-host synchronization while fixing the correctness issues described in [#16002](https://github.com/NVIDIA-NeMo/Speech/issues/16002).

The PR also adds an optional learnable silence embedding for speaker-cache compression; detailed inference profiling; metadata-validated prediction-tensor caching; native-resolution, hysteresis-aware vectorized diarization post-processing; optional encoder compilation; and streaming ONNX export fixes. The new profiling and caching components are implemented as reusable Sortformer utilities to keep the inference entrypoint focused on orchestration.

**Collection**: ASR / Speaker diarization

# Changelog

## Output resolution

- Add `high_resolution` mode with a trainable subpixel upsampling layer that produces predictions at the preprocessor frame resolution (10 ms in the standard configurations).
- Add `output_subsampling_factor` to control the returned prediction frame step independently of the encoder subsampling factor.
- Align both legacy and Lhotse diarization targets with the configured output resolution.
- Use the configured output resolution in diarization post-processing.
- Reject streaming configurations where chunk boundaries would split an output-downsampling window.
- Ensure the Sortformer output resolution matches the ASR encoder resolution inside `ParallelExpertEncoder`.

## Sortformer model and training

- Replace exhaustive PIL/ATS speaker-permutation search with Hungarian assignment.
- Add configurable arrival-time tolerance to ATS target generation.
- Support different numbers of reference speakers and prediction streams.
- Retain reference speakers beyond the model capacity during evaluation so they are counted as false negatives.
- Add an optional learnable silence embedding for speaker-cache compression; the running mean remains the default.
- Retain the legacy `hidden_to_spks` module for checkpoint compatibility while freezing its unused parameters.
- Allow `transformer_encoder` to be omitted from new model configurations.
- Remove the need for `find_unused_parameters=True`, allowing standard DDP.

## Streaming correctness and asynchronous inference

- Track speaker-cache compression state and refresh cached predictions from the current forward pass before the first compression in both synchronous and asynchronous streaming.
- Base FIFO eviction on each sample's actual chunk length in ragged asynchronous batches.
- Add fully vectorized asynchronous state updates, organized into focused private helpers for prediction gathering, FIFO-length calculation and updates, silence-profile updates, and speaker-cache updates, eliminating per-row Python loops and scalar device-to-host synchronization.
- Vectorize state concatenation and high-resolution chunk-prediction extraction for ragged asynchronous batches.
- Flush the remaining FIFO when an asynchronous row is finalized with a zero-length chunk.
- Add optional `async_pad_to_max` behavior to keep encoder input shapes fixed at their maximum capacity.
- Add optional `async_desync_updates` behavior to randomize initial FIFO eviction and stagger speaker-cache updates across batch rows.
- Add regression coverage verifying that a short stream is unaffected by longer streams in the same batch.

## Inference utilities and export

- Replace the boolean `save_preds_tensors` option with the explicit `out_preds_tensors` cache path. Validate existing cache metadata and write new caches atomically; use `overwrite_preds_tensors=true` to replace an existing cache.
- Add detailed inference profiling, including timings for preprocessing, streaming-state concatenation, encoding, cache compression, and state updates.
- Change the default inference precision used by `e2e_diarize_speech.py` from FP32 to BF16. Set `precision=32` to retain the previous behavior.
- Vectorize diarization binarization and post-processing, substantially reducing post-processing overhead.
- Add optional `torch.compile` support for the encoder through `compile_encoder`.
- Add model-sized default input examples for streaming export.
- Allow callers to provide an explicit input example for streaming export.
- Keep the concatenation width fixed during ONNX tracing while allowing logical cache, FIFO, and chunk lengths to vary at runtime.
- Add an ONNX Runtime regression test covering empty, partially filled, and fully populated streaming states.

## Unit tests

- high-resolution training, inference, and streaming;
- output-length calculation and target-resolution alignment;
- models configured without `transformer_encoder`;
- legacy checkpoint loading;
- learnable silence embeddings;
- Hungarian PIL/ATS assignment at higher speaker counts and with different numbers of reference and predicted speakers;
- evaluation with more reference speakers than the model output capacity;
- `ParallelExpertEncoder` output-resolution alignment;
- prediction-cache metadata, validation, atomic overwrite, malformed payload handling, and inference profiling;
- vectorized binarization hysteresis, empty outputs, multiple speaker streams, recording offsets, and native prediction-frame timing;
- first-compression correctness and vectorized asynchronous state updates for ragged batches;
- asynchronous row finalization and zero-length chunk flushing;
- fixed-capacity asynchronous encoder padding;
- vectorized high-resolution asynchronous chunk-prediction extraction;
- streaming ONNX export with empty, partially filled, and fully populated runtime states.

# Usage

## Model configuration

Existing Sortformer model configurations remain backward-compatible. When `high_resolution` is omitted or set to `false`, the native prediction resolution remains tied to the encoder subsampling factor, and `output_subsampling_factor` defaults to that same value.

For a standard 10 ms preprocessor frame step, high-resolution training with 10 ms returned predictions can be configured as:

```yaml
model:
  high_resolution: true
  output_subsampling_factor: 1
```

In high-resolution mode, `output_subsampling_factor` is expressed in preprocessor frames. With a 10 ms preprocessor frame step, values of `1`, `2`, `3`, `8`, and `16` correspond to returned prediction steps of 10, 20, 30, 80, and 160 ms, respectively.

By default, speaker-cache compression uses the running mean silence embedding. A learnable silence embedding can be enabled with:

```yaml
model:
  sortformer_modules:
    use_learnable_sil_emb: true
```

For new architectures that do not require the secondary transformer encoder, the `transformer_encoder` section may be omitted. Previously released model configurations containing this section continue to instantiate it normally.

## Streaming output constraints

Streaming output downsampling is applied independently to each chunk, so chunk boundaries must align with complete downsampling windows. Configurations whose chunk length is incompatible with `output_subsampling_factor` are rejected with a descriptive error.

## Inference script options

The inference script `e2e_diarize_speech.py` provides the following asynchronous-streaming, caching, and performance controls:

```yaml
precision: bf16
out_preds_tensors: null
overwrite_preds_tensors: false
async_streaming: false
async_pad_to_max: false
async_desync_updates: false
compile_encoder: false
profile_inference: true
```

`async_streaming: true` enables asynchronous batched state updates; `async_pad_to_max` and `async_desync_updates` only affect this mode. 

`async_pad_to_max: true` keeps encoder inputs at a fixed maximum shape, trading additional padded computation for stable input shapes.

`async_desync_updates: true` randomizes the initial FIFO eviction to stagger speaker-cache updates across batch rows.

`out_preds_tensors` specifies a prediction-cache path and enables both loading and saving of prediction tensors. Existing caches are validated before reuse. Set `overwrite_preds_tensors: true` to recompute and replace an existing cache.

The default inference precision is BF16. Set `precision: 32` to retain the previous FP32 behavior.

Setting `compile_encoder: true` applies `torch.compile(..., dynamic=True)` to the encoder only; it does not compile the complete Sortformer model.

Detailed inference profiling is enabled by default and can be disabled with `profile_inference: false`.

## ONNX export

Streaming export uses a model-sized default input example, and callers may provide an explicit input example when different physical capacities are required.

The exported ONNX graph has a fixed batch size and fixed physical cache, FIFO, and chunk capacities. Their logical lengths may vary independently for each batch row at runtime, including empty, partially filled, and fully populated states.

# GitHub Actions CI

**Pre checks**:

- [x] Did you write any new necessary tests? Necessary unit tests were added.
- [x] Did you add or update any necessary documentation? New configuration fields include inline documentation.
  
**PR Type**:

- [x] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Fixes [#16002](https://github.com/NVIDIA-NeMo/Speech/issues/16002).


examples/speaker_tasks/diarization/neural_diarizer/e2e_diarize_speech.py
```diff
@@ -121,6 +133,201 @@ class DiarizationConfig:
     optuna_n_trials: int = 100000
 
 
+class InferenceProfiler:
```
tango4j: Move this to parts.utils folder (maybe diarization utils)

examples/speaker_tasks/diarization/neural_diarizer/e2e_diarize_speech.py
```diff
@@ -143,24 +350,116 @@ def optuna_suggest_params(postprocessing_cfg: PostProcessingParams, trial) -> Po
...
+    model_id = model_path.name.replace(".ckpt", "").replace(".nemo", "")
+    model_id = f"{model_id}_sf{cfg.output_subsampling_factor}"
+    if cfg.out_preds_tensors:
+        tensor_path = Path(cfg.out_preds_tensors).expanduser().absolute()
+    else:
+        tensor_path = None
+    return None if tensor_path is None else str(tensor_path), model_id, tensor_filename
+
+
+def get_prediction_cache_metadata(cfg, diar_model, infer_audio_rttm_dict) -> Dict:
```
tango4j: These four functions should move to diarization_utils.py

nemo/collections/asr/data/audio_to_diar_label.py
```diff
@@ -55,6 +55,8 @@ def get_subsegments_to_timestamps(
         ts (torch.tensor):
             A tensor containing the scaled and rounded timestamps for each subsegment.
     """
+    if len(subsegments) == 0:
+        return torch.zeros((0, 2), dtype=torch.long)
```
tango4j, line 60: Either add comment on this shape or create a constant variable for 2.

nemo/collections/asr/modules/sortformer_modules.py
```diff
@@ -423,83 +531,166 @@ def streaming_update_async(self, streaming_state, chunk, chunk_lengths, preds, l
...
         chunk_lengths = (chunk_lengths - lc).clamp(min=0, max=max_chunk_len)
-        updated_fifo = torch.zeros((batch_size, max_fifo_len + max_chunk_len, emb_dim), device=preds.device)
-        updated_fifo_preds = torch.zeros((batch_size, max_fifo_len + max_chunk_len, n_spk), device=preds.device)
-        updated_spkcache = torch.zeros((batch_size, max_spkcache_len + max_pop_out_len, emb_dim), device=preds.device)
-        updated_spkcache_preds = torch.full(
-            (batch_size, max_spkcache_len + max_pop_out_len, n_spk), 0.0, device=preds.device
+
+        # Extract current predictions for all state regions with one rectangular gather.
+        spkcache_positions = torch.arange(max_spkcache_len, device=preds.device).unsqueeze(0)
+        fifo_positions = torch.arange(max_fifo_len, device=preds.device).unsqueeze(0)
```
tango4j, line 576: I suggest to create a class that handles async updates. Also recommend to create some class methods (function that manipulates the variables).

nemo/collections/asr/parts/utils/asr_multispeaker_utils.py
```diff
@@ -183,6 +185,169 @@ def get_pil_targets(labels: torch.Tensor, preds: torch.Tensor, speaker_permutati
     return max_score_permed_labels  # (batch_size, num_speakers, num_classes)
 
 
+def _validate_hungarian_target_inputs(
```
tango4j, line 188: Please add docstrings with this format: (sortformer_diar_model.py)

variable_name (type): description

nemo/collections/asr/parts/utils/asr_multispeaker_utils.py
```diff
@@ -183,6 +185,169 @@ def get_pil_targets(labels: torch.Tensor, preds: torch.Tensor, speaker_permutati
...
+        labels.shape[2],
+        preds.shape[2],
+    ):
+        raise ValueError(
+            f"assignment_mask must have shape {(labels.shape[0], labels.shape[2], preds.shape[2])}, "
+            f"got {assignment_mask.shape}"
+        )
+
+
+def get_pil_targets_hungarian(
```
tango4j, line 224: Make this docstrings follow the style of sortformer_diar_models.py

variable (type) : descriptions

nemo/collections/asr/parts/utils/asr_multispeaker_utils.py
```diff
@@ -183,6 +185,169 @@ def get_pil_targets(labels: torch.Tensor, preds: torch.Tensor, speaker_permutati
...
+        speaker_indices[batch_indices, batch_col_indices] = torch.where(
+            matched_speakers_are_active,
+            batch_row_indices,
+            torch.full_like(batch_row_indices, -1),
+        )
+
+    return aligned_labels, speaker_indices
+
+
+def get_ats_targets_hungarian(
```
tango4j, line 312: Same changes needed for docstrings

tango4j: Ask agent to follow the docstring convention of sortformer_diar_models.py

Which is 
Args:
Returns:
vairable_name (type) : description

tango4j (changes requested):
The most important things to be fixed globally in this PR are:
(1) The docstrings should follow the conventions in `sortformer_diar_models.py` where there is variable name, type and descriptions.
(2) Consider modularizing the sortformer modules code for async updates - creating a class, moving variable inits and functions to this async update class.
(3) `e2e_diarize_speech.py` needs to be maintained under principle of "call sight thinness". Please move the latency measuring parts to `diarization_utils.py` or if needed, you can create a new file in `nemo.collections.parts.utils` folder.

nemo/collections/asr/modules/sortformer_modules.py
tango4j: Instead created functions that handle variables and operations to keep each function short

tango4j:
/ok to test [a784ba6](https://github.com/NVIDIA-NeMo/Speech/pull/16032/commits/a784ba6779c06f339736b6fc00329afd4f568441)

tango4j (approved):
The requested changes are reflected. Approving

PR #16033, opened 2026-08-05
fix: report manifest format mismatch instead of failing in the wrong parser

### LINKED ISSUE #14287: TypeError: SupervisionSegment.__init__() got an unexpected keyword argument 'audio_filepath'
**Describe the bug**
error when running canary-qwen-2.5b using manifest.jsonl

adityasingh2400:
# What does this PR do?

`guess_parse_cutset` routes a string input by extension: `.json` means a NeMo manifest and `.jsonl` or `.jsonl.gz` means a Lhotse manifest. Nothing checked that the file contents agreed with the extension, so a mislabeled manifest was handed to the wrong parser and failed deep inside it with an error that never mentioned the extension.

That is issue #14287. A NeMo manifest named `input_manifest.jsonl` reaches the Lhotse parser, which treats each record as a Lhotse object and dies with:

```
TypeError: SupervisionSegment.__init__() got an unexpected keyword argument 'audio_filepath'
```

Two people hit this independently on `examples/speechlm2/salm_generate.py` and neither could tell from the message that the filename was the problem. The mirror case has the same shape: a Lhotse manifest named `cuts.json` reaches the NeMo JSON Lines parser and dies with a bare `JSONDecodeError: Expecting value: line 2 column 1`.

This PR detects the format from the manifest's first record and raises a `ValueError` that names the convention and the fix, for both directions.

# Approach

Every record of a NeMo non-tarred manifest carries an `audio_filepath` field and Lhotse records never do, so one record is enough to tell the formats apart. A leading `[` means a JSON array, which only Lhotse uses. `_sniff_manifest_format` applies those two rules and `_check_manifest_format` raises when the result contradicts the extension.

Three properties keep this safe on the paths that already work:

Detection reads at most 64 KiB, so a large manifest serialized as a single line is not pulled into memory. A record longer than that window yields no opinion rather than a guess.

Detection returns `None`, meaning no opinion, for anything it cannot read or parse. Missing paths, shard patterns such as `shard_{0..3}.jsonl`, remote URIs, empty files and non-JSON files all fall into this bucket, and the caller keeps its existing behaviour untouched.

`open_best` is used for reading, so gzipped and remote manifests are sniffed the same way as plain local ones.

The check runs only in `guess_parse_cutset`, where the extension convention is applied. Configs that name `manifest_filepath` or `cuts_path` explicitly are unchanged, since there the user has already stated the format.

# Verification

Verified on CPU against `upstream/main` at `40ace43`.

Reverting only `cutset.py` to the base revision reproduces both reported failures, and restoring the fix turns both into the new actionable error:

```
FAIL-BEFORE  NeMo manifest named .jsonl   TypeError: SupervisionSegment.__init__() got an unexpected keyword argument 'audio_filepath'
FAIL-BEFORE  Lhotse manifest named .json  JSONDecodeError: Expecting value: line 2 column 1 (char 2)
PASS-AFTER   NeMo manifest named .jsonl   ValueError: Cannot read '...' as a Lhotse manifest: ... 'audio_filepath' field, which makes it a NeMo manifest ...
PASS-AFTER   Lhotse manifest named .json  ValueError: Cannot read '...' as a NeMo manifest: ... its contents are a Lhotse manifest ...
```

New tests in `TestGuessParseCutsetManifestFormat` cover both mismatch directions, `.jsonl` and `.jsonl.gz` for the mislabeled NeMo case, the three correctly named happy paths, and the inconclusive inputs that must keep the old behaviour. They fail at the base revision and pass with the fix, 11 passed.

The surrounding suite is unaffected:

```
pytest tests/collections/common/test_lhotse_dataloading.py -q -m "not pleasefixme" --cpu
78 passed, 1 skipped
```

`isort` 5.13.2 and `black` 24.10.0 at the versions pinned in `.pre-commit-config.yaml` report both changed files clean. `flake8 --config=.flake8` and `--config=.flake8.speech` report nothing new. The three findings that remain in the test file are present at the base revision and are untouched by this change.

# Issues

Fixes #14287

# PR Type

- [x] Bugfix
- [ ] New Feature
- [ ] Documentation
- [ ] Test

# Additional notes

Commit is signed off per CONTRIBUTING. As a community fork PR, `copy-pr-bot` will not start CI on its own, so a maintainer comment of `/ok to test <sha>` is needed to run it.

### DIFF
--- a/nemo/collections/common/data/lhotse/cutset.py
+++ b/nemo/collections/common/data/lhotse/cutset.py
@@ -14,6 +14,7 @@
 """Lhotse CutSet utilities and Parquet manifest support for NeMo."""
 
 import io
+import json
 import logging
 import random
 import re
@@ -31,7 +32,7 @@ from lhotse import CutSet, Features, MonoCut, Recording, SupervisionSegment
 from lhotse.array import Array, TemporalArray
 from lhotse.cut import Cut, MixedCut, PaddingCut
 from lhotse.lazy import LazyIteratorChain
-from lhotse.serialization import load_yaml
+from lhotse.serialization import load_yaml, open_best
 from lhotse.utils import fastcopy
 from omegaconf import DictConfig, ListConfig, OmegaConf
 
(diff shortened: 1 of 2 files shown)


adityasingh2400:
Checking in on this one. It has been open a month without a review and all three checks are green.

It fixes #14287, which two people hit independently on `examples/speechlm2/salm_generate.py`, where a mislabeled manifest reaches the wrong parser and fails with an error that never mentions the extension.

The diff is +150 -1, confined to `guess_parse_cutset` in `cutset.py` plus its tests. Detection returns no opinion for anything it cannot read or parse, including shard patterns, remote URIs and empty files, so the paths that already work are untouched.

Happy to rebase or split it up if that would make review easier.

PR #16035, opened 2026-08-05
Enable offloading of validation models to CPU during training runs for EMTTS

blisc:
# What does this PR do ?

Enable offloading of validation models to CPU during training runs for EMTTS

**Collection**: tts

# Changelog

- Enable offloading of validation models to CPU during training runs for EMTTS
  
**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

### DIFF
--- a/nemo/collections/tts/models/easy_magpietts.py
+++ b/nemo/collections/tts/models/easy_magpietts.py
@@ -129,6 +129,7 @@ class EasyMagpieTTSModel(EasyMagpieTTSInferenceModel):
 
         # Validation inference with metrics (optional)
         self.run_val_inference = cfg.get('run_val_inference', False)
+        self.offload_validation_models = cfg.get('offload_validation_models', True)
         self.use_multilingual_asr = cfg.get('use_multilingual_asr', False)
         if self.run_val_inference:
             logging.info("Loading eval models for validation inference (ASR and speaker verification)...")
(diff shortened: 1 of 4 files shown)


rfejgin (commented):
In general looks good, but I had some top level questions: Is the purpose to free up GPU memory to enable larger batch sizes in training? If so, could you say a bit on how much GPU memory the validation models use to start with?

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -1278,6 +1411,11 @@ def process_batch(
         )
 
     def training_step(self, batch, batch_idx):
+        uses_uncached_audio = 'context_audio_codes' not in batch or 'audio_codes' not in batch
```
Edresson, line 1414: We should extend this if statement to check whether the codec is required for multi-turn user-audio conditioning:

```
self.cfg.get("use_multiturn_dataset", False)
and batch["user_audio_turn_splitted"] is not None
and self.cfg.get("condition_on_user_speech", False)
```

Without this check, when context_audio_codes and audio_codes are included in the multi-turn data, user-audio conditioning feature extraction may run with the codec on the CPU, which would be extremely slow or raise errors.

nemo/collections/tts/models/easy_magpietts.py
```diff
@@ -1296,6 +1434,8 @@ def training_step(self, batch, batch_idx):
             audio_lens = batch['audio_lens']
             audio_codes, audio_codes_lens = self._codec_helper.audio_to_codes(audio, audio_lens)
 
+        if uses_uncached_audio and self._should_offload_validation_models():
```
Edresson, line 1437: Why is this necessary? Could we keep the codec on the GPU throughout an epoch whenever any batch in that epoch requires it?

If you still want to do it, please move it to the after the following block, otherwise multiturn training data will fails.
```
 if (
            self.cfg.get("use_multiturn_dataset", False)
            and batch["user_audio_turn_splitted"] is not None
            and self.cfg.get("condition_on_user_speech", False)
        ):
```

Edresson (commented):
Why is this change necessary?

Also, have you measured its impact on validation performance? Validation is already relatively slow, and I’m concerned that this PR may introduce additional overhead.

PR #16036, opened 2026-08-05
Add an AutoModel Path for EMTTS In Addition to Local NemotronH file

blisc:
# What does this PR do ?

Add an AutoModel Path for EMTTS In Addition to Local NemotronH file

**Collection**: tts

# Changelog

- Add an AutoModel Path for EMTTS In Addition to Local NemotronH file

**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation


nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -230,6 +230,106 @@ class EasyMagpieTTSInferenceModel(ModelPT):
     EasyMagpieTTSModel subclasses this to add training, validation, and data loading.
     """
 
+    @staticmethod
+    def _to_container(cfg_value):
```
pzelasko: IMO would be cleaner to have some of these utilities as private free functions at the bottom of the file rather than class-level static methods. My rule of thumb: if it's generic (eg config/dict manipulation), make it a function; if it's tightly coupled with the class, make it a method.

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -516,8 +613,71 @@ def __init__(self, cfg: DictConfig, trainer: 'Trainer' = None):
...
+            self.transformer_backend_config = self._replace_config_values(
+                self.transformer_backend_config,
+                self._get_nemotron_h_config_dict(cfg),
+            )
+
+            automodel_kwargs = self._to_container(cfg.get('automodel_kwargs', {}))
+            logging.info("NeMo AutoModel kwargs: %s", automodel_kwargs)
+            # Some Automodel NemotronV3 builds leave mixer tensors from scratch init uninitialized after
+            # from_config(); reset those tensors before running Automodel's regular init/rescaling path.
+            with torch.device('cpu'):
```
pzelasko, line 533: If you're going to train Nano-sized or larger models with distributed tensor sharding, you'll need to initialize the model on "meta" device and move to GPU directly later to avoid a spike in GPU memory or super long CPU init time. You can consult `SALMAutomodel.configure_model()` for reference how to put it together with PTL trainer

nemo/collections/tts/models/easy_magpietts_inference.py
```diff
@@ -230,6 +230,106 @@ class EasyMagpieTTSInferenceModel(ModelPT):
...
+            config.update(values)
+            return config
+        for key, value in values.items():
+            setattr(config, key, value)
+        return config
+
+    @staticmethod
+    @torch.no_grad()
+    def _initialize_automodel_scratch_parameters(module: nn.Module):
+        """Reset NemotronV3 tensors that some Automodel scratch-init paths leave uninitialized."""
```
pzelasko: CC @akoumpa @hemildesai looks like some parameters in Automodel Nemotron3 implementation are not being initialized when training from scratch.

nemo/collections/tts/models/easy_magpietts_inference.py
blisc: Re-worked the static functions into standalone ones.

nemo/collections/tts/models/easy_magpietts_inference.py
blisc: TBH, this was inherited from the first revision. Not sure if I still need this or not. Until we test its removal, I'd rather keep it

nemo/collections/tts/models/easy_magpietts_inference.py
blisc: I'd keep it as only supporting simple DDP for now. When we get to larger models, I can consider refactoring.

tools/easymagpie_vllm_omni/scripts/convert_to_vllm.py
paarthneekhara, line 1: Is the VLLM inference tested for this?

Edresson (approved):

Edresson (commented):
I added a small commit that fix the multiturn prefill phase. Now all seems good to me.

PR #16044, opened 2026-08-07
Support flat native NeMo lists in idxpacks

pzelasko:
## Summary

- allow packed native NeMo datasets to use non-empty flat `manifest_filepath` lists
- allow matching flat `tarred_audio_filepaths` lists and preserve their existing declaration-position pairing
- aggregate each flat list into one runtime-addressable idxpack collection
- reject nested/weighted and mixed scalar/list packed forms with explicit errors

## Backward compatibility

Scalar path specs retain the existing expansion and numeric shard-alignment validation unchanged. Non-idxpack list handling is unchanged. For the newly supported flat packed form, each declared manifest/tar pair is expanded independently, checked for equal cardinality, and concatenated in declaration order; numeric filenames are intentionally not compared across the pair.

This does not change any model or training-step signature.

## Validation

- `17 passed`: focused idxpack and dataset-scope tests
- `74 passed`: broader Lhotse dataloading and NeMo adapter tests
- syntax, formatting, import ordering, and diff checks pass
- dry-run against both production Hero Riva blend YAMLs succeeds; each discovers 72 collections and 295,544 ordered paths

### DIFF
--- a/nemo/collections/common/data/lhotse/cutset.py
+++ b/nemo/collections/common/data/lhotse/cutset.py
@@ -1662,23 +1662,45 @@ def read_nemo_manifest(config) -> tuple[CutSet, bool]:
     }
     tar_kwargs_extra = {"indexed": indexed, **indexed_extra, **pack_extra} if indexed else {}
     is_tarred = config.get("tarred_audio_filepaths") is not None
+    manifest_filepath = config.manifest_filepath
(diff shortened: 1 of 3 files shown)


tango4j (approved):
This PR is working and fixing the new dataloader issue.

PR #16048, opened 2026-08-10
Support chunk microbatching and idxpack for multimodal SpeechLM data

KunalDhawan:
This PR improves SpeechLM training support for long-form multimodal conversation data and ParallelExpertEncoder-style perception models.

Changes include:
- Add dataset-level `.idxpack` support for `multimodal_conversation` JSONL manifests.
- Support direct archive-member audio references such as `path/to/archive.tar/member.wav` in multimodal conversation manifests.
- Add `encoder_chunk_batch_size` to encode long-audio chunks in smaller perception forwards.
- Keep FSDP-sharded perception ranks synchronized for audio-free/text-only local batches using zero-valued dummy forwards.
- Make distributed oomptimizer compatible with SALM Automodel’s dataloader-iterator training path.


scripts/speechlm2/distributed_oomptimizer.py
```diff
@@ -1221,6 +1239,10 @@ def _is_oom_like(error: RuntimeError) -> bool:
         "cuFFT error: CUFFT_INTERNAL_ERROR" in error_msg
         or "CUDA out of memory" in error_msg
         or "CUDACachingAllocator" in error_msg
+        or (
+            "torch.utils.checkpoint.CheckpointError" in error_msg
+            and "Recomputed values for the following tensors have different metadata" in error_msg
```
pzelasko: is this really an OOM?

scripts/dataloading/convert_indexes_to_idxpack.py
```diff
@@ -361,6 +365,11 @@ def discover_pack_collections(
                     "Packed ShareGPT supports JSONL manifests with direct/remote "
                     "audio paths, not paired audio tar files."
                 )
+            if typ == "multimodal_conversation":
+                raise NotImplementedError(
+                    "Packed multimodal_conversation supports JSONL manifests with "
+                    "direct/remote audio paths, not paired audio tar files."
```
pzelasko, line 371: I don't think there is anything preventing multimodal_conversation + idxpack from working with paired audio tar files (as long as the path in the manifest encoded as `/path/to/audio.tar/actual_audio.wav`)

scripts/speechlm2/distributed_oomptimizer.py
KunalDhawan: Good point, this should not be classified as OOM. It is a checkpointing metadata/consistency failure, and treating it as OOM could hide a real issue or bias the search. I’ll remove this branch from `_is_oom_like` so it surfaces as a normal probe/child failure.

scripts/dataloading/convert_indexes_to_idxpack.py
KunalDhawan: Agreed for the case where the manifest audio value already encodes the archive member path, e.g. `/path/to/audio.tar/actual_audio.wav`; with that representation packed `multimodal_conversation` can work from the JSONL collection alone.

I’m hesitant to support `tarred_audio_filepaths` as a separate packed tar collection in this PR because a multimodal-conversation row can contain multiple audio turns, so one manifest row does not necessarily map 1:1 to one tar member like native NeMo manifests do. I’ll keep/narrow the contract to archive-member references in the manifest unless we add an explicit mapping rule for paired tar sidecars.

pzelasko: ok let's skip tarred

pzelasko (dismissed):

PR #16054, opened 2026-08-11
Fix Mamba FP32 holder names in SpeechLM vLLM mapping

pzelasko:
## Summary

- Normalize Automodel Mamba/GDN FP32 holder names at the SpeechLM NeMo-to-vLLM boundary:
  - `._fp32_params.A_log` -> `.A`
  - `._fp32_params.dt_bias` -> `.dt_bias`
  - `._fp32_params.D` -> `.D`
- Restrict the conversion to `.mixer._fp32_params.` paths.
- Preserve tensor objects and values; vLLM remains responsible for the `A_log` numerical transform.
- Add focused mapping, negative-path, and CPU-only vLLM loader regression tests.

## Root cause

Current NeMo/Automodel Nemotron hybrid checkpoints serialize these Mamba parameters inside the `_fp32_params` holder. The SALM vLLM backend forwarded those names unchanged, while vLLM expected canonical Mamba names and failed at model initialization with a missing `..._fp32_params.A` key.

## Validation

- `pytest tests/collections/speechlm2/test_vllm_plugin.py -k HybridBackendWeightMapping`
  - 11 passed, 44 deselected
  - Includes the real vLLM 0.20 CPU-only loader regression and verifies `-exp` is applied exactly once.
- `isort --check-only`: passed
- `python -m py_compile`: passed
- `git diff --check`: passed
- Black was applied before commit; the available local Black is 22.3.0 while this checkout requires Black 24+, so the configured Black check could not run in this environment.

### DIFF
--- a/nemo/collections/speechlm2/vllm/salm/backends.py
+++ b/nemo/collections/speechlm2/vllm/salm/backends.py
@@ -217,9 +217,9 @@ class TransformerBackend(_BaseBackend):
 
 
 _FP32_PARAM_HOLDER_RENAMES = {
-    "._fp32_params.A_log": ".A",
-    "._fp32_params.dt_bias": ".dt_bias",
-    "._fp32_params.D": ".D",
+    ".mixer._fp32_params.A_log": ".mixer.A",
+    ".mixer._fp32_params.dt_bias": ".mixer.dt_bias",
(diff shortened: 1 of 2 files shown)


DongjiGao (approved):
LGTM.

PR #16059, opened 2026-08-13
Preserve valid codec subframe before audio EOS

vklimkov-nvidia:
## Summary

- forward the terminal packed audio row from Stage 0 instead of discarding the entire row when it contains a control code
- after codec synthesis, trim only the terminal subframes starting at the first control position
- cover control codes in both packed subframe positions

## Validation

- `102 passed, 1 skipped` in `tests/collections/tts/easymagpie_vllm_omni/serving`
- style checks passed for all four changed files
- `git diff --check` passed

## Large-scale endpoint check

Compared 128 Torch generations with CFG disabled against 128 vLLM generations on the corrected, weight-matched checkpoint (96 VCTK texts and 32 longer assistant-style texts; temperature 0.7, top-k 80).

- no convincing mid-word cutoff in either backend
- final-word mismatches: Torch 1/128; vLLM 2/128, attributable to `tonight`/`night`, spelling normalization, and tokenization of `Dunfermline`
- median Whisper final-word-to-EOF margin: Torch 320 ms; vLLM 330 ms
- median energy-defined terminal silence: Torch 88.7 ms; vLLM 102.1 ms
- normalized WER: Torch 1.75%; vLLM 1.56%

The change remains useful for correctness: when EOS occurs in the second packed subframe, the first subframe contains approximately 40 ms of valid audio and should reach synthesis.


tools/easymagpie_vllm_omni/easymagpie_vllm_omni/codec/model.py
```diff
@@ -167,10 +167,21 @@ def forward(
             )
         packed_audio = self.codec(codes)
         outputs: list[torch.Tensor] = []
+        frame_offset = 0
         offset = 0
         for frames in frame_counts:
             samples = frames * self.config.samples_per_frame
-            outputs.append(packed_audio[offset : offset + samples].float())
+            valid_samples = samples
+            if frames:
```
blisc: Can we make this if statement more strict? What data type is this?

blisc (dismissed):
Approved from my end, but left a minor comment

tools/easymagpie_vllm_omni/easymagpie_vllm_omni/codec/model.py
vklimkov-nvidia: Done. `frames` is a Python `int`: `frame_counts` is typed as `list[int]`, and its entries are constructed with `int(...)`. I tightened the condition to `if frames > 0:` in abf95289a.

blisc (approved):

vklimkov-nvidia:
ok to merge ac697970309c34e3f417c3564921c39f14e6076e

PR #11770, opened 2025-01-07
Update requirements_lightning.txt

pablo-garay:
PTL to 2.5.0.post0 (used in 24.12 release)

PR #16076, opened 2026-08-18
fix: relax lightning cap to <2.7.0 to allow the load_from_checkpoint security fix

### LINKED ISSUE #15706: Relax lightning <= 2.4.0 cap to allow >= 2.6.1 (security fix for LightningModule.load_from_checkpoint deserialization)
## Problem
`requirements/requirements_lightning.txt` pins `lightning>2.2.1,<=2.4.0` on every released `nemo-toolkit` version (up to `2.7.3`) and on `main`. `lightning<=2.4.0` is affected by a Critical-severity advisory in `LightningModule.load_from_checkpoint()` (PyTorch unsafe-unpickling defaults → arbitrary code execution on checkpoint load). The fix is in `lightning>=2.6.1`, but the NeMo upper-bound prevents downstream services from upgrading.

## Request
Relax `requirements/requirements_lightning.txt` to allow `lightning>=2.6.1` — or document the technical reason the upper bound must stay at `2.4.0` so downstreams can plan a mitigation.

## References
- Affected file (main): https://github.com/NVIDIA-NeMo/NeMo/blob/main/requirements/requirements_lightning.txt
- Previous bump attempt (closed stale): https://github.com/NVIDIA/NeMo/pull/11770
- CVE: CVE-2026-31221 (NVD analysis pending)

Thanks!

udsy19:
# What does this PR do ?

Relaxes the `lightning>2.2.1,<=2.4.0` cap to `lightning>2.2.1,<2.7.0` so downstream users can install `lightning>=2.6.1` (which contains the security fix for the `LightningModule.load_from_checkpoint()` unsafe-deserialization advisory, CVE-2026-31221), and fixes the three concrete incompatibilities that previously made NeMo unimportable/broken on lightning >= 2.6.

Fixes #15706

**Collection**: core (`nemo.core`, `nemo.lightning`, `nemo.utils`) — affects all collections transitively.

# Changelog

- `pyproject.toml`: relax `lightning>2.2.1,<=2.4.0` to `lightning>2.2.1,<2.7.0` in all 8 extras (`core`, `lightning`, `common`, `asr`, `audio`, `tts`, `speechlm2`, `all`). lightning 2.7.0 does not exist yet (2.6.5 is current), so the new cap still guards against an unreviewed future minor.
- `uv.lock`: regenerated with `uv lock` — **only the 8 specifier strings change; the locked `lightning` version stays at 2.4.0** (and `pytorch-lightning` at 2.6.1, unchanged from `main`), so CI containers and `uv sync --locked` resolve exactly as they do today. Bumping the pinned version is left as a separate maintainer decision (see "Known limitation" below).
- `nemo/lightning/one_logger_callback.py`: guard the `nv_one_logger.training_telemetry.integration.pytorch_lightning` import. All released `nv-one-logger` versions (<= 2.3.1) subclass `lightning.pytorch.Trainer` and fail at import time with `TypeError` against lightning >= 2.6 (`Trainer.save_checkpoint`'s `weights_only` changed from `bool = False` to `Optional[bool] = None`, and the override validator rejects the stale signature). Previously this made `import nemo.core` raise. Now `OneLoggerNeMoCallback` degrades to a no-op `BaseCallback` with a one-time warning naming the underlying error. Catching `ImportError` also makes NeMo work when only `nv-one-logger-core` / `-training-telemetry` are installed without the separate `-pytorch-lightning-integration` distribution.
- `nemo/utils/exp_manager.py`: make the `NeptuneLogger` import optional — it was removed from `lightning.pytorch.loggers` in lightning 2.6.0 and previously broke `import nemo.utils.exp_manager` outright. Requesting `create_neptune_logger: true` on lightning >= 2.6 now raises a clear `ValueError`; all other loggers are unaffected.
- `nemo/utils/__init__.py`: register omegaconf's container/node classes with `torch.serialization.add_safe_globals`. Lightning >= 2.6 loads checkpoints with `weights_only=True` — that *is* the CVE fix — but NeMo stores its config as omegaconf objects in `hyper_parameters`, so every resume (`resume_if_exists`) and `load_from_checkpoint` failed with `UnpicklingError`. Registering these classes keeps NeMo checkpoints loadable **under the safe weights-only unpickler**, rather than pushing users back to `weights_only=False`. See the security note below. Registration is a no-op on torch versions without `add_safe_globals`.
- `tests/core/test_serialization.py`: new unit test asserting a checkpoint with omegaconf `hyper_parameters` loads under `torch.load(..., weights_only=True)`.
- `tests/lightning/test_one_logger_callback.py`: new `TestOneLoggerFallback` regression test (simulates an unimportable integration module and asserts NeMo degrades to a working no-op instead of failing to import), plus `skipif` markers on the 8 tests that need the real nv-one-logger PTL integration. Nothing is skipped on today's locked lightning 2.4.0.

Deliberately **not** included: an `on_train_end` escape hatch that passes `weights_only=False` when re-loading the best checkpoint. I had this in an earlier revision, but it is unnecessary — with the safe-globals registration in place, `tests/core/test_exp_manager.py -k save_best_model` passes on lightning 2.6.5 with the secure default — and it would have been unsound, because `ModelCheckpoint.load_state_dict` restores `best_model_path` verbatim from the resumed checkpoint's callback state. On a `resume_if_exists` run that path is attacker-influenced, not "a file this job just wrote", so forcing a full unpickle there would have re-opened the exact hole the advisory closes. `nemo/utils/callbacks/nemo_model_checkpoint.py` is therefore untouched by this PR.

# Security note on `add_safe_globals`

This is the only part of the PR that changes what a NeMo process will deserialize, so spelling it out. The registered set is exactly:

- `omegaconf.DictConfig`, `omegaconf.ListConfig`
- `omegaconf.base.ContainerMetadata`, `omegaconf.base.Metadata`
- `omegaconf.nodes.{AnyNode, ValueNode, StringNode, IntegerNode, FloatNode, BooleanNode, BytesNode, PathNode, EnumNode, InterpolationResultNode}` (each looked up with `getattr` so the set adapts to the installed omegaconf)
- `typing.Any`, `collections.defaultdict`, and the builtins `dict, list, tuple, int, float, str, bool, bytes`

The non-omegaconf entries are not padding — pickling a real `DictConfig` emits `GLOBAL` opcodes for `typing.Any` (the default `ContainerMetadata.element_type`), `collections.defaultdict` (the metadata flags container), and the protocol-2 builtin names (`__builtin__.long` / `unicode` / `bool` / `bytes` / `dict` / `float` / `list`, which torch remaps to `int` / `str` / ...). Without them the load still fails.

Why these are inert under torch's `_weights_only_unpickler`:

- None of the omegaconf classes define `__reduce__`, so there is no arbitrary-callable gadget. `NEWOBJ` resolves to plain `object.__new__`.
- The only `__setstate__` in the set is `omegaconf.base.Node.__setstate__`, which is `self.__dict__.update(state)` plus a cache reset — no dispatch, no side effects.
- `REDUCE` can only invoke an allowlisted callable, and every allowlisted callable here is an inert constructor. In particular `collections.defaultdict` cannot be weaponised: its `default_factory` argument must itself come through a `GLOBAL` opcode and therefore must also be allowlisted, and the allowlist contains only data constructors.

I checked this by construction rather than by inspection alone — attempts to load pickles carrying `os.system`, `defaultdict(os.system)`, `str(os.popen)`, and a `DictConfig` with an attacker-supplied `__dict__` are all rejected with `UnpicklingError` after the registration is applied. (`sys`/`os`/`posix`/`nt` are also on torch's hard module blocklist, which `add_safe_globals` cannot override.)

Two things worth stating plainly rather than burying:

1. Registration is process-global and happens at `import nemo.utils`, so it widens the weights-only allowlist for every `torch.load` in the process, not just NeMo's own. A `safe_globals()` context manager would be tighter, but the loads happen inside lightning, out of NeMo's reach. Given the classes are inert this seemed like the right trade; happy to scope it differently if you'd prefer.
2. A malicious checkpoint can now carry a `DictConfig` containing interpolations (e.g. `${oc.env:...}`). These do **not** resolve at load time — I verified `repr()`/`str()` do not trigger resolution — but they do resolve on attribute access by application code, so a hostile config can still read env vars when it is later used. That is inherent to NeMo consuming a config from a checkpoint at all, and is strictly narrower than the `weights_only=False` status quo.

Net effect on the pinned lightning 2.4.0: 2.4.0's `lightning.fabric.utilities.cloud_io._load` hard-codes `weights_only=False`, so nothing about NeMo's loading behavior changes there. The registration is purely additive.

# Why the cap existed, and why it can be relaxed

The cap was added in NVIDIA/NeMo#11692, opened 2024-12-20T22:48Z — about eight hours after lightning 2.5.0 was published (2024-12-20T14:21Z) — and merged the next morning. It is a one-line change, its PR body is the single line "Update requirements_lightning.txt", and there is no linked breakage, failing test, or discussion. The same one-line change landed on `r2.1.0` in parallel via NVIDIA/NeMo#11694, hours before #11692 merged to `main`. That pattern reads as a release-stability pin taken during the 2.1.0 freeze, not a fix for a known incompatibility. The follow-up relaxation attempt (NVIDIA/NeMo#11770) was closed unmerged.

An audit of every `lightning`/`lightning.fabric` import in `nemo/` — including the ~20 private `_symbol` imports (`_TrainingEpochLoop`, `_CheckpointConnector`, `_jit_is_scripting`, `_is_local_file_protocol`, `_ResultCollection`, `_WrappingCheckpointIO`, `_convert_fp_tensor`, `_TunerExitException`, ...) — confirms all of them still exist and import cleanly on lightning 2.6.5. The only real breakages are the three fixed above: two import-time (nv-one-logger, NeptuneLogger) and one runtime, the last being a direct downstream effect of the checkpoint-security change itself.

# Usage

```bash
# downstream services can now do:
pip install nemo-toolkit[asr] "lightning>=2.6.1"
```

# Testing done locally (macOS arm64, CPU only, Python 3.12, torch 2.12)

Exact commands and results. Nothing here ran on GPU, and no collection-level (asr/tts/audio/speechlm2) suites were run locally.

With **lightning 2.6.5 / pytorch-lightning 2.6.5** installed over the repo's `asr` + `common` extras:

| Command | Result |
| --- | --- |
| `pytest tests/core/test_exp_manager.py -m "not pleasefixme"` | **33 passed, 1 skipped, 1 deselected** (~25 min). Without the safe-globals change the same suite is **4 failed, 29 passed**, all four failing with `UnpicklingError` on omegaconf hparams: `test_nemo_checkpoint_restore_model`, `test_last_checkpoint_saved`, `test_resume_checkpoint_skip_validation`, `test_invalid_checkpoints_removed_from_topk`. |
| `pytest tests/core/test_exp_manager.py -k save_best_model` | 2 passed — these are the tests that exercise `NeMoModelCheckpoint.on_train_end`'s best-checkpoint re-load, and they pass with lightning 2.6's secure `weights_only` default. |
| `pytest tests/core/test_save_restore.py -m "not pleasefixme"` | 32 passed, 4 skipped (skips are pre-existing `--with_downloads` gates) |
| `pytest tests/core/test_optimizers_schedulers.py -m "not pleasefixme"` | 23 passed, 2 skipped |
| `pytest tests/lightning/test_one_logger_callback.py tests/core/test_serialization.py` | 23 passed, 8 skipped (the 8 skips need the real nv-one-logger PTL integration — see limitation below) |
| `pytest tests/core_ptl/test_resumable_dataloader_iter.py` | passed |
| import smoke test | `nemo.core`, `nemo.utils.exp_manager`, `nemo.utils.trainer_utils`, `nemo.utils.callbacks.nemo_model_checkpoint`, `nemo.utils.callbacks.dist_ckpt_io`, `ModelPT`, `CallbackGroup` all import and instantiate |

With **lightning 2.4.0** (what the lockfile still pins — regression check that nothing changes at the pinned version):

| Command | Result |
| --- | --- |
| `pytest tests/lightning/test_one_logger_callback.py tests/core/test_serialization.py` | **31 passed, 0 skipped** — the real nv-one-logger integration is still used |
| runtime introspection | `HAVE_ONE_LOGGER_PTL is True`, `exp_manager.NeptuneLogger` resolves to the real class, and `_CheckpointConnector.restore` has no `weights_only` parameter — i.e. every new branch takes its pre-existing path |

**Known local-environment noise, not caused by this change:** `tests/core_ptl/test_ptl_stateless_timer.py::TestStatelessTimer::test_stateless_timer` fails identically on lightning 2.4.0 and 2.6.5 on this host because it requests `strategy=ddp`, which macOS/MPS rejects.

**Not validated locally / deferred to CI** (community PR — needs a maintainer to comment `/ok to test 5ad1ba76f690061fbd19a1ba33be04262b8d414d`):

- the full unit + integration matrix on the locked Linux environment (lightning 2.4.0 + pytorch-lightning 2.6.1)
- all GPU and multi-GPU/distributed tests
- collection-level suites (asr / tts / audio / speechlm2)
- `tests/core/test_exp_manager.py::TestExpManager::test_trainer_neptune_logger` — always skipped, because `neptune` is not in any dependency group. Worth flagging: its final assertion is `isinstance(test_trainer.logger, pl.loggers.NeptuneLogger)`, which would raise `AttributeError` on lightning >= 2.6 if anyone did install `neptune`. I left it alone since CI never reaches it and touching it would be unrelated churn, but it should get a version guard whenever the lockfile does move to 2.6.

# Known limitation / follow-up for maintainers

`nv-one-logger` (<= 2.3.1, the latest release) cannot be imported against lightning >= 2.6, so OneLogger telemetry is disabled (with a warning) in environments that choose lightning >= 2.6. This PR intentionally keeps the *lockfile* at lightning 2.4.0 so NVIDIA CI and the containers keep full telemetry; bumping the locked lightning to 2.6.x can be a separate decision once an nv-one-logger release compatible with the lightning 2.6 `Trainer.save_checkpoint` signature ships. That release has to come from NVIDIA — it is the one part of this that I cannot fix from outside.

# GitHub Actions CI

**Pre checks**:

  - [x] Reviewer: Does the PR have correct import guards for all optional libraries?

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information

- Related to #15706 (community request; CVE-2026-31221 affects `lightning<=2.4.0` per the advisory referenced there — NVD analysis was still pending at the time the issue was filed)
- Previous bump attempt that went stale: NVIDIA/NeMo#11770; cap introduced without stated rationale in NVIDIA/NeMo#11692 (and NVIDIA/NeMo#11694 on `r2.1.0`)


udsy19:

PR #16096, opened 2026-08-18
fix(asr): stop GreedyTDTInfer skipping a frame when the symbol budget is exhausted

### LINKED ISSUE #16095: GreedyTDTInfer skips an extra encoder frame when the symbol budget is exhausted
**Describe the bug**

`GreedyTDTInfer` (TDT models decoded with `strategy: greedy`) advances the encoder
time index one frame too far whenever the per-frame symbol budget is exhausted.
With `max_symbols_per_step=1` the budget is exhausted on *every* frame, so the
decoder skips `duration + 1` frames each step: tokens are dropped and every
surviving timestamp is late.

`nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py:2726-2747`
(`GreedyTDTInfer._greedy_decode`):

```python
                # Increment token counter.
                symbols_added += 1
                time_idx += skip          # TDT: advance by the predicted duration
                need_loop = skip == 0

            # this rarely happens, but we manually increment the `skip` number
            # if blank is emitted and duration=0 is predicted. This prevents possible
            # infinite loops.
            if skip == 0:
                skip = 1
            ...
            if symbols_added == self.max_symbols:
                time_idx += 1             # fires even when time has already advanced
```

The inner loop exits either because a non-zero duration was predicted (`need_loop`
is `False`, and `time_idx` has already been advanced by that duration) or because
the symbol budget ran out. The forced `time_idx += 1` is only needed in the second
case *when the last prediction had duration 0* — but as written it fires in the
first case too.

CPU-only, no downloads, no checkpoint. A constant joint makes the correct answer
computable by hand: the model always emits token 1 with duration 2, so with `T`
frames the tokens must land at `t = 0, 2, 4, ...`.

```python
import torch
import nemo.collections.asr.parts.submodules.rnnt_greedy_decoding as gd
from nemo.collections.asr.modules import RNNTDecoder, RNNTJoint

TOKENS = [" ", "a", "b", "c"]
V, BLANK, D = len(TOKENS), len(TOKENS), 4
DURATIONS = [0, 1, 2, 4]


class ConstJoint(RNNTJoint):
    def joint_after_projection(self, f, g):
        B = f.shape[0]
        T = f.shape[1] if f.dim() == 3 else 1
        U = g.shape[1] if g.dim() == 3 else 1
        out = torch.full((B, T, U, (V + 1) + len(DURATIONS)), -10.0)
        out[..., 1] = 5.0                    # always token 1
        out[..., (V + 1) + 2] = 5.0          # always duration DURATIONS[2] == 2
        return out


def run(cls, ms, T, **kw):
    torch.manual_seed(0)
    dec = RNNTDecoder({'pred_hidden': D, 'pred_rnn_layers': 1}, V)
    jnt = ConstJoint({'encoder_hidden': D, 'pred_hidden': D, 'joint_hidden': D, 'activation': 'relu'},
                     V, vocabulary=TOKENS, num_extra_outputs=len(DURATIONS))
    dec.eval(); jnt.eval()
    d = cls(dec, jnt, blank_index=BLANK, durations=DURATIONS,
            max_symbols_per_step=ms, include_duration=True, **kw)
    with torch.no_grad():
        h = d(encoder_output=torch.zeros(1, D, T),
              encoded_lengths=torch.tensor([T], dtype=torch.int32))[0][0]
    return list(map(int, h.timestamp)), list(map(int, h.token_duration))


T = 10
print("expected              :", list(range(0, T, 2)))
ts, dur = run(gd.GreedyTDTInfer, 1, T)
print("GreedyTDTInfer  ms=1  :", ts, "token_duration:", dur)
ts, dur = run(gd.GreedyBatchedTDTInfer, 1, T, use_cuda_graph_decoder=False)
print("GreedyBatchedTDT ms=1 :", ts, "token_duration:", dur)
```

Output:

```
expected              : [0, 2, 4, 6, 8]
GreedyTDTInfer  ms=1  : [0, 3, 6, 9] token_duration: [2, 2, 2, 2]
GreedyBatchedTDT ms=1 : [0, 2, 4, 6, 8] token_duration: [2, 2, 2, 2, 2]
```

`GreedyTDTInfer` emits 4 tokens where 5 are correct, and every timestamp after the
first is late. It still reports `token_duration = [2, 2, 2, 2]`, so the hypothesis
is internally inconsistent: `timestamp[k] + token_duration[k] != timestamp[k + 1]`.

`preserve_alignments=True` on the same input shows the same desync from the other
side — the alignment buffer is built from `skip` while the time index advances by
`skip + 1`:

```
shipped:  T=10  len(alignments)=8   timestamps=[0, 3, 6, 9]   non-empty slots=[0, 2, 4, 6]
```

Ten encoder frames produce only eight alignment slots, and the slots that hold
logits (`0, 2, 4, 6`) do not agree with the reported timestamps (`0, 3, 6, 9`).

**It is not limited to `max_symbols_per_step=1`.** The budget can also be exhausted
on a non-zero duration when an earlier step in the same frame predicted duration 0.
With a joint that alternates duration 0 and duration 2 on successive calls and
`max_symbols_per_step=2`, the correct timestamps are `[0, 0, 2, 2, 4, 4, 6, 6, 8, 8]`
and `GreedyTDTInfer` returns `[0, 0, 3, 3, 6, 6, 9, 9]`.

Over a sweep of random weights (3 seeds x 3 duration sets x 3 batch shapes x
`max_symbols_per_step` in {1, 2, 3, 5} = 468 utterance comparisons),
`GreedyTDTInfer` disagreed with `GreedyBatchedTDTInfer` on `(tokens, timestamps)`
for 135 utterances — 111/117 at `max_symbols_per_step=1`, and 9, 9 and 6 at 2, 3
and 5 respectively.

**Expected behavior**

`GreedyTDTInfer` should advance the time index by exactly the predicted duration,
and force a one-frame advance only when the predicted duration was 0 and the symbol
budget stopped the loop — which is what `GreedyBatchedTDTInfer` does.

Since the inner loop can only exit with `skip == 0` because the budget ran out, the
forced advance folds into the existing zero-duration branch and the trailing
conditional can be deleted:

```python
            if skip == 0:
                skip = 1
                time_idx += 1
            ...
            # if symbols_added == self.max_symbols:   <- delete
            #     time_idx += 1
```

With that change all 468 comparisons above agree, and the alignment buffer length
matches the encoder length again.

**Environment overview**

- Environment location: Bare-metal (macOS arm64), CPU only
- Method of NeMo install: from source, `main` @ `d0a07c056`

**Environment details**

- PyTorch 2.12.0
- Python 3.10.18

udsy19:
# What does this PR do ?

Stops `GreedyTDTInfer._greedy_decode` advancing the encoder time index one frame too far whenever
the per-frame symbol budget is exhausted. With `max_symbols_per_step=1` the budget is exhausted on
every frame, so the decoder skipped `duration + 1` frames every step — dropping tokens and making
every surviving timestamp late.

**Collection**: ASR

# Changelog

- `nemo/collections/asr/parts/submodules/rnnt_greedy_decoding.py`
  - `GreedyTDTInfer._greedy_decode`: move the forced `time_idx += 1` into the existing
    zero-duration branch and delete the trailing `if symbols_added == self.max_symbols:` advance
    (net −2 lines). The inner loop has two exits: `need_loop` became `False` (`skip != 0`, so
    `time_idx` has *already* advanced by the predicted duration), or the symbol budget ran out. The
    forced advance is only needed for the second exit **with `skip == 0`**, but it fired for every
    budget-exhausted exit, including those that also satisfied the first exit's condition.
- `tests/collections/asr/test_asr_rnnt_encdec_model.py`
  - `test_tdt_greedy_decoding_advances_by_predicted_duration` — constant joint (always token 1,
    duration 2), parametrized over `max_symbols_per_step` in {1, 2, 5}; asserts the hand-computed
    timestamps and durations for both `GreedyTDTInfer` and `GreedyBatchedTDTInfer`.
  - `test_tdt_greedy_decoding_exhausted_symbol_budget` — alternating-duration joint; covers budget
    exhaustion on a non-zero duration with `max_symbols_per_step=2`.

  Both fail on `main` and pass with this change, and no other test in the file changes status.

# Why this is wrong today

Two independent symptoms of the same off-by-one:

- `token_duration` still reports the predicted duration, so the returned hypothesis is
  self-contradictory — `timestamp[k] + token_duration[k] != timestamp[k + 1]`. Since
  `rnnt_decoding.py` enables `tdt_include_token_duration` automatically when `compute_timestamps`
  is set, and `_compute_offsets_tdt` builds `end_offset` as `start + duration`, both the start and
  the end offsets are wrong.
- The `preserve_alignments` / `preserve_frame_confidence` buffers are extended by `skip` entries per
  outer iteration, so they desync from `time_idx` by exactly the same one frame.

## Could this allow a non-advancing loop?

This is the obvious reviewer objection, since the deleted branch sits next to the comment about
infinite-loop protection. It cannot, and both the argument and the measurement are below.

`max_symbols_per_step` is validated `> 0` (or `None`) in `_GreedyRNNTInfer.__init__`, so the inner
loop body always executes at least once and `skip` is always bound. Every outer iteration then
advances `time_idx` by at least 1:

| exit | `skip` | before | after |
|---|---|---|---|
| `need_loop == False` | `!= 0` | `+skip` | `+skip` (unchanged) |
| budget exhausted, last duration 0 | `== 0` | `+1` (deleted branch) | `+1` (new line) — **identical** |
| budget exhausted, last duration `!= 0` | `!= 0` | `+skip+1` | `+skip` — **the bug being fixed** |
| `max_symbols is None` | `!= 0` | `+skip` | `+skip` (unchanged) |

A `skip == 0` exit implies `symbols_added == self.max_symbols`: the loop condition is
`need_loop and (max_symbols is None or symbols_added < max_symbols)` and `need_loop` is `skip == 0`,
so a `skip == 0` exit can only come from the budget clause. The old and new guards therefore fire on
exactly the same states in the zero-duration case. When `max_symbols is None` the loop cannot exit
with `skip == 0` at all, and the old `int == None` comparison was always `False`, so behaviour there
is unchanged too.

# Usage

Constant joint (always token 1, duration 2) makes the ground truth hand-computable — tokens must
land at `t = 0, 2, 4, …`:

```
expected              : [0, 2, 4, 6, 8]
GreedyTDTInfer  ms=1  : [0, 3, 6, 9]     token_duration: [2, 2, 2, 2]      <- before
GreedyTDTInfer  ms=1  : [0, 2, 4, 6, 8]  token_duration: [2, 2, 2, 2, 2]   <- after
GreedyBatchedTDT ms=1 : [0, 2, 4, 6, 8]  token_duration: [2, 2, 2, 2, 2]
```

Not limited to `max_symbols_per_step=1`: with a joint alternating duration 0 and 2 per call and
`max_symbols_per_step=2`, the correct answer is `[0, 0, 2, 2, 4, 4, 6, 6, 8, 8]` and the shipped
decoder returns `[0, 0, 3, 3, 6, 6, 9, 9]`.

# Validation

Random-weight sweep against `GreedyBatchedTDTInfer` as the reference — 3 seeds x 3 duration sets x
3 batch shapes x `max_symbols_per_step` in {1, 2, 3, 5}, 468 utterance-level
`(tokens, timestamps)` comparisons, counting disagreements:

```
before: ms=1: 111/117   ms=2: 9/117   ms=3: 9/117   ms=5: 6/117   total 135/468
after:  ms=1:   0/117   ms=2: 0/117   ms=3: 0/117   ms=5: 0/117   total   0/468
```

Alignment buffers, `preserve_alignments=True`, `T=10`, `ms=1`:

```
before: len(alignments)=8    timestamps=[0, 3, 6, 9]      non-empty slots=[0, 2, 4, 6]
after:  len(alignments)=10   timestamps=[0, 2, 4, 6, 8]   non-empty slots=[0, 2, 4, 6, 8]
```

Termination bound. "Every outer iteration advances by at least 1" implies the joint is called at
most `T * max_symbols_per_step` times. Adversarial sweep of 135 cases — a joint that always predicts
duration 0 with a non-blank token, one that always predicts duration 0 with blank, and random
logits, over 3 duration sets each containing 0, `max_symbols_per_step` in {1, 2, 3, 5, 10}, 3 seeds,
`T=25`, with a 20 s watchdog:

```
before: cases: 135   worst joint-calls / (T * max_symbols_per_step): 1.000   violations: 0
after:  cases: 135   worst joint-calls / (T * max_symbols_per_step): 1.000   violations: 0
```

The bound is tight (the all-zero-duration case hits it exactly) and is not exceeded before or after.
No hangs. The fix does not weaken termination.

Test suite:

```
# on main, with the new tests applied
FAILED ...::test_tdt_greedy_decoding_advances_by_predicted_duration[1]
    assert [0, 3, 6, 9] == [0, 2, 4, 6, 8]
FAILED ...::test_tdt_greedy_decoding_exhausted_symbol_budget
    assert [0, 0, 3, 3, 6, 6, ...] == [0, 0, 2, 2, 4, 4, ...]
2 failed, 2 passed

# with this change
4 passed, 64 deselected

# whole file
pytest tests/collections/asr/test_asr_rnnt_encdec_model.py -m "not pleasefixme"
62 passed, 6 deselected
```

Environment: Python 3.10.18, PyTorch 2.12.0, CPU-only, `main` at `d0a07c056`.

# Scope

Only `GreedyTDTInfer` (TDT models, `strategy: greedy`) is touched. `GreedyBatchedTDTInfer`,
`GreedyRNNTInfer`, `GreedyMultiblankRNNTInfer` and the beam decoders are unchanged — the batched TDT
decoder is used here only as the reference the fixed decoder now agrees with.

**Pre checks**:

- [ ] Did you add or update any necessary documentation? — not needed, no user-facing API change

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

# Additional Information

- Fixes #16095


artbataev (approved):
Thanks for the fix, makes sense. Approved.

PR #14809, opened 2025-09-25
Clipping, lowpass and lossy codec online augmentations for Lhotse dataloader/sampler

racoiaws:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

Adds three new randomized augmentations: soft and hard clipping, lossy codec copression (Opus/MP3/Vorbis/GSM), and lowpass filters. Can be used in any training run that uses Lhotse dataloaders/samplers.

Depends on latest Lhotse features https://github.com/lhotse-speech/lhotse/pull/1512 https://github.com/lhotse-speech/lhotse/pull/1510
as well as on a PR that is currently in review https://github.com/lhotse-speech/lhotse/pull/1511

**Collection**: common

# Changelog

- Add clipping/saturation augmentation
- Add lowpass augmentation (via resampling back-and-forth)
- Add lossy codec augmentation (Opus/MP3/Vorbis/GSM)

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information

- Needs to be updated when https://github.com/NVIDIA-NeMo/NeMo/pull/14322 gets merged, likely via a new PR

PR #14322, opened 2025-07-24
Flexible augmentation config for Lhotse dataloaders

pzelasko:
> [!IMPORTANT]  
> The `Update branch` button must only be pressed in very rare occassions.
> An outdated branch is never blocking the merge of a PR.
> Please reach out to the automation team before pressing that button.

# What does this PR do ?

Adds two new configuration fields `augment_examples` and `augment_batch` that let the user specify their own augmentations in a custom order.

This is following a design descussion with @stevehuang52. 

**Collection**: ASR, Audio, speechlm2

# Changelog

- Add specific line by line info of high level changes in this PR.

# Usage

- You can potentially add a usage example below

```python
# Add a code snippet demonstrating how to use this 
```

# GitHub Actions CI

The Jenkins CI system has been replaced by GitHub Actions self-hosted runners.

The GitHub Actions CI will run automatically when the "Run CICD" label is added to the PR.
To re-run CI remove and add the label again.
To run CI on an untrusted fork, a NeMo user with write access must first click "Approve and run".

# Before your PR is "Ready for review"

**Pre checks**:

- [ ] Make sure you read and followed [Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md)
- [ ] Did you write any new necessary tests?
- [ ] Did you add or update any necessary documentation?
- [ ] Does the PR affect components that are optional to install? (Ex: Numba, Pynini, Apex etc)
  - [ ] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.
[Contributor guidelines](https://github.com/NVIDIA/NeMo/blob/main/CONTRIBUTING.md) contains specific people who can review PRs to various areas.

# Additional Information

- Related to # (issue)

PR #16101, opened 2026-08-18
fix(lhotse): seed the RIR augmentation from shard_seed, matching its sibling transforms

### LINKED ISSUE #16100: Lhotse RIR augmentation is seeded from seed instead of shard_seed, so all ranks draw identical reverberation
**Describe the bug**

In `nemo/collections/common/data/lhotse/dataloader.py`, the RIR augmentation is the
only augmentation transform attached to the sampler that is seeded from
`config.seed` instead of `config.shard_seed`:

```python
# nemo/collections/common/data/lhotse/dataloader.py:1011-1018
if config.rir_enabled:
    sampler = sampler.map(
        ReverbWithImpulseResponse(
            rir_recordings=RecordingSet.from_file(config.rir_path) if config.rir_path is not None else None,
            p=config.rir_prob,
            randgen=random.Random(config.seed),      # <-- here
        )
    )
```

Every sibling transform in the same function uses `seed=config.shard_seed`:

| line | transform | seed argument |
|------|-----------|---------------|
| 861  | on-the-fly noise mix (`cuts.mix`) | `seed=config.shard_seed` |
| 939  | `DynamicBucketingSampler` | `seed=config.shard_seed` |
| 962  | `DynamicCutSampler` | `seed=config.shard_seed` |
| 995  | `LowpassUsingResampling` | `seed=config.shard_seed` |
| 1007 | `ClippingTransform` | `seed=config.shard_seed` |
| **1016** | **`ReverbWithImpulseResponse`** | **`randgen=random.Random(config.seed)`** |
| 1032 | `Compress` | `seed=config.shard_seed` |

`shard_seed` defaults to `"trng"` (`:109`) and is resolved *inside each process*
(`resolve_seed("trng")` is `secrets.randbelow(2**31)`), which is what makes ranks
differ. `config.seed` defaults to `0` (`:162`) and is resolved to a concrete
integer up front, before the sampler is built:

```python
# :584-585  (and the same at :678 for the multi-config entry point)
config.seed = resolve_seed(config.seed)
fix_random_seed(config.seed)
```

so it is bit-identical on every data-parallel rank.

**Net effect:** every DP rank applies the same reverberation coin-flip sequence and
selects the same impulse responses for the n-th cut it sees. (Each rank still gets
a *different subset* of cuts — DDP sharding itself is fine — but the augmentation
*decision sequence* is duplicated across ranks.) Effective RIR augmentation
diversity is far lower than the config implies. Nothing errors; the only symptom is
a model that generalises worse than expected.

Build the sampler for rank 0 and rank 1 and compare the RNG the RIR transform holds.
This runs on CPU with no model and no real data.

```python
import tempfile
from pathlib import Path

from lhotse import CutSet
from lhotse.testing.dummies import DummyManifest
from lhotse.dataset import LowpassUsingResampling, ReverbWithImpulseResponse

from nemo.collections.common.data.lhotse.dataloader import get_lhotse_dataloader_from_config


class Identity:
    def __getitem__(self, cuts):
        return cuts


root = Path(tempfile.mkdtemp())
cuts = DummyManifest(CutSet, begin_id=0, end_id=10, with_data=True)
for c in cuts:
    c.features = None
    c.custom = None
    c.supervisions[0].custom = None
p = root / "cuts.jsonl.gz"
cuts.save_audios(root / "audio").to_file(p)

config = OmegaConf.create(
    {
        "cuts_path": str(p),
        "batch_size": 2,
        "seed": 0,
        "shard_seed": "trng",   # the default
        "rir_enabled": True,
        "rir_prob": 0.5,
        "lowpass_enabled": True,   # a sibling augmentation, used as a control
        "lowpass_prob": 0.5,
    }
)


def build(rank):
    return get_lhotse_dataloader_from_config(
        config=config, global_rank=rank, world_size=2, dataset=Identity()
    ).sampler


def find(sampler, cls):
    return [t for t in sampler._transforms if isinstance(t, cls)][0]


s0, s1 = build(0), build(1)

rir0, rir1 = find(s0, ReverbWithImpulseResponse), find(s1, ReverbWithImpulseResponse)
print("RIR rank0   :", [round(rir0.random.random(), 6) for _ in range(6)])
print("RIR rank1   :", [round(rir1.random.random(), 6) for _ in range(6)])

lp0, lp1 = find(s0, LowpassUsingResampling), find(s1, LowpassUsingResampling)
print("lowpass rank0:", [round(lp0.rng.random(), 6) for _ in range(6)])
print("lowpass rank1:", [round(lp1.rng.random(), 6) for _ in range(6)])
```

Output (the `lowpass` values come from the OS TRNG and differ on every run; only the
rank0 != rank1 relationship is stable. The `RIR` values are `random.Random(0)` and are
reproducible):

```
RIR rank0   : [0.844422, 0.757954, 0.420572, 0.258917, 0.511275, 0.404934]
RIR rank1   : [0.844422, 0.757954, 0.420572, 0.258917, 0.511275, 0.404934]
lowpass rank0: [0.680268, 0.335713, 7.5e-05, 0.819685, 0.195258, 0.820385]
lowpass rank1: [0.620671, 0.550698, 0.337479, 0.266303, 0.173753, 0.753269]
```

The RIR streams are bit-identical (they are `random.Random(0)`, i.e. the default
`seed: 0`). The sibling `LowpassUsingResampling`, seeded from `shard_seed`, differs
— that is the control showing the mechanism works and the harness is sound.

I also ran this with rank 0 and rank 1 in two genuinely separate OS processes and
iterated the samplers end to end. Because `reverb_rir(affix_id=True)` appends
`_rvb` to the cut id, the per-position decisions are directly readable:

```
rank0 cut order        : [0000, 0001, 0004, 0005, 0008, 0009, 0012, 0013, 0016, 0017]
rank1 cut order        : [0002, 0003, 0006, 0007, 0010, 0011, 0014, 0015, 0018, 0019]
rank0 reverb applied?  : [0, 0, 1, 1, 0, 1, 0, 1, 1, 0]
rank1 reverb applied?  : [0, 0, 1, 1, 0, 1, 0, 1, 1, 0]     <-- identical
rank0 lowpass applied? : [0, 0, 0, 0, 0, 1, 0, 1, 1, 0]
rank1 lowpass applied? : [1, 0, 0, 0, 0, 1, 0, 1, 0, 0]     <-- differs (control)
```

**Expected behavior**

RIR augmentation randomness is differentiated per rank/worker via `shard_seed`,
like every other online augmentation in the same function.

**Environment overview**
- Environment location: bare-metal virtualenv
- Method of install: source checkout of `main` @ `0361552a6`

**Environment details**
- OS: macOS 26.5.1 (darwin, arm64)
- Python 3.10.18 / PyTorch 2.12.0 / lhotse 2.0.0a3

**Additional context**

CPU only; no GPU, model download, or real dataset needed.

Note that `lhotse.dataset.ReverbWithImpulseResponse` (the CutSet-level transform
imported here — not the same-named `lhotse.augmentation` AudioTransform) accepts
only a pre-built `randgen: random.Random` and has **no** `seed` parameter, so this
is not a kwarg rename. The siblings do `self.rng = random.Random(resolve_seed(self.seed))`
in their own `__post_init__`; the same has to be done at the call site here:

```diff
-                randgen=random.Random(config.seed),
+                randgen=random.Random(resolve_seed(config.shard_seed)),
```

`resolve_seed` is already imported in this module (`:40`). With an explicit integer
`shard_seed` this is a no-op (`resolve_seed(n) == n`), so reproducible runs and the
existing `test_dataloader_with_synth_rir` (which sets `shard_seed: 0`) are unaffected.

Related: open PR #14322 ("Flexible augmentation config for Lhotse dataloaders")
moves this block into a new `augment.py` and carries the `config.seed` line over
unchanged, but the new `ReverbRIR` dataclass it introduces there already does
`rng = random.Random(resolve_seed(self.seed))` with `seed: int | str = "trng"` —
i.e. exactly the behaviour proposed above. This report is about the current
default path on `main`.

Happy to open a PR; I have the fix and a regression test ready.

udsy19:
# What does this PR do ?

Seeds the Lhotse RIR augmentation from `shard_seed` instead of `seed`, so that
data-parallel ranks stop drawing bit-identical reverberation decisions and impulse
responses.

Fixes #16100.

**Collection**: common (ASR, Audio, speechlm2 — anything using the Lhotse dataloader
with `rir_enabled: true`)

## The problem

`get_lhotse_sampler_from_config` built the RIR transform with
`randgen=random.Random(config.seed)` (`nemo/collections/common/data/lhotse/dataloader.py:1016`).
`config.seed` is resolved to a concrete integer *before* the sampler is built
(`config.seed = resolve_seed(config.seed)`, `:584`, and `:678` for the multi-config
entry point), so it is bit-identical on every rank.

Every sibling augmentation attached to the same sampler already uses
`seed=config.shard_seed` — the noise mix (`:861`), `LowpassUsingResampling` (`:995`),
`ClippingTransform` (`:1007`) and `Compress` (`:1032`). `shard_seed` defaults to
`"trng"` and is resolved inside each process, which is what differentiates ranks.

Result: each rank applied the same reverb coin-flip sequence and picked the same
impulse response for the n-th cut it saw. DDP sharding itself was fine — each rank
still got a different subset of cuts — but the augmentation *decision sequence* was
duplicated across the whole job. Nothing errored; the only symptom was weaker
augmentation than the config implied.

# Changelog

- `nemo/collections/common/data/lhotse/dataloader.py`: pass
  `randgen=random.Random(resolve_seed(config.shard_seed))` to
  `ReverbWithImpulseResponse`, with a comment explaining why the resolution happens
  at the call site.
- `tests/collections/common/test_lhotse_multirank_rng.py`: two regression tests.

## Why `resolve_seed(...)` rather than `seed=config.shard_seed`

`lhotse.dataset.ReverbWithImpulseResponse` (the CutSet-level transform imported
here — distinct from the same-named `lhotse.augmentation` AudioTransform) takes only
`randgen: random.Random` and has no `seed` parameter, so passing `seed=` would raise
`TypeError`. The sibling transforms accept `seed` and then do
`self.rng = random.Random(resolve_seed(self.seed))` in `__post_init__`; this change
performs that same step at the call site. `resolve_seed` was already imported in this
module.

# Usage

```yaml
# Any Lhotse dataloader config with RIR augmentation enabled; no config change is required.
rir_enabled: true
rir_path: /path/to/rir_cuts.jsonl.gz
shard_seed: trng      # the default
```

With this change each data-parallel rank draws its own reverberation decisions and impulse
responses under the default `shard_seed`. An explicit integer `shard_seed` keeps every rank
identical, exactly as before.

## Backwards compatibility

With an explicit integer `shard_seed`, `resolve_seed(n) == n`, so every rank still
shares the RNG and fully reproducible runs are unaffected. The existing
`test_dataloader_with_synth_rir` (`tests/collections/common/test_lhotse_dataloading.py`,
`shard_seed: 0`) keeps its exact expectations about which cuts get reverberated and
passes unchanged.

**Scope.** This differentiates ranks under the default `shard_seed="trng"`. With
`shard_seed="randomized"` the RIR RNG stays identical across ranks, because
`resolve_seed("randomized")` only returns a per-rank/per-worker value inside a dataloader
worker, and these sampler-level `.map` transforms are constructed in the main process.
That limitation is pre-existing and shared with `LowpassUsingResampling`,
`ClippingTransform` and `Compress`; this change brings the RIR transform to parity with
them rather than fixing it, and the remaining gap is worth a separate issue.

## Tests

`tests/collections/common/test_lhotse_multirank_rng.py`:

- `test_rir_augmentation_rng_differs_across_ranks` — builds rank-0 and rank-1
  dataloaders under the default `shard_seed="trng"` and asserts the RIR RNG streams
  differ. **This test fails on `main`**, where both ranks produce the
  `random.Random(0)` stream (`0.8444218515250481, 0.7579544029403025, ...`). It also
  asserts the same for the sibling `LowpassUsingResampling` as a control, so the
  assertion is testing the seed source rather than the harness.
- `test_rir_augmentation_rng_reproducible_with_fixed_shard_seed` — asserts an
  explicit integer `shard_seed` still gives both ranks identical RIR randomness.
  This one passes both before and after; it is a guard against a future change that
  makes RIR unconditionally random and breaks reproducible runs.

## Checks run

```
pytest tests/collections/common/test_lhotse_multirank_rng.py      -> 4 passed
pytest tests/collections/common/test_lhotse_dataloading.py        -> 67 passed, 1 skipped
pytest tests/collections/common/ -k lhotse                        -> 279 passed, 1 skipped
git diff --check                                                  -> clean
```

Environment: Python 3.10.18, PyTorch 2.12.0, lhotse 2.0.0a3, CPU only.

## Documentation

No documentation change. `docs/source/dataloaders.rst` already documents
`shard_seed` as the knob controlling per-rank/per-worker randomness; this change
makes the RIR transform conform to that documented contract rather than altering it.

One thing I deliberately left alone: that section describes `shard_seed` only in
terms of *shard* randomization and does not mention that it also seeds the online
augmentation RNGs. That gap predates this change — the on-the-fly noise mix has been
seeded from `shard_seed` since well before the clipping/lowpass/codec augmentations were
added in #14809 — so I kept it out of scope; happy to add a clarifying sentence here if
you'd prefer it in this PR.

## Related

Open PR #14322 ("Flexible augmentation config for Lhotse dataloaders") relocates this
block into a new `augment.py` and carries the `config.seed` line across unchanged.
Its new `ReverbRIR` dataclass, however, already does
`rng = random.Random(resolve_seed(self.seed))` with `seed: int | str = "trng"` — the
same behaviour this PR gives the current default path. If #14322 lands first, this
change should be applied to the moved line; I'm happy to rebase either way.

# GitHub Actions CI

**Pre checks**:

- [x] Did you add or update any necessary documentation? (N/A — see Documentation above)

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Fixes #16100
- Related open PR #14322 ("Flexible augmentation config for Lhotse dataloaders") relocates the changed block; see "Related" above.

### DIFF
--- a/nemo/collections/common/data/lhotse/dataloader.py
+++ b/nemo/collections/common/data/lhotse/dataloader.py
@@ -1013,7 +1013,13 @@ def get_lhotse_sampler_from_config(config, global_rank, world_size, tokenizer=No
             ReverbWithImpulseResponse(
                 rir_recordings=RecordingSet.from_file(config.rir_path) if config.rir_path is not None else None,
                 p=config.rir_prob,
-                randgen=random.Random(config.seed),
+                # ``ReverbWithImpulseResponse`` takes a pre-built ``random.Random`` rather than a
+                # ``seed`` argument, so resolve ``shard_seed`` here to give every rank/worker its
+                # own RIR randomness. This mirrors what the sibling augmentations above and below
+                # do internally with ``seed=config.shard_seed``. Using ``config.seed`` would make
+                # every rank draw the same impulse responses, because it is resolved to a fixed
+                # integer before the sampler is built.
+                randgen=random.Random(resolve_seed(config.shard_seed)),
             )
         )
 
--- a/nemo/collections/common/data/lhotse/dataloader.py
+++ b/nemo/collections/common/data/lhotse/dataloader.py
@@ -1014,7 +1014,7 @@ def get_lhotse_sampler_from_config(config, global_rank, world_size, tokenizer=No
                 rir_recordings=RecordingSet.from_file(config.rir_path) if config.rir_path is not None else None,
                 p=config.rir_prob,
                 # ``ReverbWithImpulseResponse`` takes a pre-built ``random.Random`` rather than a
-                # ``seed`` argument, so resolve ``shard_seed`` here to give every rank/worker its
+                # ``seed`` argument, so resolve ``shard_seed`` here to give every rank its
                 # own RIR randomness. This mirrors what the sibling augmentations above and below
                 # do internally with ``seed=config.shard_seed``. Using ``config.seed`` would make
                 # every rank draw the same impulse responses, because it is resolved to a fixed
--- a/tests/collections/common/test_lhotse_multirank_rng.py
+++ b/tests/collections/common/test_lhotse_multirank_rng.py
@@ -201,3 +201,68 @@ def test_dataloader_multiple_ranks_trng(nemo_tarred_manifest_path: tuple[str, st
     assert b0_batches != b1_batches
     assert b0_incrseed_batches != b1_batches
     assert b0_batches != b0_incrseed_batches
+
+
+def _find_transform(sampler, cls):
+    """Return the first sampler-level transform of type ``cls`` (they are attached via ``sampler.map``)."""
+    matches = [t for t in sampler._transforms if isinstance(t, cls)]
+    assert matches, f"{cls.__name__} was not attached to the sampler; transforms={sampler._transforms}"
+    return matches[0]
+
+
+def _rir_dataloader(cutset_path: Path, rank: int, shard_seed):
+    config = OmegaConf.create(
+        {
+            "cuts_path": str(cutset_path),
+            "sample_rate": 16000,
+            "use_lhotse": True,
+            "num_workers": 0,
+            "batch_size": 2,
+            "seed": 0,
+            "shard_seed": shard_seed,
+            # RIR is the transform under test; lowpass is a sibling used as a control.
+            "rir_enabled": True,
+            "rir_prob": 0.5,
+            "lowpass_enabled": True,
+            "lowpass_prob": 0.5,
+        }
+    )
+    return get_lhotse_dataloader_from_config(config=config, global_rank=rank, world_size=2, dataset=_Identity())
+
+
+def test_rir_augmentation_rng_differs_across_ranks(cutset_path: Path):
+    """RIR augmentation must draw its randomness from ``shard_seed``, like every other online
+    augmentation attached to the sampler.
+
+    ``seed`` is resolved to a concrete integer before the sampler is built, so it is bit-identical
+    on every data-parallel rank; ``shard_seed`` defaults to ``"trng"`` and is resolved per process.
+    Seeding the RIR transform from ``seed`` therefore made every rank apply the same reverberation
+    coin flips and pick the same impulse responses, silently collapsing RIR augmentation diversity.
+    """
+    from lhotse.dataset import LowpassUsingResampling, ReverbWithImpulseResponse
+
+    dl0 = _rir_dataloader(cutset_path, rank=0, shard_seed="trng")
+    dl1 = _rir_dataloader(cutset_path, rank=1, shard_seed="trng")
+
+    rir0 = _find_transform(dl0.sampler, ReverbWithImpulseResponse)
+    rir1 = _find_transform(dl1.sampler, ReverbWithImpulseResponse)
+    assert [rir0.random.random() for _ in range(32)] != [rir1.random.random() for _ in range(32)]
+
+    # Control: a sibling augmentation that already used ``shard_seed`` differentiates the same way,
+    # so the assertion above is testing the seed source and not the harness.
+    lowpass0 = _find_transform(dl0.sampler, LowpassUsingResampling)
+    lowpass1 = _find_transform(dl1.sampler, LowpassUsingResampling)
+    assert [lowpass0.rng.random() for _ in range(32)] != [lowpass1.rng.random() for _ in range(32)]
+
+
+def test_rir_augmentation_rng_reproducible_with_fixed_shard_seed(cutset_path: Path):
+    """Counterpart to the test above: an explicit integer ``shard_seed`` must still give every rank
+    the same RIR randomness, so that fully reproducible runs remain possible."""
+    from lhotse.dataset import ReverbWithImpulseResponse
+
+    dl0 = _rir_dataloader(cutset_path, rank=0, shard_seed=1234)
+    dl1 = _rir_dataloader(cutset_path, rank=1, shard_seed=1234)
+
+    rir0 = _find_transform(dl0.sampler, ReverbWithImpulseResponse)
+    rir1 = _find_transform(dl1.sampler, ReverbWithImpulseResponse)
+    assert [rir0.random.random() for _ in range(32)] == [rir1.random.random() for _ in range(32)]


udsy19:

PR #16103, opened 2026-08-18
fix(core): make save_adapters/load_adapters round-trip

### LINKED ISSUE #16102: save_adapters/load_adapters do not round-trip: subset saves fail to load, and names are matched by substring
**Describe the bug**

`AdapterModelPTMixin.save_adapters()` / `load_adapters()` do not round-trip. There are two
independent defects, both in `nemo/core/classes/mixins/adapter_mixins.py`.

**(1) A file saved for a subset of adapters cannot be loaded with the documented default
`name=None`.**

`save_adapters` writes state-dict entries only for the adapters it was asked to save
(`:891-901`) but stores the *whole* model adapter config as `__cfg__` (`:929`).
`load_adapters(name=None)` then takes its list of adapters to restore from that config rather
than from the state dict (`:972-973`), and raises `KeyError` for every adapter the file never
contained.

**(2) Adapter state-dict entries are selected and stripped by substring, so adapters with
overlapping names contaminate each other.**

`module.adapter_layer` is an `nn.ModuleDict` holding every adapter of that module, so its
`state_dict()` keys are `<adapter name>.<...>` for all of them. `save_adapters` filters those
with `if adapter_name in k` (`:923`), so saving `an4` also serialises all of `an4_v2`'s tensors
into a file advertised as holding only `an4`. `load_adapters` then strips the name with
`k.replace(f"{adapter_name}.", "")` (`:1043`), which leaves the foreign keys intact, and the
per-module `load_state_dict(..., strict=True)` fails.

This is the workflow documented at `docs/source/core/adapters/intro.rst:132` ("storing just the
adapter module(s) separately from the Model, so that you can use the same base model, and share
just the Adapter modules") and taught in `tutorials/02_NeMo_Adapters.ipynb`.

Both reproducers use the fixtures already in
`tests/core/mixins/adapters/test_adapter_model_mixin.py`. CPU only, no GPU or pretrained model.

```python
import os, tempfile
os.environ["TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD"] = "1"   # see load_adapters() docstring, adapter_mixins.py:963

import torch
from tests.core.mixins.adapters.test_adapter_model_mixin import (
    DefaultAdapterModel, get_adapter_cfg, get_model_config,
)

# (1) save one of two adapters, then load with the documented default
m = DefaultAdapterModel(cfg=get_model_config(in_features=50))
m.add_adapter("encoder:en", cfg=get_adapter_cfg())
m.add_adapter("encoder:de", cfg=get_adapter_cfg())
with tempfile.TemporaryDirectory() as d:
    fp = os.path.join(d, "en.pt")
    m.save_adapters(fp, name="encoder:en")
    DefaultAdapterModel(cfg=get_model_config(in_features=50)).load_adapters(fp)
```

```
KeyError: "Requested to load adapter with name `encoder:de`, but could not the adapter in the
state dict. \nAvailable adapter names in state dict are: ['encoder:en']"
```

```python
# (2) two adapters where one name is a prefix of the other
m = DefaultAdapterModel(cfg=get_model_config(in_features=50))
m.add_adapter("encoder:an4", cfg=get_adapter_cfg())
m.add_adapter("encoder:an4_v2", cfg=get_adapter_cfg())
with tempfile.TemporaryDirectory() as d:
    fp = os.path.join(d, "an4.pt")
    m.save_adapters(fp, name="encoder:an4")
    print(sorted(torch.load(fp, weights_only=False)["encoder:an4"][0].keys()))
    DefaultAdapterModel(cfg=get_model_config(in_features=50)).load_adapters(fp, name="encoder:an4")
```

```
['an4.module.0.bias', 'an4.module.0.weight', 'an4.module.1.weight', 'an4.module.3.weight',
 'an4_v2.module.0.bias', 'an4_v2.module.0.weight', 'an4_v2.module.1.weight', 'an4_v2.module.3.weight']

RuntimeError: Error(s) in loading state_dict for LinearAdapter:
        Unexpected key(s) in state_dict: "an4_v2.module.0.weight", "an4_v2.module.0.bias",
        "an4_v2.module.1.weight", "an4_v2.module.3.weight".
```

Note the first line: the "small shareable adapter file" silently contains a second adapter's
weights.

Controls, to show neither result is an artefact of the reproducer:

- In (1), `load_adapters(fp, name="encoder:en")` succeeds, and
  `save_adapters(fp, name=None)` -> `load_adapters(fp)` succeeds.
- In (2), renaming `an4_v2` to `other` makes the same script succeed and produce a file with only
  the four `an4.*` keys.

**Expected behavior**

- `load_adapters(path)` restores exactly the adapters the file contains.
- A file saved for adapter `X` contains only `X`'s parameters, regardless of whether another
  adapter's name has `X` as a prefix.

**Environment overview**
- Environment location: bare-metal virtualenv
- Method of install: source checkout of `main` @ `d0a07c056`

**Environment details**
- OS: macOS 26.5.1 (darwin, arm64)
- PyTorch 2.12.0
- Python 3.10.18

**Additional context**

Both are prefix-vs-substring / wrong-source-of-truth issues:
`k.startswith(f"{adapter_name}.")` and `k[len(prefix):]` instead of `in` / `.replace()`, and
`name = list(state_dict.keys())` instead of `list(config.keys())`.

Unrelated but adjacent, and offered only as context: the duplicate-adapter-name guard at
`adapter_mixins.py:741` reads
`if hasattr(self, '_restoring_adapters') and self._restoring_adapters is not True:`.
`_restoring_adapters` exists only *during* restore (`:687` sets it, `:693` deletes it), so on the
normal `add_adapter` path `hasattr()` is `False` and the check never executes. Same-module
duplicates are still rejected by `AdapterModuleMixin.add_adapter` (`:260-263`), so only the
cross-module case (`encoder:tuning` then `decoder:tuning`) slips through. Enabling the guard
would be a behaviour change rather than a bug fix, so I have deliberately left it alone.

I have a branch with the fix for (1) and (2) plus regression tests and am happy to open a PR.

udsy19:
# What does this PR do ?

Fixes two independent defects that stop `AdapterModelPTMixin.save_adapters()` /
`load_adapters()` from round-tripping. One commit each.

**Collection**: core

**1. `load_adapters` restored the wrong set of adapters.**
`save_adapters` writes state-dict entries only for the adapters it was asked to save, but stores
the source model's *entire* adapter config as `__cfg__`. `load_adapters(name=None)` derived its
list from that config instead of from the state dict, so the documented default
`load_adapters(path)` raised `KeyError` for every adapter that existed in the source model but
was never written to the file. It now derives the list from the state dict, which is by
construction exactly what the file holds.

**2. Adapter state-dict keys were matched and stripped by substring.**
`module.adapter_layer` is an `nn.ModuleDict` holding every adapter of that module, so its keys
are `<adapter name>.<...>` for all of them. `if adapter_name in k` and
`k.replace(f"{adapter_name}.", "")` therefore let overlapping names collide: saving `an4` when
`an4_v2` also exists silently wrote all of `an4_v2`'s tensors into the file, and loading it
failed with `RuntimeError: Unexpected key(s) in state_dict`. Both sites now compare and strip
the exact `<adapter_name>.` prefix.

**Collection:** [Core]

# Changelog

- `nemo/core/classes/mixins/adapter_mixins.py`
  - `load_adapters`: derive the adapter list from the state dict rather than from the stored
    `__cfg__`.
  - `save_adapters` / `load_adapters`: select and strip adapter state-dict keys by exact
    `<adapter_name>.` prefix instead of substring.
- `tests/core/mixins/adapters/test_adapter_model_mixin.py`
  - `test_save_adapter_subset_then_load_with_default_name`
  - `test_save_load_adapter_with_name_that_is_prefix_of_another`

# Usage

```python
model.add_adapter("encoder:en", cfg=adapter_cfg)
model.add_adapter("encoder:de", cfg=adapter_cfg)

# ship only the English adapter
model.save_adapters("en.pt", name="encoder:en")

# recipient, using the documented default -- raised KeyError before this PR
other_model.load_adapters("en.pt")
```

# GitHub Actions CI

**Pre checks**:

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation

## Who can review?

# Additional Information

- Fixes #16102

- Both fixes are backward compatible. Files written with `save_adapters(name=None)` are
  unaffected by change 1 (the two key sources describe the same set). Files written by the old
  code that contain contaminated keys now load successfully under the new prefix strip, where
  before they raised; and files written by the new code load fine under old code, since a prefix
  match is a subset of a substring match.
- The two commits are independent - each one alone leaves the existing suite green and fixes only
  its own regression test - so they can be split if preferred.
- Verified locally with `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1`: `tests/core/mixins` and
  `tests/collections/asr/mixins/adapters` -> 143 passed, 2 skipped. Both new tests fail on `main` and pass with the fix.
- Note for reviewers running these tests locally: seven pre-existing tests in this file fail on
  `main` under PyTorch >= 2.6 unless `TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD=1` is set, as the
  `load_adapters` docstring instructs. That is unrelated to this PR (the same seven fail before
  and after); the two new tests set it via `monkeypatch` so they pass either way.

### DIFF
--- a/nemo/core/classes/mixins/adapter_mixins.py
+++ b/nemo/core/classes/mixins/adapter_mixins.py
@@ -918,9 +918,14 @@ class AdapterModelPTMixin(AdapterModuleMixin):
                             # inside of the state dict.
                             # It will be normalized in the corresponding `load_adapters()` call.
                             adapter_state_dict = module.adapter_layer.state_dict()
+                            # `adapter_layer` is a ModuleDict holding every adapter of this module,
+                            # so its keys are `<some adapter name>.<...>`. Match on the exact
+                            # `<adapter_name>.` prefix - a substring test would also pick up any
+                            # adapter whose name merely contains `adapter_name`.
+                            adapter_key_prefix = f'{adapter_name}.'
                             state_dict = {}
                             for k, v in adapter_state_dict.items():
-                                if adapter_name in k:
+                                if k.startswith(adapter_key_prefix):
                                     state_dict[k] = v
 
                             output_dict[key].append(state_dict)
@@ -1038,12 +1043,15 @@ class AdapterModelPTMixin(AdapterModuleMixin):
             for state, module in zip(adapter_state, modules_to_load):
                 # Note that state is a list of multiple state dicts for 1:1 Module mapping.
                 # However, the state_dict keys are of the form `adapter_name.<module hierarchy with dots>`.
-                # We therefore strip the `adapter_name.` part of the state dict
+                # We therefore strip the leading `adapter_name.` prefix of the state dict
                 # And then directly load each module with its 1:1 state dict.
+                # Only the exact prefix is matched/stripped, so keys belonging to another adapter
+                # whose name contains `adapter_name` are neither picked up nor mangled.
+                adapter_key_prefix = f'{adapter_name}.'
                 sub_dict = {}
                 for k, v in state.items():
-                    if adapter_name in k:
-                        k_ = k.replace(f"{adapter_name}.", "")
+                    if k.startswith(adapter_key_prefix):
+                        k_ = k[len(adapter_key_prefix) :]
                         sub_dict[k_] = v
 
                 module.load_state_dict(sub_dict, strict=strict)
--- a/nemo/core/classes/mixins/adapter_mixins.py
+++ b/nemo/core/classes/mixins/adapter_mixins.py
@@ -970,7 +970,10 @@ class AdapterModelPTMixin(AdapterModuleMixin):
             name = [name]
 
         if name is None:
-            name = list(config.keys())
+            # Restore exactly the adapters that this file contains. `config` is the *source model's*
+            # entire adapter config, which is a superset of the file whenever `save_adapters` was
+            # called with an explicit `name`, so it cannot be used to derive the list to restore.
+            name = list(state_dict.keys())
 
         # For all module:adapter names (note, for global modules, we ignore the module: part)
         for module_adapter_name in name:
--- a/tests/core/mixins/adapters/test_adapter_model_mixin.py
+++ b/tests/core/mixins/adapters/test_adapter_model_mixin.py
@@ -1141,6 +1141,40 @@ class TestAdapterModelMixin:
             assert (original_state_dict[ogkey] - restored_state_dict[newkey]).abs().mean() < 1e-6
 
     @pytest.mark.unit
+    def test_save_adapter_subset_then_load_with_default_name(self, monkeypatch):
+        # `load_adapters(filepath)` with the default `name=None` must restore exactly the
+        # adapters the file contains, even when the source model had more adapters than were
+        # saved. Otherwise the documented "share just the adapter module" workflow is unusable.
+        monkeypatch.setenv('TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD', '1')  # see load_adapters() docstring
+
+        cfg = get_model_config(in_features=50)
+        model = DefaultAdapterModel(cfg)
+
+        model.add_adapter(name='encoder:adapter_0', cfg=get_adapter_cfg(dim=5))
+        model.add_adapter(name='encoder:adapter_1', cfg=get_adapter_cfg(dim=5))
+
+        with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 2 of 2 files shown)


udsy19:

PR #16105, opened 2026-08-18
fix(asr): pad random_segment with an integer sample count

### LINKED ISSUE #16104: random_segment augmentation raises TypeError: pad_width must be of integral type whenever it has to pad
**Describe the bug**

The `random_segment` audio augmentation (`RandomSegmentPerturbation`) raises
`TypeError: `pad_width` must be of integral type.` whenever it actually has to pad, i.e.
whenever an utterance is shorter than `duration_sec` and `pad_to_duration: true` is set. It
survives only when `duration_sec` happens to be written as an integer in YAML — the class
default, `duration_sec=32.0`, and any float value crash.

Root cause, `nemo/collections/asr/parts/preprocessing/perturb.py:1098`:

```python
pad_size = self._duration_sec * data.sample_rate - data.num_samples
data.pad(pad_size=pad_size)
```

`duration_sec` is a float, so `pad_size` is a float, and `AudioSegment.pad` forwards it
straight to `numpy.pad` (`nemo/collections/asr/parts/preprocessing/segment.py:588,598`), which
requires an integral `pad_width`.

Separately, `docs/source/asr/ssl/configs.rst` presents `random_segment` as the way to keep SSL
batch durations uniform and states that "samples below the provided segment length will be
padded", but the snippet it gives does not set `pad_to_duration` and the default is `False`. So
with the documented config short utterances are passed through untouched — the opposite of what
the doc says, and it defeats the stated purpose of masking an equal number of patches per
utterance for the contrastive loss.

Together this means the documented SSL recipe either silently does not pad, or crashes if you
enable padding.

```python
import numpy as np
from nemo.collections.asr.parts.preprocessing.perturb import RandomSegmentPerturbation
from nemo.collections.asr.parts.preprocessing.segment import AudioSegment

seg = AudioSegment(np.zeros(16000 * 5, dtype=np.float32), 16000)   # 5 s of audio
p = RandomSegmentPerturbation(duration_sec=32.0, pad_to_duration=True)  # class defaults
p.perturb(seg)
```

```
  ...
  File "nemo/collections/asr/parts/preprocessing/perturb.py", line 1099, in perturb
    data.pad(pad_size=pad_size)
  File "nemo/collections/asr/parts/preprocessing/segment.py", line 598, in pad
    self._samples = np.pad(
  File "numpy/lib/_arraypad_impl.py", line 754, in pad
    raise TypeError('`pad_width` must be of integral type.')
TypeError: `pad_width` must be of integral type.
```

The same happens through the public config entry point, i.e. straight from a training YAML:

```python
from nemo.collections.asr.parts.preprocessing.perturb import process_augmentations
aug = process_augmentations({"random_segment": {"prob": 1.0, "duration_sec": 16.0, "pad_to_duration": True}})
aug.perturb(AudioSegment(np.zeros(16000 * 5, dtype=np.float32), 16000))   # same TypeError
```

```yaml
model:
  train_ds:
    augmentor:
      random_segment:
        prob: 1.0
        duration_sec: 16.0     # any float -> crash
        pad_to_duration: true
```

Controls that isolate it to the un-rounded `pad_size`:

* `RandomSegmentPerturbation(duration_sec=32, pad_to_duration=True)` (int) succeeds and returns
  512000 samples.
* `RandomSegmentPerturbation(duration_sec=16.0, pad_to_duration=True)` on 25 s of audio succeeds
  — the float is fine as long as the padding branch is not taken.

And for the documentation half:

```python
aug = process_augmentations({"random_segment": {"prob": 1.0, "duration_sec": 16}})  # the documented snippet
seg = AudioSegment(np.zeros(16000 * 5, dtype=np.float32), 16000)
aug.perturb(seg)
print(len(seg.samples))   # 80000 -- unchanged; the docs say it should be 256000
```

**Expected behavior**

A short utterance is zero-padded and a segment of `duration_sec` is returned, for both int and
float `duration_sec`; and the documented SSL config actually pads, as the surrounding text
promises.

- Environment location: Bare-metal (virtualenv)
- Method of NeMo install: from source — checkout of `main` @ `0361552a6`, run with `PYTHONPATH`
  pointing at the checkout

**Environment details**

- OS version: macOS 26.5.1 (arm64)
- PyTorch version: 2.12.0
- Python version: 3.10.18
- NumPy version: 2.2.6

**Additional context**

CPU only; no GPU, model or manifest is required to reproduce. There is currently no unit test
covering `RandomSegmentPerturbation` (`grep -rn "RandomSegment\|random_segment" tests/` → no
hits).

On the fix: `int(round(...))` is the obvious conversion and is what `AudioSegment.subsegment`
uses (`segment.py:630-631`), but it is not safe here, because `perturb` immediately calls
`subsegment(0.0, duration_sec)` and `subsegment` raises `ValueError` when `end_time` exceeds the
segment duration. Rounding to nearest can land below `duration_sec * sample_rate` and merely
trade the `TypeError` for a `ValueError` — e.g. `duration_sec=0.35` at 22050 Hz gives
`ValueError: The slice end position (0.350000 s) is out of bounds (> 0.349977 s)`. Rounding up
avoids that, and `subsegment` then trims to exactly `round(duration_sec * sample_rate)` samples:

```diff
-            pad_size = self._duration_sec * data.sample_rate - data.num_samples
+            pad_size = math.ceil(self._duration_sec * data.sample_rate) - data.num_samples
```

I have a branch with this fix, the one-line docs correction, and a `TestRandomSegmentPerturbation`
regression class in `tests/collections/asr/test_preprocessing_segment.py` (6 of its cases fail on
`main`, all pass with the fix). Happy to open a PR.

udsy19:
# What does this PR do ?

Fixes `RandomSegmentPerturbation` (`random_segment` augmentation) crashing with
`TypeError: `pad_width` must be of integral type.` whenever it has to pad a short utterance, and
corrects the SSL configuration doc that documents the padding behaviour but omits the flag that
enables it.

**Collection**: ASR

# Changelog

- `nemo/collections/asr/parts/preprocessing/perturb.py`: compute `pad_size` as an integer number
  of samples with `math.ceil`, so `AudioSegment.pad` -> `numpy.pad` receives an integral
  `pad_width`. Previously `duration_sec * sample_rate - num_samples` was a float whenever
  `duration_sec` was a float — including the constructor default `32.0` — and every short
  utterance raised `TypeError`. `math.ceil` rather than `round`/truncation is deliberate: the
  next statement is `subsegment(0.0, duration_sec)`, which raises `ValueError` if the segment
  ends up shorter than `duration_sec`, so the padded length must not round *down*. `subsegment`
  then trims to exactly `round(duration_sec * sample_rate)` samples, matching the non-padding
  branch.
- `docs/source/asr/ssl/configs.rst`: add `pad_to_duration: true` to the `random_segment` snippet.
  The surrounding text says "samples below the provided segment length will be padded", but the
  snippet omitted the flag and it defaults to `False`, so the documented recipe silently did not
  pad — which defeats its stated purpose of keeping durations uniform inside an SSL batch.
- `tests/collections/asr/test_preprocessing_segment.py`: add `TestRandomSegmentPerturbation`
  covering int and float `duration_sec`, the zero-fill placement, sample rates where
  `duration_sec * sample_rate` is not a whole number (22050 Hz), the `pad_to_duration=False`
  default, and the non-padding branch. There was previously no test for this perturbation.

# Usage

```python
import numpy as np
from nemo.collections.asr.parts.preprocessing.perturb import RandomSegmentPerturbation
from nemo.collections.asr.parts.preprocessing.segment import AudioSegment

seg = AudioSegment(np.zeros(16000 * 5, dtype=np.float32), 16000)      # 5 s
RandomSegmentPerturbation(duration_sec=32.0, pad_to_duration=True).perturb(seg)
len(seg.samples)   # before: TypeError    after: 512000
```

Equivalently, from a training config:

```yaml
model:
  train_ds:
    augmentor:
      random_segment:
        prob: 1.0
        duration_sec: 16.0
        pad_to_duration: true
```

# GitHub Actions CI

Checks run locally (macOS, CPU, Python 3.10.18, torch 2.12.0, numpy 2.2.6):

- `pre-commit run --from-ref main --to-ref HEAD` — all hooks pass (isort, black 24.10.0,
  check-case-conflict, detect-private-key, check-added-large-files).
- `pytest tests/collections/asr/test_preprocessing_segment.py -m "not pleasefixme"` — 73 passed.
- `pytest tests/collections/asr -m "not pleasefixme"` — 1402 passed, 402 skipped, 26 failed,
  12 errors. All 38 failures/errors are pre-existing on `main` and environmental on a CPU-only
  macOS box (`*_gpu` parametrisations, ONNX/TorchScript export, model-download errors); none is
  in preprocessing or perturbation code. Verified by re-running the 7 affected files with
  `perturb.py` reverted to `main`: the failure sets are byte-identical, so this change
  introduces no new failures.
- Regression check: with `perturb.py` reverted to `main` and the new tests in place, 6 of the 10
  new cases fail (`TypeError: `pad_width` must be of integral type.`); the 4 that pass are the
  int-`duration_sec`, `pad_to_duration=False` and long-audio controls. All 10 pass with the fix.
- `git diff --check` — clean.
- Not run: the Sphinx docs build. The doc change is one line inside an existing
  `.. code-block:: yaml` and adds no directive, role or reference; the snippet was parsed with
  `yaml.safe_load` to confirm it is still valid.

**Pre checks**:

- [x] Make sure you read and followed Contributor guidelines
  - No. `math` is stdlib; no new dependency and no optional import.

**PR Type**:

- [ ] New Feature
- [x] Bugfix
- [x] Documentation

# Additional Information

- Fixes #16104
- The documentation change is included here rather than split out because it is the same
  user-visible failure: on `main` the documented `random_segment` recipe does not pad, and the
  doc line would be actively misleading — it would walk readers into the `TypeError` — if landed
  without the code fix.

### DIFF
--- a/docs/source/asr/ssl/configs.rst
+++ b/docs/source/asr/ssl/configs.rst
@@ -28,6 +28,7 @@ dataset config:
     random_segment:
       prob: 1.0
       duration_sec: 16 # specify the duration you want
+      pad_to_duration: true # zero-pad utterances shorter than duration_sec (defaults to false)
 
 3) You can also use bucketing to ensure similar utterance lengths within batches.
 See :ref:`Bucketing documentation <Bucketing_Datasets>`.
--- a/nemo/collections/asr/parts/preprocessing/perturb.py
+++ b/nemo/collections/asr/parts/preprocessing/perturb.py
@@ -35,6 +35,7 @@
 import copy
 import inspect
 import io
+import math
 import os
 import random
 import subprocess
@@ -1095,7 +1096,10 @@ class RandomSegmentPerturbation(Perturbation):
                 # don't do anything if pad_to_duration is False
                 return
             start_time = 0.0
-            pad_size = self._duration_sec * data.sample_rate - data.num_samples
+            # `pad_size` must be an integer number of samples for np.pad, and `math.ceil`
+            # (rather than round/truncate) keeps the padded segment at least `duration_sec`
+            # long so that the subsegment() call below stays in bounds.
+            pad_size = math.ceil(self._duration_sec * data.sample_rate) - data.num_samples
             data.pad(pad_size=pad_size)
         else:
             start_time = random.uniform(0.0, data.duration - self._duration_sec)
--- a/tests/collections/asr/test_preprocessing_segment.py
+++ b/tests/collections/asr/test_preprocessing_segment.py
@@ -22,7 +22,12 @@ import numpy as np
 import pytest
 import soundfile as sf
 
-from nemo.collections.asr.parts.preprocessing.perturb import NoisePerturbation, ShiftPerturbation, SilencePerturbation
+from nemo.collections.asr.parts.preprocessing.perturb import (
+    NoisePerturbation,
+    RandomSegmentPerturbation,
+    ShiftPerturbation,
+    SilencePerturbation,
+)
 from nemo.collections.asr.parts.preprocessing.segment import AudioSegment, select_channels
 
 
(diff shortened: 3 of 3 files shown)


udsy19:

